PR #1226, opened 2021-09-30
Tweak blockchain-api logging

silasbw:
* Don't log to the google cloud stream unless running in GAE
* Use a pretty printer for the console stream


gnardini (approved):
Thanks!!

PR #1228, opened 2021-09-30
Use deterministic blinding factor when calling ODIS

codyborn:
### Description

[See this ticket for more details](https://app.zenhub.com/workspaces/identity-sprint-board-61015de9a849530010b49fc6/issues/celo-org/celo-monorepo/7950).

The [first PR](https://github.com/celo-org/react-native-blind-threshold-bls/pull/5) has been merged to update the React Native BLS library to use an external source of randomness. This allows Valora to use a deterministic (to the phone number) source of randomness such that calls to ODIS with the same phone number have the same blinding factor. This prevents ODIS from double counting against a user's quota when requerying the same number. This lowers the likelihood of a user running out of quota on wallet restore scenarios.

### Tested

- Tested with Android and iOS emulators.
- Verified phone number (tests saga.ts)
- Sent funds to other numbers (tests privateHashing.ts)
- Restored account and verified blinding factors are the same (with same DEK)

### How others should test

- Onboard with phone number verification
- Send payment to other phone numbers

### Related issues

- Fixes [#7950](https://app.zenhub.com/workspaces/identity-sprint-board-61015de9a849530010b49fc6/issues/celo-org/celo-monorepo/7950)

### Backwards compatibility

Is compatible with older and newer versions of ODIS.


alecps (approved):
Looks great!

gnardini (approved):
We don't have the logic for querying ODIS for phone numbers without doing the captcha yet, but this will help a lot when we do 🙌

codyborn:
@gnardini KomenciKit actually takes the BLS blinding client as input. This code change updates the BLS blinding client but still uses KomenciKit (and reCAPTCHA) during sign up. Additionally, the user needs to look up their friends' wallets using ODIS which happens in `privateHashing.ts`

PR #1229, opened 2021-10-01
Fix typo on Spanish translation

Pedro-vk:
### Description

_Typo on Spanish translation_


silasbw (approved):

silasbw:
Thanks @Pedro-vk :clinking_glasses:

jeanregisser (commented):
Thanks! We manage translations via Crowdin so we won't merge this one as it will be overwritten by the next Crowdin sync.
Instead I'll relay this change to Paula.

PR #1232, opened 2021-10-01
Get token info and balances into redux

### LINKED ISSUE #1222: Fetch user balance for all tokens
On issue #1124 we will be writing token information to a Firebase node.

On this issue we need to fetch that information when the app opens, fetch the balance the user has of each token and store it in the Redux store.

To fetch the user balance of an ERC20 token we need to use the [ERC20 abi](https://unpkg.com/browse/@uniswap/v2-periphery@1.1.0-beta.0/build/IERC20.json):
```
const contract = new kit.web3.eth.Contract(erc20Abi, tokenAddress);
const balance = await contract.methods.balanceOf(account.address).call();
```

isabellewei:
### Description

This pulls token info from firebase, finds the user's balance for each token, and then adds that info into redux for all non-zero token balances.
### Tested

tried it with test wallet on alfajores, saw cusd and celo balances in redux

### How others should test

other issues depending on this will be tested

### Related issues

- Fixes #1222

### Backwards compatibility
yes

### DIFF
--- a/packages/mobile/src/tokens/reducer.ts
+++ b/packages/mobile/src/tokens/reducer.ts
@@ -1,0 +1,45 @@
+import { RehydrateAction } from 'redux-persist'
+import { getRehydratePayload, REHYDRATE } from 'src/redux/persist-helper'
+import { Actions, ActionTypes } from 'src/tokens/actions'
+
+export interface Token {
+  address: string
+  decimals: number
+  imageUrl: string
+  name: string
+  symbol: string
+}
+
+export interface TokenBalance extends Token {
+  balance: number
+}
+
+export interface TokenBalances {
+  [address: string]: TokenBalance
+}
+
+export interface State {
+  tokenBalances: TokenBalances
+}
+
+export const initialState = {
+  tokenBalances: {},
+}
+
+export const reducer = (state: State = initialState, action: ActionTypes | RehydrateAction) => {
+  switch (action.type) {
+    case REHYDRATE: {
+      return {
+        ...state,
+        ...getRehydratePayload(action, 'tokens'),
+      }
+    }
+    case Actions.SET_TOKEN_BALANCES:
+      return {
+        ...state,
+        tokenBalances: action.balances,
+      }
+    default:
+      return state
+  }
+}
--- a/packages/mobile/src/tokens/saga.ts
+++ b/packages/mobile/src/tokens/saga.ts
@@ -1,10 +1,10 @@
 import { CeloTransactionObject } from '@celo/connect'
-import { CeloContract, StableToken } from '@celo/contractkit'
+import { CeloContract, ContractKit, StableToken } from '@celo/contractkit'
 import { GoldTokenWrapper } from '@celo/contractkit/lib/wrappers/GoldTokenWrapper'
 import { StableTokenWrapper } from '@celo/contractkit/lib/wrappers/StableTokenWrapper'
 import { retryAsync } from '@celo/utils/lib/async'
 import BigNumber from 'bignumber.js'
-import { call, put, take } from 'redux-saga/effects'
+import { call, put, select, spawn, take } from 'redux-saga/effects'
 import { showErrorOrFallback } from 'src/alert/actions'
 import { AppEvents } from 'src/analytics/Events'
 import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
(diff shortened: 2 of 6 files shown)


packages/mobile/src/tokens/saga.ts
```diff
@@ -248,3 +253,23 @@ export async function getCurrencyAddress(currency: Currency) {
       return contractKit.registry.addressFor(CeloContract.StableTokenEUR)
   }
 }
+
+export function* importTokenInfo() {
+  const tokens: Token[] = yield readOnceFromFirebase('tokensInfo')
+  let token: Token
```
gnardini, line 259: nit: why declare token here instead of doing `for (const token of tokens)` in line 263?

packages/mobile/src/tokens/saga.ts
```diff
@@ -248,3 +253,23 @@ export async function getCurrencyAddress(currency: Currency) {
...
+  const kit: ContractKit = yield call(getContractKit)
+  const address: string = yield select(currentAccountSelector)
```
gnardini, line 261: small nit: The account wording here I think might cause confusion. What do you think about copying it into a `walletAddressSelector` and marking the `currentAccountSelector` as deprecated?

packages/mobile/src/tokens/saga.ts
```diff
@@ -248,3 +253,23 @@ export async function getCurrencyAddress(currency: Currency) {
...
+  const balances: TokenBalances = {}
+  for (token of tokens) {
```
gnardini, line 263: Lets do this in parallel so its faster :D

To do that I think we might need to extract the code inside the for loop to it's own function `fetchTokenBalance` and then:
```
const balances = yield all(tokens.map((token: Token) => call(fetchTokenBalance, token))
```

Lets also add a try/catch around each individual balance fetch so that if one fails we still get the data for all the others. Maybe just leave the balance as `null` in that case.

packages/mobile/src/tokens/saga.ts
```diff
@@ -248,3 +253,23 @@ export async function getCurrencyAddress(currency: Currency) {
...
+    const contract = new kit.web3.eth.Contract(erc20.abi, token.address)
+    const balance: number = yield call(contract.methods.balanceOf(address).call)
```
gnardini, line 265: For the balance, I think it makes sense to store it in the readable format so that it can be used directly elsewhere without worrying about the decimals.

This means calling `new BigNumber(balance).dividedBy(new BigNumber(10).exponentiatedBy(token.decimals))`

packages/mobile/src/tokens/reducer.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export interface State {
+  tokenBalances: TokenBalances
+}
+
+export const initialState = {
+  tokenBalances: {},
+}
+
+export const reducer = (state: State = initialState, action: ActionTypes | RehydrateAction) => {
```
gnardini, line 29: One consideration is that we're retying to move towards [redux toolkit](https://redux-toolkit.js.org/) which is less verboragic about creating all these things. One example is the recipients reducer.

No need to change it now, but I might if I get some extra time here or there if you don't :)

gnardini (commented):
Thanks for putting this together so quickly! Looks really good, just left a couple comments :)

packages/mobile/src/tokens/saga.ts
```diff
@@ -248,3 +253,23 @@ export async function getCurrencyAddress(currency: Currency) {
       return contractKit.registry.addressFor(CeloContract.StableTokenEUR)
   }
 }
+
+export function* importTokenInfo() {
```
gnardini, line 257: It would also be nice if this function would has a few tests :D 

I think some test-cases to cover are:
- Happy path, having two tokens and seeing that their balances are fetched correctly.
- Have tokens with different number of decimals show up properly.
- If fetching one token fails, the rest are still returned and the failing one has the balance as `null`

packages/mobile/src/tokens/saga.ts
gnardini: If you need a token to test this decimals thing you can use this test token: https://test-token-faucet.web.app/

packages/mobile/src/tokens/saga.ts
isabellewei: this would be a big refactor, since `currentAccountSelector` is used in so many places. I'll open up a separate PR for that.

gnardini: Yeah, no I was not proposing changing everywhere, what I meant is:

Right now the selectors file says:
```
export const currentAccountSelector = (state: RootState) =>
  state.web3.account?.toLowerCase() ?? null
```
I suggested changing it to:
```
export const walletAddressSelector = (state: RootState) =>
  state.web3.account?.toLowerCase() ?? null
// @deprecated please use walletAddressSelector instead.
export const currentAccountSelector = walletAddressSelector
```

That way we don't need to make any big changes but we can start using a more descriptive selector name :)

What do you think?

packages/mobile/src/tokens/reducer.ts
```diff
@@ -0,0 +1,45 @@
+import { RehydrateAction } from 'redux-persist'
+import { getRehydratePayload, REHYDRATE } from 'src/redux/persist-helper'
+import { Actions, ActionTypes } from 'src/tokens/actions'
+
+export interface Token {
```
dievazqu, line 5: I didn't do it yet, but this model will contain also the current usd price in a number field. We can call it `usdPrice`. And it should be optional (because we might not have enough information to obtain a accurate price)

isabellewei: ok, added :)

PR #299, opened 2021-04-30
Tomm/flipper upate

MuckT:
### Description

Flipper will be updated from `0.69.0` to `0.87.0` to allow the use of Flipper with Xcode 12.5.  This also allows those developing on the project to use the direct download [Flipper](https://fbflipper.com/) without custom setup increasing developer accessibility for the project.

### Other changes

- Cocoapods is updated to from `1.10.0` to `1.10.1`
-  Use OpenSSL-Universal `1.1.180` instead of a custom [Headers Only Version of OpenSSL-Universal](https://github.com/celo-org/OpenSSL-headers.git)

### Tested

- [x] Tested with xCode 12.5
- [x] E2E tests run on Android.
- [x] E2E tests run on iOS
- [x] Application built in dev and release mode.
- [x] Works with latest version of [Flipper](https://fbflipper.com/) 

```Bash
# install cocoapods and bundler if you don't already have them from the /packages/mobile/ios directory
gem install cocoapods
gem install bundler

# Navigate to /packages/mobile
bundle install

# Navigate to /packages/mobile/ios
bundle exec pod install

# Flipper works with iOS
yarn dev:ios
# Flipper works with Android
yarn dev:ios
```

### Backwards compatibility

This change will require developers to download Flipper `v0.87.0` and run the following commands after pulling the latest code
```bash
# Navigate to /packages/mobile
bundle install

# Navigate to /packages/mobile/ios
bundle exec pod install
```

PR #1234, opened 2021-10-04
Fix crash on iOS device in debug mode

jeanregisser:
### Description

Fix crash on iOS device in debug mode due to incompatible OpenSSL versions between `react-native-fast-crypto` and `Flipper`.

`react-native-fast-crypto` was updated to use `OpenSSL-Universal` instead of bundling its own version.
The fact that it was previously bundling its own version was the cause of the issue. I believe this regression was introduced in #299.

This depends on https://github.com/celo-org/react-native-bip39/pull/7

### Tested

- iOS works in debug mode on a device and in the simulator.
- Android works in debug mode in an emulator.

### How others should test

N/A

### Related issues

- Discussed on [Slack](https://valora-app.slack.com/archives/C025V1D6F3J/p1633106238156700?thread_ts=1632855718.132100&cid=C025V1D6F3J).

### Backwards compatibility

Yes


silasbw (approved):
:1234:

gnardini (approved):
Thanks a lot for fixing this! 🙌

PR #1251, opened 2021-10-12
Exchanges rate

### LINKED ISSUE #1124: Calculate and store token prices
We need to be able to know from Valora the USD price of each token. We'd need a way to fetch for each token we support:
- Name
- Symbol
- Logo
- USD price
- Contract address

**Edit** Another option to what is described below is using the Ubeswap SDK to calculate prices: https://github.com/Ubeswap/ubeswap-sdk I think this one doesn't support the Moola steps (CELO-mCELO and such) not sure how much that matters, I think it could potentially give bad results since cUSD/cEUR have very low liquidity (mcUSD/mcEUR are the ones with the liquidity)

The most difficult of the above is calculating the token price and deciding which tokens to support. My suggestion for this is supporting all tokens that have at least one pool in Ubeswap. If we agree on that, we can do the following:

First, fetch all the existing Ubeswap pools. Pools are created by the [factory contract](https://github.com/Ubeswap/ubeswap/blob/master/contracts/uniswapv2/UniswapV2Factory.sol) which emits an event `PairCreated` whenever a new pair is created. The address for the contract is `0x62d5b84bE28a183aBB507E125B384122D2C25fAE`
We can interact with a contract by calling `new kit.web3.eth.Contract(abi, address)` The abi for the contract can be found [here](https://unpkg.com/browse/@uniswap/v2-periphery@1.1.0-beta.0/build/IUniswapV2Factory.json) (it's the Uniswap version but the abi is the same)

Once we have the contract we can get all `PairCreated` events with this:
```
const factory = new kit.web3.eth.Contract(abi, address)
const pairsEvents = await factory.getPastEvents('PairCreated', {
  fromBlock: '0',
  toBlock: 'latest'
})
const pairs = pairsEvents.map(pairEvent => ({
  token0: pairEvent.returnValues.token0,
  token1: pairEvent.returnValues.token1,
  pairAddress: pairEvent.returnValues.pair
}))
```
From the pair address, using the [Pair abi](https://unpkg.com/browse/@uniswap/v2-periphery@1.1.0-beta.0/build/IUniswapV2Pair.json) we can get the exchange rate between `token0` and `token1` by doing something like this ([this is the math that the contracts does which includes the protocol fee](https://github.com/Ubeswap/ubeswap/blob/master/contracts/uniswapv2/libraries/UniswapV2Library.sol#L45)):
```
const { reserve0, reserve1 } = await pairContract.methods.getReserves().call();
const numerator = 997 * reserve0;
const denominator = reserve1 * 1000 + 997;
const exchangeRate = numerator / denominator;
```

We also need to take into account that the cUSD/mcUSD cEUR/mcEUR and CELO/mCELO exchange rates are always 1 even if there is no pool for them since those can be traded 1 for 1 on Moola.

Once we have all the pools and the exchange rate between the tokens, we need to exclude the pools with very low liquidity since the exchange rates there don't mean much since any trade would fluctuate the price widely. I don't have a good suggestion to do this right now, whoever implements this issue needs to figure it out.

From the remaining pools, we create a graph and we need to find the path from each token to cUSD which provides the best exchange rate. I think we can probably just all possible combinations and keep the one with the highest liquidity?

To make all these calculations we could create a Cloud Function that runs every X (10?) minutes which stores the exchange rates in a Firebase node and Valora would read the values from there.

dievazqu:
### Description

This PR is adding a new cloud function that will run every 10 minutes.
This function will fetch the liquidity pools from Ubeswap and calculate the price of each token.
Then, it will store the prices in firebase.

### Tested

Running locally + unit tests

- Fixes #1124


tarikbellamine:
> > I haven't given this a deep review so please lmk if I am misunderstanding.
> > Should we consider maintaining a list of assets that we use as a top-of-funnel filter? Or first doing a check on liquidity for a given token before determining if it should be included in any exchange graphs? I know we are doing liquidity checks but it looks like we don't filter out those routes until AFTER we've already done all the expensive computations.
> > Listing assets on these exchanges is permission-less so the number of crappy assets can get very big very quickly. I don't think it's premature to worry about it given the exponential time and space complexity involved with this feature
> 
> @tarikbellamine Yes, having a list of assets to filter will be nice. But we will have to maintain that list. Any time a new "valid" token appears we should add it to the list. Not only because the price of that token couldn't be obtain, but also because it may vary the prices of other tokens. In the current implementation, the filtering is done one step before the most expensive computation step (which is the O(n^3) in the floyd algorithm). However, it's true that there are some part of the code that could be improved if we do that kind of filtering at the beginning. I think it's a matter of deciding if we want to manually maintain that list of not. We are currently using a list to whitelist the tokens that we are showing in the app. We can use the same list but we should take care of having that list updated.

Thanks for the response @dievazqu. I just wanted to point this out to make sure it's something you all have considered. I am OK with whichever approach you and @gnardini feel is best given the tradeoffs.

packages/cloud-functions/src/firebase.ts
```diff
@@ -31,3 +31,12 @@ export function saveTxHashProvider(address: string, txHash: string, provider: st
...
     )
 }
+
+export async function fetchFromFirebase(path: string) {
+  const snapshot = await database().ref(path).once('value')
+  return snapshot.val()
+}
+
+export async function updateFirebase(path: string, value: any) {
+  await database().ref(path).set(value)
```
gnardini, line 41: The function is called `update` but it's actually using `set` which do different things (`update` doesn't delete fields you don't send, `set` does). Lets change one of them to make it clearer :)

packages/cloud-functions/src/exchangeRate/FirebasePriceUpdater.ts
```diff
@@ -0,0 +1,88 @@
...
+  try {
+    await updatePrices()
+  } catch (err) {
+    console.error('There was an error while refreshing token prices:', (err as Error).message)
+    callCloudFunction('updateFirebasePricesByRequest', 0).catch((e) => console.error(e?.message))
+  }
+}
+
+export const updateFirebasePricesScheduled = functions.pubsub
+  .schedule('*/10 * * * *') // every 10 mins
```
gnardini, line 69: Can we change this to be every minute instead? I just don't see any harm in updating the price more often to avoid the possibilities of showing slightly wrong values :)

packages/cloud-functions/src/exchangeRate/FirebasePriceUpdater.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    const fetchTime = Date.now()
+
+    await asyncPool(
+      MAX_CONCURRENCY,
+      Object.entries(tokensInfoRaw),
+      async ([key, token]: [string, any]) => {
+        const address = token?.address?.toLowerCase()
+        if (address && prices[address]) {
+          await updateFirebase(`${FIREBASE_NODE_KEY}/${key}/usdPrice`, prices[address].toString())
```
gnardini, line 43: If we do an actual `update` instead of `set` we can do here:

```
await updateFirebase(`${FIREBASE_NODE_KEY}/${key}`, { 
  usdPrice: prices[address].toString(), 
  priceFetchedAt: fetchTime 
})
```

which updates in one go instead of making two calls :)

gnardini (commented):
Looks great, just left a small thing on using `update` vs `set` on Firebase :)

packages/cloud-functions/src/firebase.ts
dievazqu: lol, I'm sorry :(

packages/cloud-functions/src/exchangeRate/FirebasePriceUpdater.ts
dievazqu: Sure

gnardini (approved):
LGTM!

MuckT:
@dievazqu I removed the auto merge label as this branch will display multi-token features that we likely do not want in the 1.22.x release.

dievazqu:
> @dievazqu I removed the auto merge label as this branch will display multi-token features that we likely do not want in the 1.22.x release.

I think this feature can be merged anyways, it's adding prices information in Firebase but it isn't showing any of that in the app. But we can wait until the release is done if you feel more confident about it. :)

PR #1253, opened 2021-10-12
Create Token Balances screen

### LINKED ISSUE #1063: List of tokens that user holds

isabellewei:
### Description

Screen that shows balances of tokens that user holds, as well as the balances in their local currency if an exchange rate exists. Currently there is nowhere in the app that leads to this screen

### Tested

Tested on an Android device, temporarily linked the `Send` button to lead to this screen.

### How others should test

Yes

### Related issues

- Fixes #1063


isabellewei:
@gnardini I wasn't aware there were updated designs, the issue still links to the old designs. I've update the screen :)

gnardini:
@isabellewei Really sorry about that, I should have realized.

It was decided recently that there's going to be a 'tab' called 'Final' for each project and engineering should only implement what's on that one and the rest are 'Draft' pages for design to iterate. I will pay extra attention to link to only tabs which are marked as 'Final' going forward.

PR #1288, opened 2021-10-20
Fix yielding of IP address fetch

tarikbellamine:
### Description
Fix yielding of IP address fetch. We were `calling` `response.json` instead of just yielding it, resulting in an error.

### Other changes
No

### Tested
Manually

### How others should test
Ensure IP address is correct for your local device

### Related issues
N/A

### Backwards compatibility
Yes


MuckT:
@ValoraQA you should be able to test by changing the IP on the device and seeing the providers change based on IP.

PR #1294, opened 2021-10-21
Add token balances to homescreen

### LINKED ISSUE #1062: Display total balance on the home screen
Add to the home screen the total sum of all assets that user holds.
Remove balances from the side menu.
It is OK if some tokens that user holds don't have a price, just skip them.

 [the mockups](https://www.figma.com/file/8USELhisKuP1YlCt95WrD2/Balances-on-the-Home-Screen?node-id=562%3A7723) 

For now, ignore the error cases but make sure to include all the scenarios where the user holds 0-N tokens.

isabellewei:
### Description

Between the notifications and transaction list on the homescreen, the user can see their total balance of all tokens. If they only have one token, they are shown info for that one balance, otherwise they see their total balance and a link to see a breakdown of all their token balances.

### Tested

Screenshot_20211020-224641_Alfajores (dev)
Screenshot_20211020-224623_Alfajores (dev)

### How others should test

View the homescreen when the user has 0, 1 or multiple token balances

### Related issues

- Fixes #1062 

### Backwards compatibility

yes


packages/mobile/src/home/HomeTokenBalance.tsx
```diff
@@ -0,0 +1,124 @@
...
+    ValoraAnalytics.track(HomeEvents.viewTokenBalances)
+    navigate(Screens.TokenBalances)
+  }
+
+  return (
+    <View style={styles.container}>
+      <View style={styles.title}>
+        <View style={styles.row}>
+          <Text style={styles.totalValue}>{t('totalValue')}</Text>
+          {!noBalances && <InfoIcon size={14} color={Colors.gray3} />}
```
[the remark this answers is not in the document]
isabellewei: oops, added now!

packages/mobile/src/home/HomeTokenBalance.tsx
```diff
@@ -0,0 +1,124 @@
+import Colors from '@celo/react-components/styles/colors'
```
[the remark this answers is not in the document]
isabellewei: added!

isabellewei:
Also removed the `BalancesDisplay` from the sidebar!

packages/mobile/src/home/HomeTokenBalance.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+      },
+    })
+
+    const tree = render(
+      <Provider store={store}>
+        <TokenBalancesScreen {...mockScreenProps} />
+      </Provider>
+    )
+
+    expect(tree).toMatchSnapshot()
```
gnardini, line 52: Can we check the actual value that is shown? Just query for the total and check that it matches the expected value. You can do that with the `getElementText` function like this:
```
expect(getElementText(getByTestId('TestId'))).toEqual('R$15.00') // Change to actual value.
``` 

Same for the other tests here

Snapshot files are hard to read when changes are made and very easy to just overwrite so they don't provide a very good safety net :/

packages/mobile/src/home/HomeTokenBalance.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const localCurrencySymbol = useSelector(getLocalCurrencySymbol)
+  const totalBalance = useSelector(totalTokenBalanceSelector)
+  if (tokenBalances.length == 0) {
+    return (
+      <Text style={styles.balance}>
+        {localCurrencySymbol}
+        {'0.00'}
+      </Text>
+    )
+  } else if (tokenBalances.length == 1) {
```
gnardini, line 31: I know this probably works fine for this case but lets use `===` for equals checks :) The `==` is prone to false positives in many cases. Foe example, if `tokenBalances.length` returns `undefined`, `tokenBalances.length == 0` will still equal `true`.

It's just generally a good practice to use the strict equality operator `===`

packages/mobile/src/home/HomeTokenBalance.tsx
```diff
@@ -0,0 +1,146 @@
...
+    return (
+      <View style={styles.oneBalance}>
+        <Image source={{ uri: tokenBalances[0].imageUrl }} style={styles.tokenImg} />
+        <View style={styles.column}>
+          <Text style={styles.balance}>
+            {localCurrencySymbol}
+            {totalBalance}
+          </Text>
+          <Text style={styles.tokenBalance}>
+            {new BigNumber(tokenBalances[0].balance).toFixed(2)} {tokenBalances[0].name}
```
gnardini, line 41: For some tokens just two decimals son't be enough :/ For example, for BTC $10 are `0,00016 BTC`. There's a `calculateDecimalsToShow` function in `TokenDisplay` you can use to figure out how many decimals to show.

gnardini (commented):
Thanks for making the changes :) Left a couple extra comments and would you mind syncing with `isabellewei/tokenChanges` as well? There are a lot of unrelated changes in this PR atm :)

packages/mobile/src/home/HomeTokenBalance.test.tsx
isabellewei: added :)

packages/mobile/src/home/HomeTokenBalance.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+function TokenBalance() {
+  const tokenBalances = useSelector(tokensWithBalanceSelector)
+  const localCurrencySymbol = useSelector(getLocalCurrencySymbol)
+  const totalBalance = useSelector(totalTokenBalanceSelector)
+  if (tokenBalances.length === 0) {
+    return (
+      <Text style={styles.balance} testID={'TotalTokenBalance'}>
+        {localCurrencySymbol}
+        {'0.00'}
```
gnardini, line 28: The period depends on the locale, in some places (like Argentina) they use a comma instead. How about using `new BigNumber(0).toFixed(2)` instead?

packages/mobile/src/home/HomeTokenBalance.tsx
```diff
@@ -0,0 +1,147 @@
...
+    return (
+      <View style={styles.oneBalance}>
+        <Image source={{ uri: tokenBalances[0].imageUrl }} style={styles.tokenImg} />
+        <View style={styles.column}>
+          <Text style={styles.balance} testID={'TotalTokenBalance'}>
+            {localCurrencySymbol}
+            {totalBalance}
+          </Text>
+          <Text style={styles.tokenBalance}>
+            {tokenBalance.toFixed(calculateDecimalsToShow(tokenBalance))} {tokenBalances[0].name}
```
gnardini, line 42: Hey, sorry I know I mentioned this but the usage has now changed, instead of exposing the decimals to show you can now just show `formatValueToDisplay(tokenBalance)`

gnardini, line 42: Also, I think we should probably use `tokenBalances[0].symbol` instead of `name`

gnardini (commented):
Couple final things :)

packages/mobile/src/home/HomeTokenBalance.tsx
isabellewei: great suggestion!

gnardini (approved):
Great job! 🚀

PR #1340, opened 2021-10-29
Remove legacy blockchain-api code

silasbw:
This code now lives at https://github.com/valora-inc/blockchain-api


dievazqu (approved):
LGTM

gnardini (approved):
Should we add packages/blockchain-api to the root gitignore? To avoid people accidentally pushing the code again or see it in their diff when changing between branches.

Other than that looks great, approving

Issue #1126, opened 2021-09-16
Enable sending any token through send flow


aleksey-nov:
Atm only cUSD / cEUR can be send through send/request flow.
Now when we are adding support for all Celo tokens, it makes sense to also enable users sending any token they own (including Celo) through regular send flow.

PR #1342, opened 2021-10-29
Allow sending or inviting using multiple tokens

gnardini:
### Description

The title says it all :D

### Other changes

- N/A

### Tested

Manually and with unit tests

### How others should test

This is disabled by a feature flag, but we'll have instructions soon. Basically just need to send or invite users normally using different tokens and try to find issues!

### Related issues

- Part of #1126

### Backwards compatibility

N/A


packages/mobile/src/invite/saga.ts
```diff
@@ -48,59 +45,36 @@ export async function getInviteFee(
...
   feeInfo?: FeeInfo
 ) {
   try {
     ValoraAnalytics.track(InviteEvents.invite_start, {
-      escrowIncluded: true,
       amount: amount.toString(),
+      tokenAddress,
     })
 
-    const web3: Web3 = yield call(getWeb3)
```
dievazqu, line 60: I'm wondering, why this code is removed or why was it there in the first place... Just to understand it can you give me more context? :)

packages/mobile/src/send/saga.ts
```diff
@@ -188,17 +200,124 @@ function* sendPayment(
...
+) {
+  const contract: Contract = yield call(getERC20TokenContract, tokenAddress)
+  const coreContract: CeloTokenWrapper<any> | undefined = yield call(
+    getTokenContractFromAddress,
+    tokenAddress
+  )
+  const convertedAmount: string = yield call(tokenAmountInWei, amount, tokenAddress)
+
+  const kit: ContractKit = yield call(getContractKit)
+  return coreContract
```
dievazqu, line 228: What does it mean that `coreContract` is `undefined`?

dievazqu (approved):
I left a couple of questions to understand a bit more about the context, but the code seems fine.

packages/mobile/src/invite/saga.ts
gnardini: Yes, so this code hasn't been used for a long time.

Originally, before we had Komenci, you needed to pay $0.15 to verify your phone number. So, if you wanted to invite someone you were asked to pay $0.15. The funds were put in a temporary account and an "invite code" was created. This invite code had access to the private key of the temporary address.

In the onboarding flow there was a screen where you could put in the code to get the $0.15 transferred from the temporary wallet to your new one you just created.

This stopped being used late last year when Komenci was implemented to pay for the verification fees, so that's the deleted code here :)

packages/mobile/src/send/saga.ts
gnardini: There is a `coreContract` for cUSD, cEUR and CELO. These core contracts have the `transferWithComment` function that the normal ERC20 interface doesn't.

So, what we do here is: if there is a core contract, call `transferWithComment` on it, otherwise (for all other tokens) just do a normal `transfer` without any comment.

packages/mobile/src/escrow/saga.test.ts
```diff
@@ -12,9 +12,14 @@ import {
...
-import { mockAccount, mockE164Number, mockE164NumberHash, mockE164NumberPepper } from 'test/values'
+import {
+  mockAccount,
+  mockCusdAddress,
+  mockE164Number,
+  mockE164NumberHash,
+  mockE164NumberPepper,
+} from 'test/values'
 
 describe(transferToEscrow, () => {
```
jeanregisser, line 24: Looks like we're missing a happy path test for this feature 😢

packages/mobile/src/send/SendConfirmation.test.tsx
```diff
@@ -239,9 +240,38 @@ describe('SendConfirmation', () => {
   it('renders correct modal for invitations', async () => {
     const { getByTestId, queryAllByTestId } = renderScreen({}, mockInviteScreenProps)
 
-    expect(queryAllByTestId('InviteAndSendModal')[0].props.visible).toBe(false)
-    // Fire event press not working here so instead we call the onClick directly
-    getByTestId('ConfirmButton').props.onClick()
```
jeanregisser, line 244: Why removing? it checked the modal wasn't present before clicking, which seems is still relevant.

packages/mobile/src/tokens/saga.ts
```diff
@@ -328,6 +328,17 @@ export function* calculateTotalTokenBalance() {
   yield put(setTotalTokenBalance(totalBalance.toFixed(2).toString()))
 }
 
+export function* tokenAmountInWei(amount: BigNumber, tokenAddress: string) {
```
jeanregisser, line 331: Shouldn't this be called `tokenAmountInSmallestUnit` or something?
`1 wei` is really only `10^-18 eth`, right?

packages/mobile/src/tokens/saga.test.ts
```diff
@@ -54,3 +56,33 @@ describe(importTokenInfo, () => {
...
+            tokenBalances: {
+              [mockAddress]: {
+                address: mockAddress,
+                decimals: 5,
+              },
+            },
+          },
+        }).getState()
+      )
+      .returns((10e5).toFixed(0))
```
jeanregisser, line 77: nit: usually better to avoid intermediary representations in expectations
```suggestion
      .returns('1000000')
```

jeanregisser (approved):
Looks great! 🎉 
Left some small comments but already approved.

packages/mobile/src/send/SendConfirmation.test.tsx
gnardini: Yup, I don't know what I was thinking, I may have deleted it by mistake

packages/mobile/src/escrow/saga.test.ts
gnardini: Added!

ValoraQA:
Hey @MuckT, We have verified this issue using Latest iOS Alfajores Nightly V1.23.0, Android Alfa jores nightly V1.23.0 and added the result in below sheet.

**Sheet Link:**
https://docs.google.com/spreadsheets/d/1LcYXmQcuLTBIHZrzUjoYaUvnzisq2ReRvxVwmn-xGEA/edit#gid=0

**Devices:** iPhone 13 mini (15.1.1), One plus 7t (11.0)
CC: @gnardini 
Thanks..!!

PR #1346, opened 2021-11-01
Make token picker on bottom sheet scrollable

gnardini:
### Description

The token picker is now scrollable.

https://user-images.githubusercontent.com/6062888/139738067-b5aad11c-5723-4c3f-bc9a-20a73784bcf6.mov

### Other changes

Fixed a warning due to a missing `key` attribute.

### Tested

I made a few copies of the tokens I had to test with a large token list.

### How others should test

Get a lot of different tokens and check that they can all be seen from the token picker

### Related issues

- Part of #1126

### Backwards compatibility

N/A


packages/mobile/src/components/TokenBottomSheet.tsx
```diff
@@ -102,6 +102,9 @@ function TokenBottomSheet({ isVisible, origin, onTokenSelected, onClose }: Props
     setPickerHeight(height)
   }
 
+  const maxHeight = Dimensions.get('window').height - 100
```
silasbw, line 105: can you pull up the `100` into a module-level or function-level variable with a descriptive name?

packages/mobile/src/components/TokenBottomSheet.tsx
```diff
@@ -111,24 +114,21 @@ function TokenBottomSheet({ isVisible, origin, onTokenSelected, onClose }: Props
...
       </Animated.View>
-      <Animated.View
-        style={[
-          styles.contentContainer,
-          { paddingBottom: Math.max(safeAreaInsets.bottom, Spacing.Thick24) },
-          animatedPickerPosition,
-        ]}
+      <Animated.ScrollView
+        style={[styles.contentContainer, { paddingBottom, maxHeight }, animatedPickerPosition]}
+        contentContainerStyle={{ paddingBottom: pickerHeight >= maxHeight ? 50 : 0 }}
```
silasbw, line 119: same with the `50`?

silasbw (commented):
nit to help me understand the code better.

if `100` and `50` are well-defined values everyone should know, let me know!

packages/mobile/src/components/TokenBottomSheet.tsx
```diff
@@ -111,24 +114,21 @@ function TokenBottomSheet({ isVisible, origin, onTokenSelected, onClose }: Props
           testID={'BackgroundTouchable'}
         />
       </Animated.View>
-      <Animated.View
-        style={[
-          styles.contentContainer,
-          { paddingBottom: Math.max(safeAreaInsets.bottom, Spacing.Thick24) },
-          animatedPickerPosition,
-        ]}
+      <Animated.ScrollView
```
dievazqu, line 117: Is it possible to add a test for this new behavior?

dievazqu (approved):
It looks ok! It would be nice to test the scrollable behavior somehow...

packages/mobile/src/components/TokenBottomSheet.tsx
gnardini: I don't see how, this is a UI thing. I imagine with an e2e test, but Detox tests are hard to write for this kind of thing I think.

I'll keep this in mind to consider when adding e2e tests for this flow.
cc @MuckT

MuckT: We can add a e2e test to ensure it's scrollable, but testing one component isn't ideal for e2e tests. Alternatively, we might be able to test this with [React Native Testing Library](https://callstack.github.io/react-native-testing-library/docs/api/#fireeventscroll-element-reacttestinstance-data-arrayany--void) which seems more appropriate if possible.

ValoraQA:
Hey @MuckT, We have verified this issue using Latest iOS Alfajores Nightly V1.23.0, Android Alfa jores nightly V1.23.0 and added the result in below sheet.

**Sheet Link:**
https://docs.google.com/spreadsheets/d/1LcYXmQcuLTBIHZrzUjoYaUvnzisq2ReRvxVwmn-xGEA/edit#gid=1458277593

**Devices:** iPhone 13 mini (15.1.1), One plus 7t (11.0)
CC: @gnardini
Thanks..!!

PR #1347, opened 2021-11-01
Increased required code coverage to 75%

MuckT:
### Description

Currently code coverage is around 75.06% with a target of 70%. This PR will set the target to 75%.

### Other changes

Pull requests will require a minimum of 75% total coverage to pass status checks.

### Tested

Target observed on checks.

### How others should test

N/A

### Backwards compatibility

This change is backwards compatible.


jeanregisser (approved):
👍

PR #1348, opened 2021-11-01
Refactor token selectors

isabellewei:
### Description

- `totalTokenBalance` is no longer a redux state, instead `totalTokenBalancesSelector` just directly calculates it
- get rid of duplicated `tokenBalancesSelector` in the reducer

### Tested

Yes

### How others should test
View the token balances screen


packages/mobile/src/tokens/selectors.ts
```diff
@@ -60,3 +61,23 @@ export const defaultTokenSelector = createSelector(tokensListSelector, (tokens)
 
   return maxTokenAddress
 })
+
+export const totalTokenBalanceSelector = createSelector(
```
gnardini, line 65: Can you add a test for this selector? I think there should be some examples in the selectors.test file :)

packages/mobile/src/tokens/selectors.ts
```diff
@@ -60,3 +61,23 @@ export const defaultTokenSelector = createSelector(tokensListSelector, (tokens)
...
+    let totalBalance = new BigNumber(0)
+
+    for (const token of tokenBalances) {
+      const tokenAmount = new BigNumber(token.balance)
+        .multipliedBy(token.usdPrice)
+        .multipliedBy(usdRate)
+      totalBalance = totalBalance.plus(tokenAmount)
+    }
+
+    return totalBalance.toFixed(2).toString()
```
gnardini, line 81: I can't think of any case where we might want to use the total as a `BigNumber` instead of a string, but I think exposing this as a `BigNumber` gives us more flexibility in the future in case such a use case does pop up. This is an optional comment though.

packages/mobile/src/tokens/selectors.ts
isabellewei: added!

gnardini (approved):
Great work! 🚀

PR #1357, opened 2021-11-02
Fix address on SendConfirmation screen when going through SecureSend

gnardini:
### Description

Going through SecureSend with the new multi-token support had a bug where it looked like an invite instead of a send.

### Other changes

N/A

### Tested

Manually and with unit test

### How others should test

Go through SecureSend using the new flow and see that you can send funds normally.

### Related issues

- Part of #1126

### Backwards compatibility

N/A


packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -89,6 +95,44 @@ function SendConfirmation(props: Props) {
...
+      )
+      return {
+        ...paramRecipient,
+        // Setting it explicitly so Typescript doesn't complain
+        e164PhoneNumber: paramRecipient.e164PhoneNumber,
+        address: recipientAddress ?? undefined,
+      }
+    }
+    return paramRecipient
+  }, [paramRecipient])
```
jeanregisser, line 125: Feels a bit weird to do this here, should we handle this somewhere else?

gnardini: I moved it to a hook, but I'm not sure if that's what you had in mind 🤔

PR #1359, opened 2021-11-03
WIP adding analytics event for when a notification is sent

cajubelt:
NOTE this may not work work! (need a valid way to import the analytics event dependencies...)

### Description

Publishing an event when a notification is sent.

I'm playing around with analytics events to learn about how they work and how to add a new one. We have an event for when someone opens a notification, but not for when one is sent, which is important because you can't always rely on a user opening a notification-- sometimes they'll just see it and come into the app "organically" as a result.

### Other changes

na

### Tested

### How others should test

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

### DIFF
--- a/packages/cloud-functions/src/notifications/index.ts
+++ b/packages/cloud-functions/src/notifications/index.ts
@@ -87,6 +87,7 @@ export async function sendNotification(
     ValoraAnalytics.track(AppEvents.push_notification_opened, {
       id: data.id,
       type: data.type,
+      address,
     })
   } catch (error) {
     console.error(
--- a/packages/cloud-functions/src/notifications/index.ts
+++ b/packages/cloud-functions/src/notifications/index.ts
@@ -2,6 +2,8 @@ import * as admin from 'firebase-admin'
 import i18next, { TFunction } from 'i18next'
 import { NOTIFICATIONS_TTL_MS } from '../config'
 import { database, messaging } from '../firebase'
+import ValoraAnalytics from '@celo/mobile/src/analytics/ValoraAnalytics'
+import { AppEvents } from '@celo/mobile/src/analytics/Events'
 
 const TAG = 'Notifications'
 
@@ -82,6 +84,10 @@ export async function sendNotification(
         latency: data.timestamp ? Date.now() - Number(data.timestamp) : null,
       })
     )
+    ValoraAnalytics.track(AppEvents.push_notification_opened, {
+      id: data.id,
+      type: data.type,
+    })
   } catch (error) {
     console.error(
       JSON.stringify({
--- a/packages/cloud-functions/src/notifications/index.ts
+++ b/packages/cloud-functions/src/notifications/index.ts
@@ -84,7 +84,7 @@ export async function sendNotification(
         latency: data.timestamp ? Date.now() - Number(data.timestamp) : null,
       })
     )
-    ValoraAnalytics.track(AppEvents.push_notification_opened, {
+    ValoraAnalytics.track(AppEvents.push_notification_sent, {
       id: data.id,
       type: data.type,
       address,
--- a/packages/mobile/src/analytics/Events.tsx
+++ b/packages/mobile/src/analytics/Events.tsx
@@ -11,6 +11,7 @@ export enum AppEvents {
   fetch_balance_error = 'fetch_balance_error',
   redux_keychain_mismatch = 'redux_keychain_mismatch',
   redux_store_recovery_success = 'redux_store_recovery_success',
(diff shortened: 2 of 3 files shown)


packages/cloud-functions/src/notifications/index.ts
```diff
@@ -2,6 +2,8 @@ import * as admin from 'firebase-admin'
 import i18next, { TFunction } from 'i18next'
 import { NOTIFICATIONS_TTL_MS } from '../config'
 import { database, messaging } from '../firebase'
+import ValoraAnalytics from '@celo/mobile/src/analytics/ValoraAnalytics'
+import { AppEvents } from '@celo/mobile/src/analytics/Events'
```
cajubelt, line 6: Is this a valid way of importing from another package in the monorepo?

jeanregisser: We should avoid doing this. This will make it harder to split cloud-functions out of the wallet monorepo.

Also ValoraAnalytics is a client only implementation which has dependencies on React Native libs. So won't work at runtime in a cloud function environment.

Do we really want/need to share this code between client and server?
I'd favor splitting client and server side tracking because they have different constraints and dependencies.

packages/cloud-functions/src/notifications/index.ts
```diff
@@ -82,6 +84,11 @@ export async function sendNotification(
         latency: data.timestamp ? Date.now() - Number(data.timestamp) : null,
       })
     )
+    ValoraAnalytics.track(AppEvents.push_notification_opened, {
```
jeanregisser, line 87:
```suggestion
    ValoraAnalytics.track(AppEvents.push_notification_sent, {
```

packages/cloud-functions/src/notifications/index.ts
gnardini: I think for this to work from Cloud Functions you need to push the dependency to npm and import it here. Otherwise it will only work locally but not when you deploy :/ That's a google limitation with yarn workspaces I think (or used to be a few months ago when I ran into this).

In any case, I don't think this is what we want. `ValoraAnalytics` is tailored to the mobile environment and it holds all the setup data for that use case. Stuff like mobile screen size, OS, memory available, app version, user address and a bunch of other data. It would be quite complicated to make that one generic enough for it to be used on our backend services.

For our backend services we have two options:

- In the `cloud-functions` directory we have a `src/bigQuery` file with a `trackEvent` function which pushes directly to BigQuery. This is the legacy way of doing it and it has a few drawbacks like not having some fields added automatically (like timestamp) and having to create and maintain the table on BigQuery manually. Usually Segment does this for us which takes me to:
- In the `notification-service` directory we have a dependency called `analytics-node`. This one pushes directly to Segment. You can think of Segment as a middleman, it receives events from various sources, makes some changes (mainly adds fields) and sends the events to targets. One of the possible targets is BigQuery, where it creates the necessary tables and adds columns if new ones are added. Very convenient. There's example usage in that service.

Here's a screenshot of what Segment looks like: I just sent you an invite as well
Screen Shot 2021-11-03 at 02 32 17

To clarify on the underlying goal of this PR which is to add events when push notifications are sent, we are indeed sending events when a payment push notification is sent, which happens in the notification service. What we are missing is sending the event for payment request push notifications, which is what is being done here in this cloud function.

One final note is that we're using the `analytics-node` library directly but it might make sense to create a wrapper library to customize it a bit more for our usage. Not a priority, but a nice to have at some point imo.

cajubelt: Thanks for the in depth explanations @jeanregisser @gnardini !!

Ok, it sounds like this PR was misguided on multiple levels lol. I just opened it to start learning about how we manage analytics events though, it's not an urgent change by any means. I'll close it for now and try again with a more limited-scope change, like a tracking event in the `mobile` package.

That being said, it does seem like the `events` and `properties` could benefit from a single source of truth being available somewhere, to prevent things like collisions between analytics events being published from different services.

And having a helper library for publishing tracking events to segment/bigquery sounds potentially useful, in case we ever wanted to switch out those dependencies.

cajubelt: I'll create GH issues for those 2 higher-level improvement ideas and tag you both on them.

cajubelt: cool, here is the issue @jeanregisser @gnardini . Feel free to comment or edit. https://app.zenhub.com/workspaces/acquisition-squad-sprint-board-6010683afabec1001a090887/issues/valora-inc/wallet/1369

PR #1363, opened 2021-11-03
Switch to Valora owned KMS keys

jeanregisser:
### Description

The repo secrets were still using a KMS key from `celo-testnet`.
It's now using `celo-mobile-mainnet` and `celo-mobile-alfajores`.

Note: this is just a quick change so we don't rely on `celo-testnet` which is owned by cLabs. Actual goal is to use Google's Secret Manager instead, but requires a bit more work.

See https://valora-app.slack.com/archives/C025V1D6F3J/p1635811586043200


silasbw (approved):

silasbw:
:hot_pepper: :hot_pepper: :hot_pepper: :hot_pepper: :hot_pepper: :hot_pepper: :hot_pepper: :hot_pepper: :hot_pepper: :hot_pepper: :hot_pepper:

Issue #305, opened 2021-05-03
[Analytics] Update list of parameters to track on every event, when possible


aleksey-nov:
**What's important about this?**
Adding additional parameters that we track on every event fired in the Valora app will allow us significantly improve the depth of analysis and segmenting our user base. This will allow us to make more data-driven decisions in product development.

**What is done?**
Where possible append all of the parameters listed below to every analytics event:

**userAddress** : waller address of user in lower case
**language** : send in the form it is available in.
**CELOBalance** : send the float value of the celo balance upto three digits
**cUSDBalance** : send the float value of the cusd balance upto three digits
**cEURBalance** : send the float value of the ceur balance upto three digits
**country** : Send the three digit (Alpha-2 code) of the countries based on phone country code as usa, gbr or arg
**fiatCurrency** : Default currency of the app customer in lower case eur or usd or gbp
**fiatValue** : Total fiat value of the celo assets held
**screenName** : send the name of the screen where the interaction is happening in lower case
**flowType** : send the type of the screen where the interaction is happening. screen to flow type mapping is here
**phoneVerified** : yes or no mentioning whether a user has verified their phone
**accountBackupCompleted** : yes or no mentioning whether a user has verified their address.
**otherTenTokens** : Send the top ten tokens by balance other than CELO, cUSD, cEUR in the format ```{{tokencode1}}:{{tokenvalue1}},{{tokencode2}}:{{tokenvalue2}},{{tokencode3}}:{{tokenvalue3}},{{tokencode4}}:{{tokenvalue4}} ``` so on

If in any case, the parameter is not available in some cases, then the best option is to send them as undefined.

**What's involved in doing this work?**
Adding parameters listed above. Sometimes it is not possible to add them simply because they are not yet available for some reason. For example, during onborading it is not possible to add parameters like address, celo_balance, cusd_balace, ceur_balance, etc.

grvalora:
Hi @aleksey-nov , 

I have refined the parameters, we should be good to proceed with the following. Please review and all is fine update the ticket above with the parameters mentioned here please. 

**address** : waller address of user in lower case
**language** : send the two digit ISO language code of the user in lower case as en or es or pt
**celoBalance** : send the float value of the celo balance
**cusdBalance** : send the float value of the cusd balance
**ceurBalance** : send the float value of the ceur balance
**country** : Send the three digit (Alpha-3 code) of the countries based on phone country code as **usa, gbr or arg**
[...]

PR #1383, opened 2021-11-05
Fix some require cycles

jeanregisser:
### Description

Fix some require cycles which were causing issues for the work I'm doing for #305
- between account/reducer and backup/utils
- src/identity/reducer.ts -> src/identity/verification.ts -> src/app/selectors.ts -> src/identity/reducer.ts

Require cycles can create subtle issues, we should aim to remove all of them to avoid surprises in the future.

Before this change, the RN packager was listing 89 require cycle warnings.
After this change, we're down to 80 require cycle warnings.


gnardini (approved):

silasbw (approved):
:100:

Issue #1085, opened 2021-09-07
[analytics] Optimising the fetch_balance event conditions on segment


grvalora:
**Context :**
Currently fetch_balance event triggered by segment is firing thrice for each celo asset and the calls are being made not early in the stages of app being opened.

**Fix needed:**
1. fetch_balance event to trigger for app open and as early as possible.
2. trigger fetch_balance event only one and containing details about all assets.
3. include balance support for other CELO assets
4. add fiatCurrency and fiatValue to the existing parameters

**Fix needed:**
Please find the list of events and parameters here

1. fetch_balance : event to be updates as per the details here in the sheet link : https://docs.google.com/spreadsheets/d/1XbfyL1vN7RpYWA5d5uxuLB5AyOpazdNA-7CCuRsFins/edit#gid=0

2. home_all_balances

3. balance_screen_loaded

Screen shot here
![image.png](https://images.zenhubusercontent.com/610c15c3d406bbe75388a732/d20f055e-bd42-4442-86f2-3970e802f1db)

ACs
* Identify the value held by Users for each token. 
* DApp user segmentation 

**Current logs Info** 
[...]

PR #1386, opened 2021-11-05
Fetch token balances using Blockscout

gnardini:
### Description

Before this PR when fetching balances we were querying the balance directly from the token contracts.
This has the advantage that it's querying directly from Forno or other nodes that are more likely to be up to date than blockscout, which sometimes has a delay to update balances.

However, the disadvantage was that we were making one request per each token. If there are not a lot of tokens this is fine, but there are 42 tokens already on https://github.com/Ubeswap/default-token-list/blob/master/ubeswap.token-list.json and this number will continue to grow. Making so many requests each time we want to fetch balances looks like too much.

Another possible solution longer-term is indexing and updating the balances of our users ourselves using the indexer. This would give us control over the latency for block updates while not having to make so many requests at the same time.

### Other changes

Created an action which re-fetches the token balances and called it in the places where the old ones used to be called.

### Tested

Manually and with unit tests.

### How others should test

Nothing to test directly here, just make sure the balances show up and are up to date after making a transfer.
### Related issues

- Part of #1126

### Backwards compatibility

N/A

### DIFF
--- a/packages/mobile/.env.alfajores
+++ b/packages/mobile/.env.alfajores
@@ -19,3 +19,4 @@ GETH_USE_FULL_NODE_DISCOVERY=true
 GETH_USE_STATIC_NODES=true
 # Only for development, use with caution
 GETH_START_HTTP_RPC_SERVER=false
+BLOCKSCOUT_BASE_URL=https://alfajores-blockscout.celo-testnet.org/api
--- a/packages/mobile/src/escrow/saga.ts
+++ b/packages/mobile/src/escrow/saga.ts
@@ -41,7 +41,7 @@ import { VerificationStatus } from 'src/identity/types'
 import { NUM_ATTESTATIONS_REQUIRED } from 'src/identity/verification'
 import { navigateHome } from 'src/navigator/NavigationService'
 import { fetchStableBalances } from 'src/stableToken/actions'
-import { TokenBalance } from 'src/tokens/reducer'
+import { fetchTokenBalances, TokenBalance } from 'src/tokens/reducer'
 import {
   getCurrencyAddress,
   getERC20TokenContract,
@@ -324,6 +324,7 @@ function* withdrawFromEscrow(komenciActive: boolean = false) {
     }
 
     yield put(fetchStableBalances())
+    yield put(fetchTokenBalances())
     Logger.showMessage(i18n.t('inviteFlow11:transferDollarsToAccount'))
     ValoraAnalytics.track(OnboardingEvents.escrow_redeem_complete)
   } catch (e) {
--- a/packages/mobile/src/home/saga.ts
+++ b/packages/mobile/src/home/saga.ts
@@ -20,6 +20,7 @@ import { shouldFetchCurrentRate } from 'src/localCurrency/selectors'
 import { withTimeout } from 'src/redux/sagas-helpers'
 import { shouldUpdateBalance } from 'src/redux/selectors'
 import { fetchStableBalances } from 'src/stableToken/actions'
+import { fetchTokenBalances } from 'src/tokens/reducer'
 import { Actions as TransactionActions } from 'src/transactions/actions'
 import Logger from 'src/utils/Logger'
 import { getConnectedAccount } from 'src/web3/saga'
@@ -42,6 +43,7 @@ export function withLoading<Fn extends (...args: any[]) => any>(fn: Fn, ...args:
 export function* refreshBalances() {
   Logger.debug(TAG, 'Fetching all balances')
   yield call(getConnectedAccount)
+  yield put(fetchTokenBalances())
   yield put(fetchStableBalances())
   yield put(fetchGoldBalance())
   yield put(fetchSentEscrowPayments())
--- a/packages/mobile/src/tokens/saga.test.ts
+++ b/packages/mobile/src/tokens/saga.test.ts
@@ -2,16 +2,19 @@ import BigNumber from 'bignumber.js'
 import { expectSaga } from 'redux-saga-test-plan'
 import { call, select } from 'redux-saga/effects'
 import { readOnceFromFirebase } from 'src/firebase/firebase'
-import { setTokenBalances, StoredTokenBalance } from 'src/tokens/reducer'
-import { getERC20TokenBalance, importTokenInfo, tokenAmountInSmallestUnit } from 'src/tokens/saga'
+import { setTokenBalances, StoredTokenBalances } from 'src/tokens/reducer'
+import { fetchTokenBalances, importTokenInfo, tokenAmountInSmallestUnit } from 'src/tokens/saga'
 import { walletAddressSelector } from 'src/web3/selectors'
 import { createMockStore } from 'test/utils'
-import { mockAccount, mockTokenBalances, mockTokenBalances2 } from 'test/values'
+import { mockAccount, mockTokenBalances } from 'test/values'
 
-const firebaseTokenInfo: StoredTokenBalance[] = [
-  {
+const poofAddress = '0x00400FcbF0816bebB94654259de7273f4A05c762'
+const cUsdAddress = '0x874069Fa1Eb16D44d622F2e0Ca25eeA172369bC1'
+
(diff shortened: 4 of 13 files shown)


packages/mobile/src/tokens/saga.ts
```diff
@@ -266,44 +269,58 @@ export async function getERC20TokenContract(tokenAddress: string) {
...
-  const balance: number | null = yield call(getERC20TokenBalance, token, address)
-  return {
-    ...token,
-    balance:
-      balance || balance === 0
-        ? new BigNumber(balance)
-            .dividedBy(new BigNumber(10).exponentiatedBy(token.decimals))
-            .toString()
-        : null,
+export function* fetchTokenBalancesFromBlockscout(address: string) {
```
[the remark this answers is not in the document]
dievazqu: Also, it would be nice to test this function by really calling blockscout api. So we can be sure that the returned model is the one that we are expecting. Maybe we can use `alfajores` api, and use a fixed testing address?
Not sure if that is the best approach, but I think having a e2e test for this api call adds some value :)

dievazqu (approved):
LGTM!

jeanregisser (changes requested):
I understand the underlying need for this, but this is at odds with users of the light client.

I think we should sync as a team on the full implication of this before merging.

jeanregisser:
- Could we avoid this change by using batch requests?
- If we were to go with fetching the balance from blockscout, what do you think of abstracting it through `blockchain-api`?

gnardini:
I agree with routing through blockchain-api, I'll make that change.

I don't see how to batch requests using contractKit unfortunately and since there could be a large number of tokens getting the balance for all of them client-side doesn't seem very data-efficient :/

packages/mobile/src/tokens/saga.ts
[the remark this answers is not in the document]
gnardini: > Can we add that this returns a BlockscoutTokenBalance?

We can't actually :( These generator functions have typing issues, that's why you see errors with red underlines for all `yield` usages except if you put in the type manually, which isn't super safe of course. I think there was talk in the past to fix this, but we never ended up prioritizing it I think. cc @jeanregisser 

re: calling the real blockscout api I think that's where e2e tests come in. We usually mock these things in unit tests and have an e2e test that tests the entire flow, there's an action item pending which is adding e2e tests for this that will call the actual API and test everything, does that sound like enough to you?

gnardini:
@jeanregisser Here's the PR on blockchain-api: https://github.com/valora-inc/blockchain-api/pull/42 I updated this one to use it as well, it requires it to be deployed to work

packages/mobile/.env.alfajores
```diff
@@ -19,3 +19,4 @@ GETH_USE_FULL_NODE_DISCOVERY=true
 GETH_USE_STATIC_NODES=true
 # Only for development, use with caution
 GETH_START_HTTP_RPC_SERVER=false
+BLOCKSCOUT_BASE_URL=https://alfajores-blockscout.celo-testnet.org/api
```
jeanregisser, line 22: We can remove these now that we go through `blockchain-api`.

packages/mobile/src/escrow/saga.ts
```diff
@@ -324,6 +324,7 @@ function* withdrawFromEscrow(komenciActive: boolean = false) {
     }
 
     yield put(fetchStableBalances())
```
jeanregisser, line 326: Do we still need thise?

packages/mobile/src/home/saga.ts
```diff
@@ -42,6 +43,7 @@ export function withLoading<Fn extends (...args: any[]) => any>(fn: Fn, ...args:
 export function* refreshBalances() {
   Logger.debug(TAG, 'Fetching all balances')
   yield call(getConnectedAccount)
+  yield put(fetchTokenBalances())
   yield put(fetchStableBalances())
   yield put(fetchGoldBalance())
   yield put(fetchSentEscrowPayments())
```
jeanregisser, line 49: Hopefully one day we can have a single action 😅

packages/mobile/src/tokens/saga.test.ts
```diff
@@ -29,6 +49,20 @@ const firebaseTokenInfo: StoredTokenBalance[] = [
     decimals: 18,
     balance: null,
   },
+}
+
+const fetchBalancesResponse = [
+  {
+    tokenAddress: poofAddress,
+    balance: (5 * Math.pow(10, 18)).toString(),
```
jeanregisser, line 57: nit: feel free to skip ;)
```suggestion
    balance: (5 * 10**18).toString(),
```

packages/mobile/src/transactions/saga.ts
```diff
@@ -115,6 +116,8 @@ export function* sendAndMonitorTransaction<T>(
     if (STABLE_CURRENCIES.some((stableCurrency) => balancesAffected.has(stableCurrency))) {
       yield put(fetchStableBalances())
     }
```
jeanregisser, line 118: Should we remove this?

packages/mobile/src/tokens/saga.ts
```diff
@@ -266,44 +270,69 @@ export async function getERC20TokenContract(tokenAddress: string) {
...
-  const balances: StoredTokenBalances = {}
-  for (const tokenBalance of fetchedTokenBalances) {
-    balances[tokenBalance.address] = tokenBalance
+  try {
+    // In e2e environment we use a static token list since we can't access Firebase.
+    const tokens: StoredTokenBalances = isE2EEnv
+      ? e2eTokens()
+      : yield call(readOnceFromFirebase, 'tokensInfo')
+    const address: string = yield select(walletAddressSelector)
+    const tokenBalances: FetchedTokenBalance[] = yield call(fetchTokenBalancesForAddress, address)
```
jeanregisser, line 315: Can we add new analytics events for this?
We were discussing this with @grvalora for #1085
I'm now thinking we should add a new event name for this new multi token context, maybe `fetch_token_balances`, `fetch_token_balances_error`?
Can be done in a separate PR though. But is important to ship along with this feature otherwise gonna miss some data.
Though of course #1409 may be enough.

Well, let's sync to make sure we're on the same page?

packages/mobile/src/tokens/saga.ts
```diff
@@ -266,44 +270,69 @@ export async function getERC20TokenContract(tokenAddress: string) {
...
+      address,
+    },
+    fetchPolicy: 'network-only',
+    errorPolicy: 'all',
+  })
+
+  return response.data.userBalances.balances
+}
+
 export function* importTokenInfo() {
```
jeanregisser, line 308: Should we rename this?
```suggestion
export function* fetchTokenBalances() {
```

jeanregisser (commented):
Great!

packages/mobile/src/tokens/saga.ts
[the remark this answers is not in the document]
jeanregisser: Yes we want to use `redux-typed-saga`.
By default vscode uses its own TypeScript version, but we can switch to our project version by clicking `{}` next to TypeScript in the status bar. This avoid seeing all the errors.

jeanregisser (approved):

packages/mobile/src/escrow/saga.ts
gnardini: We do for places that use the old balance, like the Celo tab or the add/withdraw screen :/

packages/mobile/src/home/saga.ts
gnardini: 😅

packages/mobile/src/transactions/saga.ts
gnardini: Like I said before, eventually yes, but while we still have the balances in the `stableToken` reducer we need it to be up to date :/

packages/mobile/src/tokens/saga.ts
gnardini: I added a `fetch_balance` event and verified that the updated balances are included per the changes on #1409

PR #1388, opened 2021-11-05
Add feature flags to use for some multi-token support features

### LINKED ISSUE #1387: Add feature flags for multi-token support
Add 3:
- multiTokenShowHomeBalances: If enabled, the total balance will be shown in the home screen, including all supported tokens.
- multiTokenUseSendFlow: If enabled, the send flow will use the new flow that supports multiple tokens except for payment requests
- multiTokenUseUpdatedFeed: If enabled, the feed will use the new feed using the new endpoints.

gnardini:
### Description

Added three feature flags to remote config:
- multiTokenShowHomeBalances: If enabled, the total balance will be shown in the home screen, including all supported tokens.
- multiTokenUseSendFlow: If enabled, the send flow will use the new flow that supports multiple tokens except for payment requests
- multiTokenUseUpdatedFeed: If enabled, the feed will use the new feed using the new endpoints.

For now only `multiTokenUseSendFlow` works, `multiTokenShowHomeBalances` needs the changes in #1294 and `multiTokenUseUpdatedFeed` is still a WIP

### Other changes

Changed the firebase remote config fetch time to 0 since we are already waiting an hour to call that function from outside.

### Tested

Manually

### How others should test

No need for QA to test this specifically I think since they don't have access to Firebase but to test you need to change the value of the feature flag on Firebase and see that the new features are seen when they are set to true.

### Related issues

- Fixes #1387

### Backwards compatibility

N/A

### DIFF
--- a/packages/mobile/test/schemas.ts
+++ b/packages/mobile/test/schemas.ts
@@ -792,8 +792,16 @@ export const v19Schema = {
     },
     totalBalance: '40',
   },
+}
+
+export const v20Schema = {
+  ...v19Schema,
+  _persist: {
+    ...v19Schema._persist,
+    version: 20,
+  },
   app: {
-    ...v18Schema.app,
+    ...v19Schema.app,
     multiTokenShowHomeBalances: false,
     multiTokenUseSendFlow: false,
     multiTokenUseUpdatedFeed: false,
@@ -801,5 +809,5 @@ export const v19Schema = {
 }
 
 export function getLatestSchema(): Partial<RootState> {
-  return v19Schema as Partial<RootState>
+  return v20Schema as Partial<RootState>
 }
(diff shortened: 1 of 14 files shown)


packages/mobile/test/schemas.ts
```diff
@@ -792,6 +792,12 @@ export const v19Schema = {
     },
     totalBalance: '40',
   },
+  app: {
+    ...v18Schema.app,
+    multiTokenShowHomeBalances: false,
+    multiTokenUseSendFlow: false,
+    multiTokenUseUpdatedFeed: false,
+  },
```
jeanregisser, line 800: nit: I'd prefer we bumped the schema because these properties are new.
As a rule we should increase the schema version every time we deviate from last schema published in an official release.
This avoids making wrong assumption for migrations as to what properties were part of a specific release.

jeanregisser (commented):
Great!

jeanregisser (approved):

Issue #1341, opened 2021-10-29
After Send Flow while transaction is pending an 'Your balance is too low' banner briefly shows incorrectly.


jh2oman:
### Current behavior

After clicking 'Send' in the Send/Request flow while the transaction is pending a 'Your balance is too low' alert briefly shows which seems very confusing. Despite this alert, the transaction processes fine and shows so in the Home view. 

It seems to happen when you try to send an amount > 50% of the total in your current wallet.  

I think the UI state is updating the wallets current funds before the transaction is actually processed which is triggering the alert that the user doesn't have enough funds to make the transaction.

**In this scenario i had $1.90 cUSD and was sending $1 to my Mom**
![IMG_9944](https://user-images.githubusercontent.com/8432644/139501396-352907e7-bd75-40f3-bdc5-66ea45ef2454.PNG)

### Desired behavior

The alert should not show when for this scenario.
### Steps to Reproduce

Send over 50% of the funds in your wallet to somebody and see the alert flash when you are redirected to the home view
### Versions

iPhone 12 Pro

ValoraQA:
Hey @jh2oman We have verfied this issue using Android internal & Test Flight build V1.21.0 7 observed that we are able to repro this issue.
**Verified On Devices:**  One Plus 7T (11.0), iPhone 12 (15), iPhone 6+ (12.4.5), Redmi Note 8 (9.0)
**Attachment:** [ErrorForValidAmount](https://drive.google.com/file/d/17k99UrcWI9IBWLAQRqllsjAHYYIuIlM1/view?usp=sharing)

PR #1389, opened 2021-11-05
Fix SendConfirmation screen insufficient funds bug

jh2oman:
### Description

Load the balance only once for the SendConfirmationScreen so that it doesn't pick up any balance changes from completed transactions and potentially throw an error.

### Tested

I couldn't come up with an easy way to add a unit test for this bug, suggestions welcome! 

Manually test by following the repro steps on https://github.com/valora-inc/wallet/issues/1341 and seeing that the bug no longer repros. 

### Related issues

- Fixes # https://github.com/valora-inc/wallet/issues/1341

### Backwards compatibility

The balance is never refreshed during this flow anyways, so there should be no unintended side effects from only loading it once.

### DIFF
--- a/packages/mobile/src/send/SendConfirmationLegacy.tsx
+++ b/packages/mobile/src/send/SendConfirmationLegacy.tsx
@@ -131,7 +131,8 @@ function SendConfirmationLegacy(props: Props) {
   )
   const account = useSelector(currentAccountSelector)
   const isSending = useSelector(isSendingSelector)
-  const balance = useBalance(currency)
+  // Only load the balance once to prevent race conditions with transactions updating balance
+  const [balance, _] = useState(useBalance(currency))
   const celoBalance = useBalance(Currency.Celo)
   const appConnected = useSelector(isAppConnected)
   const isDekRegistered = useSelector(isDekRegisteredSelector) ?? false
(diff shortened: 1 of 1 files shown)


gnardini (approved):
Small nit, looks great, thanks for debugging and fixing this!

I don't think it's necessary to add a unit test for this because this code will go away soon, but if you did want to test this you would need to call `rerender` I think, there are a couple of examples in the codebase.

ValoraQA:
Hi @MuckT We have verified the issue on latest Android Internal build V:1.22.0 and latest iOS Test Flight build V:1.22.0 and updated the observation in Gsheet 

Link : https://docs.google.com/spreadsheets/d/1UVmYHl_g7ZWKqX3zBtc5OdO9wEmdAdj4PW9zgwJxUys/edit#gid=603719806
Note : We are not able to complete send flow as error is seen while performing send flow 

New Bug Link : https://github.com/valora-inc/wallet/issues/1438

CC: @jh2oman

ValoraQA:
Hi @MuckT We have verified the issue on the latest Android Internal build V:1.22.0 and iOS Test Flight build V:1.22.0 and updated the observation in Gsheet

**Link :** https://docs.google.com/spreadsheets/d/1UVmYHl_g7ZWKqX3zBtc5OdO9wEmdAdj4PW9zgwJxUys/edit#gid=603719806

**Devices :** Samsung Galaxy J5 Prime (8.0), iPhone 12(15), Redmi Note 8 (10.0)

CC: @jh2oman

ValoraQA:
Hi @MuckT we have verified the above issue in the latest iOS Test Flight build V: 1.22.0 and updated status in Test Quality Tool
Link :  https://valora.testquality.com/project/12128/plan/22712/run/152953/result/2573241
Verified on devices : iPhone SE (13.6.1)
CC: @jh2oman

PR #1390, opened 2021-11-05
deleting RaiseLimitScreen UI

lisaqyli:
This is a pr to delete raise limit screen from help menu


jeanregisser:
Cool, but is there more context as to why?

PR #1391, opened 2021-11-05
Analytics: add more user properties

jeanregisser:
### Description

This adds more user properties (think user profile in CleverTap or Firebase).
This allows creating campaigns targeting users with these new properties (for instance users with cEUR balance >= X, etc).

### Tested

Added new unit tests.

### How others should test

- Create or restore an account
- CleverTap profile for the account should contain the new properties

### Related issues

- Part of #305

### Backwards compatibility

Yes


packages/mobile/src/analytics/selectors.ts
```diff
@@ -0,0 +1,79 @@
...
+    const currencyAddresses = new Set(
+      Object.values(tokensByCurrency).map((token) => token?.address)
+    )
+
+    // Don't rename these unless you have a really good reason!
+    // They are used in users analytics profiles + super properties
+    return {
+      accountAddress,
+      walletAddress,
+      currency, // TODO: is this the right name in multi token context?
```
gnardini, line 55: What is property meant to be used for? The `defaultCurrencySelector` was just used to decide the initial selected token in the legacy send flow but will go away soon probably. Can we just remove this?

packages/mobile/src/analytics/selectors.ts
```diff
@@ -0,0 +1,79 @@
...
+      name,
```
gnardini, line 56: Is `name` considered PII? Why do we want this in our analytics?

packages/mobile/src/analytics/selectors.ts
```diff
@@ -0,0 +1,79 @@
...
+      countryCodeAlpha2,
+      language,
+      totalBalanceUsd,
+      tokenCount: tokensByUsdBalance.length,
+      otherTenTokens: tokensByUsdBalance
+        .filter((token) => !currencyAddresses.has(token.address))
+        .slice(0, 10)
```
gnardini, line 63: Why are we limiting to 10? Is there a limit to the size of a string allowed?

gnardini (commented):
Left a couple questions, but great to see this! We've needed it for a while now 😄

packages/mobile/src/analytics/selectors.ts
jeanregisser: Good question, name is indeed considered PII.
I copied what we were already sending. It's a good time to check this.

packages/mobile/src/analytics/selectors.ts
jeanregisser: There's always a limit, but should be high enough.
I'll check with @grvalora.

packages/mobile/src/analytics/selectors.ts
jeanregisser: I copied what we were already sending.
It was useful to target the "default" currency used by others, but now in a multi token context I'm not sure anymore.

packages/mobile/src/analytics/selectors.ts
```diff
@@ -0,0 +1,85 @@
...
+  totalTokenBalanceSelector,
+} from 'src/tokens/selectors'
+import { Currency } from 'src/utils/currencies'
+import { accountAddressSelector, walletAddressSelector } from 'src/web3/selectors'
+
+export const getCurrentUserTraits = createSelector(
+  [
+    walletAddressSelector,
+    accountAddressSelector,
+    defaultCurrencySelector,
```
gnardini, line 22: I think this `defaultCurrencySelector` isn't used anymore?

gnardini (approved):
Amazing! 🚀

PR #1393, opened 2021-11-05
Dont allow writing comments for non-core tokens

gnardini:
### Description

Comments are only supported when sending (not inviting) using cUSD, cEUR or CELO. You can no longer write a comment if you're inviting someone or sending any other token.

### Other changes

N/A

### Tested

Manually and with unit tests

### How others should test

Using the new token transfer flow:
- Try sending cUSD, cEUR and CELO and see that you can write a comment.
- Try inviting someone with any token or sending any other token and see that there's no commend field.

### Related issues

- Part of #1126

### Backwards compatibility

N/A


packages/mobile/src/tokens/utils.test.tsx
```diff
@@ -62,3 +73,39 @@ describe(getHigherBalanceCurrency, () => {
...
+    const isCoreToken = useIsCoreToken(tokenAddress)
+
+    return (
+      <View>
+        <Text testID="isCoreToken">{isCoreToken}</Text>
+      </View>
+    )
+  }
+
+  it('cUSD, cEUR and CELO are core tokens', async () => {
```
dievazqu, line 88: Why can't you just check like this?
```
[mockCusdAddress, mockCeurAddress, mockCeloAddress].each((token) => expect(useIsCoreToken(token).toBeTruthy())
```

dievazqu: Oh, I see, is it because you need the `store={createMockStore()}`?
Can't we just mock `useSelector(tokensListSelector)`? 
I mean, it's ok but it seems a lot of extra code to test `useIsCoreToken`

dievazqu (approved):
LGTM!
The test of utils seems a bit odd, but maybe that's the only way to do it.

jeanregisser (commented):
It seems we may want this "coreToken" attribute to be part of the remote list so we can update it easily. cBRL is just around the corner for instance.
Also in that case I feel we could name it differently: "canTransferWithComment" (or something) to more accurately represent what it can do.
What do you think?

packages/mobile/src/tokens/utils.test.tsx
gnardini: It's because testing hooks is a PITA :( I agree it's not great.

You can't just call a hook, you must do it inside a React component. So the only way to test a hook (that I found) is to create a fake component and print the result in the screen

gnardini: I deleted the hook so these tests don't exist anymore :)

packages/mobile/src/tokens/utils.ts
```diff
@@ -19,3 +21,10 @@ export function getHigherBalanceCurrency(
   }
   return maxCurrency
 }
+
+export function useIsCoreToken(tokenAddress: string) {
+  const tokens = useSelector(tokensListSelector)
+  const token = tokens.find((token) => token.address === tokenAddress)
+  if (!token) return false
+  return ['cUSD', 'cEUR', 'CELO'].includes(token.symbol)
```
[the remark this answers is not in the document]
jeanregisser: yep sounds good to me, as long as we document somewhere what a coreToken is 😄

packages/mobile/src/tokens/reducer.ts
```diff
@@ -9,6 +9,7 @@ interface BaseToken {
   imageUrl: string
   name: string
   symbol: string
+  isCoreToken?: boolean
```
jeanregisser, line 12: Could you add a comment about what this attribute does?

jeanregisser (approved):
Awesome! 👍

ValoraQA:
Hey @MuckT We have verified this issue using Android internal build and iOS test flight build V:1.22.1 & updated observation in Gsheet
**Gsheet Link:** https://docs.google.com/spreadsheets/d/1UVmYHl_g7ZWKqX3zBtc5OdO9wEmdAdj4PW9zgwJxUys/edit#gid=572111532&range=P2
**Verified On Devices:** Samsung Galaxy F41(11.0), iPhone 12(15.0)
CC: @gnardini 
Thanks..!!

PR #1394, opened 2021-11-08
Reference correct Valora gcloud projects and tables in CICO webhooks

jophish:
### Description

This PR updates our CICO webhooks to point to the Valora BigQuery projects for mainnet/alfajoress, as well as the correct datasets. I checked where `src/bigQuery.ts` was used, and I believe this should not have any side-effects re: other, non-CICO data.

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

We'll need to update our analytics dashboards to point to these new datasources once the old data has been migrated over from cLabs to Valora.

### DIFF
--- a/packages/cloud-functions/src/bigQuery.ts
+++ b/packages/cloud-functions/src/bigQuery.ts
@@ -1,9 +1,9 @@
 import { BigQuery } from '@google-cloud/bigquery'
 
 const gcloudProject = process.env.GCLOUD_PROJECT
-export const bigQueryProjectId = 'celo-testnet-production'
-export const bigQueryDataset =
-  gcloudProject === 'celo-mobile-alfajores' ? 'mobile_wallet_dev' : 'mobile_wallet_production'
+export const bigQueryProjectId = gcloudProject
+export const bigQueryExternalDataset = 'external_data_stream' // Dataset used for storing events sourced from external providers
+export const bigQueryInternalDataset = bigQueryProjectId === 'celo-mobile-alfajores' ? 'mobile_wallet_dev' : 'mobile_wallet_production'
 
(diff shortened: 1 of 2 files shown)


packages/cloud-functions/src/bigQuery.ts
```diff
@@ -1,9 +1,9 @@
 import { BigQuery } from '@google-cloud/bigquery'
 
 const gcloudProject = process.env.GCLOUD_PROJECT
-export const bigQueryProjectId = 'celo-testnet-production'
-export const bigQueryDataset =
-  gcloudProject === 'celo-mobile-alfajores' ? 'mobile_wallet_dev' : 'mobile_wallet_production'
+export const bigQueryProjectId = gcloudProject
+export const bigQueryExternalDataset = 'external_data_stream' // Dataset used for storing events sourced from external providers
```
jophish, line 5: We're moving the provider-sourced webhook events to a new `external_data_stream` dataset, though we still need to access the `{mobile_wallet_production, mobile_wallet_dev}.app_launched` tables (see below file).

silasbw (approved):

gnardini (commented):
So this will mean we will have some data on `celo-testnet-production`, some on `celo-mobile-alfajores` and some on `celo-mobile-mainnet` right?

This data is only used for dashboards/analytics right now, right? Will this data be available on Metabase after the change?


Somewhat related (no need to change things right now, but maybe at some point in the future), for analytics in general we are using segment instead of BigQuery directly, which routes to wherever we configure it to, including BigQuery. There is an example in the `notification-service` on how it works using the lib `analytics-node`. Routing through Segment means that we can easily send the information elsewhere if we need it as well, for example CleverTap for marketing efforts

PR #1397, opened 2021-11-08
Hide displaying tokens in homescreen behind a feature flag

isabellewei:
### Description

Using firebase feature flags, if `multiTokenShowHomeBalances` is true, the new multitoken UI is showed in the homescreen, otherwise the old `BalancesDisplay` is shown in the sidebar.

This is branched off of both #1294 and #1388 

### Tested

not tested

### How others should test

refer to #1388


packages/mobile/src/home/WalletHome.tsx
```diff
@@ -46,6 +47,7 @@ interface StateProps {
   numberVerified: boolean
   cashInButtonExpEnabled: boolean
   balances: Balances
+  showTokensinHome: boolean
```
gnardini, line 50: nit: typo `showTokensinHome` -> `showTokensInHome`

packages/mobile/src/home/WalletHome.tsx
```diff
@@ -174,6 +177,13 @@ export class WalletHome extends React.Component<Props, State> {
       renderItem: () => <NotificationBox key={'NotificationBox'} />,
     })
 
+    if (this.props.showTokensinHome) {
```
gnardini, line 180: I think this is repeated, we should remove the one right below this :)

gnardini (commented):
Looks great! Just left a small issue, might have been a merge issue :)

packages/mobile/src/home/WalletHome.tsx
isabellewei: hm not sure what you mean here they are the same haha

gnardini: It's the capitalization in the i showTokens **i** nHome vs showTokens **I** nHome

isabellewei: oooo I see

noisypigeon:
@gnardini @isabellewei seems there's still an issue with this PR.

gnardini (approved):

PR #1404, opened 2021-11-09
Show message on failed token fetches

### LINKED ISSUE #1125: Handle error situations when displaying tokens
Once #1062 is done, we need to handle errors when fetching the balance of a token fails.
The most common one is probably being offline, but we should also handle just random failures on one of the calls to fetch balances, probably add some sort of retrying mechanism.

See the [mockups here](https://www.figma.com/file/8USELhisKuP1YlCt95WrD2/Balances-on-the-Home-Screen?node-id=562%3A7723) and ask the design or product teams if you have any questions.

isabellewei:
### Description

If there are issues fetching token balances from blockscout, a banner prompting the user to refresh their app is shown. It comes up both in the homescreen and token list screen, and can be dismissed when the user clicked the `Refresh` CTA

### Other changes
Create a function that allows you to fetch data from firebase with a timeout
### Tested

on an Android physical device
Screenshot_20211109-120123_Alfajores (dev)

### How others should test

Set the testing device to Airplane mode to block connectivity, then refresh the app.

### Related issues

- Fixes #1125 

### Backwards compatibility

Yes


packages/mobile/src/home/WalletHome.tsx
```diff
@@ -115,6 +123,20 @@ export class WalletHome extends React.Component<Props, State> {
     // Waiting 1/2 sec before triggering to allow
     // rest of feed to load unencumbered
     setTimeout(this.tryImportContacts, 500)
+    if (tokenBalancesError || tokenBalancesLoading) {
```
gnardini, line 126: Not sure I understand this, this is only being run on `componentDidMount`, not on update? And we're showing the error even if the token fetching is loading, I see it all the time when opening the app.

Also, the banner in the top is hiding the hamburger menu, and if you dismiss it there is no way of refreshing again other than restarting the app. Did Nitya sign off on this?

isabellewei:
@gnardini Thanks for the feedback! The app was getting stuck in the token `loading` state because the fetch from firebase was hanging. I modified the fetch from firebase with a timeout so that it would return an error. 

Also showing the out of sync banner on all refreshes of the homescreen now

PR #1406, opened 2021-11-09
Add ODIS circuit breaker implementation

### LINKED ISSUE #1396: Implement ODIS Circuit breaker Firebase function
More context [here](https://clabsco.notion.site/ODIS-Circuit-Breaker-ccd2411af34e4d9e898d04e5a485d3d3).

Essentially, this ticket covers implementing and deploying a new Firebase function with two methods.

1. Method accepting a ciphertext string that performs asymmetric decryption on it using an HSM-based key from KMS and returns the plaintext.
2. A healthcheck method.

This ticket also covers creating/deploying the HSM-based keys in both out mainnet and alfajores GCPs.

AC:
* HSM keys created and added to GCPs
* Cloud functions deployed to mainnet and alfajores

jophish:
### Description

Refer [here](https://clabsco.notion.site/ODIS-Circuit-Breaker-ccd2411af34e4d9e898d04e5a485d3d3) for more information.

This PR adds an implementation for the ODIS circuit breaker. HSM keys have been deployed to mainnet/alfajores GCPs as well. Circuit breaker operation can be performed by temporarily disabling or permanently destroying the decryption keys. Once this PR is approved, I'll deploy the cloud functions in both mainnet/alfajores.

### Related issues

- Fixes #1396


packages/cloud-functions/src/circuitBreaker/circuitBreaker.ts
```diff
@@ -0,0 +1,89 @@
...
+app.get('/health', async (req, res) => {
+  return res.status(200).json({
+    status: 'ok',
+  })
+})
+
+app.get('/status', async (req, res) => {
+  const keyState = await getKeyState()
+  switch (keyState) {
+    case 'ENABLED':
```
jophish, line 29: switching on string literals directly is obviously not ideal; not sure how to wrangle the `CryptoKeyVersionState` object to get this to work as desired, however.

packages/cloud-functions/src/circuitBreaker/circuitBreaker.ts
```diff
@@ -0,0 +1,89 @@
...
+      return res.status(200).json({
+        status: 'temporarily disabled',
+      })
+    case 'DESTROYED':
+      return res.status(200).json({
+        status: 'permanently disabled',
+      })
+    default:
+      console.error(`Unexpected key state: ${keyState}`)
+      return res.status(200).json({
```
silasbw, line 43: 200 -> 400?

packages/cloud-functions/src/circuitBreaker/circuitBreaker.ts
```diff
@@ -0,0 +1,89 @@
...
+        status: 'unknown',
+      })
+  }
+})
+
+app.post('/unwrapKey', async (req, res) => {
```
silasbw, line 49: fwiw, i think uppercase in path names is weird because the handling of case is inconsistent across the broader "internet toolchain".  i'd suggest using kebab style:

```ts
app.post('/unwrap-key', async (req, res) => {
```

silasbw (commented):
Worth adding testing?  I'm more concerned about a potential problem popping up and someone who is on-call trying to fix it, but not having tests to validate expected behavior against.

packages/cloud-functions/src/circuitBreaker/circuitBreaker.ts
jophish: d'oh, yes, i agree with this 100%

packages/cloud-functions/src/circuitBreaker/circuitBreaker.ts
jophish: not really a 400 though (imo); the request itself is fine, we're just getting an unexpected state for the key internally (i.e., a dev probably manually changed it to something other than `enabled`, `disabled`, or `destroyed`.)

jophish:
@silasbw definitely worth adding tests, was planning on it. may have been better to submit this pr in a draft state; primarily just wanted some feedback on the design itself.

packages/cloud-functions/src/circuitBreaker/circuitBreaker.ts
silasbw: ah, makes sense.

packages/cloud-functions/src/circuitBreaker/circuitBreaker.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+})
+
+app.post('/unwrap-key', async (req, res) => {
+  const ciphertext = req.body?.ciphertext
+  if (!ciphertext) {
+    return res.status(400).json({
+      error: '"ciphertext" parameter must be provided',
+    })
+  }
```
nategraf, line 55: Input `ciphertext` should be an encoded buffer, and it seems implied here that it is `base64`. I'd recommend we make this explicit and check upfront that this is the case, especially because `Buffer.from(buffer, 'base64')` will output an empty buffer if the input is not valid base64 (I know, wtf).
```suggestion
  const BASE64_REGEXP = /^(?:[A-Za-z0-9+\/]{4})*(?:[A-Za-z0-9+\/]{2}==|[A-Za-z0-9+\/]{3}=)?$/
  const ciphertext = req.body?.ciphertext
  if (!ciphertext) {
    return res.status(400).json({
      error: '"ciphertext" parameter must be provided',
    })
  } else if (!BASE64_REGEXP.test(ciphertext)) {
    return res.status(400).json({
      error: '"ciphertext" parameter must be a base64 encoded buffer',
    })
  }
```

packages/cloud-functions/src/circuitBreaker/circuitBreaker.ts
```diff
@@ -0,0 +1,88 @@
...
+      } catch (error) {
+        console.error(`Error while decrypting ciphertext: ${error.message}`)
+        return res.status(500).json({
+          error: 'Error while decrypting ciphertext',
+        })
+      }
+
+      const plaintext = decryptResponse[0].plaintext as string
+      return res.status(200).json({
+        plaintext: plaintext.toString(),
```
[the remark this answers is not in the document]
nategraf: My understanding is based on https://github.com/googleapis/nodejs-kms/blob/main/samples/decryptAsymmetric.js

packages/cloud-functions/src/circuitBreaker/circuitBreaker.ts
```diff
@@ -0,0 +1,88 @@
...
+    case 'ENABLED':
+      let decryptResponse
+      try {
+        decryptResponse = await client.asymmetricDecrypt({
+          name: CIRCUIT_BREAKER_KEY_PATH,
+          ciphertext,
+        })
+      } catch (error) {
+        console.error(`Error while decrypting ciphertext: ${error.message}`)
+        return res.status(500).json({
```
nategraf, line 68: This may actually be a `400` error, if the `ciphertext` is mangled, but I don't see any obvious way to determine what caused the decryption failure, so returning `500` here seems fine to me.

packages/cloud-functions/src/circuitBreaker/circuitBreaker.test.ts
```diff
@@ -0,0 +1,106 @@
+import request from 'supertest'
```
nategraf, line 1: Nice tests 👌

nategraf (approved):
Nice! I left a few comments, but the implementation looks good to me overall.

I am currently working on an encrypted backup library, and so will be adding a circuit breaker client to the `@celo/identity` package as part of that work. I can tell from the code what the input and output should be, and I can match that in the client. I'll plan on adding so TS definitions as part of that for the circuit breaker request and response objects. I commented on defining `ciphertext` and `plaintext` as `base64` buffers. The other thing that would be helpful is to have a string enum with error codes that I can check as part of the client library to bubble up failure reasons.

packages/cloud-functions/src/circuitBreaker/circuitBreaker.ts
jophish: Yeah, not sure there's a simple way to figure that out; agreed that a 500 seems fine here.

packages/cloud-functions/src/circuitBreaker/circuitBreaker.ts
jophish: Yeah, it should be `base64` encoded. Good catch on adding an explicit test for that. Will add that in! (also definitely wtf re: `Buffer.from` behavior lol)

packages/cloud-functions/src/circuitBreaker/circuitBreaker.ts
```diff
@@ -0,0 +1,88 @@
...
+        })
+      }
+
+      const plaintext = decryptResponse[0].plaintext as string
+      return res.status(200).json({
+        plaintext: plaintext.toString(),
+      })
+
+    case 'DISABLED':
+      return res.status(503).send()
```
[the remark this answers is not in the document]
jophish: Maybe returning a json response with e.g. `error: "DISABLED"` and `message: "Service is temporarily disabled"` to make it easier for clients to match against the `error` field? Good suggestion though.

packages/cloud-functions/src/circuitBreaker/circuitBreaker.ts
```diff
@@ -0,0 +1,88 @@
...
+    case 'DESTROYED':
+      return res.status(410).send()
```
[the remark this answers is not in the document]
jophish: Yeah I agree; will change to 503 and add in an error code as discussed.

packages/cloud-functions/src/circuitBreaker/circuitBreaker.ts
```diff
@@ -0,0 +1,88 @@
...
+    default:
+      console.error(`Unexpected key state: ${keyState}`)
+      return res.status(200).json({
+        status: 'unknown',
+      })
+  }
+})
+
+app.post('/unwrap-key', async (req, res) => {
+  const ciphertext = req.body?.ciphertext
```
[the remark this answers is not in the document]
jophish: yep, base64

packages/cloud-functions/src/circuitBreaker/circuitBreaker.ts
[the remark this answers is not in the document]
jophish: Yeah, it is indeed a Buffer, although the official documentation says otherwise; coercing it to a Buffer now so we can call return a b64 encoded string.

packages/cloud-functions/src/circuitBreaker/circuitBreaker.ts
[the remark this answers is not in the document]
jophish: Actually, opted for a simpler implementation so we return similar objects in the `/status` and `/unwrap-key` routes for consistency.

nategraf (approved):
Thanks for addressing my comments. This seems solid!

silasbw (approved):

PR #1409, opened 2021-11-10
Analytics: add super properties

jeanregisser:
### Description

This PR adds super properties to all tracked events.
They are prefixed with `s` to avoid clashing with events props.

Note: this requires sending more data to Segment and other services. 
This isn't a concern with Segment since events are [batched and compressed](https://segment.com/blog/lifecycle-of-a-mobile-message/). 
I haven't yet checked with Firebase and CleverTap though.

### Other changes

- Removed `setUserAddress` call which was a bit redundant with `identify`

### Tested

Added new unit tests.

### How others should test

N/A

### Related issues

- Part of #305

### Backwards compatibility

Yes


noisypigeon (approved):
LGTM but I would have another person take a look!

gnardini (approved):

PR #1411, opened 2021-11-10
Fee estimation refactor

gnardini:
### Description

Refactored how fee estimation works and added the fee value to the total amount we show in the confirmation screen.

### Other changes

Uncommented a few `.only` that snuck up in previous PRs 🤦‍♂️ 

### Tested

Manually looking at the logs to see that the estimations are fired at the right time and with unit tests.

### How others should test

Go through the new send flow and see that the tx fees are estimated correctly.

### Related issues

- Part of #1126

### Backwards compatibility

N/A


packages/mobile/src/fees/reducer.ts
```diff
@@ -1,51 +1,82 @@
...
-export const reducer = combineReducers({ estimates: estimatesReducer })
+const rehydrate = createAction<any>(REHYDRATE)
+export const estimateFee = createAction<EstimateFeeAction>('FEES/ESTIMATE_FEE')
+export const feeEstimated = createAction<{
+  tokenAddress: string
+  feeType: FeeType
+  estimation: FeeEstimateState
+}>('FEES/FEE_ESTIMATED')
+
+export const reducer = createReducer(initialState, (builder) => {
```
jeanregisser, line 47: Fine for now, but in the future we should favor `createSlice` instead.
See https://github.com/valora-inc/wallet/issues/1442

packages/mobile/src/fees/reducer.ts
```diff
@@ -1,51 +1,82 @@
...
+  builder
+    .addCase(rehydrate, (state, action) => {
+      // hack to allow rehydrate actions here
+      const hydrated = getRehydratePayload((action as unknown) as RehydrateAction, 'fees')
+      return {
+        ...state,
+        ...hydrated,
+      }
+    })
```
jeanregisser, line 56: This shouldn't be needed. All reducers are rehydrated by default.

packages/mobile/src/redux/store.ts
```diff
@@ -19,10 +19,10 @@ let lastEventTime = Date.now()
 
 const persistConfig: any = {
   key: 'root',
-  version: 20, // default is -1, increment as we make migrations
+  version: 21, // default is -1, increment as we make migrations
   keyPrefix: `reduxStore-`, // the redux-persist default is `persist:` which doesn't work with some file systems.
   storage: FSStorage(),
-  blacklist: ['geth', 'networkInfo', 'alert', 'fees', 'imports'],
```
jeanregisser, line 25: Why should we persist fees now?

packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -263,7 +263,7 @@ function SendConfirmation(props: Props) {
         confirmButton={{
           action: onSendClick,
           text: isInvite ? t('sendAndInvite') : t('send'),
-          disabled: isSending || !!feeError,
```
jeanregisser, line 266: Should we really move forward if we couldn't calculate the fee?

packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -170,17 +166,21 @@ function SendConfirmation(props: Props) {
         <FeeDrawer
           testID={'feeDrawer/SendConfirmation'}
           isEstimate={true}
-          currency={feeInfo?.currency}
+          currency={Currency.Dollar}
```
jeanregisser, line 169: Shouldn't this be `feeCurrency`?

packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -242,7 +242,7 @@ function SendConfirmation(props: Props) {
         usdAmount,
         comment,
         recipient,
-        feeInfo,
+        feeEstimate?.feeInfo,
```
jeanregisser, line 245: I think we shouldn't move forward if `feeInfo` is undefined.
Otherwise we could risk sending a transaction where the actual fee hasn't been approved by the user.
or am I missing something?

packages/mobile/src/send/SendConfirmation.test.tsx
```diff
@@ -124,42 +123,47 @@ describe('SendConfirmation', () => {
...
 
-    // NOTE: CELO fees are currently not combined into the total.
-    // TODO: This should equal more than $1.33, depending on the CELO fee value.
+    // Subtotal is $1.33, which is added the fee amount.
     const totalComponent = getByTestId('TotalLineItem/Total')
-    expect(getElementText(totalComponent)).toEqual('$1.33')
+    expect(getElementText(totalComponent)).toEqual('$1.36')
   })
 
-  it('shows a generic `calculateFeeFailed` error when fee estimate fails due to an unknown error', async () => {
```
jeanregisser, line 135: Do you think showing `---` is enough for users to understand there was an error?

packages/mobile/src/send/SendAmountLegacy/SendAmount.test.tsx
```diff
@@ -144,7 +148,7 @@ describe('SendAmountLegacy', () => {
           dismissAfter: null,
           displayMethod: ErrorDisplayType.BANNER,
           message: i18n.t('needMoreFundsToSend', {
-            amountNeeded: '106.9852',
```
jeanregisser, line 147: Why do we need more now?

packages/mobile/src/fees/saga.test.ts
```diff
@@ -1,79 +1,230 @@
...
+            error: true,
+            usdFee: null,
+            lastUpdated: Date.now(),
+          },
+        })
       )
       .run()
   })
 
-  it("doesn't calculates fee if the balance is zero", async () => {
```
jeanregisser, line 70: How does it handle the 0 balance case?

packages/mobile/src/fees/saga.ts
```diff
@@ -23,88 +37,183 @@ export interface FeeInfo {
   currency: Currency
 }
 
-// TODO(victor): This fee caching mechansim is only being used by the balance check on the send
-// amount entry screen. In an effort to standardize and improve fee estimation, we should either
-// update and use this mechanism everywhere or remove it in favor of another solution.
-// Cache of the gas estimates for common tx types
-// Prevents us from having to recreate txs and estimate their gas each time
-const feeGasCache = new Map<FeeType, BigNumber>()
```
jeanregisser, line 31: Are we still caching fees?

jeanregisser (commented):
Great work! 🎉 
From what I see it fixes a bunch of edge cases we had. 

And I have a bunch of questions now 🙂

packages/mobile/src/components/TokenTotalLineItem.test.tsx
```diff
@@ -20,11 +20,13 @@ describe('TokenTotalLineItem', () => {
     tokenAddress = defaultTokenAddress,
     localCurrencyCode = LocalCurrencyCode.BRL,
     localCurrencyExchangeRate = '1.5',
+    feeToAddInUsd = undefined,
```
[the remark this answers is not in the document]
gnardini: Oh, I missed this, thanks!

packages/mobile/src/fees/reducer.ts
gnardini: Forgot about this 😓  Will keep in mind

packages/mobile/src/redux/store.ts
gnardini: I don't know why we weren't doing it before. The stored fee has a `lastUpdated` field that is used to refresh the fee after some time passed, I don't see why restarting the app should change that. I'm ok adding it back though if you think we shouldn't persist it, no strong opinion on my side.

packages/mobile/src/send/SendConfirmation.tsx
gnardini: Well, I changed this thinking of my own personal case, I would prefer to be able to send the tx even if fee estimation fails, but I guess that might not be true to everyone and we shouldn't change it without discussing, I'll revert this

packages/mobile/src/send/SendConfirmation.tsx
gnardini: No, you're correct, see my other comment, fixed!

packages/mobile/src/send/SendConfirmation.test.tsx
gnardini: Not really, no. When fee estimation fails the user also gets an error thing on the top saying it failed, but that happens in the previous screen so it's hard for them to make the connection. I'm just following how it used to work, but I agree we should probably improve on this, I'll create an issue

packages/mobile/src/send/SendAmountLegacy/SendAmount.test.tsx
gnardini: Need more funds because sending now also includes the fee (it didn't before) so the funds they need are a bit higher.

packages/mobile/src/send/SendConfirmation.tsx
gnardini: Yes

packages/mobile/src/fees/saga.ts
gnardini: Yes, but in redux. This was I think legacy or something, it was storing fees in this map which is not a pattern we use anywhere else. The comment said `In an effort to standardize and improve fee estimation, we should either
update and use this mechanism everywhere or remove it in favor of another solution.` I chose the second option, removing it in favor of storing fee estimations on Redux

packages/mobile/src/fees/saga.test.ts
gnardini: It throws an error. If the balance is exactly zero there is a check that will fail and mark the estimation with `error` and if it's slightly more than zero but less than the required fee the fee estimation fails

packages/mobile/src/redux/store.ts
jeanregisser: Yes no strong opinion, but if we decide to persist it we should be able to recover if it's stored as loading and the app is then killed and restarted.

gnardini: Yeah, I think this is not an issue since we always estimate the fee information if there's stale data in the store, even if it's already set to 'loading' so I don't think the UI can get stuck like that.

packages/mobile/src/send/SendConfirmation.test.tsx
gnardini: Here's the issue: https://app.zenhub.com/workspaces/retention-squad-sprint-board-612f2ef98ef2b800116e6043/issues/valora-inc/wallet/1549

jeanregisser (approved):

ValoraQA:
Hi @MuckT, We have verified this issue using iOS Test flight build V1.23.0 & updated status in below Gsheet.

**Sheet Link** :
https://docs.google.com/spreadsheets/d/15Z15zu4roT6XFD1dbi9__c3Q6aQZpIiIZka1rKVrCgw/edit#gid=176031153

**Devices** : iPhone 13 mini (15.1.1), iPhone 7(13.3), One Plus 7T(11.0)

CC: @gnardini 
Thanks..!!

ValoraQA:
Hi @MuckT , We have verified this issue using iOS Test flight build V1.23.0 & Android Internal build V: 1.23.0 and updated status in below Gsheet.

**Sheet Link :**
https://docs.google.com/spreadsheets/d/15Z15zu4roT6XFD1dbi9__c3Q6aQZpIiIZka1rKVrCgw/edit#gid=176031153

**Observation :** If we enter amount as 1$ then estimation fees is seen as $0.001 , if we enter amount as 3$ then also the estimation fees is seen as $0.001 so is this calculation fine or there is some formula by which it is calculated 

Devices : iPhone 13mini(15.1.1) , iPhone SE(14.2) , OnePlus 7T(11.0) , Samsung Galaxy M40 (10.0)
CC: @gnardini
Thanks..!!

ValoraQA:
Hi @MuckT , We have verified this issue using Latest Android Internal build V: 1.23.0 and updated status in below Gsheet.

**Sheet Link :**
https://docs.google.com/spreadsheets/d/15Z15zu4roT6XFD1dbi9__c3Q6aQZpIiIZka1rKVrCgw/edit#gid=176031153

**Observation :** If we enter amount as 1$, 3$, 0.01$ & 0.02$ then estimation fees displayed remains same as $0.001 , so is this calculation fine or there is some formula by which it is calculated.

**New Bug:** [Link](https://github.com/valora-inc/wallet/issues/1649)

**Devices :** Samsung Galaxy A8+(8.0), Redmi Note 5 Pro (9.0)
CC: @gnardini
Thanks..!!

PR #1412, opened 2021-11-10
WalletConnect V1 E2E Tests

### LINKED ISSUE #1141: Add E2E test for WalletConnect
WalletConnect doesn't have any test yet in Valora.
Given the importance it's taking, it's becoming critical that we cover it.

We should cover both WalletConnect v1 and v2.

One idea to make it work with Detox would be to run a test WalletConnect dapp within the Detox env.
And then generate the `wc:XXXX` link with it that can be used to initiate the pairing (via deeplink).

We should also cover the different [supported methods](https://github.com/valora-inc/wallet/blob/949aa7ecaf38165862b47af3d7b445191cb1ab9e/packages/mobile/src/walletConnect/constants.ts#L3) (`eth_sendTransaction`, `personal_sign`)

MuckT:
### Description

Contains E2E tests for the WalletConnectV1. These tests are working on both Android and iOS

### Other changes

- Updated misc testID's and dynamically rendered setting items to avoid `undefined/value` testID's.
- Moved WalletConnect e2e specific utils to their own module.

### Tested

Tested locally on iOS and Android. Some tests have been skipped as they are currently failing.

```
PASS e2e/src/WalletConnect.spec.js (137.49 s)
  Given Wallet Connect
    When V1
      ✓ Then is able to establish a session (30677 ms)
      ✓ Then is able to send a transaction (eth_sendTransaction) (21855 ms)
      ✓ Then is able to sign a personal message (personal_sign) (7627 ms)
      ✓ Then is able to sign message (eth_sign) (7903 ms)
      ✓ Then should be able to disconnect a session (17695 ms)
      ○ skipped Then is able to sign a transaction
      ○ skipped Then is able to sign typed data (eth_signTypedData)
      ○ skipped Then is able to send custom request
```

### How others should test

QA does not need to test this. The PR reviewer should run the WalletConnect E2E tests with `yarn test:e2e:ios -f WalletConnect.spec.js` & `yarn test:e2e:android -f WalletConnect.spec.js`

### Related issues

- Fixes #1141 for V1 WalletConnect.

### Backwards compatibility

This change only adds additional tests for the e2e test framework and is backwards compatible.


MuckT:
These tests are passing; however, several reruns were needed on iOS due to timeout issues with wallet connect. I'd like to make these e2e tests pass 100% of the time if there is not an error with Valora.

packages/mobile/src/walletConnect/screens/Loading.tsx
```diff
@@ -13,7 +13,7 @@ import { isScreenOnForeground, navigate } from 'src/navigator/NavigationService'
 import { Screens } from 'src/navigator/Screens'
 import { StackParamList } from 'src/navigator/types'
 
-const CONNECTION_TIMEOUT = 10_000
+const CONNECTION_TIMEOUT = 20_000
```
kathaypacific, line 16: was this increased for the CI? i think 20 seconds is a long time for a user to wait (but then again, 10 seconds is already a long time) - my only thought is in mainnet we'd lose some of the wallet connect timeout analytics after 10 seconds if the user abandons the app after that time. not sure if this is a big issue, it could be fixed by moving this to config.ts and changing it for the test env if we want :)

packages/mobile/src/firebase/remoteConfigValuesDefaults.e2e.ts
```diff
@@ -14,8 +14,8 @@ export const REMOTE_CONFIG_VALUES_DEFAULTS: Omit<
   inviteRewardsEnabled: false,
   inviteRewardCusd: 1,
   inviteRewardWeeklyLimit: 20,
-  walletConnectV1Enabled: false,
-  walletConnectV2Enabled: false,
+  walletConnectV1Enabled: true,
+  walletConnectV2Enabled: true,
```
kathaypacific, line 18: just checking we want/need these changes here? i thought wallet connect v2 was disabled because it was causing issues?

packages/mobile/e2e/src/usecases/WalletConnectV2.js
```diff
@@ -0,0 +1,71 @@
+import WalletConnectClient, { CLIENT_EVENTS } from '@walletconnect/client'
```
kathaypacific, line 1: this spec will never be run because it's skipped in the WalletConnect.spec...i'm not clear on what the plan is for wallect connect v2, it just feels kind of odd to be creating new files that aren't being run, are there plans to enable WC v2 soon?

packages/mobile/e2e/src/usecases/WalletConnectV1.js
```diff
@@ -0,0 +1,324 @@
...
+      .withTimeout(10 * 1000)
+
+    // Wait for signature
+    let signature = await result
+
+    // Verify the signature - currently both return valid
+    const valid = verifySignature(JSON.stringify(typedData), signature, fromAddress)
+    const invalid = verifySignature(JSON.stringify(typedData), signature, toAddress)
+    jestExpect(valid).toStrictEqual(true)
+    jestExpect(invalid).toStrictEqual(false)
```
kathaypacific, line 273: wondering if there's any difference between `jestExpect` and `expect` - i see `expect` in most other specs?

packages/mobile/e2e/src/usecases/WalletConnectV1.js
```diff
@@ -0,0 +1,324 @@
...
+
+  it('Then is able to sign a personal message (personal_sign)', async () => {
+    const message = `My email is valora.test@mailinator.com - ${+new Date()}`
+    const msgParams = [
+      // Both Required
+      utf8ToHex(message),
+      fromAddress,
+    ]
+    let result = walletConnector.signPersonalMessage(msgParams)
+    await waitFor(element(by.text('Sign a data payload')))
```
kathaypacific, line 165: was wondering if you think it'd be a good idea to abstract the logic for lines 165-211 so that it can be reused? just noticed that it's repeated in a few tests, might be easier to maintain if it's centralised

kathaypacific (approved):
Nice work! I am having issues running the tests locally on my M1 macbook so i couldn't test (i will fix this but it's unrelated to this PR) but the code looks good :)

MuckT:
These tests started failing a lot, converted to draft while I figure out why.

packages/mobile/src/firebase/remoteConfigValuesDefaults.e2e.ts
MuckT: Good catch! I'll disable V2 until we enable it.

packages/mobile/e2e/src/usecases/WalletConnectV1.js
MuckT: Detox overwrites jest's expect with it's own. This reimports jest's expect under a different name.

MuckT:
@Jean @silasbw we might want to wait for https://github.com/valora-inc/wallet/issues/1559 to be fixed prior to merging these tests.

packages/mobile/e2e/src/usecases/WalletConnectV2.js
MuckT: This is just some pre-work for WalletConnectV2 when we turn it on in Valora.

MuckT:
These test should pass once #1729 is merged.

PR #1414, opened 2021-11-11
add i18n gate to initialise i18n before app is loaded

kathaypacific:
### Description

This PR tries to make the i18n init flow more robust, in preparation for over the air translations.

`i18n.init` returns a promise, and we shouldn't be using `i18n` before it is properly initialised. This PR aims to solve the following:
- In the current implementation, we invoke the `i18n.init` promise without waiting for it to resolve. This is okay for now since the promise resolves pretty quickly but this won't work reliably if we were to cache the translations (from over the air, for example) and read them on i18n init.
- the `currentLanguageSelector` is returning data from the store OR from `i18n` which I thought was unexpected. The store value for `language` is only `null` on first app load but at this time `i18n` could also be not initialised.
- the `setLanguage` action is dispatched on every app load. This kicks off the `syncLanguageSelection` and updates the user data in Firebase. This isn't necessary on every app load if the user hasn't changed their language.

What is the change:
- explicitly wait for i18n init before app loads (this might make the app load screen show for a bit longer if the i18n is slow to initialise, but if we don't wait then the app text isn't being properly translated)
- remove the language detector from i18n and init i18n with language from persisted redux value, or device language for new app install, or set i18n to show english (therefore showing the language selection screen in english)

Why is this needed?
- for over the air translations, if we want to cache the translations and load them on app start, this is an asynchronous action and we need to handle the loading state

### Other changes

N/A

### Tested

Scenarios I tested:
- after changing language, the app should reflect the new language
- after changing language and closing the app, the new language should be reflected on the next app load
- a new app install should pick up the "best" language as default (based on choice made by react-native-localize)

### How others should test

Smoke test for app load/app install/changing languages - you should observe no difference. 

### Related issues

N/A

### Backwards compatibility

N/A


packages/mobile/src/app/I18nGate.tsx
```diff
@@ -0,0 +1,40 @@
...
+  useEffect(() => {
+    const i18nInit = async () => {
+      try {
+        await initI18n(language || bestLanguage || DEFAULT_APP_LANGUAGE)
+        if (!language && bestLanguage) {
+          dispatch(setLanguage(bestLanguage))
+        }
+        setIsInitialised(true)
+      } catch (reason) {
+        Logger.error('i18n', 'Failed init i18n', reason)
```
silasbw, line 31: What is the app experience on error?  Is this app usable if `initI18n` fails?

packages/mobile/src/i18n.ts
```diff
@@ -47,34 +44,13 @@ function getAvailableResources() {
...
-  cacheUserLanguage: (lng: string) => {
-    Logger.debug(TAG, `Skipping user language cache ${lng}`)
-  },
-}
-
-i18n
-  .use(languageDetector)
-  .use(initReactI18next)
-  .init({
+export const initI18n = async (lng: string) => {
```
silasbw, line 47: nit: `lng` -> `language` ("language" is used in other variable names)

packages/mobile/src/i18n.ts
```diff
@@ -47,34 +44,13 @@ function getAvailableResources() {
...
+  await i18n.use(initReactI18next).init({
```
silasbw, line 48: code golf:

```ts
return i18n.use(initReactI18next).init({
```

packages/mobile/src/i18n.ts
silasbw: code golf (drop the asyc, see my comment below) and nit (use the function expression because that's the convention):
```
export function initI18n(language: string) {
```

packages/mobile/src/i18n.ts
```diff
@@ -47,34 +44,13 @@ function getAvailableResources() {
...
     fallbackLng: {
-      default: ['en-US'],
+      default: [DEFAULT_APP_LANGUAGE],
       'es-US': ['es-LA'],
     },
+    lng,
```
silasbw, line 53: ```lng: language```

silasbw (commented):
:hot_pepper: !

packages/mobile/src/app/I18nGate.tsx
kathaypacific: Ah, thanks for pointing this out! In this implementation the app will get stuck on the loading state indeed . we can either:
1. show the generic error screen here - unfortunately even the generic error screen requires us to use `t` to display the error, so if the `init` fails then we will see the translation keys
2. keep the current implementation, where we will allow the app to function normally. for a non english speaking user to see english dev keys, i'm not sure that's helpful at all...

i chose the error screen personally because i18n init is unlikely to fail, if it is failing then there are probably problems elsewhere so we can't guarantee that the app is working as expected. let me know if you think the second option is better, and i will change it :)

packages/mobile/src/i18n.ts
kathaypacific: good one, i've updated :)

packages/mobile/src/i18n.ts
kathaypacific: updated :D

packages/mobile/src/i18n.ts
kathaypacific: also updated :)

packages/mobile/src/app/App.tsx
```diff
@@ -111,7 +112,9 @@ export class App extends React.Component<Props> {
             <PersistGate loading={<AppLoading />} persistor={persistor}>
               <StatusBar backgroundColor="transparent" barStyle="dark-content" />
               <ErrorBoundary>
-                <NavigatorWrapper />
+                <I18nGate fallback={<AppLoading />}>
```
jeanregisser, line 115: Why not directly as a child of `PersistGate`?

jeanregisser (commented):
Looking great! 👍 

There's one current behavior I'm curious about:
If we fail to determine the best language to use (for instance, phone is in Italian but we don't have Italian translations), do we still display the LanguageScreen? or directly show the Welcome screen in English?

kathaypacific:
> Looking great! 👍
> 
> There's one current behavior I'm curious about: If we fail to determine the best language to use (for instance, phone is in Italian but we don't have Italian translations), do we still display the LanguageScreen? or directly show the Welcome screen in English?

I wanted to keep the behaviour the same, so show the LanguageScreen (in english) if react-native-localise cannot detect a best language :)

packages/mobile/src/app/App.tsx
kathaypacific: I tried to block as little as possible and thought maybe a good idea for the I18nGate to have the ErrorBoundary to fall back on...although now I can see that ErrorBoundary also depends on i18n :/ so yeah good shout, i will move this below PersistGate

packages/mobile/src/app/I18nGate.tsx
silasbw: sgtm!

kathaypacific:
@jeanregisser this is ready for re-review :)

gnardini (commented):
Looks great! Just left a small comment

silasbw (approved):
lgtm, but I defer to the expert reviewers (@gnardini and @jeanregisser).

packages/mobile/src/app/I18nGate.tsx
```diff
@@ -0,0 +1,45 @@
...
+}
+
+const I18nGate = ({ loading, children }: Props) => {
+  const [isInitialised, setIsInitialised] = useState(false)
+  const [initFailure, setInitFailure] = useState(false)
+  const dispatch = useDispatch()
+  const language = useSelector(currentLanguageSelector)
+  const bestLanguage = findBestAvailableLanguage(Object.keys(locales))?.languageTag
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
kathaypacific: makes sense :) i didn't see that useAsync hook before, i've updated the component to use this now!

packages/mobile/src/app/I18nGate.tsx
```diff
@@ -0,0 +1,46 @@
...
+    }
+    return true
+  }, [])
+
+  useEffect(() => {
+    if (i18nInitResult.error) {
+      Logger.error('i18n', 'Failed init i18n', i18nInitResult.error)
+      navigateToError('appInitFailed')
+    }
+  }, [i18nInitResult.error])
```
jeanregisser, line 36: nit: `useAsync` has an `onError` option which would simplify things here a bit.
See https://github.com/slorber/react-async-hook/blob/db317141ed72ec2917b3eb7b1b344454116a23c3/test/useAsync.test.ts#L37-L46

jeanregisser (approved):
Amazing work! 🎉

packages/mobile/src/app/I18nGate.tsx
kathaypacific: good idea, i have updated to use this :)

Issue #1377, opened 2021-11-04
[M1] Separate Persona component out from Tarik's PR


lisaqyli:
Here is Tarik's PR on Persona Integration https://github.com/valora-inc/wallet/pull/1241
The PR includes the work to add Persona package into the build, the component to trigger and manage Persona flow and integration with the raise transaction limit flow. 

AC:

- [ ] Persona npm package is added to wallet and build is successful
- [ ] The component to trigger and manage Person is added to the wallet repo
- [ ] kycStatus is stored and managed in redux and saga

PR #1419, opened 2021-11-11
Persona Component

lisaqyli:
### Description
This PR is adapted from Tarik's previous PR https://github.com/valora-inc/wallet/pull/1241/files
This PR creates the persona component, which launches Persona identity verification flow when triggered. It also calls IHL to create an account when the component is launched for the first time. 

### Tested
Unit tested.


packages/mobile/src/account/Persona.tsx
```diff
@@ -0,0 +1,118 @@
...
+    if (!accountAddress) {
+      Logger.error(TAG, "Can't render Persona because accountAddress is null")
+      return
+    }
+
+    Inquiry.fromTemplate(templateId)
+      .referenceId(accountAddress)
+      .environment(networkConfig.personaEnvironment)
+      .onSuccess((inquiryId: string, attributes: InquiryAttributes) => {
+        console.log(
```
jh2oman, line 52: Should there be a TODO in here to add more logic?

packages/mobile/src/account/saga.ts
```diff
@@ -112,6 +114,30 @@ export function* watchDailyLimit() {
   }
 }
 
+export function* watchKycStatus() {
+  const account = yield call(getWalletAddress)
+  const channel = yield call(kycStatusChannel, account)
```
cajubelt, line 119: I think there's a bug here-- to get the user's kyc from firebase you'll need their MTW address, not their EOA address

https://docs.celo.org/celo-codebase/protocol/identity/valora-accounts

cajubelt: this also means users need an MTW set up already (which happens during phone verification) to access the feature. If that comes as a surprise we should discuss further to make sure the requirement is accounted for in our designs and GH issues.

cajubelt (commented):
Thanks for getting this started! Left some comments

packages/mobile/src/account/Persona.tsx
```diff
@@ -0,0 +1,124 @@
...
+    <Button
+      onPress={launchPersonaInquiry}
+      text={t('raiseLimitBegin')}
+      type={BtnTypes.PRIMARY}
+      size={BtnSizes.FULL}
+      testID="PersonaButton"
+      disabled={
+        !personaAccountCreated
+        // ONLY FOR TESTING PURPOSES
+        // || !templateId
```
jh2oman, line 118: I commented this out temporarily just to get a working test. All affects from hooks happen after the initial render, so we need to be able to wait for some UI change to verify things like `expect(mockFetch).toHaveBeenCalledTimes(1)`.

One key principle of react-testing-library is that you should keep most of the tests focused on the visual changes the user experiences, in this case the button becomes enabled / disabled based on results from Persona and IHL. In order to truly test this we need to also be able to mock / control `templateId` since it blocks us from testing whether or not the button becomes enabled /disabled.

Next step should be mocking `templateId` from firebase so that you can test this button more effectively

packages/mobile/src/account/saga.ts
lisaqyli: It looks like if the user doesnt have phone number verified, there's a `unverifiedMtwAddress` we can get from Komenci

packages/mobile/src/account/Persona.tsx
lisaqyli: oh sweet, just mocked templateId! ty for looking into it

packages/mobile/src/firebase/firebase.ts
```diff
@@ -349,6 +349,10 @@ export async function cUsdDailyLimitChannel(address: string) {
   return simpleReadChannel(`registrations/${address}/dailyLimitCusd`)
 }
 
+export async function kycStatusChannel(address: string) {
```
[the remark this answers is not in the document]
lisaqyli: are you suggesting to rename it to mtwAddress?

packages/mobile/src/account/Persona.tsx
```diff
@@ -24,7 +24,7 @@ export interface Props {
 }
 
 const Persona = ({ kycStatus }: Props) => {
-  const { t } = useTranslation(Namespaces.accountScreen10)
+  const { t } = useTranslation(Namespaces.translation)
   const [personaAccountCreated, setPersonaAccountCreated] = useState(!!kycStatus)
 
   const accountAddress = useSelector(accountAddressSelector)
```
cajubelt, line 30: accountAddress defaults to the EOA address, which won't work. IHL requires this to be the mtwAddress. It would be better to use the mtwAddressSelector here.

cajubelt (dismissed):
There are remaining issues with mtw vs eoa address. 

Also, there isn't enough code coverage to pass CI.

packages/mobile/locales/en-US/global.json
```diff
@@ -0,0 +1,210 @@
+{
```
jh2oman, line 1: Is this file here by mistake?  Most of these strings don't have anything to do with Persona

packages/mobile/package-lock.json
```diff
@@ -0,0 +1,14 @@
+{
```
jh2oman, line 1: Do you know why the package-lock.json changed? It doesn't look like any changes were made to package.json

jh2oman (approved):
Looks good after some more tests to get the code coverage up 👍

cajubelt (approved):

cajubelt:
cc #1377

cajubelt:
@lisaqyli did you merge this before CI checks passed?

Issue #1332, opened 2021-10-28
Connected name option is shown in “Connected apps” screen after Tapping on “Allow” button for second time from wallet connect prompt as user Declined permission for same earlier


ValoraQA:
**Frequency:** 100%

**Repro on build version:** Test flight Build V1.21.0(1635160132), Android Internal Build V1.21.0
**Repro on devices:** iPhone 12 (15.0), One plus 7t (11.0), iPhone 6+(12.5.4), iPhone SE (14.2) 
**Browser Info:** Safari V15.0, Safari V12.5.4, Safari V14.2, Chrome V94.0

**Pre-condition:** 
1. User must have installed the app and completed onboarding flow by verifying the phone number.
2. User must have opened url: https://app.sushi.com/swap in the mobile browser. 
3. User must have tap on “Cancel” button from the wallet connect prompt for the first time in app, after tapping in “Connect to a wallet” button from browser.

**Repro Steps:**
1.  Navigate back to browser window after tapping on “Cancel” button from the wallet connect prompt in app.
2. Close the “select wallet” popup from the browser
3. Tap on “Connect to a wallet” button again from the browser
4. Tap on “Wallet Connect” option from select Wallet popup
5. Tap on “Valora” option from wallet connect popup.
6. Tap on “Allow” button from wallet connect prompt in the app
7. “Success! Please..” message appear in blue banner.
8. Navigate to “Connected apps” screen in & observe the dapp option is not displayed.  
9. Navigate back to browser window & observe the dapp appears connected.
10. Tap on “Disconnect” button & disconnect dapp from browser.
11.  Repeat Step 3,4,5
12. Tap on “Allow” button for the wallet connect prompt
13. “Success! Please..” message appear in blue banner.
14. Navigate to “Connected apps” screen in & observe the dapp option is displayed now.

**Current Behavior:** Connected dapp option is visible in “Connected apps” screen after Tapping on “Allow” button for second time from wallet connect prompt in app

**Expected Behavior:** Connected dapp option should be visible in “Connected apps” screen after Tapping on “Allow” button for first time from wallet connect prompt in app, after Tapping on “Cancel” button once from wallet connect prompt.

**Investigation:** This issue is reproducible for both mWeb & Web browsers. 

**Impact:** Bad user experience.  

**Attachment:** [Dapp issue.mp4](https://drive.google.com/file/d/1yOeBHAlN5CsGkF6mgsZsYd9QNAW3o6gh/view?usp=sharing)

[...]

Issue #935, opened 2021-07-30
“Connecting……” text is shown continuously for longer time when user purchased celo.


ValoraQA:
**Frequency:** 100%
**Repro on build version:**  Android internal build V1.17.0, Playstore build V1.16 & Testflight build V1.17.0(71), Appstore build V1.1
**Repro on devices:** Vivo V11 pro (10),  Redmi Note 8 (9.0),iPhone 6+(12.5.4), iPhone 12 (14.7.1)
**Testing Account:**     (680) 333 1978/ [Account Key](https://docs.google.com/spreadsheets/d/1gXY1I4PBkT9f1uB3zxWgUUJJ1zjVeJcqcWLjiIls1f0/edit#gid=0&range=B9)
**Pre-condition:** Make sure user should wait for 3-5mins after purchase on the activity screen. 
**Repro Steps:**

1. Launch the application & Login with test account 
2. Go to celo buy/Sell page 
3. Tap on Buy button 
4. enter valid amount 
5. Tap on review button 
6. Tap on celo buy button
7. wait for 5mins on activity & observe 

**Current Behavior:**  Connecting activity is shown for longer time
**Expected Behavior:**  Activity should confirm after some time 
**Observations** 

- Sold activity is shown confirmed after some time 
- Amount get updated after some time for purchased 
- Activity is not shown after kill & relaunch when it was in connecting state 

**Impact:** Bad user experience 
**Attachment:** [ConnectingText.mp4](https://drive.google.com/file/d/1_Ethdnum-TToshiQXOOqrj1EiVcvRfTe/view?usp=sharing )

Celoqa:
https://valora.testquality.com defect [D82](https://valora.testquality.com/project/12128/result/2450418/defect/8017).

ValoraQA:
https://valora.testquality.com defect [D93](https://valora.testquality.com/project/12128/result/2464175/defect/8064).

ValoraQA:
https://valora.testquality.com defect [D104](https://valora.testquality.com/project/12128/result/2470969/defect/8183).

ValoraQA:
https://valora.testquality.com defect [D105](https://valora.testquality.com/project/12128/result/2470987/defect/8184).

ValoraQA:
https://valora.testquality.com defect [D112](https://valora.testquality.com/project/12128/result/2471065/defect/8191).

ValoraQA:
[...]

Issue #931, opened 2021-07-29
"Select Language" page does not appear on opening the android app & iOS app after fresh install 


ValoraQA:
**Frequency:** 100%
**Repro on build version:** Android internal build V1.17.0
**Repro on devices:** Google Pixel 4a (11.0), Samsung Galaxy J5 Prime (8.0) , iPhone 6+(12.5.4), Samsung Galaxy M40(10.0),Google Pixel 2XL (11.0)
**Does not Repro Devices:** iPhone 12 (14.7.1)
**Does not Repro build version:** Test flight build V1.17.0

**Precondition:** Make sure user has freshly installed the application.
**Repro Steps:**
1. Launch the freshly installed application.
2. Observe that.

**Current Behavior:** "Select Language" page does not appear on opening the freshly installed app.
**Expected Behavior:** "Select Language" page must appear on opening the freshly installed app.
**Impact:** Bad user experience.
Attachment: [Select Language Page Issue Android.mp4](https://drive.google.com/file/d/1XqQpmFG-ppUFJaNMK9THUcFBMVrBrff2/view?usp=sharing)
[Select Language Page Appear For iOS.mp4](https://drive.google.com/file/d/1whXcFToE5nKxlTgxohqGa5hlK9m7tFsW/view?usp=sharing)
[language_page_not_seen.mp4](https://drive.google.com/file/d/1XXZngT9s2Amsrt71qDu9GbmS9Hob98gW/view?usp=sharing)

Celoqa:
Hi, We have verified this issue in Latest Android Internal Build V1.17.0 & the issue is repro on Android 7.0 & does not not repro on Android 6.0.
**Repro Device:** Samsung Galaxy A5 (7.0)
**Does Not Repro Device:** LG  K7i (6.0)

ValoraQA:
https://valora.testquality.com defect [D99](https://valora.testquality.com/project/12128/result/2468995/defect/8156).

ValoraQA:
https://valora.testquality.com defect [D128](https://valora.testquality.com/project/12128/result/2479300/defect/8214).

ValoraQA:
https://valora.testquality.com defect [D130](https://valora.testquality.com/project/12128/result/2481953/defect/8221).

ValoraQA:
https://valora.testquality.com defect [D178](https://valora.testquality.com/project/12128/result/2514529/defect/8329).

ValoraQA:
[...]

PR #1423, opened 2021-11-11
Change WalletHome to be a function component

gnardini:
### Description

Changed the `WalletHome` component to be a function component. This is the main screen of the app and was due for the update. No logic should have been changed.

### Other changes

- Changed in some places the wording of `gold` for `celo`. Gold is the original name of the token but was changed a longtime ago. Didn't go through all the code to change it, only the places that were affected by this PR.

### Tested

Manually and updated the unit tests.

### How others should test

Just go through the test plan normally and make sure everything continues to work.

### Related issues

- Part of #546

### Backwards compatibility

N/A


packages/mobile/src/home/WalletHome.tsx
```diff
@@ -1,212 +1,156 @@
...
+        t('testnetAlert.0', { testnet: _.startCase(DEFAULT_TESTNET) })
+      )
+    )
+  }
 
-    this.scrollPosition = new Animated.Value(0)
-    this.onScroll = Animated.event([{ nativeEvent: { contentOffset: { y: this.scrollPosition } } }])
-    this.state = {
-      isMigrating: false,
+  const tryImportContacts = async () => {
```
isabellewei, line 66: do we no longer want to check if the user's number is verified before importing contacts?

isabellewei (approved):
LGTM! Thanks for doing this :)

packages/mobile/src/home/WalletHome.tsx
```diff
@@ -1,212 +1,156 @@
...
 
-  shouldShowCashInBottomSheet = () => {
-    const hasStable = STABLE_CURRENCIES.some((currency) =>
-      this.props.balances[currency]?.isGreaterThan(STABLE_TRANSACTION_MIN_AMOUNT)
-    )
-    const hasGold = this.props.balances[Currency.Celo]?.isGreaterThan(GOLD_TRANSACTION_MIN_AMOUNT)
-    const isAccountBalanceZero = !hasStable && !hasGold
+  useEffect(() => {
+    if (appState === AppState.Active) {
+      dispatch(refreshAllBalances())
```
jeanregisser, line 92: not directly related to this PR, but I noticed due to the way the sagas are initialized, this action isn't handled by any saga (still waiting in the blocking `appInit` saga).
But maybe we can remove this soon anyway with the multi token changes?

packages/mobile/src/home/WalletHome.test.tsx
```diff
@@ -1,139 +1,133 @@
...
-    )
-    expect(tree).toMatchSnapshot()
+    const { getByTestId } = renderScreen({
+      ...zeroBalances,
+      app: {
+        cashInButtonExpEnabled: true,
+      },
+    })
+
+    expect(getByTestId('cashInBtn')).toBeTruthy()
```
jeanregisser, line 120: Love seeing more useful tests and not just a big snapshot test that nobody looks at ;)

jeanregisser (approved):
Amazing!! Thanks so much! 🙏 💯

packages/mobile/src/home/WalletHome.tsx
gnardini: Yeah, I don't know why I changed that, didn't mean to, undid that change, thanks!

packages/mobile/src/home/WalletHome.tsx
gnardini: We still need to fetch balances on app load, for now I just added `yield call(refreshBalances)` to the saga initialization, but we should think about how to handle the token fetching in a more robust way. I think a good time to do that is when we remove the old code

ValoraQA:
Hey @MuckT We have verified this issue using Android internal build and iOS test flight build V:1.22.1 &  updated observation in BVT test plan from test quality tool
**BVT test quality tool link:**  https://valora.testquality.com/project/12128/plan/20403/run/153293/result/2581215
**Query/Issues:**
We have already filed below Query/Bugs in the whole application while performing BVT 
- https://github.com/valora-inc/wallet/issues/931
- https://github.com/valora-inc/wallet/issues/1452
- https://github.com/valora-inc/wallet/issues/935
- https://github.com/valora-inc/wallet/issues/1332
**Verified On Devices:** Samsung Galaxy F41(11.0), iPhone 12(15.0)
CC: @gnardini 
Can you please let us know if any other things need to test here ?
Thanks..!!

PR #1427, opened 2021-11-11
Fix broken ipapi schema for `fetchUserLocationData` cloud function

jophish:
### Description

Bit of a post-mortem incoming:

_Note:_ As a stopgap until this PR is approved and deployed, I've manually redeployed previous cloud function version from the August 2 commit linked below, which should stop the 500 errors we're seeing. The actual user impact of this should have been low; I _believe_ the only effect is that some non-US users could _potentially_ be being shown an incorrect list of CICO providers.

 
Earlier today, I redeployed our CICO-related cloud functions so that they would send data to our own Valora GCP. While doing so, I also redeployed the `fetchUserLocationData` cloud function; this wasn't *strictly* necessary, but it did import code from files that had been modified to support the GCP move, so I redeployed it along with the other functions.

A little while ago, I noticed that the e2e tests for one of my PRs [were failing](https://github.com/valora-inc/wallet/runs/4183312901?check_suite_focus=true), for reasons related to CICO provider eligibility (my PR had nothing to do with anything related to that). I don't exactly know how the e2e tests work for the app; but I wondered if they were hitting the actual production cloud functions (or at least the `fetchProviders`/`fetchUserLocationData` functions). Navigating to the [logs](https://console.cloud.google.com/functions/details/us-central1/fetchUserLocationData?project=celo-mobile-mainnet&tab=metrics) for our mainnet deploy of `fetchUserLocationData`, I noticed intermittent 500 errors, beginning around when I redeployed the function.

I was under the (reasonable?) assumption that the code checked into `fetchUserLocationData.ts` was the most-recent, currently-deployed version of the code, in which case redeploying the function would be completely safe. Navigating to, and downloading the source of the previously deployed version of the code ([here](https://console.cloud.google.com/storage/browser/_details/gcf-sources-1027349420744-us-central1/fetchUserLocationData-e97b6a5b-2125-404b-83d8-97df402a351d/version-3/function-source.zip;tab=live_object?project=celo-mobile-mainnet), version 3, deployed August 2 2021), I noticed that it was completely different than the code currently checked in. A bit of digging reveals that the previously deployed version of the code actually corresponds to *around* [this commit](https://github.com/valora-inc/wallet/commit/c5a7a99863129caf825e5e73186973d21dd97de8) (not sure exactly since Git commit information is not contained in the old CF source I pulled from Google Cloud), from August 2, 2021, though the last change to `fetchUserLocationData.ts` directly is from [this commit](https://github.com/valora-inc/wallet/commit/7e90de155fb8cd00fc6909b6990ab5e9b1255c3f#diff-2d97dcf9fab599b5c4f1c41780a4c1eb8c309d4b5d0e712ecfb83fe63543b49d) from June 1, 2021.

There were two commits ([here](https://github.com/valora-inc/wallet/commit/6b57faf4f493707eb7c378c8b869691b21a7c8dd#diff-2d97dcf9fab599b5c4f1c41780a4c1eb8c309d4b5d0e712ecfb83fe63543b49d), and [here](https://github.com/valora-inc/wallet/commit/1b2f1516c7bb631980f8189fd2e6ade2c9bb59da#diff-2d97dcf9fab599b5c4f1c41780a4c1eb8c309d4b5d0e712ecfb83fe63543b49d)) between that deploy and now; the main difference is that schema validation was added for the response from the IP API service.

The 500s are being thrown from a block that gets executed when the IP API response doesn't match the schema. Some manual testing by spoofing my IP with VPN revealed that for some (non-US) countries, the region/city/etc fields are set to `null` . While we require `region_code` explicitly for the response of this function, the version of the function deployed until today was returning `null` for `region_code` *anyways*, so allowing these fields to be nullable puts this function back at parity with the previously deployed version.


jophish:
Issue for adding testing and alerting: #1428

silasbw (approved):

jeanregisser (approved):

PR #1433, opened 2021-11-12
Crowdin OTA translations

kathaypacific:
### Description

1. use crowdin ota client
2. on app load, read from ota translations file if it exists (otherwise use the bundled translations as usual) - this means that if a user fetches the ota translations once, the won't see the bundled translations on subsequent app loads
3. ota functionality hidden behind remote config flag - if the flag is turned on, launch ota saga
4. add ota saga that fetches + caches the ota translations for the selected language only if there has been an update/on language change/on app update. Caching manually in a file was the most reliable way to cache the translations (the crowdin project uses in memory cache only)
3. ota saga is run after fetching feature flags, and also on change language. it is NOT run on app init because the feature flag values are from the previous app init

What is pending?
- [x] I am working with Paula to align our Crowdin setup - at the moment english is not a target language therefore we cannot fetch ota updates for english until Paula's changes are made
- [x] the ota distribution combines all results into one blob,. Since we want to combine our translations into one file anyway, i am only playing with the global.json translations for this branch (as a POC). I will migrate our translations into 1 file before picking up this branch again
- [x] decide on what behaviour is acceptable when the remote config flag is toggled. as fetching feature flags is not blocking the app from loading, if the flag is true for session 1 but false for session 2, what should happen while the feature flags are being fetched at the start of session 2? (at the moment, i reinitialise i18n after the feature flag is fetched, so the user sees the app copy updated when the feature flag is updated - not sure if this is acceptable) @jeanregisser @gnardini @silasbw 
- [x] unit tests for the ota saga and initI18n function are coming

### Other changes

N/A

### Tested

If the remote config flag for "allowOtaTranslations" is true:
1. modify the bundled translations OR the Crowdin translations, so that there is a mismatch between the 2 sources
2. load the app the first time - see the bundled translations first, and ota translations are fetched in the background
3. on navigating away/refreshing the components, see the new ota translations
4. on restart of the app, only ota translations are seen
5. the same sequence of behaviours observed on language change in app settings

If the remote config flag for "allowOtaTranslations" is false:
The app should run as it does now, with bundled translations

### How others should test

Same steps above

### Related issues

N/A

### Backwards compatibility

None. But good to note that once this feature is rolled out, it becomes harder to update our translations implementation (e.g. translation file structure, key names etc) since we use one source of truth in Crowdin to service multiple app versions in the wild


packages/mobile/.env.alfajores
```diff
@@ -19,3 +19,4 @@ GETH_USE_FULL_NODE_DISCOVERY=true
 GETH_USE_STATIC_NODES=true
 # Only for development, use with caution
 GETH_START_HTTP_RPC_SERVER=false
+CROWDIN_DISTRIBUTION_HASH=e-f9f6869461793b9d1a353b2v7c
```
silasbw, line 22: will we ever need to update this value?

silasbw:
re: feature flags / remote parameters

We don't expect the remote parameter to change often for each user.  Most likely we'll use it to gradually roll out from 0% to 100%, so we'd expect it to change once for each user.

With that in mind, the "copy updated flicker" seems OK IMO.  @jeanregisser @gnardini @kathaypacific ?

yarn.lock
```diff
@@ -7295,6 +7302,13 @@ axios-retry@^3.0.2:
...
 
+axios@0.21.3:
+  version "0.21.3"
+  resolved "https://registry.yarnpkg.com/axios/-/axios-0.21.3.tgz#f85d9b747f9b66d59ca463605cedf1844872b82e"
+  integrity sha512-JtoZ3Ndke/+Iwt5n+BgSli/3idTvpt5OjKyoCmz4LX5+lPiY5l7C1colYezhlxThjNa/NhngCUWZSZFypIFuaA==
+  dependencies:
+    follow-redirects "^1.14.0"
+
 axios@^0.21.1:
   version "0.21.4"
```
jeanregisser, line 7313: yarn doesn't do a good job at this, but if you manually delete both axios from `yarn.lock` and re-run `yarn` it will use a single version, since they are compatible.

packages/mobile/src/app/saga.ts
```diff
@@ -68,6 +78,39 @@ export function* appInit() {
...
+    const timestamp = yield otaClient.getManifestTimestamp()
+    if (lastFetchTime < timestamp) {
+      yield spawn(otaTranslationsSaga)
+    }
+  }
+}
+
+export function* otaTranslationsSaga() {
+  try {
+    const timestamp = yield otaClient.getManifestTimestamp()
```
jeanregisser, line 94: We should prefer `yield call(...)` instead of just `yield something` so we can more easily test this saga.
See https://stackoverflow.com/a/53468319/158525

jeanregisser (commented):
W00t! awesome work! 🎉 
Mostly small comments.

packages/mobile/.env.alfajores
kathaypacific: i'll check with Paula about this.

i thought having a separate distribution for alfajores/mainnet would be a good idea, even though we only have one source of truth for the translation files we can still choose to publish updates at different times (i.e. alfajores can be a "preview" - maybe this could be useful?)

i also thought it'd be a good idea to bundle this distribution hash with the app itself, in case of future updates to the structure of the translation files we could have the flexibility to serve different distributions for different apps. again, not sure if this is useful for us - what do you think?

packages/mobile/src/app/I18nGate.tsx
```diff
@@ -32,7 +34,7 @@ const I18nGate = ({ fallback, children }: Props) => {
       }
     }
     void i18nInit()
-  }, [])
+  }, [allowOtaTranslations])
```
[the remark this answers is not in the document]
kathaypacific: ah yes good one. i will find another way to do this, i know this looks weird :/ i needed this when playing with toggling the feature flag and expecting the app to refresh with the ota/bundled translations, but this is probably not a valid scenario. update coming :)

packages/mobile/src/app/saga.ts
kathaypacific: ah yes, good one! updated :)

packages/mobile/src/i18n.ts
```diff
@@ -31,27 +32,39 @@ export enum Namespaces {
...
-  const resources = {}
-  for (const [key, value] of Object.entries(locales)) {
-    Object.defineProperty(resources, key, {
-      get: () => value!.strings,
+async function getAvailableResources(allowOtaTranslations: boolean) {
+  let cachedTranslations: Resource = {}
+
+  if (allowOtaTranslations && (await RNFS.exists(OTA_TRANSLATIONS_FILE))) {
+    cachedTranslations = JSON.parse(await RNFS.readFile(OTA_TRANSLATIONS_FILE))
+  }
```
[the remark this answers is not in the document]
kathaypacific: good one, i didn't think about this! i think that is a good idea, i added another variable to track this in redux and the comparison is there now :)

packages/mobile/src/app/saga.ts
```diff
@@ -68,6 +78,39 @@ export function* appInit() {
     navigate(Screens.SetClock)
     return
   }
+
+  const allowOtaTranslations = yield select(allowOtaTranslationsSelector)
+  if (allowOtaTranslations) {
+    const lastFetchTime = yield select(otaTranslationsLastUpdateSelector)
+    const timestamp = yield otaClient.getManifestTimestamp()
```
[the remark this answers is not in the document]
kathaypacific: it does indeed send a network request..the saga doesn't progress past this if there is no network connectivity. i updated and added a try/catch block to log the error now :)

yarn.lock
kathaypacific: i couldn't get this to work - yarn puts both versions back every time :( 

i could force this to work by editing the file manually to be:
`axios@^0.21.1, axios@0.21.3:
  version "0.21.3"
  resolved "https://registry.yarnpkg.com/axios/-/axios-0.21.3.tgz#f85d9b747f9b66d59ca463605cedf1844872b82e"
  integrity sha512-JtoZ3Ndke/+Iwt5n+BgSli/3idTvpt5OjKyoCmz4LX5+lPiY5l7C1colYezhlxThjNa/NhngCUWZSZFypIFuaA==
  dependencies:
    follow-redirects "^1.14.0"`
buuuut this didn't feel quite right because we're downgrading axios for the analytics-node package :/ if you think this is better then i can push it :) having more dependencies than we need isn't great but locking to one version also doesn't seem ideal, so i'm not sure which is better

packages/mobile/.env.alfajores
kathaypacific: I removed these from the env files because Paula said she only needs one distribution for all the apps, this value lives in the config file now :)

kathaypacific:
> re: feature flags / remote parameters
> 
> We don't expect the remote parameter to change often for each user. Most likely we'll use it to gradually roll out from 0% to 100%, so we'd expect it to change once for each user.
> 
> With that in mind, the "copy updated flicker" seems OK IMO. @jeanregisser @gnardini @kathaypacific ?

Can I confirm if the following behaviours are okay with you guys? I started to overthink and with additional logic these items can be resolved, but i wasn't sure if the added complexity is worth the effort:
1. When ota translations are refreshed, we call `i18n.addResourceBundle` but this does not update the app right away - the new translations are reflected when components are refreshed
2. if a user has ota translations, and we turn the feature flag off -> the user will see ota translations when they relaunch the app, and see the bundled translations _on the second time_ they launch the app. this is because redux persist will load the app with the previous feature flag, so that value will be there when the app/i18n initialises
3. @jeanregisser suggested that we skip the on disk translations if the app version is different to avoid loading old strings - a new scenario is then a user has an updated app + the translations have not been updated. instead of add extra logic, i've just allowed the ota fetch to happen again - the network request is ~16kb.

Let me know :)

packages/mobile/src/app/saga.ts
```diff
@@ -310,10 +327,58 @@ export function* handleSetAppState(action: SetAppState) {
   }
 }
 
+export function* handleSaveOtaTranslations(language: string, translations: Resource) {
+  const hasPreviouslyFetchedTranslations = yield RNFS.exists(OTA_TRANSLATIONS_FILEPATH)
+  if (hasPreviouslyFetchedTranslations) {
```
silasbw, line 332: I just wanted to double-check: is this existence check necessary?  It might cause a time-to-check-time-to-use bug with reading.  Consider this inter-leaving:

1. https://github.com/valora-inc/wallet/pull/1433/files#diff-42a622ab015727ac614a8c7c378e96f0ddd2a323930c2c3541c29c8fdeba37a7R23
2. https://github.com/valora-inc/wallet/pull/1433/files#diff-ff3f8a62a1a862b5c63324a903a40de66fef2fe88527f5024500dfb574a19758R333
3. https://github.com/valora-inc/wallet/pull/1433/files#diff-42a622ab015727ac614a8c7c378e96f0ddd2a323930c2c3541c29c8fdeba37a7R25

The atomic way to update contents is with:
1. create new tmp file with new contents
2. move tmp file to OTA_TRANSLATIONS_FILEPATH location

packages/mobile/src/app/saga.ts
```diff
@@ -310,10 +327,58 @@ export function* handleSetAppState(action: SetAppState) {
   }
 }
 
+export function* handleSaveOtaTranslations(language: string, translations: Resource) {
```
silasbw, line 330: Can you centralize the caching functionality into the same file?  It's spread out across this file and packages/mobile/src/i18n.ts.

packages/mobile/src/i18n.ts
```diff
@@ -1,36 +1,56 @@
...
+    (await RNFS.exists(OTA_TRANSLATIONS_FILEPATH))
+  ) {
+    cachedTranslations = JSON.parse(await RNFS.readFile(OTA_TRANSLATIONS_FILEPATH))
+  }
+
+  const resources: Resource = {}
+  for (const [lng, value] of Object.entries(locales)) {
+    Object.defineProperty(resources, lng, {
+      get: () => ({
+        translation: cachedTranslations[lng] || value!.strings.translation,
```
silasbw, line 32: having a single root object in the cache with a single property (the language) is effectively part of the logic to store in the cache, but here the logic is kinda separate from the cache loading.  it would probably decrease the likelihood of an "oops" from a future contributor if the lng check happened above when reading from the cache (or as part of the logic loading the cache).  You might want to define a function:  `loadOtaTranslations(language: string)`

packages/mobile/src/app/I18nGate.tsx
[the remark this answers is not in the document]
kathaypacific: I removed this in the end, and start the ota saga when the feature flag request is finished

packages/mobile/src/app/saga.ts
kathaypacific: good idea! i have moved these to the i18n file :)

packages/mobile/src/i18n.ts
kathaypacific: sorry i'm not sure i understand...do you mean that something like `translation: loadOtaTranslations(lng) || value!.strings.translation` would be safer? for a filed with a singled cached language, something like `loadOtaTranslations(language: string)` would be unnecessarily reading the file multiple times to return undefined (if i understand the intention correctly), or we read the cached translations into memory and just have the `loadOtaTranslations` function handle only the `cachedTranslations[lng]` internally?

The single root object with a single language property that is written into the cache file allows us to retain the `Resource` type from i18next for the translations. This is defined to be `{ [language: string]: ResourceLanguage; }` so i'm not sure i agree that the `cachedTranslations[lng]` is very custom/unexpected?

packages/mobile/src/app/saga.ts
kathaypacific: good one! i can't remember why i added this actually, i was under the impression that `writeFile` appends instead of overwrites the file contents but this is wrong. i just tested and it looks like `writeFile` does the overwriting already, so i removed this check :)

MuckT:
@kathaypacific it looks like the text changes are failing an e2e test
Failed-e2e
:

packages/mobile/src/utils/otaTranslations.ts
```diff
@@ -0,0 +1,15 @@
+import { Resource, ResourceLanguage } from 'i18next'
+import * as RNFS from 'react-native-fs'
+import { OTA_TRANSLATIONS_FILEPATH } from 'src/config'
+
+export function saveOtaTranslations(language: string, translations: ResourceLanguage) {
+  return RNFS.writeFile(OTA_TRANSLATIONS_FILEPATH, JSON.stringify({ [language]: translations }))
+}
+
+export async function getOtaTranslations() {
```
silasbw, line 9: if `save` takes `language`, shouldn't `get` take a `language` to avoid exposing the implementation details of the cache?  i.e., the `{ [language]: ... }` structure defined in `save`.

silasbw (approved):

packages/mobile/src/utils/otaTranslations.ts
kathaypacific: i made the `saveOtaTranslations(resource: Resource)` -> `getOtaTranslations(): Resource` change we talked about, it does look much better 😀

packages/mobile/test/schemas.ts
```diff
@@ -832,6 +832,10 @@ export const v20Schema = {
     multiTokenShowHomeBalances: false,
     multiTokenUseSendFlow: false,
     multiTokenUseUpdatedFeed: false,
+    allowOtaTranslations: false,
+    otaTranslationsLastUpdate: 0,
+    otaTranslationsAppVersion: '0',
+    otaTranslationsLanguage: '',
```
jeanregisser, line 838: We should create a new schema + migration (empty migration is fine here because the app reducer handles the REHYDRATE action manually and uses the reducers initial state) for these new props.

packages/mobile/src/app/saga.test.ts
```diff
@@ -1,29 +1,70 @@
...
+
 jest.mock('src/utils/time', () => ({
   clockInSync: () => true,
 }))
 
 jest.mock('src/dappkit/dappkit')
 
-const MockedAnalytics = ValoraAnalytics as any
+const MockedAnalytics = ValoraAnalytics as jest.Mocked<typeof ValoraAnalytics>
+const MockedI18n = i18n as jest.Mocked<typeof i18n>
```
jeanregisser, line 67: nit: There's an even more simple syntax:
```suggestion
import { mocked } from 'ts-jest/utils'

const mockedAnalytics = mocked(ValoraAnalytics)
const mockedI18n = mocked(i18n)
```

packages/mobile/src/app/saga.ts
```diff
@@ -184,6 +197,9 @@ export function* appRemoteFeatureFlagSaga() {
       if (configValues) {
         yield put(updateRemoteConfigValues(configValues))
         lastLoadTime = Date.now()
+        if (configValues.allowOtaTranslations) {
+          yield put(fetchOtaTranslations())
```
jeanregisser, line 201: Though this works, we could decouple this and have a watcher for the `UPDATE_REMOTE_CONFIG_VALUES` action type. The watcher could also listen to the `SET_LANGUAGE` action type to trigger the OTA fetch.
And the explicit `fetchOtaTranslations` wouldn't be needed.

Though the current codebase is not a good example of this, it's preferable to avoid dispatching multiple actions sequentially.
See https://redux.js.org/style-guide/style-guide#avoid-dispatching-many-actions-sequentially and https://redux.js.org/style-guide/style-guide#model-actions-as-events-not-setters

packages/mobile/src/app/actions.ts
```diff
@@ -34,6 +34,8 @@ export enum Actions {
   APP_UNMOUNTED = 'APP/APP_UNMOUNTED',
   VERIFICATION_MIGRATION_RAN = 'APP/VERIFICATION_MIGRATION_RAN',
   ANDROID_MOBILE_SERVICES_AVAILABILITY_CHECKED = 'APP/ANDROID_MOBILE_SERVICES_AVAILABILITY_CHECKED',
+  SET_OTA_TRANSLATIONS_LAST_UPDATE = 'APP/SET_OTA_TRANSLATIONS_LAST_UPDATE',
```
jeanregisser, line 37: It's [recommended to model actions as events instead of setters](https://redux.js.org/style-guide/style-guide#model-actions-as-events-not-setters).

```suggestion
  OTA_TRANSLATIONS_UPDATED = 'APP/OTA_TRANSLATIONS_UPDATED',
```

packages/mobile/src/config.ts
```diff
@@ -156,6 +157,8 @@ export const CASH_IN_FAILURE_DEEPLINK = 'celo://wallet/cash-in-failure'
 
 export const APP_STORE_ID = Config.APP_STORE_ID
 export const DYNAMIC_DOWNLOAD_LINK = Config.DYNAMIC_DOWNLOAD_LINK
+export const CROWDIN_DISTRIBUTION_HASH = 'e-f9f6869461793b9d1a353b2v7c'
+export const OTA_TRANSLATIONS_FILEPATH = `file://${DocumentDirectoryPath}/translations`
```
jeanregisser, line 161: Should we use the Cache directory instead?

> Data that can be downloaded again or regenerated should be stored in the <Application_Home>/Library/Caches directory. Examples of files you should put in the Caches directory include database cache files and downloadable content, such as that used by magazine, newspaper, and map applications.

Source: https://developer.apple.com/icloud/documentation/data-storage/

packages/mobile/src/i18n.test.ts
```diff
@@ -1,58 +1,96 @@
...
+      expect(ptLoaded).toBe(false)
+    })
+
+    it('displays the bundled translation if cached translation is for a different language', async () => {
+      await i18n.changeLanguage('pt-BR')
+      expect(i18n.t('someKey')).toEqual('Oi!')
+      expect(enLoaded).toBe(false)
+      expect(esLoaded).toBe(false)
+      expect(ptLoaded).toBe(true)
+    })
```
jeanregisser, line 94: Great tests! 👍

packages/mobile/src/utils/otaTranslations.test.ts
```diff
@@ -0,0 +1,34 @@
...
+  it('should save the translations', async () => {
+    const writeToFileSpy = jest.spyOn(RNFS, 'writeFile')
+
+    await saveOtaTranslations(translation)
+
+    expect(writeToFileSpy).toHaveBeenCalledTimes(1)
+    expect(writeToFileSpy).toHaveBeenCalledWith(OTA_TRANSLATIONS_FILEPATH, stringTranslation)
+  })
+
+  it('should return the cached translations', async () => {
```
jeanregisser, line 18: nit: ```suggestion
  it('should return the cached translations when present', async () => {
```

File: packages/mobile/src/utils/otaTranslations.test.ts [hunk 18]
```diff
@@ -0,0 +1,34 @@
...
+    jest.spyOn(RNFS, 'exists').mockResolvedValueOnce(true)
+    jest.spyOn(RNFS, 'readFile').mockResolvedValueOnce(stringTranslation)
+
+    const result = await getOtaTranslations()
+
+    expect(result).toEqual(translation)
+  })
+
+  it('should return no cached translations', async () => {
```
jeanregisser, line 27:
```suggestion
  it('should return no cached translations when no present', async () => {
```

packages/mobile/src/app/saga.test.ts
```diff
@@ -279,4 +320,41 @@ describe('App saga', () => {
...
+        [call(saveOtaTranslations, { de: translations }), null],
+        ...defaultProviders,
+      ])
+      .put(setOtaTranslationsLastUpdate(timestamp, appVersion, 'de'))
+      .run()
+
+    await expectSaga(handleFetchOtaTranslations)
+      .provide([[select(otaTranslationsAppVersionSelector), '0.9.0'], ...defaultProviders])
+      .put(setOtaTranslationsLastUpdate(timestamp, appVersion, 'en-US'))
+      .run()
```
jeanregisser, line 356: Could you add comments explaining what each invocation of `handleFetchOtaTranslations` tests here so it's easier for people looking at it?

jeanregisser (commented):
Amazing work!! 🚀 ⭐ 

One thing I realize now, it seems we could/should move this new logic into a new `i18n` folder, to avoid bloating the `app` module.
But this could be done in a separate PR.

What do you think?

packages/mobile/src/app/saga.test.ts
kathaypacific: oh cool i didn't know this existed! super nice, i have updated :)

packages/mobile/src/app/saga.ts
kathaypacific: yeah good call out, i've removed the `fetchOtaTranslations` action and it's simplified the logic a lot i think :)

packages/mobile/src/app/actions.ts
kathaypacific: updated!

packages/mobile/src/config.ts
kathaypacific: updated!

packages/mobile/src/utils/otaTranslations.test.ts
kathaypacific: updated!

packages/mobile/src/utils/otaTranslations.test.ts
kathaypacific: updated!

packages/mobile/src/app/saga.test.ts
kathaypacific: updated!

packages/mobile/test/schemas.ts
kathaypacific: i've incremented the redux version and added a migration :)

kathaypacific:
> Amazing work!! 🚀 ⭐
> 
> One thing I realize now, it seems we could/should move this new logic into a new `i18n` folder, to avoid bloating the `app` module. But this could be done in a separate PR.
> 
> What do you think?

I think this is a good idea - I had a go at doing this now quickly but 

> Amazing work!! 🚀 ⭐
> 
> One thing I realize now, it seems we could/should move this new logic into a new `i18n` folder, to avoid bloating the `app` module. But this could be done in a separate PR.
> 
> What do you think?

I think this is a good idea! i will create a follow up PR for this, to avoid to many more changes to this one :)

jeanregisser (approved):

ValoraQA:
Hey @MuckT, We have verified this task on Android Alfajores Nightly, Test flight Alfajores Nightly build V1.23.0 & observed that few components & text are not getting translated.

**Bug Link:**

**Querry link:**
[Querry1](https://github.com/valora-inc/wallet/issues/1692)

Thanks..!!

PR #1445, opened 2021-11-15
Fix Moonpay webhook schema

### LINKED ISSUE #1336: Fix Moonpay schema for CICO webhook and backfill logs
### Description
Currently, the schema we use for our Moonpay webhook for CICO is out of date, which is causing us to drop webhook events. We should update the schema to match the payload that's being sent to the Webhook, and backfill any data that got dropped using the stored logs.

### ENG
- [ ] Joe says Tarik has a script to help with backfilling

### Acceptance Criteria
- [ ] Moonpay schema updated to be correct
- [ ] Old/missing events backfilled from log data

jophish:
Fixes #1336 

Updates the Moonpay webhook schema to match the data that we're actually receiving from Moonpay. The actual schemas in our BigQuery tables have been updated, and old data has already been backfilled.


cajubelt (approved):
worth adding a regression test?

silasbw (approved):

PR #1453, opened 2021-11-16
Analytics: fix balances serialization

jeanregisser:
### Description

Fix incorrect serialization of balances sent to analytics backends.

`BigNumber` implements `toJSON` so the snapshot test was actually misleading and showing the `string` serialization whereas the value sent to Segment was the raw object serialization (Example: `"sCeloBalance": { "c": Array [20], "e": 1,"s": 1 }` instead of `"sCeloBalance": "20"`).

### Tested

Updated unit tests.

### How others should test

Balance properties in Segment appear as expected.

### Related issues

- Discussed on [Slack](https://valora-app.slack.com/archives/C02A3FQ7151/p1637061161016400).

### Backwards compatibility

Yes


packages/mobile/src/analytics/selectors.ts
```diff
@@ -76,7 +76,7 @@ export const getCurrentUserTraits = createSelector(
       ...Object.fromEntries(
         (Object.keys(tokensByCurrency) as Currency[]).map((currency) => [
           `${currency === Currency.Celo ? 'celo' : currency.toLowerCase()}Balance`,
-          tokensByCurrency[currency]?.balance,
```
jeanregisser, line 79: This is the one line fix.

gnardini (approved):
💯

PR #1468, opened 2021-11-16
Adjust Simulator Parameter 

MuckT:
### Description

A very small change to `run_app.sh` to allow launching on multiple devices. This work was largely done and just needed a small tweak.

### Other changes

No other changes

### Tested

- `run_app.sh` runs the same without the parameter.
- App can be launched on multiple devices by providing simulator parameter. 
- Navigate to `packages/mobile` after running `yarn && yarn build` in the root of the repo.
```sh
# Runs by default on iPhone 11
yarn dev:ios

# Runs on an iPhone 12 mini
yarn dev:ios -s "iPhone 12 mini"

# Runs on an iPad
yarn dev:ios -s "iPad Pro (9.7-inch)"
```

### How others should test

This does not need to be tested by QA. For those reviewing the PR please ensure you can run without providing a parameter and when providing one.

### Backwards compatibility

This change is backwards compatible and should only impacts iOS dev builds when the -s or --simulator parameter is passed.


jeanregisser (approved):
Thanks!

PR #1469, opened 2021-11-17
Remove it.only blocks

MuckT:
### Description

A few stray it.only blocks were present in some tests. This PR removes them.

### Other changes

N/A

### Tested

Tests run locally with `yarn test` in `packages/mobile`.

### How others should test

N/A - Check that tests are passing in CI

### Related issues

N/A

### Backwards compatibility

This change is backwards compatible and only enables existing tests.


jeanregisser (approved):
Thanks!

PR #1471, opened 2021-11-17
Update documentation around WalletConnect and DAppKit

jeanregisser:
### Description

Make it clear that WalletConnect v1 should be the go to integration with Valora and add link to the example that @critesjosh created on Celo docs.


gnardini (approved):
🙌

Issue #1196, opened 2021-09-28
[Query] “Time out” error message appears with “Go Back” Button on scanning QR code using use-contractkit from scan section of connected apps.


ValoraQA:
**Frequency:** 100%

**Repro on build version:** Android Internal Build V1.20.1 & Test flight Build V1.20.1
**Repro on devices:** OnePlus 7T (11.0), Redmi note 8 (10.0), Redmi note 8 (9.0), iPhone 12 (14.7.1), iPhone 6+ (12.5.4)

**Pre-condition:**
1. User must have opened https://use-contractkit.vercel.app/ in Browser.
2. User must have opened valora wallet QR code by tapping on connect button.
3. User must have installed app and login into the app.
 
**Repro Steps:**
1.Launch the app.
2.Tap on “Settings” option from hamburger menu.
3. Tap on “Connected Apps” option from settings screen.
4. Tap on “Scan” button from Connected Apps screen.
5. Scan the valora wallet QR code opened in browser & observe.

**Current Behavior:** “Time out” error message appears with “Go Back” Button on scanning QR code using use-contractkit from scan section of connected apps.

**Expected Behavior:** “Time out” error message should not appear scanning QR code using use-contractkit from scan section of connected apps.

**Investigation:** This issue is repro when user scans QR code for https://ubeswap.org/ also.

**Impact:** Bad user impact.

**Attachment:** [Timeout error appears.mp4](https://drive.google.com/file/d/1FjTPQ4mc480YmyN9vSUor0hN4dABfsgP/view?usp=sharing)

gnardini:
Thanks for testing the connecting feature!
https://use-contractkit.vercel.app/ is running an outdated version and is expected not to work atm.

There are two different ways of connecting. The UI/UX is the same but we should test both because projects can use one or the other:
- WalletConnect v1 can be tested from https://app.sushi.com
- WalletConnect v2 can be tested from  https://use-contractkit-git-alexbharley-fix-wc-2-alexbharley.vercel.app/

Thanks!

ValoraQA:
https://valora.testquality.com defect [D814](https://valora.testquality.com/project/12128/result/2601472/defect/9537).

ValoraQA:
https://valora.testquality.com defect [D815](https://valora.testquality.com/project/12128/result/2601473/defect/9538).

ValoraQA:
[...]

PR #1474, opened 2021-11-17
WalletConnect: add support for eth_signTypedData_v4

jeanregisser:
### Description

This adds support for `eth_signTypedData_v4` in our WalletConnect implementation.
This is to fix removing liquidity on Sushi which uses it.

### Tested

Manually by successfully adding and removing liquidity on https://sushi.com
Also added new unit tests.

### How others should test

1. Go to https://sushi.com
2. Connect using WalletConnect (prefer scanning on Desktop as there's currently a UI refresh bug in the mobile workflow on Sushi)
3. Provide liquidity (for instance cUSD and cEUR)
4. Approve actions in Valora
5. Check your balance is updated accordingly in Valora
6. Now remove the provided liquidity
7. Approve the actions in Valora
8. See success message on Sushi
9. Check your balance is updated accordingly in Valora

### Related issues

- Discussed on [Slack](https://valora-app.slack.com/archives/C02E114V346/p1636398433018800).

### Backwards compatibility

Yes


gnardini (approved):
LGTM, but do you know what the differences are between `eth_signTypedData` and `eth_signTypedData_v4`? Looks like we're treating both the same way

jeanregisser:
Yes I see it's different implementations of the [EIP-712](https://eips.ethereum.org/EIPS/eip-712) spec.
Also https://docs.metamask.io/guide/signing-data.html#a-brief-history
But I don't fully understand the nuances and what is actually implemented on Celo.

@nategraf @nambrot could you help us and confirm we're doing the right thing?

nambrot (approved):

nategraf:
Ok, here is a summary from my research just now:
* MetaMask implemented [EIP-712](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md) at various stages of standardization. EIP-712 is ambiguously finalized right now.
* `signTypedData` is currently an alias for `signTypedData_v3` in Metamask.
* `signTypedData_v3` [does not support hashing arrays](https://github.com/MetaMask/eth-sig-util/blob/main/src/sign-typed-data.ts#L203-L207), which is currently part of the EIP-712 spec. v4 does.
* `signTypedData_v4` also supports [encoding null struct values as 32-bytes of zeroes](https://github.com/MetaMask/eth-sig-util/blob/main/src/sign-typed-data.ts#L177-L184). I've read the spec a few times now, and don't actually see this mentioned anywhere, nor have I found discussion of this anywhere.
* `signTypedData_v3` does a nonstandard trick for handling missing fields, which is it [_drops the fields from the type_](https://github.com/MetaMask/eth-sig-util/blob/main/src/sign-typed-data.ts#L247-L249). This will result in a signature failing to be verified by anyone using a fixed type hash.
* If the message being signed does not have any null values fields, no differences will result in the output. The one difference is then that v3 will throw an error if presented an array.
* If the message does have null value fields, then it will produce a signature which will be considered invalid by the receiver if they use a different method for handling null values. (e.g. is using a different version of the verification)

nategraf:
I recently [did some work on our implementation of typed data signing](https://github.com/celo-org/celo-monorepo/commit/f4dd249e9cef6a541b909c3dc7ae01888b088de2)* and added array support. If someone provides a `null` value, our implementation will not handle it, and will throw IIRC (I can confirm this if needed). So our `signTypedData` is somewhere in between v3 and v4 from MetaMask, where we _do_ support arrays but _don't_ support null values (which are non-standard). This may cause errors in to surface when trying to sign data, but will not result in invalid signatures, which is good.

TL;DR This PR looks good to me. Callers will get an error if they provide null data.

nategraf:
* caveat above is that I assume Valora is still using `sign-typed-data-utils` from `@celo/utils`.

jeanregisser:
Thanks so much for looking and explaining the details!!

> * caveat above is that I assume Valora is still using sign-typed-data-utils from @celo/utils.

Yes we are, but we're using version ~1.2.0 from what I see. Which doesn't contain your recent changes.
We should update, right?

nategraf:
In order to get array support, which apps may use, yes.

ValoraQA:
Hey @MuckT, We have verified this issue using Latest Android build V1.22.1, Test flight build V1.22.1 & updated the result in below Gsheet.

**Sheet Link:**
https://docs.google.com/spreadsheets/d/1UVmYHl_g7ZWKqX3zBtc5OdO9wEmdAdj4PW9zgwJxUys/edit#gid=481485423

**Query:** [Link](https://github.com/valora-inc/wallet/issues/1196)
CC: @jeanregisser 
Thanks..!!

PR #1482, opened 2021-11-17
chore: separate out createAccount callback

cajubelt:
The Persona component had a lot going on, and when a test failed it wasn't clear why. I separated out the piece that creates the persona account, so that it can be unit tested independently from the component itself (and mocked so the other things in the component could be tested)

This isn't perfect or done, but I think it could help as a starting place for decomposing the Persona component


jeanregisser (commented):
Quick feedback, love the intent.

Now it seems this logic could/should probably live in a saga like most of the business logic we have. This way we can easily test and decouple from the component logic.

If it's really component specific, I would extract the whole hook instead of creating a separate function which take various parameters, including react state setters which in my opinion should remain close to the component or in a hook.
I'm happy to discuss this further together.

PR #1487, opened 2021-11-18
update mobile readme google cloud instructions

kathaypacific:
### Description

When I was setting up the project locally, I found that we don't really need access to celo-testnet and there is a broken link for instructions on logging in with google credentials - I added the official google cloud docs for this as I couldn't find the corresponding readme in the celo project (I found those were helpful enough to get me started, but please let me know if there is a better link to use!)

### Other changes

N/A

### Tested

N/A

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

N/A

### DIFF
--- a/packages/mobile/README.md
+++ b/packages/mobile/README.md
@@ -58,10 +58,10 @@ brew install jq
 ```
 
 #### Google Cloud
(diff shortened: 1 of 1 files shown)


jeanregisser (approved):
Thanks! 👍

PR #1512, opened 2021-11-22
prevent whole firebase registrations registry from being loaded into memory

### LINKED ISSUE #1381: Don't load the whole Firebase registrations node into memory in the notification service
The notification service is loading the whole _registrations_ node into memory. This node has the language and firebase token for each user which we use to send push notifications. This takes a lot of memory though and has caused issues because of it in the past.  We have increased the instance size to temporarily address this.

This issue involves not doing that anymore and fetching the language and firebase token data for addresses when we actually need it instead.

AC:
* Fetch (language, fcm token) tuple on demand from Firebase RTDB (don't hold in memory!)

kathaypacific:
### Description

Instead of loading the whole firebase registrations db into memory, query the db for the relevant registration on demand when a push event needs it

### Other changes

N/A

### Tested

Running the notifications service locally, log the output of `getRegistration` on push event and validating that it is matching what is in firebase rtdb

### How others should test

Same as above

### Related issues

- Fixes #1381 
### Backwards compatibility

N/A

### DIFF
--- a/packages/notification-service/src/firebase.ts
+++ b/packages/notification-service/src/firebase.ts
@@ -107,13 +107,17 @@ export function initializeDb() {
     firebaseFetchError('Rewards senders')
   )
 
-  database.ref('/inviteRewardAddresses').on('value', (snapshot) => {
-    const addresses = snapshot?.val() ?? []
-    console.debug(`inviteRewardAddresses fetched: ${addresses}`)
-    if (addresses.length) {
-      inviteRewardsSenders = addresses
-    }
-  })
+  database.ref('/inviteRewardAddresses').on(
+    'value',
+    (snapshot) => {
+      const addresses = snapshot?.val() ?? []
+      console.debug(`inviteRewardAddresses fetched: ${addresses}`)
+      if (addresses.length) {
+        inviteRewardsSenders = addresses
+      }
+    },
+    firebaseFetchError('Invite rewards addresses')
+  )
 
   knownAddressesCache = new KnownAddressesCache()
   knownAddressesCache.startListening(database)
--- a/packages/notification-service/src/firebase.ts
+++ b/packages/notification-service/src/firebase.ts
@@ -119,9 +119,9 @@ export function initializeDb() {
   knownAddressesCache.startListening(database)
 }
 
-export function getRegistration(address: string) {
+export async function getRegistration(address: string) {
   let registration: Registration | undefined
-  registrationsRef.on(
+  await registrationsRef.once(
     'value',
     (snapshot) => {
       registration = (snapshot?.val() || {})[address]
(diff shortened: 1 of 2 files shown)


packages/notification-service/src/firebase.ts
```diff
@@ -139,12 +119,26 @@ export function initializeDb() {
   knownAddressesCache.startListening(database)
 }
 
+export function getRegistration(address: string) {
+  let registration: Registration | undefined
+  registrationsRef.on(
```
jeanregisser, line 124: I believe we should use `once` instead and make this function async.
Otherwise we create a listener while we're only interested in a single read.

packages/notification-service/test/firebase.test.ts
```diff
@@ -33,8 +33,17 @@ const mockedKnownAddressesCache = {
   },
 }
 
+const mockedRegistrationsRef = {
+  on: jest.fn((event, callback) =>
+    // @ts-ignore: Only mocking the `val` property
```
kathaypacific, line 38: Looking for suggestions on how to do this test better - now that we are not holding the db in memory anymore, i think we can mock either the database itself or the database reference (are there more options?). It is hard to mock these manually because these objects have many properties that we don't need. I saw this type of `ts-ignore` in this file already so i took a shortcut and contributed to the problem here. Is there a better way to mock the registrations? should i spend some time to find a lib that helps us with testing firebase-admin?

packages/notification-service/src/firebase.ts
kathaypacific: you're so right on both fronts, thanks for catching this! i have updated

packages/notification-service/src/firebase.ts
```diff
@@ -122,29 +104,44 @@ export function initializeDb() {
...
 }
 
-export function getTokenFromAddress(address: string) {
-  return registrations[address]?.fcmToken ?? null
+export async function getRegistration(address: string) {
+  let registration: Registration | undefined
+  await registrationsRef.once(
+    'value',
+    (snapshot) => {
+      registration = (snapshot?.val() || {})[address]
```
gnardini, line 131: This is still fetching the whole `registrations` node into memory and then getting just one value and ignoring the rest :/ 

What we should do instead is fetching the node `registrations/${address}` directly instead of fetching `registrations` and then getting the value for a single address in memory. To do this instead of the `registrationsRef` we can use `database.ref(`/registrations/${address}`)` or maybe there's a way to ask a sub-node from the `registrationsRef` directly.

packages/notification-service/src/firebase.ts
```diff
@@ -122,29 +104,44 @@ export function initializeDb() {
...
+    },
+    firebaseFetchError('Registration')
+  )
+  return registration
+}
+
+export async function getTokenFromAddress(address: string) {
```
gnardini, line 138: Doing this this way means making two requests for each transaction to send: one for fetching the language and another for fetching the fcmToken. I suggest instead fetching the registration node for an address only once on both the `sendPaymentNotification` and `sendInviteNotification` functions and adding a parameter to `sendNotification` called `fcmToken` to pass it directly we that we can avoid this :)

gnardini (commented):
Thanks for working on this!

packages/notification-service/src/firebase.ts
kathaypacific: ooohhhhh i see what you mean! thanks for this, i'm on it :)

kathaypacific:
Closing this PR as the notification service has been moved out to its own project, the new PR is https://github.com/valora-inc/notification-service/pull/2

PR #1527, opened 2021-11-23
add a patch so that when hasStable or hasCelo is undefined, we dont s…

lisaqyli:
…how cash in bottom sheet

### Description

We had a support issue when fetch balance api fails, user w/ balances would see the cash in bottom sheet. This PR explicitly checks whether `hasStable` and `hasCelo` is equal to false and we only display the bottom sheet when both value is false. Previously, an undefined value would also display the bottom sheet.

More context -> https://valora-app.slack.com/archives/C018VLA3YAK/p1637704572096200

### DIFF
--- a/packages/mobile/src/home/WalletHome.tsx
+++ b/packages/mobile/src/home/WalletHome.tsx
@@ -102,7 +102,7 @@ function WalletHome() {
       balances[currency]?.isGreaterThan(STABLE_TRANSACTION_MIN_AMOUNT)
     )
     const hasCelo = balances[Currency.Celo]?.isGreaterThan(CELO_TRANSACTION_MIN_AMOUNT)
-    const isAccountBalanceZero = !hasStable && !hasCelo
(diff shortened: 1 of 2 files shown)


packages/mobile/src/home/WalletHome.tsx
```diff
@@ -102,7 +102,7 @@ function WalletHome() {
       balances[currency]?.isGreaterThan(STABLE_TRANSACTION_MIN_AMOUNT)
     )
     const hasCelo = balances[Currency.Celo]?.isGreaterThan(CELO_TRANSACTION_MIN_AMOUNT)
-    const isAccountBalanceZero = !hasStable && !hasCelo
+    const isAccountBalanceZero = hasStable === false && hasCelo === false
```
gnardini, line 105: Ohh, quite unfortunate :/ Good catch!

gnardini (approved):

PR #1528, opened 2021-11-24
Replace default recipient wallet address with e2e test faucet address

MuckT:
### Description

Changed the `DEFAULT_RECIPIENT_ADDRESS` to the e2e test faucet address. This should help e2e tests from running low on funds.

### Other changes

N/A

### Tested

Tested with CI.

### How others should test

N/A - only impacts e2e tests no need for QA to test.

### Related issues

N/A

### Backwards compatibility

These changes only impact e2e tests and are backwards compatible.


kathaypacific (commented):
cool! 🤩

what do you think about also having the test faucet address as a constant here? then you'd only need to update this in 1 place if it were to change, and it can be used in the fund-e2e-accounts file too. something like receipientAddress = testFaucetAddress would make it super clear to read 😀

MuckT:
@kathaypacific for now this was a quick fix. Thinking about a follow up PR to populate the constants from a .env file and use the existing constants as fallbacks with the goal of being able to run e2e tests on mainnet.

kathaypacific (approved):

PR #1529, opened 2021-11-24
Fix crash when no local amount is returned by the blockchain-api

jeanregisser:
### Description

This improves our resilience to currencylayer API outages / or missing exchange rate data preventing currency conversion to happen in blockchain-api.

Now Valora gracefully degrades and tries to do currency conversions locally using the cached rate of the day.
And if there's no cached rate, it just displays `-` or nothing at all, which is not great but at least it doesn't crash.

This should be improved even more with the switch to the new transaction feed with multiple tokens where the actual underlying token would always be displayed.

### Tested

Added new unit tests.

### How others should test

This is hard to test manually because it needs a local test instance of the blockchain-api which wouldn't return the local amount.
Or an actual outage from currencylayer.

### Related issues

- See [discussion on Slack](https://valora-app.slack.com/archives/C02N3AR2P2S/p1637725955019100).
- This also addresses the following incident item from [the previous incident](https://www.notion.so/valora-inc/Valora-Transaction-Feed-Not-Loading-b34d9470388d4827a41ee49ecc7f27f8):
  - When currency conversions fail, the whole feed doesn't load at all. The UI should degrade gracefully instead.

### Backwards compatibility

Yes

### DIFF
--- a/packages/mobile/src/exchange/ExchangeConfirmationCard.test.tsx
+++ b/packages/mobile/src/exchange/ExchangeConfirmationCard.test.tsx
@@ -3,25 +3,33 @@ import * as React from 'react'
 import 'react-native'
 import { Provider } from 'react-redux'
 import ExchangeConfirmationCard from 'src/exchange/ExchangeConfirmationCard'
-import { createMockStore } from 'test/utils'
+import { createMockStore, getElementText } from 'test/utils'
 
 const localAmount = {
   value: '1.23',
   exchangeRate: '0.555',
   currencyCode: 'EUR',
 }
-const makerAmount = { value: '20', currencyCode: 'cGLD', localAmount }
-const takerAmount = { value: '1.99', currencyCode: 'cUSD', localAmount }
 
 const store = createMockStore({})
 
-it('renders correctly with no exchange rate', () => {
+it('renders correctly', () => {
   const tree = render(
     <Provider store={store}>
-      <ExchangeConfirmationCard makerAmount={makerAmount} takerAmount={takerAmount} />
+      <ExchangeConfirmationCard
+        makerAmount={{ value: '20', currencyCode: 'cGLD', localAmount }}
+        takerAmount={{ value: '1.99', currencyCode: 'cUSD', localAmount }}
+      />
     </Provider>
   )
   expect(tree).toMatchSnapshot()
+
+  expect(getElementText(tree.getByTestId('CeloAmount/value'))).toEqual('20.000')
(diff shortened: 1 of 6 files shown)


packages/mobile/src/transactions/GoldTransactionFeedItem.tsx
```diff
@@ -35,16 +35,18 @@ export function ExchangeFeedItem(props: Props) {
   const isPending = status === TransactionStatus.Pending
   // We always show Local Currency to cGLD exchage rate
   // independent of transaction type
-  const localAmount = (isSellGoldTx ? makerAmount : takerAmount).localAmount!
```
jeanregisser, line 38: This was the cause of the crash in the feed.

packages/mobile/src/exchange/ExchangeConfirmationCard.tsx
```diff
@@ -34,16 +34,18 @@ export default function ExchangeConfirmationCard(props: Props) {
   const totalFee = new BigNumber(tobinTax).plus(fee)
   const feeCurrency = Currency.Dollar
 
-  const localAmount = (isSellGoldTx ? makerAmount : takerAmount).localAmount!
```
jeanregisser, line 37: This would cause a crash when tapping the exchange feed item when no local amount is set.

packages/mobile/src/exchange/ExchangeConfirmationCard.test.tsx
```diff
@@ -3,25 +3,33 @@ import * as React from 'react'
...
+        takerAmount={{ value: '1.99', currencyCode: 'cUSD', localAmount }}
+      />
     </Provider>
   )
   expect(tree).toMatchSnapshot()
+
+  expect(getElementText(tree.getByTestId('CeloAmount/value'))).toEqual('20.000')
+  expect(getElementText(tree.getByTestId('CeloExchangeRate/value'))).toEqual('€0.55')
+
+  expect(getElementText(tree.getByTestId('TotalLineItem/Total/value'))).toEqual('$2.64')
```
dievazqu, line 30: nit: I'm not sure how these values are obtained, maybe a small comment would help to understand it

gnardini (approved):

packages/mobile/src/exchange/ExchangeConfirmationCard.test.tsx
jeanregisser: yep you're right it's quite confusing given the fake localAmount used for testing.

PR #662, opened 2021-07-05
Use buy and sell methods of Celo exchange contract

Pedro-vk:
### Description

_Use buy and sell methods of Cello contracts doing the exchange more accurate_

### Tested

_Tested on Android 11. Going to Celo section, performing buy and sells, and checking the amount on Blockscout to ensure that the Celo is the amount typed on the exchange input._

### How others should test

_Buy and sell Celo and check it on Blockscout._

### Related issues

- Fixes #792

PR #1535, opened 2021-11-25
Update detox version

MuckT:
### Description

Update [Detox](https://www.npmjs.com/package/detox) version from `18.20.3` to `19.1.0`.

### Other changes

- Updated `getDeviceModel()` util function to work with new detox version. This also appears to fix [Detox Recorder](https://www.npmjs.com/package/detox-recorder) launch which would occasionally fail.
- Skipped a failing exchange test as the minimum amount allowed fluctuates with the price of CELO.

### Tested

Tested locally on Android and iOS.

### How others should test

No need for QA to test. 
Reviewer should be able to run Detox Recorder - [recording a test](https://github.com/valora-inc/wallet/tree/main/packages/mobile/e2e#recording-a-test).

### Related issues

N/A

### Backwards compatibility

This change is backwards compatible with existing e2e tests.


packages/mobile/e2e/src/usecases/ExchangeCelo.js
```diff
@@ -120,8 +120,9 @@ export default ExchangeCelo = () => {
     // TODO Check that transaction appears in home feed
   })
 
+  // Note: Amount fluctuates based on CELO value
   jest.retryTimes(2)
-  it('Then Sell Minimum CELO', async () => {
+  it.skip('Then Sell Minimum CELO', async () => {
```
jeanregisser, line 125: I wonder if this will be fixed with https://github.com/valora-inc/wallet/pull/662
I'll try it.

PR #1539, opened 2021-11-26
Reset app state when keychain/keystore has no matching password hash

### LINKED ISSUE #729: Restoring from iCloud puts user into bad state
### Current Behavior

1. Set up new device from iCloud
2. Valora was auto-downloaded (I _think_ it downloaded 1.3.0 even though my prior phone was using the Testflight build, not sure)
3. Opening it up, I was dropped into Home and could see contacts in my feed so no onboarding
4. I didn't poke around much at the time because I kept testing on the original device
5. At some point updated to latest Testflight build
6. Today I tried to send a transaction and got an Incorrect PIN message
7. Went to Account Key to trigger PIN again, also got Incorrect PIN error, was taken to Account Key screen anyway, but it displayed an empty Account Key field

Though I could see my feed, it seems like my account was maybe not restored fully. Deleting and re-installing (and restoring with onboarding) fixed things.

### Expected Behavior

Auto-log user out if we are unable to detect an account phrase for them.

From chat with @tarikbellamine: "so what happens is we have app state but then we use the phone’s secure storage for some stuff like your account phrase so app state is being restored, but not the stuff we are storing on your phone... that would make it unusable. sounds like we should auto-log the user out if we are unable to detect an account phrase for them"

jeanregisser:
### Description

This PR fixes an incorrect state causing all account key unlock operations to fail.
This can happen specifically after restoring from backups. See #729.

### Other changes

- Added type to persistConfig.

### Tested

Manually and added unit tests.

### How others should test

See steps to reproduce in #729

### Related issues

- Fixes #729

### Backwards compatibility

Yes


gnardini (approved):
LGTM

packages/mobile/src/utils/accountChecker.test.ts
```diff
@@ -0,0 +1,63 @@
...
+  jest.clearAllMocks()
+})
+
+describe('resetStateOnInvalidStoredAccount', () => {
+  it("returns the same state when there's no account", async () => {
+    const state = getMockStoreData({ web3: { account: null } })
+    expect(walletAddressSelector(state)).toEqual(null)
+    const result = await resetStateOnInvalidStoredAccount(state)
+
+    expect(result === state).toEqual(true)
```
kathaypacific, line 23: this is the first time i've seen this 😄  it clearly works but is the more conventional form `expect(result).toEqual(state)`?

kathaypacific (approved):
Also LGTM 🚢

packages/mobile/src/utils/accountChecker.test.ts
```diff
@@ -0,0 +1,63 @@
...
+import { resetStateOnInvalidStoredAccount } from 'src/utils/accountChecker'
+import { walletAddressSelector } from 'src/web3/selectors'
+import { getMockStoreData } from 'test/utils'
+import { mocked } from 'ts-jest/utils'
+
+jest.mock('src/geth/geth')
+
+const mockedKeychain = mocked(Keychain)
+
+beforeEach(() => {
```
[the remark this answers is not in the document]
jeanregisser: No, this still runs before each individual test.
[Jest’s ‘beforeEach’ may not be running the way you think it does…](https://joshua-toth.medium.com/jests-beforeeach-may-not-be-running-the-way-you-think-it-does-c81599d83649)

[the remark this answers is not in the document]
kathaypacific: my whole life has been a lie 🙈

packages/mobile/src/utils/accountChecker.test.ts
jeanregisser: Yes I wanted to avoid ambiguity for readers 😄 
I initially wrote it as `expect(result).toBe(state)` which uses `Object.is` under the hood.
But I really wanted identity comparison. `toEqual` doesn't fit either.
https://stackoverflow.com/a/63003041/158525

kathaypacific: ah, i see what you're trying to do now! makes sense

PR #1540, opened 2021-11-26
Remove duplication in v0Schema

jeanregisser:
### Description

`v0Schema` was copying most props from `vNeg1Schema` making it hard to see what was different.

This is in preparation for the switch to PHP as local currency for unit tests.

### How others should test

No need to test this.

### Backwards compatibility

Yes


kathaypacific:
LGTM 🚢

kathaypacific (approved):

PR #1541, opened 2021-11-26
Make PHP the default local currency for unit tests

jeanregisser:
### Description

The default local currency was set to MXN which unfortunately uses the same $ symbol as the USD $. 
This makes it harder to understand unit tests displaying converted money amounts.
PHP is now used instead which uses ₱ as symbol.

### Tested

Updated unit tests.

### How others should test

No need to test.

### Related issues

- Discussed on [Slack](https://valora-app.slack.com/archives/CL7BVQPHB/p1637765197181800).

### Backwards compatibility

Yes


gnardini (approved):
Nice!

PR #1542, opened 2021-11-26
move i18n related logic to separate folder and create separate i18n redux store

kathaypacific:
### Description

As suggested by Jean, to avoid bloating the app folder/state then we should move all i18n related logic to a separate folder. In this PR:
- create new i18n folder and move existing i18n/ota translations files + logic into this new folder
- create new i18n redux state, move language and ota translations state into this store
- update redux migrations
- add useChangeLanguage hook to manage dispatching the `setLanguage` action and calling `changeLanguage` on i18n instance. Previously, we were dispatching the action from components and the action calls `changeLanguage` but since the new i18n store is created using redux toolkit, we cannot do this anymore (actions are auto created by the slice)
- add react testing library for hooks, add test for new useChangeLanguage hook

### Other changes

N/A

### Tested

Manually - click around the app and see it still works. Also tested new install, changing languages, fetching OTA translations

### How others should test

Same as above, if you want

### Related issues

N/A

### Backwards compatibility

N/A


packages/mobile/src/firebase/saga.ts
```diff
@@ -81,18 +83,18 @@ function* initializeFirebase() {
   }
 }
 
-export function* syncLanguageSelection({ language }: SetLanguage) {
+export function* syncLanguageSelection(action: PayloadAction<string>) {
```
jeanregisser, line 86: I propose we do the following to avoid the type to diverge from the original action:
```suggestion
export function* syncLanguageSelection(action: ReturnType<typeof setLanguage>) {
```

packages/mobile/src/i18n/i18nSlice.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const initialState: State = {
+  language: null,
+  allowOtaTranslations: REMOTE_CONFIG_VALUES_DEFAULTS.allowOtaTranslations,
+  otaTranslationsLastUpdate: 0,
+  otaTranslationsAppVersion: '0',
+  otaTranslationsLanguage: '',
+}
+
+export interface OtaTranslationsUpdatedAction {
```
jeanregisser, line 22: Looks like this doesn't need to be exported.

packages/mobile/src/i18n/i18nSlice.ts
```diff
@@ -0,0 +1,60 @@
...
+  otaTranslationsLastUpdate: number
+  otaTranslationsAppVersion: string
+  otaTranslationsLanguage: string
+}
+
+export const i18nSlice = createSlice({
```
jeanregisser, line 28: Should we just name is `slice`? and same for the file?

packages/mobile/src/i18n/useChangeLanguage.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+  it('should handle changing language', async () => {
+    const setLanguageSpy = jest.fn().mockResolvedValue(true)
+    mockedI18n.changeLanguage.mockImplementationOnce(setLanguageSpy)
+
+    const { result } = renderHookWithProvider()
+    await result.current('pt-BR')
+
+    expect(setLanguageSpy).toHaveBeenCalledWith('pt-BR')
+    expect(changeLanguageSpy).toHaveBeenCalledWith('pt-BR')
+  })
```
jeanregisser, line 41: Can we check that the redux action is actually dispatched?

packages/mobile/src/i18n/useChangeLanguage.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+jest.mock('src/i18n')
+
+const mockedI18n = mocked(i18n)
+
+const changeLanguageSpy = jest.spyOn(I18nSlice, 'setLanguage')
+const loggerErrorSpy = jest.spyOn(Logger, 'error')
+
+const renderHookWithProvider = () => {
+  const store = createMockStore()
+  return renderHook(() => useChangeLanguage(), {
```
jeanregisser, line 20: cool for testing hooks! 👍

packages/mobile/src/redux/migrations.ts
```diff
@@ -329,5 +329,14 @@ export const migrations = {
...
+  22: (state: any) => ({
+    ...state,
+    i18n: {
+      language: state.app.language,
+      allowOtaTranslations: false,
+      otaTranslationsLastUpdate: 0,
+      otaTranslationsAppVersion: '0',
+      otaTranslationsLanguage: '',
+    },
+  }),
```
jeanregisser, line 341: Can you add a test for this migration?
Also as good practice we should set `state.app.language` to `undefined`.

jeanregisser (commented):
Amazing!! Love the boilerplate reduction 💯 👍

packages/mobile/src/i18n/i18nSlice.ts
kathaypacific: ah yes you're right! i've removed 🙈

packages/mobile/src/i18n/useChangeLanguage.test.tsx
kathaypacific: it is! i don't know what is wrong with my brain today....but i named these the wrong way around. i've updated this now - the `setLanguageSpy` is watching the setLanguage action on the i18n slice (and `changeLanguageSpy` watching the i18n instance) :)

packages/mobile/src/redux/migrations.ts
kathaypacific: yes i've added this now :) and good shout about the setting language to undefined...i did this for the schema file, but not for this 🙈  my brain .....

as an aside, i was really tempted to do a one liner `expect(migratedSchema).toEqual(v22Schema)` since from the schemas it is already easy to visualise what is changed. but i noticed not many other tests were doing this and i didn't want to look lazy, so i asserted just the changes i was looking for. is there any reason why we don't have a loop test that just goes through every migration and does the `expect(migratedSchema).toEqual(v(n-1)Schema)`? you reckon it'd slow the test run time too much?

packages/mobile/src/i18n/i18nSlice.ts
kathaypacific: nice, i updated this.

i find generic names a bit harder to work with when there are many files open (e.g. when they all say reducer.js in the tab title) so was a natural reflex that i did the specific naming like this :/ but indeed consistency is key and we already have the generic naming pattern in the project so we should keep this :)

packages/mobile/src/i18n/useChangeLanguage.test.tsx
jeanregisser: Looks good now.

packages/mobile/src/redux/migrations.test.ts
```diff
@@ -385,4 +386,16 @@ describe('Redux persist migrations', () => {
     expect(migratedSchema2.walletConnect.v1).toBe(v18Schema.walletConnect.v1)
     expect(migratedSchema2.walletConnect.v2).toBeUndefined()
   })
+  it('works for v21 to v22', () => {
+    const migratedSchema = migrations[22](v21Schema)
+
+    expect(migratedSchema.i18n).toEqual({
+      language: v21Schema.app.language,
```
jeanregisser, line 393: nit: what if `v21Schema.app.language` was actually `undefined`? I know it's not the case here but it'd slightly better to use the actual string. in case it changes in the future we would spot it.

packages/mobile/src/redux/migrations.ts
jeanregisser: There's another test which tries all the migrations from the initial schema to be sure they all succeed: https://github.com/valora-inc/wallet/blob/3062f866e458406ca38084c87a0c7153fcc0ea9c/packages/mobile/src/redux/createMigrate.test.ts#L19

Sometimes depending on the migration we need more complex test cases.
But I like what you're proposing! Not sure we've been thorough enough to make it work for all schemas, but feel free to try and open a PR.

jeanregisser (approved):

ValoraQA:
Hey @MuckT, We have verified this task on Android Alfajores Nightly build & iOS Test flight Alfajores Nightly build V1.23.0 & observed that few components & text are not seen translated.

**Bug Link:**

**Query link:**
[Query1](https://github.com/valora-inc/wallet/issues/1692)

CC: @kathaypacific
Thanks..!!

Issue #1521, opened 2021-11-23
Continuous loading appears on “Awaiting Payment” screen in Bidali page & User is blocked to buy gift card.


ValoraQA:
**Frequency:** 100%

**Repro on build version:** iOS Alfajores V1.22.1
**Repro on devices:** iPhone 12 (15.0), iPhone 6+ (12.5.4)

**Does Not Repro on build version:** iOS Alfajores 1.21.0
**Does Not Repro on devices:** iPhone 12 (15)
**Note:** For android alfa jores build V1.22.1 is not available.

**Pre-condition:** 
1. User must have installed the app and completed onboarding flow.
2. User must have sufficient balance in the account.
3. User must be on bidali screen. 

**Repro Steps:**
1. Tap on any brand.
2.  Select any option of gift card amount.
3. Tap on “Purchase” button.
4. Enter valid email address.
5. Tap on “Next” button & Observe.

**Current Behavior:** Continuous loading appears on “Awaiting Payment” screen in bidali page & User is blocked to buy gift card.

**Expected Behavior:**  Continuous loading must not appear on “Awaiting Payment” screen in bidali page & User must be able submit order.

**Impact:** Bad user experience as user is blocked to submit order

**Attachment:** [Not able to buy Gift Card.mp4](https://drive.google.com/file/d/1UKgG6EXSycVQ_rXZ83tTGbcjt_jG9YUX/view?usp=sharing)

ValoraQA:
https://valora.testquality.com defect [D672](https://valora.testquality.com/project/12128/result/2573987/defect/9312).

MuckT:
@ValoraQA can you attach logs for this issue?

ValoraQA:
Hey @MuckT, We have attached the logs for the issue below.

**Attachment:** [Celo logs.txt](https://drive.google.com/file/d/1sq721A0I0yC33gmQyqVWZKYebNYi_4LU/view?usp=sharing) 
**Device:** iPhone 12 (15.0)
Thanks..!!

ValoraQA:
Hey @MuckT, We have verified this issue using Latest iOS Alfajores Build V1.22.2 & observed that the issue is fixed now.

**Device:** iPhone 12(14.7.1)
**Attachment**:[Able to buy gift card.mp4](https://drive.google.com/file/d/1DpLfkc6eZVuBS-ohqQxHL5bj8Qi3B1KH/view?usp=sharing) 
CC: @jeanregisser 
Thanks..!!

PR #1501, opened 2021-11-19
Convert Payment Request Confirmation to Functional Component

MuckT:
### Description

Changes payment request confirmation from a class based component to a functional component.

### Other changes

N/A

### Tested

Locally tested on iOS and Android and existing unit tests are passing.

### How others should test

Requester: account making the request
Giver: account receiving the request

#### Payment Request Successful

1. From requester send a payment request to the giver - no comment.
2. On requester see that the request is populated in the requests call to action (CTA).
3. From giver see that the request is received.

#### Comments
1. From requester add a comment and send payment request.
2. On requester see that the request is populated in the requests call to action (CTA).
3. From giver see that the request is received.

#### Back
1. From requester go back from the payment request screen
2. Request amount input loads.

### Related issues

Prerequisite for - #1395

### Backwards compatibility

This change is backwards compatible.

PR #1554, opened 2021-11-30
Add Test for Bidali Send Flow

MuckT:
### Description

Adds a test for the Bidali web view purchase flow. Currently testing web views with Detox is currently [only available in Android](https://github.com/wix/Detox/releases/tag/18.7.0). I've also used the `:android:` tag in the test name to enable this test only on Android. This test would have caught #1521; as this defect was found late in manual regression testing process it makes sense to have an e2e test for it.

### Other changes

No other changes

### Tested

Tested locally with e2e test on Android.

### How others should test

### Related issues

- Tests #1521 #1501

### Backwards compatibility

This change only impacts e2e tests and is backwards compatible.


MuckT:
These tests pass locally, but fail in CI. This is might be a good candidate for [mocking](https://github.com/wix/Detox/blob/master/docs/Guide.Mocking.md) the responses from Bidali.

MuckT:
Closing this PR as the interactions with webviews in Android are failing in CI 😢 . I recommend we mock the expected responses from all providers and use those in e2e tests instead.

Issue #1557, opened 2021-12-01
Operation Translations Failing to Populate in WalletConnect


MuckT:
### Current behavior
Operation translations are failing to populate in Wallet Connect.

### Desired behavior
Operation translations should be present on the Wallet Connect `ActionRequest.tsx` screen.

### Steps to Reproduce
1. Connect to a Dapp using WalletConnect like [PoolTogether](https://app.pooltogether.com/).
2. Accept the session.
3. Attempt to perform an action that would trigger the `ActionsRequest.tsx` screen e.g. adding to a pool.
4. Observe section under 'Operation' is not populated.

### Screenshot
<img src="https://user-images.githubusercontent.com/26950305/144338460-f836156c-c107-4c2c-9bc5-1f37e5b8ce84.png" alt="Wallet Connect Action Request" width="50%" height="auto">

### Versions
v1.22.2 & v1.22.3 Does not reproduce on 1.21.0

ValoraQA:
Hey @MuckT, We have verified this issue on Latest Test flight build V:1.22.4 and observed that issue is fixed now.
Observation: 
- "Send a CELO tx" text is displayed under the operation section and also after tapping on "Allow"  button the transaction is completed successfully.
**Device:** iPhone XR(14.6), iPhone SE(13.6.1)
**Attachments:** 
[Operation Section text.jpg](https://drive.google.com/file/d/1UjfjS_4ji5nzEY5sY-zif_hGg_qdtHXu/view?usp=sharing)
[Transaction Successfull.jpg](https://drive.google.com/file/d/1DQYhkodqu1MacxNl61nfnRFJ3J9yK_iH/view?usp=sharing)
Thanks..!!

ValoraQA:
Hey @MuckT we have commented in the ticket #1563

PR #1561, opened 2021-12-02
Fix missing translations in ActionRequest

kathaypacific:
### Description

Pass the `t` function from components to the static functions used to generate translations, to prevent translated strings from being initialised before the i18n instance is initialised and empty strings being rendered. 

### Other changes

N/A

### Tested

Manually

### Related issues

- Fixes https://github.com/valora-inc/wallet/issues/1557

### Backwards compatibility

N/A

### DIFF
--- a/packages/mobile/src/walletConnect/constants.ts
+++ b/packages/mobile/src/walletConnect/constants.ts
@@ -1,4 +1,4 @@
-import i18n from 'src/i18n'
+import { TFunction } from 'i18next'
 
 export enum SupportedActions {
   eth_accounts = 'eth_accounts',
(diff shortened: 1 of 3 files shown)


jeanregisser (commented):
Thanks!

packages/mobile/src/walletConnect/constants.ts
```diff
@@ -12,35 +12,23 @@ export enum SupportedActions {
...
-  [SupportedActions.eth_signTypedData_v4]: i18n.t('action.sign'),
-  [SupportedActions.eth_sign]: i18n.t('action.sign'),
-  [SupportedActions.personal_sign]: i18n.t('action.sign'),
-  [SupportedActions.personal_decrypt]: i18n.t('action.decrypt'),
-  [SupportedActions.computeSharedSecret]: i18n.t('action.computeSharedSecret'),
-}
-
 export function isSupportedAction(action: string) {
-  return action in actionTranslations
+  return Object.values(SupportedActions).includes(action as SupportedActions)
```
kathaypacific, line 16: @jeanregisser this change i'm iffy about, what do you think

jeanregisser: Looks good! 
Nothing to do now, we could have skipped the enum and used `type SupportedActions = "eth_accounts" | ...`

jeanregisser (approved):

ValoraQA:
Hey @MuckT we have updated the status in the ticket #1563

PR #1565, opened 2021-12-03
merge bundled translations with cached OTA translations on app load

kathaypacific:
### Description

During development you may need to add a new translation key as part of your new feature. While this feature/new translation key is not merged to main, it will not be registered in Crowdin and will therefore not come through as part of the OTA translations. As a result you'll see raw translation keys during development which doesn't give confidence that the feature is working correctly.

In this PR, when loading the resources, use the bundled translations as the base and merge cached OTA translations on top so that we can pick up the OTA changes but also have new translation keys that were added in the app.

Note that this means we always load the bundled resource for the selected language (as opposed to before when we could save this resource load if there are cached translations)

### Other changes

N/A

### Tested

Manually add a new key to the translation.json in feature branch and verify that it is displayed correctly in the app

### How others should test

Same as above

### Related issues

N/A

### Backwards compatibility

N/A

### DIFF
--- a/packages/mobile/README.md
+++ b/packages/mobile/README.md
@@ -1,40 +1,42 @@
 # Mobile (Valora)
 
-  - [Overview](#overview)
-  - [Architecture](#architecture)
-  - [Setup](#setup)
-    - [iOS](#ios)
-      - [Enroll in the Apple Developer Program](#enroll-in-the-apple-developer-program)
-      - [Install Xcode](#install-xcode)
-      - [Install Cocopods, Bundler, and download project dependencies](#install-cocopods-bundler-and-download-project-dependencies)
-    - [Android](#android)
-      - [Install Java](#install-java)
-      - [Install Android Dev Tools](#install-android-dev-tools)
-      - [Optional: Install an Android emulator](#optional-install-an-android-emulator)
-  - [Running the mobile wallet](#running-the-mobile-wallet)
-    - [iOS](#ios-1)
-    - [Android](#android-1)
-    - [Running in forno (data saver) mode](<#running-in-forno-(data-saver)-mode>)
-  - [Debugging & App Profiling](#debugging--app-profiling)
-    - [Debugging](#debugging)
-      - [Optional: Install React Native Debugger](#optional-install-react-native-debugger)
-    - [App Profiling](#app-profiling)
-  - [Testing](#testing)
-    - [Snapshot testing](#snapshot-testing)
-    - [React component unit testing](#react-component-unit-testing)
-    - [Saga testing](#saga-testing)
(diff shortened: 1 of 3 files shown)


jeanregisser (commented):
Looks great!

Now I think there's one additional edge case which may be confusing:
Someones edits the value of an existing English string during development. It will still display the OTA value. Leading to the developer jumping out of the window: why am I not seeing the string I've set 😄 

I'm thinking we could display a big WARNING in the app log in dev mode to warn about this.
```
Crowdin Over-the-Air update is currently enabled, existing values which are updated in the localized string files WILL NOT reflect in the app.
```
or something like this.

kathaypacific:
> Looks great!
> 
> Now I think there's one additional edge case which may be confusing: Someones edits the value of an existing English string during development. It will still display the OTA value. Leading to the developer jumping out of the window: why am I not seeing the string I've set 😄
> 
> I'm thinking we could display a big WARNING in the app log in dev mode to warn about this.
> 
> ```
> Crowdin Over-the-Air update is currently enabled, existing values which are updated in the localized string files WILL NOT reflect in the app.
> ```
> 
> or something like this.

I think this is a good idea! I tried adding something like
```
useEffect(() => {
    if (__DEV__ && allowOtaTranslations) {
      Logger.warn(
        TAG,
        'WARNING! Crowdin Over-the-Air update is currently enabled, existing values which are updated in the localized string files WILL NOT reflect in the app.'
      )
    }
  }, [allowOtaTranslations])
```
to I18nGate but the problem is we have soooo much logging that it's easy to miss this (unless i get quite creative and make the message like 
```
###########
## BEWARE ##
###########
```
?

I was also thinking maybe OTA translations should be turned off in dev mode...or we can do `_.merge(otaTranslations, bundledTranslations)` in dev mode to take the bundled translations over the ota ones? Having dev builds handle translations differently to production builds is not really ideal though.

Or maybe I can update our readme with a paragraph about this so that only devs who don't read the readme will tear out their hair when trying to update translations locally? What do you think 🙈

jeanregisser:
Yes the log is quite easy to miss 😅 
I agree with you, we should avoid differences in dev mode unless really necessary.

I really like adding a note about this in our README.
We could explain the few differences we have (this edge case and also analytics being disabled come to mind).

jeanregisser (approved):
Shiiip it 👍 :shipit: 🐑 🚢

ValoraQA:
Hey @MuckT, We have verified this task on Android Alfajores Nightly build & iOS Test flight Alfajores Nightly build V1.23.0 & observed that few components & text are not seen translated.

**Bug Link:**

**Query link:**
[Query1](https://github.com/valora-inc/wallet/issues/1692)

CC: @kathaypacific 
Thanks..!!

Issue #1574, opened 2021-12-06
Show the name of the invitee correctly in the transaction feed


gnardini:
When inviting someone, the transaction feed shows the item as 'Unknown'. We need to query the blockchain to know which identifier was invited in that tx and use the local identifier to recipient cache to show the right name.

Also, when reclaiming invites it shows as unknown with a weird message. It would be cool to fix that too.

kathaypacific:
Closing this as a duplicate of #2215

Issue #1573, opened 2021-12-06
Add Standby Transactions to multi-token support Transaction Feed


gnardini:
In the new transaction feed with multi-token support standby transactions don't show up. We need to fix that.

There are a few places where we need to make changes:

- In the send saga: https://github.com/valora-inc/wallet/blob/main/packages/mobile/src/send/saga.ts#L270 we need to add a "standby transaction" to the Redux store.
- In the invites saga: https://github.com/valora-inc/wallet/blob/main/packages/mobile/src/invite/saga.ts#L69 we also need to add a standby transaction.
- Also add a standby transaction when withdrawing CELO from the CELO tab and when buying/selling CELO
- In the Transaction Feed (this PR: https://github.com/valora-inc/wallet/pull/1575) we need to append these standby transactions.
- When the transaction finishes (https://github.com/valora-inc/wallet/blob/main/packages/mobile/src/transactions/saga.ts#L102 ) we are already dispatching an action! We need to check that the `status` field of the transaction is `true` and show the transaction as confirmed when that happens instead of 'Confirming'.
- Every time a new transaction is returned from Blockscout we are already checking for (legacy) standby transactions and removing them if found. We need to do the same for the new standby transactions as well
: https://github.com/valora-inc/wallet/blob/main/packages/mobile/src/transactions/saga.ts#L211

Issue #1572, opened 2021-12-06
Transaction Feed: Add multi-token support to Transaction Details Screen


gnardini:
In #1262 we started implementing a new Transaction Feed with multi-token support.

When tapping on an item, it's not open the detail screen. We need to implement it and open it. We can either reuse existing ones or create a new one.

Work on this is in progress in this PR: https://github.com/valora-inc/wallet/tree/transaction-feed-details
For now it works for the Payment Sent transactions but is missing payments received, invite sent/received and exchanging. Also tests.

PR #1575, opened 2021-12-06
Add Transaction Feed with multi-token support

### LINKED ISSUE #1262: Show token transfer feed in the app
After #1132, we need to make changes to the app to show transactions in any token.

Here the decision to make is if we want to reuse the existing `TransactionsList` or create a new one. The existing one is still a class component and is based around the `Currency` enum. My suggestion is to create a new component to show txs based on a `tokenAddress` instead and use the existing code as a template and copy the relevant parts.

Acceptance criteria (can take several PRs):
1. Create new component for TransactionList that is based on tokenAddress instead of currencyCode.
2. Make changes to supporting components like the Transaction Drilldown screen.
3. Make sure pending transactions (just after they are sent) are rendered correctly.
4. Make sure existing e2e tests work and consider adding new ones.
5. Delete the code for the legacy component.

gnardini:
### Description

I know what you're thinking "1600 lines? That's way too much!" BUT over 1k lines are tests and snapshots and this is already a small subset of the feature. I couldn't find a way of meaningfully reduce the number of changed lines but if you have suggestions on what to move to another PR I'm happy to do it if you think that will help reviewing.

This PR is the first step to add multi-token support to the transaction feed. The relevant changes are mostly in three files:
- `packages/mobile/src/transactions/feed/TransferFeedItem.tsx` (new)
- `packages/mobile/src/transactions/feed/TransactionFeed.tsx` (new)
- `packages/mobile/src/transactions/transferFeedUtils.ts` (added 2 new hooks)

There are a few things missing which I opened new issues for:
- Transaction detail: https://github.com/valora-inc/wallet/issues/1572
- Standby transactions: https://github.com/valora-inc/wallet/issues/1573
- Invite names: https://github.com/valora-inc/wallet/issues/1574

### Other changes

Removed the invites reducer since it was never being written to :( In production right now you see `Unknown` from the get go. I added an issue to fix this.

### Tested

Manually and from unit tests.

### How others should test

Not ready to be tested, we need to implement the listed issues first, but to test we need to check that the transaction feed always show the expected transactions and that we can see the details by tapping on it.

### Related issues

- Fixes #1262

### Backwards compatibility

N/A


packages/mobile/src/transactions/feed/TransactionFeed.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const address = useSelector(walletAddressSelector)
+  const localCurrencyCode = useSelector(getLocalCurrencyCode)
+  const dispatch = useDispatch()
+
+  // Update the counter variable every |POLL_INTERVAL| so that a query is made to the backend.
+  const [counter, setCounter] = useState(0)
+  useEffect(() => {
+    const timer = setInterval(() => setCounter((n) => n + 1), POLL_INTERVAL)
+    return () => clearInterval(timer)
+  }, [])
```
kathaypacific, line 43: We use this setInterval pattern a lot in the project already but I was wondering if you had any thoughts about a custom hook like `useInterval`? Our favourite Dan Abramov has a nice custom hook that we could use, it helps to eliminate some edge case bugs that are not an issue here but might be helpful for the future and it'd reduce the logic in this hook :) https://overreacted.io/making-setinterval-declarative-with-react-hooks/

packages/mobile/src/transactions/feed/TransactionFeed.test.tsx
```diff
@@ -0,0 +1,126 @@
...
+    mockFetch.mockResponse(JSON.stringify(MOCK_RESPONSE))
+
+    const tree = renderScreen({})
+
+    await waitFor(() => tree.getByTestId('TransactionList'))
+
+    expect(tree.queryByTestId('NoActivity/loading')).toBeNull()
+    expect(tree.queryByTestId('NoActivity/error')).toBeNull()
+
+    expect(tree).toMatchSnapshot()
```
kathaypacific, line 83: This is probably a wider discussion but thought this might be a good conversation starter - do we want to keep using snapshot tests? My personal opinion is that they capture mostly implementation detail, and while sometimes during a refactor it can be a nice prompt that this component was affected, most of the time we will update sanpshot tests without a second thought and the snapshot updates are not reviewed. The end result is that they add extra maintainence overhead and don't really give us any more confidence that the component is working correctly...they also do a less good job of acting as documentation (it's easy to see that in this scenario we should not see the loading or error states in the first 2 `expect` statements) 

so TL;DR I think that the rest of the tests are great and this snapshot isn't really needed - what do you think?

packages/mobile/src/geth/networkConfig.ts
```diff
@@ -89,6 +89,7 @@ const networkConfigs: { [testnet: string]: NetworkConfig } = {
     nodeDir: `.${Testnets.alfajores}`,
     syncMode: DEFAULT_SYNC_MODE,
     initiallyForno: FORNO_ENABLED_INITIALLY,
+    // blockchainApiUrl: 'http://127.0.0.1:8080',
```
kathaypacific, line 92: just checking that you want to keep this?

packages/mobile/src/redux/migrations.ts
```diff
@@ -344,4 +344,5 @@ export const migrations = {
     },
   }),
   23: (state: any) => state,
+  24: (state: any) => state,
```
kathaypacific, line 347: wondering if you want to do a `invites: undefined` in the state? i guess it is not necessary but would be nice to have it here or in the test schemas file to see what changed?

kathaypacific:
I think it looks good!

packages/mobile/src/transactions/feed/TransactionFeed.tsx
```diff
@@ -0,0 +1,128 @@
...
+    Logger.warn(
+      TAG,
+      `Found errors when querying the transaction feed: ${JSON.stringify(result.errors)}`
+    )
+  }
+
+  useEffect(() => {
+    if (result?.data?.tokenTransactionsV2.transactions.length) {
+      dispatch(updateTransactions(result.data.tokenTransactionsV2.transactions))
+    }
```
jeanregisser, line 71: nit: There's a [`onSuccess` callback in react-use-async](https://github.com/slorber/react-async-hook/blob/f8569186fbf77607b0f6d46a62639cd838b4d966/src/index.ts#L73) which would avoid adding this extra `useEffect`.

packages/mobile/src/transactions/feed/TransactionFeed.tsx
```diff
@@ -0,0 +1,128 @@
...
+  // Update the counter variable every |POLL_INTERVAL| so that a query is made to the backend.
+  const [counter, setCounter] = useState(0)
+  useEffect(() => {
+    const timer = setInterval(() => setCounter((n) => n + 1), POLL_INTERVAL)
+    return () => clearInterval(timer)
+  }, [])
+
+  // TODO: Extract this to a more generic function/hook so that it can be reused
+  const { loading, error, result } = useAsync(async () => {
+    const response = await fetch(`${config.blockchainApiUrl}/graphql`, {
```
jeanregisser, line 47: Preparing the move away from apollo client, right? 😄

jeanregisser (commented):
Great to see this!

I think this PR should have left out the changes around the invites.
These are 2 different things which should be discussed separately.

packages/mobile/src/transactions/feed/TransactionFeed.test.tsx
gnardini: I agree in general, snapshot files don't seem very useful. I'm not sure about starting in this PR though, maybe this is something we can talk over together in our next team meeting? So that we can all agree and move forward in agreement :)

packages/mobile/src/geth/networkConfig.ts
gnardini: Yeah, it's just commented out. It saves a little time when testing locally so that I only need to uncomment this and comment the other line instead of writing the whole URL from scratch. Not a huge deal of course, if you don't like it I can delete it, it's just a minor convenience thing.

packages/mobile/src/transactions/feed/query.ts
```diff
@@ -0,0 +1,88 @@
+export const TRANSACTIONS_QUERY = `
```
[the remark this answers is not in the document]
gnardini: I tried, but it doesn't work with it. I get this error:
`GraphQL queries must be strings. It looks like you're sending the internal graphql-js representation of a parsed query in your request instead of a request in the GraphQL query language. You can convert an AST to a string using the `print` function from `graphql`, or use a client like `apollo-client` which converts the internal representation to a string for you`

packages/mobile/src/transactions/feed/TransferFeedItem.test.tsx
```diff
@@ -0,0 +1,385 @@
...
+    verifyDisplay({
+      getByTestId,
+      expectedTitleSections: ['feedItemReceivedTitle', 'Simplex'],
+      expectedSubtitleSections: ['tokenDeposit', 'cUSD'],
+      expectedAmount: '+₱13.30',
+      expectedTokenAmount: '10.00 cUSD',
+    })
+  })
+
+  // TODO: Also test with encrypted comment.
```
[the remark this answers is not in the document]
gnardini: It was already super big, didn't want to make it even bigger.. I figured it would be easier to review, no?

packages/mobile/src/transactions/feed/TransactionFeed.tsx
```diff
@@ -0,0 +1,128 @@
...
+    switch (tx.__typename) {
+      case 'TokenExchangeV2':
+        // TODO
+        return <View key={tx.transactionHash} />
+      case 'TokenTransferV2':
+        // TODO: We should standarize this. The tokens in Firebase are usually
+        // (not always for some reason) in the checksum format (with some capital
+        // letters) but they come in all lowercase from blockchain-api.
+        if (
+          !tokensInfo[toChecksumAddress(tx.amount.tokenAddress)] &&
```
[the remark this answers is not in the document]
gnardini: Yeah, I was under the impression that we had already agreed to use lowercase letters, we use them almost everywhere, the only exception in the tokens info in Firebase. I think that's an oversight from when we implemented that but we should probably just fix it there and then remove this condition which should be unnecessary.

packages/mobile/src/transactions/feed/TransactionFeed.test.tsx
kathaypacific: yeah of course! leave it for now and i'll bring it up for the next meeting :)

packages/mobile/src/transactions/feed/TransactionFeed.tsx
gnardini: Oh, nice! I didn't know about this, much better solution :)

packages/mobile/src/components/TokenDisplay.tsx
```diff
@@ -91,7 +93,7 @@ function TokenDisplay({
 
   const amountToShow = showLocalAmount ? amountInLocalCurrency : new BigNumber(amount)
 
-  const sign = amountToShow.isNegative() ? '-' : showExplicitPositiveSign ? '+' : ''
+  const sign = hideSign ? '' : amountToShow.isNegative() ? '-' : showExplicitPositiveSign ? '+' : ''
```
dievazqu, line 96: Why do we need to hide the sign?

packages/mobile/src/geth/networkConfig.ts
kathaypacific: i don't really have strong feelings about this (my eye only twitched once!), just wanted to check :) if it is saving you time then i don't mind it being here

packages/mobile/src/transactions/types.ts
```diff
@@ -67,3 +68,66 @@ type TransferTransactionType =
   | TokenTransactionType.InviteSent
   | TokenTransactionType.InviteReceived
   | TokenTransactionType.NetworkFee
+
+export type TokenTransaction = TokenTransfer | TokenExchange
```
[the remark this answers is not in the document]
gnardini: They are not generated, they are created manually. I know generating them automatically is possible but it is not currently working. I don't see the point of using them, it adds a dependency on GraphQL for no reason if we're just going to update it manually.

packages/mobile/src/transactions/feed/TransactionFeed.tsx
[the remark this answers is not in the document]
dievazqu: Yep, the admin dashboard uploads the token address exactly as the user inputs it. But we should map it to lowercase. However, the new logic (the one that automatically uploads token info) already uses lowercase.

packages/mobile/src/transactions/types.ts
```diff
@@ -67,3 +68,66 @@ type TransferTransactionType =
...
+  fees: [Fee]
+}
+
+interface TokenExchangeMetadata {
+  title?: string
+  subtitle?: string
+}
+
+export interface Fee {
+  type: string
```
dievazqu, line 131: I think this could be a type too, the same as `TokenTransactionTypeV2` for example.

packages/mobile/src/components/TokenDisplay.tsx
gnardini: The Figma mockups say that. The amount in local currency has the + or - but the token amount doesn't have any sign: https://www.figma.com/file/8USELhisKuP1YlCt95WrD2/Support-for-Multiple-Tokens?node-id=951%3A29262

packages/mobile/src/transactions/feed/query.ts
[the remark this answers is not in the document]
jeanregisser: Ah yes I see, we're not using Apollo client with this one. Nevermind.

packages/mobile/src/transactions/feed/TransferFeedItem.tsx
```diff
@@ -0,0 +1,136 @@
...
+  subtitle: {
+    ...fontStyles.small,
+    color: colors.gray4,
+    paddingTop: 2,
+  },
+  amount: {
+    ...fontStyles.regular500,
+    marginLeft: 'auto',
+    paddingLeft: 10,
+    width: '40%',
```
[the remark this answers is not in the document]
jeanregisser: Yes feel free to skip in this PR then, if it's copied from the existing component.

packages/mobile/src/transactions/feed/TransferFeedItem.test.tsx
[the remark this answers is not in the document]
jeanregisser: Yep

packages/mobile/src/transactions/feed/TransactionFeed.tsx
jeanregisser: Yes the Apollo lib we're using is quite outdated, I feel you about this pain point.
I'm all for removing it from the code base too. 👍

packages/mobile/src/transactions/types.ts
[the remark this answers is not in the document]
jeanregisser: Yes, if we're not continuing to use GraphQL this looks good!
I'd love to avoid manually typing server responses though. Let's discuss it in our next team meeting.

packages/mobile/src/transactions/feed/TransactionFeed.tsx
```diff
@@ -0,0 +1,123 @@
...
+      },
+    }
+  )
+
+  if (result?.errors) {
+    Logger.warn(
+      TAG,
+      `Found errors when querying the transaction feed: ${JSON.stringify(result.errors)}`
+    )
+  }
```
jeanregisser, line 73: nit: Could be nice to put this one in the `useAsync` `onError` option.

jeanregisser (approved):
Amazing! 🎉

packages/mobile/src/transactions/feed/TransactionFeed.tsx
gnardini: The problem is that this isn't an error, GraphQL returns 200 for errors as well, but I can put it in `onSuccess`

jeanregisser: oh yep! 👍

PR #1578, opened 2021-12-06
Remove strict mode from mergify configuration

jh2oman:
### Description

strict mode is being deprecated and today (December 6th) there is a global brownout of the configuration, so none of our builds will pass today unless we fix this :) 

https://blog.mergify.com/strict-mode-deprecation/


jh2oman:
@Mergifyio refresh

lisaqyli (approved):

PR #1582, opened 2021-12-07
add app version to firebase registrations node

kathaypacific:
### Description

Add app version as a new property in the registrations node so that we can capture the app version in the Firebase RTDB and manage backwards compatibility with our BE services.

The immediate use case for this is push notifications for all supported tokens - as the push notifications allow a deep link into the app, we want to prevent incompatible notifications from being sent to the app. In the future this will be useful for feature support for various app versions.

This PR only adds the new db property, as this needs to be deployed before we can test writing to it.


kathaypacific:
this doesn't need to be deployed separately, closing as this change is included in 1583

Issue #1127, opened 2021-09-16
Support for all supported tokens in push notifications


aleksey-nov:
Eng: 
* Investigate if notification-service can call blockscount endpoint to get all token transfers
* If the above doesn't exist, call one endpoint for each token (we maintain a whitelist of tokens)

Acceptance criteria:
* User should receive push notification when they receive any token, not just cusd/celo/ceur
* Identify what the current bottleneck is and what the bottleneck will be with querying for N tokens
* When you open a push notification it takes you to the right transaction detail screen

* ensure existing clients don't crash

PR #1583, opened 2021-12-07
update app version in registrations node on app init

kathaypacific:
### Description

Update app version in the registrations firebase rtdb node on app init to more easily manage backwards compatibility with our BE services.

The immediate use case for this is push notifications for all supported tokens - as the push notifications allow a deep link into the app, we want to prevent incompatible notifications from being sent to the app. In the future this will be useful for feature support for various app versions.

### Other changes

N/A

### Tested

Manually and update unit test

### How others should test

Manually

### Related issues

Relates to https://github.com/valora-inc/wallet/issues/1127


packages/mobile/src/firebase/firebase.ts
```diff
@@ -382,15 +367,20 @@ export async function readOnceFromFirebase(path: string) {
     .then((snapshot) => snapshot.val())
 }
 
-export async function setUserLanguage(address: string, language: string | null) {
+export async function setRegistrationProperties(address: string, properties: Record<string, any>) {
```
jeanregisser, line 370: I think we should be explicit about what props we allow here to prevent mistakes:
```suggestion
export async function setRegistrationProperties(address: string, properties: { language?: string | null, fcmToken?: string, appVersion?: string}) {
```

jeanregisser (commented):
Great! 👍

packages/mobile/src/firebase/firebase.ts
kathaypacific: yeah good idea :D applied

jeanregisser (approved):

PR #1590, opened 2021-12-07
feat(persona): Implement new Persona KYC states

### LINKED ISSUE #1567: Support additional Persona KYC statuses in frontend
We recently updated our backend to support additional KYC statuses from Persona (namely, `KycStatus.Approved`, `KycStatus.Declined`, and `KycStatus.NeedsReview`). We should update the frontend Persona flow to support these statuses appropriately.

 AC:
* KYC flow updated to support new Persona KYC statuses

jophish:
### Description

We recently learned that we need to support additional KYC statuses from Persona in order to implement the KYC flow correctly; this PR adds those states in and adjusts when certain parts of the flow are displayed to the user.

### Related issues

Fixes #1567


packages/mobile/src/account/LinkBankAccountScreen.tsx
```diff
@@ -96,6 +98,10 @@ function StepOne({ kycStatus }: StepOneProps) {
           <Text style={styles.description}>{t('linkBankAccountScreen.pending.description')}</Text>
         </View>
       )
+    case KycStatus.NotCreated:
+    case KycStatus.Created:
+    case KycStatus.Pending:
+    case KycStatus.Expired:
```
jophish, line 104: We don't *need* these states here, but I thought it would be better to be explicit. Also, unsure if the fall-through should be considered an anti-pattern or not.

packages/mobile/src/account/LinkBankAccountScreen.test.tsx
```diff
@@ -53,19 +53,23 @@ describe('LinkBankAccountScreen', () => {
...
-    const tree = render(
-      <Provider store={store}>
-        <LinkBankAccountScreen />
-      </Provider>
-    )
-    fireEvent.press(tree.getByTestId('SupportContactLink'))
-    expect(navigate).toBeCalledWith(Screens.SupportContact, {
-      prefilledText: 'linkBankAccountScreen.failed.contactSupportPrefill',
+  describe('redirects correctly to SupportContact when button is clicked', async () => {
+    ;[KycStatus.Failed, KycStatus.Declined].forEach((kycValue) => {
```
cajubelt, line 57: I'd recommend using it.each for this

jh2oman (approved):
+1 After Charlie's nit

Issue #1588, opened 2021-12-07
Finish implementation of nightly builds


aleksey-nov:
Create a standalone nightly build app

- Add new build targets;
- Update release automation repo to use new config.

PR #1599, opened 2021-12-08
Add nightly apps

jeanregisser:
### Description

Add nightly apps for both mainnet and alfajores, for iOS and Android.

They use new bundle ids so they can be installed alongside the production apps.

Also created the apps on the App Store / Play Store.

Next: update the `release-automation` project to build these nightly apps.

### Tested

Built the nightly targets locally (iOS and Android)

### How others should test

N/A

### Related issues

- Part of #1588

### Backwards compatibility

Yes


kathaypacific (approved):
🚀

PR #1602, opened 2021-12-09
Misc E2E: Test ID's and update to fund-e2e-accounts.sh

MuckT:
### Description

Add a testID to the verification learn more model and assert based on testID presence or absence instead of displayed text which can change in Crowdin.

### Other changes

- Updated snapshots.
- E2E tests will now fail prior to running if balance of the e2e test wallet is below 10 for any currency prior to running the e2e tests. This should allow e2e test runs to fail prior to running Detox.

### Tested

E2E Tests and unit tests run locally.

### How others should test

### Related issues

N/A

### Backwards compatibility

Yes


jeanregisser (approved):
🙏

MuckT:
@Jean updated the `fund-e2e-accounts.sh` to error if the balance is low prior to running Detox tests.

PR #1613, opened 2021-12-13
refactor(in-house-liquidity): Put in-house-liquidity helpers in their own separate file

jh2oman:
### Description

Create separate IHL file to prepare for more IHL endpoints being called by the app. Also fixes several issues with the original call to IHL.

### Tested

`yarn test:all`

### Related issues

[- Fixes #[issue number here]](https://app.zenhub.com/workspaces/acquisition-squad-sprint-board-6010683afabec1001a090887/issues/valora-inc/wallet/1476)


jophish (approved):
LGTM, pending e2e tests passing!

packages/mobile/src/inHouseLiquidity/index.test.ts
```diff
@@ -0,0 +1,48 @@
+import { createPersonaAccount } from 'src/inHouseLiquidity'
```
cajubelt, line 1: let's use kebab-case for the IHL directory. So, `packages/mobile/src/in-house-liquidity` instead of `packages/mobile/src/inHouseLiquidity`

packages/mobile/src/inHouseLiquidity/index.ts
```diff
@@ -0,0 +1,42 @@
+import networkConfig from 'src/geth/networkConfig'
+import { serializeSignature, signMessage } from '@celo/utils/lib/signatureUtils'
+import { getStoredMnemonic } from 'src/backup/utils'
+import { generateKeys } from '@celo/utils/lib/account'
+
+export const createPersonaAccount = async (accountMTWAddress: string): Promise<Response> => {
+  const message = `post /account/create ${JSON.stringify({ accountMTWAddress })}`
```
cajubelt, line 7: the Date header needs to be in the signature too. see packages/mobile/src/inHouseLiquidity/index.test.ts

cajubelt (commented):
Nice, this is great for unit test-ability. See comments

packages/mobile/src/in-house-liquidity/index.ts
```diff
@@ -36,7 +40,13 @@ const getSerializedSignature = async (
   const signature = signMessage(message, privateKey, accountMTWAddress)
   return serializeSignature(signature)
 }
-const getAuthHeader = async (message: string, accountMTWAddress: string): Promise<string> => {
+const getAuthHeader = async (
+  endpoint: string,
+  accountMTWAddress: string,
+  body: string = ''
+): Promise<string> => {
+  const dateHeader = new Date().getUTCDate()
```
cajubelt, line 48: This won't work because IHL will expect the date header of the request to match the date that is used in the signed message. It's mean to help prevent replay attacks.

cajubelt (commented):
Thanks for making the updates. Looks like there's still a disconnect on that date header issue though.

packages/mobile/src/in-house-liquidity/index.ts
jh2oman: Ah, I understand how.

packages/mobile/src/in-house-liquidity/index.test.ts
```diff
@@ -47,7 +50,7 @@ describe('In House Liquidity Calls', () => {
       )
       // Has the correct Signature
       expect(signMessage).toHaveBeenCalledWith(
-        `post /account/create 14 ${expectedBody}`,
+        `post persona/account/create ${date.toUTCString()} ${expectedBody}`,
```
cajubelt, line 53:
```suggestion
        `post /persona/account/create ${date.toUTCString()} ${expectedBody}`,
```
I think the path needs a leading slash

packages/mobile/src/in-house-liquidity/index.ts
```diff
@@ -5,16 +5,16 @@ import { generateKeys } from '@celo/utils/lib/account'
 
 export const createPersonaAccount = async (accountMTWAddress: string): Promise<Response> => {
   const body = { accountMTWAddress }
-  const authorization = await getAuthHeader(
-    'post /account/create',
+  const headers = await getAuthAndDateHeaders(
+    'post persona/account/create',
```
cajubelt, line 9:
```suggestion
    'post /persona/account/create',
```
same here

packages/mobile/src/in-house-liquidity/index.ts
```diff
@@ -40,13 +40,18 @@ const getSerializedSignature = async (
   const signature = signMessage(message, privateKey, accountMTWAddress)
   return serializeSignature(signature)
 }
-const getAuthHeader = async (
+
+const getAuthAndDateHeaders = async (
```
cajubelt, line 44: sorry to just bring this up now, but this could probly use a docstring explaining how to use it

cajubelt: the fact that you need an HTTP verb prefix, in particular, is not intuitive

packages/mobile/src/in-house-liquidity/index.ts
```diff
@@ -40,13 +40,18 @@ const getSerializedSignature = async (
...
   endpoint: string,
   accountMTWAddress: string,
   body: string = ''
-): Promise<string> => {
-  const dateHeader = new Date().getUTCDate()
-  const message = `${endpoint} ${dateHeader} ${body}`
+): Promise<{ Date: string; Authorization: string }> => {
+  const date = new Date().toUTCString()
+  const message = `${endpoint} ${date} ${body}`
```
cajubelt, line 50: this actually won't work for the GET case because it will have a trailing space that isn't expected

cajubelt (commented):
I noticed a few more things 😬

packages/mobile/src/in-house-liquidity/index.ts
jh2oman: let me refactor it a bit and actually write some unit tests for it too

packages/mobile/src/in-house-liquidity/index.ts
```diff
@@ -5,51 +5,69 @@ import { generateKeys } from '@celo/utils/lib/account'
...
     },
     body: JSON.stringify(body),
   })
 }
 
 // Helper Functions
 
-const getPrivateKey = async (accountMTWAddress: string): Promise<string> => {
+const getKeys = async (accountMTWAddress: string) => {
   const mnemonic = await getStoredMnemonic(accountMTWAddress)
```
cajubelt, line 20: I think the input for getStoredMnemonic is supposed to be an EOA, not an MTW address

cajubelt: in `BackupPhrase.tsx` the same function is used, but an EOA address is supplied (rather than an MTW)

jh2oman: I tried it just now with both. Using MTW address works

Using EOA address returns a different privateKey 😮 which IHL then says is not a signature that matches the call.

jeanregisser: We shouldn't use the mnemonic this way.

The EOA or DEK private keys are already stored securely and should be accessed via the `wallet` interface (see [`getWallet`](https://github.com/valora-inc/wallet/blob/8a2718ba8b7b74540d93c98cf8929f83364e144c/packages/mobile/src/web3/contracts.ts#L170)
Its `signTypedData` function should probably be preferred.

Not entirely familiar with IHL, but the DEK may be more appropriate to use.

Anyway looks like we need to discuss to share more context.

lisaqyli: Following, would like to learn more about this when you guys discuss

packages/mobile/src/in-house-liquidity/index.test.ts
```diff
@@ -0,0 +1,149 @@
...
+      const headers = await getAuthAndDateHeaders(
+        'GET',
+        '/account/foo',
+        MOCK_USER.accountMTWAddress
+      )
+
+      // Creates the correct Headers
+      expect(headers).toMatchObject({
+        Date: expectedDateString,
+        Authorization: `Valora ${MOCK_USER.walletAddress}:0x1c8689cf5de841119091ef4f0c49170c4130909b402dfba1af7bcab4f047a5f23735d0fe8ab631804cd51a098bf335cd2a8243d94630435e6d854fc96998f8a313`,
```
lisaqyli, line 54: nit: extract the expected serialized signature into its own variable for readability

lisaqyli:
overall. LGTM, thanks for fixing the IHL signatures! Would love to learn about Mnemonic that Jean is talking about

packages/mobile/src/in-house-liquidity/index.ts
```diff
@@ -0,0 +1,90 @@
...
+  options: RequestInit
+): Promise<Response> => {
+  const authAndDateHeaders = await getAuthAndDateHeaders(
+    options.method,
+    path,
+    accountMTWAddress,
+    walletAddress,
+    options.body as string
+  )
+  return fetch(`${networkConfig.inhouseLiquditiyUrl}${path}`, {
```
cajubelt, line 52: `inhouseLiquditiyUrl` isn't properly camel cased

cajubelt: it also has a typo in the "liquidity" part

packages/mobile/src/in-house-liquidity/index.ts
```diff
@@ -0,0 +1,90 @@
...
+ * Gets the auth and date headers that IHL expects as a signature on mosts requests
+ *
+ * The behavior is slightly different between GET requests and other types of requests
+ *
+ * @param {httpVerb} string GET, POST
+ * @param {requestPath} string like /persona/get/foo
+ * @param {accountMTWAddress} accountAddress
+ * @param {walletAddress} walletAddress
+ * @param {requestBody} string optional request body
+ * @returns {{Date, Authorization}} date and authorization headers
```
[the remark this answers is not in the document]
cajubelt: then you wouldn't need the fancy mocking

cajubelt (commented):
Nice!! Just a typo and some nits, I think.

packages/mobile/src/in-house-liquidity/index.ts
```diff
@@ -0,0 +1,90 @@
...
+ */
+export const getAuthAndDateHeaders = async (
+  httpVerb: string = '',
+  requestPath: string,
+  accountMTWAddress: string,
+  walletAddress: string,
+  requestBody?: BodyInit | null
```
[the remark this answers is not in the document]
jh2oman: Actually do need the null since RequestInit.body can be BodyInit | null

lisaqyli (approved):

cajubelt (approved):
🥳 🥳 🥳 🥳 🥳

ValoraQA:

Issue #1461, opened 2021-11-16
Initial performance instrumentation


silasbw:
ENG:

* You're encouraged to use Sentry: https://docs.sentry.io/platforms/react-native/performance/
* You're also encouraged to look at alternatives to understand the landscape better. E.g., https://instabug.com/product/app-performance-monitoring
* Note that the [automatic instrumentation](https://docs.sentry.io/platforms/react-native/performance/instrumentation/automatic-instrumentation/) from Sentry helps with the AC below

AC:
A performance report / dashboard summarizing RUM for Valora. Including but not limited to:
* TTI (Time To Interactive) aka warm start and cold start in Sentry docs
* UI Hangs per screen (A hang is when the app isn't responding to the user's input for more than 250 ms.)
* Network data transfers per session (KB up and down) -- **only do this if it's provided by automatic instrumentation**
* HTTP fetch calls and latencies
* navigation latencies
* Memory / CPU / Disk usage
* Anything else that Sentry automatic instrumentation might provide

* Ensure general app performance is not visibly impacted by the added instrumentation (there's a "sampling" parameter you can set with Sentry).  If possible (i.e., less than ~1hr of work) make this tunable by RemoteConfig

PR #1620, opened 2021-12-14
Update Sentry to latest version

kathaypacific:
### Description

This PR updates Sentry to the latest version. According to the migration docs, there were no changes needed from our side (there were very few breaking changes and we are not using the changed features). 

https://docs.sentry.io/platforms/react-native/migration/

### Tested

Rebuilt both iOS and Android apps, manually (created and) triggered an error and verified that it came through in Sentry.

### How others should test

There should be no change to the app behaviour with this PR, so verify that Sentry is working as expected. Errors that you trigger should be received by Sentry as usual.

### Related issues

Prepares for #1461

### DIFF
--- a/packages/mobile/ios/Podfile.lock
+++ b/packages/mobile/ios/Podfile.lock
@@ -628,9 +628,9 @@ PODS:
     - React
   - RNSecureRandom (1.0.0):
     - React
-  - RNSentry (2.1.0):
+  - RNSentry (3.2.8):
     - React-Core
-    - Sentry (= 6.0.9)
+    - Sentry (= 7.5.4)
   - RNShare (3.3.0):
     - React
   - RNSVG (12.1.1):
@@ -655,9 +655,9 @@ PODS:
     - Firebase (~> 7.7)
     - Firebase/Core (~> 7.7)
     - FirebaseAnalytics (~> 7.7)
-  - Sentry (6.0.9):
-    - Sentry/Core (= 6.0.9)
-  - Sentry/Core (6.0.9)
+  - Sentry (7.5.4):
+    - Sentry/Core (= 7.5.4)
+  - Sentry/Core (7.5.4)
   - Toast (4.0.0)
   - TOCropViewController (2.6.0)
   - Yoga (1.14.0)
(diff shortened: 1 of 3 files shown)


packages/mobile/ios/Podfile.lock
```diff
@@ -1101,19 +1101,19 @@ SPEC CHECKSUMS:
...
-  Sentry: 388c9dc093b2fd3a264466a5c5b21e25959610a9
+  Sentry: 5c5dd4005f3b7b9765d5a8871232cddbd0d888b7
   Toast: 91b396c56ee72a5790816f40d3a94dd357abc196
   TOCropViewController: 3105367e808b7d3d886a74ff59bf4804e7d3ab38
   Yoga: 4bd86afe9883422a7c4028c00e34790f560923d6
   YogaKit: f782866e155069a2cca2517aafea43200b01fd5a
 
 PODFILE CHECKSUM: d88bd91187a580924250d0d0c8281a6e164b4a68
 
-COCOAPODS: 1.10.1
```
jeanregisser, line 1119: We should be using our "pinned" version of CocoaPods from our Gemfile.lock.
Looks like `pod ...` was run instead of `bundle exec pod ...`

It's fine to update CocoaPods, but would be preferable in a different PR.

jeanregisser (commented):
Great! 👍 
Almost good, just a small thing about the CocoaPods version.

packages/mobile/ios/Podfile.lock
kathaypacific: good catch! that's exactly what happened indeed (had some issues installing pods on my machine, i did end up doing a `pod install` instead of `bundle exec pod install` - i got lost in the weeds) 

for this i uninstalled cocoapods + reinstalled to the version we have in the podfile.lock + ran `pod update Sentry`. ran `bundle exec pod install` afterwards to check no more changes were added.

jeanregisser (approved):

ValoraQA:

kathaypacific:
hey @ValoraQA no testing is needed from your side for this one :)

PR #1621, opened 2021-12-14
feat(analytics): Add in client-side events for M1 PFP

### LINKED ISSUE #1464: [M1] tracking events implementation
AC:
- tracking events created for each of the touch points we want to track (see spike issue)
- able to analyze m1 tracking events from dashboard (Lisa says this is automatic, should confirm)

jophish:
### Description

This PR adds support for tracking events related to the M1 PFP integration. After speaking with Gowtham, I learned that relying on `screen_view` events is not well supported, since they can apparently be pretty flaky. I've updated the [Notion doc](https://www.notion.so/valora-inc/M1-Tracking-Events-3f0e92b500994acd8c7dbe7093144763) with a new design for emitting client-side (and server-side) events.

We will emit the following events:

* `SettingsEvent.settings_link_bank_account`: user presses the "Link Bank Account" button with a phone # connected.
* `SettingsEvent.settings_number_not_connected`: user presses the "Link Bank Account" button _without_ a phone # connected.
* `CICOEvents.link_bank_account_cancel`: user presses the back button after entering the "Link Bank Account" screen with a phone # connected.
* `CICOEvents.persona_kyc_start`: user presses the "Begin" button for Persona KYC
* `CICOEvents.connect_phone_start`: user presses the "Connect" button after trying to link a bank account with no phone # connected.
* `CICOEvents.connect_phone_cancel`: user presses the back button after trying to link a bank account with no phone # connected.
* `CICOEvents.persona_kyc_success`: triggered when the `onSuccess` hook of the Persona inquiry is called in the client.
* `CICOEvents.persona_kyc_cancel`: triggered when the `onCancel` hook of the Persona inquiry is called in the client.
* `CICOEvents.persona_kyc_error`: triggered when the `onError` hook of the Persona inquiry is called in the client.

### Related issues

Fixes #1464.


packages/mobile/src/navigator/Headers.tsx
```diff
@@ -103,6 +104,20 @@ export const headerWithBackButton: StackNavigationOptions = {
   headerLeft: ({ canGoBack }) => (canGoBack ? <BackButton /> : null),
 }
 
+export function headerWithBackButtonEvent({
```
lisaqyli, line 107: wouldn't it make more sense to just put the `eventName` prop in `headerWithBackButton` since it's optional anyways?

packages/mobile/src/navigator/Headers.tsx
```diff
@@ -126,6 +141,26 @@ export const headerWithCloseButton: StackNavigationOptions = {
   headerLeftContainerStyle: { paddingLeft: 20 },
 }
 
+export function headerWithCloseButtonEvent({
```
lisaqyli, line 144: Same comment as the previous one

lisaqyli:
Also I think it would be a good idea to add tracking events for KYC success / fail / cancel here https://github.com/valora-inc/wallet/blob/main/packages/mobile/src/account/Persona.tsx#L62

packages/mobile/src/navigator/Headers.tsx
jophish: Yeah, we could change the `headerWithBackButton` (and `headerWithCloseButton`) to take an optional `eventName` param; I considered that, but wasn't sure if we should modify it since it's currently being used in a bunch of different places and would require changes to a bunch of files to accommodate that change.

jophish:
@lisaqyli Good idea about the events for the Persona callbacks; I've added those events to the Notion doc and updated the PR to include them.

cajubelt (commented):
is it worth adding assertions in our tests too? to verify that the ValoraAnalytics.track method was called for the different button clicks etc. This might go into the realm of "testing implementation, rather than behavior", so if you'd rather not I understand

packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -442,12 +443,12 @@ const settingsScreens = (Navigator: typeof Stack) => (
     <Navigator.Screen
       name={Screens.LinkBankAccountScreen}
       component={LinkBankAccountScreen}
-      options={headerWithBackButton}
+      options={headerWithBackButtonEvent({ eventName: CICOEvents.link_bank_account_cancel })}
```
jeanregisser, line 446: Between iOS and Android, there are 3 ways to leave a screen:
- hitting the back button in the navigator bar
- using the system back button (Android only)
- swipe to navigate back (iOS only)

This event will only catch the back button in the navigation bar.
One way to cover all cases is by using the technique described in https://reactnavigation.org/docs/preventing-going-back/
But we'd need to be sure it doesn't trigger when not intended. See when `beforeRemove` is actually called.

jeanregisser (commented):
Looking almost good!
See edge cases for the cancel event.

packages/mobile/src/analytics/Events.tsx
```diff
@@ -37,6 +37,8 @@ export enum SettingsEvents {
   settings_profile_edit = 'settings_profile_edit',
   settings_profile_name_edit = 'settings_profile_name_edit',
   settings_verify_number = 'settings_verify_number',
+  settings_link_bank_account = 'settings_link_bank_account',
+  settings_number_not_connected = 'settings_number_not_connected',
```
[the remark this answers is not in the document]
jophish: Oops, good catch -- will add. Also in regard to your other comment; I think it's fine to add those in; part of the "behavior" is that we should log an event.

packages/mobile/src/navigator/Navigator.tsx
jophish: @jeanregisser Interesting -- thanks for the info... I'm not sure I understand why it would trigger unintentionally; from the docs, it looks like the only case where we wouldn't want to fire an event is when `Some action such as pop or reset was dispatched which removes the screen from the state`. Is something like this possible? It seems like we would just need to add a listener for `beforeRemove` events and log an event there..

packages/mobile/src/analytics/Properties.tsx
```diff
@@ -1114,6 +1117,16 @@ interface DappKitProperties {
   }
 }
 
+interface CICOEventsProperties {
+  [CICOEvents.link_bank_account_cancel]: undefined
```
jh2oman, line 1121: Why are these all `undefined`?

jophish: @jh2oman I was following the pattern used by other events which don't require additional properties to be sent; I'm not totally sure if this is object is needed, but felt it would be good to be consistent with our other events. I'm happy to remove this if it's totally unnecessary.

jh2oman: Ah, don't require additional properties. I think its fine to continue the pattern. Its nice and explicit then.

jh2oman (approved):

packages/mobile/src/account/ConnectPhoneNumberScreen.tsx
```diff
@@ -17,6 +21,14 @@ function onPressConnectButton() {
 export default function ConnectPhoneNumberScreen() {
   const { t } = useTranslation()
 
+  // Log a cancel event on a "back" action (hardware back button, swipe, or normal navigate back)
+  const navigation = useNavigation()
+  React.useEffect(() => {
+    navigation.addListener('beforeRemove', (e) => {
+      ValoraAnalytics.track(CICOEvents.link_bank_account_cancel)
+    })
+  })
```
jeanregisser, line 30: We should always pass the dependencies array (here empty is fine as `navigation` doesn't change), otherwise it will apply the effect on each re-render.
Also it's necessary to remove the listener when unmounting. This is done by returning the result from `addListener`. See https://reactnavigation.org/docs/navigation-events/#navigationaddlistener
```suggestion
  React.useEffect(
    () =>
      navigation.addListener('beforeRemove', (e) => {
        ValoraAnalytics.track(CICOEvents.link_bank_account_cancel)
      }), 
    []
  )
```

packages/mobile/src/account/LinkBankAccountScreen.tsx
```diff
@@ -17,11 +17,23 @@ import LoadingSpinner from 'src/icons/LoadingSpinner'
...
 }
 
 function LinkBankAccountScreen() {
+  // Log a cancel event on a "back" action (hardware back button, swipe, or normal navigate back)
+  const navigation = useNavigation()
+  React.useEffect(() => {
+    navigation.addListener('beforeRemove', (e) => {
+      ValoraAnalytics.track(CICOEvents.link_bank_account_cancel)
+    })
+  })
```
jeanregisser, line 35: Same as above.

jeanregisser (commented):
Almost good!

packages/mobile/src/navigator/Navigator.tsx
jeanregisser: Yeah if there's no `pop` or `reset` involved in this flow it should be good.

lisaqyli (approved):

PR #1627, opened 2021-12-14
Fail E2E CI check if E2E wallet address balance is too low prior to running tests

MuckT:
### Description

Prevent e2e tests from running when the balance is too low.

### Other changes

N/A

### Tested

Tested in CI - [Run with fail on low balance](https://github.com/valora-inc/wallet/runs/4544841078?check_suite_focus=true).

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/.github/workflows/e2e-android.yml
+++ b/.github/workflows/e2e-android.yml
@@ -53,10 +53,9 @@ jobs:
           echo TWILIO_ACCOUNT_SID=$TWILIO_ACCOUNT_SID >> .env
           echo VERIFICATION_PHONE_NUMBER=$VERIFICATION_PHONE_NUMBER >> .env
       - name: Fund E2E accounts
-        continue-on-error: false
         run: |
           cd packages/mobile/e2e/scripts
-          node ./fund-e2e-accounts.ts
+          node --unhandled-rejections=strict ./fund-e2e-accounts.ts
       - name: Run E2E tests
         run: |
           export ANDROID_SDK_ROOT=$HOME/android-tools
--- a/.github/workflows/e2e-android.yml
+++ b/.github/workflows/e2e-android.yml
@@ -53,6 +53,7 @@ jobs:
           echo TWILIO_ACCOUNT_SID=$TWILIO_ACCOUNT_SID >> .env
           echo VERIFICATION_PHONE_NUMBER=$VERIFICATION_PHONE_NUMBER >> .env
       - name: Fund E2E accounts
+        continue-on-error: false
         run: |
           cd packages/mobile/e2e/scripts
           node ./fund-e2e-accounts.ts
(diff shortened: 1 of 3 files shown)


packages/mobile/e2e/scripts/fund-e2e-accounts.ts
```diff
@@ -14,11 +14,16 @@ const kit = ContractKit.newKitFromWeb3(web3)
...
 const balanceError = async (address = valoraE2ETestWallet, minBalance = 10) => {
-  let balanceObject = await getBalance(address)
-  for (const balance in balanceObject) {
-    if (balanceObject[balance] < minBalance) {
-      throw new Error(`Balance of ${address} is below ${minBalance}`)
+  try {
+    let balanceObject = await getBalance(address)
+    for (const balance in balanceObject) {
+      if (balanceObject[balance] < minBalance) {
+        throw new Error(`Balance of ${address} is below ${minBalance}`)
```
kathaypacific, line 21: nice! i was wondering if you think it'd be a good idea to also add some text about how to fund the accounts to resolve the error? i am imagining myself seeing this error and wanting to resolve it to keep the pipeline going. it'd be super helpful to see something like `Balance of ${address} is below ${minBalance}, follow these instructions (on the readme? notion?) to top up the balance`?

.github/workflows/e2e-android.yml
```diff
@@ -53,6 +53,7 @@ jobs:
           echo TWILIO_ACCOUNT_SID=$TWILIO_ACCOUNT_SID >> .env
           echo VERIFICATION_PHONE_NUMBER=$VERIFICATION_PHONE_NUMBER >> .env
       - name: Fund E2E accounts
+        continue-on-error: false
```
jeanregisser, line 56: AFAIK this is already the default and can be removed.

.github/workflows/e2e-ios.yml
```diff
@@ -59,6 +59,7 @@ jobs:
           echo TWILIO_ACCOUNT_SID=$TWILIO_ACCOUNT_SID >> .env
           echo VERIFICATION_PHONE_NUMBER=$VERIFICATION_PHONE_NUMBER >> .env
       - name: Fund E2E accounts
+        continue-on-error: false
```
jeanregisser, line 62: ditto

packages/mobile/e2e/scripts/fund-e2e-accounts.ts
jeanregisser: +1

packages/mobile/e2e/scripts/fund-e2e-accounts.ts
```diff
@@ -14,11 +14,16 @@ const kit = ContractKit.newKitFromWeb3(web3)
...
+      }
     }
+  } catch (err) {
+    console.log(err)
+    process.exit(1)
   }
```
jeanregisser, line 27: Though this works for this specific function, there's a better alternative which would exit with a non-zero status code on any unhandled promise rejection:

```js
process.on("unhandledRejection", (error) => {
  throw error;
});
```

or alternatively call the script with:
`node --unhandled-rejections=throw some-script.js`

`--unhandled-rejections=throw` is now the default since Node 15, see https://nodejs.org/api/cli.html#--unhandled-rejectionsmode

jeanregisser (approved):

PR #1632, opened 2021-12-15
Add basic performance monitoring with Sentry

kathaypacific:
### Description

Add tracing for react native, to get out of the box performance monitoring from Sentry.

In this PR
- initialise Sentry with ReactNativeTracing
- use sampling rate from remote config - remote config variables are stored in redux, which creates a dependency on redux. Once Sentry is initialised, we cannot change the sampling rate but at the same time we should be initialising Sentry as soon as possible. The compromise: initialise Sentry in appInit saga, which is run after redux rehydrate. This means that the first time the app is loaded, the default sampling rate is used. On subsequent app loads, the previously fetched remote config value is used.

Performance metrics measured 
- TTI: app warm and cold startup times
- navigation latencies
- http fetch latencies
- UI hangs per screen

What doesn't come out of the box:
- Memory/CPU/disk usage of device

### Other changes

N/A

### Tested

Verify in Sentry that performance events are displayed
Verify manually that app runs as normal (without major downgrade in performance) - it was hard to check this without performance monitoring 🙈 what I did do:
- to check the app load time was unaffected: in debug mode check the log timestamps between app load and splash screen dismiss for app with/without tracing enabled. I saw no change.
- to check the "feel" of the app: make a release build on device (I only have an iPhone, so couldn't check on Android) and compare with Valora app. I didn't notice any change to the app performance, and sampling rate or 0.2 vs 1 made no noticeable difference.

### How others should test

Same as above

### Related issues

- Fixes #[1461]

### Backwards compatibility

Yes


packages/mobile/src/sentry/Sentry.ts
```diff
@@ -1,32 +1,66 @@
...
 
-// This should be called as early in the lifecycle of the app as possible.
-function installSentry() {
   if (!SENTRY_CLIENT_URL) {
     Logger.info(TAG, 'installSentry', 'Sentry URL not found, skiping instalation')
     return
   }
+
+  const tracesSampleRate = yield select(sentryTracesSampleRateSelector)
+  const tracingOrigins = [
```
silasbw, line 34: for the services we control, this makes sense.  by why add the tracing header to the 3rd party services?

https://docs.sentry.io/platforms/react-native/performance/instrumentation/automatic-instrumentation/#tracingorigins

silasbw (commented):
:mage:

packages/mobile/src/sentry/Sentry.ts
kathaypacific: we don't have control over the performance of 3rd party services but if they are causing our app performance to drop, it is nice to know this upfront so that we don't waste time investigating the cause? i thought the extra visibility was nice, but i don't have strong feelings about this so i can remove if you think we should not track

kathaypacific: (but yes i forgot to mark this PR as WIP before - the list of tracingOrigins is something I still need to refine)

packages/mobile/src/config.ts
```diff
@@ -163,3 +163,5 @@ export const OTA_TRANSLATIONS_FILEPATH = `file://${CachesDirectoryPath}/translat
 export const FETCH_TIMEOUT_DURATION = 15000 // 15 seconds
 
 export const DEFAULT_APP_LANGUAGE = 'en-US'
+
+export const DEFAULT_SENTRY_TRACES_SAMPLE_RATE = 0.2
```
kathaypacific, line 167: The docs recommended choosing a lower number if unsure of what to use. 0 = no events are sent, 1 = 100% events are sent. happy to take suggestions, i thought 0.2 might be a nice starting point.

packages/mobile/src/sentry/Sentry.ts
kathaypacific: 🤔  i tried to go through the urls that we have in config to see what they do/if they should be included, but i'm not sure if i've understood everything correctly. i've excluded third party urls that are being used to generate webviews that we don't want to measure, but maybe there are still things in there we don't want to include as part of our traces?

packages/mobile/src/sentry/Sentry.ts
```diff
@@ -1,32 +1,57 @@
...
+    DEFAULT_FORNO_URL,
+    networkConfig.blockchainApiUrl,
+    networkConfig.odisUrl,
+    networkConfig.komenciUrl,
+    networkConfig.providerFetchUrl,
+    networkConfig.simplexApiUrl,
+    networkConfig.fetchUserLocationDataUrl,
+    networkConfig.komenciLoadCheckEndpoint,
+    networkConfig.walletConnectEndpoint,
+    networkConfig.inhouseLiquditiyUrl,
```
jeanregisser, line 39: Some of these are cloud functions sharing the same domain, can we track the base domain instead?

jeanregisser (approved):
Great! 🚀

packages/mobile/src/sentry/Sentry.ts
silasbw: My expectation for this feature is to provide inter-service traces that help pinpoint failures and bottlenecks across all our services.  e.g., you can follow a request from the wallet -> blockchain API -> some other service.  that feature requires all services in the trace to be using sentry and our account.  So it makes sense to configure that information for services we own (and might use sentry today, or will sometime in the future).  But it's unclear why we'd do if for services we don't own (and won't ever run sentry with our account).

kathaypacific: yeah you're right, i guess i'm not 100% sure about which services you consider we "own"? the thing i was trying to achieve with these tracingOrigins is to provide more detail to the out of the box navigation tracing that Sentry provides - without these tracing urls, we would only see the "navigation" trace without any other information. i went through again in more detail and removed any urls that do not directly influence the screen navigation events. of course there are lots of other things we could be tracking to give this navigation trace more detail (like react profiling and custom instrumentation) but i've stuck to the minimum for now...the react profiling is very intrusive, we need to export every component we want to track wrapped in `Sentry.withProfiler` 🙈

Screenshot 2021-12-20 at 12 10 35

jeanregisser: I'm not too sure what's the recommended way to do this with Sentry, but even if we don't own a service (ODIS, segment, firebase, etc), we'd still want to know how long it takes to request from them.
Reading https://docs.sentry.io/product/sentry-basics/tracing/distributed-tracing/ it seems adding them as origins would be the way to go.
What do you think @kathaypacific @silasbw?

kathaypacific: This is also my original thought, tracing all services we use so that we can see their impact on our own trace events.

I ended up removing most of them because I couldn't verify that the traces were coming through correctly - either they did not contribute to the traces currently being traced (i.e. the ui load and navigation) or the services are not allowing the sentry trace header to be added to the requests. I inspected the network requests going through flipper on screen navigation and checked the corresponding events in Sentry.

To make the trace work, the docs say `You will need to configure your web server CORS to allow the sentry-trace header` so that got me doubtful about leaving url's in the `tracingOrigins` when they might not be working. I thought it would be better to add to the `tracingOrigins` if we expand our use of Sentry performance monitoring?

silasbw: We're talking about two things:
* distributed tracing
* profiling

We should do profiling, which includes transactions times for services we don't own.

We should also do distributed tracing, but it's only going to work for services we own (and that also use sentry!).

silasbw: ...i guess i have no idea how sentry tracing works or if tracing is coupled with distributed tracing (but if that's the case it's implied that traces can't include non-HTTP transactions, which doesn't seem very useful) :confused: 

@gnardini halp

silasbw (approved):
A, but it would be good to solidify our understanding of Sentry's tracing and distributed tracing.

packages/mobile/src/sentry/Sentry.ts
gnardini: I'm confused by how all this works.

I agree in principle of the difference between profiling vs tracing. 

Sentry has two concepts for performance monitoring: A `Transaction` is the main thing which can have several `Span`s. HTTP requests should be spans within a transaction.

Sentry's React Native docs claim that they add all http requests as spans to the active transaction. However, they don't. When seeing the transaction details I see that HTTP requests show up as 'Breadcrumbs' which are a different thing not performance-related so they are detecting them but not adding them as spans. However, when we add the tracing origins thing they do show up as Spans. I tried googling around a bit but it doesn't look like there's a ton of information around this.

I'd say we can keep these things as tracing originas for now?

packages/mobile/src/sentry/Sentry.ts
gnardini: I think we should also add Komenci's URL? If we have it..

packages/mobile/src/sentry/Sentry.ts
```diff
@@ -1,32 +1,51 @@
...
+  }
 
-// This should be called as early in the lifecycle of the app as possible.
-function installSentry() {
   if (!SENTRY_CLIENT_URL) {
     Logger.info(TAG, 'installSentry', 'Sentry URL not found, skiping instalation')
     return
   }
+
+  const tracesSampleRate = yield select(sentryTracesSampleRateSelector)
```
gnardini, line 28: We can also add custom transactions in places we care about performance. Using `const transaction = Sentry.startTransaction({ name: "actionName" });` and `transaction.finish()`.

I think it might make sense to add some of these in some places we care about? For example the send/invite sagas? Maybe it's something for a future PR 🤔

packages/mobile/src/sentry/Sentry.ts
kathaypacific: i had this but removed because i couldn't verify that the traces for it are coming through...in this PR we are only tracking the navigation traces and if i understand correctly, the http calls to komenci do not impact the navigation? 

do you think we should include all endpoints in this tracingOrigins array? not sure if you saw my comment[https://github.com/valora-inc/wallet/pull/1632#discussion_r772528463] but that is how i ended up with this current list, let me know if you think we should go back to including everything?

packages/mobile/src/sentry/Sentry.ts
kathaypacific: i agree! for this PR i just focused on getting the tracing integrated. in the future it'd be nice to have custom transactions but i think we should put some thought into exactly where we want them :)

packages/mobile/src/sentry/Sentry.ts
gnardini: I think this is fine for now, we can look more into it when/if we add specific performance monitoring to the verification flow

gnardini (approved):

MuckT:
@kathaypacific I am not sure why it wasn't caught by the main being merge into CI. But it looks like this pull request might cause detox to hang on Android. When running detox I am getting the messages below and detox hangs and doesn't run the tests.
```
detox[32673] INFO:  [APP_STATUS] The app is busy with the following tasks:
• There are enqueued timers.
detox[32673] INFO:  [APP_STATUS] The app is busy with the following tasks:
• There are enqueued timers.
```

PR #1658, opened 2021-12-20
Standby transaction legacy

dievazqu:
### Description

* Move current stand-by transaction logic to a "legacy" code.
* Display new stand-by transactions in the feed.
* Add a stand-by transaction in sending transactions.

### Tested

Manually in a emulator + unit tests

### How others should test

Anyone should be able to see the new stand-by transaction after doing a transaction. Make sure you have the ff on for the new flow.

### Related issues

- Fixes some bullet points on #1573

### Backwards compatibility

Compatible


packages/mobile/src/transactions/saga.ts
```diff
@@ -48,15 +52,15 @@ const RECENT_TX_RECIPIENT_CACHE_LIMIT = 10
 
 // Remove standby txs from redux state when the real ones show up in the feed
 function* cleanupStandbyTransactions({ transactions }: NewTransactionsInFeedAction) {
```
gnardini, line 54: I think we also need to clean up the non-legacy transactions when the actual tx is found in the feed, no?

packages/mobile/src/transactions/feed/TransactionFeed.tsx
```diff
@@ -74,12 +80,58 @@ function useQueryTransactionFeed() {
...
 
   const { loading, error, transactions } = useQueryTransactionFeed()
-  const tokenTransactions: TokenTransaction[] = transactions ?? cachedTransactions
+  const confirmedTokenTransactions: TokenTransaction[] = transactions ?? cachedTransactions
+
+  const standbyTransactions = useSelector(standbyTransactionsSelector).map((tx) =>
+    mapStandbyTransactionToTokenTransaction(tx)
+  )
+
+  const tokenTransactions = [...confirmedTokenTransactions, ...standbyTransactions]
```
gnardini, line 134: Shouldn't these by in the reverse order? ie. First the standby transactions and then the confirmed?

Also, when testing this the standby transaction doesn't show up as 'Confirming' what we used to have is an item that looked the same except that the subtitle said `Confirming...` (the `confirmingTransaction` translation text)

packages/mobile/src/send/saga.ts
```diff
@@ -267,19 +272,18 @@ function* sendPayment(
...
-    //     address: recipientAddress,
-    //   })
-    // )
+    yield put(
+      addStandbyTransaction({
+        context,
+        type: TokenTransactionTypeV2.Sent,
+        comment,
+        status: TransactionStatus.Pending,
+        value: amount.toString(),
```
gnardini, line 281: I _think_ we need to add a `-` here, since this is an outgoing transaction, the value change is negative. I think this is an issue with the legacy stuff as well. Don't worry about the legacy (not sure if it's even a problem), but for the new feed we need it negative I think.

gnardini (commented):
Looks really good overall, thanks for woking on this!

packages/mobile/src/escrow/saga.ts
```diff
@@ -138,7 +138,7 @@ export function* transferToEscrow(action: EscrowTransferPaymentAction) {
 
     // Tranfser the funds to the Escrow contract.
     Logger.debug(TAG + '@transferToEscrow', 'Transfering to escrow')
-    yield call(registerStandbyTransaction, context, amount.toString(), escrowWrapper.address)
+    yield call(registerStandbyTransactionLegacy, context, amount.toString(), escrowWrapper.address)
```
gnardini, line 141: I think the non-legacy code also uses this. Maybe we can register both legacy and non-legacy here? So invites also show up in the standby state.

packages/mobile/src/transactions/saga.ts
dievazqu: This is done in the next PR: https://github.com/valora-inc/wallet/pull/1669

packages/mobile/src/transactions/feed/TransactionFeed.tsx
dievazqu: Hmm, let me check this again

packages/mobile/src/send/saga.ts
dievazqu: Make sense, I will change it

packages/mobile/src/escrow/saga.ts
dievazqu: This is done in the next PR: https://github.com/valora-inc/wallet/pull/1669

packages/mobile/src/transactions/feed/TransactionFeed.test.tsx
```diff
@@ -123,4 +141,21 @@ describe('TransactionFeed', () => {
...
+        standbyTransactions: MOCK_STANDBY_TRANSACTIONS,
+      },
+    })
+
+    await waitFor(() => tree.getByTestId('TransactionList'))
+
+    expect(tree.queryByTestId('NoActivity/loading')).toBeNull()
+    expect(tree.queryByTestId('NoActivity/error')).toBeNull()
+
+    expect(tree).toMatchSnapshot()
```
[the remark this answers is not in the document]
dievazqu: Okay... I will try with the testID :)

packages/mobile/src/transactions/feed/TransferFeedItem.tsx
```diff
@@ -12,28 +13,36 @@ import TokenDisplay from 'src/components/TokenDisplay'
...
   const openTransferDetails = () => {
     navigate(Screens.TransactionDetailsScreen, { transaction: transfer })
     ValoraAnalytics.track(HomeEvents.transaction_feed_item_select)
   }
 
-  const { title, subtitle, recipient } = useTransferFeedDetails(transfer)
+  let { title, subtitle, recipient } = useTransferFeedDetails(transfer)
+  const isPending = transfer.status === TransactionStatus.Pending
+
+  // I feel this should be inside useTransferFeedDetails() instead of here. What do you think?
```
gnardini, line 42: Yes, I think that probably makes sense :)

packages/mobile/src/transactions/feed/TransferFeedItem.tsx
```diff
@@ -12,28 +13,36 @@ import TokenDisplay from 'src/components/TokenDisplay'
...
+  subtitle = isPending ? t('confirmingTransaction') : subtitle
   const colorStyle = new BigNumber(amount.value).isPositive() ? { color: colors.greenUI } : {}
 
   return (
-    <Touchable disabled={false} onPress={openTransferDetails}>
+    <Touchable disabled={isPending} onPress={openTransferDetails}>
```
gnardini: I think we should still be able to open the details for pending transactions?

packages/mobile/src/transactions/feed/TransferFeedItem.tsx
```diff
@@ -35,14 +37,14 @@ function TransferFeedItem({ transfer }: Props) {
   }
 
   let { title, subtitle, recipient } = useTransferFeedDetails(transfer)
-  const isPending = transfer.status === TransactionStatus.Pending
+  const isStandbyTransaction = transfer.status === TransactionStatus.Pending
 
   // I feel this should be inside useTransferFeedDetails() instead of here. What do you think?
-  subtitle = isPending ? t('confirmingTransaction') : subtitle
+  subtitle = isStandbyTransaction ? t(STAND_BY_TRANSACTION_SUBTITLE_KEY) : subtitle
```
gnardini: I think doing it this way will prevent us from being able to do some static replace thing that I don't remember well. @jeanregisser mentioned it a few times, it's preferred if we don't pass variables to the `t` function, just strings. I think the constant is only used in the test? In that case I think we can just hard-code it there

gnardini (commented):
Looks good 🙌  !

gnardini (approved):
Great job! 👏

ValoraQA:
Hey @MuckT, **Need clarification for standby transaction**,We are assuming that when we buy CELO than at first Confirming activity appears, before getting completed. Can we call it as stand by transaction ?
If not can you please let us know what is stand by transaction.
thanks..!!

PR #1661, opened 2021-12-21
Some Persona Cleanups 

lisaqyli:
### Description

This is a PR to clean up Persona, here are some notable changes:
1. should pass user's MTW address to firebase
2. UIs for different Persona Inquiry states
3. on inquiry Cancelled flow, kycStatus will be updated to `Pending`, updated Pending UI to let user to start over.

### Tested
Manually tested


packages/mobile/src/account/LinkBankAccountScreen.tsx
```diff
@@ -98,10 +128,33 @@ function StepOne({ kycStatus }: StepOneProps) {
           <Text style={styles.description}>{t('linkBankAccountScreen.pending.description')}</Text>
         </View>
       )
-    case KycStatus.NotCreated:
-    case KycStatus.Created:
```
cajubelt, line 102: So what will happen now for the notcreated and created cases? Maybe I'm missing something.

cajubelt (approved):
Thanks for cleaning things up! 🚢 

Left a question, probably fine though so I'm also approving.

In the future, please try for single-purpose PR's.

packages/mobile/src/account/LinkBankAccountScreen.tsx
jh2oman: The not created and created pieces get the 'default' from the case/switch, which is the "Verify your Identity" view

packages/mobile/src/account/LinkBankAccountScreen.tsx
```diff
@@ -98,10 +128,33 @@ function StepOne({ kycStatus }: StepOneProps) {
...
+    case KycStatus.Completed:
```
jh2oman, line 131: I thought that "Completed" meant that the user completed the KYC but it was not yet approved, which was why we previously showed the "Verification Pending" screen. Why are we now showing the "Verification Complete" screen?

lisaqyli: you can see the description of various persona lifecycle states here: https://docs.withpersona.com/docs/models-lifecycle
And also i tested this, if the inquiry was successful without needing to be manually reviewed, `completed` will be the updated state

ValoraQA:

PR #1669, opened 2021-12-21
Adding StandBy Transactions with multitoken support

### LINKED ISSUE #1573: Add Standby Transactions to multi-token support Transaction Feed

dievazqu:
### Description

It adds the new stand-by transaction when there is an invite sent, and exchange or a celo withdraw.
It removes the stand-by transaction that are already fetched from blockchain-api.

### Tested

Unit tests (I gonna test this locally soon)

### Related issues

- Fixes #1573

### Backwards compatibility

Compatible


packages/mobile/src/escrow/saga.ts
```diff
@@ -139,6 +140,13 @@ export function* transferToEscrow(action: EscrowTransferPaymentAction) {
     // Tranfser the funds to the Escrow contract.
     Logger.debug(TAG + '@transferToEscrow', 'Transfering to escrow')
     yield call(registerStandbyTransactionLegacy, context, amount.toString(), escrowWrapper.address)
+    yield call(
+      registerStandbyTransaction,
+      context,
+      amount.toString(),
```
gnardini, line 146: I haven't checked but I think we might need to negate this as well?

packages/mobile/src/tokens/saga.ts
```diff
@@ -248,7 +245,7 @@ export function tokenTransferFactory({ actionName, tag }: TokenTransferFactory)
         )
       } catch (error) {
         Logger.error(tag, 'Error transfering token', error)
-        yield put(removeStandbyTransactionLegacy(context.id))
+        yield put(removeStandbyTransaction(context.id))
```
gnardini, line 248: Hmm, we are creating a legacy pending tx but removing the non-legacy one?

I think for now we need to both and and remove both legacy and non-legacy so that it works weather we show the new or old feed.

Alternatively, we could use one or the other depending on the status of the feature flag for the new transaction feed

packages/mobile/src/transactions/actions.ts
```diff
@@ -24,10 +24,8 @@ export enum Actions {
   REFRESH_RECENT_TX_RECIPIENTS = 'TRANSACTIONS/REFRESH_RECENT_TX_RECIPIENTS',
   UPDATE_RECENT_TX_RECIPIENT_CACHE = 'TRANSACTIONS/UPDATE_RECENT_TX_RECIPIENT_CACHE',
   UPDATE_TRANSACTIONS = 'TRANSACTIONS/UPDATE_TRANSACTIONS',
-  // Remove legacy actions once the multitoken support feature is fully released
+  // Remove legacy action once the multitoken support feature is fully released
   ADD_STANDBY_TRANSACTION_LEGACY = 'TRANSACTIONS/ADD_STANDBY_TRANSACTION_LEGACY',
-  REMOVE_STANDBY_TRANSACTION_LEGACY = 'TRANSACTIONS/REMOVE_STANDBY_TRANSACTION_LEGACY',
```
gnardini, line 29: This flag is not on yet, right? So we shouldn't remove this yet?

packages/mobile/src/tokens/saga.ts
dievazqu: Sorry, maybe it's a bit confusing but `removeStandbyTransaction` removes from the legacy and the non-legacy: https://github.com/valora-inc/wallet/pull/1669/files#diff-5ba66e05e1de1827234d9fbc6273e0f406fc733151024de936747ce2e77b252fR80
At first I did separately, but then I realized it was the same code duplicated and we could do it like this.
Let me know if you think isn't clear enough

packages/mobile/src/transactions/actions.ts
dievazqu: Same as before, I merged both "remove logic" in one.

packages/mobile/src/escrow/saga.ts
dievazqu: Yep, it makes sense :)

packages/mobile/src/tokens/saga.ts
gnardini: Ohh, yeah that makes a lot of sense, nevermind.

gnardini (approved):
Amazing! 🙌

ValoraQA:
Hey @MuckT, **Need clarification for standby transaction**,We are assuming that when we buy CELO than at first Confirming activity appears, before getting completed. Can we call it as stand by transaction ?
If not can you please let us know what is stand by transaction.
thanks..!!

PR #1670, opened 2021-12-21
Add workflow to automerge Crowdin PRs daily

jeanregisser:
### Description

One less thing to do manually.

### How others should test

N/A

### Related issues

- #1588

### Backwards compatibility

Yes


kathaypacific (approved):
This looks good to me, really love the idea!! would be keen to merge this and test it out, add refinements if needed later as Jean is on vacation. @gnardini you reckon?

gnardini (approved):
Amazing! I agree @kathaypacific lets merge it and see how it does :)

My only slight concern is that if the PR has conflicts we might be merging them and that could break the build? Although if they do one of the checks might fail I think, so merging would fail. Something to monitor.

PR #1676, opened 2021-12-22
Covert Backup Complete to Functional Component

MuckT:
### Description

This PR changes the `BackupComplete.tsx` to a functional component from a class based component.

### Other changes

- Added assert that the `BackComplete.tsx` screen is served during the onboarding e2e tests.
- Updated snapshots with new testID.

### Tested

Tested locally and with e2e tests.

### How others should test

Complete the new account onboarding an ensure that the backup complete screen with "You're all set" screen is served.

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/packages/mobile/src/backup/BackupComplete.tsx
+++ b/packages/mobile/src/backup/BackupComplete.tsx
@@ -7,6 +7,7 @@ import { StyleSheet, Text, View } from 'react-native'
 import { SafeAreaView } from 'react-native-safe-area-context'
 import { OnboardingEvents } from 'src/analytics/Events'
 import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
+import { backupCompletedSelector } from 'src/backup/selectors'
 import { navigate } from 'src/navigator/NavigationService'
 import { Screens } from 'src/navigator/Screens'
 import { StackParamList } from 'src/navigator/types'
@@ -22,11 +23,11 @@ type Props = StackScreenProps<StackParamList, Screens.BackupComplete>
 
 function BackupComplete({ route }: Props) {
   const navigatedFromSettings = route.params?.navigatedFromSettings ?? false
-  const backupCompleted = useSelector((state) => state.account.backupCompleted)
+  const backupCompleted = useSelector(backupCompletedSelector)
   const { t } = useTranslation()
 
   useEffect(() => {
-    setTimeout(() => {
+    const timer = setTimeout(() => {
       if (navigatedFromSettings) {
         navigate(Screens.Settings, { promptConfirmRemovalModal: true })
       } else if (backupCompleted) {
@@ -36,6 +37,7 @@ function BackupComplete({ route }: Props) {
         throw new Error('Backup complete screen should not be reachable without completing backup')
       }
     }, 2000)
+    return () => clearTimeout(timer)
   }, [])
 
   return (
--- a/packages/mobile/src/backup/BackupComplete.tsx
+++ b/packages/mobile/src/backup/BackupComplete.tsx
@@ -2,65 +2,48 @@ import Checkmark from '@celo/react-components/icons/Checkmark'
 import fontStyles from '@celo/react-components/styles/fonts'
 import { StackScreenProps } from '@react-navigation/stack'
 import * as React from 'react'
-import { WithTranslation } from 'react-i18next'
+import { useTranslation } from 'react-i18next'
 import { StyleSheet, Text, View } from 'react-native'
 import { SafeAreaView } from 'react-native-safe-area-context'
-import { connect } from 'react-redux'
 import { OnboardingEvents } from 'src/analytics/Events'
 import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
-import { withTranslation } from 'src/i18n'
 import { navigate } from 'src/navigator/NavigationService'
 import { Screens } from 'src/navigator/Screens'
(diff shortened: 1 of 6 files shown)


packages/mobile/src/backup/BackupComplete.tsx
```diff
@@ -2,65 +2,48 @@ import Checkmark from '@celo/react-components/icons/Checkmark'
...
-    }, 2000)
-  }
+type Props = StackScreenProps<StackParamList, Screens.BackupComplete>
+
+function BackupComplete({ route }: Props) {
+  const navigatedFromSettings = route.params?.navigatedFromSettings ?? false
+  const backupCompleted = useSelector((state) => state.account.backupCompleted)
+  const { t } = useTranslation()
+
+  setTimeout(() => {
```
gnardini, line 28: This is setting the timeout on every render, but we only want it to happen on mount looks like. To achieve that we need a `useEffect` hook, like so:
```
useEffect(() => {
  setTimeout(...)
}, [])
```
notice the empty array as a second parameter, that array is a list of dependencies, whenever one of those changes the function in the first parameter is called (it's also called the first time). Since we're passing an empty array nothing can change so the whole thing is only ran once.

gnardini (commented):
Thanks for working on this! left just a comment :)

packages/mobile/src/backup/BackupComplete.tsx
```diff
@@ -48,19 +36,16 @@ class BackupComplete extends React.Component<Props> {
         throw new Error('Backup complete screen should not be reachable without completing backup')
       }
     }, 2000)
-  }
+  }, [])
```
kathaypacific, line 39: nit: i think we should also return a `clearTimer` function here to cleanup the `useEffect` after component unmount? this was missing in the previous implementation but for whatever reason if this component is unmounted before 2 seconds then this setTimeout will be hanging out :)

packages/mobile/src/backup/BackupComplete.tsx
```diff
@@ -1,44 +1,32 @@
...
-  static navigationOptions = { header: null }
 
-  componentDidMount() {
-    // Show success check for a while before leaving screen
-    const { backupCompleted } = this.props
+type Props = StackScreenProps<StackParamList, Screens.BackupComplete>
+
+function BackupComplete({ route }: Props) {
+  const navigatedFromSettings = route.params?.navigatedFromSettings ?? false
+  const backupCompleted = useSelector((state) => state.account.backupCompleted)
```
kathaypacific, line 25: nit: should we create a selector here, to keep consistency? i'm always unsure about the benefit of using selectors vs doing the selection directly like this, i think for simple selectors like this there isn't much difference but there are already some simple selectors in the `src/account/selectors.ts` file so would be nice to keep the pattern?

kathaypacific (commented):
😎

packages/mobile/src/backup/BackupComplete.tsx
MuckT: Should be fixed in https://github.com/valora-inc/wallet/pull/1676/commits/ddc035bd8314453a01fde0f89275674f373f3a86!

packages/mobile/src/backup/BackupComplete.tsx
MuckT: Should be fixed in https://github.com/valora-inc/wallet/pull/1676/commits/ddc035bd8314453a01fde0f89275674f373f3a86!

kathaypacific (approved):
Looks good to me!

ValoraQA:
Hey @MuckT, We have verified this task using Android Nightly Build V1.24.0 & updated results in below sheet.

**Sheet Link:** https://docs.google.com/spreadsheets/d/1LcYXmQcuLTBIHZrzUjoYaUvnzisq2ReRvxVwmn-xGEA/edit#gid=1640061259

**New Query:** [Link](https://github.com/valora-inc/wallet/issues/1699)

**Devices:** Google Pixel XL (10.0), Redmi Note 8 (11.0)
Thanks..!!

ValoraQA:
Hey @MuckT, We have verified this task using Android Internal Build V1.24.1, Test flight Build V1.24.1 & updated results in below sheet.

**Sheet Link:** https://docs.google.com/spreadsheets/d/15nGo3iJAEVJ76wHae9PvYuWI7S2bovKAQPX_edtwPEI/edit#gid=573249288

**Devices:** iPhone 13 mini (15.1.1), iPhone XS Max (13.5.1), iPhone 12 (14.7.1), Google pixel XL (10.0), Google pixel 2xl (11.0)
Thanks..!!

ValoraQA:
Hey @MuckT, We have verified this task using Android Internal release Build V1.24.1 & updated results in below sheet.

**Sheet Link**: https://docs.google.com/spreadsheets/d/15nGo3iJAEVJ76wHae9PvYuWI7S2bovKAQPX_edtwPEI/edit#gid=573249288

**Devices**: Samsung Galaxy A8+(8.0)
Thanks..!!

ValoraQA:
Hey @MuckT, We have verified this task using Android Internal Release Build V1.24.1 & updated results in below sheet.

**Sheet Link:**https://docs.google.com/spreadsheets/d/15nGo3iJAEVJ76wHae9PvYuWI7S2bovKAQPX_edtwPEI/edit#gid=573249288

**Devices:** Redmi Note 5 Pro (9.0)
Thanks..!!

Issue #1677, opened 2021-12-22
 Application crashes when user navigates to "Phone number" page while creating & restoring account


ValoraQA:
**Frequency:** 100%

**Repro on build version:** Testflight Build V1.23.2
**Repro on devices:** iPhone 13 mini (15.1.1), iPhone 12 (14.7.1)

**Does Not Repro on build version:** Testflight Build V1.23.1, Android Internal Build V1.23.2
**Does Not Repro on devices:** iPhone 13 mini (15.1.1), iPhone 12 (14.7.1), Samsung Galaxy A8+(8.0), One plus 7T(11.0)

**Pre-condition:**
1. User must have installed the application 
2. User must be on Create Account Step 1 of 3 screen & added photo & name into respective fields.

**Repro Steps:**
1. Tap on "Next" button after adding photo and name.
2. Enter a valid PIN on Create PIN step 2of 3 screen.
3. Enter same valid PIIN again to confirm PIN & observe.

**Current Behavior:** Application crashes when user navigates to phone number page while creating & restoring account

**Expected Behavior:** Application must not crash when user navigates to phone number page while creating & restoring account & able to complete both flows. 

**Investigation:** This issue is not reproducible on Android Platform

**Impact:** User is blocked to create or restore account.

**Attachment:** 
[App crashes.mp4](https://drive.google.com/file/d/1I1iNT7kyPEXve5GX4jAXEXfrXIzD-C8N/view?usp=sharing) 
[Celo Crash Logs](https://drive.google.com/file/d/1BFkDBdrbL0y12ewjt4diHUD_zSLGFHwg/view?usp=sharing)

ValoraQA:
Hey @MuckT, We have verified this issue on Test flight build V1.23.2 and observed that application is crashing when user opens the valora application.

**Device:** iPhone Xs Max(13.5.1) 
**Attachment:** [App crash.mp4](https://drive.google.com/file/d/11fmi5aytjlfS4OK5b96CX6qo8Vl5RD-d/view?usp=sharing)
Thanks..!!

ValoraQA:
Hey @MuckT, We have verified this issue using latest Test flight Build V1.23.3 & observed that issue is fixed now. The app is not crashing on app load, also application is not crashing on "Phone number" page and user is able to complete the create & restore account flow.

**Devices:** iPhone Xs Max(13.5.1), iPhone 13 mini (15.1.1), iPhone 12 (14.7.1)
**Attachment:** 
[...]

PR #1679, opened 2021-12-22
Make sure totalBalanceUsd is not NaN

gnardini:
### Description

We had a bad token in Firebase which caused the total balance to return `NaN`. This should avoid it from crashing the app.

### Other changes

N/A

### Tested

Locally

### How others should test

Open the app, make sure it doesn't crash.

### Related issues

- Fixes #1671

### Backwards compatibility

N/A


kathaypacific (approved):

ValoraQA:
Hey @MuckT we have verified the issue on the latest iOS testflight build V:1.23.3 and observed the issue is not reproducible and also we have updated the comment on the ticket #1677 

Thanks..!

PR #1688, opened 2021-12-27
Setup Test Butler for Android E2E Stability

MuckT:
### Description

Adds [Test Butler](https://github.com/linkedin/test-butler) to our Android E2E tests. Occasionally we get Android fails in e2e tests such [as this](https://github.com/valora-inc/wallet/actions/runs/1628052396) where the Android emulator serves an error and all subsequent tests fail. By adding Test Butler we should not encounter these types of failures in the future. This was largely based on [this example](https://github.com/wix/Detox/tree/master/examples/demo-react-native-jest/e2e).

### Other changes

- Updated `packages/mobile/e2e/.gitignore`
- Added Test Butler as an androidTestImplementation similar to Detox.
- Added `e2e/global-setup.js` to handle Test butler setup
- Added `e2e/global-teardown.js` to more gracefully handle detox cleanup.
- Detox update to [19.4.1](https://github.com/wix/Detox/releases).

### Tested

Tested locally and in CI.

### How others should test

This does not need to be tested by QA. All runs after the last commit should be passing in CI.

[Android CI runs using this workflow](https://github.com/valora-inc/wallet/actions/workflows/e2e-android.yml?query=branch%3Atomm%2Fe2e-ci-211227)

### Backwards compatibility

Yes


packages/mobile/e2e/global-setup.js
```diff
@@ -0,0 +1,42 @@
+const fs = require('fs-extra')
+const { execSync } = require('child_process')
+const detox = require('detox')
+
+async function globalSetup() {
+  const config = resolveSelectedConfiguration() || {}
+  downloadTestButlerAPKIfNeeded(config)
+  await detox.globalInit()
```
kathaypacific, line 8: do we need to handle an error if this promise fails to resolve for whatever reason? same with the `globalCleanup`?

packages/mobile/e2e/global-setup.js
```diff
@@ -0,0 +1,42 @@
...
+function downloadTestButlerAPKIfNeeded(config) {
+  if (isAndroidConfig(config)) {
+    downloadTestButlerAPK()
+  }
+}
+
+function downloadTestButlerAPK() {
+  const version = '2.2.1'
+  const artifactUrl = `https://repo1.maven.org/maven2/com/linkedin/testbutler/test-butler-app/${version}/test-butler-app-${version}.apk`
+  const filePath = `./e2e/test-butler-app.apk`
```
kathaypacific, line 20: these feel like they should be in a config file somewhere but i remember you saying that we will make a global change for it at some point. maybe these should live at the top of the file, as constants?

kathaypacific (approved):
Nice! Looks good to me :D

packages/mobile/e2e/global-setup.js
MuckT: [Test Butler](https://github.com/linkedin/test-butler/releases) doesn't release often. This code was largely from [this example](https://github.com/wix/Detox/tree/master/examples/demo-react-native-jest/e2e).

packages/mobile/e2e/global-setup.js
MuckT: I was able to run it 10 times over the course of today and didn't see any failures. If it fails I believe the e2e test suite should fail.

PR #1697, opened 2021-12-29
Upgrade React Native 0.66.4

### LINKED ISSUE #1145: Upgrade React Native version
ENG:
- @jeanregisser might have some tips for doing the upgrade easily

AC:
- [ ] wallet uses version 0.66.X+ of React Native
- [ ] CI passes reliably
- [ ] Run some ad-hoc smoke tests on an iOS device and an Android device

kathaypacific:
### Description

Upgrade react native to 0.66.4.

Notable changes in this PR include:
- Android: upgrade to gradle version 4.2.2
- Android: use mavenCentral as [JCenter has been sunset](https://github.com/facebook/react-native/releases/tag/v0.65.0) - [exoplayer is not available on mavenCentral yet](https://github.com/DoubleSymmetry/react-native-track-player/issues/1235) so make an exception for that in build.gradle
- Patch react-native-fast-crypto which is [not compatible with gradle 4 yet](https://github.com/EdgeApp/react-native-fast-crypto/issues/36)
- Update Hermes to compatible version
- Update react native/react/react-test-renderer which required many test snapshots to be updated
- Remove references to deprecated `AppState.removeListener` (https://reactnative.dev/docs/appstate#removeeventlistener)
- Verified with activation team that Plaid button is still working correctly on Android. There was a runtime error on app launch which RN suggested to remove the Plaid package from Android Manifest, no official mention of this in the react-native-plaid-link-sdk project but following the RN suggestion worked (even though the react-native-plaid-link-sdk docs still mention [these lines in the setup steps](https://github.com/plaid/react-native-plaid-link-sdk#2-add-plaidpackage-to-your-application))

### Other changes

N/A

### Tested

E2E tests should work as usual, app should build and function as usual.

### How others should test

A basic regression - click around the app and see no broken functionalities

### Related issues

- Fixes #1145 

### Backwards compatibility

N/A


packages/mobile/android/app/src/main/res/values/styles.xml
```diff
@@ -2,7 +2,7 @@
   <style name="AppTheme" parent="Base.V0.AppTheme"/>
 
   <!--  Base theme, also see API specific overrides in values-vXX/styles.xml -->
-  <style name="Base.V0.AppTheme" parent="Theme.AppCompat.Light.NoActionBar">
+  <style name="Base.V0.AppTheme" parent="Theme.AppCompat.DayNight.NoActionBar">
```
jeanregisser, line 5: Why do we need to change this?

packages/mobile/android/app/src/main/res/values/styles.xml
```diff
@@ -11,7 +11,6 @@
     <item name="android:windowDrawsSystemBarBackgrounds">true</item>
     <!-- Without this we briefly see the theme preview before our own splash screen -->
     <item name="android:windowDisablePreview">true</item>
-    <item name="android:textColor">#000000</item>
```
jeanregisser, line 14: Why removing?

patches/react-native-fast-crypto+2.0.0.patch
```diff
@@ -0,0 +1,20 @@
...
+-
+-    // If you want Gradle to package prebuilt native libraries
+-    // with your APK, modify the default source set configuration
+-    // to include the directory of your prebuilt .so files as follows.
+-    sourceSets {
+-        main {
+-            jniLibs.srcDirs 'jni/libs/'
+-        }
+-    }
+ }
```
jeanregisser, line 18: What's the context around this patch?

jeanregisser (commented):
OMG, amazing!! 🚀 🤩 🎉

patches/react-native-fast-crypto+2.0.0.patch
kathaypacific: i had mentioned in the description (patch react-native-fast-crypto which is not compatible with gradle 4 yet) but noticed i only linked the issue - this is the real reason https://developer.android.com/studio/releases/gradle-plugin#cmake-imported-targets - basically this block is conflicting with gradle version 4 and should be removed

packages/mobile/android/app/src/main/res/values/styles.xml
kathaypacific: i don't have a better answer than i tried to copy the changes here https://react-native-community.github.io/upgrade-helper/?from=0.63.4&to=0.66.4 - my naive thinking is that newer versions of android has light/dark themes but i didn't look into it much. do you prefer to keep it unchanged?

packages/mobile/android/app/src/main/res/values/styles.xml
kathaypacific: same as above :)

packages/mobile/android/app/src/main/res/values/styles.xml
jeanregisser: Sounds good. Maybe we should double check how the app is impacted when the dark theme is selected (Settings -> Display -> Theme).

packages/mobile/android/app/src/main/res/values/styles.xml
jeanregisser: ok 👍

patches/react-native-fast-crypto+2.0.0.patch
jeanregisser: My bad missed this 👍

jeanregisser (approved):

jh2oman (approved):
Was able to build after switching my terminal to Rosetta mode.

packages/mobile/android/app/src/main/res/values/styles.xml
kathaypacific: I quickly checked this, the app looks the same as usual in dark mode (except for the native things like keyboard and alerts which are dark) 🙈

PR #1698, opened 2021-12-29
Setting status complete when transaction is successful

### LINKED ISSUE #1573: Add Standby Transactions to multi-token support Transaction Feed

dievazqu:
### Description

Setting the status of the stand by as complete if we already know the transaction was successful.

### Tested

Manually with the local emulator

### How others should test

### Related issues

- Fixes #1573

### Backwards compatibility

Compatible


packages/mobile/src/transactions/feed/TransactionFeed.tsx
```diff
@@ -144,6 +144,8 @@ function TransactionFeed() {
     mapStandbyTransactionToFeedTokenTransaction(tx)
   )
 
+  Logger.debug(TAG, `Standby transactions: ${JSON.stringify(standbyFeedTransactions)}`)
```
gnardini, line 147: De you think it's valuable to keep this log or was it for debugging mainly and we can delete it? I ask because the logs are already quite noisy and worry a bit about making them more so, especially because this will be printed very often I think

gnardini (approved):
This looks great, I'm just wondering now, does the comment show up whe nthe tx is comfirmed by this mechanism? I think we're not sending the comment to the standby tx thing, but I might be wrong.

Anways, amazing job! 👏

packages/mobile/src/transactions/feed/TransactionFeed.tsx
dievazqu: I think we can remove it

dievazqu:
> This looks great, I'm just wondering now, does the comment show up whe nthe tx is comfirmed by this mechanism? I think we're not sending the comment to the standby tx thing, but I might be wrong.
> 
> Anways, amazing job! 👏

Yep, I have never added the comment to the standby tx. However, I think this was working like this in the previous version. Right?
I'm testing with the previous flow, and I can't see the comment anywhere. :/

ValoraQA:
Hey @MuckT, We will verify all the transaction status, while performing regression test plan.
Thanks...!!

PR #1703, opened 2021-12-31
Disable sentry in e2e tests

MuckT:
### Description

This PR disables sentry in e2e test which can cause Android Detox e2e tests to hang.

### Other changes

- Prevent app reuse on ci runs.

### Tested

Tested locally and in CI.

### How others should test

### Related issues

N/A

### Backwards compatibility

Yes


packages/mobile/src/sentry/Sentry.ts
```diff
@@ -33,16 +33,19 @@ export function* initializeSentry() {
...
     dsn: SENTRY_CLIENT_URL,
     environment: DeviceInfo.getBundleId(),
     enableAutoSessionTracking: true,
-    integrations: [
-      new Sentry.ReactNativeTracing({
-        routingInstrumentation: sentryRoutingInstrumentation,
-        tracingOrigins,
-      }),
-    ],
+    integrations: isE2EEnv
```
silasbw, line 41: IMO it would be more straightforward (and understandable to other contributors) to have a "sentryEnabled" flag that's `false` by default and explicitly enabled when we actually want to configure logging to sentry.  That approach makes it safe for anyone to run the repo in any mode w/o worrying about logging to sentry; and the variable rename communicates the semantics (enable/disable sentry) without requiring people to dive into code.

silasbw (changes requested):

packages/mobile/src/sentry/Sentry.ts
MuckT: @silasbw would we want to have this enabled by default and disabled in e2e tests. I could see an issue where the remoteConfigValues are set to false, and only updated after the remote config fetch leading to missing metrics of app load in Sentry.

kathaypacific: there is already a `SENTRY_ENABLED` variable that we use to stop Sentry from being initialised completely for dev builds, should we just include the `isE2EEnv` variable there too? what we have here is Sentry minus the performance tracing for e2e tests - have we been using Sentry to evaluate errors generated by E2E tests? to me it's more clear to exclude Sentry completely....

MuckT: Opted to disable sentry entirely during e2e tests. This should reduce noise from sentry to user facing issues.

packages/mobile/scripts/run_e2e.sh
```diff
@@ -112,7 +116,7 @@ runTest() {
     --retries $RETRIES \
     --headless \
     "${test_match}" \
-    --reuse
+    "${extra_param}"
```
kathaypacific, line 119: just checking you still want to make this change, if we are opting to disable Sentry for e2e?

packages/mobile/src/sentry/Sentry.ts
```diff
@@ -11,7 +11,8 @@ const TAG = 'sentry/Sentry'
 
 // Set this to true, if you want to test Sentry on dev builds
 // Set tracesSampleRate: 1 to capture all events for testing performance metrics in Sentry
-export const SENTRY_ENABLED = !__DEV__ || false
+// Set SENTRY_ENABLED to false during e2e tests
+export const SENTRY_ENABLED = (!isE2EEnv && !__DEV__) || false
```
kathaypacific, line 15: sorry i know we talked about this but now that i stare at it....not sure why we needed the `|| false` before (`true || false`/`false || false` seems redundant 😕  )

kathaypacific (approved):
looks good to me, maybe nice to get a second pass from @silasbw to see if his concern was addressed by this :)

MuckT:
@silasbw do these changes look good to you?

packages/mobile/scripts/run_e2e.sh
MuckT: We still want this change. We never want CI to use a previous build; this was why the build was passing but then failing later after merging into main.

packages/mobile/src/sentry/Sentry.ts
MuckT: My reasoning here is that e2e tests run in release mode and we want to enable sentry on release modes that are not e2e runs.

silasbw: I'm :-1: on implicit behavior like this.  I.e., isE2EEnv implies that sentry is disabled.  It's not obvious to others what the entire implied behavior of isE2EEnv is.  I'd rather have a config value of SENTRY_ENABLED (i.e., in src/config.ts or whatever) that we explicitly set.

MuckT: Currently e2e tests run in release mode on Alfajores, but we could in the future run e2e tests on mainnet. Since we cannot base SENTRY_ENABLED solely on the network or release v.s. dev mode. It might be best to explicitly set it in the run_e2e.sh

silasbw: +1 to explicitly setting it and avoiding intermixing configuration with code via magic config variables like `isE2EEnv`.

silasbw (approved):
:artist: !

PR #1706, opened 2022-01-03
Track "taps" on the hamburger

silasbw:


gnardini (approved):
🚀

cajubelt (approved):

packages/mobile/src/navigator/DrawerTopBar.tsx
```diff
@@ -29,14 +31,15 @@ function DrawerTopBar({ middleElement, rightElement, scrollPosition, testID }: P
...
+
   return (
     <Animated.View testID={testID} style={viewStyle}>
-      <TouchableOpacity
-        style={styles.hamburger}
-        // @ts-ignore Only used in a drawer
-        onPress={navigation.toggleDrawer}
-        hitSlop={iconHitslop}
-      >
+      <TouchableOpacity style={styles.hamburger} onPress={hamburgerOnPress} hitSlop={iconHitslop}>
```
silasbw, line 42: changing `onPress` from `navigation.toggleDrawer` to `hamburgerOnPress` (or even `() => { navigation.toggleDrawer() }`) causes the snapshot tests to fail with this diff:

```
          <View
            accessible={true}
    -       focusable={false}
    +       focusable={true}
            hitSlop={
              Object {
                "bottom": 10,
                "left": 10,
                "right": 10,
```

That's unexpected and can't figure out why it's happening.  @jeanregisser or @kathaypacific any ideas?

packages/mobile/src/navigator/DrawerTopBar.tsx
```diff
@@ -29,14 +31,15 @@ function DrawerTopBar({ middleElement, rightElement, scrollPosition, testID }: P
     [scrollPosition]
   )
 
+  const hamburgerOnPress = () => {
```
jeanregisser, line 34: nit: we use this convention for naming event handlers.
```suggestion
  const onPressHamburger = () => {
```

jeanregisser (approved):
Thanks!

ValoraQA:

PR #1723, opened 2022-01-06
Allow seeing and sending tokens without USD price

### LINKED ISSUE #1722: Allow seeing and sending tokens without USD price
Currently tokens without a USD price aren't listed in the app and can't be sent.

AC:
Users should be able to see tokens that don't have USD price in the app and should be able to send them.

gnardini:
### Description

There are a few rough edges UI-wise but users can now see their tokens even if they don't have a USD price and can send them as well.

Screen Shot 2022-01-06 at 02 21 51

https://user-images.githubusercontent.com/6062888/148335028-8895a59f-0b48-474a-9c12-27d430c50bd1.mov

### Other changes

- The tokens in the tokens list are now ordered by USD balance.
- Now ignoring token USD prices if their `priceFetchedAt` is too old.

### Tested

Manually and with unit tests.

### How others should test

- Get a token that doesn't have a USD price. For ease of testing you can use https://test-token-faucet.web.app/
- Once you have some, go to the token list and check that you can see the token balance.
- Try to send some tokens and make sure you can do that.

### Related issues

- Fixes #1722

### Backwards compatibility

N/A


packages/mobile/src/analytics/ValoraAnalytics.test.ts
```diff
@@ -83,11 +83,11 @@ describe('ValoraAnalytics', () => {
       sHasVerifiedNumber: false,
       sLanguage: 'es-419',
       sLocalCurrencyCode: 'PHP',
-      sOtherTenTokens: 'TT:0,MOO:0',
```
gnardini, line 86: @jeanregisser This change caused 0-balance tokens to not show up here anymore. Is that ok? Right now in production we are sending the same tokens for everyone with zero balance, right?

packages/mobile/src/fees/saga.ts
```diff
@@ -211,7 +211,7 @@ function* mapFeeInfoToUsdFee(feeInfo: FeeInfo) {
     tokensByCurrencySelector
   )
   const tokenInfo = tokensInfo[feeInfo.currency]
-  if (!tokenInfo) {
+  if (!tokenInfo?.usdPrice) {
     throw new Error(`No token info found for ${feeInfo.currency}`)
```
silasbw, line 215: nit: I'm not sure if this error.message is logged, but its a misleading after the change above (either tokenInfo or usdPrice can by falsey).  Suggestion:

```
`Missing tokenInfo or tokenInfo.usdPrice for ${feeInfo.currency}`
```

packages/mobile/src/tokens/selectors.ts
```diff
@@ -30,21 +30,32 @@ export const tokensListSelector = createSelector(tokensByAddressSelector, (token
   return Object.values(tokens).map((token) => token!)
 })
 
-export const tokensWithBalanceSelector = createSelector(tokensListSelector, (tokens) => {
+type TokenBalanceWithUsdPrice = TokenBalance & {
+  usdPrice: BigNumber
+}
+
+export const tokensWithUsdBalanceSelector = createSelector(tokensListSelector, (tokens) => {
```
silasbw, line 37: naming nit: "USD Value" instead of "USD Balance" (i.e., `tokensWithUsdValueSelector`)

I don't really have a great reason and basing this on conversational language (we say "what's the value in USD?" vs "what's the balance in USD?")

packages/mobile/src/tokens/selectors.ts
```diff
@@ -30,21 +30,32 @@ export const tokensListSelector = createSelector(tokensByAddressSelector, (token
...
   return tokens.filter((tokenInfo) =>
-    tokenInfo.balance.multipliedBy(tokenInfo.usdPrice).gt(STABLE_TRANSACTION_MIN_AMOUNT)
-  )
+    tokenInfo.balance.multipliedBy(tokenInfo.usdPrice ?? 0).gt(STABLE_TRANSACTION_MIN_AMOUNT)
+  ) as TokenBalanceWithUsdPrice[]
 })
 
-export const coreTokensSelector = createSelector(tokensListSelector, (tokens) => {
-  return tokens.filter((tokenInfo) => tokenInfo.isCoreToken === true)
+export const tokensWithTokenBalanceSelector = createSelector(tokensListSelector, (tokens) => {
```
silasbw, line 43: naming nit: `tokensWithMinimumBalanceSelector`

packages/mobile/src/tokens/selectors.ts
```diff
@@ -60,11 +71,11 @@ export const tokensByCurrencySelector = createSelector(tokensListSelector, (toke
 })
 
 // Returns the token with the highest usd balance to use as default.
-export const defaultTokenSelector = createSelector(tokensListSelector, (tokens) => {
+export const defaultTokenSelector = createSelector(tokensWithTokenBalanceSelector, (tokens) => {
   let maxTokenAddress: string = ''
   let maxBalance: BigNumber = new BigNumber(-1)
   for (const token of tokens) {
-    const usdBalance = token.balance.multipliedBy(token.usdPrice)
+    const usdBalance = token.balance.multipliedBy(token.usdPrice ?? 0)
```
silasbw, line 78: If a token "doesn't have a USD price" is that equivalent to a USD price of 0?

packages/mobile/src/analytics/ValoraAnalytics.test.ts
jeanregisser: Yes I think this is fine.

jeanregisser:
Will do a proper review tomorrow, but a quick question: Shouldn't the TT token be displayed last in the Select Token list in the video?

packages/mobile/src/fees/saga.ts
gnardini: Yes, to clarify this shouldn't happen since the feeCurrency should only be CELO, cUSD, cEUR or cREAL, all of which have usd price, but it's good to make the message clearer

packages/mobile/src/tokens/selectors.ts
gnardini: Hmm, I changed it but just for the record I find balance clearer than value, maybe because English is not my first language? I just don't use the word 'value' very often in this context

packages/mobile/src/tokens/selectors.ts
gnardini: Hmm, not sure, I changed this to be based on `tokensWithUsdValueSelector` instead so we don't have to answer that :)

gnardini:
@jeanregisser Yup, that makes sense to me, switched to order in the bottom sheet by balance as well.

Screen Shot 2022-01-06 at 15 41 40

packages/mobile/src/tokens/selectors.ts
```diff
@@ -19,7 +19,7 @@ export const tokensByAddressSelector = createSelector(
       tokenBalances[tokenAddress] = {
         ...storedState,
         balance: new BigNumber(storedState.balance),
-        usdPrice: usdPrice.isNaN() ? new BigNumber(0) : usdPrice,
+        usdPrice: usdPrice.isNaN() ? null : usdPrice,
```
dievazqu, line 22: I'm concern about showing prices that were fetched long time ago. Right now, if we can't calculate the price for a token we are not deleting its previous usdPrice. I see two options here, one would be to remove every usdPrice before calculating new prices in the cloud function code. Other option would be to use the `priceFetchedAt` property to see if the price is too old.
In any case, we could do that in another PR, but I just wanted you to remind you that's something we need to address.

packages/mobile/src/send/SendAmount/SendAmount.test.tsx
```diff
@@ -352,6 +355,34 @@ describe('SendAmount', () => {
         },
       })
     })
+
+    it("Doesn't allow inputting in local currency if token has no usd price", () => {
```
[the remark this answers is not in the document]
gnardini: In that scenario it switches to token based input, but if you switch back to a token with a USD price then it goes back to local currency input. I can add a test for it.

packages/mobile/src/send/SendAmount/SendAmountValue.tsx
```diff
@@ -95,41 +97,52 @@ function SendAmountValue({
...
-                <Text
-                  allowFontScaling={false}
-                  adjustsFontSizeToFit={true}
-                  numberOfLines={1}
-                  style={styles.secondarySymbol}
-                >
-                  {localCurrencySymbol || localCurrencyCode}
+          {tokenHasUsdPrice && (
+            <View style={styles.valueContainer}>
+              {!usingLocalAmount && (
```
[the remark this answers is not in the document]
gnardini: Yup, this is the subtitle, not the main input. If inputting using tokens, the subtitle is shown in local currency and viceversa.

packages/mobile/src/tokens/selectors.ts
gnardini: I can address it here, just add a condition based on `priceFetchedAt` and set the `usdPrice` as `null` if it's too old (is 1 day too much?)

packages/mobile/src/home/HomeTokenBalance.tsx
```diff
@@ -15,10 +15,14 @@ import ProgressArrow from 'src/icons/ProgressArrow'
...
-import { tokensWithBalanceSelector, totalTokenBalanceSelector } from 'src/tokens/selectors'
+import {
+  tokensWithTokenBalanceSelector,
+  tokensWithUsdValueSelector,
+  totalTokenBalanceSelector,
+} from 'src/tokens/selectors'
 
 function TokenBalance() {
-  const tokenBalances = useSelector(tokensWithBalanceSelector)
+  const tokenBalances = useSelector(tokensWithUsdValueSelector)
```
jeanregisser, line 25: nit: Should we rename the variable too?
```suggestion
  const tokensWithUsdValue = useSelector(tokensWithUsdValueSelector)
```

packages/mobile/src/tokens/selectors.ts
jeanregisser: I'm also on @gnardini's side here, as I'm seeing this function listing the tokens for which we have a balance that can be represented in USD.
Isn't it what this function is?

packages/mobile/src/tokens/selectors.ts
```diff
@@ -16,10 +17,11 @@ export const tokensByAddressSelector = createSelector(
         continue
       }
       const usdPrice = new BigNumber(storedState.usdPrice)
+      const tokenUsdPriceIsStale = storedState.priceFetchedAt < Date.now() - ONE_DAY_IN_MILLIS
```
jeanregisser, line 20: It seems we should surface this more explicitly with a dedicated config constant in `src/config.ts`. What do you think?

packages/mobile/src/tokens/utils.ts
```diff
@@ -19,3 +20,9 @@ export function getHigherBalanceCurrency(
   }
   return maxCurrency
 }
+
+export const sortByUsdBalance = (token1: TokenBalance, token2: TokenBalance) => {
+  const token1UsdBalance = token1.balance.multipliedBy(token1.usdPrice ?? 0)
+  const token2UsdBalance = token2.balance.multipliedBy(token2.usdPrice ?? 0)
+  return token2UsdBalance.minus(token1UsdBalance).toNumber()
+}
```
jeanregisser, line 28: `BigNumber` has a [`comparedTo` method](https://mikemcl.github.io/bignumber.js/#cmp) for this purpose which doesn't require converting to a `number` which could overflow.
also nit: plain old `function` declaration is a few keystrokes shorter than `const` + `arrow function` 😄 
```suggestion
// Sort by usd balance (descending)
export function sortByUsdBalance(token1: TokenBalance, token2: TokenBalance) {
  const token1UsdBalance = token1.balance.multipliedBy(token1.usdPrice ?? 0)
  const token2UsdBalance = token2.balance.multipliedBy(token2.usdPrice ?? 0)
  return token2UsdBalance.comparedTo(token1UsdBalance)
}
```

packages/mobile/test/schemas.ts
```diff
@@ -645,6 +645,7 @@ export const v19Schema = {
         usdPrice: '1',
         balance: '10',
         isCoreToken: true,
+        priceFetchedAt: Date.now(),
```
jeanregisser, line 648: We should avoid changing past schemas. The main reason is to make it easier for us to track down how it evolves over time and prevent unexpected issues for existing users.

packages/mobile/src/tokens/selectors.ts
```diff
@@ -30,21 +32,32 @@ export const tokensListSelector = createSelector(tokensByAddressSelector, (token
...
-    b.balance.multipliedBy(b.usdPrice).comparedTo(a.balance.multipliedBy(a.usdPrice))
-  )
+export const tokensByUsdBalanceSelector = createSelector(
+  tokensWithUsdValueSelector,
+  (tokensList) => {
+    return tokensList.sort((a, b) =>
+      b.balance.multipliedBy(b.usdPrice).comparedTo(a.balance.multipliedBy(a.usdPrice))
+    )
+  }
+)
```
jeanregisser, line 57: I'm having 2nd thoughts on this. I propose we do this:
```suggestion
// Note: tokens which don't have a usd price will be listed last
export const tokensByUsdBalanceSelector = createSelector(
  tokensListSelector,
  (tokensList) => {
    return tokensList.sort(sortByUsdBalance)
  }
)
```

jeanregisser (commented):
Great work! 🚀

packages/mobile/src/tokens/TokenBalances.tsx
```diff
@@ -59,18 +60,24 @@ function TokenBalancesScreen({ navigation }: Props) {
...
         </View>
       </View>
     )
   }
 
-  return <ScrollView style={styles.scrollContainer}>{tokens.map(getTokenDisplay)}</ScrollView>
+  return (
+    <ScrollView style={styles.scrollContainer}>
+      {tokens.sort(sortByUsdBalance).map(getTokenDisplay)}
+    </ScrollView>
```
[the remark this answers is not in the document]
gnardini: Agreed, I will keep it in mind.

packages/mobile/src/tokens/reducer.ts
```diff
@@ -9,6 +9,7 @@ interface BaseToken {
   imageUrl: string
   name: string
   symbol: string
+  priceFetchedAt: number
```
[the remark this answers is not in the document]
gnardini: I was trying to avoid it tbh. This should be overwritten as soon as the token info is fetched the next time, so I figured it was fine not to migrate?

packages/mobile/test/schemas.ts
gnardini: I wasn't sure how to go about adding this, but I just copy/pasted the tokenBalances object and added the extra field, let me know if you think we should do it differently :)

packages/mobile/src/tokens/reducer.ts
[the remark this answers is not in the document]
jeanregisser: ah yes, or maybe `priceFetchedAt?: number` then?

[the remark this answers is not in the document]
gnardini: Sounds good, done!

jeanregisser (approved):
This is great!! thanks! 💯

packages/mobile/src/tokens/selectors.ts
silasbw: I don't often hear "what is the balance of CELO in USD?".  Balance connotes offsetting in accounting.  E.g., "What's the number I need to balance this?"

I do often hear "what is the value of CELO in USD?".

packages/mobile/src/tokens/selectors.ts
gnardini: It depends on how you think about "having a price of zero", but I don't think it should ever be zero. 

Not having a usd price happens when the token is not traded in any exchanges.

Having a price of zero should not ever happen.

I don't understand what you're aiming at though 🤔

ValoraQA:
Hey @MuckT We have verified this task on the latest Android Internal build V:1.24.1 and latest iOS TestFlight Build V:1.24.1 and updated the status in Gsheet 

**Sheet link** :https://docs.google.com/spreadsheets/d/15nGo3iJAEVJ76wHae9PvYuWI7S2bovKAQPX_edtwPEI/edit#gid=1469788558

**Devices :** iPhone 12(14.7.1) , Google Pixel 2XL(11.0)

CC: @gnardini

ValoraQA:
Hey @MuckT We have verified this task on the latest Android Internal build V:1.24.1 and latest iOS TestFlight Build V:1.24.1 and updated the status in Gsheet

**Sheet Link**:
https://docs.google.com/spreadsheets/d/15nGo3iJAEVJ76wHae9PvYuWI7S2bovKAQPX_edtwPEI/edit#gid=1469788558

**Devices** : Samsung Galaxy M40 (10.0)

CC: @gnardini

ValoraQA:
Hey @MuckT we have verified the issue on latest Android Internal release build V1.24.1, latest iOS Test Flight release build V:1.24.1 and updated the status in below sheet.

**Sheet link :** https://docs.google.com/spreadsheets/d/15nGo3iJAEVJ76wHae9PvYuWI7S2bovKAQPX_edtwPEI/edit#gid=1469788558

**Devices:** iPhone 7(13.3), Redmi Note 5 Pro (9.0)
CC: @gnardini
Thank..!!

PR #1724, opened 2022-01-06
Allow parallelizeBuildables on iOS builds to improve build time

kathaypacific:
### Description

The purpose of this PR is to enable us to use "dependency order" for the "build order". This is the default setting for xcode 13 and also since RN 0.62. The benefits include better build performance through more parallelisation and reduced dependency cycle errors.

Screenshot 2022-01-06 at 15 26 55

To build in this way, dependencies need to require the `React-Core` pod instead of `React` . For this reason, many dependencies needed to be updated. Where possible I have updated to the latest versions, some libs were deprecated so I have updated to the new recommended option. 

Of note:
- react-native-camera has been deprecated in favour of [react-native-vision-camera](https://github.com/mrousavy/react-native-vision-camera), however react-native-vision-camera seems to serve a different purpose and doesn't have the QR barcode scanning functionalities we are needing. I opted to upgrade to the latest version of react-native-camera even though it is deprecated, as we were one major version behind and the latest version did have the podspec change we need for this piece of work.
- react-native-exit-app also seems a little abandoned, i used a commit hash because https://github.com/wumke/react-native-exit-app/pull/45#issuecomment-865867215

New patches:
- react-native-keep-awake is deprecated in favour of [expo-keep-awake](https://www.npmjs.com/package/expo-keep-awake), i opted to patch react-native-keep-awake since we don't have/want expo 
- react-native-sms/react-native-fast-crypto seem abandoned, i opted to patch instead of waiting for a new release with these 14 month old PRs [here](https://github.com/tkporter/react-native-sms/pull/90) and [here](https://github.com/EdgeApp/react-native-fast-crypto/pull/32)
- clevertap-react-native doesn't have the podspec change we need in the latest release, so i have patched the existing version for now
- react-native-securerandom seems abandoned, we are on the latest version. 
- react-native-svg we are on the latest version, there is a new pre-release that contains the fix we need but it has been on pre-release since 08/2021. there is an open [issue](https://github.com/react-native-svg/react-native-svg/issues/1683) to publish the pre-release so for now i have patched.
- react-native-netinfo - after upgrading to the latest, the android CI started failing due to not correctly detecting the network connection status. rather than holding up the rest of this work, patch for now and investigate later. more context [here](https://valora-app.slack.com/archives/CL7BVQPHB/p1642086517037100)

### Other changes

Unlink react-native-geth and react-native-securerandom to resolve the below build warning. This was actually pretty straightforward, didn't cause any problems.

Screenshot 2022-01-07 at 14 23 48

### Tested

Manually

### How others should test

A regression to see that app functionality hasn't been affected. e.g. QR scanning, app data persistence when force closing/opening the app.

### Backwards compatibility

Yes


jeanregisser:
w00t w00t!! Amazing.
It seems e2e run is ~4mins faster thanks to this! ❤️ 🚀

packages/mobile/e2e/src/usecases/WalletConnectV1.js
```diff
@@ -323,6 +323,7 @@ export default WalletConnect = () => {
     await element(by.text('Connected Apps')).tap()
     await element(by.text('Tap to Disconnect')).tap()
     await element(by.text('Disconnect')).tap()
+    await element(by.id('BackChevron')).tap()
```
kathaypacific, line 326: after apps are disconnected, the user is still on the Sessions screen but the next assertion for `ConnectedApplications/value` is on the Settings screen. because this assertion is in the `afterAll` hook, the tests failed without being captured by the test reports or test artifacts. i think for correctness the command to navigate back to the settings screen should exist...

(i also ran this test locally on the main branch, and it seems that the tests pass even though the app is still on the Sessions screen and the `ConnectedApplications/value` is not found :/ i'm not sure why it can pass, but the behaviour of disconnecting a connected app has not changed in this PR)

jeanregisser (approved):
Amazing! 💯 🚀

PR #1729, opened 2022-01-06
Fix WalletConnect `eth_signTransaction` requests handling

### LINKED ISSUE #1559: 'Sign a transaction' failing with WalletConnect - Optional Gas Required
### Current behavior
When creating e2e tests for WalletConnect 'Sign a transaction' was failing after accepting. 

### Desired behavior
WalletConnectV1 should handle sign transaction (eth_signTransaction)​ with or without gas [based on the docs](https://docs.walletconnect.com/1.0/quick-start/dapps/node#sign-transaction-eth_signtransaction).

### Steps to Reproduce
Reproducible on [e2e-walletconnect](https://github.com/valora-inc/wallet/tree/tomm/e2e-walletconnect).
1. Enable the test "Then is able to sign a transaction" in `usecases/WalletConnectV1.js`
2. `yarn install && yarn build && cd packages/mobile && yarn test:e2e:ios -f WalletConnect.spec.js`

### Relevant Logs With Gas Excluded
```
redux/saga@logger/{"type":"WALLETCONNECT/ACCEPT_REQUEST_V1","request":{"id":1638408490234137,"jsonrpc":"2.0","method":"eth_signTransaction","params":[{"from":"0x6131a6d616a4be3737b38988847270a64bc10caa","to":"0xe5f5363e31351c38ac82dbadead91fd5a7b08846","gasPrice":"0x1","value":"0x1","data":"0x"}]},"peerId":"d387e766-1327-4048-a49b-9c6c5317f5a7"}
WalletConnect/saga-v1@acceptRequest/"gas" is missing
ValoraAnalytics/Tracking event wc_request_accept_error with properties: {"timestamp":1638408492160,"sessionId":"42d817afd6f9d5be938253fad9904d460998f2a8da7fba4fb9ed476654692e","userAddress":"0x6131a6d616a4be3737b38988847270a64bc10caa","celoNetwork":"alfajores","sAccountAddress":"0x6131a6d616a4be3737b38988847270a64bc10caa","sWalletAddress":"0x6131a6d616a4be3737b38988847270a64bc10caa","sPhoneCountryCallingCode":"+1","sPhoneCountryCodeAlpha2":"US","sCountryCodeAlpha2":"US","sLanguage":"en-US","sTotalBalanceUsd":787.530008012507,"sTokenCount":3,"sOtherTenTokens":"","sCusdBalance":49.07247683830868,"sCeurBalance":165.29173707525877,"sCeloBalance":83.60204329621429,"sLocalCurrencyCode":"USD","sHasVerifiedNumber":false,"sHasCompletedBackup":true,"sDeviceId":"06D56C71-324C-4C3D-9B2B-21BA25B250B3","sAppVersion":"1.21.0","sAppBuildNumber":"82","sAppBundleId":"org.celo.mobile.test","sCurrentScreenId":"WalletConnectActionRequest","sPrevScreenId":"WalletHome","version":1,"dappName":"WalletConnectV1 E2E","dappUrl"<…>
```

### Relevant Logs with Gas Included
```
redux/saga@logger/{"type":"WALLETCONNECT/ACCEPT_REQUEST_V1","request":{"id":1638408969953139,"jsonrpc":"2.0","method":"eth_signTransaction","params":[{"from":"0x6131a6d616a4be3737b38988847270a64bc10caa","to":"0xe5f5363e31351c38ac82dbadead91fd5a7b08846","gasPrice":"0x2540be400","gas":"0x9c40","value":"0x1","data":"0x"}]},"peerId":"f701bb2f-44d4-4edc-bafb-3fc63f1cf838"}
WalletConnect/saga-v1@acceptRequest/One of the values "chainId", "gasPrice", or "nonce" couldn't be fetched: {"gasPrice":"0x2540be400"}
```

jeanregisser:
### Description

WalletConnect v1 utils strips some important tx fields (on the dapp side) which prevents `eth_signTransaction` calls from working.
See all details in #1559.

Ideally this could/should be fixed in WalletConnect v1 utils too, but it's not clear whether such a fix would be accepted and it may take quite some time before it's released and deployed. We're discussing this in [this Slack thread](https://valora-app.slack.com/archives/C014SRL4URY/p1641463814007200).

Working around it in Valora fixes the issue directly and also ensures we're more robust with missing or incorrect tx params.

Note: I'll open another PR to re-use the same logic for `eth_sendTransaction`. That one already works but the new logic added in `eth_signTransaction` is a bit better and should be shared.

### Tested

- Added new unit tests
- Manually, swapping on https://mento.finance/ works once connected via WalletConnect with the following situations:
  - when I only have CELO in my balance
  - when I have only cUSD in my balance

### How others should test

See above.

### Related issues

- Fixes #1559
- celo-org/use-contractkit#89

### Backwards compatibility

Yes


packages/mobile/src/walletConnect/request.ts
```diff
@@ -29,14 +31,41 @@ export function* handleRequest({ method, params }: { method: string; params: any
...
+        // Pass undefined to use CELO to pay for gas.
+        const feeCurrencyAddress: string | undefined =
+          feeCurrency === Currency.Celo ? undefined : yield call(getCurrencyAddress, feeCurrency)
+
+        rawTx.feeCurrency = feeCurrencyAddress
+        // Note: we're resetting gas and gasPrice here because if the feeCurrency has changed, we need to recalculate them
+        rawTx.gas = undefined
+        rawTx.gasPrice = undefined
+      }
+      const tx: CeloTx = yield call(normalizer.populate.bind(normalizer), rawTx)
```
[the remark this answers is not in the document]
nategraf: It looks like this eventually makes an `eth_gasPrice` RPC call to the supporting Geth node. This will suggest an appropriate gas price, which by default is 5x the GPM. So this seems like it should result in a setting that will work.

https://github.com/celo-org/celo-blockchain/blob/8cd59acbb3ff70033c45867d2d62dbda0032c36a/contracts/gasprice_minimum/gasprice_minimum.go#L55-L60

packages/mobile/src/walletConnect/request.ts
```diff
@@ -29,14 +31,41 @@ export function* handleRequest({ method, params }: { method: string; params: any
...
+        // This will use CELO to pay for fees if the user has a balance,
+        // otherwise it will fallback to the first currency with a balance
+        const feeCurrency: Currency = yield call(chooseFeeCurrency, Currency.Celo)
+        // Pass undefined to use CELO to pay for gas.
+        const feeCurrencyAddress: string | undefined =
+          feeCurrency === Currency.Celo ? undefined : yield call(getCurrencyAddress, feeCurrency)
+
+        rawTx.feeCurrency = feeCurrencyAddress
+        // Note: we're resetting gas and gasPrice here because if the feeCurrency has changed, we need to recalculate them
+        rawTx.gas = undefined
```
nategraf, line 56: One caveat here is that there are cases where gas estimation will fail even if the transaction will end up clearing. Clearing the gas may result in errors for the dApp. E.g. If a dApp developer is doing a two step approve and exchange and requesting both signatures together, they will set the gas on the second transaction because if estimateGas is run before the approve completes, execution will fail.

I saw in the Slack thread there is some mention of `estimateGas` being broken for non-CELO fee currencies, but I am not aware of any issue right now. It is true though that using a non-CELO fee currency uses more gas, by about 50k, so if the caller estimates using CELO and this code switches it to cUSD, the transaction may end up with too little gas.

Of these two scenarios, failures due to clearing the gas value seem more likely to me than failures due to not clearing it.

MuckT:
@jeanregisser I locally brought these changes into the https://github.com/valora-inc/wallet/pull/1412 and the e2e test around sign a transaction is passing.

packages/mobile/src/walletConnect/request.ts
gnardini: Interesting problem, nice catch!

I think not having CELO to pay for gas is a somewhat common situation, especially for Valora users where they are encouraged to cash in cUSD/cEUR and not CELO. dApps are also usually not very good at catching it and show unhelpful errors when this happens.

My point is that the switch from CELO fee to cUSD/cEUR fee might happen somewhat often here. Don't really have the data to back it up, just my intuition (we could look it up).

How about just plain adding 50k to the `gas` field when this happens? Would that help solve for both use-cases or are there other complications I might be missing?

packages/mobile/src/walletConnect/request.ts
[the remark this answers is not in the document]
jeanregisser: Great! Thanks for checking 💪

packages/mobile/src/walletConnect/request.ts
jeanregisser: > E.g. If a dApp developer is doing a two step approve and exchange and requesting both signatures together, they will set the gas on the second transaction because if estimateGas is run before the approve completes, execution will fail.

Do you mean that the dapp may call `eth_signTransaction` without having called estimateGas on their end and sent  individually?
So if it does that for approve and exchange, it will indeed fail for us.
Well, great catch!! 🥇 

So yeah ideally we should really not touch what the dapp is providing. But since we have this bug with WalletConnect v1 utils, we have to do something.
I really like what @gnardini is proposing then.

In addition, I'll also limit the workaround to when `type` is not set. Because in that case the TX data is sent as is. 
See https://github.com/WalletConnect/walletconnect-monorepo/blob/c6b26481c34848dbc9c49bb0d024bda907ec4599/packages/helpers/utils/src/ethereum.ts#L45

packages/mobile/src/walletConnect/request.ts
```diff
@@ -29,14 +36,56 @@ export function* handleRequest({ method, params }: { method: string; params: any
...
+      // so it's nice to be flexible and still allow the transaction to be signed (and sent) successfully
+
+      const rawTx = { ...params[0] }
+      let tx
+      // Provide an escape hatch for dapp developers who don't want any normalization
+      if (rawTx.__skip_normalization) {
+        // Remove this custom field which may cause issues down the line
+        delete rawTx.__skip_normalization
+        tx = rawTx
+      } else {
```
jeanregisser, line 53: I haven't tested this escape hatch end-to-end yet.
Could this cause issues between the dapp and Valora processing it.

Of course because of the WC v1 utils stripping fields, someone wanted to use it would have to set both `type` and `__skip_normalization`.

I'll document this if we decide this escape hatch is viable.

gnardini (approved):
Everything looks good to be, great job on this!

packages/mobile/src/walletConnect/request.ts
nategraf: Adding 50k to the gas value when setting a non-CELO fee currency seems reasonable. That should be the correct value as it is the value currently set in the `intrinsicGasForAlternativeFeeCurrency` on-chain 👌

packages/mobile/src/walletConnect/request.ts
```diff
@@ -29,14 +36,56 @@ export function* handleRequest({ method, params }: { method: string; params: any
...
+      // IMPORTANT: We need to normalize the transaction parameters
+      // WalletConnect v1 utils currently strips away important fields like `chainId`, `feeCurrency`, `gatewayFee` and `gatewayFeeRecipient`
+      // See https://github.com/WalletConnect/walletconnect-monorepo/blame/c6b26481c34848dbc9c49bb0d024bda907ec4599/packages/helpers/utils/src/ethereum.ts#L66-L86
+      // Also the dapp developer may have omitted some of the needed fields,
+      // so it's nice to be flexible and still allow the transaction to be signed (and sent) successfully
+
+      const rawTx = { ...params[0] }
+      let tx
+      // Provide an escape hatch for dapp developers who don't want any normalization
+      if (rawTx.__skip_normalization) {
```
jmrossy, line 49: Will WC strip this field out though?

packages/mobile/src/walletConnect/request.ts
```diff
@@ -29,14 +36,56 @@ export function* handleRequest({ method, params }: { method: string; params: any
   yield call(unlockAccount, account)
 
   switch (method) {
-    case SupportedActions.eth_signTransaction:
-      return (yield call(wallet.signTransaction.bind(wallet), params[0])) as EncodedTransaction
+    case SupportedActions.eth_signTransaction: {
```
jmrossy, line 39: Should we normalize eth_send and eth_sign the same way?

jmrossy (approved):
LGTM with two optional comments

packages/mobile/src/walletConnect/request.ts
jeanregisser: Yes it will for v1 requests, so developers wanting to use it would have to both set `type = null` and `__skip_normalization`.

packages/mobile/src/walletConnect/request.ts
jeanregisser: Yes we should, I'll open a PR applying the same logic to both.

ValoraQA:
Hey @MuckT we have verified the issue on Latest Android Internal build V:1.24.1 and Latest iOS TestFlight build V:1.24.1 and updated the status in Gsheet 
**Sheet link** : https://docs.google.com/spreadsheets/d/15nGo3iJAEVJ76wHae9PvYuWI7S2bovKAQPX_edtwPEI/edit#gid=2037765399

**Observation** : There is no option to connect wallet to given link and also we would require cREAL for the transaction in our account balance 

we have raised query for the same : [Query link](https://github.com/valora-inc/wallet/pull/1760)

**Devices** : Google Pixel 2XL(11.0) , Google Pixel (10.0) , iPhone 12(14.7), iPhone 13 Mini (15.1.1)

ValoraQA:
Hey @MuckT we have verified the issue on Latest iOS Test Flight build V:1.24.1 and updated the status in Gsheet 

**Sheet link:**  https://docs.google.com/spreadsheets/d/15nGo3iJAEVJ76wHae9PvYuWI7S2bovKAQPX_edtwPEI/edit#gid=2037765399

**Observation** : There is no option to connect wallet to given link and also we would require cREAL for the transaction in our account balance 

We have raised query for the same : [Query link](https://github.com/valora-inc/wallet/pull/1760)

**Devices:**iPhone XS Max (13.5.1)
CC: @jeanregisser 
Thanks..!!

ValoraQA:
Hey @MuckT we have verified the issue on Latest Android Internal release build V:1.24.1 and updated the status in Gsheet

**Sheet link** : https://docs.google.com/spreadsheets/d/15nGo3iJAEVJ76wHae9PvYuWI7S2bovKAQPX_edtwPEI/edit#gid=2037765399

**Observation** : There is no option to connect wallet to given link and also we would require cREAL for the transaction in our account balance

We have raised query for the same : [Query link](https://github.com/valora-inc/wallet/issues/1760)

**Devices** : Samsung Galaxy A8+(8.0)
CC: @jeanregisser 
Thanks..!!

ValoraQA:
Hey @MuckT we have verified the issue on Latest Android Internal Release build V:1.24.1 and updated the status in Gsheet

**Sheet link:** https://docs.google.com/spreadsheets/d/15nGo3iJAEVJ76wHae9PvYuWI7S2bovKAQPX_edtwPEI/edit#gid=2037765399

**Observation:** There is no option to connect wallet to given link and also we would require cREAL for the transaction in our account balance

We have raised query for the same : [Query link](https://github.com/valora-inc/wallet/issues/1760)

**Devices:**  Redmi Note 5 Pro (9.0)
CC: @jeanregisser 
Thanks..!!

jmrossy:
@ValoraQA You can test mento-fi with the pre-release version here:
https://mento-fi-q5d61619k-celotools.vercel.app/

ValoraQA:
Hey @MuckT we have verified the issue on Latest Android Internal Release build and Latest iOS Test Flight Internal  Build V:1.24.1 and observed that we are able to swap tokens on "https://mento-fi-q5d61619k-celotools.vercel.app/" 

**Observation:** After clicking on "Swap" button it asks the transaction permission twice on the valora app

**Attachment :**  [mento.mp4](https://drive.google.com/file/d/1sAtFHkEsXRNgketDnngBolJu6xQTjzHF/view?usp=sharing)

**Devices:** OnePlus 7T (11.0) , iPhone 13 mini (15.1.1)

**CC:** @jmrossy 

Thanks..!!

PR #1732, opened 2022-01-07
Replace faker with @faker-js/faker

MuckT:
### Description

Replaces `faker` with `@faker-js/faker` some context around this change can be [read here](https://github.com/faker-js/faker#faq---what-happened-to-the-original-fakerjs).

### Other changes

N/A

### Tested

N/A

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

Yes


jeanregisser (commented):
Thanks! We also need to update the import since the lib as a new `@withshpherd/` suffix.

ValoraQA:

PR #1733, opened 2022-01-07
Fix fee calculation for non cUSD tokens

### LINKED ISSUE #1727: “Estimated fees” changes automatically on “Sending” screen when user taps on “Send” button for completing sending flow.
**Frequency:** 100%

**Repro on build version:** Android Alfajores Nightly V1.24.0,  iOS test flight Alfajores Nighlty V1.24.0
**Repro on devices:** Google Pixel 2XL (11.0), Google Pixel XL (10.0),  iPhone 12 (14.7.1), iPhone 13 mini (15.1.1)

**Pre-condition:** 
1. User must have installed the app & complete onboarding flow.
2. User must have valid balance
3. User must be on Sending screen on sending flow

**Repro Steps:**
1. Tap on “Send” Button on send screen & observe

**Current Behavior:**  “Estimated fees” changes automatically on “Sending” screen when user taps on “Send” button for completing sending flow.  

**Expected Behavior:**  “Estimated fees” should not change  automatically on “Sending” screen when user taps on “Send” button for completing sending flow. 

**Investigation:** This issue is reproducible for iOS platform when user send cEUR currency. 

**Impact:** Bad user experience.

**Attachment:** [EstimateChnages.mp4](https://drive.google.com/file/d/1FcOaNFbe0VTEhOhlPWmfWR-YZ30e1oLk/view?usp=sharing)

gnardini:
### Description

There were a couple of issues.
First, the token to use as fee currency was being chosen at random. Not a huge deal I guess, but switched back to using the token with highest balance for consistency.
The more important issue is that we were sending to the `FeeDrawer` the `totalFee` in USD but also sending the fee currency and the exchange rate, so in the end we were transforming the value and ended up showing something inaccurate. Since we're already calculating the fee in USD, we are not just passing the currency as `Currency.Dollar` and letting the `FeeDrawer` component transform it into the local currency normally (`CurrencyDisplay` ends up doing that).

### Other changes

N/A

### Tested

I did a TDD approach here, if you run the new test with the old code it fails, but after the changes it works.

### How others should test

Try to send using different fee currencies (decided by the highest balance, so try while having more cUSD than cEUR/CELO and then by having more cEUR than cUSD/CELO) and see that the fee currency has an expected value. Not sure how to decide what "expected value" is though 🤔 

### Related issues

- Fixes #1727

### Backwards compatibility

N/A

### DIFF
--- a/packages/mobile/src/send/SendConfirmation.test.tsx
+++ b/packages/mobile/src/send/SendConfirmation.test.tsx
@@ -188,7 +188,7 @@ describe('SendConfirmation', () => {
     const feeComponent = getByTestId('feeDrawer/SendConfirmation/totalFee/value')
     expect(getElementText(feeComponent)).toEqual('₱0.0266')
 
-    // Subtotal is $1.33, which is added the fee amount.
+    // Subtotal is $1.33, which is added to the fee amount.
     const totalComponent = getByTestId('TotalLineItem/Total')
     expect(getElementText(totalComponent)).toEqual('₱1.36')
   })
--- a/packages/mobile/src/send/SendConfirmation.test.tsx
+++ b/packages/mobile/src/send/SendConfirmation.test.tsx
@@ -60,6 +60,26 @@ type ScreenProps = StackScreenProps<
   Screens.SendConfirmation | Screens.SendConfirmationModal
 >
 
+const mockFeeEstimates = {
+  [FeeType.SEND]: {
+    usdFee: '0.02',
+    lastUpdated: 500,
+    loading: false,
+    error: false,
+    feeInfo: mockFeeInfo,
+  },
+  [FeeType.INVITE]: {
+    usdFee: '0.04',
+    lastUpdated: 500,
+    loading: false,
+    error: false,
+    feeInfo: mockFeeInfo,
+  },
+  [FeeType.EXCHANGE]: undefined,
+  [FeeType.RECLAIM_ESCROW]: undefined,
+  [FeeType.REGISTER_DEK]: undefined,
+}
+
 describe('SendConfirmation', () => {
   beforeEach(() => {
     jest.clearAllMocks()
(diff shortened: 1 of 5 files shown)


packages/mobile/src/send/SendConfirmation.test.tsx
```diff
@@ -154,6 +157,42 @@ describe('SendConfirmation', () => {
...
+
+    fireEvent.press(getByText('feeEstimate'))
+
+    jest.runAllTimers()
+    await flushMicrotasksQueue()
+
+    const feeComponent = getByTestId('feeDrawer/SendConfirmation/totalFee/value')
+    expect(getElementText(feeComponent)).toEqual('₱0.0266')
+
+    // Subtotal is $1.33, which is added the fee amount.
```
jeanregisser, line 191:
```suggestion
    // Subtotal is $1.33, which is added to the fee amount.
```

jeanregisser (approved):
LGTM 👍

ValoraQA:
Hey @MuckT we have verified the issue on latest Android Internal Build V:1.24.1 and latest iOS Test Flight build V:1.24.1 and updated the status in sheet and also commented in ticket #1727 

**Sheet link :** https://docs.google.com/spreadsheets/d/15nGo3iJAEVJ76wHae9PvYuWI7S2bovKAQPX_edtwPEI/edit#gid=833354919

**Devices :** iPhone 12(14.7.1) , Google Pixel (11.0)

ValoraQA:
Hey @MuckT, We have verified this task using Latest Android Internal build and Test flight build V1.24.1 and added the observation in below sheet.

**Sheet Link**:
https://docs.google.com/spreadsheets/d/15nGo3iJAEVJ76wHae9PvYuWI7S2bovKAQPX_edtwPEI/edit#gid=833354919

**Observation**:  “Add description” textbox is not appearing on Sending screen when user tried to Send CELO by tapping on “Send” button from home screen and we have raised [query](https://github.com/valora-inc/wallet/issues/1800) for the same.

**Devices** : Samsung Galaxy M40 (10.0)

CC: @gnardini
Thanks..!!

ValoraQA:
Hey @MuckT we have verified the issue on latest iOS Test Flight build V:1.24.1 and updated the status in below sheet 

**Observation** : "One Time" Encryption fees is not seen on iOS platform and we have raised [query](https://github.com/valora-inc/wallet/issues/1799) for it

**Sheet link** : https://docs.google.com/spreadsheets/d/15nGo3iJAEVJ76wHae9PvYuWI7S2bovKAQPX_edtwPEI/edit#gid=833354919

**Devices** : iPhone 7(13.3) 
CC: @gnardini

PR #1734, opened 2022-01-07
Add error handling for fetching token info/balances

### LINKED ISSUE #1125: Handle error situations when displaying tokens

gnardini:
### Description

Improved the error handling based on the changes here: https://github.com/valora-inc/wallet/pull/1404/files

Some notes:
- The UX is not ideal IMO. The banner on top looks like a weird way to do refresh and it covers the hamburger menu. A button close to where the total balance is supposed to be would be more intuitive imo.
- Since blockchain-api doesn't return the symbol of the tokens, we can't show the token amount for feed items if we haven't fetched the token info, so the whole feed shows up with `-` for token price. This can be improved by returning the symbol in the blockchain-api, we should be able to populate the entire feed since the symbol is the only missing piece. Also, there appears to be a bug there because the fiat amount should be visible but isn't. We're not showing the value that comes from the blockchain-api correctly.

### Other changes

Removed some logs that were very noisy. The value they were giving was pretty low compared to how much they were spamming the logs.

### Tested

Manually and with unit tests

### How others should test

Try to open the app without an internet connection and see that the error banner shows up.

### Related issues

- Fixes #1125

### Backwards compatibility

N/A


packages/mobile/src/transactions/feed/TransactionFeed.tsx
```diff
@@ -166,10 +166,6 @@ function TransactionFeed() {
       case 'TokenExchangeV2':
         return <ExchangeFeedItem key={tx.transactionHash} exchange={tx} />
       case 'TokenTransferV2':
-        if (!tokensInfo[tx.amount.tokenAddress]) {
```
gnardini, line 169: The blockchain-api should be the one that decides to show or not an item, not the client. If we haven't fetched the token list the feed shows up empty otherwise

packages/mobile/src/transactions/feed/TransactionFeed.test.tsx
```diff
@@ -46,20 +46,6 @@ const MOCK_RESPONSE: { data: { tokenTransactionsV2: { transactions: TokenTransac
...
-            tokenAddress: '0xrandomAddress',
-            value: '0.1',
-          },
-          block: '8648978',
-          fees: [],
-          metadata: {},
-          timestamp: 1542306118,
-          transactionHash: '0x544367eaf2b01622dd1c7b75a6b19bf278d72127aecfb2e5106424c40c268e8b',
-          type: TokenTransactionTypeV2.Received,
-        },
```
jeanregisser, line 62: Why removing this one?

packages/mobile/src/tokens/reducer.ts
```diff
@@ -38,12 +38,14 @@ export interface State {
   tokenBalances: StoredTokenBalances
   loading: boolean
   error: boolean
+  lastSuccessfulFetch: number
```
jeanregisser, line 41: nit: For correctness and avoiding a migration:
```suggestion
  lastSuccessfulFetch?: number
```

packages/mobile/src/localCurrency/saga.ts
```diff
@@ -52,6 +52,7 @@ export function* fetchLocalCurrencyRateSaga() {
     if (!localCurrencyCode) {
       throw new Error("Can't fetch local currency rate without a currency code")
     }
+    // TODO: Remove EUR and CELO rates, everything is based off dollar now.
```
jeanregisser, line 55: 👍

jeanregisser (approved):
Amazing! I agree the top banner is probably too much.

Already approved but left some small questions/comments.

packages/mobile/src/transactions/feed/TransactionFeed.test.tsx
gnardini: This one was supposed to be an item with an address that the app doesn't recognize, so we weren't showing it in the feed. In this PR I'm changing it to show whatever comes from blockcchain-api, so any filtering we want should be made there.

ValoraQA:
Hey @MuckT, We have verified this task using Latest Test flight release build and Android Internal release build V1.24.1 and added the observation in below sheet.

**Sheet Link:**
https://docs.google.com/spreadsheets/d/1AFv9JWXnKfXJewhXf3nKlahB43d2v9xBU965ygXQ294/edit#gid=1656427720

**Devices:** iPhone 13 mini (15.1.1), Samsung Galaxy M40 (10.0)
Thanks..!!

ValoraQA:
Hey @MuckT, We have verified this task using Latest Android Internal release build V1.24.1 and added the observation in below sheet.

**Sheet Link**:
https://docs.google.com/spreadsheets/d/15nGo3iJAEVJ76wHae9PvYuWI7S2bovKAQPX_edtwPEI/edit#gid=1668676485

**Devices**: Samsung Galaxy A8+(8.0)

CC: @gnardini 
Thanks..!!

ValoraQA:
Hey @MuckT, We have verified this task using Latest Test flight release build V1.24.1 and added the observation in below sheet.

**Sheet Link**:
https://docs.google.com/spreadsheets/d/15nGo3iJAEVJ76wHae9PvYuWI7S2bovKAQPX_edtwPEI/edit#gid=1668676485

**Devices**: iPhone 7(13.3)

CC: @gnardini
Thanks..!!

PR #1736, opened 2022-01-08
feat(CashInBottomSheet): Setup a new experiment linking directly to ramp

jh2oman:
### Description

**Still needs Tests**

The experiment compares the existing Cash In Bottom sheet which goes through typical CICO flow to a new one that links directly to Ramp. The goal is to drive up Cash In engagement.

[Zenhub](https://app.zenhub.com/workspaces/acquisition-squad-sprint-board-6010683afabec1001a090887/issues/valora-inc/wallet/1687)
[Experiment Design
](https://www.notion.so/Experiment-2-Change-CTA-to-Add-cUSD-cEUR-10-cbb3e82e2c2c4687a3bc7c42defe5cd9)

Screen Shot 2022-01-07 at 4 57 12 PM
Screen Shot 2022-01-07 at 4 57 27 PM

### Other changes

### Tested

### How others should test

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

### DIFF
--- a/packages/mobile/src/analytics/Events.tsx
+++ b/packages/mobile/src/analytics/Events.tsx
@@ -364,7 +364,9 @@ export enum FiatExchangeEvents {
   // Add fund flow entered through home screen cash in bottom sheet
   cico_add_funds_bottom_sheet_selected = 'cico_add_funds_bottom_sheet_selected',
   cico_add_funds_bottom_sheet_impression = 'cico_add_funds_bottom_sheet_impression',
-  cico_add_funds_bottom_sheet_ramp = 'cico_add_funds_bottom_sheet_ramp',
+  cico_add_funds_bottom_sheet_ramp_selected = 'cico_add_funds_bottom_sheet_ramp_selected',
+  cico_add_funds_bottom_sheet_ramp_available = 'cico_add_funds_bottom_sheet_ramp_available',
+
   cico_cash_out_selected = 'cico_cash_out_selected',
   cico_spend_selected = 'cico_spend_selected',
   cico_fund_info = 'cico_fund_info',
--- a/packages/mobile/src/analytics/Events.tsx
+++ b/packages/mobile/src/analytics/Events.tsx
@@ -364,6 +364,7 @@ export enum FiatExchangeEvents {
   // Add fund flow entered through home screen cash in bottom sheet
   cico_add_funds_bottom_sheet_selected = 'cico_add_funds_bottom_sheet_selected',
   cico_add_funds_bottom_sheet_impression = 'cico_add_funds_bottom_sheet_impression',
+  cico_add_funds_bottom_sheet_ramp = 'cico_add_funds_bottom_sheet_ramp',
   cico_cash_out_selected = 'cico_cash_out_selected',
   cico_spend_selected = 'cico_spend_selected',
   cico_fund_info = 'cico_fund_info',
--- a/packages/mobile/src/home/CashInBottomSheet.test.tsx
+++ b/packages/mobile/src/home/CashInBottomSheet.test.tsx
@@ -25,56 +25,19 @@ const mockRampProvider = {
 }
 
 const mockRampProviderUnavailable = {
-  name: 'Ramp',
-  restricted: false,
-  paymentMethods: [PaymentMethod.Card, PaymentMethod.Bank],
-  url: 'www.fakewebsite.com',
-  logo:
-    'https://firebasestorage.googleapis.com/v0/b/celo-mobile-mainnet.appspot.com/o/images%2Framp.png?alt=media',
-  quote: [
-    { paymentMethod: PaymentMethod.Card, digitalAsset: 'cusd', returnedAmount: 100, fiatFee: 0 },
-  ],
-  cashIn: true,
-  cashOut: false,
+  ...mockRampProvider,
   unavailable: true,
 }
 
 const mockRampProviderRestricted = {
-  name: 'Ramp',
+  ...mockRampProvider,
   restricted: true,
-  paymentMethods: [PaymentMethod.Card, PaymentMethod.Bank],
-  url: 'www.fakewebsite.com',
-  logo:
-    'https://firebasestorage.googleapis.com/v0/b/celo-mobile-mainnet.appspot.com/o/images%2Framp.png?alt=media',
-  quote: [
-    { paymentMethod: PaymentMethod.Card, digitalAsset: 'cusd', returnedAmount: 100, fiatFee: 0 },
-  ],
-  cashIn: true,
-  cashOut: false,
-  unavailable: false,
 }
 
 const mockRampProviderNoCashIn = {
-  name: 'Ramp',
-  restricted: false,
-  paymentMethods: [PaymentMethod.Card, PaymentMethod.Bank],
-  url: 'www.fakewebsite.com',
-  logo:
-    'https://firebasestorage.googleapis.com/v0/b/celo-mobile-mainnet.appspot.com/o/images%2Framp.png?alt=media',
-  quote: [
-    { paymentMethod: PaymentMethod.Card, digitalAsset: 'cusd', returnedAmount: 100, fiatFee: 0 },
-  ],
+  ...mockRampProvider,
   cashIn: false,
-  cashOut: false,
-  unavailable: false,
 }
-
-// jest.mock('src/fiatExchanges/utils', () => ({
-//   ...(jest.requireActual('src/fiatExchanges/utils') as any),
-//   fetchProviders: jest.fn(() => [
-//     mockRampProvider
-//   ]),
-// }))
 
 describe('CashInBottomSheet', () => {
   const mockFetch = fetch as FetchMock
@@ -124,7 +87,7 @@ describe('CashInBottomSheet', () => {
     const { getByTestId } = render(
       <Provider
         store={createMockStore({
-          app: { rampCashInButtonExpEnabled: false },
+          app: { rampCashInButtonExpEnabled: true },
         })}
       >
         <CashInBottomSheet />
@@ -137,7 +100,7 @@ describe('CashInBottomSheet', () => {
     const { getByTestId } = render(
       <Provider
         store={createMockStore({
-          app: { rampCashInButtonExpEnabled: false },
+          app: { rampCashInButtonExpEnabled: true },
         })}
       >
         <CashInBottomSheet />
@@ -150,7 +113,7 @@ describe('CashInBottomSheet', () => {
     const { getByTestId } = render(
       <Provider
         store={createMockStore({
-          app: { rampCashInButtonExpEnabled: false },
+          app: { rampCashInButtonExpEnabled: true },
         })}
       >
         <CashInBottomSheet />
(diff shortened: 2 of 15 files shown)


packages/mobile/src/home/CashInBottomSheet.tsx
```diff
@@ -26,13 +43,59 @@ function CashInBottomSheet() {
...
+      const providers = await fetchProviders({
+        userLocation,
+        walletAddress: account,
+        fiatCurrency: localCurrency,
+        digitalAsset: currencyToBuy,
+        fiatAmount: 20,
+        digitalAssetAmount: 20,
+        txType: 'buy',
+      })
+      setModalVisible(true)
```
jeanregisser, line 66: For correctness, we should not be calling these setters within the async code.
The component may be unmounted by the time they are called.
We can pass [`onSuccess`/`onError` options to `react-async-hook`](https://github.com/slorber/react-async-hook/blob/f8569186fbf77607b0f6d46a62639cd838b4d966/test/useAsync.test.ts#L43-L44) to trigger the needed state updates.

jeanregisser (commented):
Looking good! 👍

packages/mobile/src/home/CashInBottomSheet.test.tsx
```diff
@@ -1,12 +1,90 @@
...
+  cashOut: false,
+  unavailable: false,
+}
+
+// jest.mock('src/fiatExchanges/utils', () => ({
+//   ...(jest.requireActual('src/fiatExchanges/utils') as any),
+//   fetchProviders: jest.fn(() => [
+//     mockRampProvider
+//   ]),
+// }))
```
cajubelt, line 77: stale code
```suggestion
```

packages/mobile/src/home/CashInBottomSheet.tsx
```diff
@@ -26,13 +43,66 @@ function CashInBottomSheet() {
...
+          rampURL: rampProvider?.url,
+        }
+      } catch (error) {
+        Logger.error(TAG, 'Failed to fetch CICO providers')
+      }
+    },
+    [],
+    {
+      onSuccess: () => setModalVisible(true),
+      onError: () => setModalVisible(true),
```
cajubelt, line 84: is this intentional? setting visible to true on error?

packages/mobile/src/home/CashInBottomSheet.tsx
```diff
@@ -54,15 +124,35 @@ function CashInBottomSheet() {
...
-          size={BtnSizes.FULL}
-          onPress={goToAddFunds}
-          style={styles.addFundBtn}
-          testID={'cashInBtn'}
-        />
+        {rampAvailable && rampCashInButtonExpEnabled ? (
+          <>
+            <Text style={styles.title}>
+              {t('cashInBottomSheet.titleRamp', {
+                currency: localCurrency === LocalCurrencyCode.EUR ? Currency.Euro : Currency.Dollar,
```
cajubelt, line 131: it annoys me that Currency and CiCoCurrency are different.. but that's not your fault

packages/mobile/src/home/CashInBottomSheet.tsx
```diff
@@ -54,15 +124,35 @@ function CashInBottomSheet() {
...
-        <Text style={styles.subtitle}>{t('cashInBottomSheet.subtitle')}</Text>
-        <Button
-          text={t('cashInBottomSheet.addFunds')}
-          size={BtnSizes.FULL}
-          onPress={goToAddFunds}
-          style={styles.addFundBtn}
-          testID={'cashInBtn'}
-        />
+        {rampAvailable && rampCashInButtonExpEnabled ? (
+          <>
```
cajubelt, line 128: Frontend noob question-- is this supposed to be `<>`?  Should it be `<View>`?

packages/mobile/src/home/CashInBottomSheet.test.tsx
```diff
@@ -1,12 +1,90 @@
...
+  url: 'www.fakewebsite.com',
+  logo:
+    'https://firebasestorage.googleapis.com/v0/b/celo-mobile-mainnet.appspot.com/o/images%2Framp.png?alt=media',
+  quote: [
+    { paymentMethod: PaymentMethod.Card, digitalAsset: 'cusd', returnedAmount: 100, fiatFee: 0 },
+  ],
+  cashIn: false,
+  cashOut: false,
+  unavailable: false,
+}
```
cajubelt, line 70:
```suggestion
const mockRampProviderNoCashIn = {
  ...mockRampProvider,
  cashIn: false
}
```
optional nit. could help readability

packages/mobile/src/home/CashInBottomSheet.test.tsx
```diff
@@ -28,4 +106,71 @@ describe('CashInBottomSheet', () => {
...
+      </Provider>
+    )
+    await waitFor(() => expect(getByTestId('cashInBtn')))
+  })
+  it('Regular Version is visible when ramp is unavailable', async () => {
+    mockFetch.mockResponse(JSON.stringify([mockRampProviderUnavailable]))
+    const { getByTestId } = render(
+      <Provider
+        store={createMockStore({
+          app: { rampCashInButtonExpEnabled: false },
```
cajubelt, line 127:
```suggestion
          app: { rampCashInButtonExpEnabled: true },
```
I think this test isn't testing the ramp response right now, just the feature flag (since the feature flag is set to false)

packages/mobile/src/home/CashInBottomSheet.test.tsx
```diff
@@ -28,4 +106,71 @@ describe('CashInBottomSheet', () => {
...
+      </Provider>
+    )
+    await waitFor(() => expect(getByTestId('cashInBtn')))
+  })
+  it('Regular Version is visible when ramp is restricted', async () => {
+    mockFetch.mockResponse(JSON.stringify([mockRampProviderRestricted]))
+    const { getByTestId } = render(
+      <Provider
+        store={createMockStore({
+          app: { rampCashInButtonExpEnabled: false },
```
cajubelt, line 140: same here

packages/mobile/src/home/CashInBottomSheet.test.tsx
```diff
@@ -28,4 +106,71 @@ describe('CashInBottomSheet', () => {
...
+      </Provider>
+    )
+    await waitFor(() => expect(getByTestId('cashInBtn')))
+  })
+  it('Regular Version is visible when ramp does not have cash in', async () => {
+    mockFetch.mockResponse(JSON.stringify([mockRampProviderNoCashIn]))
+    const { getByTestId } = render(
+      <Provider
+        store={createMockStore({
+          app: { rampCashInButtonExpEnabled: false },
```
cajubelt, line 153: and here

cajubelt (changes requested):
Looks good, thanks Jacob!

I think a few tests are slightly broken. Also, the screenshot in the PR says something weird about CUSD in the US, is that acceptable?

packages/mobile/src/analytics/Events.tsx
```diff
@@ -364,6 +364,7 @@ export enum FiatExchangeEvents {
   // Add fund flow entered through home screen cash in bottom sheet
   cico_add_funds_bottom_sheet_selected = 'cico_add_funds_bottom_sheet_selected',
   cico_add_funds_bottom_sheet_impression = 'cico_add_funds_bottom_sheet_impression',
+  cico_add_funds_bottom_sheet_ramp = 'cico_add_funds_bottom_sheet_ramp',
```
lisaqyli, line 367: Can we name this analytics event more explicitly? Maybe `cico_add_funds_bottom_sheet_ramp_selected`. The intention is to distinguish between an impression event and an interaction event so when DS run analysis, it's easier for them to guess what's the event is about

packages/mobile/src/home/CashInBottomSheet.tsx
```diff
@@ -54,15 +124,35 @@ function CashInBottomSheet() {
...
-        <Text style={styles.title}>{t('cashInBottomSheet.title')}</Text>
-        <Text style={styles.subtitle}>{t('cashInBottomSheet.subtitle')}</Text>
-        <Button
-          text={t('cashInBottomSheet.addFunds')}
-          size={BtnSizes.FULL}
-          onPress={goToAddFunds}
-          style={styles.addFundBtn}
-          testID={'cashInBtn'}
-        />
+        {rampAvailable && rampCashInButtonExpEnabled ? (
```
lisaqyli, line 127: just a thought! don't forget to ramp the experiment to regions where ramp is available

packages/mobile/src/home/CashInBottomSheet.test.tsx
lisaqyli: though I would imagine we would only ramp the experiment to regions where user has ramp available.

packages/mobile/src/analytics/Events.tsx
jh2oman: Yes, I can change that event. good idea

packages/mobile/src/home/CashInBottomSheet.tsx
jh2oman: Isn't it already limited by the check in the useAsync hook on `.unavailable` and `.restricted` ?

lisaqyli: It is already limited but if we dont limit the experiment to ramp only users, we will get diluted results from firebase.

jh2oman: Ah I think there is a better way to do that with activation events. I'll add an event specifically for ramp availablity so that we can use it in the experiment!
Screen Shot 2022-01-11 at 1 13 12 PM

packages/mobile/src/home/CashInBottomSheet.tsx
jh2oman: Its called a Fragment, very similar to a view https://reactjs.org/docs/fragments.html#short-syntax

packages/mobile/src/home/CashInBottomSheet.tsx
jh2oman: ikr!

packages/mobile/src/home/CashInBottomSheet.tsx
jh2oman: Yep! This async function is solely for getting the ramp availability info. If it errors out, there's no reason we can't still show the normal cash in button

jh2oman:
@cajubelt The warning in the screenshot is from their test network that we use for alfajores so I think it should be fine.

packages/mobile/src/home/CashInBottomSheet.tsx
jh2oman: https://github.com/valora-inc/wallet/pull/1736/files#diff-9536823e710c7215fefeaaa2da0274e5220c8e7629ff1aa6061442b5ba981fb4R73-R77

cajubelt: very cool, nice find @jh2oman

packages/mobile/src/home/CashInBottomSheet.tsx
cajubelt: thanks

cajubelt (approved):
lgtm, nice work!

packages/mobile/src/home/CashInBottomSheet.tsx
lisaqyli: Ooo interesting! nice find!

lisaqyli (approved):

MuckT:
@jh2oman Can we enable rampCashInButtonExpEnabled on alfajores so that QA can test this flow in the nightly release cycle?

ValoraQA:
Hey @MuckT please let know what need to be tested here so we can provide our feedback onto this ticket after checking it

PR #1739, opened 2022-01-10
Use the localAmount sent from blockchain-api

### LINKED ISSUE #1738: Use the exchange rate that comes from blockchain-api in the feed
The `blockchain-api` returns values with the exchange rate at the time of the transactions. We weren't using them correctly and using the current exchange rate instead. This PR fixes that.

gnardini:
### Description

The `blockchain-api` returns values with the exchange rate at the time of the transactions. We weren't using them correctly and using the current exchange rate instead. This PR fixes that.

### Other changes

N/A

### Tested

Manually and with unit tests

### How others should test

Look at all transactions and make sure that the exchange rate in them is the exchange rate at the time, not the current one.

### Related issues

- Fixes #1738

### Backwards compatibility

N/A


packages/mobile/src/components/TokenTotalLineItem.tsx
```diff
@@ -44,11 +49,17 @@ export default function TokenTotalLineItem({
...
+              `${
+                LocalCurrencySymbol[localAmount.currencyCode as LocalCurrencyCode]
+              }${formatValueToDisplay(new BigNumber(localAmount.exchangeRate))}`
+            ) : (
+              <TokenDisplay
+                amount={new BigNumber(1)}
+                tokenAddress={tokenAddress}
+                showLocalAmount={true}
+              />
+            )}
```
jeanregisser, line 62: We should avoid concatenating strings this way as we can't properly translate them.
Can we add a new translation key instead?

packages/mobile/src/components/TokenTotalLineItem.tsx
```diff
@@ -44,11 +49,17 @@ export default function TokenTotalLineItem({
...
             {' @ '}
-            <TokenDisplay
-              amount={new BigNumber(1)}
-              tokenAddress={tokenAddress}
-              showLocalAmount={true}
-            />
+            {localAmount?.exchangeRate ? (
+              `${
+                LocalCurrencySymbol[localAmount.currencyCode as LocalCurrencyCode]
+              }${formatValueToDisplay(new BigNumber(localAmount.exchangeRate))}`
```
jeanregisser, line 55: Why not using `CurrencyDisplay` here?
Looks like there's an opportunity to have a component that can display both Token and CurrencyCode values.

jeanregisser (commented):
Thanks! 👍 

For later: Would love to see/discuss how we could converge on a solution for displaying token and currency values. We have very similar concerns split across `CurrencyDisplay` and `TokenDisplay`.

packages/mobile/src/components/TokenTotalLineItem.tsx
gnardini: Isn't this the same as what we're doing on `TokenDisplay` and `CurrencyDisplay`? We are doing string concatenation there as well. Is there any language where showing symbol and then value is not the convention?

gnardini:
> For later: Would love to see/discuss how we could converge on a solution for displaying token and currency values. We have very similar concerns split across CurrencyDisplay and TokenDisplay.

Yeah, so when we started with all these changes trying to force `CurrencyDisplay` to work for multiple tokens seemed like would be very complex and the code there had already been through many patches to accomodate one-off use cases, many of which weren't used anymore, so I started over with `TokenDisplay`. I think eventually there should be no more uses of `CurrencyDisplay` and we could switch to `TokenDisplay` for all amounts that show a token amount.

packages/mobile/src/components/TokenTotalLineItem.tsx
gnardini: Honestly because it's too complex, it has too many things and it's hard to see which transformations will be made to the amount sent to it, in this case there's one number and we want to print the number as-is, with no transformations. This is a "localAmount" but both `CurrencyDisplay` and `TokenDisplay` have rules to multiply by exchange rates so since it was such a simple use case instead fo adding special rules to them I figured it was simpler to just print the item directly.

packages/mobile/src/components/TokenTotalLineItem.tsx
jeanregisser: There are 2 concerns.

Here I meant a key for `<Something> @ <Rate>` like we have in https://github.com/valora-inc/wallet/blob/c1e0fae65108e064ce313f56ae85cbba84b1810f/packages/mobile/src/components/TotalLineItem.tsx#L66-L78

And the 2nd concern (which doesn't need to be addressed here) is the rules for formatting monetary amounts which we are not doing correctly right now as we assume the symbol is always first and then the value.
For instance in France the convention is to show it like `1.000 €`.
And there are more exotic formats like for Cape Verde Escudo (CVE) where the decimal separator is using the $ sign: `20$50` indicates 20 Escudos and 50 cents.

packages/mobile/src/components/TokenTotalLineItem.tsx
jeanregisser: Ok I see, I just don't like too much to multiply the ways we display monetary amounts. Would prefer to have all places wrapped in a function call or component so it's easier to change in the future.
I guess here's since we're using `formatValueToDisplay` it's fine.

packages/mobile/src/components/TokenTotalLineItem.tsx
gnardini: Got it, added the translation key for now

packages/mobile/src/components/TokenTotalLineItem.tsx
gnardini: I agree there should be just one, I just really did feel that this was the solution that reduced the added complexity the most, but I would prefer to go through `TokenDisplay` in most situations

jeanregisser (approved):
👍

ValoraQA:
Hey @MuckT "Look at all transactions and make sure that the exchange rate in them is the exchange rate at the time, not the current one."
Can you please provide us from which link we can identify or get the updated exchange rate for blockchain API

PR #1741, opened 2022-01-11
chore: upgraded ethereumjs-util to 7.1.3

cajubelt:
### Description

I want to use `stripHexPrefix`, which is available in version 7, so I updated ethereumjs-util

### Other changes

fixed a place where the upgrade changed the function signature of sha256, and added test coverage for it

### Tested

just automated tests

### How others should test

n/a

### Backwards compatibility


packages/mobile/src/analytics/ValoraAnalytics.test.ts
```diff
@@ -21,6 +22,10 @@ jest.mock('react-native-permissions', () => ({}))
 jest.mock('@sentry/react-native', () => ({ init: jest.fn() }))
 jest.mock('src/redux/store', () => ({ store: { getState: jest.fn() } }))
 
+const mockDeviceId = 'abc-def-123'
+const expectedSessionId = 'b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'
+jest.spyOn(DeviceInfo, 'getUniqueId').mockReturnValue(mockDeviceId)
```
jeanregisser, line 27: Looks like we should apply this change globally in https://github.com/valora-inc/wallet/blob/main/packages/mobile/__mocks__/react-native-device-info.ts

yarn.lock
```diff
@@ -10111,6 +10098,17 @@ ethereumjs-util@^7.0.8:
...
+ethereumjs-util@^7.3.1:
+  version "7.1.3"
+  resolved "https://registry.yarnpkg.com/ethereumjs-util/-/ethereumjs-util-7.1.3.tgz#b55d7b64dde3e3e45749e4c41288238edec32d23"
+  integrity sha512-y+82tEbyASO0K0X1/SRhbJJoAlfcvq8JbrG4a5cjrOks7HS/36efU/0j2flxCPOUM++HFahk33kr/ZxyC4vNuw==
+  dependencies:
+    "@types/bn.js" "^5.1.0"
+    bn.js "^5.1.2"
+    create-hash "^1.1.2"
+    ethereum-cryptography "^0.1.3"
+    rlp "^2.2.4"
```
jeanregisser, line 10110: yarn is not very good at deduplicating multiple compatible versions. We can help it by manually deleting both versions in the lock file and running `yarn install` again.

packages/mobile/package.json
```diff
@@ -88,7 +88,7 @@
     "date-fns": "^2.12.0",
     "dot-prop-immutable": "^1.4.0",
     "es6-promisify": "^6.1.1",
-    "ethereumjs-util": "^6.2.0",
+    "ethereumjs-util": "^7.3.1",
```
jeanregisser, line 91: The last published version is [`7.1.3`](https://www.npmjs.com/package/ethereumjs-util)
```suggestion
    "ethereumjs-util": "^7.1.3",
```

jeanregisser:
Great! 
As a side note Celo SDK also has [`trimLeading0x`](https://github.com/celo-org/celo-monorepo/blob/b65eddd98873c09b7c211dbad2f871d5c4908b72/packages/sdk/base/src/address.ts#L13) util too.

packages/mobile/src/analytics/ValoraAnalytics.test.ts
```diff
@@ -21,6 +21,9 @@ jest.mock('react-native-permissions', () => ({}))
 jest.mock('@sentry/react-native', () => ({ init: jest.fn() }))
 jest.mock('src/redux/store', () => ({ store: { getState: jest.fn() } }))
 
+const mockDeviceId = 'abc-def-123' // mocked in __mocks__/react-native-device-info.ts (but importing from that file causes weird errors)
```
cajubelt, line 24: this is at least not worse than what's already happening in this test file, where we hardcode expected bundle id's, for instance, to `'org.celo.mobile.debug'`, which mirrors another mocked value from `__mocks__/react-native-device-info.ts`

cajubelt: but if I'm missing some nicer way to do this I'd still be curious to hear about it

jeanregisser: Good, that's the way to do it. 👍 
We don't want to import from the `__mocks__` folder directly.

yarn.lock
jeanregisser: No luck deduplicating this?

cajubelt: just pushed changes

jeanregisser (approved):

ValoraQA:

CC : @cajubelt

PR #1744, opened 2022-01-11
Update dependencies to use https:// instead of no longer supported git:// security protocol

kathaypacific:
### Description

Background: https://github.blog/2021-09-01-improving-git-protocol-security-github/ - the final brownout for this change was today.

The git:// security protocol is no longer supported and our CI was unable to finish fetching dependencies. Update to use https://

### Other changes

N/A

### Tested

No testing is needed, as we are just updating the protocol for fetch one dependency. The app should be unaffected.

### How others should test

As above.

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,18 +1,39 @@
+queue_rules:
+  - name: default
+    # With `speculative_check: 1` and `batch_size: 3`, if there are 3 PRs in the queue,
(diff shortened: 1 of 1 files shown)


jeanregisser (approved):
Thanks! 👍

ValoraQA:
Hey @MuckT  please let us know what need to be tested here so we can provide our feedback accordingly

PR #1751, opened 2022-01-11
(Biometry 1/4) Add enable biometry screen to onboarding flow

### LINKED ISSUE #1683: Add biometry opt-in to onboarding flow
### Description

_For context, the Valora PIN cannot be replaced by biometrics. New critical functionality relies on this PIN. Also, in the past we have seen Android devices revoke access to biometry for the app randomly._

A new screen will be added to the **end** of the onboarding flow, to allow users to opt-in to using biometry instead of their Valora pin. Designs [here](https://www.figma.com/file/j84MQn5xaOJ9RBjuIZgxPN/Onboarding-Experiments?node-id=802%3A16669)

Note that for this experiment, only iOS users will be included due to the high risk of issues with Android seen in the past.

### Acceptance Criteria
- The new onboarding screen is enabled depending on the remote config value (boolean) that the user receives. 
- The screen headers in the onboarding flow are correctly updated to reflect the accurate number of onboarding steps.
- This new onboarding screen is the last screen of the onboarding flow, so the user should be navigated to the wallet home screen regardless of if they opt in or out of biometry.
- The result of the opt-in should be saved (in redux)

kathaypacific:
### Description

In this PR:
- add remote config value for `biometricsEnabled`
- add new "enable biometrics" onboarding screen to both the create account and restore account flows. it will be the new step 3, after the pincode set screen.
- on tap of the "turn on biometrics" button, a state variable is saved to remember the selection. Note that this PR does not include updating the keychain.
- The enable biometrics screen is not displayed if the device does not support biometrics.
- The new biometrics screen looks like below and the designs are [here](https://www.figma.com/file/j84MQn5xaOJ9RBjuIZgxPN/Onboarding-Experiments?node-id=802%3A16669)

Image from iOS

NOTE:
- This feature only needs to be tested on iOS for now. Only iOS devices are in scope for the initial rollout. This feature is controlled by remote config and will only be enabled for iOS devices.

### Other changes

N/A

### Tested

Manually - this needs to be tested on a physical device. On simulators, biometrics is not supported. This biometrics onboarding screen will not be displayed if biometrics is not supported by the device.

### How others should test

On a physical device, test that the new onboarding screen is present and the onboarding screen header shows the correct step number/total number of steps for each screen.

### Related issues

- Fixes #1683 

### Backwards compatibility

Yes


packages/mobile/locales/base/translation.json
```diff
@@ -666,6 +666,17 @@
...
   },
+  "enableBiometry": {
+    "title": "Secure your wallet",
+    "description": "Face ID secures your wallet and makes it easier to use Valora",
+    "cta": {
+      "useFaceId": "Turn on Face ID",
+      "useTouchId": "Turn on Touch ID",
+      "useFingerprint": "Turn on fingerprint",
+      "useFace": "Turn on face unlock",
+      "useIris": "Turn on iris recognition"
```
kathaypacific, line 677: i've been searching for what the common phrase for the android fingerprint/face/iris but couldn't find :(

packages/mobile/src/onboarding/registration/EnableBiometry.tsx
```diff
@@ -0,0 +1,193 @@
...
+import Svg, { Path } from 'svgs'
+
+type Props = StackScreenProps<StackParamList, Screens.EnableBiometry>
+
+const biometryImageMap: { [key in Keychain.BIOMETRY_TYPE]: JSX.Element } = {
+  [Keychain.BIOMETRY_TYPE.FACE_ID]: <FaceIDImage />,
+  [Keychain.BIOMETRY_TYPE.TOUCH_ID]: <TouchIDImage />,
+  [Keychain.BIOMETRY_TYPE.FINGERPRINT]: <FingerprintImage />,
+  [Keychain.BIOMETRY_TYPE.FACE]: <FaceImage />,
+  [Keychain.BIOMETRY_TYPE.IRIS]: <FaceImage />,
```
kathaypacific, line 30: i didn't pay too much attention to these for now, since this experiement is for iOS users (fingerprint/face/iris is for android)

packages/mobile/src/app/reducers.ts
```diff
@@ -196,6 +208,7 @@ export const appReducer = (
         multiTokenUseUpdatedFeed: action.configValues.multiTokenUseUpdatedFeed,
         linkBankAccountEnabled: action.configValues.linkBankAccountEnabled,
         sentryTracesSampleRate: action.configValues.sentryTracesSampleRate,
+        biometryEnabled: action.configValues.biometryEnabled && Platform.OS === 'ios',
```
kathaypacific, line 211: i thought it would be best to add this platform check in the code, so that we can keep using this feature flag in future releases if we want to enable for android

packages/mobile/src/app/reducers.ts
```diff
@@ -46,6 +47,9 @@ export interface State {
   multiTokenUseUpdatedFeed: boolean
   linkBankAccountEnabled: boolean
   sentryTracesSampleRate: number
+  supportedBiometryType: BIOMETRY_TYPE | null
+  biometryEnabled: boolean
+  useBiometry: boolean
```
kathaypacific, line 52: i think i will remove this and instead use the PincodeType to deduce if biometry is used in the next PR

packages/mobile/src/app/reducers.ts
gnardini: I think we can do this from Remote Config as well (have different values for android and iOS) but this is fine as well I suppose :)

packages/mobile/src/app/selectors.ts
```diff
@@ -108,4 +108,20 @@ export const linkBankAccountEnabledSelector = (state: RootState) => state.app.li
...
+
+export const biometryEnabledSelector = (state: RootState) =>
+  state.app.biometryEnabled && !!state.app.supportedBiometryType
+
+export const useBiometrySelector = (state: RootState) => state.app.useBiometry
+
+export const totalRegistrationStepsSelector = createSelector(
+  [choseToRestoreAccountSelector, biometryEnabledSelector],
+  (chooseRestoreAccount, biometryEnabled) => {
+    if (biometryEnabled) {
```
gnardini, line 121: This is fine, just sharing how I think it'd be slightly nicer to read imo:
```
let steps = 3
if (chooseRestoreAccount) { 
  steps++ 
}
if (biometryEnabled) { 
  steps++ 
}
return steps
```

this is subjective though, so no need to change if you don't like it

packages/mobile/src/verify/VerificationEducationScreen.tsx
```diff
@@ -88,9 +92,18 @@ function VerificationEducationScreen({ route, navigation }: Props) {
   const currentState = useSelector(currentStateSelector)
   const shouldUseKomenci = useSelector(shouldUseKomenciSelector)
   const verificationStatus = useSelector(verificationStatusSelector)
+  const biometryEnabled = useSelector(biometryEnabledSelector)
   const choseToRestoreAccount = useSelector(choseToRestoreAccountSelector)
   const registrationStep = useRegistrationStep(route.params?.choseToRestoreAccount ? 4 : 3)
 
+  const handleProceedToNextStep = () => {
+    if (biometryEnabled) {
+      navigate(Screens.EnableBiometry)
```
gnardini, line 101: So biometry goes after phone number verification? I think if the user does verify their phone number then they never get to see the biometry screen, right? I think this could be added in `OnboardingSuccessScreen`

I think it would be better to do it before though, since phone number verification can be quite long it would make sense to not add extra steps after that, but if this has already been discussed and decided it's fine.

gnardini (commented):
Looks great! Just left a nit and a question :)

packages/mobile/src/app/reducers.ts
kathaypacific: yeah i think configuring remote config based on OS is the plan! i added this check because in a future release we might enable biometrics for everyone using this feature flag, but for an android user who hasn't upgraded and is stuck on this version of the app still, i don't want biometrics to be enabled for them when the value of this flag changes (we haven't tested Android properly)

packages/mobile/src/app/selectors.ts
kathaypacific: i like this, it does show the relationship between the steps more clearly! although i do question if this current relationship will hold true always... i know there is some work to reorganise the onboarding flow, so i think there could be the possibility that we have 2 steps for restore account and 4 steps for new account, for example?

gnardini: Hmm, in that case we'd need to make changes anyways I think? :) Not sure, I'm fine either way

packages/mobile/src/verify/VerificationEducationScreen.tsx
kathaypacific: oh Gonza, this was such a good catch. both Nitya and I forgot about this `OnboardingSuccessScreen` 🙈

i spoke with Nitya and have moved this screen to after the pincode screen. this brought an annoying problem of how to keep track of all the steps (not just the total steps, but also which step the screen is!) - i updated the useRegistrationStep hook, it's very verbose and i'm not sure i like it but my intention is for screens not to need to know too much about the number. really wanted to avoid the phone verification/import wallet screens needing to know about biometrics.....let me know what you think?

i'd hoped to get rid of `totalRegistrationStepsSelector` actually, but `PincodeSet` and `VerificationInputScreen` are still classes and can't use the `useRegistrationStep` hook 😭

packages/mobile/src/app/selectors.ts
kathaypacific: yeah you're right. i've updated :)

packages/mobile/src/onboarding/registration/EnableBiometry.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { default as useSelector } from 'src/redux/useSelector'
+
+type Props = StackScreenProps<StackParamList, Screens.EnableBiometry>
+
+const biometryImageMap: { [key in Keychain.BIOMETRY_TYPE]: JSX.Element } = {
+  [Keychain.BIOMETRY_TYPE.FACE_ID]: <FaceID />,
+  [Keychain.BIOMETRY_TYPE.TOUCH_ID]: <TouchID />,
+  [Keychain.BIOMETRY_TYPE.FINGERPRINT]: <Fingerprint />,
+  [Keychain.BIOMETRY_TYPE.FACE]: <Face />,
+  [Keychain.BIOMETRY_TYPE.IRIS]: <Face />,
```
jeanregisser, line 35: Doesn't matter now, but should we have an iris icon?

jeanregisser (commented):
This is great! 👍 

Could you add screenshots so it's easier for reviewers/testers to understand what has changed visually?

packages/mobile/locales/base/translation.json
```diff
@@ -666,6 +666,17 @@
     "step2": "With {{appName}}, it’s easy to access and easy to grow your crypto.",
     "step3": "All you need is a mobile phone to send, spend and explore."
   },
+  "enableBiometry": {
+    "title": "Secure your wallet",
+    "description": "Face ID secures your wallet and makes it easier to use Valora",
```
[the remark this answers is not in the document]
kathaypacific: ah you're right, i forgot 🙈 i've updated this key and also modified how we handle the other keys (i opted to translate each biometrics option separately to avoid having multiple variations for each biometrics option on every key we need. unfortunately the biometrics options we get from the keychain lib is not exactly what we need)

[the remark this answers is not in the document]
kathaypacific: updated!

packages/mobile/test/schemas.ts
```diff
@@ -865,6 +865,9 @@ export const v26Schema = {
   },
   app: {
     ...v25Schema.app,
+    biometryEnabled: false,
+    supportedBiometryType: null,
+    useBiometry: false,
```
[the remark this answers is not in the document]
kathaypacific: ha, good spot! so outdated now...i've created the v29 schema now :)

packages/mobile/src/onboarding/registration/EnableBiometry.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const onPressUseBiometry = async () => {
+    dispatch(setUseBiometry(true))
+    // do some stuff to use biometry
+
+    navigate(choseToRestoreAccount ? Screens.ImportWallet : Screens.VerificationEducationScreen)
+  }
+
+  return (
+    <SafeAreaView style={styles.container}>
+      <ScrollView contentContainerStyle={styles.contentContainer}>
```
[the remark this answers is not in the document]
kathaypacific: oh i didn't think of this! i have updated to have the elements in this order now

packages/mobile/src/onboarding/registration/EnableBiometry.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+      activeScreen: Screens.EnableBiometry,
+    },
+    account: {
+      choseToRestoreAccount: false,
+    },
+  })
+
+  return render(
+    <Provider store={store}>
+      <NavigationContainer>
```
kathaypacific, line 27: I wanted to add a test for the skip button in the screen header but i cannot get it to render even with this here - any ideas? when i debug what is rendered by the `render` it starts from the ScrollView of the screen, not with navigation wrapper....

packages/mobile/src/app/selectors.ts
```diff
@@ -108,6 +113,89 @@ export const linkBankAccountEnabledSelector = (state: RootState) => state.app.li
...
+export const registrationStepsSelector = createSelector(
+  [
+    choseToRestoreAccountSelector,
+    biometryEnabledSelector,
+    activeScreenSelector,
+    recoveringFromStoreWipeSelector,
+  ],
+  (chooseRestoreAccount, biometryEnabled, activeScreen, recoveringFromStoreWipe) => {
+    let step = 0
+    let totalSteps = 3
```
gnardini, line 176: Shouldn't the base `totalSteps` be `4` instead if the user is restoring an account?

packages/mobile/src/app/selectors.ts
```diff
@@ -108,6 +113,89 @@ export const linkBankAccountEnabledSelector = (state: RootState) => state.app.li
...
+      step = restoreAccountSteps[activeScreen as RestoreAccountScreens]
+    } else {
+      step = createAccountSteps[activeScreen as CreateAccountScreens]
+    }
+
+    // biometry screen inserted as third screen
+    if (biometryEnabled) {
+      totalSteps++
+      if (activeScreen === Screens.EnableBiometry) {
+        step = 3
```
gnardini, line 193: Is Biometry also available when restoring? If so I think this won't work well :/

packages/mobile/src/app/selectors.ts
```diff
@@ -108,6 +113,89 @@ export const linkBankAccountEnabledSelector = (state: RootState) => state.app.li
...
+  [Screens.NameAndPicture]: 1,
+  [Screens.PincodeSet]: 2,
+  [Screens.ImportWallet]: 3,
+  [Screens.VerificationEducationScreen]: 4,
+  [Screens.VerificationInputScreen]: 4,
+}
+
+// The logic in this selector should be moved to a hook when all registration
+// screens are function components
+export const registrationStepsSelector = createSelector(
```
gnardini, line 167: This selector is a bit tricky with all the different cases!

I'm not sure what the best way to implement this is, but here's a suggestion, feel free to ignore it. It's more boilerplate-y and less generic but I feel like easier to read and understand.

The idea is to just re-write the code 3 times for each of the 3 flows. I don't remember exactly but I think this is what the original version looked like? In that case I'm sorry for making your life difficult :( So something like:

```
if (recoveringFromStoreWipe) {
  return {
    step: storeWipeRecoverySteps[activeScreen]
    totalSteps: 3,
  }
} else if (chooseRestoreAccount) {
  const totalSteps = biometryEnabled ? 5 : 4
  let step = restoreAccountSteps[activeScreen] // Note that I'd add the BiometricScreen to the restoreAccountSteps object with the number 4.
  if (biometricsEnabled && isVerificationScreen(activeScreen)) {
    step = 5
  }
  return { totalSteps, step }
} else {
  const totalSteps = biometryEnabled ? 4 : 3
  let step = createAccountSteps[activeScreen] // Note that I'd add the BiometricScreen to the createAccountSteps object with the number 3.
  if (biometricsEnabled && isVerificationScreen(activeScreen)) {
    step = 4
  }
  return { totalSteps, step }
}
```

I don't think this is the best option either to be honest, not sure how to do it in a way that's tidy and scalable into the future, but I think this is good for now? To me particularly having the separate `if` for each flow helps me read that particular block thinking just about that one flow which is easier than juggling the three flows in a more generic fashion, but maybe it's just me.

In the existing code, the part that made me think too hard and I don't fully understand is the `if (biometryEnabled)`, since we're handling two flows there I think the indexes might be slightly wrong, but maybe I'm wrong.


In any case, if the current code works feel free to leave as it is as I'm not convinced that my version is too much better either :/

gnardini (commented):
I think I might have made things more confusing with my comment in the first review, sorry about that if so :(

Another disadvantage about how this works right now is that the individual screens are the ones deciding where to navigate to next and we have that same logic duplicated here.

One option I think might be good in the future is to create a "Navigation Plan" when starting the flow that would build the sequence of screens to go through at the very start (so basically an array of screens) so each individual screen would know where it is in the flow (ie step 3 of 5) and which one to go to next without doing any ifs or logic. Maybe it would have other issues though, not really sure.

packages/mobile/src/onboarding/registration/EnableBiometry.test.tsx
jeanregisser: This is because the screen here is not being rendered via `react-navigation`.
We would need to create a stack for testing with the route.

Something like:
```tsx
<NavigationContainer>
      <Stack.Navigator>
        <Stack.Screen
          name={Screens.EnableBiometry}
          component={ EnableBiometry }
          initialParams={ params }
        />
      </Stack.Navigator>
    </NavigationContainer>
```

But there might be some additional problem with this.

jeanregisser (approved):
yeah! 👍

packages/mobile/src/app/selectors.ts
kathaypacific: yes biometry is also available when restoring. it's also step 3 there because steps 1 and 2 (name and picture/pincode set) are the same for both the restore and create flows

packages/mobile/src/app/selectors.ts
kathaypacific: i agree! in the original i had the blocks separated, i can go back to that. i do think it's clearer like what you have above with each flow separated, especially after adding the recover from store wipe flow. separate blocks makes it looks less magic that the numbers add up in the end (i did write some unit tests for the flows, so the current version does work but readability is key so i will update!)

packages/mobile/src/app/selectors.ts
kathaypacific: yes if the user is restoring account i have a `totalSteps++` in the `if (chooseRestoreAccount)` block...but i will make this less confusing by making the separate blocks

packages/mobile/src/onboarding/registration/EnableBiometry.tsx
kathaypacific: update: Nitya couldn't find an iris icon. let's address this when we plan the rollout for android, she suggested asking Tinka/Bryan for an icon but it wouldn't be the standard android icon

packages/mobile/src/app/selectors.ts
kathaypacific: hopefully the update is more clear? i took your suggestion to add the biometrics screen to the objects. i think let's not spend too much time on it because it sounds like there is work in the pipeline to remove the verification steps from the onboarding altogether, so this logic would become redundant in the near future :)

packages/mobile/src/onboarding/registration/EnableBiometry.test.tsx
kathaypacific: great idea!! we actually have a mocked navigator (unused) which i could refurbish and use here, works like a charm :D am glad that i could cover the skip button, since it's unlikely this can be covered by e2e

gnardini (approved):
Sorry about the confusion with the steps thing, great job on this Kathy!

ValoraQA:
Hey @MuckT, We are not able access figma screen [Url](https://www.figma.com/file/j84MQn5xaOJ9RBjuIZgxPN/Onboarding-Experiments?node-id=802%3A16669) mentioned above. 
Please do needful.
Thanks..!!

ValoraQA:
Hey @MuckT, We have verified the above task in Latest Test flight Alfa jores build V1.26.0 and updated the status in test rails.
**Note:** 

**Test rail links:**
**iOS 13:** https://valoraapp.testrail.io/index.php?/tests/view/9997

**iOS 14:** https://valoraapp.testrail.io/index.php?/tests/view/9742

**iOS 15:** https://valoraapp.testrail.io/index.php?/tests/view/9679

CC: @kathaypacific 
Thanks..!!

PR #1756, opened 2022-01-11
refactor(auth): use JWTs for IHL auth

cajubelt:
### Description

We've updated IHL auth to use JWTs instead of a homegrown token schema, for maintainability. This update switches IHL queries over to the new auth scheme.

### Other changes

also fixes how we access the DEK private keys for a user

### Tested

automated tests
doing integration tests with IHL next...

### How others should test

n/a

### Related issues

fixes https://app.zenhub.com/workspaces/acquisition-squad-sprint-board-6010683afabec1001a090887/issues/valora-inc/in-house-liquidity/446

### Backwards compatibility

n/a (feature not released yet on mainnet)


packages/mobile/src/in-house-liquidity/index.ts
```diff
@@ -74,23 +76,10 @@ export const createPersonaAccount = async ({
...
-  message,
-  accountMTWAddress,
-}: {
-  message: string
-  accountMTWAddress: string
-}): Promise<string> => {
-  const contractKit = await getContractKitAsync()
-  const accountWrapper = await contractKit.contracts.getAccounts()
-  const dataEncryptionKey = await accountWrapper.getDataEncryptionKey(accountMTWAddress)
-  return signWithDEK({ message, dataEncryptionKey })
```
cajubelt, line 87: This actually didn't work-- the "dek" returned by the smart contract was the *public* key of the DEK pair, so signing with the public key was not useful. That's why the new DEK access pattern looks different.

packages/mobile/src/web3/dataEncryptionKey.ts
```diff
@@ -50,25 +50,9 @@ import {
...
 
-// Based on https://github.com/celo-org/celo-monorepo/blob/f4dd249e9cef6a541b909c3dc7ae01888b088de2/packages/sdk/identity/src/odis/query.ts#L111
-export function signWithDEK({
-  message,
-  dataEncryptionKey,
-}: {
-  message: string
-  dataEncryptionKey: string
-}) {
-  const key = ec.keyFromPrivate(hexToBuffer(dataEncryptionKey))
```
cajubelt, line 68: this was where we assumed the DEK provided was a private key (it was not)

cajubelt: I tried to use "private" and "public" in all the variable names for DEK's to help prevent this kind of thing from happening again. It's very confusing to name both elements of a key pair with the same name and just rely on context clues to figure out which one it is. Also pretty dangerous..

packages/mobile/src/account/PlaidLinkButton.tsx
```diff
@@ -34,9 +39,13 @@ const PlaidLinkButton = ({ disabled }: { disabled: boolean }) => {
       Logger.error(TAG, "Can't render Plaid because walletAddress is null")
       return
     }
+    if (!dekPrivate) {
+      Logger.error(TAG, "Can't render Plaid because DEK is null")
```
cajubelt, line 43:
```suggestion
      Logger.error(TAG, "Can't render Plaid because dekPrivate is null")
```

packages/mobile/src/account/Persona.tsx
```diff
@@ -84,9 +84,14 @@ const Persona = ({ kycStatus, text, onCancelled, onPress }: Props) => {
         return
       }
 
+      if (!dekPrivate) {
+        Logger.error(TAG, 'Cannot render Persona because DEK is null')
```
cajubelt, line 88:
```suggestion
        Logger.error(TAG, 'Cannot render Persona because dekPrivate is null')
```

packages/mobile/src/in-house-liquidity/index.ts
```diff
@@ -99,73 +88,48 @@ interface SignAndFetchParams {
...
-  return {
-    Date: dateString,
-    Authorization: authorization,
-  }
+  dekPrivate,
+}: {
+  accountMTWAddress: string
+  dekPrivate: string
+}): Promise<string> => {
+  const keyEncoder = new KeyEncoder('secp256k1')
```
jh2oman, line 125: Does this need to be done on every request or can it instead be done once at the top of the file?

cajubelt: Nothing in the docs about this. I looked at the code for encodePublic and encodePrivate and neither appears to mutate the keyEncoder. (Even if they did, the private/public DEKs for the user should not change.) So I think it's fine to have it at the top.

ValoraQA:

CC : @cajubelt

PR #1766, opened 2022-01-12
Allow sending with comment on any core token

### LINKED ISSUE #1767: Allow sending comments for cUSD/cEUR/cREAL
### What would you like?
The new multi token flow is not allowing to add a comment with cUSD/cEUR/cREAL tokens. We should allow this!

gnardini:
### Description

Sending with comment was broken because we were comparing uppercase letters with lowercase ones. Sending with comment is currently not working on the 1.24 release. This fixes it while also adding cREAL as possible token to send with comment.

### Other changes

Added cREAL in a few places, but it's not used, not really useful..

### Tested

Manually sending core tokens and not core tokens.

### How others should test

- Send cUSD/cEUR/cREAL and see that you can send a comment with it.
- Send CELO or any other token and check that you can't send a comment with it.

### Related issues

- Fixes #1767
### Backwards compatibility

N/A


packages/mobile/src/send/saga.ts
```diff
@@ -223,19 +225,22 @@ export function* buildSendTx(
   comment: string
 ) {
   const contract: Contract = yield call(getERC20TokenContract, tokenAddress)
-  const coreContract: CeloTokenWrapper<any> | undefined = yield call(
-    getTokenContractFromAddress,
-    tokenAddress
-  )
+  const coreContract: Contract = yield call(getStableTokenContract, tokenAddress)
```
gnardini, line 228: Using the contract via the abi so new stable tokens work out of the box in the future, with no need to update the code

packages/mobile/src/exchange/saga.ts
```diff
@@ -110,6 +110,8 @@ export async function getExchangeContract(token: StableCurrency) {
       return contractKit.contracts.getExchange(StableToken.cUSD)
     case Currency.Euro:
       return contractKit.contracts.getExchange(StableToken.cEUR)
+    case Currency.Real:
+      return contractKit.contracts.getExchange(StableToken.cREAL)
```
gnardini, line 114: This isn't used, not sure why I added it tbh, but shouldn't hurt?

PR #1784, opened 2022-01-14
Dapp explorer

### LINKED ISSUE #1506: Add 'DeFi Explorer' screen to the app
We want to add a new screen to the app from which users can see the existing dApps distributed by categories and interact with them. See the epic for more info on the future.

Here are the mockups: https://www.figma.com/file/zdK8XvxNXvTjxnM5V33pHw/dAPP-Directory?node-id=831%3A19659

Some notes and considerations:
- There must be an item in the side menu that opens this screen. Don't worry about the in-app notification since that's added dynamically from the admin dashboard.
- The information about the dApps and categories must be fetched from the endpoint that will be created on #1507. Check that issue for the fields that will be returned.
- The copy and icons doesn't look like it's finalized, so check with design/product before implementing, especially the  side menu icon, the info modal and the confirmation bottom sheet.
- For the bottom sheet, there's an example on the component `TokenBottomSheet` but it's not very extendable. You have three options: 
  - Create a new component and copy the necessary stuff. It's not a lot of code anyways.
  - Create a generic `BottomSheet` component that is shared by both.
  - Look for and use an existing library for the bottom sheet.

AC: 
- Users can open the 'DeFi Explorer' screen from the side menu and open the dApps that are available.
- The UI matches the mockups.
- This feature is unit tested.
- This feature should be initially only available behind a feature flag.

dievazqu:
### Description

DApps Screen.
Designs: https://www.figma.com/file/zdK8XvxNXvTjxnM5V33pHw/dAPP-Directory?node-id=831%3A19659

### Tested

Manually in the Android and iOS emulator

### How others should test

Turn on the feature flag `dappsExplorerEnabled`, then you should be able to interact with the new screen.

### Known Issue
In the new Dapps screen, if the user presses on Bidali (which it's not visible yet) and then goes back, it doesn't return to the Dapps screen.

### Related issues

- Fixes #1506

Screenshots (this is a older version, some wording has changed):
Screen Shot 2022-01-14 at 13 30 09
Screen Shot 2022-01-14 at 13 30 35
Screen Shot 2022-01-14 at 13 31 58
Screen Shot 2022-01-14 at 13 32 16


packages/mobile/src/components/BottomSheet.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+interface Props {
+  isVisible: boolean
+  onBackgroundPress?: () => void
+  children?: JSX.Element
+}
+
+const MIN_EMPTY_SPACE = 100
+
+function BottomSheet({ children, isVisible, onBackgroundPress }: Props) {
```
dievazqu, line 19: Mainly I copy pasted this class from `TokenBottomSheet`.
I will create a next PR with a refactor to start using only this class instead of having both.

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,342 @@
...
+          </View>
+          <View style={styles.centerContainer}>
+            <Text style={{ ...fontStyles.h2, textAlign: 'center', paddingVertical: 16 }}>
+              {t('dappsScreenBottomSheet.title', { dappName: dappSelected?.name })}
+            </Text>
+            <Text style={{ ...fontStyles.regular, textAlign: 'center', flex: 1 }}>
+              {t('dappsScreenBottomSheet.message')}
+            </Text>
+
+            {/* <Text>Dont show again Message</Text> */}
```
dievazqu, line 160: There is a discussion regarding if we want this feature or not: https://valora-app.slack.com/archives/C02TH9GNERL/p1642102112018200

packages/mobile/locales/en-US/translation.json
```diff
@@ -496,6 +496,7 @@
   "helpFindAccount": "Help your contact find their Account Address",
   "close": "Close",
   "home": "Home",
+  "dappsExplorer": "dAPP Explorer",
```
gnardini, line 499: I think this text should be in the `base` directory as well?

packages/mobile/src/app/reducers.ts
```diff
@@ -85,6 +86,7 @@ const initialState = {
   multiTokenUseUpdatedFeed: REMOTE_CONFIG_VALUES_DEFAULTS.multiTokenUseUpdatedFeed,
   linkBankAccountEnabled: REMOTE_CONFIG_VALUES_DEFAULTS.linkBankAccountEnabled,
   sentryTracesSampleRate: REMOTE_CONFIG_VALUES_DEFAULTS.sentryTracesSampleRate,
+  dappsExplorerEnabled: REMOTE_CONFIG_VALUES_DEFAULTS.dappsExplorerEnabled,
```
gnardini, line 89: nit: Can we put this variable next to `dappListApiUri `? Since both are related to the same thing :)

gnardini: I think we also need to add `dappsExplorerEnabled` to the reducer for `Actions.UPDATE_REMOTE_CONFIG_VALUES` as well

packages/mobile/src/icons/DappsExplorerLogo.tsx
```diff
@@ -0,0 +1,78 @@
...
+        fill="#2E3338"
+      />
+    </Svg>
+  )
+}
+
+/*
+<Svg width="94" height="106" viewBox="0 0 94 106" fill="none" xmlns="http://www.w3.org/2000/svg">
```
gnardini, line 54: What is this commented stuff? Can we delete it?

packages/mobile/src/navigator/DrawerNavigator.tsx
```diff
@@ -179,6 +181,8 @@ function CustomDrawerContent(props: DrawerContentComponentProps<DrawerContentOpt
 export default function DrawerNavigator() {
   const { t } = useTranslation()
   const isCeloEducationComplete = useSelector((state) => state.goldToken.educationCompleted)
+  const isDappsExplorerEnabled = useSelector((state) => state.app.dappsExplorerEnabled)
```
gnardini, line 184: To avoid having to handle the error in the other screen, can we check as well if the URI is available? And if it's not not show the item since we know it will fail :)

packages/mobile/test/schemas.ts
```diff
@@ -866,6 +866,7 @@ export const v26Schema = {
   app: {
     ...v25Schema.app,
     rampCashInButtonExpEnabled: false,
+    dappsExplorerEnabled: false,
```
gnardini, line 869: I think that when we add new stuff we need to create a new schema, ie`v27Schema` with the new feature flag.

I think it will also require making a redux store migration.

packages/mobile/src/components/BottomSheet.tsx
```diff
@@ -0,0 +1,141 @@
...
+    position: 'absolute',
+    bottom: 0,
+    opacity: 1,
+    width: '100%',
+    backgroundColor: colors.light,
+    padding: Spacing.Thick24,
+    borderTopRightRadius: Spacing.Regular16,
+    borderTopLeftRadius: Spacing.Regular16,
+  },
+  title: {
```
gnardini, line 102: I like the new generic bottom sheet :) We should just remove the unused styles here

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,349 @@
...
+          result.applications.forEach((app: any) => {
+            categoriesById[app.categoryId].items.push({
+              id: app.id,
+              name: app.name,
+              iconUrl: app.logoUrl,
+              description: app.description,
+              dappUrl: app.url,
+            })
+          })
+          setcategoriesWithItems(Object.values(categoriesById))
```
gnardini, line 103: typo: `setcategoriesWithItems` -> `setCategoriesWithItems`

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
gnardini: Lets delete this commented out stuff in the meantime, if we decide to add it we'll need to add extra logic anyways :)

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,349 @@
...
+        <Text style={{ ...styles.categoryText, color: categoryWithItems.fontColor }}>
+          {categoryWithItems.name}
+        </Text>
+      </View>
+      <>{categoryWithItems.items.map((item) => renderItem(item, onItemPress))}</>
+    </View>
+  )
+}
+
+function renderItem(item: DappItem, onItemPress: (dapp: DappItem) => void) {
```
gnardini, line 228: I think in general instead of items I'd call them `dApps` or something, in the categories object above as well, I think it's more descriptive

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,349 @@
...
+
+  return (
+    <Card style={styles.card} rounded={true} shadow={Shadow.Soft} key={`item-${item.id}`}>
+      <TouchableOpacity style={styles.pressableCard} onPress={onPress}>
+        <Image source={{ uri: item.iconUrl }} style={styles.dappIcon} />
+        <View style={styles.itemTextContainer}>
+          <Text style={styles.itemTitleText}>{item.name}</Text>
+          <Text style={styles.itemSubtitleText}>{item.description}</Text>
+        </View>
+        <Image source={require('src/images/link-arrow.png')} style={styles.linkArrow} />
```
gnardini, line 243: I think for more icons we add the png in `src/images/Images.ts` and import it wherever it's used instead of requiring them inline like this.

Not sure why we do that, and I realize that there's one more place that does it like this, but I think it makes sense to copy the most common pattern?

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,349 @@
...
+  if (!dappsListUrl) {
+    return (
+      <View style={styles.centerContainer} testID="DAppExplorerScreen/error">
+        <Text style={fontStyles.large}> Configuration error. </Text>
+        <Text style={fontStyles.large}> dappsListUrl is not set. </Text>
+      </View>
+    )
+  }
+
+  const { loading } = useAsync(
```
gnardini, line 68: One thing that's potentially missing is an error. If this fails there will be no items and we won't show the loading indicator anymore but no content either. We can ask Nitya but maybe a simple 'There was an error loading the dApps' could be enough? I'd check with her in any case.

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
gnardini, line 103: So this has the potential to cause an error (not sure if leak memory or what) because this is an async function that keeps a reference to the `setcategoriesWithItems` callback. If the user goes back or something this `setcategoriesWithItems` callback may not be defined anymore or not sure exactly what can happen. I think we can just return the `Object.values(categoriesById)` and it will be available in the `result` field that `useAsync` returns.

Maybe @jeanregisser can correct me if that's not the real problem or add more context 😅

gnardini (commented):
Amazing work! Can't wait to see all this in action :)

packages/mobile/locales/en-US/translation.json
dievazqu: My bad, I think we can remove this one :)

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
dievazqu: It makes sense :)

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
dievazqu: Yep, you are right, I just copy pasted but it makes sense to continue with the pattern

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
dievazqu: Oh, nice catch! I will change it, so it returns the value instead of calling `setCategoryiesWithItems`

packages/mobile/src/navigator/DrawerNavigator.tsx
dievazqu: Yep, I'll do that

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,349 @@
...
+export function DAppsExplorerScreen() {
+  const { t } = useTranslation()
+  const dappsListUrl = useSelector(dappsListApiUrlSelector)
+  const [categoriesWithItems, setcategoriesWithItems] = useState<CategoryWithItems[]>([])
+  const [isHelpDialogVisible, setHelpDialogVisible] = useState(false)
+  const [isBottomSheetVisible, setBottomSheetVisible] = useState(false)
+  const [dappSelected, setDappSelected] = useState<DappItem>()
+
+  if (!dappsListUrl) {
+    return (
```
kathaypacific, line 60: i think this guy should be moved further down the component, just before the last `return` because if react will complain about mismatched number hooks between re-renders if the code does return early here :) (alternatively `useAsync` could be moved before this return, but it's easier to accidentally add another hook after this return later on by accident) (https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level)

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,349 @@
...
+          )}
+          {!loading &&
+            categoriesWithItems.map((category) => renderCategoryWithItems(category, onItemPress))}
+        </>
+      </ScrollView>
+    </SafeAreaView>
+  )
+}
+
+function renderCategoryWithItems(
```
kathaypacific, line 205: was wondering if we have any preferences for patterns here...this way of course is fine and works :) but it is equivalent to doing `function CategoryWithItems(props) {...}` and calling it in line 198 like 
```
categoriesWithItems.map(
  <CategoryWithItems
    {...someProps}
  />
)}
```
this function is producing a react component so i personally prefer invoking it like a component but not sure what the consensus is, so thought i'd ask! we do use the `renderXXX` function pattern if the `renderXXX` is inside a component/class already which makes sense, but this one is outside and standalone.

kathaypacific (commented):
Nice work!!!! 🔥

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
dievazqu: Yes! When I was doing it I thought that using react component would be better, I'll change it :)

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,362 @@
...
+            <View style={{ ...styles.centerContainer, marginTop: 96 }}>
+              <Text style={fontStyles.regular}>{t('dappsScreen.errorMessage')}</Text>
+            </View>
+          )}
+          {!loading &&
+            !error &&
+            result &&
+            result.map((category: CategoryWithDapps) => (
+              <CategoryWithDapps category={category} onDappPress={onDappPress} />
+            ))}
```
jeanregisser, line 215: This is ok when we have few Dapps / Categories, but will become a performance issue if we have more than 10 or so of them. At that point we should favor a `FlatList`.

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,362 @@
...
+  centerContainer: {
+    alignItems: 'center',
+    justifyContent: 'center',
+    flex: 1,
+  },
+  categoryContainer: {
+    alignItems: 'flex-start',
+    justifyContent: 'center',
+    flex: 1,
+    marginVertical: 16,
```
jeanregisser, line 284: We haven't used this everywhere sofar, but we have standard spacing constants. For instance: https://github.com/valora-inc/wallet/blob/c1e0fae65108e064ce313f56ae85cbba84b1810f/packages/mobile/src/verify/VerificationLoadingScreen.tsx#L290

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,362 @@
...
+    ...fontStyles.regular,
+    fontSize: 12.5,
+    borderRadius: 100,
+    paddingHorizontal: 11,
+    paddingVertical: 4,
+  },
+  itemTitleText: {
+    ...fontStyles.small,
+    fontWeight: '500',
+    color: '#2E3338',
```
jeanregisser, line 332: Can't we use the colors from our design system?

https://github.com/valora-inc/wallet/blob/6bf13e1bb5a3cfda63eebf0477f829430085e598/packages/react-components/styles/colors.tsx#L1-L36

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,362 @@
...
+  categoryText: {
+    ...fontStyles.regular,
+    fontSize: 12.5,
+    borderRadius: 100,
+    paddingHorizontal: 11,
+    paddingVertical: 4,
+  },
+  itemTitleText: {
+    ...fontStyles.small,
+    fontWeight: '500',
```
jeanregisser, line 331: I forgot the exact details now, but there are differences between iOS and Android related to `fontWeight` and our custom font, so this might not render the way we expect.
We should stick to the styles defined in our design system to avoid this.

https://github.com/valora-inc/wallet/blob/66cdc5a6bd34cd0957e968d599e223cccaf6496e/packages/react-components/styles/fonts.tsx#L38-L115

packages/mobile/src/app/selectors.ts
```diff
@@ -107,3 +107,5 @@ export const multiTokenUseUpdatedFeedSelector = (state: RootState) =>
 export const linkBankAccountEnabledSelector = (state: RootState) => state.app.linkBankAccountEnabled
 
 export const sentryTracesSampleRateSelector = (state: RootState) => state.app.sentryTracesSampleRate
+
+export const dappsListApiUrlSelector = (state: RootState) => state.app.dappListApiUri
```
jeanregisser, line 111: nit: So should we use `uri` or `url`? 😄 
Looks like we've already mixed both, seems `url` should be preferred in my opinion.

jeanregisser (commented):
Looking great!! 🤩 
Left mostly some small nits/comments.

packages/mobile/src/components/Dialog.tsx
```diff
@@ -17,8 +17,8 @@ interface Props {
   image?: ImageSourcePropType
   title?: string | React.ReactNode
   children: React.ReactNode
-  actionText: string
-  actionPress: () => void
+  actionText?: string
+  actionPress?: () => void
```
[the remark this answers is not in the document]
dievazqu: Well, I wasn't sure about this.
The primary action has some style that is different than the secondary action (green color text instead of gray).
In this case what we want is a Dialog that can be dismissed, and I think this will be a normal use case. That's why I allowed this behavior, but if you don't agree, what we can do is pass some style for the actionText.
However, I think it's great to have the possibility to use the Dialog in this case. Then it would be developer's responsibility to see how it's the best way to use the Dialog.

packages/mobile/src/icons/LinkArrow.tsx
```diff
@@ -1,8 +1,8 @@
 import * as React from 'react'
 import { Image } from 'react-native'
 
-const LinkArrow = () => {
-  return <Image source={require('src/images/link-arrow.png')} style={{ width: 32, height: 32 }} />
+const LinkArrow = ({ style = { width: 32, height: 32 } }) => {
+  return <Image source={require('src/images/link-arrow.png')} style={style} />
```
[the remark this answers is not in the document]
dievazqu: I like this, thanks!

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
dievazqu: Yes, we can! Thanks!

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
dievazqu: I added that to honor the design parameters. But I just removed the `fontWeight` property and it's showing the same, so we can remove it.

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
dievazqu: It makes sense, I will change it

jeanregisser: Or ˋSectionList`

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
dievazqu: You are right, already changed :)

dievazqu:
@gnardini @kathaypacific @jeanregisser 
Thank you for your reviews!
I think I have addressed all the comments, and the PR is ready to be re-review again. 😃

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,389 @@
...
+            name: app.name,
+            iconUrl: app.logoUrl,
+            description: app.description,
+            dappUrl: app.url,
+          })
+        })
+
+        return Object.values(categoriesById)
+      } catch (error) {
+        Logger.error(TAG, 'onError', error as Error)
```
gnardini, line 122: Maybe add some small detail to the message? Like `Error fetching dApp list` instead of `onError`

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,389 @@
...
+              />
+            </View>
+          )}
+          {!loading && error && (
+            <View style={{ ...styles.centerContainer, marginTop: 96 }}>
+              <Text style={fontStyles.regular}>{t('dappsScreen.errorMessage')}</Text>
+            </View>
+          )}
+          {!loading && !error && result && (
+            <SectionList
```
gnardini, line 227: I'm not sure about having the `SectionList` inside the `ScrollView` since they are both scrollable, I think the containing `ScrollView` should be just a View? I'm not sure how the interaction works though, mentioning it just in case.

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,389 @@
...
+
+function parseResultIntoSections(categoriesWithDapps: CategoryWithDapps[]): SectionData[] {
+  return categoriesWithDapps.map((category) => ({
+    data: category.dapps,
+    category: category,
+  }))
+}
+
+function CategoryHeader({ category }: { category: CategoryWithDapps }) {
+  Logger.debug(TAG, `Render category ${JSON.stringify(category)}`)
```
gnardini, line 255: Is this log helpful? It looks like it would be a bit spammy since it's logged on every re-render, so it would keep logging the same thing over and over.

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,389 @@
...
+        }}
+      >
+        <Text style={{ ...styles.categoryText, color: category.fontColor }}>{category.name}</Text>
+      </View>
+    </View>
+  )
+}
+
+function Dapp({ dapp, onPressDapp }: DappProps) {
+  Logger.debug(TAG, `Render item ${JSON.stringify(dapp)}`)
```
gnardini, line 272: Same related to the frequency of logging

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,389 @@
...
+
+  const onPress = () => {
```
gnardini, line 274: super nit: I think this would look slightly nicer as `const onPress = () => onPressDapp(dapp)`

gnardini (commented):
Left a couple final nits, looks great :)

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
jeanregisser: Exactly @gnardini 👍

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,381 @@
...
+        Logger.debug(TAG, `fetch onSuccess: ${JSON.stringify(result)}`)
+      },
+      onError: (error) => {
+        Logger.error(TAG, 'fetch onError', error as Error)
+      },
+    }
+  )
+
+  const onPressDapp = (dapp: Dapp) => {
+    setDappSelected(dapp)
```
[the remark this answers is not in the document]
dievazqu: Hmm, you are right. Let me see what I can do. Nice catch!

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
dievazqu: I'll try removing the ScrollView and see how that works :)

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
dievazqu: I found it helpful while developing. But maybe we can delete them now.
It's true that it's logged on every re-render, however, there shouldn't be any of them, except the first one (after fetching the json file). So I also don't find it spammy.
On my opinion I will let them there. But happy to delete them if you still think we should.

dievazqu: I removed them 😋

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,386 @@
...
+  },
+  descriptionImage: {
+    height: 106,
+    width: 94,
+    marginLeft: Spacing.Smallest8,
+  },
+  bottomSheetHeader: {
+    flex: 1,
+    flexDirection: 'row-reverse',
+    alignItems: 'flex-start',
```
kathaypacific, line 375: super nit: i think you can have `alignItems: 'flex-end'` and remove the flexDirection to achieve the same effect?

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,386 @@
...
+function CategoryHeader({ category }: { category: CategoryWithDapps }) {
+  return (
+    <View style={styles.categoryContainer}>
+      <View
+        style={{
+          ...styles.categoryTextContainer,
+          backgroundColor: category.backgroundColor,
+        }}
+      >
+        <Text style={{ ...styles.categoryText, color: category.fontColor }}>{category.name}</Text>
```
kathaypacific, line 271: i notice that there are a lot of inline styles in this component, do we want to keep them like this or move to a stylesheet? i think having styles in the stylesheet has some performance benefits (not recreating the style every rerender) and slightly easier to read but i see there are other places in the app where we use inline style so i'm not sure! just thought i'd ask, what do you think?

packages/mobile/src/redux/migrations.ts
```diff
@@ -362,4 +362,11 @@ export const migrations = {
       rampCashInButtonExpEnabled: false,
     },
   }),
+  27: (state: any) => ({
+    ...state,
+    app: {
+      ...state.app,
+      dappsExplorerEnabled: false,
```
kathaypacific, line 369: i was expecting to see `dappListApiUrl` here, maybe nice to add it for correctness? (i also notice `dappListApiUrl` was added to the v7 schema in the test schemas 🙈)

kathaypacific (approved):
This is awesome Diego!! Looks like you've been doing RN for years 🔥 I just left a few super small comments but honestly, looks great to me 🚀

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
jeanregisser: We're trying to avoid inlining whenever possible. But I guess here it would look like this, which is really equivalent:
```suggestion
        <Text style={[styles.categoryText, { color: category.fontColor }]}>{category.name}</Text>
```

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
kathaypacific: actually you could even just add `style={{ alignSelf: 'flex-end' }}` to the `TopBarIconButton` and remove the need for the surrounding `View` too

packages/mobile/src/components/Dialog.tsx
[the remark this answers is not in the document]
jeanregisser: ah I see, then maybe we could have add a `highlightPrimaryAction?: boolean`, which would default to `true`.
This way the highlighted style on the primary action would be configurable and we would ensure there's an action handler passed.
What do you think?

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
gnardini: Oh, sorry about that, it's true that it's not re-rendered that often, so I would have been fine if you preferred to keep them, but I guess it's fine either way :)

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,386 @@
...
+  }
+
+  const onPressHelp = () => {
+    setHelpDialogVisible(true)
+  }
+
+  const onPressNavigationButton = () => {
+    if (!dappSelected) {
+      // Should never happen
+      Logger.warn(TAG, 'Internal error. There was no dapp selected')
```
gnardini, line 152: Should this be `Logger.error` instead? I'm sorry, just now saw it :(

gnardini (approved):
I think the conflicts are my fault, sorry about that :( Looks amazing, can't wait to have this in-app!

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
dievazqu: I think you are right, let me try it. Thanks!

packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
dievazqu: Sure

packages/mobile/src/redux/migrations.ts
dievazqu: Oh, it makes sense, I'll fix it.

packages/mobile/src/components/Dialog.tsx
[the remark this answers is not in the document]
dievazqu: Hmm... I'm not sure. What will be the expected behavior if `highlightPrimaryAction = false` and we also pass a secondaryAction?
I think it's ok to have the possibility to create a Dialog without any action. Why should we force it to have one?

I'm gonna try to merge this, but we can continue this discussion offline if you want (or we can keep using this thread), and I will create a follow up PR if needed, that sounds ok?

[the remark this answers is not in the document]
jeanregisser: Yes that's fine.

ValoraQA:
Hey @MuckT we have verified the above task on Latest iOS Test Flight Release Build V:1.25.1 and updated the status in test rails

**Test rail link** :
**iOS 13:** https://valoraapp.testrail.io/index.php?/tests/view/4401&group_by=cases:section_id&group_order=asc&group_id=30

[Bug link
](https://github.com/valora-inc/wallet/issues/1818)
**Devices** : iPhone 7 (13.3)
cc : @dievazqu 
Thanks..!

ValoraQA:
Hey @MuckT, we have verified the above task on Latest iOS Test Flight Alfajores Build V:1.25.1 & updated the status in test rails

**Test rail link :**
**iOS 15:**https://valoraapp.testrail.io/index.php?/tests/view/4464

**Observation :** We are able to view Dapp section in Hamburger menu and we are also able to access the page.

**Bug:** [Link](https://github.com/valora-inc/wallet/issues/1818)
**Devices :** iPhone 13 mini (15.1.1)
Cc : @dievazqu
Thanks..!!

ValoraQA:
Hey @MuckT, we have verified the above task on Latest Android Alfajores Build V:1.25.2 & updated the status in test rails

**Test rail link:**
**Android 10:** https://valoraapp.testrail.io/index.php?/tests/view/7082&group_by=cases:section_id&group_order=asc&group_id=30

**Observation**: We are able to view Dapp section in Hamburger menu and we are also able to access the page.

[Bug link
](https://github.com/valora-inc/wallet/issues/1818)

**Devices** : Samsung Galaxy M40(10.0)

**Cc:** @dievazqu 
Thanks..!!

ValoraQA:
Hey @MuckT, we have verified the above task on Latest Android Alfajores Build V:1.25.2 & updated the status in test rails


Android 8: 

https://valoraapp.testrail.io/index.php?/tests/view/7145&group_by=cases:section_id&group_order=asc&group_id=30

**Observation:** We are able to view Dapp section in Hamburger menu and we are also able to access the page.



Devices : Samsung Galaxy J5 Prime(8.0.0)

Cc: @dievazqu
Thanks..!!

ValoraQA:
Hey @MuckT, We have verified the above task on Latest Android Alfajores Build V:1.25.2 & updated the status in test rails. User is able to view Dapp section in Hamburger menu and we are also able to access the page.

**Test rail link:**
**Android 9:**  https://valoraapp.testrail.io/index.php?/tests/view/7469

**Bug:** [Link](https://github.com/valora-inc/wallet/issues/1818)
**Devices :** Redmi Note 5 Pro (9.0)
Cc: @dievazqu
Thanks..!!

PR #1791, opened 2022-01-17
Send error logs to Sentry

### LINKED ISSUE #1459: Improve usage of Sentry logging to capture error events
ENG:
* Configure Sentry logging to send `.warn` and `.error` occurrences
* Check the expected amount of events sent to sentry (and check with Silas if we need to increase our quota)

AC:
* Consistent use of `log.error` to indicate a user is prevented from using the application in an intended way due to an unexpected error that is **not transient**
* Consistent use of `log.error to indicate a user is prevented from using the application in an intended way due to a **transient** error that is **not handled gracefully** (e.g., silently dropping error and not responding to user tap)
* Consistent use of `log.warn` to indicate a user is prevented from using the application in an intended way due to a **transient** error that is **handled gracefully** (e.g., "please try again")
* An engineer is able to see occurrences (a "report") of unhandled errors and log.error messages
* Ensure expected events sent to Sentry are within our quota

gnardini:
### Description

The `Logger.error` calls now send the errors to Sentry. Also went through all the calls to `error` and changed a few for `warn`. In many cases the error was being logged twice (captured, logged then re-thrown and re-captured -> re-logged).

### Other changes

- Added a breadcrumb when navigating to have some extra info in error reports. Might be nice to add more breadcrumbs at some point.
- In the dappkit screen the phone number was required and threw an error if not set. There was no real requirement for this so I removed that check.
- Moved the `ReactNativeLogger` to the `mobile` package (it was in `react-components`).

### Tested

I added some `Logger.error`s to test and checked that they showed up on Sentry.

### How others should test

No need to test this specifically I think?

### Related issues

- Fixes #1459

### Backwards compatibility

N/A


packages/mobile/src/services/ReactNativeLogger.ts
```diff
@@ -33,6 +34,7 @@ export default class ReactNativeLogger {
     const sanitizedError =
       error && shouldSanitizeError ? this.sanitizeError(error, valueToPurge) : error
     const errorMsg = this.getErrorMessage(sanitizedError)
+    Sentry.captureException(error, { extra: { tag, message, errorMsg, source: 'Logger.error' } })
```
kathaypacific, line 37: i was wondering if we want to prevent errors from being sent to Sentry in dev mode? we can filter by envs in Sentry so we don't have to look at them but if we are paying by volume of captured events then probably not worth sending these?

packages/mobile/src/navigator/NavigatorWrapper.tsx
```diff
@@ -150,6 +151,11 @@ export const NavigatorWrapper = () => {
         currentScreen: currentRouteName,
       })
       dispatch(activeScreenChanged(currentRouteName))
+      Sentry.addBreadcrumb({
+        category: 'navigation',
+        message: `Navigated to ${currentRouteName}`,
+        level: Sentry.Severity.Info,
+      })
```
kathaypacific, line 158: was wondering if these breadcrumbs provide different info to the `sentryRoutingInstrumentation` which tracks the navigation already? i tried looking at the error events/navigation events of the last few days but i couldn't see any extra breadcrumbs in Sentry. if you could point me to an example i'm keen to see what the breadcrumbs look like!

kathaypacific:
Looks great!!! 🔥

packages/mobile/src/escrow/EscrowedPaymentListItem.tsx
```diff
@@ -38,7 +38,7 @@ function EscrowedPaymentListItem({ payment }: Props) {
         }),
       })
     } catch (error) {
-      Logger.error(TAG, `Error sending reminder to ${recipient.e164PhoneNumber}`, error)
+      Logger.warn(TAG, `Error sending reminder to ${recipient.e164PhoneNumber}`, error)
```
jeanregisser, line 41: This is interesting to capture as well.

packages/mobile/src/dappkit/DappKitAccountScreen.tsx
```diff
@@ -64,10 +64,6 @@ class DappKitAccountAuthScreen extends React.Component<Props> {
       Logger.error(TAG, 'No account set up for this wallet')
       return
     }
-    if (!phoneNumber) {
-      Logger.error(TAG, 'No phone number set up for this wallet')
-      return
-    }
```
jeanregisser, line 70: What's the impact for Dapps using DAppKit?
Thinking mainly about ImpactMarket and Ubeswap.

packages/mobile/src/navigator/NavigationService.ts
```diff
@@ -132,7 +132,7 @@ export async function ensurePincode(): Promise<boolean> {
     await requestPincodeInput(true, false)
     return true
   } catch (error) {
-    Logger.error(`${TAG}@ensurePincode`, `PIN entering error`, error, true)
+    Logger.warn(`${TAG}@ensurePincode`, `PIN entering error or cancelled`, error, true)
```
jeanregisser, line 135: I think we should capture the error unless the user actually cancelled.

packages/mobile/src/utils/contacts.ts
```diff
@@ -51,11 +51,13 @@ export async function getAllContacts(): Promise<MinimalContact[] | null> {
       if (error) {
         Logger.error(TAG, 'Error getting all contacts', error)
         reject(error)
+        return
       }
 
       if (!contacts) {
         Logger.error(TAG, 'Contacts is null')
         reject('Contacts is null')
+        return
```
jeanregisser, line 60: Nice catch! 👍

packages/mobile/src/utils/useClipboard.ts
```diff
@@ -33,7 +33,7 @@ export function useClipboard(): [boolean, string, () => Promise<string>] {
 
         setClipboardContent(newClipboardContent)
       } catch (error) {
-        Logger.error('Error checking clipboard contents', error)
+        Logger.warn('Error checking clipboard contents', error)
```
jeanregisser, line 36: What do you think of capturing this too instead?

jeanregisser (commented):
Great work!! 💯 👍 

I've proposed to keep some logs as errors. Main reason is to be sure we're not missing important things. Then we can tweak and ignore some of them in Sentry.
But let me know, I might have made some incorrect assumptions on some of them.

About:
> In the dappkit screen the phone number was required and threw an error if not set. There was no real requirement for this so I removed that check.
Thanks for fixing it! I think we have an issue opened but can't find it now.
We should tests existing Dapps using DAppkit to be sure this change is not gonna break them.

packages/mobile/src/services/ReactNativeLogger.ts
gnardini: I think they aren't sent in dev mode because Sentry isn't initialized, so it no-ops. See `src/sentry/Sentry.ts#16`

packages/mobile/src/navigator/NavigatorWrapper.tsx
kathaypacific: 🤔 the traces sample rate is 0.2 i think, so is it possible that the first event was sampled and the second event wasn't? i see this event from 1.24.1 that has a navigation breadcrumb already https://sentry.io/organizations/valora-inc/performance/celo-mobile:055880cf850442238b44cd855d127305/?environment=co.clabs.valora&project=1250733&query=transaction.duration%3A%3C15m+event.type%3Atransaction&statsPeriod=14d&transaction=WalletHome&unselectedSeries=p100%28%29 which looks pretty similar to the breadcrumb provided by this snippet...

kathaypacific: having said that, i'm not sure if the react navigation tracing leaves breadcrumbs on error events. that event i linked is from the performances trace, so may be different. i'm having a hard time finding the navigation breadcrumb on the recent errors, so not really sure what's going on. in any case, i am happy to keep this block here!

packages/mobile/src/services/ReactNativeLogger.ts
kathaypacific: oh of course 🙈

packages/mobile/src/navigator/NavigatorWrapper.tsx
gnardini: I'll try a couple more times to see how consistent it is. I would expect the tracing limit to affect performance stuff but not errors! If there's an error I want to see the breadcrumbs all the time, not just 20% of the time :P I'll test and report back

packages/mobile/src/dappkit/DappKitAccountScreen.tsx
gnardini: I tested on Ubeswap and it worked well! 

For ImpactMarket do you mean the web or the app? Not sure how to test with the app but I can give the web a try

packages/mobile/src/escrow/EscrowedPaymentListItem.tsx
gnardini: Hmm, the phone number is considered PII, right? I just removed it from the log, don't think it adds a ton of value having it anyways.

packages/mobile/src/dappkit/DappKitAccountScreen.tsx
gnardini: Tested with ImpactMarket and connected successfully as well :)

packages/mobile/src/navigator/NavigatorWrapper.tsx
gnardini: Yeah, I tested 4 more times, the two times that I added the breadcrumb both were there: https://sentry.io/organizations/valora-inc/issues/2940380905/?project=1250733&query=is%3Aunresolved and https://sentry.io/organizations/valora-inc/issues/2940386168/?project=1250733&query=is%3Aunresolved and the two times I didn't no navigation breadcrumb showed up: https://sentry.io/organizations/valora-inc/issues/2940386748/?project=1250733&query=is%3Aunresolved and https://sentry.io/organizations/valora-inc/issues/2940387891/?project=1250733&query=is%3Aunresolved

packages/mobile/src/escrow/EscrowedPaymentListItem.tsx
jeanregisser: Yes, I meant the error itself, we can redact the phone number.

kathaypacific (approved):
🚀 looks great!! 🚢

ValoraQA:

PR #1810, opened 2022-01-19
Remove unused e2e use case

jeanregisser:
### Description

Stumbled upon this unused e2e use case which is already covered in `NewAccountOnboarding.js`.

@MuckT let me know if you had another intention with it.

### How others should test

N/A


ValoraQA:

CC : @jeanregisser

PR #1811, opened 2022-01-19
(Biometry 2/4) Use biometrics instead of pincode enter screen when enabled

### LINKED ISSUE #1684: Use biometry to access PIN
### Description
_For context, the Valora PIN cannot be replaced by biometrics. New critical functionality relies on this PIN. Also, in the past we have seen Android devices revoke access to biometry for the app randomly. Note that for this experiment, only iOS users will be included due to the high risk of issues with Android seen in the past._

If biometry is enabled, try to use it to retrieve the PIN when the PIN is required. e.g. when making a new transaction.

To discuss: do we want to require the PIN sometimes, to ensure users do not forget? Are there certain functionalities that we want to use PIN but not biometry (e.g. change PIN)? (Team agreed, use biometry everywhere)

### Acceptance Criteria
- If biometry is enabled:
    - the secret password hash we store on the Keychain should be retrievable by biometry
    - when a user tries to perform an action that we require the pin for (such as send flow):
        - instead of showing the "enter PIN" screen, use biometry instead
        - display native animation for biometry check loading/success/fail
        - allow 2 failed attempts to validate biometrics
        - if biometric validation fails, then display the "enter PIN" screen as usual (**not** the device pin)
- If biometry is disabled, no existing behaviour should change.

kathaypacific:
### Description

- Create new keychain item that stores pincode protected by biometrics. I opted for using an async function rather than a saga (this was how biometrics was originally implemented) because I think it is nice for the consumer of the function to decide how to handle the outcome. On the EnableBiometry screen, this means staying on the screen while loading/error, and navigating further on success.
- For flows that require pincode, if pincodeType is PhoneAuth then try to fetch pincode via keychain and biometrics. Fallback to pincode enter screen.

### Other changes

Remove unused `isSettingPin` from redux state, remove unused setPincode action

### Tested

Manually & unit tests.

### How others should test

Test flows that require pin with biometrics (face id, touch id). Technically you can test this on a simulator but it doesn't actually match your face/fingerprint so it is better to test on a physical device.

Also test with face/fingerprint obscured to see fallback flow.

Flows like sending tokens, seeing recovery phrase, reset pin should all work with biometrics instead of requiring pin.

Please note that this feature is only available for iOS currently.

### Related issues

- Fixes #1684 

### Backwards compatibility

Yes


packages/mobile/src/onboarding/registration/EnableBiometry.tsx
```diff
@@ -72,10 +77,13 @@ export default function EnableBiometry({ navigation, route }: Props) {
...
-    dispatch(setUseBiometry(true))
-    // do some stuff to use biometry
-
-    navigate(choseToRestoreAccount ? Screens.ImportWallet : Screens.VerificationEducationScreen)
+    try {
+      await setPincodeWithBiometrics()
+      dispatch(setPincodeSuccess(PincodeType.PhoneAuth))
+      navigate(choseToRestoreAccount ? Screens.ImportWallet : Screens.VerificationEducationScreen)
+    } catch (error) {
+      Logger.warn(TAG, 'Error enabling biometrics', error)
```
kathaypacific, line 85: I've chosen `warn` over `error` for most of the logging because the user action of cancelling the biometric authentication (after 2 failed attempts, if the user is wearing a mask or something) would trigger this error. there isn't a list of error codes provided by the keychain lib to exclude this error - there is only a hard coded string error displayed and it seems OS specific

packages/mobile/src/pincode/authentication.ts
```diff
@@ -167,6 +170,18 @@ function storePasswordHash(hash: string, account: string) {
   return storeItem({ key: passwordHashStorageKey(account), value: hash })
 }
 
+function storePinWithBiometrics(pin: string) {
+  return storeItem({
+    key: STORAGE_KEYS.PIN,
+    value: pin,
```
kathaypacific, line 176: should we encrypt this value somehow? 🤔

jeanregisser (commented):
Amazing work! 🔒 🔑 

What happens when the user restores their iPhone account to a new device?
Answering myself: nothing special, they need to restore their Valora account. So we're good on that front 😄 

Can we cover all paths with tests in the new code added in `authentication.ts` so codecov doesn't show any warning about this?
This is a critical part of the app and we should aim for maximum coverage to avoid surprises.
Here it's mostly about testing some more error cases from what I see.

packages/mobile/ios/celo/Info.plist
```diff
@@ -74,6 +74,8 @@
 	<string>Connecting your camera allows you to scans codes for payments.</string>
 	<key>NSContactsUsageDescription</key>
 	<string>Adding your contacts makes it easy to send and request payments with your friends.</string>
+	<key>NSFaceIDUsageDescription</key>
+	<string>This is required for you to use Face ID to secure your account.</string>
```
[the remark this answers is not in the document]
kathaypacific: ah good one, i totally forgot! i have done now :)

packages/mobile/src/navigator/NavigationService.ts
```diff
@@ -123,11 +127,20 @@ export async function ensurePincode(): Promise<boolean> {
...
     Logger.error(TAG + '@ensurePincode', `Unsupported Pincode Type ${pincodeType}`)
     return false
   }
 
+  if (pincodeType === PincodeType.PhoneAuth) {
+    try {
+      await getPincodeWithBiometrics()
+      return true
+    } catch (error) {
+      Logger.warn(`${TAG}@ensurePincode`, `Retrieve PIN by biometrics error`, error)
```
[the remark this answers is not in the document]
kathaypacific: okay, i've updated to Logger.error and we can always silence it in Sentry/update it later if it gets noisy.

the chain of calls to fetch the pin with biometrics is `getPincodeWithBiometrics` -> `retrieveStoredItem` -> `Keychain.getGenericPassword`. for the user cancelled action in case of failed biometrics auth,`Keychain.getGenericPassword` will reject with a user cancelled error and this is already logged with `Logger.error` in `retrieveStoredItem`. The new error that is thrown from `retrieveStoredItem` is a generic `ErrorMessages.KEYCHAIN_STORAGE_ERROR` so in this function above, we would be sending the generic `ErrorMessages.KEYCHAIN_STORAGE_ERROR` to Sentry on top of the actual keychain error sent by `retrieveStoredItem`.....my thinking is that most errors that come from this event will be user cancelled errors but let's see in Sentry :)

packages/mobile/src/pincode/PincodeSet.tsx
```diff
@@ -71,6 +72,7 @@ function mapStateToProps(state: RootState): StateProps {
 const mapDispatchToProps = {
   setPincode,
```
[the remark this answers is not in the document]
kathaypacific: good catch! not anymore, i've removed this action and the `isSettingPin` state that wasn't used as well :D

packages/mobile/src/pincode/authentication.ts
```diff
@@ -252,6 +267,43 @@ export function* getPasswordSaga(account: string, withVerification?: boolean, st
 
 type PinCallback = (pin: string) => void
 
+export async function setPincodeWithBiometrics() {
+  let pin = getCachedPin(DEFAULT_CACHE_ACCOUNT)
+  if (!pin) {
+    pin = await requestPincodeInput(true, true)
+  }
```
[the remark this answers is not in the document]
kathaypacific: i'm not sure what you mean here? i really wanted `setPincodeWithBiometrics` to be simple to call, because this will be called from the biometry screen + the settings screen. i didn't want those screens to know about how to fetch the pin. there is a `storePinWithBiometrics` function a few lines down that does explicitly require the pin as a param, is this what you mean?

packages/mobile/src/pincode/authentication.test.ts
```diff
@@ -65,8 +75,23 @@ describe(getPincode, () => {
...
+      password: mockPin,
+      username: 'username',
+      service: 'service',
+      storage: 'storage',
+    })
+    mockedKeychain.getGenericPassword.mockImplementationOnce(getPasswordSpy)
+    await getPincode()
+
+    expect(getPasswordSpy).toHaveBeenCalledTimes(1)
+    expect(getPasswordSpy).toHaveBeenCalledWith({ service: 'PIN' })
```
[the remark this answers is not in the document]
kathaypacific: 🙈 because i didn't know you could do that 🙈 i've updated, i'm happy to have learned this!

kathaypacific:
> Amazing work! 🔒 🔑
> 
> What happens when the user restores their iPhone account to a new device? Answering myself: nothing special, they need to restore their Valora account. So we're good on that front 😄
> 
> Can we cover all paths with tests in the new code added in `authentication.ts` so codecov doesn't show any warning about this? This is a critical part of the app and we should aim for maximum coverage to avoid surprises. Here it's mostly about testing some more error cases from what I see.

Actually really happy that you suggested to do more testing, in the process i found that the `getPasswordSaga` had a global mock so we were testing a mock 🥲

packages/mobile/src/pincode/authentication.test.ts
```diff
@@ -82,6 +163,174 @@ describe(getPincode, () => {
...
+
+    expect(getCachedPin(DEFAULT_CACHE_ACCOUNT)).toEqual(mockPin)
+    expect(mockedKeychain.setGenericPassword).toHaveBeenCalledWith(
+      'CELO',
+      expectedPasswordHash,
+      expect.objectContaining({ service: expectedAccountHash })
+    )
+    expect(mockedKeychain.setGenericPassword).toHaveBeenCalledWith(
+      'CELO',
+      expect.any(String), // TODO test that this can be decrypted correctly
```
kathaypacific, line 300: @jeanregisser any suggestions on how to test that the correct encrypted mnemonic is stored? the encryption doesn't resolve to a single value so i couldn't hard code this, and we're mocking the keychain so I couldn't retrieve it to test it 🥲

packages/mobile/src/pincode/authentication.ts
```diff
@@ -252,6 +267,43 @@ export function* getPasswordSaga(account: string, withVerification?: boolean, st
...
+    // safely here
+    await storePinWithBiometrics(pin)
+
+    const retrievedPin = await retrieveStoredItem(STORAGE_KEYS.PIN)
+
+    if (retrievedPin !== pin) {
+      throw new Error('Retrieved incorrect pin with biometrics after saving')
+    }
+  } catch (error) {
+    Logger.error(TAG, 'Failed to save pin with biometrics', error)
```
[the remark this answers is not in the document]
gnardini: Yes!

packages/mobile/src/pincode/authentication.ts
```diff
@@ -319,8 +381,12 @@ export async function updatePin(account: string, oldPin: string, newPin: string)
     if (updated) {
       clearPasswordCaches()
       setCachedPin(DEFAULT_CACHE_ACCOUNT, newPin)
-      const hash = await getPasswordHash(newPassword)
+      const hash = getPasswordHash(newPassword)
       await storePasswordHash(hash, account)
+      const pincodeType = pincodeTypeSelector(store.getState())
+      if (pincodeType === PincodeType.PhoneAuth) {
```
gnardini, line 387: I'm _a bit_ worried about what happens if an intermediate steps fail, but I'm not sure if the worry makes sense or not 🤔  We have a blanket error catch, but say we save the PIN this way but then updating the mnemonic fails. I guess that's where the asking for PIN fallback comes in 😄 

Yeah, writing this I think I convinced myself that it's fine, but sending the message anyways just in case

gnardini (commented):
Amazing job! Can't wait to see this live 😄

packages/mobile/src/navigator/NavigationService.ts
```diff
@@ -123,11 +127,20 @@ export async function ensurePincode(): Promise<boolean> {
...
     Logger.error(TAG + '@ensurePincode', `Unsupported Pincode Type ${pincodeType}`)
     return false
   }
 
+  if (pincodeType === PincodeType.PhoneAuth) {
+    try {
+      await getPincodeWithBiometry()
+      return true
+    } catch (error) {
+      Logger.error(`${TAG}@ensurePincode`, `Retrieve PIN by biometry error`, error)
```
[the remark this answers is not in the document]
kathaypacific: i've added a comment explaining! we need the flow to continue since we do not provide a fallback for the biometric auth - if face/touch id fails, then we need to continue in the function to activate the pincode input screen :)

packages/mobile/src/pincode/authentication.ts
```diff
@@ -252,6 +267,43 @@ export function* getPasswordSaga(account: string, withVerification?: boolean, st
...
+    // safely here
+    await storePinWithBiometry(pin)
+
+    const retrievedPin = await retrieveStoredItem(STORAGE_KEYS.PIN)
+
+    if (retrievedPin !== pin) {
+      throw new Error('Retrieved incorrect pin with biometry after saving')
+    }
+  } catch (error) {
+    Logger.error(TAG, 'Failed to save pin with biometry', error)
```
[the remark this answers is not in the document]
kathaypacific: i've updated :D

packages/mobile/src/pincode/authentication.ts
```diff
@@ -260,6 +312,16 @@ export async function getPincode(withVerification = true) {
     return cachedPin
   }
 
+  const pincodeType = pincodeTypeSelector(store.getState())
+  if (pincodeType === PincodeType.PhoneAuth) {
+    try {
+      const retrievedPin = await getPincodeWithBiometry()
+      return retrievedPin
+    } catch (error) {
+      Logger.error(TAG, 'Failed to retrieve pin with biometry', error)
```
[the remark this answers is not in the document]
kathaypacific: ditto, i've added the comment here as well explaining the pin input fallback :)

packages/mobile/src/pincode/authentication.ts
kathaypacific: yeah indeed, i've tried to make the phone auth as "optional" as possible - basically just try and if it fails then the user just gets the flow they currently have

gnardini (approved):
Amazing job!

packages/mobile/src/pincode/authentication.ts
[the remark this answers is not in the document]
jeanregisser: Ah yes I see, fine as is then 👍

packages/mobile/src/navigator/NavigationService.ts
```diff
@@ -134,10 +134,12 @@ export async function ensurePincode(): Promise<boolean> {
 
   if (pincodeType === PincodeType.PhoneAuth) {
     try {
-      await getPincodeWithBiometrics()
+      await getPincodeWithBiometry()
       return true
     } catch (error) {
-      Logger.warn(`${TAG}@ensurePincode`, `Retrieve PIN by biometrics error`, error)
+      Logger.error(`${TAG}@ensurePincode`, `Retrieve PIN by biometry error`, error)
```
jeanregisser, line 140: I saw Sentry captured an error about user cancelling this: https://sentry.io/organizations/valora-inc/issues/2956635589/?environment=org.celo.mobile.alfajores.dev&project=1250733&query=canceled&statsPeriod=14d
We should check that here and just warn for that case.

packages/mobile/src/pincode/authentication.test.ts
```diff
@@ -174,13 +254,92 @@ describe(setPincodeWithBiometrics, () => {
...
+    })
+  })
+
+  const oldPin = '123123'
+  // expectedPasswordHash generated from mockPin
+  const expectedPasswordHash = '9853810edb88b031bf6ac1505f5689cb423876fbeb14f7a3037c97ec4531b6ae'
+  // expectedAccountHash generated from normalizeAddress(mockAccount)
+  const expectedAccountHash = 'PASSWORD_HASH-0000000000000000000000000000000000007e57'
+  // mockEncryptedMneumonic generated using mockMneumonic, mockAccount, old pin
+  const mockEncryptedMneumonic =
```
jeanregisser, line 290:
```suggestion
  // mockEncryptedMnemonic generated using mockMnemonic, mockAccount, old pin
  const mockEncryptedMnemonic =
```

packages/mobile/src/pincode/authentication.test.ts
jeanregisser: Since it's a unit test, it's fine if it's not entirely real.
But can we have a deterministic string here?

No need to do it here, but testing this e2e would ensure everything is really working as expected.

jeanregisser (commented):
Great job covering this + existing things with more tests! 👍 

Left a couple of small comments.

packages/mobile/src/pincode/authentication.test.ts
kathaypacific: haha good catch. i've updated!

packages/mobile/src/navigator/NavigationService.ts
kathaypacific: yeah this was the issue i mentioned above, where the keychain lib throws an error for the user cancelled event and we were catching it in `retrieveStoredItem` and throwing a generic `throw new Error(ErrorMessages.KEYCHAIN_STORAGE_ERROR)`

i've attempted to exclude the user cancelled error from the genuine errors the same way as we exclude network errors, would be nice to see if you agree with the handling :)

kathaypacific:
@jeanregisser to your comment above 

> Since it's a unit test, it's fine if it's not entirely real.
> But can we have a deterministic string here?
> 
> No need to do it here, but testing this e2e would ensure everything is really working as expected.

I was having issues making the encryption deterministic before, but ended up mocking the crypto-js lib and checking that it was called correctly :)

packages/mobile/src/navigator/NavigationService.ts
```diff
@@ -123,11 +128,24 @@ export async function ensurePincode(): Promise<boolean> {
...
   }
 
+  if (pincodeType === PincodeType.PhoneAuth) {
+    try {
+      await getPincodeWithBiometry()
+      return true
+    } catch (error) {
+      if (!isUserCancelledError(error)) {
+        Logger.error(`${TAG}@ensurePincode`, `Retrieve PIN by biometry error`, error)
+      }
```
jeanregisser, line 143: Great!

packages/mobile/src/pincode/authentication.test.ts
```diff
@@ -7,45 +8,92 @@ import { navigate, navigateBack } from 'src/navigator/NavigationService'
...
+    expectPincodeEntered()
+  })
+  it('should throw an error for unset pincode type', async () => {
+    try {
+      await expectSaga(getPasswordSaga, mockAccount, false, false)
+        .provide([[select(pincodeTypeSelector), PincodeType.Unset]])
+        .run()
+    } catch (error) {
+      expect(error).toEqual(Error('Pin has never been set'))
+    }
```
jeanregisser, line 79: This will happily pass if no error is actually thrown.
Related to our discussion in https://github.com/valora-inc/wallet/pull/1816#discussion_r793482703

packages/mobile/src/pincode/authentication.test.ts
```diff
@@ -7,45 +8,92 @@ import { navigate, navigateBack } from 'src/navigator/NavigationService'
...
+  })
+  it('should throw an error for unexpected pincode type', async () => {
+    try {
+      await expectSaga(getPasswordSaga, mockAccount, false, false)
+        .provide([[select(pincodeTypeSelector), 'unexpectedPinType']])
+        .run()
+    } catch (error) {
+      expect(error).toEqual(Error('Unsupported Pincode Type unexpectedPinType'))
+    }
```
jeanregisser, line 88: ditto

packages/mobile/src/pincode/authentication.test.ts
```diff
@@ -82,6 +180,186 @@ describe(getPincode, () => {
...
+  })
+
+  it('throws an error if a null pin was retrieved', async () => {
+    mockedKeychain.getGenericPassword.mockResolvedValue(false)
+
+    try {
+      await getPincodeWithBiometry()
+    } catch (error) {
+      expect(error).toEqual(expect.any(Error))
+    }
```
jeanregisser, line 205: ditto

packages/mobile/src/pincode/authentication.test.ts
```diff
@@ -82,6 +180,186 @@ describe(getPincode, () => {
...
+      username: 'username',
+      service: 'PIN',
+      storage: 'storage',
+    })
+
+    try {
+      await setPincodeWithBiometry()
+    } catch (error) {
+      expect(error).toEqual(expect.any(Error))
+    }
```
jeanregisser, line 277: ditto

packages/mobile/src/storage/keychain.tsx
```diff
@@ -36,6 +46,9 @@ export async function retrieveStoredItem(key: string) {
     }
     return item.password
   } catch (error) {
+    if (isUserCancelledError(error)) {
+      throw error
+    }
```
jeanregisser, line 51: Maybe we should add a comment here why we're still re-throwing when cancelled?

packages/mobile/src/pincode/authentication.test.ts
kathaypacific: ✅ i've updated all these already in https://github.com/valora-inc/wallet/pull/1816

packages/mobile/src/storage/keychain.tsx
kathaypacific: done! i hope it's clear from the comment..but i think it is still necessary/easier for the consumer of this function to be aware that fetching from the keychain failed (to know whether to log an error, to default to the pin input screen etc) - i toyed with the idea of returning null instead of throwing an error in this case, but then consumers need to know that null value means either the user cancelled or the keychain returned a falsey value :/ this seemed like the worse option

kathaypacific: i thought about this in the weekend, i don't think it should return an error....i will change this

kathaypacific:
@jeanregisser ready for another pass:
- i removed the keychain store/retrieve/remove functions to throw the actual error instead of ErrorMessages.KeychainError. i think this is less confusing than throwing a generic error + a real error for the user cancelled case....and ErrorMessages.KeychainError is not currently being used so i am confident i haven't modified any existing behaviours.
- I didn't remove KeychainError from ErrorMessages, i think for a future PR we should evaluate where the keychain errors are unhandled and display a generic keychain error to the user. (e.g. in the pincode screens where we use `checkPin` we should catch the error)
- since the Keychain functions are already using Logger.error (for non user cancelled errors), i demoted the new biometrics logging that use the keychain functions to `warn` to avoid sending the same error to sentry twice
- i moved the test assertions changes in #1816 to here to make it less confusing

jeanregisser (approved):
🎉

kathaypacific:
@mergifyio refresh

ValoraQA:
**Note:**

**Test rail links:**

**iOS 13:** https://valoraapp.testrail.io/index.php?/tests/view/9999

**iOS 14:** https://valoraapp.testrail.io/index.php?/tests/view/9744

**iOS 15:** https://valoraapp.testrail.io/index.php?/tests/view/9681

**CC:** @kathaypacific 

Thanks..!!

PR #1816, opened 2022-01-20
(Biometry 3/4) Update settings screen items layout and add item for biometrics

### LINKED ISSUE #1685: Add biometry opt-in/opt-out in Settings
### Description
_For context, the Valora PIN cannot be replaced by biometrics. New critical functionality relies on this PIN. Also, in the past we have seen Android devices revoke access to biometry for the app randomly. Note that for this experiment, only iOS users will be included due to the high risk of issues with Android seen in the past._

If a user has opted in to use biometry in the onboarding flow, we need to give them the option to opt out for whatever reason.

### Acceptance Criteria
- In the Settings screen, create item to switch biometry on/off. This involves a new settings item and slight shuffle of the existing settings screen items, designs [here](https://www.figma.com/file/j84MQn5xaOJ9RBjuIZgxPN/Onboarding-Experiments?node-id=849%3A10011)
- This biometry settings item should exist based on the remote config value received for the user
- Turning biometry off: the secret password hash stored on the Keychain should no longer be accessible with biometry. (Verify: is any visual feedback necessary here?)
- Turning biometry on: validate on the spot that the secret password hash stored on the Keychain is accessible with biometry (the user should see the native biometry verification animation). Asking for pin confirmation before enabling biometry may be a technical requirement.
- Errors encountered during switch on/off (to do with keychain access, not biometric authentication) should be logged to Sentry and an error banner displayed (?)

kathaypacific:
### Description

- Break up the "security and data" section of the settings screen into separate "security" and "data" sections. Designs https://www.figma.com/file/j84MQn5xaOJ9RBjuIZgxPN/Onboarding-Experiments?node-id=849%3A10011
- Add a "Use xxx" toggle to enable supported biometry option

Also included in this branch are https://github.com/valora-inc/wallet/pull/1832 and https://github.com/valora-inc/wallet/pull/1823

### Other changes

Also included in this branch are https://github.com/valora-inc/wallet/pull/1832 and https://github.com/valora-inc/wallet/pull/1823

### Tested

Manually

### How others should test

- Verify that the new item for toggling biometrics is available in the settings screen (for iOS only, for devices that support biometrics and where the user has enrolled in biometrics on a device level)
- Verify that toggling the option off means that all flows requiring pin now show the pin screen (e.g. go to the recovery phrase screen, the pin screen should appear)
- Verify that toggling the option on means that all flows requiring pin will use biometrics instead (e.g. go to the recovery phrase screen, the screen should be unlocked by biometrics)
- Toggling the option on will require either the pin to already be cached (from previous use) or the pincode input screen should appear.
- Test the "require with pin" setting item toggled on, then unlocking the app with and without biometry enabled.

### Related issues

- Fixes #1685 

### Backwards compatibility

Yes


gnardini (approved):
Looks great! 🔥

packages/mobile/src/pincode/authentication.ts
```diff
@@ -182,6 +182,10 @@ function storePinWithBiometrics(pin: string) {
   })
 }
 
+export function removeStoredPin() {
+  return removeStoredItem(STORAGE_KEYS.PIN)
```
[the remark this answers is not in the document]
kathaypacific: oopsies forgot to reply, i have added! :D

packages/mobile/locales/base/translation.json
```diff
@@ -677,6 +677,7 @@
       "useIris": "Turn on Iris recognition"
     }
   },
+  "useBiometryOption": "Use {{option}}",
```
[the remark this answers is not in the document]
kathaypacific: yeah good callout, i have updated :D

packages/mobile/src/pincode/authentication.test.ts
```diff
@@ -340,6 +341,25 @@ describe(updatePin, () => {
...
+    expect(mockedKeychain.resetGenericPassword).toHaveBeenCalledWith({ service: 'PIN' })
+  })
+  it('should throw an error if item could not be removed from keychain', async () => {
+    mockedKeychain.resetGenericPassword.mockRejectedValue('some error')
+
+    try {
+      await removeStoredPin()
+    } catch (error) {
+      expect(error).toEqual('some error')
+    }
```
jeanregisser, line 359: Problem with this is if it doesn't throw an error, we're not expecting anything and the test will happily pass 😅 

Here's one way to avoid this problem:
```suggestion
    expect(await removeStoredPin()).rejects.toThrowError('some error')
```

kathaypacific: oh you're right! i didn't really think about this before. i went through and updated all the instances of this try/catch type of assertion in this test file, there were a few.

one interesting thing i found was `rejects.toThrowError` didn't work for this particular assertion and the assertion on line 170, i had to use `rejects.toEqual`. i could verify that the tests fail if the `toEqual` value doesn't match but not really sure why the `toThrowError` doesn't work... 🤔 for me the `rejects.toEqual` assertion is okay enough here, but if you have any ideas on why `toThrowError` doesn't work as expected then i'd like to fix this!

jeanregisser: Ah yes I guess it's because the mocks rejects to a `string` and `toThrowError` probably expects an actual `Error` instance?

ValoraQA:
**Note:**

**Test rail links:**

**iOS 13:** https://valoraapp.testrail.io/index.php?/tests/view/10002

**iOS 14:** https://valoraapp.testrail.io/index.php?/tests/view/9747

**iOS 15:** https://valoraapp.testrail.io/index.php?/tests/view/9684

**CC:** @kathaypacific 
Thanks..!!

PR #1822, opened 2022-01-21
Replace user address in dapp URLs

### LINKED ISSUE #1821: Replace user address in dapp URLs
This will allow some dApps to open with the user already 'connected'. The motivating example is CeloTracker, where you can put the address in the URL like `https://celotracker.com/?address={{address}}` and it will show that address's balance.

AC:
- A dapp's url that includes the `{{address}}` string has it replaced with the user's address

gnardini:
### Description

This will allow some dApps to open with the user already 'connected'. The motivating example is CeloTracker, where you can put the address in the URL like `https://celotracker.com/?address={{address}}` and it will show that address's balance.

### Other changes

N/A

### Tested

Manually

### How others should test

No way for QA to test this at this time, but whenever Celo Tracker shows up in the Dapps screen then it should open with the user's account already set when opened

### Related issues

- Fixes #1821

### Backwards compatibility

N/A

### DIFF
--- a/packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
+++ b/packages/mobile/src/dappsExplorer/DAppsExplorerScreen.tsx
@@ -21,7 +21,6 @@ import { SafeAreaView } from 'react-native-safe-area-context'
 import { useDispatch, useSelector } from 'react-redux'
 import { openUrl } from 'src/app/actions'
 import { dappsListApiUrlSelector } from 'src/app/selectors'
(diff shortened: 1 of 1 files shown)


silasbw (approved):
nice.

i added https://github.com/valora-inc/dapp-list#templating

silasbw (approved):
:pray: !

ValoraQA:
Hey @MuckT please let us know what to be tested here so we can provide our feedback accordingly

PR #1830, opened 2022-01-22
Check that sentry is enabled before making calls

MuckT:
### Description

Sentry calls were being made when they shouldn't be e.g. e2e tests. This resulted in a large amount of sentry errors. This PR uses the SENTRY_ENABLED flag to check if we should be making the call prior to making the sentry call.

Slack Thread: https://valora-app.slack.com/archives/C02N3AR2P2S/p1642875669024700

### Tested

Tested in CI & sentry errors observed to be reduced (WIP).

[Issue 1](https://sentry.io/share/issue/3c309012982f42159758d39b91f4eb91)
[Issue 2](https://sentry.io/share/issue/d24cfb5430c0498eb173a92f162c7893)

### How others should test

No need for QA to test. Reviewers please check that the number of errors from Sentry is decreasing.

### Related issues

N/A

### Backwards compatibility

Yes


jh2oman (approved):

MuckT:
Closing as the calls to sentry were not from e2e tests.

PR #1832, opened 2022-01-24
(Biometry 5/4) - Use biometry to unlock app if "require pin on app start" enabled

kathaypacific:
### Description

- If "require pin on app start" and biometry are both enabled, attempt to fetch pin with biometry on app start. Display branding background image while biometry auth is in progress
- If biometry auth fails, display pin input screen to unlock app
- If user cancels biometry auth after failed attempt, display pin input screen
-  If "require pin on app start" is enabled and biometry is **not** enabled, the flow should remain unchanged. the pin input screen should be displayed.

### Other changes

Update PincodeLock tests and remove snapshot.

### Tested

Manually and unit tests for PincodeLock

### How others should test

As in the description, test unlocking the app with and without biometry enabled.

### Related issues

- Relates to [#1684](https://github.com/valora-inc/wallet/issues/1684)

### Backwards compatibility

Yes


gnardini (approved):
Good catch!

packages/mobile/src/pincode/PincodeLock.tsx
```diff
@@ -54,6 +64,22 @@ function PincodeLock() {
...
+      void onCompletePin(pin)
+    }
+  }, [])
+
+  if (shouldGetPinWithBiometry && !getPinWithBiometryError) {
+    return (
+      <SafeAreaView style={styles.loadingContainer}>
+        <Image testID="BackgroundImage" source={background} style={styles.backgroundImage} />
+        <Logo type={LogoTypes.LIGHT} height={70} />
+      </SafeAreaView>
```
jeanregisser, line 79: This is trying to recreate the native app splash, right?
I think we should avoid doing that as it won't fully match it in term of position of the elements. Leading to a slight jump of things, which doesn't look very good.
Instead I propose we use maybe a blurview or wait for the biometry check to happen before dismissing the real splash (though this one probably is not so trivial, given how the app inits right now).

jeanregisser (commented):
Great!

packages/mobile/src/pincode/PincodeLock.tsx
kathaypacific: yeah i wanted to recreate the native splash. i don't think the jump is noticeable - actually we can't even see the logo because the native animation covers the central part of the screen. i could remove the logo? i just need the background really, to give the illusion that the biometric verification is happening while the app is loading

jeanregisser: Yes let's remove the logo here as a compromise then.
Maybe one day we'll have a more direct app init where we could do the biometry check before we trigger the call to dismiss the native splash.

kathaypacific: sounds like a good compromise, i have removed the logo :)

jeanregisser (approved):

PR #1833, opened 2022-01-24
Resolve node-fetch to version 2.6.7 which contains security patch to fix CI

kathaypacific:
### Description

Our vulnerabilities CI has identified node-fetch as a new vulnerability and recommends to upgrade to version 2.6.7. This is the error identified https://github.com/advisories/GHSA-r683-j2x4-v87g

Example of the error in CI: https://github.com/valora-inc/wallet/runs/4918866732?check_suite_focus=true

Version 2.6.7 of node-fetch is a security patch update: https://github.com/node-fetch/node-fetch/releases/tag/v2.6.7

Most of our dependencies are using version 2 of node-fetch already so this resolution is low risk. Note that only `graphql-codegen` depended on node-fetch version 1.x.x, it is only used for the `build:gen-graphql-types` script but as far as I know this isn't being actively used.

We need to resolve this to make our CI green again 🟢

### Other changes

N/A

### Tested

N/A

### How others should test

Does not need to be tested

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,6 +105,7 @@
     "css-what": "5.0.1",
     "normalize-url": "6.0.1",
     "trim-newlines": "^4.0.2",
-    "glob-parent": "^5.1.2"
+    "glob-parent": "^5.1.2",
+    "node-fetch": "^2.6.7"
   }
 }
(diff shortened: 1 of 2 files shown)


gnardini (approved):
Thanks for taking care of this!

ValoraQA:

 CC : @kathaypacific

Issue #1385, opened 2021-11-05
Remove require cycles


jeanregisser:
### Description

There are currently >= 80 require cycle warnings shown by the React Native packager.
Some of them are in external libs, but most of them are in the wallet codebase.

Require cycles can create subtle issues, we should aim to remove all of them to avoid surprises in the future.

See also #1383.

### Acceptance Criteria

- We should address all require cycles present in the wallet codebase. We can leave require cycles in external libs we don't control.
- We should prevent new require cycles from landing in the codebase. This could be done using eslint [`import/no-cycle`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md).

PR #1834, opened 2022-01-24
Fix broken import for Sentry enabled config variable that was causing some error events to be lost

kathaypacific:
### Description

During a refactor, the SENTRY_ENABLED variable moved to config.ts but in the index.js we were still importing from the previous place. This means that SENTRY_ENABLED is always undefined and we are not sending the root errors to Sentry.

This PR fixes the broken import.

### Other changes

Remove unused sentryEnabled variable.

### Tested

N/A

### How others should test

This doesn't need to be tested by QA team

### Related issues

N/A

### Backwards compatibility

Yes


packages/mobile/index.js
```diff
@@ -9,12 +9,15 @@ if (__DEV__) {
...
 // This needs to happen early so any errors (including in the store) get caught
-import { SENTRY_ENABLED } from 'src/sentry/Sentry'
+import Config from 'react-native-config'
+import { stringToBoolean } from '@celo/utils/lib/parsing'
 import App from 'src/app/App'
 import * as Sentry from '@sentry/react-native'
 import 'react-native-gesture-handler'
 import { Text, TextInput } from 'react-native'
 
+const SENTRY_ENABLED = stringToBoolean(Config.SENTRY_ENABLED || 'false')
```
kathaypacific, line 19: i tried to import this from Config but ran into some runtime errors (typeerror: undefined is not an object evaluating _config.TOS_LINK) - i thought this was relatively low impact so went with this rather than spend time on the errir

silasbw (approved):

kathaypacific:
@silasbw totally agree, Sentry should be initialised as soon as possible. i think it already is actually - it is being initialised in the `appInit` saga which is called after the redux rehydrate https://github.com/valora-inc/wallet/blob/f7118cd9eceb2827001649ef56073dbe34b160c2/packages/mobile/src/redux/sagas.ts#L104 i don't think the app is functional until the redux state is rehydrated so i think this is as early as we can safely do it?

packages/mobile/index.js
jeanregisser: > i tried to import this from Config but ran into some runtime errors (typeerror: undefined is not an object evaluating _config.TOS_LINK) - i thought this was relatively low impact so went with this rather than spend time on the errir

Ah because of the [require cycles](https://github.com/valora-inc/wallet/issues/1385) we have :/
Looks good then, until we can properly address those.

jeanregisser (approved):

ValoraQA:

Cc: @kathaypacific

PR #1837, opened 2022-01-25
Android E2E Multiple Emulators Fixes 

MuckT:
### Description

Misc fixes to `walletconnect.spec.js` and changes to the `run_e2e.sh` to better support multiple emulators / unblock CI.

### Other changes

N/A

### Tested

Tested locally and in CI.

### How others should test

This does not need to be tested by QA. Reviewers should check the history of runs of this branch in iOS and Android E2E tests.

### Related issues

N/A

### Backwards compatibility

Yes


.github/workflows/e2e-android.yml
```diff
@@ -65,7 +65,7 @@ jobs:
           export PATH=$ANDROID_SDK_ROOT/platform-tools:$PATH
           export PATH=$ANDROID_SDK_ROOT/emulator:$PATH
           cd packages/mobile
-          yarn run test:e2e:android -w 3
+          yarn run test:e2e:android -w 1
```
kathaypacific, line 68: is this back to 1 on purpose? 🥲

packages/mobile/e2e/src/WalletConnect.spec.js
```diff
@@ -7,6 +7,7 @@ describe('Given Wallet Connect', () => {
     await quickOnboarding()
   })
 
+  // Testing to see if walletConnectV1 is the issue
```
kathaypacific, line 10: is this comment necessary?

packages/mobile/e2e/src/usecases/WalletConnectV1.js
```diff
@@ -149,7 +149,7 @@ export default WalletConnect = () => {
     let result = walletConnector.signTransaction(tx)
     await waitFor(element(by.text('Sign a Celo TX')))
       .toBeVisible()
-      .withTimeout(10 * 1000)
+      .withTimeout(15 * 1000)
```
kathaypacific, line 152: i wonder if it's worthwhile to make a wrapper for this line and the one above, like `toBeVisibleWithTimeout()` and inside that function you could control the timeout...makes it easier to maintain the timeout values?

packages/mobile/scripts/run_e2e.sh
```diff
@@ -190,12 +189,6 @@ if [ $PLATFORM = "android" ]; then
   # Run Detox Tests
   runTest
 
-  # Close active emulators
```
kathaypacific, line 193: is detox closing the emulators after the tests by default? or are we leaving the emulators running until the next run?

.github/workflows/e2e-android.yml
MuckT: Was attempting to see if that was the issue, but I've bumped it back up to 3.

packages/mobile/e2e/src/usecases/WalletConnectV1.js
MuckT: The Detox Maintainers have indicated they want to move away form `waitFor` syntax, but I am unsure on the priority in their roadmap: https://github.com/wix/Detox/issues/2190. Can wrap this in the utilities in another PR.

packages/mobile/scripts/run_e2e.sh
MuckT: Added this back in, as it does appear as if it was closing the emulators in CI after the run.

packages/mobile/e2e/src/WalletConnect.spec.js
MuckT: Removed in https://github.com/valora-inc/wallet/pull/1837/commits/7609f7d602b2cf6e67e6b41843c51bd3d993a984!

packages/mobile/e2e/src/usecases/WalletConnectV1.js
gnardini: +1 to moving it to an utils or somewhere where we can put a default timeout so it's easier to change. If they do end up changing the `waitFor` syntax we would only need to change it in one place (probably).

gnardini (approved):
Not sure I fully understand what's going on with all the `disableSynchronization`s and their order with other assertions but it looks like it's unblocking CI so that's nice :D

ValoraQA:
Hey @MuckT since "How others should test has n/a, " hence we are skipping this PR

PR #1841, opened 2022-01-25
Update translation.json

mpgaarciaa:


packages/mobile/locales/base/translation.json
```diff
@@ -1069,7 +1069,7 @@
     "titleRamp": "Add {{currency}} to start using Valora",
     "subtitleRamp": "Add funds with zero fees through one of our trusted external providers."
   },
-  "balances": "Balances",
+  "balances": "Assets",
```
MuckT, line 1072: If we're changing `Balances` to `Assets` we should probably also change the key too as to avoid confusion. Same thing with `"viewBalanaces": "View Assets"`

kathaypacific: i don't think we should be changing the keys every time we change the translations - having keys stay the same and updating translations without updating code is one of the big perks of having translations separately in a file like this...also changing translation keys makes our translations backwards incompatible with previous app versions (would break OTA translations)

kathaypacific: indeed if we decide as a business rule that we now refer to "balances" as "assets" then we should update the code to use proper business domain lingo but we'd need to either create a new key + keep this key for OTA translations, or create a new distribution list for OTA translations and not push this change to older app versions. this particular change doesn't strike me as a business domain change though, so i didn't think we needed an app wide change?

kathaypacific: @MuckT maybe we can take this offline, if you feel strongly that we should update the translation key and the code :) i'll approve and automerge this PR to unblock Paula (until this is merged she cannot do the translations for this key in Crowdin, the new workflow is to edit the source language in github instead of Crowdin) - either way if we do make the change to the translation key it will be one of us and not Paula who does it

kathaypacific (approved):

ValoraQA:

PR #1853, opened 2022-01-26
Use default genesis block to connect to testnet

nategraf:
### Description

Related to https://github.com/celo-org/celo-blockchain/pull/1811, this PR removes the logic to
download and manually set the genesis block when instantiating the blockchain client. When used with
a mobile client that includes the PR linked above, it will use the genesis block for the network
corresponding to the given network ID. This avoids the need to download the genesis block, such that
clients that do not use static nodes (e.g. clients using Forno) will not require downloading any
files before starting their node.

### Other changes

None

### Tested

WIP

### Backwards compatibility

Should be fully backwards compatible, but this should be tested.

### DIFF
--- a/packages/mobile/__mocks__/@celo/network-utils/index.ts
+++ b/packages/mobile/__mocks__/@celo/network-utils/index.ts
@@ -1,7 +1,3 @@
-export const GenesisBlockUtils = jest.fn()
-GenesisBlockUtils.getGenesisBlockAsync = jest.fn()
-GenesisBlockUtils.getChainIdFromGenesis = jest.fn()
-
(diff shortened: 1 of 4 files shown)


MuckT:
Looks like this has an error with `throw new Error("Private key must be 32 bytes long");` @Jean would this be an error which any non-org member hits on a PR? Also I am surprised that this PR attempted to run our checks without approval.

jeanregisser:
> Looks like this has an error with `throw new Error("Private key must be 32 bytes long");` @jean would this be an error which any non-org member hits on a PR?

Yes, PR from forked repos don't have access to the secrets for security reasons.

nategraf:
No longer relevant now that light client mode is disabled

PR #1865, opened 2022-01-27
Prevent network errors from being sent to Sentry

kathaypacific:
### Description

Prevent "network request failed" errors from being sent to Sentry when the user is offline, to reduce the noise. 

In this PR
- Make the Logger aware of if the user is online
- Make the Logger aware of the network errors that we want to exclude from Sentry
- Network errors should still be sent to Sentry if the user is online (this is a genuine error)
- Add a remote config value so that we can add network errors that we want to exclude on the fly

### Other changes

N/A

### Tested

Manually

### How others should test

This task does not require testing from QA team

### Related issues

N/A

### Backwards compatibility

Yes


packages/mobile/src/networkInfo/saga.ts
```diff
@@ -37,11 +37,15 @@ const isErrorResponse = (response: UserLocationData | ErrorResponse): response i
...
+  let isNetworkConnected = isConnected(connectionInfo)
+  yield put(setNetworkConnectivity(isNetworkConnected))
+  Logger.setIsNetworkConnected(isNetworkConnected)
   while (true) {
     try {
       connectionInfo = yield take(networkStatusChannel)
-      yield put(setNetworkConnectivity(isConnected(connectionInfo)))
+      const isNetworkConnected = isConnected(connectionInfo)
+      yield put(setNetworkConnectivity(isNetworkConnected))
+      Logger.setIsNetworkConnected(isNetworkConnected)
```
jeanregisser, line 48: In case changing the network state triggers an error, it's preferable to put it before the dispatch.
```suggestion
      const isNetworkConnected = isConnected(connectionInfo)
      Logger.setIsNetworkConnected(isNetworkConnected)
      yield put(setNetworkConnectivity(isNetworkConnected))
```

kathaypacific: yeah good idea, i swapped the order!

jeanregisser (approved):
Amazing! thanks so much! 🌐 👍

packages/mobile/src/services/ReactNativeLogger.ts
```diff
@@ -34,13 +41,38 @@ export default class ReactNativeLogger {
...
       console.info(console.trace())
     }
   }
 
+  setIsNetworkConnected = (isConnected: boolean) => {
+    this.isNetworkConnected = isConnected
+  }
+
+  setNetworkErrors = (errors: string[]) => {
+    this.networkErrors = [...this.networkErrors, ...errors]
```
jeanregisser, line 73: Oh sorry one small thing, this has the potential of growing indefinitely everytime the feature flags are fetched.
Would be be better to have a `DEFAULT_NETWORK_ERRORS` array to avoid this.

kathaypacific: oh yeah you're right, i overlooked this. i added a 'DEFAULT_SENTRY_NETWORK_ERRORS' to the global config.ts file so that it can live with the other sentry config variable, hope this is okay

packages/mobile/src/redux/migrations.ts
```diff
@@ -387,6 +387,13 @@ export const migrations = {
     },
   }),
   30: (state: any) => ({
+    ...state,
+    app: {
+      ...state.app,
+      sentryNetworkErrors: [],
```
gnardini, line 393: Not a big deal because biometry isn't live yet, but I think the order between 30 and 31 should be reversed :)
(ie biometry in 30 and network errors in 31)

packages/mobile/src/services/ReactNativeLogger.ts
```diff
@@ -34,13 +42,38 @@ export default class ReactNativeLogger {
...
       console.info(console.trace())
     }
   }
 
+  setIsNetworkConnected = (isConnected: boolean) => {
+    this.isNetworkConnected = isConnected
+  }
+
+  setNetworkErrors = (errors: string[]) => {
+    this.networkErrors = [...DEFAULT_SENTRY_NETWORK_ERRORS, ...errors]
```
gnardini, line 74: The `set` naming makes me think it'd override everything instead of appending. I think it makes sense to override anyways? Unless I'm missing something

gnardini (approved):
Couple minor nits, thanks a lot for taking care of this!!

packages/mobile/src/redux/migrations.ts
kathaypacific: oops sorry i should have added a comment - this was a hack on purpose, i think we want to patch 1.25 with this PR and i accidentally merged biometry before this 🙈

packages/mobile/src/services/ReactNativeLogger.ts
kathaypacific: oh yeah, this is true. if we don't have any defaults in the code and rely only on remote config values, we will have an edge case first time the user opens the app + they are offline and couldn't fetch remote config....but as i'm writing this i realise we can just set the default value of the remote config value in the code and do a redux migration. i'll go with that!

kathaypacific: hum. i removed the appending so this setter now just sets :) i thought i would get rid of the config variable altogether but i ended up keeping it. this logger is initialised way before the appInit saga where this setter is first called, so it makes sense to init this Logger with a default value to avoid sending network errors on app start. so now the default value is being use here and the remote config defaults, so i thought it's better to keep the source of truth in the config file

ValoraQA:

CC : @kathaypacific

PR #1869, opened 2022-01-27
refactor: Recipient Picker to be functional

dievazqu:
### Description

Refactoring RecipientPicker to be a functional component.


packages/mobile/src/recipients/RecipientPicker.tsx
```diff
@@ -47,37 +46,29 @@ interface Props {
   isOutgoingPaymentRequest: boolean
 }
 
-interface StateProps {
-  recipientInfo: RecipientInfo
-  showSendToAddressWarning: boolean
-}
+function RecipientPicker(props: RecipientProps) {
+  const recipientInfo = useSelector(recipientInfoSelector)
```
gnardini, line 50: nit: If you import `useSelector` from `src/redux/useSelector` instead of `react-redux` you don't need to set the `RootState` type.

packages/mobile/src/recipients/RecipientPicker.tsx
```diff
@@ -47,37 +46,29 @@ interface Props {
   isOutgoingPaymentRequest: boolean
 }
 
-interface StateProps {
-  recipientInfo: RecipientInfo
-  showSendToAddressWarning: boolean
-}
+function RecipientPicker(props: RecipientProps) {
```
gnardini, line 49: No need to change it here because it'd be annoying, just mentioning it for other situations, but the pattern we usually do it destructuring the props here, like `function RecipientPicker({ searchQuery, sections, etc }: Props) {` instead of using the `props` thing directly.

I'm not really sure if one or the other is better, just sharing how we usually do it

gnardini (approved):
Left a very small nit but looks great! 🚀

ValoraQA:
Thanks..!!

MuckT:
@ValoraQA we can cover this in our build verification tests and regression testing.

PR #1871, opened 2022-01-27
Don't allow sending tokens if exchange rate wasn't loaded successfully

### LINKED ISSUE #1870: Don't allow sending tokens if exchange rate wasn't loaded successfully
This fixes this error: https://sentry.io/organizations/valora-inc/issues/2963488661/?project=1250733&query=is%3Aunresolved+null+is+not+an+object&statsPeriod=14d

gnardini:
### Description

If the exchange rate isn't loaded, then the Send flow can't show the correct amounts. We are now preventing navigation to that screen if the exchange rate wasn't loaded.

### Other changes

N/A

### Tested

Manually and with unit tests.

### How others should test

To test this change you need the exchange rate query to fail. I did this by editing the code directly, but I don't know how others can reproduce this really :(

### Related issues

- Fixes #1870

### Backwards compatibility

N/A


packages/mobile/src/send/selectors.ts
```diff
@@ -7,3 +10,11 @@ export const getRecentPayments = (state: RootState) => {
 export const isSendingSelector = (state: RootState) => {
   return state.send.isSending
 }
+
+export const canSendTokensSelector = createSelector(
+  tokensWithTokenBalanceSelector,
+  localCurrencyToUsdSelector,
+  (tokensList, usdExchangeRate) => {
+    return tokensList.length > 0 && usdExchangeRate !== null
```
kathaypacific, line 18: just for my understanding, are we only checking for usdExchangeRate here (as opposed to the exchange rate for  euro or some other currency the user might have as default) because all the exchange rates are set at the same time so if one is not null then all of them will be safely not null?

kathaypacific (approved):
LGTM! 🔥

packages/mobile/src/send/selectors.ts
gnardini: Nope, we are only using the usd exchange rate in the new flow. The EUR and CELO exchange rates are there for legacy reasons.

Right now, to get the value for a user in local currency we transform from the token to USD and then from USD to local currency.

ValoraQA:

PR #1875, opened 2022-01-27
docs: updated firebase db rules for IHL node

cajubelt:
### Description

This will keep our db-rules.json file in sync with the current set of rules in the firebase UI.

I think we can improve on these rules ahead of M2/3, but I think this is fine for M1. More thoughts on this here: https://app.zenhub.com/workspaces/acquisition-squad-sprint-board-6010683afabec1001a090887/issues/valora-inc/in-house-liquidity/478

### Other changes

na

### Tested

ran a manual test with this rule set to confirm that someone can still go through the persona flow and see their KYC updates

### How others should test

n/a

### Related issues

na

### Backwards compatibility

na


packages/mobile/src/firebase/db-rules.json
```diff
@@ -88,9 +88,10 @@
       ".read": true
     },
     "inHouseLiquidity": {
+      ".write": "auth.uid === 'ihl-service-worker'",
+      ".read": "auth.uid === 'ihl-service-worker'",
       "$address": {
-        ".write": "auth.uid === 'ihl-service-worker'",
-        ".read": "(auth.uid != null && $address === root.child('users').child(auth.uid).child('address').val()) || auth.uid === 'ihl-service-worker'"
```
cajubelt, line 93: this 'read' rule didn't actually work

cajubelt: note that IHL does not store any PII (in Postgres or firebase)

jeanregisser (approved):
Thanks! 👍 

As a side note, we should probably move this out of the wallet repo.

PR #1878, opened 2022-01-28
Fix some network errors still incorrectly being sent to Sentry

kathaypacific:
### Description

Previously we were checking only the `error` argument of the logger error function for network errors, we should also check the `message` argument. As a consequence, some network errors were still being sent to Sentry.

The `message` argument of `Logger.error` is supposed to be a string but it sometimes receives an object. Typescript does not complain about calling `try {...} catch (error) { Logger.error(TAG, error) }` because `error` has type `any`. So I've added a `.toString()` before checking the value for network errors.

### Other changes

N/A

### Tested

Manually

### How others should test

No testing needed by QA

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/packages/mobile/src/services/ReactNativeLogger.ts
+++ b/packages/mobile/src/services/ReactNativeLogger.ts
@@ -42,8 +42,10 @@ export default class ReactNativeLogger {
     const sanitizedError =
       error && shouldSanitizeError ? this.sanitizeError(error, valueToPurge) : error
     const errorMsg = this.getErrorMessage(sanitizedError)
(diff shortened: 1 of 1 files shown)


jeanregisser (approved):
Thanks! 👍 
Next step upgrading TypeScript where the error type is now `unknown` in try catch blocks.
https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#using-unknown-in-catch-variables

Which will force us to check the type and prevent this.

kathaypacific:
@mergifyio refresh

ValoraQA:

CC : @kathaypacific

PR #1887, opened 2022-01-31
Resolve simple-get to version 4.0.1 which contains security patch to fix CI

kathaypacific:
### Description

Our vulnerabilities CI has identified simple-get as a new vulnerability and recommends to upgrade to version 4.0.1. This is the error identified https://github.com/advisories/GHSA-wpg7-2c88-r8xv

Example of the error in CI: https://github.com/valora-inc/wallet/runs/4988071468?check_suite_focus=true

The simple-get releases are not well documented so it is difficult to know which (breaking) changes are in which release. Several of our dependencies are using version 2 and 3 of simple-get (komenci, wallet connect, several celo projects) so we are relying on manual regression tests and the e2e CI to verify that resolving simple-get to v4 works for us.

We need to resolve this to make our CI green again 🟢

### Other changes

N/A

### Tested

N/A

### How others should test

Does not need to be tested explicitly (implicitly covered during regression tests)

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/.github/workflows/e2e-android.yml
+++ b/.github/workflows/e2e-android.yml
@@ -65,7 +65,7 @@ jobs:
           export PATH=$ANDROID_SDK_ROOT/platform-tools:$PATH
           export PATH=$ANDROID_SDK_ROOT/emulator:$PATH
           cd packages/mobile
-          yarn run test:e2e:android -w 1
+          yarn run test:e2e:android -w 3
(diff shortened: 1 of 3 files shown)


jeanregisser (approved):
Thanks so much! 👍

ValoraQA:
Thanks..!!

PR #1905, opened 2022-02-01
Improve analytics for pincode events to help understand if biometry is useful

kathaypacific:
### Description

In a discussion, Jean mentioned it would be helpful to have some more events to help us learn about the impact of biometry on the user journey. In particular:
- what's the success rate when unlocking the PIN when biometry is not enabled and when it is enabled?
- when biometry is enabled, how often does it fallback to really asking the PIN?

In this PR:
- add sPincodeType as super property that is sent with all events
- add events to tack pincode fetch start/success/error flows

Example sequence for failed biometry attempt where user had to use pin instead:
get_pincode_start (sPincodeType = "PhoneAuth")
get_pincode_with_biometry_start
get_pincode_with_biometry_error
get_pincode_with_input_start
get_pincode_with_input_complete
get_pincode_complete

### Other changes

Fix a typo in a base string. Removed snapshot tests that could be replaced by a normal test.

### Tested

Unit tests

### How others should test

No impact to customer journey, doesn't need to be tested

### Related issues

N/A

### Backwards compatibility

Yes


packages/mobile/src/analytics/ValoraAnalytics.test.ts
```diff
@@ -121,67 +159,17 @@ describe('ValoraAnalytics', () => {
     // Now that init has finished track should have been called
     expect(mockedAnalytics.track).toHaveBeenCalledTimes(1)
     expect(mockedAnalytics.track).toHaveBeenCalledWith(HomeEvents.drawer_navigation, {
-      celoNetwork: 'alfajores',
```
gnardini, line 124: NIce improvement to this tests's readability! 👏

packages/mobile/src/navigator/NavigationService.ts
```diff
@@ -148,13 +150,15 @@ export async function ensurePincode(): Promise<boolean> {
...
     await requestPincodeInput(true, false)
+    ValoraAnalytics.track(AuthenticationEvents.get_pincode_complete)
     return true
   } catch (error) {
     if (error === CANCELLED_PIN_INPUT) {
       Logger.warn(`${TAG}@ensurePincode`, `PIN entering cancelled`, error)
     } else {
       Logger.error(`${TAG}@ensurePincode`, `PIN entering error`, error)
     }
+    ValoraAnalytics.track(AuthenticationEvents.get_pincode_error)
```
gnardini, line 161: If the user cancelled do we still count it as an error? Maybe we do (ie they don't remember their PIN so they cancel), just asking

gnardini (approved):
Great idea adding these events :)

packages/mobile/src/navigator/NavigationService.ts
kathaypacific: yeah good question! i think we do want this error regardless if it is a user action or genuine failure, because if these events are meant to help us understand the success rate of the pin flow with and without the biometrics, then we need to capture the user cancelled/forgot their pin error and count that towards the failure rate. but yeah i'm only 95% convinced about this, i'd like to leave it like this and ask gowtham to have a play with the nightly build tomorrw to see if the events are what he expects. and i can tweak if they are not :)

PR #1913, opened 2022-02-02
Android CI Fixes

### LINKED ISSUE #1907: Fix Multiple Workers in CI
### Description
Currently we are running three workers for Android and five workers for iOS in CI. The success rate for iOS and Android is 58.54% and 69.23% respectively. Ideally we should have a success rate of E2E tests of greater than 80% and a time of less than 25 minutes. 

| OS | Duration (Past 7 Days) | Success Rate (Past 7 Days) |
| ----- | ----- | ----- | 
| iOS | 30:14 | 69.23% |
| Android | 24:52 | 58.54% |

MuckT:
### Description

CI fixes for Android

### Other changes

Additional tests and test stubs for deep links. 

### Tested

Tested locally and in CI.

### How others should test

### Related issues

- Fixes #1907

### Backwards compatibility

Yes


packages/mobile/e2e/src/DeepLinkSend.spec.js
```diff
@@ -1,7 +1,8 @@
 import HandleDeepLinkSend from './usecases/HandleDeepLinkSend'
 import { quickOnboarding } from './utils/utils'
+import { launchApp } from './utils/retries'
```
kathaypacific, line 3: is this being used?

packages/mobile/scripts/run_e2e.sh
```diff
@@ -155,11 +155,6 @@ if [ $PLATFORM = "android" ]; then
 
     startPackager
 
-    # Kill all emulators if present
-    killall -9 qemu-system-x86_64 || true
```
kathaypacific, line 159: does removing this help to improve test stability, or is it just not needed? i was checking the history of this file and i see that we've had some code block to kill emulators that are alive at the time of running this script for a long time. github actions automatically cancels active jobs if new changes are pushed to the branch, so doesn't removing this block introduce problems?

packages/mobile/e2e/src/DeepLinkSend.spec.js
```diff
@@ -1,7 +1,8 @@
...
 
-describe('Deep link without account send', () => {
+describe('Given Deep Link', () => {
```
kathaypacific, line 5: is there a convention for how we name tests/specs? i see some of the top level specs are verbs but some are nouns and this one is "Given"...i don't have a preference but would be nice to have some consistent pattern to make the test reports really easy to read :) in the mobile unit tests the top level describe is usually the function or file name, so maybe this one could be `describe('Deep Link Send')` or something?

kathaypacific:
just a few questions from me :) looks good

packages/mobile/e2e/src/DeepLinkSend.spec.js
MuckT: I prefer the Given / When / Then behavior-driven development syntax for E2E tests, but this can also be a bit cumbersome for those not used to it.

MuckT:
@kathaypacific I tried several ways to detect existing emulators and all were failing. When I remove that check Android CI passes as expected.

packages/mobile/e2e/src/DeepLinkSend.spec.js
kathaypacific: yeah i like the given/when/then too, although i'm more a fan of consistency than anything else :) i mention this here because i see most other specs look like `describe('Account', () => {})` or `describe('Send cUSD', () => {})` etc, so for me the version before this change was more consistent. if there is a plan to follow up this PR with an update of the naming conventions for other tests then that's all good, just wanted to flag it

packages/mobile/e2e/src/usecases/HandleDeepLinkSend.js
```diff
@@ -1,13 +1,47 @@
...
+    await device.terminateApp()
+    await sleep(5000)
+    await launchApp({ url: PAY_AMOUNT_URL, newInstance: true })
+    await sleep(5000)
+    await dismissBanners()
+    // Correct name displayed
+    await expect(element(by.text('TestFaucet'))).toBeVisible()
+
+    // Tap Pay
+    await element(by.text('Pay')).tap()
```
kathaypacific, line 35: following our discussion on slack, should we use test id's for new tests instead of text? i was thinking about how to use translation keys instead but i'm not sure if this is very easy to do, since detox is testing a user facing app and we don't expose the translation keys there :/

packages/mobile/e2e/src/usecases/HandleDeepLinkSend.js
```diff
@@ -17,11 +51,20 @@ export default HandleDeepLinkSend = () => {
...
-    await inputNumberKeypad('1.5')
+    await inputNumberKeypad('0.1')
     await element(by.id('Review')).tap()
 
     // Correct name displayed
-    await expect(element(by.text('Crypto4BlackLives'))).toBeVisible()
+    await expect(element(by.text('TestFaucet'))).toBeVisible()
+
+    // Tap Pay
+    await element(by.text('Send')).tap()
```
kathaypacific, line 61: same as above, should we use id here?

kathaypacific:
> @kathaypacific I tried several ways to detect existing emulators and all were failing. When I remove that check Android CI passes as expected.

i'm not sure i follow, i see the change you made to detect existing emulators and the CI is green? looks good to me!

packages/mobile/scripts/run_e2e.sh
MuckT: Ended up killing the emulators on Android in a different way:
```
NUM_DEVICES=`adb devices -l | grep emulator- | wc -l` || echo "0" | bc
if [ $NUM_DEVICES -ge 1 ]; then
  echo "$NUM_DEVICES emulator(s) already running or attached, removing"
  adb devices | grep emulator | cut -f1 | while read line; do adb -s $line emu kill; done
fi
 ```

packages/mobile/e2e/src/usecases/HandleDeepLinkSend.js
MuckT: fixed in https://github.com/valora-inc/wallet/pull/1913/commits/aa3e205916c5ffe73f85246d4e5b60aa2fe37f5b. I opted for the text selector for TestFaucet since that text is based on the deeplinkrul which is controlled by the test.

kathaypacific (approved):
looks great! 🚀

PR #1919, opened 2022-02-02
feat(settings): Vary the behavior and appearance of the Link Bank Account settings item based on user state

jh2oman:
### Description

All to match the [figma designs ](https://www.figma.com/file/DAcpLsmPdeHT9Gt4eXhtWe/Finclusive-(US)?node-id=681%3A12616)

### Other changes

### Tested

### How others should test

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


packages/mobile/src/account/Settings.tsx
```diff
@@ -414,15 +423,85 @@ export class Account extends React.Component<Props, State> {
...
-      verificationPossible,
+      kycStatus,
       linkBankAccountEnabled,
+      hasLinkedBankAccount,
       mtwAddress,
+      linkBankAccountStepTwoEnabled,
+      t,
     } = this.props
+
+    // Not enabled
```
jh2oman, line 436: The meat of this PR is in all of these if statements. I could have made it in less lines of code, but I felt it was better to be more verbose and explicit about the business logic

lisaqyli: i think being explicit always > less lines of code

packages/mobile/locales/base/translation.json
```diff
@@ -1131,8 +1131,9 @@
     "delete": "Delete"
   },
   "syncingBankAccount": "Syncing bank account...",
-  "linkBankAccountSettingsTitle": "Link Bank Accounts",
+  "linkBankAccountSettingsTitle": "Linked Bank Accounts",
```
lisaqyli, line 1134: shouldn't this title vary based on whether user has linked bank account yet? I would imagine it would be called `linked bank accounts` if there are bank accounts linked, while called `link bank accounts` if there are not bank accounts linked.

jh2oman: The designs always call for `Linked bank accounts`. And i think its fine that it isn't a CTA since it also says `get access`

lisaqyli (approved):

packages/mobile/src/redux/migrations.ts
```diff
@@ -409,4 +409,11 @@ export const migrations = {
       isSettingPin: undefined,
     },
   }),
+  32: (state: any) => ({
```
cajubelt, line 412: not sure we need v32 just yet, I don't see a release branch cut for v31 yesterday... just the hotfix which doesn't include any schema migration

cajubelt: also, I'm not sure we need this to begin with, since the migration will produce a falsy value for hasLinkedBankAccount anyway

cajubelt (commented):
I reviewed everything but the tests and it looks good so far-- leaving to someone else to approve if they can look at the tests too

packages/mobile/src/redux/migrations.ts
jh2oman: Asked in [#engineering-team](https://valora-app.slack.com/archives/C025V1D6F3J/p1643913042569519)

Apparently we should update it every time!

ValoraQA:
Hey @MuckT since "how other should test" has a statement "Does this need to be tested by QA in the next release cycle? If so please give a brief explanation of how to test these changes." so please provide more information about we can test the Pr , so we can write test cases, perform it and share feedback accordingly 

Thanks..!!

MuckT:
@jh2oman can you please respond to @ValoraQA?

PR #1922, opened 2022-02-03
Add some delay for biometry authentication to allow native verification animation to run fully

kathaypacific:
### Description

Add delay in biometry get/set functions so that the native animations can run fully before moving the user on. Without the delay, the animation can be running while the app is navigating to the next screen already.

With a 800ms delay, the animation runs like this:

https://user-images.githubusercontent.com/20150449/152314145-9f21302e-1f6f-4123-823b-889b1930b645.MP4

### Other changes

N/A

### Tested

Unit tests and manually

### How others should test

In the onboarding flow when enabling biometrics, and in flows where you need biometry to get the pin (recovery phrase screen, send flow, change pin, etc.) see that the biometric verification animation run to the end before the app moves on.

### Related issues

N/A

### Backwards compatibility

N/A

### DIFF
--- a/packages/mobile/src/pincode/authentication.test.ts
+++ b/packages/mobile/src/pincode/authentication.test.ts
@@ -100,6 +100,7 @@ describe(getPasswordSaga, () => {
 describe(getPincode, () => {
   beforeEach(() => {
     jest.clearAllMocks()
+    jest.useRealTimers()
     mockedNavigate.mockReset()
     clearPasswordCaches()
   })
@@ -131,6 +132,7 @@ describe(getPincode, () => {
       storage: 'storage',
     })
     const pin = await getPincode()
+    jest.runAllTimers()
 
     expect(pin).toEqual(mockPin)
     expect(mockedKeychain.getGenericPassword).toHaveBeenCalledTimes(1)
@@ -192,6 +194,7 @@ describe(getPincode, () => {
 
 describe(getPincodeWithBiometry, () => {
   it('returns the correct pin and populates the cache', async () => {
+    jest.useRealTimers()
     clearPasswordCaches()
     mockedKeychain.getGenericPassword.mockResolvedValue({
       password: mockPin,
@@ -200,6 +203,7 @@ describe(getPincodeWithBiometry, () => {
       storage: 'storage',
     })
     const retrievedPin = await getPincodeWithBiometry()
+    jest.runAllTimers()
 
     expect(retrievedPin).toEqual(mockPin)
     expect(getCachedPin(DEFAULT_CACHE_ACCOUNT)).toEqual(mockPin)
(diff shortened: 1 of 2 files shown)


packages/mobile/src/pincode/authentication.ts
```diff
@@ -276,6 +277,12 @@ export function* getPasswordSaga(account: string, withVerification?: boolean, st
 
 type PinCallback = (pin: string) => void
 
+async function addBiometryVerificationDelay() {
+  // insert artificial delay so that the native biometry verification
+  // animation can run the full course
+  return new Promise((resolve) => setTimeout(resolve, BIOMETRY_VERIFICATION_DELAY))
+}
+
```
jeanregisser, line 285: I noticed we already have 2 helpers that can do this:

https://github.com/valora-inc/wallet/blob/aaa9de2af0e32814f214f5b846839c2d9ed7826c/packages/mobile/src/geth/saga.ts#L2

https://github.com/valora-inc/wallet/blob/57cd5a341cbffc62903380dffbbb09850f6ec828/packages/mobile/src/pincode/authentication.ts#L12

We should pick one and remove the other (removal in a different PR) 😅

packages/mobile/src/pincode/authentication.test.ts
```diff
@@ -100,6 +100,7 @@ describe(getPasswordSaga, () => {
 describe(getPincode, () => {
   beforeEach(() => {
     jest.clearAllMocks()
+    jest.useRealTimers()
```
jeanregisser, line 103: nit: can we avoid this since we're calling `runAllTimers`? otherwise it seems we're introducing a 800ms delay several times in our test suite.

jeanregisser (commented):
Looking great! 🤩

packages/mobile/src/pincode/authentication.ts
kathaypacific: noooo i didn't even see there was a sleep function right there 🤦🏻‍♀️ i went with the sleep from the @celo project because it's used in more places (and we also use a retry function from there) so feels like it's easier to remove `sleep-promise`

packages/mobile/src/pincode/authentication.test.ts
kathaypacific: i think my brain is not working, i don't know why i did `useRealTimers` with `runAllTimers`. `runAllTimers` didn't work well, I think because the sleep is inside the tested function. now that i am using the sleep function from the @celo project i just mocked the sleep function and removed these real timers

jeanregisser (approved):

ValoraQA:

**Note:**
-We have verified face ID for iOS 14,15 & touch ID for iOS13
-As this features is iOS specific, so we have blocked android test cases
**Test rail links:**

**iOS 13:** https://valoraapp.testrail.io/index.php?/tests/view/10030

**iOS 14:** https://valoraapp.testrail.io/index.php?/tests/view/9775

**iOS 15:** https://valoraapp.testrail.io/index.php?/tests/view/9712

**CC:** @kathaypacific 
Thanks..!!

PR #1925, opened 2022-02-03
[Send screen] Make max button use max token value instead of max fiat value

### LINKED ISSUE #1653: The max amount in fiat and crypto don't match
There is different max amount of Moo token, depending from if user taps ‘max’ when in fiat mode, or in crypto.

AC:
- [ ] clicking MAX always maxes the token amount (and then the fiat value of approximated from the max token amount) except for "gas fee tokens"
- [ ] for gas fee tokens MAX always maxes the token amount less the gas fee
- [ ] when the user presses the swap button, the values for both token and estimated value are set to 0

kathaypacific:
### Description

This PR updates:
- pressing the swap button clears the previous value entered
- changing the token type clears the previous value entered
- pressing the max button maxes out the token value always (even if the fiat currency is focused) 

https://user-images.githubusercontent.com/20150449/152537874-64641d98-31ae-4677-b4c6-b319a551c3d9.MP4

### Other changes

N/A

### Tested

Manually and unit tests

### How others should test

On the send screen, try with different tokens and press the max button with fiat and token currency as input. Also verify that inputting values manually (without max button) works as usual. Pressing the swap button/changing token type should clear the value.

### Related issues

- Fixes #1653 

### Backwards compatibility

Yes


packages/mobile/src/send/SendAmount/SendAmount.test.tsx
```diff
@@ -206,6 +206,47 @@ describe('SendAmount', () => {
...
+      const { getByTestId } = render(
+        <Provider store={store}>
+          <SendAmount {...mockScreenProps()} />
+        </Provider>
+      )
+
+      fireEvent.press(getByTestId('MaxButton'))
+
+      // The value expected here is |balance| - the send fee (1 cUSD)
+      expect(getElementText(getByTestId('InputAmountContainer'))).toEqual('$21.85')
```
kathaypacific, line 239: i suppose we round down so that we don't give the illusion that we are sending more money than the user has? but then we display 21.86cUSD below? i'm not sure what the desired behaviour is to be honest, but we do use 2 different ways to format the 2 lines (i.e. on the input line, we round to 2 decimal but on the line below, we round to 2 significant figures)

Screenshot 2022-02-03 at 17 41 44
Screenshot 2022-02-03 at 17 41 53

kathaypacific: i made an executive decision to format the numbers consistently with the send confirmation screen. i just don't think it makes sense to show something like $21.85 on this screen then $21.86 on the send confirmation screen :/

packages/mobile/src/send/SendAmount/index.tsx
```diff
@@ -152,31 +162,38 @@ function SendAmount(props: Props) {
   )
   const maxBalance = tokenInfo?.balance.minus(feeEstimate) ?? ''
   const maxInLocalCurrency = useTokenToLocalAmount(maxBalance, transferTokenAddress)
+  const maxAmountValue = showInputInLocalAmount
+    ? maxInLocalCurrency
+      ? maxInLocalCurrency?.toFixed()
+      : ''
+    : maxBalance.toFixed(TOKEN_MAX_DECIMALS)
```
gnardini, line 169: Calling the `toFixed` here is losing some precision. Should we do just `toFixed()` or keep it as a `BigNumber`?

gnardini (commented):
There is one more issue that is pre-existing (sorry!). I've tested this and got a few reverted transactions: https://alfajores-blockscout.celo-testnet.org/tx/0x434e826982068681d6e22580e9ee83a1381d296fbb42b1120884d870cd125c3d/token-transfers

The problem is that it's trying to send `4746273679565546845` cEUR but the balance is `4746273679565546844` (notice the last digit).

This is because the `SendAmount` screen is sending the `SendConfirmation` screen  the input amount, and if the input amount is in local currency it needs to transform it back to the token amount, which loses precision. I guess this wasn't an issue before because of the rounding but now that we are sending the right value it became one.

So, maybe what we could do is send the `SendConfirmation` screen always the token amount. The precision loss should not affect the visible digits which is what it's used for. What do you think?

packages/mobile/src/send/SendAmount/index.tsx
```diff
@@ -152,31 +162,38 @@ function SendAmount(props: Props) {
   )
   const maxBalance = tokenInfo?.balance.minus(feeEstimate) ?? ''
   const maxInLocalCurrency = useTokenToLocalAmount(maxBalance, transferTokenAddress)
+  const maxAmountValue = showInputInLocalAmount
+    ? maxInLocalCurrency
+      ? maxInLocalCurrency?.toFixed()
```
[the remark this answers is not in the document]
kathaypacific: to reduce rounding errors i ended up keeping this `maxAmountValue` as a` BigNumber` so removed the need for a nested ternary :)

kathaypacific:
> There is one more issue that is pre-existing (sorry!). I've tested this and got a few reverted transactions: https://alfajores-blockscout.celo-testnet.org/tx/0x434e826982068681d6e22580e9ee83a1381d296fbb42b1120884d870cd125c3d/token-transfers
> 
> The problem is that it's trying to send `4746273679565546845` cEUR but the balance is `4746273679565546844` (notice the last digit).
> 
> This is because the `SendAmount` screen is sending the `SendConfirmation` screen the input amount, and if the input amount is in local currency it needs to transform it back to the token amount, which loses precision. I guess this wasn't an issue before because of the rounding but now that we are sending the right value it became one.
> 
> So, maybe what we could do is send the `SendConfirmation` screen always the token amount. The precision loss should not affect the visible digits which is what it's used for. What do you think?

@gnardini thank you for testing!! this is exactly the type of feedback i was hoping for. i took your advice and made the `SendConfirmation` screen always the token amount, i think it makes sense. i'm not getting bounced transactions anymore so i think it is working!

packages/mobile/src/send/saga.ts
```diff
@@ -391,16 +389,7 @@ export function* sendPaymentOrInviteSaga({
...
-        sendPayment,
-        recipient.address,
-        amount,
-        amountInLocalCurrency,
-        usdAmount,
-        tokenAddress,
-        comment,
-        feeInfo
-      )
+      yield call(sendPayment, recipient.address, amount, usdAmount, tokenAddress, comment, feeInfo)
```
kathaypacific, line 392: `amountInLocalCurrency` exists to be passed into the `sendPayment` saga but it wasn't being used there so i removed it and also removed from the action as well

ValoraQA:
Hey @MuckT. we have verified the above task on Latest Android internal Build V:1.26.0 and updated status in Test rails.

**Test rail link:**
**Android 8:**
https://valoraapp.testrail.io/index.php?/tests/view/9825&group_by=cases:section_id&group_order=asc&group_id=35

**Android 10:**
https://valoraapp.testrail.io/index.php?/tests/view/9951&group_by=cases:section_id&group_order=asc&group_id=35

**Devices:** Samsung Galaxy A8+(8.0), Samsung Galaxy M40(10.0)

**CC:** @kathaypacific 
Thanks..!!

ValoraQA:
Hey @MuckT, We have verified the above task on Latest Test flight Build V1.26.0 and updated status in Test rails.

**Test rail link:**
**iOS 13:**
https://valoraapp.testrail.io/index.php?/tests/view/10017&group_by=cases:section_id&group_order=asc&group_id=35

**Devices:** iPhone 7(13.3)

**CC:** @kathaypacific
Thanks..!!

PR #1936, opened 2022-02-04
remove edu screens from onboarding flow

### LINKED ISSUE #1612: Setup an experiment to remove education from the onboarding flow.
[Experiment proposal in notion](https://www.notion.so/valora-inc/Remove-education-from-on-the-onboarding-flow-edf2be9017e645e18c21874ad6e28c6c)

Eng notes:
- Since this is the first screen of the app, we still don't have the feature flag values here, and it would be weird to show the screen and then remove it.
- One option is to extend the Splash Screen a few seconds the first time to fetch the value of the feature flag and then decide which screen to show first.

AC:
- experiment according to specs in the proposal is ready to launch from Firebase.

lisaqyli:
### Description
This PR implements an experiment to remove education screen from onboarding flow. During implementation, I discovered that remote config values are fetched after the initial screen is launched. I calculated there was ~0.8s of lag between when the initial screen is launched and when the firebase value comes back. Therefore, I implemented a hardcoded 50/50 experiment, the config value will be written into the redux store when someone first installed Valora app. 

There are 2 tracking events implemented for analyzing the impact of this experiment to our onboarding funnel. 

### Tested

I added a unit test for the change language flow. We didn't have tests around navigator to begin with. I have some attempts (see my commit history) but they were not successful.

### Related issues

- Fixes #1612


packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -592,25 +600,56 @@ export function MainStackScreen() {
       pincodeType,
       account,
       hasSeenVerificationNux,
+      shouldSkipOnboardingEducationScreen,
```
cajubelt, line 603: I think the code could be easier to understand and safer from bugs if this were named differently, like `savedShouldSkipEducationScreen`. That way there's less ambiguity over whether you should use this or the underscore-prefixed variable in the actual logic later on

cajubelt: so you could have this as 
```suggestion
      shouldSkipOnboardingEducationScreen: savedShouldSkipEducationScreen,
```
and then refer to it as `savedShouldSkipEducationScreen` from here on

packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -592,25 +600,56 @@ export function MainStackScreen() {
...
+    let _shouldSkipOnboardingEducationScreen = shouldSkipOnboardingEducationScreen
+    if (shouldSkipOnboardingEducationScreen == undefined) {
+      if (Math.random() < 0.5) {
+        _shouldSkipOnboardingEducationScreen = true
+        dispatch(skipOnboardingEducationScreen(true))
+      } else {
+        _shouldSkipOnboardingEducationScreen = false
+        dispatch(skipOnboardingEducationScreen(false))
+      }
+    }
```
cajubelt, line 618:
```suggestion
    const _shouldSkipEducationScreen = savedShouldSkipEducationScreen ?? (Math.random() < 0.5)
    dispatch(skipOnboardingEducationScreen(_shouldSkipEducationScreen))
```

cajubelt: code golf

packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -592,25 +600,56 @@ export function MainStackScreen() {
...
       // User didn't go far enough in onboarding, start again from education
-      initialRoute = Screens.OnboardingEducationScreen
-    } else if (!account) {
-      initialRoute = choseToRestoreAccount
-        ? Screens.ImportWallet
+
+      ValoraAnalytics.track(
+        _shouldSkipOnboardingEducationScreen
+          ? OnboardingEvents.onboarding_education_skipped
+          : OnboardingEvents.onboarding_education_not_skipped
```
cajubelt, line 630: why are these two different events? Wouldn't it be simpler to have a single event with a boolean field on it indicating whether education was skipped?

packages/mobile/src/language/Language.test.tsx
```diff
@@ -35,4 +36,25 @@ describe('Language', () => {
     expect(changeLanguageSpy).toHaveBeenCalledWith('es-419')
     expect(navigate).toHaveBeenCalledWith(Screens.OnboardingEducationScreen)
   })
+
+  it('navigates to welcome screen if skip onboarding education screen flag is turned on in redux store', () => {
```
cajubelt, line 40: could you also add a test for when `shouldSkipOnboardingEducationScreen` is false? Also, I recommend mocking Math.random to make absolutely sure these tests don't pass by random accident. So in this case, you could mock Math.random to return 0.75, which means if the saved value were ignored and a new `shouldSave...` value were selected instead, the test would fail

cajubelt (commented):
nice! almost there, just added a suggestion on testing

packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -592,25 +600,56 @@ export function MainStackScreen() {
       pincodeType,
       account,
       hasSeenVerificationNux,
+      shouldSkipOnboardingEducationScreen,
     } = mapStateToProps(store.getState())
 
+    // Remove Onboarding Education Screen Experiment: Because remote configs are fetched after the initial route is launched,
```
MuckT, line 606: Can we use a remote config and pull it in early like we did with sentry? If it's a random 50 / 50 hardcoded split e2e tests will fail 50% of the time unless they are adjusted.

lisaqyli: The reason was fetching from firebase takes extra time, i estimated it to be about .8s so didn't want to prolong the initial loading time. It's possible that by doing some sort of parallel request, the actual time added to initial loading would be a lot less than .8s. Also another reason is just want to get a quick exp out of door without spending a lot of time on it. 

That was a good point about e2e testing. I've updated the e2e test and hopefully it will pass

packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -592,25 +598,51 @@ export function MainStackScreen() {
...
       account,
       hasSeenVerificationNux,
+      shouldSkipOnboardingEducationScreen: savedShouldSkipEducationScreen,
     } = mapStateToProps(store.getState())
 
+    // Remove Onboarding Education Screen Experiment: Because remote configs are fetched after the initial route is launched,
+    // The randomization is hardcoded here to achieve a 50/50 split, the value is written into the redux store so the same experience
+    // would persist. This block of code should be removed when the experiment is done.
+    const _shouldSkipOnboardingEducationScreen =
+      savedShouldSkipEducationScreen ?? Math.random() < 0.5
```
jh2oman, line 608: I think that a more reliable and cleaner way to ensure persistance would be to use a random number generator that uses a seed instead of Math.random() and use deviceId as a seed.

For example, using [seedrandom](https://www.npmjs.com/package/seedrandom)
```
const rng = seedrandom(DeviceInfo.getDeviceId())
const randomNumber = rng() // Always the same random number for a given deviceId
```

lisaqyli: great suggestion

packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -592,25 +599,53 @@ export function MainStackScreen() {
       pincodeType,
       account,
       hasSeenVerificationNux,
+      shouldSkipOnboardingEducationScreen: savedShouldSkipEducationScreen,
     } = mapStateToProps(store.getState())
 
+    // Remove Onboarding Education Screen Experiment: Because remote configs are fetched after the initial route is launched,
+    // The randomization is hardcoded by device id here to achieve a 50/50 split, the value is written into the redux store so the same experience
+    // would persist. This block of code should be removed when the experiment is done.
+    const rng = seedrandom('deviceInfoModule.getUniqueId()')
```
jh2oman, line 608: Is this a mistake?  It looks like just a hardcoded string instead of a function call

packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -592,25 +599,53 @@ export function MainStackScreen() {
...
+    const randomNumber = rng()
+    const _shouldSkipOnboardingEducationScreen =
+      savedShouldSkipEducationScreen ?? randomNumber < 0.5
+    if (savedShouldSkipEducationScreen == undefined) {
+      dispatch(skipOnboardingEducationScreen(_shouldSkipOnboardingEducationScreen))
```
jh2oman, line 613: Now that you are using seedrandom which guarantees consistency, is there any need to have a value in redux for this?

packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -594,23 +597,46 @@ export function MainStackScreen() {
...
+          : OnboardingEvents.onboarding_education_not_skipped
+      )
+      initialRoute = _shouldSkipOnboardingEducationScreen
+        ? Screens.Welcome
         : Screens.OnboardingEducationScreen
+    } else if (!account) {
+      if (choseToRestoreAccount) {
+        initialRoute = Screens.ImportWallet
+      } else {
+        ValoraAnalytics.track(
```
jophish, line 626: nit: these ternary expressions are duplicated; maybe worth pulling them out somehow? similar story with `initialRoute`, but don't want to sacrifice clarity for compactness.

packages/mobile/src/utils/seedRandom.ts
```diff
@@ -0,0 +1,7 @@
+import deviceInfoModule from 'react-native-device-info'
+import seedrandom from 'seedrandom'
+
+export function getRandomByUUID() {
+  const rng = seedrandom(deviceInfoModule.getUniqueId())
```
jophish, line 5: nit: 
```suggestion
  return seedrandom(deviceInfoModule.getUniqueId())()
```

jh2oman (approved):

packages/mobile/src/navigator/Navigator.tsx
```diff
@@ -594,23 +597,46 @@ export function MainStackScreen() {
...
+          _shouldSkipOnboardingEducationScreen
+            ? OnboardingEvents.onboarding_education_skipped
+            : OnboardingEvents.onboarding_education_not_skipped
+        )
+        initialRoute = _shouldSkipOnboardingEducationScreen
+          ? Screens.Welcome
+          : Screens.OnboardingEducationScreen
+      }
```
cajubelt, line 634:
```
    } else if (!name || !acceptedTerms || pincodeType === PincodeType.Unset || (!account && !choseToRestoreAccount)) {
      // User didn't go far enough in onboarding, start again from education
      if (_shouldSkipOnboardingEducationScreen) {
        initialRoute = Screens.Welcome
        ValoraAnalytics.track(OnboardingEVents.onboarding_education_skipped)
      } else {
        initialRoute = Screens.OnboardingEducationScreen
        ValoraAnalytics.track(OnboardingEvents.onboarding_education_not_skipped)
      }
    } else if (!account && choseToRestoreAccount) {
      initialRoute = Screens.ImportWallet
```
an idea for solving Joe's comment

cajubelt: (removes the dupe ternaries)

packages/mobile/src/navigator/Navigator.tsx
cajubelt: this can be done with CleverTap

packages/mobile/src/navigator/Navigator.tsx
cajubelt: if you wanna go with this: https://github.com/valora-inc/wallet/pull/2011

ValoraQA:
Hey @MuckT we have verified the above task on Latest Android Internal Release Build & Test flight Release Build V1.27.0 and updated status in Test rails.

**Test rail link:**
**Android 10:** https://valoraapp.testrail.io/index.php?/tests/view/12910&group_by=cases:section_id&group_order=asc&group_id=37

**Android 11:** https://valoraapp.testrail.io/index.php?/tests/view/12941&group_by=cases:section_id&group_order=asc&group_id=37

**iOS 14:** https://valoraapp.testrail.io/index.php?/tests/view/12972&group_by=cases:section_id&group_order=asc&group_id=37

**iOS 15:** https://valoraapp.testrail.io/index.php?/tests/view/13003&group_by=cases:section_id&group_order=asc&group_id=37

**Observation:** For few devices walkthrough screens flow are appearing when user installs the Valora app and opens the app and selects any language option. We have raised a [Query](https://github.com/valora-inc/wallet/issues/2049) for it.

**Devices:**  iPhone 13 mini(15.1.1), iPhone 12 (14.7.1), Google Pixel 2XL (11.0), Samsung Galaxy M40 (10.0)

**CC:** @lisaqyli 

Thanks..!!

Issue #1417, opened 2021-11-11
Investigate the nonce issue


aleksey-nov:
Follow up to the bug with nonce management that was discussed on slack [here](https://valora-app.slack.com/archives/CL7BVQPHB/p1636131905092600)

To do for Jean: 
Estimate how many users are impacted;
Write down acceptance criteria before refinement meeting.

PR #1954, opened 2022-02-07
Throw error when signing a transaction with a 0 gas price

jeanregisser:
### Description

TXs with a 0 gas price are currently stuck in the txpool.
With this change, we'll be alerted when we're signing such TXs and can get more data about why/when it's happening.

Note: starting with celo-blockchain >= 1.5.0, TXs with a 0 gas price will be evicted instead of sitting in the txpool indefinitely, but still would be good to never send such invalid TXs.

### Tested

Added new unit test and manually trying signing a TX with a 0 gas price.

### How others should test

N/A

### Related issues

- #1417
- https://github.com/celo-org/celo-blockchain/pull/1817 (fix for TXs with 0 gas price not being evicted).

### Backwards compatibility

Yes

### DIFF
--- a/packages/mobile/src/geth/GethNativeBridgeSigner.ts
+++ b/packages/mobile/src/geth/GethNativeBridgeSigner.ts
@@ -52,6 +52,12 @@ export class GethNativeBridgeSigner implements Signer {
     addToV: number,
     encodedTx: RLPEncodedTx
   ): Promise<{ v: number; r: Buffer; s: Buffer }> {
+    const { gasPrice } = encodedTx.transaction
+    if (gasPrice === '0x0' || gasPrice === '0x' || gasPrice === '0' || !gasPrice) {
+      // Make sure we don't sign and send transactions with 0 gas price
+      // This resulted in those TXs being stuck in the txpool for nodes running geth < v1.5.0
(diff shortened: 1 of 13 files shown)


packages/mobile/src/geth/GethNativeBridgeSigner.ts
```diff
@@ -52,6 +52,12 @@ export class GethNativeBridgeSigner implements Signer {
     addToV: number,
     encodedTx: RLPEncodedTx
   ): Promise<{ v: number; r: Buffer; s: Buffer }> {
+    const { gasPrice } = encodedTx.transaction
+    if (gasPrice === '0x0' || gasPrice === '0x' || gasPrice === '0' || !gasPrice) {
```
gnardini, line 56: If a WalletConnect dapp tries to sign a tx without a `gasPrice` are we adding it ourselves? It'd be nice if those don't fail.

gnardini (approved):
Small question but makes sense to me to not send txs we know will fail :)

packages/mobile/src/geth/GethNativeBridgeSigner.ts
jeanregisser: Yes we're adding it ourselves earlier in the flow. So this won't fail for them.

PR #1967, opened 2022-02-08
Make error logging consistent so that they are grouped correctly in Sentry

kathaypacific:
### Description

Now that we are sending errors to Sentry, consistency is more important than ever so that similar errors can be grouped together to help us find issues/trends faster.

We've noticed some strange errors come through on Sentry like `[object object]` and `undefined`. I think this is caused by what we are sending to the `Sentry.captureException` function. Sometimes we pass a whole error object to the `message` which is expecting a string. 

In this PR:
- explicitly turn the message sent to Sentry to a string, as Typescript currently doesn't warn us against passing objects so we can't really prevent this in the future until TS is upgraded
- go through the app and make sure the logger error function is called like `Logger.error(<some tag>, <some message>, <error object?>)`

### Other changes

N/A

### Tested

Manually

### How others should test

This does not need to be tested

### Related issues

N/A

### Backwards compatibility

N/A

### DIFF
--- a/packages/mobile/src/account/Profile.tsx
+++ b/packages/mobile/src/account/Profile.tsx
@@ -40,7 +40,7 @@ function Profile({ navigation, route }: Props) {
       // Delete old profile pictures if necessary.
       if (picturePath && picturePath !== newPictureUri) {
         RNFS.unlink(picturePath).catch((e) => {
-          Logger.error('Error deleting old profile picture:', e)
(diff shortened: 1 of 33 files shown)


gnardini (approved):
Thanks a lot for taking care of this, my bad on the Sentry PR!

jeanregisser (commented):
Thanks for going through all this! 👍

PR #1422, opened 2021-11-11
Starts work for merchant payments

dckesler:
### Description

Starts the merchant payments flow.

### Tested

Scanned a merchant payments QR code everything was correct. Can still scan other types of QR codes without errors.

### How others should test

This doesn't need QA yet.

### Backwards compatibility

Yes backwards compatible.

PR #1976, opened 2022-02-09
Merchant payments (continued from #1422)

nicolasbrugneaux:
### Description

In preparation for celo-connect where we would like the conference participants to pay with valora using the merchant-payments API, here is the PR adding that behaviour (under a feature flag)

### Other changes

Nothing normally

### Tested

[_An explanation of how the changes were tested or an explanation as to why they don't need to be._](https://user-images.githubusercontent.com/3650621/157277234-b3eea4f1-2a21-4613-b914-01176e66721a.mov)

### How others should test

We have a test shop here: https://sandbox-celo-demo-store.demo.firstdag.com/

### Backwards compatibility

This new feature is behind a feature flag, should be 100% backwards compatible.


packages/mobile/src/qrcode/utils.ts
```diff
@@ -166,3 +175,29 @@ export function* handleBarcode(
...
+}
+
+function* paymentDeepLinkHandlerDisabled(barcode: QrCode) {
+  yield put(showError(ErrorMessages.QR_FAILED_INVALID_ADDRESS))
+  Logger.error('A payment deep link was scanned without a set handler', barcode.data)
+}
+
+function* paymentDeepLinkHandlerMerchant(barcode: QrCode) {
+  const numberVerified = yield select(numberVerifiedSelector)
+  if (numberVerified || true) {
```
gnardini, line 194: Is having a verified number a requirement? Feels like it shouldn't but in any case this `|| true` shouldn't be here I assume? For testing, in the settings menu you can toggle verification status if that's useful

packages/mobile/test/schemas.ts
```diff
@@ -783,7 +784,7 @@ export const v24Schema = {
     ...v23Schema.transactions,
     transactions: [],
   },
-  invite: undefined,
+  invites: undefined,
```
gnardini, line 787: We shouldn't change this

packages/mobile/src/flags.ts
```diff
@@ -7,6 +7,7 @@ export const features = {
   SHOW_CASH_OUT: true,
   PNP_USE_DEK_FOR_AUTH: true,
   SHOW_INVITE_MENU_ITEM: false,
+  MERCHANT_PAYMENTS: false,
```
gnardini, line 10: Is this being used anywhere?

packages/mobile/src/merchantPayment/MerchantPaymentScreen.tsx
```diff
@@ -0,0 +1,286 @@
...
+        dispatch(showMessage(t('merchantPaymentSubmitSuccess')))
+        navigateHome()
+      } catch (e) {
+        setSubmitting(false)
+        dispatch(showError(t('merchantPaymentSubmitFail')))
+      }
+    }
+  }, [sdkCharge, dispatch, submitting])
+
+  useEffect(() => {
```
gnardini, line 96: For async operations we use `useAsync` instead of `useEffect`. It behaves very similarly with the difference that it allows the function to be `async` and returns a `loading`, `error` and `result` with the result of the async function.

packages/mobile/src/merchantPayment/MerchantPaymentScreen.tsx
```diff
@@ -0,0 +1,286 @@
...
+    void (async () => {
+      const kit = await getContractKitAsync()
+      const wallet = kit.getWallet()! as UnlockableWallet
+      const [account, dekAddress] = wallet.getAccounts()
+      await wallet.unlockAccount(account, await getPassword(account), UNLOCK_DURATION)
```
gnardini, line 101: This will cause the PIN to be requested as soon as the screen is opened, right? Is it necessary to do it now? Can't we wait until the user confirms the transaction first?

packages/mobile/src/merchantPayment/MerchantPaymentScreen.tsx
```diff
@@ -0,0 +1,286 @@
...
+
+      const chainHandler = new ContractKitTransactionHandler(kit)
+      const charge = new Charge(routeParams.apiBase, routeParams.referenceId, chainHandler, false)
+      try {
+        const info = await charge.getInfo()
+        setSdkCharge(charge)
+        setAmount(divideByWei(new BigNumber(info.action.amount)).decimalPlaces(2))
+        setBusiness(info.receiver.businessData)
+      } catch (e: unknown) {
+        dispatch(showError(t('merchantPaymentSetup')))
```
gnardini, line 74: Is this the right error message? We should also `Logger.error` the error

packages/mobile/src/merchantPayment/MerchantPaymentScreen.tsx
```diff
@@ -0,0 +1,286 @@
...
+    if (tokenAddress) {
+      dispatch(estimateFee({ feeType: FeeType.SEND, tokenAddress }))
+    }
+  }, [tokenAddress])
+
+  useEffect(() => {
+    if (!isDekRegistered) {
+      dispatch(estimateFee({ feeType: FeeType.REGISTER_DEK, tokenAddress }))
+    }
+  }, [isDekRegistered])
```
gnardini, line 140: Should the token address also be in the dependency array?

packages/mobile/src/merchantPayment/MerchantPaymentScreen.tsx
```diff
@@ -0,0 +1,286 @@
...
+import { useTokenInfoBySymbol } from 'src/tokens/hooks'
+import { fetchTokenBalances } from 'src/tokens/reducer'
+import { Currency } from 'src/utils/currencies'
+import { divideByWei } from 'src/utils/formatting'
+import { getContractKitAsync } from 'src/web3/contracts'
+import { isDekRegisteredSelector } from 'src/web3/selectors'
+
+type Props = StackScreenProps<StackParamList, Screens.MerchantPayment>
+
+function MerchantPaymentScreen({ route }: Props) {
```
gnardini, line 40: I don't want to be annoying because I know there's a time crunch here, but it feels like this screen is doing too much and some things can be sent to other helper components or custom hooks.

For example:
- All the fee handling should probably be in it's own component.
- All the payments stuff and interaction with the payments SDK could be a hook that holds all the logic related to it and just return whatever is needed to draw the UI.

So basically anything that is encapsulated or pure logic should be outside of the screen and just UI-related things should be in this file.

I know not all files in this repo follow the above though, there are some screens that are not properly designed but we do like to avoid it in new components.

What do you think @jeanregisser ?

gnardini (commented):
Looks good! Left a couple nits mostly. 

Would you mind also sharing the necessary steps to be able to run this locally? As in example values for the URL, a QR code value to scan and whatever else might be necessary

silasbw (changes requested):
I just skimmed this, but it looks like it lacks tests.  Can you please add testing?

packages/mobile/locales/en-US/translation.json
```diff
@@ -1195,5 +1195,12 @@
   "others": "Others",
   "nomSpaceRecipient": "Nomspace: {{name}}",
   "celoDollarAmount": "<0></0> Celo Dollar",
-  "celoDollarAmount_plural": "<0></0> Celo Dollars"
-}
\ No newline at end of file
+  "celoDollarAmount_plural": "<0></0> Celo Dollars",
+  "merchant": "Merchant Test Screen",
```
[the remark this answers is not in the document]
kathaypacific: just to add here that translations should be modified in https://github.com/valora-inc/wallet/blob/main/packages/mobile/locales/base/translation.json, not this file :)

packages/mobile/src/qrcode/utils.ts
```diff
@@ -166,3 +175,29 @@ export function* handleBarcode(
...
+    const { api_base: apiBase, reference_id: referenceId } = parse(barcode.data, true).query
+    if (typeof apiBase === 'string' && typeof referenceId === 'string') {
+      navigate(Screens.MerchantPayment, { apiBase, referenceId })
+    }
```
jeanregisser, line 198: Would be nice to show an error message when the params are incorrect.

packages/mobile/src/qrcode/utils.ts
```diff
@@ -111,6 +115,11 @@ export function* handleBarcode(
     yield call(initialiseWalletConnect, barcode.data, WalletConnectPairingOrigin.Scan)
     return
   }
+  if (barcode.data.startsWith('celo://wallet/payment')) {
```
jeanregisser, line 118: Can we use a different deeplink? We already have`celo://wallet/pay?xxx` which are unfortunately matched using `startsWith` in https://github.com/valora-inc/wallet/blob/712593a970a36f222864022751ff66ac753f02cf/packages/mobile/src/app/saga.ts#L251
So they could be incorrectly handled if scanned via the device camera and passed as a deeplink.

Maybe `celo://wallet/merchantPay`?

Additionally, it would be nice to handle these links inside the existing `handleDeepLink` method too.

packages/mobile/src/merchantPayment/MerchantPaymentScreen.tsx
```diff
@@ -0,0 +1,286 @@
...
+      try {
+        await sdkCharge.submit({ phoneNumber: e164PhoneNumber })
+        setSubmitting(false)
+        dispatch(showMessage(t('merchantPaymentSubmitSuccess')))
+        navigateHome()
+      } catch (e) {
+        setSubmitting(false)
+        dispatch(showError(t('merchantPaymentSubmitFail')))
+      }
+    }
```
jeanregisser, line 93: Though this works in most cases, the main problem is if the user navigates away from the screen before these async actions finish.
These could be handled within a saga to cancel further side effects when this happens.

packages/mobile/src/merchantPayment/MerchantPaymentScreen.tsx
```diff
@@ -0,0 +1,286 @@
...
+      const chainHandler = new ContractKitTransactionHandler(kit)
+      const charge = new Charge(routeParams.apiBase, routeParams.referenceId, chainHandler, false)
+      try {
+        const info = await charge.getInfo()
+        setSdkCharge(charge)
+        setAmount(divideByWei(new BigNumber(info.action.amount)).decimalPlaces(2))
+        setBusiness(info.receiver.businessData)
+      } catch (e: unknown) {
+        dispatch(showError(t('merchantPaymentSetup')))
+        navigateBack()
```
jeanregisser, line 75: There's an edge case here too if the user has already left the screen, this shouldn't be called.

packages/mobile/src/merchantPayment/MerchantPaymentScreen.tsx
jeanregisser: Yes exactly @gnardini.

Adding to this, this merchant payment flow should be probably be handled by a saga, taking into account errors and cancellation at any point.
There's an example of this with Bidali in https://github.com/valora-inc/wallet/blob/6d8719ed101ee48a9bd0548b271803c5484e1251/packages/mobile/src/fiatExchanges/saga.ts#L69-L103

Of course, I realize it's gonna be difficult to make these changes now and meet the deadline.

Looks like we need to chat directly about this to be aligned.

jeanregisser (commented):
Nice work!!

Could you update the PR description to follow our [PR template](https://github.com/valora-inc/wallet/blob/main/.github/pull_request_template.md)?
This is important for other engineers to have some context on these changes and our QA team to understand it and have some direction on how to test it.

It would help to have screenshots or a video showing it too. And links to documents/issues about it.

Thanks 🙏

packages/mobile/src/qrcode/utils.ts
nicolasbrugneaux: Good catch! Indeed I was bypassing it like that locally. I didn't know of the toggle, thanks for that! 👍

nicolasbrugneaux: And yes, the verified number is a requirement I had

packages/mobile/test/schemas.ts
nicolasbrugneaux: Good catch, looks like a bad merge from my side

packages/mobile/src/qrcode/utils.ts
nicolasbrugneaux: While I 100% agree, we need to sync with first-dag to know what link we should change it too. meanwhile I added it link this to take priority for now.

Screenshot 2022-03-08 at 16 38 19

packages/mobile/src/merchantPayment/MerchantPaymentScreen.tsx
nicolasbrugneaux: I need to DEK to be unlocked to be able to do requests with the payment SDK (including the `initCharge` so I'm doing it as early possible

nicolasbrugneaux:
Here is a video of the feature in effect 

Uploading Screen Recording 2022-03-08 at 17.01.32.mov…

packages/mobile/src/geth/GethNativeBridgeWallet.ts
```diff
@@ -99,4 +101,19 @@ export class GethNativeBridgeWallet
...
+   */
+  async signTypedData(address: string, typedData: EIP712TypedData): Promise<string> {
+    Logger.info(
+      `${TAG}@signTypedData`,
+      `Signing typed DATA: ${JSON.stringify({ address, typedData })}`
+    )
+    const signer = this.getSigner(address)
+    const { v, r, s } = await signer.signTypedData(typedData, address)
+    return ethUtil.toRpcSig(v, r, s)
+  }
```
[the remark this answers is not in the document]
nicolasbrugneaux: I'll update the comments but it's defined this way in the interface sadly

packages/mobile/src/qrcode/utils.ts
dckesler: I talked to our current partner about switching to `merchant-payment` but probably we should do it after celo connect

packages/mobile/src/merchantPayment/FeeContainer.test.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { createMockStore } from 'test/utils'
+
+describe('FeeContainer', () => {
+  it('renders correctly', () => {
+    const tree = render(
+      <Provider store={createMockStore({})}>
+        <FeeContainer amount={new BigNumber(1337)} />
+      </Provider>
+    )
+    expect(tree).toMatchSnapshot()
```
jeanregisser, line 16: This only check that it renders, but it's not actually useful.
We've decided to move away from this kind of test and favor of https://testing-library.com/ tests.

jeanregisser (approved):
Thanks for addressing the feedback 🙏 👍 
I'm accepting it now so it can be included in the release.

There are 2 main things that should be followed up on:
1. improving the testing
2. addressing the TODO about changing the `/payment` deeplink so it's not confusing for older versions of Valora which will match it

packages/mobile/src/merchantPayment/FeeContainer.test.tsx
jeanregisser: This comment is also for the other component tests that were added.

nicolasbrugneaux:
Closing in favor of #2133

PR #2002, opened 2022-02-11
feat(bank-accounts): Handle errors from ihl/plaid

jh2oman:
### Description

These were missed in [[M2] Handle errors for onExit hooks](https://app.zenhub.com/workspaces/acquisition-squad-sprint-board-6010683afabec1001a090887/issues/valora-inc/wallet/1447) since the BankAccounts component was still in progress when it was done. Better late than never!

### Tested

Manually triggering failed api calls

### How others should test

N/A


lisaqyli (approved):

cajubelt (approved):
🥳

PR #432, opened 2021-06-01
Various verification fixes

gnardini:
### Description

Changes in this PR (some are further explained with a PR comment in the relevant place):
- Only run SMS retrieval on Android
- If user was already verified, skip verification (it was getting stuck in the loading screen).
- If user runs out of Komenci Quota, go back to `VerificationEducationScreen` to redo recaptcha and get more.
- When resending messages, resend only the non completed ones and cancel the previous code listeners that were waiting for codes to be input, otherwise they are processed twice and one of the tries fails, which cancels the whole verification flow.
- Fix an issue with locking: We were not acquiring the lock correctly, so we got some errors due to race conditions sometimes. See PR comment for details.
- Implement lock in `receiveAttestationCodeSaga` to avoid UI glitch if many codes are paster **very** quickly together.
- Catch all sagas with an error handler that fires the fail saga if verification fails (it was getting stuck instead of showing the error modal in those cases).

### Other changes

N/A

### Tested

Manually many times. Some edge cases I tested: 
- I put some sleeps in the code to test the locks.
- I wrote code to input the three codes at the same time.
- I requested a bunch of attestations to run out of quota.
- I resent messages a few times within the same session and across different app runs.
- I input the codes in random order in every test.

### How others should test

The whole verification flow needs to be tested including all edge cases that one can think of.

### Related issues

- Fixes #377

### Backwards compatibility

N/A

Issue #381, opened 2021-05-18
“Confirmation” error is shown when a user try to verify the phone number after changing the number


ValoraQA:
**Frequency:** 100%
**Repro on Devices:** iPhone 6+ (12.4.5), Vivo V11 (10), one plus 6T (10)
**Repro on build version:**  Android internal build V1.14.1 (1004294351) , Test Flight build V1.14.1(59), Play Store & app store builds V1.13.1

**Repro Steps:**

1. Launch the application and complete education
2. Tap on create new account 
3. add name 
4. Enter phone number 
5. Tap on start button 
6. Make sure user should receive 3 attestation codes 
7. Get back to phone number page 
8. Change the phone number 
9. & try to verify phone number 
10. 
**Current Behavior:** Confirmation error is shown when user try to verify phone number after changing phone number.
**Expected Behavior:** User should be able to continue & complete verification flow if user change the phone number.
**Impact;** User get blocked to complete verification flow
**Attachment:** [ConfirmationErrorAfterchangingPhoneNumber.mp4](https://drive.google.com/file/d/1BDkXlEq1Eo9MUVypzJqtsC61tFRikDYe/view?usp=sharing )
[Logs](https://drive.google.com/file/d/1Kc8RLk05vBXE7_L5Gv53cshBhhpA_ZYt/view?usp=sharing)

MuckT:
https://valora.testquality.com requirement [R20](https://valora.testquality.com/project/12128/testcase/282097/requirement/11274). Do not remove or edit.

Celoqa:
Hi @MuckT thank you so much for query clarification.
- We are converting into a bug.

Celoqa:
Hi @MuckT  We have verified this issue on latest android internal build v1.14.2 & Test Flight Build V1.14.2 and observe the followings:
- User is able to complete the verification flow after changing the phone number.

MuckT:
https://valora.testquality.com defect [D9](https://valora.testquality.com/project/12128/result/2404491/defect/7794). Do not remove or edit.

Lss-Ankit:
Hi @jeanregisser @MuckT We have verified this issue on latest Android internal build V1.14.3 & Testflight build V1.14.3 (62) and found issue is again repro.
Repro steps: 
- Launch the app
-  Walkthrough tutorials 
-  Tap on create new account
- Add name & Photo and tap on next 
- Setup valid pin
- Enter valid phone number 
- Tap on start button & Complete captcha 
[...]

PR #2003, opened 2022-02-11
Fixes in the verification flow

gnardini:
### Description

Some context:
On May 2021 we had a huge issue with verification. There was a huge refactor that brought a bunch of issues that we ended up reverting. _Before_ reverting though we fixed a lot of bugs related to it. In the middle of those bugfixes there were some other bugfixes that were pre-existing! However, when we reverted we reverted everything so we lost those fixes.

Here's the PR from that time: https://github.com/valora-inc/wallet/pull/432

In this PR I'm bringing back two bugfixes that will help with the verification issues.

- The first one is calling a lock we were already using properly. We were using a lock in a section on the code but we were not actually `awaiting` it so it was useless. There was a delay there to "fix" the issue but it didn't really fix it. The lock should now work properly.
- The second one is adding another lock in the code that receives the verification codes. Without it, the UI glitches out a bunch and there may be other issues as well, since all the codes pasted try to get inserted in the same spot. This happens on Android where codes are auto-pasted. With this change, all codes should be pasted in a different position.

This fixes issues for new users, but we also need to run a migration to fix the redux store state for existing users that show up as verified when they are not.

### Other changes

N/A

### Tested

Manually on both OSes. I also modified the code a bit to send the three codes instantly to make sure they are handled correctly.

### How others should test

Try verifying in Android and iOS several times. Try all the edge cases you can think of, including but not limited to:
- Copying the same code several times.
- Pasting the codes very quickly.
- Asking codes to be resent.

### Related issues

- Part of #1906
This doesn't fix the issue completely, it fixes it for new users but we still need to fix the internal state for existing ones.

### Backwards compatibility

N/A


packages/mobile/src/identity/verification.ts
```diff
@@ -956,12 +968,8 @@ function* completeAttestation(
     `Completing code (${code.shortCode} ${codePosition}) for issuer: ${code.issuer}`
   )
 
-  // Make each concurrent completion attempt wait a sec for where they are relative to other codes
-  // to ensure `processingInputCode` has enough time to properly gate the tx. 0-index code
-  // will have 0 delay, 1-index code will have 1 sec delay, etc.
   if (shouldUseKomenci) {
-    yield delay(codePosition * 5000)
-    yield inputAttestationCodeLock.acquireAsync()
+    yield call([inputAttestationCodeLock, inputAttestationCodeLock.acquireAsync])
```
gnardini, line 972: The delay was necessary because the lock wasn't working as intended because it was not being called correctly. redux-saga can be non-intuitive at times :/

packages/mobile/src/identity/verification.ts
```diff
@@ -666,24 +667,29 @@ export function attestationCodeReceiver(
       action.index
     )
 
-    const attestationInputStatus = yield select(attestationInputStatusSelector)
```
gnardini, line 669: This block of code wasn't actually changed, I just moved the `try` further up the flow and added a lock

jeanregisser (approved):
Thanks so much! 🥇 💯 

Hopefully in the next iteration, we'll have better test coverage.

ValoraQA:
Hey @MuckT, We have verified the above issue on Latest Android Internal Release Build & Test flight Release Build V1.27.0 and updated status in Test rails.

- User is able to complete phone number verification by entering 3-confirmation codes. 
- But on changing phone and starting verification again "Confirmation error" popup appears. [Bug link](https://github.com/valora-inc/wallet/issues/381)

**Test rail links:**
**Android 10:** https://valoraapp.testrail.io/index.php?/tests/view/12913

**iOS 14:** https://valoraapp.testrail.io/index.php?/tests/view/12975

**iOS 15:** https://valoraapp.testrail.io/index.php?/tests/view/13006

**Devices:** iPhone 13 mini(15.1.1), iPhone 12 (14.7.1), Google Pixel 2XL (11.0)
**cc:** @gnardini 
Thanks..!!

PR #2012, opened 2022-02-15
Increase "deep link" capability for push notifications

### LINKED ISSUE #1682: Increase "deep link" capability for push notifications
### Description
At the moment we can only direct a user to the cash in screen and home screen through push notifications via CleverTap. We need to expand our ability to directly push users into different parts of the app to help them through their user journey and activate them more effectively.

For example, for Supercharge we'd like to push them to:
- Verification Screen
- CELO Screen
- Help Screen

Places we'd like to deep link to for other retention campaigns (not limited to this list):
- Language screen
- dApp marketplace (once live)
- Recovery Phrase

ENG:
- Talk to Jean (the clevertap bindings are annoying)
- Should be possible to deeplink into _most_ screens based on some set of parameters from Clevertap (i.e., we don't need to implement specific deeplink support for every screen)

AC:

Able to deeplink to:
- [ ] Verification Screen
- [ ] CELO Screen
- [ ] Help Screen
- [ ] Language screen
- [ ] dApp marketplace (once live)
- [ ] Recovery Phrase

### Related Issues
Slack thread [here](https://valora-app.slack.com/archives/C025Q9QG8JG/p1639689901163900)
[Notion Ticket](https://www.notion.so/valora-inc/More-deep-link-ability-in-push-notifications-9ca983a5aaec4e8388c895fb21ff189b)

MuckT:
### Description

Adds the ability to handle deeplinks from CleverTap on the JavaScript side of the app.

### Other changes

Updated the podfile to include CleverTap React Native.

### Tested

Tested locally on a physical iOS device. Android WIP.

#### iOS
1. Enable analytics in Dev Mode by editing `packages/mobile/src/analytics/ValoraAnalytics.ts`
```
-    return !__DEV__ && store.getState().app.analyticsEnabled
+    return store.getState().app.analyticsEnabled
```
2. Build Xcode with Alfajores targeting the physical device; Product > Destination.
3. Login to CleverTap and find your device by Account Address.
4. Send a notification with the deeplink field populated e.g. `celo://wallet/openScreen?screen=GoldEducation` with Valora in the following states: Open, Closed and background.

### Related issues

- Fixes #1682

### Backwards compatibility

Yes


packages/mobile/src/app/App.tsx
```diff
@@ -66,6 +67,23 @@ export class App extends React.Component<Props> {
   async componentDidMount() {
     await ValoraAnalytics.init()
 
+    // Handles opening Clevertap deeplinks when app is closed / in background
+    CleverTap.getInitialUrl(async (err: any, url: any) => {
+      if (err) {
+        Logger.debug('App', err)
```
gnardini, line 73: Should this be `Logger.error`? Maybe a more descriptive message? :P

packages/mobile/src/app/App.tsx
```diff
@@ -66,6 +67,23 @@ export class App extends React.Component<Props> {
...
+      } else if (url) {
+        await this.handleOpenURL({ url }, true)
+      }
+    })
+
+    // Handles opening Clevertap deeplinks when app is open
+    CleverTap.addListener('CleverTapPushNotificationClicked', async (event: any) => {
+      if (event.customExtras['wzrk_dl']) {
```
gnardini, line 81: This looks quite strange to me 😅  What is `wzrk_dl`?

packages/mobile/src/app/App.tsx
```diff
@@ -66,6 +67,23 @@ export class App extends React.Component<Props> {
...
+        const url = event.customExtras['wzrk_dl']
+        await this.handleCleverTapUrl(url, true)
```
gnardini, line 83: Can't we just do the same as in `getInitialUrl` and call `await this.handleOpenURL({ url }, true)` instead of creating a new function?

gnardini (commented):
Oh, wow, I expected this to require more changes, nice that CleverTap has a React Native integration :)

packages/mobile/src/app/App.tsx
MuckT: `wzrk_dl` appears to be the deeplink field provided by CleverTap.

packages/mobile/src/app/App.tsx
MuckT: I can give it another try like that. Unfortunately it didn't appear that the CleverTap event had a url field to be deconstructed.

gnardini: Yup, but this is the opposite of deconstruction, this is building an object with a single `url` field, just like you did a few lines above, unless I'm reading this wrong 🤔

packages/mobile/src/app/App.tsx
gnardini: Wow, that's a weird choice from them. Is this documented or did you figure it out? I just wonder if it's going to change across different installs or something :/ 

In any case I think we should probably move it to a constant, it looks a bit magic-y like this

MuckT: I haven't seen it change across installs. Their documentation calls it out [here](https://docs.clevertap.com/docs/faq#q-why-am-i-receiving-blank-push-notifications).

gnardini: Oh, I guess `dl` stands for `deep link`, I wonder what `wzrk` stands for 😅

packages/mobile/src/app/App.tsx
MuckT: Fixing in https://github.com/valora-inc/wallet/pull/2012/commits/c5ad74ee98e6a3fa0623afd73fad0d892ed3700c.

packages/mobile/src/app/App.tsx
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2012/commits/c5ad74ee98e6a3fa0623afd73fad0d892ed3700c!

packages/mobile/src/app/App.tsx
MuckT: I have no idea what `wzrk` stands for, but they use it a bunch of places across different SDKs. 🤷

MuckT: > In any case I think we should probably move it to a constant

Something like:

```JavaScript
const url = event?.customExtras['wzrk_dl']
if (url) {
  await this.handleOpenURL({ url }, true)
}
```

packages/mobile/ios/Podfile
```diff
@@ -28,7 +28,7 @@ use_frameworks!
 
 target "celo" do
   pod "CeloBlockchain", :path => "../../../node_modules/@celo/client/CeloBlockchain.podspec", :modular_headers => true
-
+  pod 'clevertap-react-native', :path => '../../../node_modules/clevertap-react-native'
```
jeanregisser, line 31: This shouldn't be needed with autolink.

packages/mobile/src/app/App.tsx
jeanregisser: Yes this seems good.

packages/mobile/src/app/App.tsx
```diff
@@ -66,6 +67,23 @@ export class App extends React.Component<Props> {
   async componentDidMount() {
     await ValoraAnalytics.init()
 
+    // Handles opening Clevertap deeplinks when app is closed / in background
+    CleverTap.getInitialUrl(async (err: any, url: any) => {
+      if (err) {
+        Logger.error('App CleverTap Deeplink on Load', err)
```
jeanregisser, line 73: The error argument is the 3rd one.
```suggestion
        Logger.error('App/componentDidMount', 'App CleverTap Deeplink on Load', err)
```

packages/mobile/src/app/App.tsx
```diff
@@ -66,6 +67,23 @@ export class App extends React.Component<Props> {
   async componentDidMount() {
     await ValoraAnalytics.init()
 
+    // Handles opening Clevertap deeplinks when app is closed / in background
+    CleverTap.getInitialUrl(async (err: any, url: any) => {
```
jeanregisser, line 71: Looks like CleverTap has types for this: https://github.com/CleverTap/clevertap-react-native/blob/master/index.d.ts#L33
```suggestion
    CleverTap.getInitialUrl(async (err, url) => {
```

jeanregisser (commented):
Great that it was way simpler than anticipated 😄

packages/mobile/ios/Podfile
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2012/commits/d2af69ea54a10a7425e5acbed011e464a7d89a5b!

packages/mobile/src/app/App.tsx
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2012/commits/f5ea9f621cfeb3bf2541fccf39c03e29335f24e1!

packages/mobile/src/app/App.tsx
MuckT: I was able to get the types for url; however, err did not appear to pickup the types from CleverTap. Fixed in https://github.com/valora-inc/wallet/pull/2012/commits/a19fa197a391f4f86e1a86ea22d3209216401370.

jeanregisser (approved):
🐑 🚢 :shipit:

ValoraQA:

ValoraQA:
Hey @MuckT we have checked the above task on latest iOS Alfajores build V:1.27.0 and Latest Android Internal Release Build V:.1.27.0 and observed that after performing deep link it is redirecting to homescreen

Devices : OnePlus 7t(11.0) , iPhone 12(14.7.1) , Google Pixel(11.0)

Thanks ..!!

PR #2013, opened 2022-02-15
Run migration to check verification state

### LINKED ISSUE #1906: Fix the bug where Valora shows that user is verified when they are actually not verified.

gnardini:
### Description

Re-running an old migration to the verification state that we ran to solve a similar issue in the past.

### Other changes

N/A

### Tested

I set the internal state to different situations and ran the migration to see it works as expected.

### How others should test

No real way to test this unfortunately without access to the code.

### Related issues

- Fixes #1906

### Backwards compatibility

N/A

### DIFF
--- a/packages/mobile/src/app/verificationMigration.test.ts
+++ b/packages/mobile/src/app/verificationMigration.test.ts
@@ -1,6 +1,7 @@
 import { AttestationStat } from '@celo/contractkit/lib/wrappers/Attestations'
 import { expectSaga } from 'redux-saga-test-plan'
 import { call, select } from 'redux-saga-test-plan/matchers'
+import { StaticProvider } from 'redux-saga-test-plan/providers'
 import { verificationMigrationRan } from 'src/app/actions'
 import { numberVerifiedSelector, ranVerificationMigrationSelector } from 'src/app/selectors'
 import { runVerificationMigration } from 'src/app/verificationMigration'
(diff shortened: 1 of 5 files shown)


packages/mobile/src/app/verificationMigration.test.ts
```diff
@@ -13,142 +14,133 @@ import { mockAccount, mockE164Number, mockE164NumberPepper } from 'test/values'
...
 
+const mockMTW = mockAccount
+
 describe('verificationMigration', () => {
   beforeEach(async () => {
     jest.clearAllMocks()
   })
 
-  it("doesn't change anything if mtwAddress is present and verified", async () => {
+  async function commonProvides({
```
gnardini, line 24: I didn't actually change the tests (or didn't intend to) they were just hard to read because of all the repeated code so I created this function so that each test only passes the relevant fields

packages/mobile/src/app/verificationMigration.ts
```diff
@@ -26,9 +23,20 @@ function* isAddressVerified(address: string | null) {
       TAG,
       `No number or pepper found - Number present? ${!!e164Number} - Pepper present? ${!!pepper}`
     )
+    return null
+  }
+  return getPhoneHash(e164Number, pepper)
+}
+
+function* isAddressVerified(address: string | null) {
```
gnardini, line 31: The diff is a bit annoying to read. I started doing more changes but ended up reverting all of them, in the end all I did is moving some stuff to separate functions: `fetchPhoneIdentifier` and `fetchMtwAddress`. No logic should have changed

packages/mobile/src/redux/migrations.ts
```diff
@@ -416,4 +416,11 @@ export const migrations = {
       hasLinkedBankAccount: false,
     },
   }),
+  33: (state: any) => ({
+    ...state,
+    app: {
+      ...state.app,
+      ranVerificationMigrationAt: null,
```
gnardini, line 423: This is the actual relevant change. By setting the field as `null` the verification migration will be run again

packages/mobile/src/app/verificationMigration.ts
gnardini: If anyone has any concerns with this I'm happy to revert, but I do think after the changes the file is better organized, it's just that this diff is a bit annoying to read.

jeanregisser (approved):
Looks great! 👍

ValoraQA:

PR #1896, opened 2022-02-01
plaid oauth handling - wallet part

lisaqyli:
### Description
This PR handles deeplink from plaid linking bank account success flow. After user successfully linked their bank account through plaid, some of the bank accounts requires user to go through OAuth flow. https://valoraapp.com/plaid-oauth-redirect is provided as the link to redirect to after oauth success. 

After the deeplink is opened from the app, we will navigate user to bankAccounts screen

### Tested

pending e2e test

### Related issues

- Fixes #1444

PR #2035, opened 2022-02-22
Fail CI if lock files are not up-to-date

jeanregisser:
### Description

We unexpectedly committed changes to `Podfile.lock` that were incorrect in https://github.com/valora-inc/wallet/pull/1896

This now makes sure `yarn.lock` (well, this one was already checked via the [`check.yml` workflow](https://github.com/valora-inc/wallet/blob/545f1974cb663b13eca4972968d0b62d503ef689/.github/workflows/check.yml)), `Gemfile.lock` and `Podfile.lock` are up-to-date.

It does that by checking there are no git differences after running the corresponding `install` commands.

I also considered using tool specific command line options (like [`yarn install --frozen-lockfiles`](https://classic.yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-frozen-lockfile)) to fail during the install phase if there would be differences with the lock files.
However for `CocoaPods` it also rewrites the Xcode project, so wanted to be extra sure there were no other file differences.
Plus the diff shows the actual difference.

### Tested

CI passes

### How others should test

N/A

### Related issues

- Discussed on [Slack](https://valora-app.slack.com/archives/CL7BVQPHB/p1645523289179829)

### Backwards compatibility

Yes

### DIFF
--- a/.github/workflows/e2e-android.yml
+++ b/.github/workflows/e2e-android.yml
@@ -38,6 +38,8 @@ jobs:
         run: mkdir -p ~/.android/avd/Pixel_API_29_AOSP_x86_64.avd/ && cp packages/mobile/e2e/conf/avd_conf.ini ~/.android/avd/Pixel_API_29_AOSP_x86_64.avd/config.ini
       - name: Install package dependencies
         run: yarn
+      - name: Fail if someone forgot to commit "yarn.lock"
+        run: git diff --exit-code
       - name: Build mobile dependencies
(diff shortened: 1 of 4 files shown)


gnardini (approved):

ValoraQA:
Hey @MuckT JFYI: We are skipping this task as in "How others should test" section it is mentioned N/A.

CC: @jeanregisser 
thanks.!

PR #2038, opened 2022-02-22
chore(FiatConnect): added fiat connect types

### LINKED ISSUE #1993: [FC UI] Generate FC typings
The work to create the FC API wrapper module (#1978) _should_ be orthogonal to the actual UI development, but if #1978 includes adding typings for FC responses/schemas/etc. to the wallet, then the UI work is somewhat blocked, since UI components will require certain FC-specific information to be passed as props.

This ticket is for generating and adding typings for FC objects/entities/schemas to the wallet repo, such that #1978 and the UI tickets can be worked on concurrently. For entities such as KYC schemas, account types, and account schemas (which are not currently defined by the spec), we should come up with sensible mock defaults and include them in the typings.

AC:
* Sensible mock defaults for currently undefined FC entities
* FC typings added to wallet repo

cajubelt:
### Description

Added types that will be useful for integrating with Fiat Connect

### Other changes

na

### Tested

none (just types, not used anywhere yet)

### How others should test

None required

### Related issues

- Fixes #1993 

### Backwards compatibility

na

### DIFF
--- a/packages/mobile/src/fiat-connect/types.ts
+++ b/packages/mobile/src/fiat-connect/types.ts
@@ -1,4 +1,4 @@
-interface Quote {
+export interface Quote {
   fiatType: FiatType
   cryptoType: string
   fiatAmount?: number
@@ -6,29 +6,29 @@ interface Quote {
   region: string
 }
 
-enum FiatType {
+export enum FiatType {
   USD = 'USD',
   EUR = 'EUR',
 }
 
-enum CryptoType {
+export enum CryptoType {
   cUSD = 'cUSD',
   cEUR = 'cEUR',
   CELO = 'CELO',
 }
 
-enum KycSchema {
+export enum KycSchema {
   NameAndAddress = 'NameAndAddress',
 }
 
-interface NameAndAddressKYCSchema {
+export interface NameAndAddressKYCSchema {
   // NOTE: this is a MOCK schema (not a real one in the spec!)
   firstName: string
   lastName: string
   address: PostalAddress
 }
 
-interface PostalAddress {
+export interface PostalAddress {
   address1: string
   address2?: string
   city: string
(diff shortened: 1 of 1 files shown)


jh2oman:
Missing [KycStatusEnum](https://github.com/fiatconnect/specification/blob/632099c0d7d547caa70e7e3827cca7df27dbe8da/fiatconnect-api.md#711-kycstatusenum)

packages/mobile/src/fiat-connect/types.ts
```diff
@@ -0,0 +1,94 @@
...
+  TransferFailed = 'TransferFailed',
+}
+
+export enum WebhookEventType {
+  // NOTE: if we use these we should update the spec (which has 'webhook' prefix for all of these)
+  KYCStatusEvent = 'KYCStatusEvent',
+  TransferInStatusEvent = 'TransferInStatusEvent',
+  TransferOutStatusEvent = 'TransferOutStatusEvent',
+}
+
```
jh2oman, line 79: When I thought about this ticket, I thought it would also include all of the Request Body, error objects, and response objects for the different endpoints such as `QuoteRequest`, `QuoteResponse`, `QuoteErrorResponse`, `TransferRequest`, `TransferResponse`, etc.

What do others think? @cajubelt @lisaqyli @jophish 

A Pro is that it is very useful to have those explicitly typed so that client libraries and partners can implement the endpoints/functions with confidence. A con is that its a bit of extra work to maintain while the endpoints are still in flux.

jophish: Hmm... this is a good point. The SDK will need to expose `xxxRequest` types to allow clients to pass in arguments to API methods, and `xxxResponse` types to return values... the API will also need exactly these types in order to implement its endpoints. I would say +1 on including those types here.

cajubelt:
@jophish is taking this so he can include it in the skeleton API he is working on

Issue #881, opened 2021-07-14
“Having trouble?” popup does not open on shaking the phone while “Invite a friend” popup is open.


ValoraQA:
**Frequency:** 100%
**Repro on build version:** Test Flight Build V1.16.2(68)
**Repro on devices:** iPhone SE 1st Generation (14.6),  iPhone 6+ (12.4.5)
**Does not repro version** Android Internal Build V1.16.0
**Does not repro devices** Redmi Note 8 (10.0), Google Pixel 4a (11.0)
**Pre-condition:** Make sure user has installed and login into the application
**Repro Steps:**
Launch the application 
Open the Hamburger menu.
Tap on “Invite” option in hamburger menu.
Shake the phone when “Invite a friend” popup is open & Observe the bug
**Current Behavior:**  “Having trouble?” popup does not open on shaking the phone.
**Expected Behavior:**  “Having trouble?” popup must open on shaking the phone while “Invite a friend” popup is open

**Impact:** Bad user experience as user is blocked to open contact screen.
**Investigation:** The above issue is not reproducible in Android platform.
**Attachment:** 
 [iOS Having Trouble Popup issue.mp4](https://drive.google.com/file/d/1peq6zSg91q22YwafFRVzeflHZc61EFCb/view?usp=sharing)

[Android Behaviour.mp4](https://drive.google.com/file/d/1vi9pU8p1yD4hJNnz_pBYZKefOHDT6dcg/view?usp=sharing)

MuckT:
Based on the feedback from the [bugapalooza](https://docs.google.com/document/d/1_E3fc2Gxyvefhe419_8WEUNnh8Xx2giMHyBYofokzVE/edit#) appears to reproduce with with the shake to support feature when any pop-up is modal is currently on the screen. @isabellewei can you take a look?

ValoraQA:
Hey @MuckT, This feature has been removed as per [PR](https://github.com/valora-inc/wallet/pull/1045) link, So we are closing this issue.
Thanks..!!

PR #2039, opened 2022-02-22
Kathy/recent dapps

### LINKED ISSUE #1745: M2: dapps marketplace - recent dapps on the home screen
[PRD](https://www.notion.so/valora-inc/PRD-in-app-dApps-Services-marketplace-c88abb7524494781b2ba0c5c48a0f96c#ed4876f58e7644afacd0ef6873802115)
[Design](https://www.figma.com/file/zdK8XvxNXvTjxnM5V33pHw/Dapps-Page?node-id=1356%3A35142)
[Experiment proposal
](https://www.notion.so/valora-inc/Dapp-browser-M2-recently-used-dapps-on-home-screen-c06ab4ff178b44a6ac9d5bbf8d616157)
There is a widget on the home screen that shows the list of 3 recently used dApps by this user.

The users who haven't interacted with one of the prelisted dApps, won't see them here.

Only dApps from our safelist can be displayed here. Meaning that if the user recently interacted with the dApp that is not part of Valora's app directory, it won't be displayed here.

If user taps on the app's icon the user will see bottom sheet notifying that this is an external service ([see Ledger Live example](https://www.notion.so/PRD-in-app-dApps-Services-marketplace-c88abb7524494781b2ba0c5c48a0f96c)). **Once they accept it, the dApp will be opened in the external browser**

The component will show up only for users who have interacted with dapps.

If user has interacted with only one dapp, only one icon will be displayed in the row.

ENG:
- When a user taps on a dApp (from the market place) it's added to a 'recently used' list on redux

AC:
- [ ] Use taps dapp in the **dapp market place** it's added to the front of the list OR user signs a transaction for all dapps OR connects the wallet for all dapps
- [ ] List of max of N items
- [ ] Eviction from the list is LRU
- [ ] Users can see this list in the home screen
- [ ] Add RemoteConfig variable for show/hiding the frequently used dapps
- [ ] Analytics TBD

kathaypacific:
### Description

Add recently used dapps to the home screen.

[Screenshot here]

In this PR:
- add remote config to show/hide recently used dapps
- store recently used dapps in redux (update this list every time the user opens a dapp from Valora)
- move some dapp logic around to be reused
- TODO: unit tests

### Other changes

N/A

### Tested

Open dapps from dapps explorer screen, see that the most recently opened dapps are displayed on the home screen.

Only the 4 most recently used dapps are displayed.

No duplicate dapps should be displayed.

The most recently used dapp should appear on the left.

### How others should test

As above

### Related issues

- Fixes #1745 

### Backwards compatibility

Yes

### DIFF
--- a/packages/mobile/src/config.ts
+++ b/packages/mobile/src/config.ts
@@ -172,4 +172,4 @@ export const DEFAULT_APP_LANGUAGE = 'en-US'
 export const DEFAULT_SENTRY_TRACES_SAMPLE_RATE = 0.2
 export const DEFAULT_SENTRY_NETWORK_ERRORS = ['network request failed']
 
-export const NUM_RECENT_DAPPS = 3
+export const NUM_RECENT_DAPPS = 4
--- a/packages/mobile/src/config.ts
+++ b/packages/mobile/src/config.ts
@@ -171,3 +171,5 @@ export const DEFAULT_APP_LANGUAGE = 'en-US'
 
 export const DEFAULT_SENTRY_TRACES_SAMPLE_RATE = 0.2
 export const DEFAULT_SENTRY_NETWORK_ERRORS = ['network request failed']
+
+export const NUM_RECENT_DAPPS = 3
--- a/packages/mobile/src/dappsExplorer/useOpenDapp.ts
+++ b/packages/mobile/src/dappsExplorer/useOpenDapp.ts
@@ -1,0 +1,81 @@
+import { useState } from 'react'
+import { useDispatch } from 'react-redux'
+import { DappExplorerEvents } from 'src/analytics/Events'
+import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
+import { openUrl, recentDappSelected } from 'src/app/actions'
+import { Dapp } from 'src/app/types'
+import { isDeepLink } from 'src/utils/linking'
+import Logger from 'src/utils/Logger'
+
+const TAG = 'DApps'
+
+// Open the dapp if deep linked, or require confirmation to open the dapp
+const useOpenDapp = () => {
+  const [showOpenDappConfirmation, setShowOpenDappConfirmation] = useState(false)
+  const [selectedDapp, setSelectedDapp] = useState<Dapp | null>(null)
+  const dispatch = useDispatch()
+
+  const onCancelOpenDapp = () => {
+    setShowOpenDappConfirmation(false)
+    if (selectedDapp) {
+      ValoraAnalytics.track(DappExplorerEvents.dapp_bottom_sheet_dismiss, {
+        categoryId: selectedDapp.categoryId,
+        dappId: selectedDapp.id,
+        dappName: selectedDapp.name,
+        horizontalPosition: 0,
+        section: selectedDapp.isFeatured ? 'featured' : 'all',
+      })
+    }
+  }
+
+  const openDapp = (dapp: Dapp) => {
+    ValoraAnalytics.track(DappExplorerEvents.dapp_open, {
+      categoryId: dapp.categoryId,
+      dappId: dapp.id,
+      dappName: dapp.name,
+      section: dapp.isFeatured ? 'featured' : 'all',
+      horizontalPosition: 0,
+    })
+    dispatch(recentDappSelected(dapp))
+    dispatch(openUrl(dapp.dappUrl, true, true))
+  }
+
+  const onOpenDapp = () => {
+    if (!selectedDapp) {
+      // Should never happen
+      Logger.error(TAG, 'Internal error. There was no dapp selected')
+      return
+    }
+    openDapp(selectedDapp)
+    setShowOpenDappConfirmation(false)
+  }
+
+  const onSelectDapp = (dapp: Dapp) => {
+    const dappEventProps = {
+      categoryId: dapp.categoryId,
+      dappId: dapp.id,
+      dappName: dapp.name,
+      horizontalPosition: 0,
+      section: dapp.isFeatured ? 'featured' : 'all',
+    }
+    ValoraAnalytics.track(DappExplorerEvents.dapp_select, dappEventProps)
+
+    if (!isDeepLink(dapp.dappUrl)) {
+      setSelectedDapp(dapp)
+      setShowOpenDappConfirmation(true)
+      ValoraAnalytics.track(DappExplorerEvents.dapp_bottom_sheet_open, dappEventProps)
+    } else {
+      openDapp(dapp)
+    }
+  }
+
+  return {
+    onSelectDapp,
+    onOpenDapp,
+    onCancelOpenDapp,
+    showOpenDappConfirmation,
+    selectedDapp,
+  }
+}
+
+export default useOpenDapp
--- a/packages/mobile/src/firebase/remoteConfigValuesDefaults.e2e.ts
+++ b/packages/mobile/src/firebase/remoteConfigValuesDefaults.e2e.ts
@@ -45,4 +45,5 @@ export const REMOTE_CONFIG_VALUES_DEFAULTS: Omit<
   dappListApiUrl:
     'https://raw.githubusercontent.com/valora-inc/dapp-list/main/translations/valora-dapp-list-base.json',
   superchargeButtonType: SuperchargeButtonType.PillRewards,
+  recentDappsEnabled: false,
 }
--- a/packages/mobile/src/home/RecentlyUsedDapps.tsx
+++ b/packages/mobile/src/home/RecentlyUsedDapps.tsx
@@ -15,12 +15,17 @@ import {
   View,
 } from 'react-native'
 import { recentDappsSelector } from 'src/app/selectors'
+import { Dapp } from 'src/app/types'
 import ProgressArrow from 'src/icons/ProgressArrow'
 import { navigate } from 'src/navigator/NavigationService'
 import { Screens } from 'src/navigator/Screens'
 import useSelector from 'src/redux/useSelector'
 
-function RecentlyUsedDapps() {
+interface Props {
+  onSelectDapp(dapp: Dapp): void
+}
+
+function RecentlyUsedDapps({ onSelectDapp }: Props) {
   const recentlyUsedDapps = useSelector(recentDappsSelector)
   const [currentIndex, setCurrentIndex] = useState(0)
   const { t } = useTranslation()
@@ -48,7 +53,7 @@ function RecentlyUsedDapps() {
         <Text style={styles.title}>{t('recentlyUsedDapps')}</Text>
         <TouchableOpacity style={styles.row} onPress={onPressAllDapps} testID="AllDapps">
           <Text style={styles.allDapps}>{t('allDapps')}</Text>
-          <ProgressArrow height={10} color={Colors.greenUI} />
+          <ProgressArrow color={Colors.greenUI} />
         </TouchableOpacity>
       </View>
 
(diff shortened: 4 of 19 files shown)


packages/mobile/src/home/WalletHome.tsx
```diff
@@ -160,6 +180,14 @@ function WalletHome() {
       />
       <SendOrRequestBar />
       {shouldShowCashInBottomSheet() && <CashInBottomSheet />}
+      {showRecentDapps && (
```
kathaypacific, line 183: any ideas on how else i could do this? ideally i want to keep the dapps logic together (i.e. have this bottom sheet inside the `RecentlyUsedDapps` component) but I could not get the bottom sheet to be on the bottom of the screen with the way flex styles work since `RecentlyUsedDapps` is inside the section list 🥲

packages/mobile/src/dappsExplorer/useOpenDapp.ts
```diff
@@ -0,0 +1,81 @@
...
+
+const TAG = 'DApps'
+
+// Open the dapp if deep linked, or require confirmation to open the dapp
+const useOpenDapp = () => {
+  const [showOpenDappConfirmation, setShowOpenDappConfirmation] = useState(false)
+  const [selectedDapp, setSelectedDapp] = useState<Dapp | null>(null)
+  const dispatch = useDispatch()
+
+  const onCancelOpenDapp = () => {
```
kathaypacific, line 18: these functions were mostly copy + paste from dapps explorer screen

packages/mobile/src/home/WalletHome.tsx
gnardini: Yup, I think it's fine. One possible nit is:

Right now, the `useOpenDapp` hook and `DAppsBottomSheet` components are implicitly connected since the component uses the state from the hook so this component (and `DAppsExplorerScreen`) is basically acting as a middle man.

Something we _could_ do (not really necessary, just thinking out load) is to make `useOpenDapp` return the component and just render it here. In that case, `useOpenDapp` would just return the existing `onSelectDapp` function and a new `bottomSheetDialog` component. And then here we would just render that `bottomSheetDialog`. That way there are less props being returned and passed to the bottom sheet here.

Again, this is very nit-picky, I think the current solution is perfectly fine 😄

packages/mobile/src/home/RecentlyUsedDapps.tsx
```diff
@@ -0,0 +1,137 @@
...
+}
+
+const DAPP_ICON_SIZE = 68
+
+function RecentlyUsedDapps({ onSelectDapp }: Props) {
+  const recentlyUsedDapps = useSelector(recentDappsSelector)
+  const [currentIndex, setCurrentIndex] = useState(0)
+  const { t } = useTranslation()
+
+  const handleScroll = (event: { nativeEvent: NativeScrollEvent }) => {
```
gnardini, line 35: I'm not sure I understand the purpose of this, why are we handling the scroll like this?

packages/mobile/src/home/RecentlyUsedDapps.tsx
```diff
@@ -0,0 +1,137 @@
...
+    const nextIndex = Math.round(event.nativeEvent.contentOffset.x / variables.width)
+    if (nextIndex === currentIndex) {
+      return
+    }
+
+    setCurrentIndex(nextIndex)
+  }
+
+  const onPressAllDapps = () => {
+    navigate(Screens.DAppsExplorerScreen)
```
gnardini, line 45: Can you check with Gowtham if we want any analytics for this (or other parts)?

packages/mobile/src/home/RecentlyUsedDapps.tsx
```diff
@@ -0,0 +1,137 @@
...
+      <ScrollView
+        horizontal={true}
+        showsHorizontalScrollIndicator={false}
+        onScroll={handleScroll}
+        testID="RecentlyUsedDapps/ScrollContainer"
+      >
+        {recentlyUsedDapps.map((recentlyUsedDapp, index) => (
+          <TouchableOpacity
+            key={recentlyUsedDapp.id}
+            onPress={() => onSelectDapp(recentlyUsedDapp)}
```
gnardini, line 71: We might want a special analytics event if the dapp is opened from here :)

packages/mobile/src/home/WalletHome.tsx
kathaypacific: i was actually thinking about this too, but didn't do it because it didn't feel "react-y" to return a component from a hook. it felt very similar to get props from a hook then pass them to a component, vs get component from hook and put it in the render function.

but now that you mentioned it, i tried it and it does save exposing a bunch of functions/state for opening the dapps bottom sheet. and i don't really know why i feel like it is "un-react" to return a component from a hook, i couldn't find a good reason online so it's probably just fine! let me know if you like how this looks?

packages/mobile/src/home/RecentlyUsedDapps.tsx
kathaypacific: ha. there is no reason, i didn't do a good job of covering up my copy+paste job from the notification box component. this was being used to track which notification is focused but we don't need that here. i've removed!

packages/mobile/src/home/RecentlyUsedDapps.tsx
kathaypacific: yes good point! there's a separate analytics [ticket](https://app.zenhub.com/workspaces/retention-squad-sprint-board-612f2ef98ef2b800116e6043/issues/valora-inc/wallet/1965) actually, i'll put the changes in a separate PR if you're okay with that

packages/mobile/src/home/WalletHome.tsx
kathaypacific: @jeanregisser maybe you have some thoughts about this too?

i think returning components from a hook is not a well liked pattern because it is hard to keep the returned component's identity the same and there are issues with react mounting/unmounting the component unnnecessarily. i think that is not a big problem here, and i've tried to reduce the impact with useMemo but still not feeling great about it (or my previous solution of having the hook provide the props for the bottom sheet...)

jeanregisser (commented):
Great work! :+1:

Like you said, I'm also concerned by the icon styles.
We're not using circular dapps icon in the Dapps page. 
It seems we made that choice for the recent dapps to match the circular icons used for transactions items.

packages/mobile/src/home/WalletHome.tsx
jeanregisser: It looks ok for now.
But I think we should consider turning BottomSheets (and Dialogs) as screens handled by the navigator.
This gives greater flexibility to really render fullscreen and trigger them from anywhere.
Avoiding the issue seen here and others we have (see https://github.com/valora-inc/wallet/issues/881).

packages/mobile/src/config.ts
```diff
@@ -172,3 +172,5 @@ export const DEFAULT_APP_LANGUAGE = 'en-US'
 
 export const DEFAULT_SENTRY_TRACES_SAMPLE_RATE = 0.2
 export const DEFAULT_SENTRY_NETWORK_ERRORS = ['network request failed']
+
+export const NUM_RECENT_DAPPS = 4
```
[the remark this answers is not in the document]
kathaypacific: ah yes good idea, i've updated the remote config to be a number.

packages/mobile/src/dappsExplorer/useOpenDapp.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const openDapp = (dapp: Dapp) => {
+    ValoraAnalytics.track(DappExplorerEvents.dapp_open, {
+      categoryId: dapp.categoryId,
+      dappId: dapp.id,
+      dappName: dapp.name,
+      section: dapp.isFeatured ? 'featured' : 'all',
+      horizontalPosition: 0,
+    })
+    dispatch(recentDappSelected(dapp))
+    dispatch(openUrl(dapp.dappUrl, true, true))
```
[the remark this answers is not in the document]
kathaypacific: yeah good point. i've moved the open url part into a saga :)

packages/mobile/src/firebase/remoteConfigValuesDefaults.e2e.ts
```diff
@@ -49,4 +49,5 @@ export const REMOTE_CONFIG_VALUES_DEFAULTS: Omit<
   dappListApiUrl:
     'https://raw.githubusercontent.com/valora-inc/dapp-list/main/translations/valora-dapp-list-base.json',
   superchargeButtonType: SuperchargeButtonType.PillRewards,
+  recentDappsEnabled: false,
```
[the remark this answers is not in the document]
kathaypacific: i enabled it for the e2e tests now since there's no harm, but initially i left it out because this feature sounded very experimental and likely to move somewhere else in the app from my discussion with Nitya :)

packages/mobile/src/home/RecentlyUsedDapps.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  if (!recentlyUsedDapps.length) {
+    return null
+  }
+
+  return (
+    <View style={styles.body} testID="RecentlyUsedDappsContainer">
+      <View style={[styles.titleContainer, styles.row]}>
+        <Text style={styles.title}>{t('recentlyUsedDapps')}</Text>
+        <TouchableOpacity style={styles.row} onPress={onPressAllDapps} testID="AllDapps">
```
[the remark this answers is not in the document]
kathaypacific: ah interesting, i didn't notice the subtle difference between the components.

i've updated to use `Touchable` but we should probably update this component at some point since `react-native-platform-touchable` is no longer maintained and they recommend using the button from `react-native-gesture-handler` instead

kathaypacific:
> Great work! 👍
> 
> Like you said, I'm also concerned by the icon styles. We're not using circular dapps icon in the Dapps page. It seems we made that choice for the recent dapps to match the circular icons used for transactions items.

yes the dapps are circles on the home screen to match the transaction feed items. depending on how successful this experiment is, this feature might be moved to another screen so we decided to not spend time optimising it for the home screen for now :)

jeanregisser (approved):
🚢 it! 👍

ValoraQA:
Hey @MuckT, We have verified above issue using Latest Android Internal Release Build V1.28.1 & updated the status in test rails.

**Test Rail Links:**
**Android 10:** https://valoraapp.testrail.io/index.php?/tests/view/16581

**Devices:** Samsung Galaxy M40(10.0)

**CC:** @kathaypacific 

Thanks..!!

ValoraQA:
Hey @MuckT, We have verified above task using Latest iOS Test flight Release build V1.28.1 & updated the status in test rails. **Observation :** Recently visited Dapps is seen on homepage after user navigates to it via Dapps page

**Test Rail Links:**
iOS 14 : https://valoraapp.testrail.io/index.php?/tests/view/17648&group_by=cases:section_id&group_order=asc&group_id=40

**New Enhancement :** [link](https://github.com/valora-inc/wallet/issues/2153)

**Devices:** iPhone 12(14.7.1)

CC: @kathaypacific
Thanks..!!

ValoraQA:
Hey @MuckT, We have verified this issue on Latest Android Internal Release Build V1.28.1, iOS Test flight Release build V1.28.1 & observed that user is not able to see Recently used Dapps on the Homepage after navigating to an external Dapp via Dapps section, so we have raised a bug for it.

**New Bug**: [Link](https://github.com/valora-inc/wallet/issues/2222)

**Devices:**  Google Pixel 2XL (11.0), iPhone 13 mini (15.1.1), iPhone 12(14.7.1)

**CC:** @kathaypacific 
Thanks..!!

PR #2054, opened 2022-02-25
Convert some Logger.errors to Logger.warns in profileInfo.ts

silasbw:
We get a raft of Errors from this code in Sentry and we ignore all of them. This
implies they're expected failures and should probably be Warns.


gnardini (approved):

ValoraQA:
HEy @muckt please let us know do we need to test it ?

Thanks..!

PR #2070, opened 2022-03-01
Disable Sentry if bundle id doesn't match the one set at build time

jeanregisser:
### Description

This is a tentative to avoid the environment pollution we currently see in Sentry.
See [Slack discussion](https://valora-app.slack.com/archives/C025V1D6F3J/p1645782671954849?thread_ts=1645758296.049519&cid=C025V1D6F3J).

### Tested

- Sentry init is skipped if bundle id doesn't match the one set at build time

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/packages/mobile/src/config.ts
+++ b/packages/mobile/src/config.ts
@@ -71,6 +71,8 @@ export const DEFAULT_FORNO_URL =
     : 'https://alfajores-forno.celo-testnet.org/'
 export const BLOCKSCOUT_BASE_URL = Config.BLOCKSCOUT_BASE_URL
 
+export const APP_BUNDLE_ID = Config.APP_BUNDLE_ID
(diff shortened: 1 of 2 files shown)


silasbw (commented):
I'd rather not have tricky enable/disable behavior like this, and it's not clear to me how this helps or with what issue it helps.

kathaypacific (approved):
nice!! i actually thought these random bundle id's in sentry were from very old testing, but i see recent events coming with them so this change is great! but yeah, how weird that this is happening 🤔

ValoraQA:
Hey @MuckT , JFYI: We are skipping this task as in "How others should test" section it is mentioned N/A.
Thanks..!!

PR #2059, opened 2022-02-25
feat(linked-bank-accounts): Use finclusive KYC to determine UI state

jh2oman:
### Description

As an end-user, I need to be able to connect my bank account with Plaid (step 2 in the "link bank account" UI) ONLY once it's possible to do so, i.e. once my Finclusive compliance check has completed.

### Other changes

Refactored some tests, got rid of some snapshots that weren't doing much

### Tested

When the M1 Persona Flow is ready to be released, ask a member of the Activation Squad to QA this feature. 

### How others should test

Manually Running through the persona flow and seeing the state changes. Its a bit difficult to do since it can only be done once and finclusive doesn't update to the final state.

### Related issues

- [[M2] Avoid allowing user to connect bank account before Finclusive KYC complete](https://app.zenhub.com/workspaces/acquisition-squad-sprint-board-6010683afabec1001a090887/issues/valora-inc/wallet/1873)

PR #2072, opened 2022-03-01
chore(kyc): rename kycStatus to personaKycStatus

jh2oman:
### Description

Rename kycStatus to personaKycStatus throughout for clarity now that we have multiple sets of kyc to deal with.

Followup to https://github.com/valora-inc/wallet/pull/2059#discussion_r815192620

### Tested

Some basic manual regression testing

### How others should test


packages/mobile/test/schemas.ts
```diff
@@ -1003,6 +1003,7 @@ export const v36Schema = {
   account: {
     ...v35Schema.account,
     finclusiveKycStatus: FinclusiveKycStatus.NotSubmitted,
+    personaKycStatus: undefined,
```
cajubelt, line 1006: shouldn't this be set to whatever `kycStatus` was before? otherwise the migration will result in data loss. This should probably be in a unit test, also

cajubelt (changes requested):
wow this looked painful, thanks for doing it though.

There's one potentially important thing related to migrations

packages/mobile/test/schemas.ts
cajubelt: oh wait, I thought this was a migrations file

jh2oman: I thought about that. The app restarts every time it upgrades right? If so, the kyc value from firebase is fetched every time the app starts, so we shouldn't have anything to worry about

cajubelt: don't we need a migration?

packages/mobile/src/redux/migrations.test.ts
```diff
@@ -407,4 +409,17 @@ describe('Redux persist migrations', () => {
     expect(migratedSchema.web3.fornoMode).toBe(true)
     expect(migratedSchema.web3.hadFornoDisabled).toBe(true)
   })
+
+  it('works for v35 to v36', () => {
```
cajubelt, line 413: 🙏

cajubelt (approved):
🥳

kathaypacific:
hey @jh2oman wanted to check if this PR is still ongoing? can we update + merge or close? :)

PR #2085, opened 2022-03-03
add logo and institution name for BankAccounts view

### LINKED ISSUE #1852: [M2] Add Institution Logo & Institution Name

lisaqyli:
### Description
When calling getFinclusiveBankAccounts endpoint to IHL, IHL would return a list of bank account information associated with the wallet. IHL now returns institution name and logo, we want to display them in the bankAccounts list view.

### Tested
Unit tested

### Related issues

- Fixes #1852


lisaqyli:
With fake data:
274805862_1190222141716284_5923268814323250007_n

When no logo avail:
274969597_633940807709391_3191677383846024400_n

packages/mobile/src/account/BankAccounts.tsx
```diff
@@ -73,24 +73,21 @@ function BankAccounts({ navigation, route }: Props) {
   }, [newPublicToken])
 
   function getBankDisplay(bank: BankAccount) {
+    // Todo: Consider adding a default placeholder image for banks without a logo available
```
cajubelt, line 76: can you write an issue for this and remove the todo? ...or just do it? doesn't sound that hard

packages/mobile/src/in-house-liquidity/index.ts
```diff
@@ -84,6 +84,8 @@ export interface BankAccount {
   accountName: string
   accountType: string
   accountNumberTruncated: string
+  institutionName: string
+  institutionLogo: string
```
cajubelt, line 88: if `institutionLogo` is expected to be nullable, that should be reflected in the type so that we get compile-time errors wherever we forgot to use null checks

packages/mobile/src/account/BankAccounts.tsx
lisaqyli: So the hard part is I dont have a default image at hand not the act of adding the placeholder image itself. Also the percentage of bank accounts plaid does not return logo is really small so I don't feel it's a ticket-worthy task as all. I also think just the gray blank image doesn't look too bad.

jh2oman: I think blank logo is fine

jh2oman (approved):

MuckT:
@lisaqyli does this need to be tested by QA in the release cycle?

PR #2128, opened 2022-03-09
feat(onboarding): Add a skipVerification flag that can skip phone verification during onboarding

jh2oman:
### Description

This PR adds a feature flag, `skipVerification`, that can be used to toggle whether or not Phone Verification should be a part of the onboarding flow. Removing phone verification from the onboarding flow will be part of an upcoming experiment.

[Experiment Proposal
](https://www.notion.so/valora-inc/Verification-c1afb404d23242d293bd80d2d40c8efb)

[Zenhub Ticket
](https://app.zenhub.com/workspaces/acquisition-squad-sprint-board-6010683afabec1001a090887/issues/valora-inc/wallet/1838)

https://user-images.githubusercontent.com/8432644/157553203-ecab871b-badf-4134-92b7-06f7610635c8.mp4

### Tested

Tested locally on an iOS emulator

### How others should test

Turn `skipVerification` feature flag to `true`. Then go through the onboarding flow (both create account, and restore account steps) and note that phone verification is skipped and the `Step X of X` header displays the correct numbers.


packages/mobile/src/app/selectors.ts
```diff
@@ -180,29 +182,48 @@ export const registrationStepsSelector = createSelector(
...
+        step =
+          step > (steps as Record<Screens.EnableBiometry, number>)[Screens.EnableBiometry]
+            ? step - 1
+            : step
       }
-      // remove biometry screen from step
-      const step = restoreAccountSteps[activeScreen as RestoreAccountScreens]
-      return { step: step > 3 ? step - 1 : step, totalSteps: 4 }
+    }
+    if (skipVerification) {
```
cajubelt, line 220:
```suggestion
    if (skipVerification && totalSteps > 0) {
```
this is pretty dumb but feels a little safer

packages/mobile/src/app/selectors.ts
```diff
@@ -180,29 +182,48 @@ export const registrationStepsSelector = createSelector(
...
+      totalSteps--
     }
 
-    if (biometryEnabled) {
-      return { step: createAccountSteps[activeScreen as CreateAccountScreens], totalSteps: 4 }
+    return {
+      step,
```
cajubelt, line 225: a hidden assumption here is that phone verification always takes place last in the list of steps. What if it didn't? Shouldn't we make sure `step` is updated to account for that? Could do this with something like `if (step >= steps[Screens.VerificationEducationScreen] || step >= steps[Screens.VerificationInputScreen]) step--`

packages/mobile/src/app/selectors.ts
```diff
@@ -180,29 +182,48 @@ export const registrationStepsSelector = createSelector(
...
+      totalSteps,
     }
-    // remove biometry screen from step
-    const step = createAccountSteps[activeScreen as CreateAccountScreens]
-    return { step: step > 3 ? step - 1 : step, totalSteps: 3 }
```
MuckT, line 206: I noticed previously the code would return out of the if blocks when conditions were hit. Here it looks like here we are switching to modifying the `totalSteps`, `step` and `steps` across the if blocks as there are now multiple conditions; e.g. I could have !biometryEnabled and also be in the experiment group so `totalSteps`, `step` and `steps` so it needs to dynamically account for those two conditions; Is this correct? I don't doubt that this works, but it's a bit hard to follow.

packages/mobile/src/import/saga.test.ts
```diff
@@ -79,6 +84,29 @@ describe('Import wallet saga', () => {
     await expectSuccessfulSagaWithPhrase(mockValidSpanishPhrase)
   })
 
+  it('initializes account and navigates to home if skipVerification is true', async () => {
```
cajubelt, line 87: is there also coverage on the case where skipVerification is false (ensuring that the verification screens are shown)?

packages/mobile/src/app/selectors.ts
jh2oman: Yes thats correct, it needs to be able to account for the possibility of both the biometrics screen and the phone verification screens being skipped.

The code looks much more confusing in diff mode than it does in split diff or on its own.

packages/mobile/src/redux/migrations.test.ts
```diff
@@ -428,6 +429,15 @@ describe('Redux persist migrations', () => {
...
     expectedSchema.app.showPriceChangeIndicatorInBalances = false
 
+    expect(migratedSchema).toMatchObject(expectedSchema)
+  })
+  it('works for v37 to v38', () => {
+    const oldSchema = v37Schema
+    const migratedSchema = migrations[38](oldSchema)
+
+    const expectedSchema: any = { ...oldSchema }
+    expectedSchema.app.skipVerification = false
```
cajubelt, line 439: we should probably use a deep copy here

cajubelt: since this mutates the oldSchema object

packages/mobile/src/import/saga.ts
```diff
@@ -160,8 +161,13 @@ export function* importBackupPhraseSaga({ phrase, useEmptyWallet }: ImportBackup
       ValoraAnalytics.track(AppEvents.redux_store_recovery_success, { account })
     }
     ValoraAnalytics.track(OnboardingEvents.wallet_import_success)
-
-    navigateClearingStack(Screens.VerificationEducationScreen)
+    const skipVerification = yield select(skipVerificationSelector)
+    if (skipVerification) {
+      yield put(initializeAccount())
```
cajubelt, line 166: seems like there's probably a good answer to this, but can you explain to me why we only need to initialize the account if we're skipping verification?

cajubelt (commented):
minor stuff, looks good overall

packages/mobile/src/redux/migrations.test.ts
jh2oman: 🤯

packages/mobile/src/app/selectors.ts
jh2oman: That would help safeguard against future changes. I can add it in.

packages/mobile/src/import/saga.test.ts
jh2oman: Adding it in 👍

packages/mobile/src/import/saga.ts
jh2oman: Currently `initializeAccount` is only called on the VerificationEducationScreen OR if it gets skipped because komenci is down or verification is disabled completely. I'm not entirely sure why those were the places chosen, but I though it probably best to continue calling `initializeAccount` when verification is skipped

packages/mobile/src/app/selectors.ts
```diff
@@ -218,7 +218,16 @@ export const registrationStepsSelector = createSelector(
       }
     }
     if (skipVerification) {
-      totalSteps--
+      if (Object.keys(steps).includes(Screens.VerificationEducationScreen)) {
+        totalSteps--
+        step =
+          step >
```
cajubelt, line 224:
```suggestion
          step >=
```
I think this needs to be >=

cajubelt: not 100% sure

cajubelt (approved):
nice! see comment, not sure if we need it to be >= in one place

packages/mobile/src/app/selectors.ts
jh2oman: Equal to would be the actual VerificationEducationScreen, which should hopefully never be shown when skipVerification is true

packages/mobile/src/app/selectors.ts
```diff
@@ -180,29 +182,57 @@ export const registrationStepsSelector = createSelector(
...
+    let steps
+    let totalSteps
+    let step
     if (recoveringFromStoreWipe) {
-      return {
-        step: storeWipeRecoverySteps[activeScreen as StoreWipeRecoveryScreens],
-        totalSteps: 3,
-      }
+      steps = storeWipeRecoverySteps
+      totalSteps = 3
```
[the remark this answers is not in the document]
jh2oman: It does happens to be the same since verification has 2 screens in the same step. I was hesitant to code that directly in b/c i'm not sure how stable that will be over time

packages/mobile/src/app/selectors.test.ts
```diff
@@ -129,6 +129,22 @@ describe('registrationStepsSelector', () => {
       }
     )
   })
+  it('should return the correct steps for create account screens with skipValidation enabled', () => {
+    ;(Object.keys(createAccountSteps) as Array<keyof typeof createAccountSteps>).forEach(
```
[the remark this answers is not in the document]
jh2oman: I don't think it actually has any purpose. I'll admit that i copy/pasted from the other tests which had this pattern

ValoraQA:
Hey @MuckT please help us to test the above PR as it has some feature to be enabled from backend

PR #2131, opened 2022-03-10
Upgrade typescript to 4.1.6

nicolasbrugneaux:
### Description

This PR upgrades typescript to the current major version. It is a stepping stone to having it up-to-date. Next step will be (from >= 4.2 yield typings are handled differently) to fix all the `yield` typings.

### Other changes

Minor types fixes

### Tested

Relying on automated tests

### How others should test

### Related issues

- Unblocks #1976 

### Backwards compatibility

Should be fully compatible.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,8 +68,8 @@
     "pretty-quick": "^2.0.1",
     "solc": "0.5.8",
     "ts-jest": "^26.5.1",
-    "ts-node": "^8.3.0",
-    "typescript": "^3.8.3"
+    "ts-node": "10.7.0",
+    "typescript": "4.1.6"
   },
   "resolutions": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -68,8 +68,8 @@
     "pretty-quick": "^2.0.1",
     "solc": "0.5.8",
     "ts-jest": "^26.5.1",
-    "ts-node": "^8.3.0",
-    "typescript": "^3.8.3"
+    "ts-node": "10.7.0",
+    "typescript": "4.1.6"
```
nicolasbrugneaux, line 72: no carret here to force this version. 4.2 is not yet compatible with valora's codebase

PR #2139, opened 2022-03-11
Fix auto version bump script failure

kathaypacific:
### Description

A few things needed to be ironed out after seeing this script in the wild:
- as this job is triggered by a tag, by default the repo is checked out with the tag as reference - need to avoid this as the tag is on the release branch which could contain extra commits (like in the case of 1.28.0) => checkout with `ref: main`
- the disclaimer part of the pre-deploy script executes asynchronously and was still ongoing while the next script step (push branch) was executing => add opt out for disclaimer. it's also nice to separate the disclaimer anyway because we will remove it from the pre-deploy script soon.
- the create PR api request was failing with error 
```
"message": "Validation Failed",
  "errors": [
    ***
      "resource": "Issue",
      "code": "missing_field",
      "field": "title"
    ***
  ],
  "documentation_url": "https://docs.github.com/rest/reference/pulls#create-a-pull-request"
```
=> add "title" to the request body (this was annoying, as the docs clearly say that this param is not required...)

### Other changes

N/A

### Tested

Manually

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/.github/scripts/create_app_version_pr.sh
+++ b/.github/scripts/create_app_version_pr.sh
@@ -30,4 +30,4 @@ curl -u "valora-bot:$VALORA_BOT_TOKEN" \
   -X POST \
   -H "Accept: application/vnd.github.v3+json" \
   https://api.github.com/repos/valora-inc/wallet/pulls \
-  -d '{ "head": "'$BRANCH_NAME'", "base": "main", "title": "'$commit_message'" }'
+  -d '{ "head": "'$BRANCH_NAME'", "base": "main", "title": "Bump app version to '$app_version'" }'
--- a/.github/scripts/create_app_version_pr.sh
+++ b/.github/scripts/create_app_version_pr.sh
@@ -1,31 +1,31 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-cd packages/mobile
+# cd packages/mobile
 
-# ensure that we are using ssh
-git remote set-url origin git@github.com:valora-inc/wallet.git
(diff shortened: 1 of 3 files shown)


packages/mobile/scripts/pre-deploy.sh
```diff
@@ -9,10 +9,12 @@ set -euo pipefail
 # --minor (Optional): Bump minor version automatically (default is to let user input new version number)
 
 MINOR=false
+NO_DISCLAIMER=false
```
jeanregisser, line 12: nit:
```suggestion
NO_LICENSE_UPDATE=false
```

.github/scripts/create_app_version_pr.sh
```diff
@@ -6,22 +6,26 @@ cd packages/mobile
...
 
-app_version="$(node -p "require('./package.json').version")"
-
-echo "Create branch"
+echo "Create version bump branch from main"
 git checkout -b $BRANCH_NAME
 
 echo "Bump app version"
-yarn pre-deploy --minor
+yarn pre-deploy --minor --no-disclaimer
```
jeanregisser, line 13: nit: clearer I think
```suggestion
yarn pre-deploy --minor --no-license-update
```

jeanregisser (approved):
Nice! 🤖 🚀

.github/scripts/create_app_version_pr.sh
kathaypacific: yes you're right!! i've updated both places now :) good idea!

Issue #2101, opened 2022-03-08
Ensure correct handling for WalletConnect and DAppKit links from the internal WebView


jeanregisser:
This task is about updating the logic for interacting with dapps using WalletConnect and DappKit from within the internal WebView for dapps.

See also [Option 3](https://www.notion.so/valora-inc/Internal-dapps-browser-design-document-27a7e9591b274e398365e4d9f52b74cc#0d36aed01dde4f559d0c04f0183ddacc) from the internal dapps browser design document.

AC:
- [ ] Connecting/signing with WalletConnect from the internal WebView should work
- [ ] Connecting/signing with DAppKit from the internal WebView should work
- [ ] WalletConnect messaging around going back to the dapp should be updated, as there's no need to go back from within Valora

PR #2144, opened 2022-03-11
Open dapps in a webview

### LINKED ISSUE #2100: Implement WebView for opening dapps
This is the implementation of [Option 3](https://www.notion.so/valora-inc/Internal-dapps-browser-design-document-27a7e9591b274e398365e4d9f52b74cc#0d36aed01dde4f559d0c04f0183ddacc) from the internal dapps browser design document.

[Designs](https://www.figma.com/file/zdK8XvxNXvTjxnM5V33pHw/Dapps-Page)

AC:
- [ ] A WebView is used to open a dapp instead of opening it using the external browser
- [ ] The WebView has basic navigation controls at the bottom
- [ ] Cookies set in the WebView should be preserved

kathaypacific:
### Description

In this PR:
- add remote config flag for `dappsWebViewEnabled`
- if `dappsWebViewEnabled`, open dapp in a WebView instead of an external browser
- Cookies are preserved by default 
- adjust the screen styles for WebView to match [designs](https://www.figma.com/file/zdK8XvxNXvTjxnM5V33pHw/Dapps-Page?node-id=1999%3A45353), add basic bottom nav bar and refresh control

Screenshot_1647349346

**NOT included** in this PR:
- connecting dapps to app (i.e. clicking "connect wallet" on dapps inside the webview - this requires some more work and will be part of https://github.com/valora-inc/wallet/issues/2101)

### Other changes

Deleted an unused file (InAppBrowser)

### Tested

Manually and some unit test for webview screen and opening dapps via webview or deeplink

### How others should test

Open dapps from dapp explorer/recently used dapps, verify that the dapp is opened in-app.
should be able to navigate back and forth within the dapp using the bottom bar nav controls (e.g. click on a button inside the dapp which leads you to a different part of the dapp, then pressing the back button on the bottom nav bar should return you to the previous dapp page)

### Related issues

- Fixes #2100 

### Backwards compatibility

Yes


packages/react-components/icons/ForwardChevron.tsx
```diff
@@ -4,30 +4,32 @@ import Svg, { Path } from 'svgs'
...
   color?: string
 }
 
-export default class ForwardChevron extends React.PureComponent<Props> {
-  static defaultProps = {
-    height: 20,
-    width: 10,
-    color: colors.gray3,
-  }
+function ForwardChevron({ color, height }: Props) {
```
kathaypacific, line 10: I noticed that the BackChevron and ForwardChevron actually didn't look the same when placed side by side. so i updated them to be the ones from https://www.figma.com/file/zdK8XvxNXvTjxnM5V33pHw/Dapps-Page?node-id=1999%3A45353

packages/mobile/src/app/WebViewScreen.tsx
```diff
@@ -37,29 +71,66 @@ function WebViewScreen({ route }: Props) {
...
+        startInLoadingState={true}
+        renderLoading={() => <ActivityIndicator style={styles.loading} size="large" />}
+        onNavigationStateChange={(navState) => {
+          setCanGoBack(navState.canGoBack)
+          setCanGoForward(navState.canGoForward)
+        }}
       />
-      {loading && <ActivityIndicator style={styles.loading} size="large" />}
-    </View>
+      <View style={styles.navBar}>
```
dievazqu, line 107: I think this is expected only for iOS according to the designs.
Is that something we can do? I mean, add some code only for iOS or Android.

gnardini: Yes! You can use the `Platform.OS` constant that can be either `android` or `ios`

gnardini: However, there would be no forward button in Android in that case, so I'm not sure that's what we want, even if we said that in the previous meeting 🤔

dievazqu: Also for @nityarao, I have to say that I'm bit concern about the UX.
We would have the refresh button at the top while the arrows for navigation at the bottom. I feel it's a bit weird.
I think it would nice to have all those buttons together.
Also, for Android users, what is the expected behaviour if they click on the back button on their devices? Should they go back to the previous webApp page, or should they return to the previous Valora page (e.g. dapps marketplace or homescreen)

nityarao: @gnardini Thats a good catch about not having a forward button in Android. I'd think that users would benefit from navigating both forward & back on dapp pages. 

@dievazqu that being said, what do you think about having the same bottom bar for both iOS & Android to start? 

The icons would all be grouped together in the bottom nav bar. This was my original idea but concerns were raised around it being repetitive with the device back button. In this case, both the device back button & browser back button can take the users back one screen. The close button at the top will close the browser.

kathaypacific: thanks @dievazqu for bringing this up! grateful for a diverse set of users amongst us 🙏🏻 indeed i didn't think too much on the layout for android before, but now I am thinking:

- while it's nice to be able to browse back and forth, it seems that this is not what Android browsers (like Chrome) easily enable. if this is not part of the "usual" browsing experience for android users, having the bottom nav there might not be useful for them (might be confusing, and provide functionality they don't use). i lean towards Diego's idea of hiding the bottom nav for Android users, to give a more native experience. @nityarao  if you see in this screenshot, it does look very bizarre to have the bottom bar + the hardware buttons and takes some screen real estate away 
Screenshot 2022-03-15 at 09 41 26

- i think for Android screen headers, it is conventional to have the left item as an icon and have the header title on the left rather than in the center. should i make this adjustment here too? instead of "close", i could use a cross icon for android. and the title would be next to the cross icon like [this ](https://developer.android.com/guide/navigation/navigation-ui) - curious for your thoughts @nityarao , it'd be cool to build a "native" feeling screen but i also see this is not the pattern we use for other screens so i'll go with what you decide

packages/mobile/src/app/WebViewScreen.tsx
```diff
@@ -37,29 +71,66 @@ function WebViewScreen({ route }: Props) {
     return true
   }
 
-  const hideLoading = () => {
-    setLoading(false)
+  const handleRefresh = () => {
+    if (webviewRef && webviewRef.current) {
+      webviewRef.current.reload()
+    }
```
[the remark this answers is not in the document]
kathaypacific: yeah good idea, i have updated :)

packages/mobile/src/app/WebViewScreen.tsx
kathaypacific: update: sorry i missed the updated designs, i have now moved the refresh icon also to the bottom nav bar. i've been doing some research and am pretty satisfied that having this bottom nav bar with all the actions isn't a strange pattern on android :) for me, having the refresh icon there makes a big difference in making the nav bar feel more useful. so i'd like to leave it like this, and get some feedback through our internal testers on the nightly builds. i'd be happy to tweak anything that comes out of the feedback! (have updated the PR description snapshot to include the change to the bottom nav bar)

packages/mobile/src/app/WebViewScreen.tsx
```diff
@@ -1,33 +1,68 @@
...
+
+  useBackHandler(() => {
+    // android hardware back button functions as either browser back button or
+    // app back button
+    if (canGoBack) {
+      handleGoBack()
+    } else {
+      navigateBack()
+    }
+    return true
```
jeanregisser, line 64: Nice!

packages/mobile/src/app/WebViewScreen.tsx
```diff
@@ -37,29 +72,61 @@ function WebViewScreen({ route }: Props) {
     return true
   }
 
-  const hideLoading = () => {
-    setLoading(false)
+  const handleRefresh = () => {
+    webviewRef?.current?.reload()
```
jeanregisser, line 76: nit: `webviewRef` itself can never be null or undefined, right?

packages/mobile/src/redux/migrations.ts
```diff
@@ -557,4 +557,11 @@ export const migrations = {
       },
     }
   },
+  41: (state: any) => ({
+    ...state,
+    app: {
+      ...state.app,
+      dappsWebViewEnabled: false,
+    },
+  }),
```
jeanregisser, line 566: We can also simply return the state. The state reconciler will add the new prop as it merges it with the new initialState.
But good to be explicit too. 👍 

```suggestion
  41: (state: any) => state,
```

jeanregisser (approved):
🚀

packages/mobile/src/app/WebViewScreen.tsx
kathaypacific: omg yes you're right, i got confused between the null value of the ref (`webviewRef.current`) and the ref (`webviewRef`) itself 🤦🏻‍♀️ i've removed the redundant null check

packages/mobile/src/redux/migrations.ts
kathaypacific: i made this change :) TBH i don't have strong feelings about it though - i know adding properties can be done automatically like this so it works, but removing properties we need to do manually... so it's an extra thing that we need to remember (rather than just blanket rule for explicitly declaring what you are doing to the redux state in a migration)? consistency is a little hard to enforce as well, although not so important for this case i guess 😅

ValoraQA:
Hey @MuckT, We have verified above task on Latest Test Flight Alfa jores build V1.29.0, Android Alfa jores Build V1.29.0 & updated status in test rails. When user taps on any dapps & taps on "Go to apps" button , user get redirected to in-app webview for dapps.

**Test rail link:**
**Android 11:** https://valoraapp.testrail.io/index.php?/tests/view/20916
**iOS13:** https://valoraapp.testrail.io/index.php?/tests/view/20853
**iOS14:** https://valoraapp.testrail.io/index.php?/tests/view/20790

Devices: Google Pixel 2XL (11.0), iPhone SE 2020 (14.8.1), iPhone 7 (13.3)
**cc:** @kathaypacific 
Thanks..!!

PR #2152, opened 2022-03-12
WIP use private key to sign JWTs for IHL auth

### LINKED ISSUE #2029: [M1] Update IHL auth in wallet to use private key to encrypt JWT payload sent to IHL
Long story short, we realized that while the FiatConnect spec requiring the user to have a DEK registered might makes sense for Valora, it's otherwise too strict of a requirement on the end user. We decided to transition FC auth to using simple public/private key signing using the user's typical private key. If we ever plan on wrapping Finclusive in a FiatConnect API, then, we'll need to modify IHL's auth to match this, since it too currently requires a DEK (which is where the phone number verification requirement comes from).

This ticket is for updating the JWT generation/signing for IHL in the wallet to use the user's private key rather than their DEK private key.

Eng:
- note that EOA (public key) will need to be added as a parameter for helpers calling IHL (so that the public key can be included in the JWT)

AC:
* IHL auth in the wallet is refactored to encrypt the JWT payload with the user's private key.
* integration test against Alfajores IHL with auth turned on (can be run manually)

### LINKED ISSUE #2030: [M1] Use EOA as identifier throughout PFP integration in the wallet
Currently, we use the user's MTW address as an identifier throughout our PFP integration. Since we now will be sending a user's EOA address to IHL, we need to change the wallet integration to use the EOA address as an identifier, replacing the MTW address wherever it appears. This includes:

* In requests to IHL
* As an identifier in Firebase

Once we change the identifier in Firebase, we need to update the Firbase security rules for the `in-house-liquidity` node to only allow reads from users who's EOA address matches the identifier in Firebase. (Can copy over from other nodes with the same rule.)

AC:
* Wallet refactored to use EOA address instead of MTW for PFP integration
* Firebase rules updated
* manual integration tests against IHL with auth turned on should pass

cajubelt:
### Description

switching to using the private key, rather than the DEK private key, to sign JWTs for IHL auth

This is part of an effort to prepare for FiatConnect. FiatConnect auth uses JWTs signed by the private key to be easier for other wallets to implement.

### Other changes

this includes a switch to EOA from MTW addresses

### Tested

TODO (ongoing)

### How others should test

TODO

### Related issues

fixes #2029 
fixes #2030

### Backwards compatibility

no issues here since the feature is still sealed off from production users


cajubelt:
closing this since we are getting rid of jwt auth

PR #2169, opened 2022-03-16
silencing some sentry dashboard errors

lisaqyli:
### Description
Silencing some of the top errors that has been flooding our dashboard but not oncall worthy
Network connect was lost - it seems like it happens when user put app to background
https://sentry.io/organizations/valora-inc/issues/2957958304/?project=1250733&query=is%3Aunresolved&sort=freq&statsPeriod=30d

CleverTap initialURL is null
https://sentry.io/organizations/valora-inc/issues/3020003220/similar/?project=1250733&query=is%3Aunresolved&statsPeriod=14d

Cancelled pin code input
https://sentry.io/organizations/valora-inc/issues/2978422876/?project=1250733&query=is%3Aunresolved&statsPeriod=14d

Picture input: user cancelled or missing permission
https://sentry.io/organizations/valora-inc/issues/3002574596/similar/?project=1250733&query=is%3Aunresolved+picture&statsPeriod=14d
https://sentry.io/organizations/valora-inc/issues/3002577036/?project=1250733&statsPeriod=14d


packages/mobile/src/tokens/saga.ts
```diff
@@ -130,7 +130,10 @@ export function* fetchToken(token: Currency, tag: string) {
     } else {
       ValoraAnalytics.track(AppEvents.fetch_balance_error, balanceLogObject)
     }
-  } catch (error) {
+  } catch (error: any) {
+    if (error?.message?.includes('The network connection was lost')) {
```
silasbw, line 134: This still `Logger.error`s below. Did you mean this to be an if/else?

packages/mobile/src/app/App.tsx
```diff
@@ -70,7 +70,7 @@ export class App extends React.Component<Props> {
     // Handles opening Clevertap deeplinks when app is closed / in background
     CleverTap.getInitialUrl(async (err: any, url) => {
       if (err) {
-        Logger.error('App/componentDidMount', 'App CleverTap Deeplink on Load', err)
+        Logger.debug('App/componentDidMount', 'App CleverTap Deeplink on Load', err)
```
silasbw, line 73: i don't know anything about this code, but downgrading to `debug` from `error` seems drastic and I would have expected a `warn` instead.  Any reason you moved it to `debug`?

kathaypacific: @lisaqyli we have a remote config variable for exact errors that we want to stop reporting on, it's called `sentryNetworkErrors` - are these errors something that we should add to this variable, rather than make changes to the code? the benefit of using `sentryNetworkErrors` is that we can exclude specific error values (rather than exclude specific parts of the code where we catch errors, where the error can be any value)

kathaypacific:
@lisaqyli we have a remote config variable for exact errors that we want to stop reporting on, it's called `sentryNetworkErrors` - are these errors something that we should add to this variable, rather than make changes to the code? the benefit of using `sentryNetworkErrors` is that we can exclude specific error values (rather than exclude specific parts of the code where we catch errors, where the error can be any value)

packages/mobile/src/app/App.tsx
jeanregisser: Could we filter out the exact error instead?

jeanregisser (commented):
Love this!! Thanks so much 💯

jeanregisser:
> @lisaqyli we have a remote config variable for exact errors that we want to stop reporting on, it's called sentryNetworkErrors - are the errors string you mention above something that we should add to this variable, rather than make changes to the code? the benefit of using sentryNetworkErrors is that we can exclude specific error values (rather than exclude specific parts of the code where we catch errors, where the error can be any value, which seems to be what you are trying to achieve here)

To me `sentryNetworkErrors` is only to filter out actual network errors happening, which we didn't have a definitive list for.
Maybe we can make it more generic.
But at the same time, I think it's good to filter out specific expected errors exactly where they should happen. Like this PR is doing. Of course it's gonna take time to address all of this.
Well, maybe we can chat again about this.

packages/mobile/src/app/App.tsx
```diff
@@ -70,7 +70,11 @@ export class App extends React.Component<Props> {
     // Handles opening Clevertap deeplinks when app is closed / in background
     CleverTap.getInitialUrl(async (err: any, url) => {
```
cajubelt, line 71: if the callback were taken out and exposed (with something like `export const cleverTapGetInitialUrlCallback = ...`) then it could be tested pretty easily

kathaypacific:
> > @lisaqyli we have a remote config variable for exact errors that we want to stop reporting on, it's called sentryNetworkErrors - are the errors string you mention above something that we should add to this variable, rather than make changes to the code? the benefit of using sentryNetworkErrors is that we can exclude specific error values (rather than exclude specific parts of the code where we catch errors, where the error can be any value, which seems to be what you are trying to achieve here)
> 
> To me `sentryNetworkErrors` is only to filter out actual network errors happening, which we didn't have a definitive list for. Maybe we can make it more generic. But at the same time, I think it's good to filter out specific expected errors exactly where they should happen. Like this PR is doing. Of course it's gonna take time to address all of this. Well, maybe we can chat again about this.

i've been thinking about this for a few days and i do think it's nice to filter out errors in this way if we know where they occur, but at the same time it just seems like a game of whack-a-mole? assuming that we don't catch these errors and already filter them out when developing, we'd need to see them in sentry -> decide we don't care about them -> update the code -> wait until the next release but people who have not updated continue sending these errors? would it not be easier to have another remote config variable like `sentryIgnoredErrors` that works in a similar way to `sentryNetworkErrors` by not sending errors that match an entry in this variable, and we can update this list on the fly? just thinking out loud :) maybe this is something we can discuss at eng sync

jeanregisser:
Yes, I wanted to ignore these directly in Sentry, instead of doing that in Valora.
But ignored errors still count toward our quota: https://help.sentry.io/account/billing/do-ignored-events-in-the-ui-count-towards-my-quota/

Your idea of having a `sentryIgnoredErrors` seems quite appealing.
My only concern is if we start ignoring errors that are expected if they originate from a specific place in the codebase, but that should otherwise be reported if they happen in a different place.

packages/mobile/src/onboarding/registration/PictureInput.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+    )
+
+    await fireEvent.press(getByText('chooseFromLibrary'))
+    // Transition time for the modal to close
+    jest.advanceTimersByTime(500)
+    await waitFor(() =>
+      expect(Logger.info).toHaveBeenCalledWith('PictureInput', 'User cancelled image selection')
+    )
+  })
+})
```
jeanregisser, line 98: Thanks for adding this test 👍

packages/mobile/src/onboarding/registration/PictureInput.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+    const store = createMockStore()
+    const tree = render(
+      <Provider store={store}>
+        <PictureInput {...mockProps} />
+      </Provider>
+    )
+    expect(tree).toMatchSnapshot()
+  })
+
+  it('onOptionChosen', async () => {
```
jeanregisser, line 53: nit: could you add a more descriptive title?

jeanregisser (approved):
🚀

Issue #1351, opened 2021-11-02
 [13 iOS Specific] Blank screen displayed when user browse “https://app.sushi.com/swap” in Safari/Edge browser


ValoraQA:
**Frequency:** 100
**Repro on Browser:** Microsoft Edge (V 95.1020.40), Safari (V 13.6.1)
**Repro on devices:** iPhone SE (13.6.1), iPhone XR (13.2.3)
**Does Not Repro on devices:** iPhone SE (14.2), Redmi Note 8 (10.0), One Plus 7T (11.0), iPhone 12 (15)
**Repro Steps:**
1. Launch the "https://app.sushi.com/swap" URL in the Safari mWeb browser 
2. Wait for some time & observe that 
**Current Behavior:** Blank screen displayed on “app.sushi.com/swap” page
**Expected Behavior:** Page should load completely & user should able to proceed to connect wallet with Valora
**Investigation:** Also we have observed the same issue for the Microsoft Edge mWeb browser 
**Impact:** User gets blocked to connect to “app.sushi.com/swap” on specific OS’s
**Attachment:** [BlankScreenForSushi.mp4](https://drive.google.com/file/d/1GPpCQ2RB_PFpAsvFeYpH_JCJDo7S4WqZ/view?usp=sharing)

ValoraQA:
https://valora.testquality.com defect [D444](https://valora.testquality.com/project/12128/result/2540319/defect/8749).

ValoraQA:
https://valora.testquality.com defect [D445](https://valora.testquality.com/project/12128/result/2540320/defect/8750).

ValoraQA:
https://valora.testquality.com defect [D446](https://valora.testquality.com/project/12128/result/2545181/defect/8751).

ValoraQA:
https://valora.testquality.com defect [D447](https://valora.testquality.com/project/12128/result/2545185/defect/8752).

ValoraQA:
https://valora.testquality.com defect [D619](https://valora.testquality.com/project/12128/result/2571425/defect/9224).

ValoraQA:
https://valora.testquality.com defect [D620](https://valora.testquality.com/project/12128/result/2571426/defect/9225).

MuckT:
Raised a bug with Sushi Team and downgraded priority.

ValoraQA:
Thanks fro confirmation:)

ValoraQA:
[...]

PR #2178, opened 2022-03-17
Ensure correct WC/Dappkit handling for dapps internal WebView

### LINKED ISSUE #2101: Ensure correct handling for WalletConnect and DAppKit links from the internal WebView

kathaypacific:
### Description

In this PR:
- when a user is launching a wallet connect or dappkit connection to Valora, ensure that we navigate back (not navigate home) 
- there is a WC loading screen that is added to the stack, which we do not want the user to land on when navigating back. i opted for a generic solution here, as we would never want this behaviour. I added a check in the `navigate` function to replace this screen (rather than push a new screen on top) if it is on top of the stack
- add redux variable to track if a dapp browsing session is active. this is used to modify the in-app messaging after a successful connection, and tell the WebView to respond to the dappkit deep link
- i had a hard time getting the dappkit connection right. it looks like changing the source uri of the WebView reloads the dapp, which is not what we want. therefore there is the javascript injection for the dappkit deep link.

https://user-images.githubusercontent.com/20150449/158836046-7b1f8dbd-3599-4245-8e37-fa95326c6ec7.mov

### Other changes

- There is some logic to handle a WC loading timeout, i moved this to a function used in the saga rather than having in the display component.
- Move twilio balance check for e2e inside the conditional block as twilio balance check failure was causing the whole e2e job to fail e.g. https://github.com/valora-inc/wallet/runs/5626058429?check_suite_focus=true

### Tested

Manually

### How others should test

Note that the in-app dapp view is hidden behind a feature flag. When enabled, then:
- opening dapps from the dapps page or recently used dapps section should be opened inside the app
- when connecting Valora to the dapp, the confirm screens should appear and you should be navigated back to the dapp on confirm/cancel
- you should be able to connect Valora to all dapps and be able to perform transactions successfully
- you should still be able to connect Valora to external dapps by scanning the dapp QR (e.g. from a computer browser or from a browser on the device) and perform transactions successfully

### Related issues

- Fixes #2101 

### Backwards compatibility

Yes


kathaypacific:
A couple of thoughts, looking for opinions:
- during WC/dappkit communication between Valora and the dapp, the navigation happens more quickly than the dapp is able to respond (so we navigate back to the dapp and it takes a moment for the dapp to do anything) - i added a success banner that says the connection was successful, is this clear? or should i add an artificial delay before navigating back to the dapp? 
- in the video i attached to the description, you can see that when a user cancels an action, the dapp doesn't know about this and is in a pending state. i *think* this is acceptable behaviour, keen for someone else's thoughts though

packages/mobile/src/navigator/NavigationService.ts
```diff
@@ -89,12 +89,24 @@ export function navigate<RouteName extends keyof StackParamList>(
...
+      // Remove the WC loading screen from the stack as we never want to land
+      // on this screen when navigating back. Ensures that the dapp webview WC
+      // navigation works correctly.
+      if (navigationRef.current?.getCurrentRoute()?.name === Screens.WalletConnectLoading) {
+        navigationRef.current?.dispatch(
+          StackActions.replace({
+            routeName,
+            params,
+          })
+        )
```
jeanregisser, line 101: Wondering, how does it look visually when it's replaced?
Is it a hard replace, without any animation?

packages/mobile/src/app/WebViewScreen.tsx
```diff
@@ -53,6 +63,14 @@ function WebViewScreen({ route, navigation }: Props) {
     })
   }, [navigation])
 
+  useEffect(() => {
+    if (dappSessionActive && dappkitDeeplink) {
+      webViewRef.current?.injectJavaScript(
+        `window.history.replaceState({}, "", "${dappkitDeeplink}"); true;`
+      )
+    }
```
jeanregisser, line 71: 👍

packages/mobile/src/app/WebViewScreen.tsx
```diff
@@ -25,22 +26,31 @@ type RouteProps = StackScreenProps<StackParamList, Screens.WebViewScreen>
...
+  const dappSessionActive = useSelector(dappSessionActiveSelector)
 
-  const webviewRef = useRef<WebViewRef>(null)
+  const webViewRef = useRef<WebViewRef>(null)
   const [canGoBack, setCanGoBack] = useState(false)
   const [canGoForward, setCanGoForward] = useState(false)
 
+  const handleCloseWebView = () => {
+    if (dappSessionActive) {
+      dispatch(dappSessionEnded())
```
jeanregisser, line 41: This needs to be called when the webView is unmounted.
Which not only happens when the Close button is touched, but also when swiping back on iOS or when the Android back button is used.

We could use a `useEffect` clean up function for this.

packages/mobile/src/dappkit/DappKitAccountScreen.tsx
```diff
@@ -65,7 +65,7 @@ class DappKitAccountAuthScreen extends React.Component<Props> {
       Logger.error(TAG, 'No account set up for this wallet')
       return
     }
-    navigateHome({ onAfterNavigate: () => this.props.approveAccountAuth(request) })
+    this.props.approveAccountAuth(request)
```
jeanregisser, line 68: Doesn't this break the normal flow without the in-app WebView?

packages/mobile/src/dappkit/DappKitSignTxScreen.tsx
```diff
@@ -54,7 +54,7 @@ class DappKitSignTxScreen extends React.Component<Props> {
   linkBack = () => {
     const request = this.getRequest()
 
-    navigateHome({ onAfterNavigate: () => this.props.requestTxSignature(request) })
+    this.props.requestTxSignature(request)
```
jeanregisser, line 57: ditto

packages/mobile/src/navigator/NavigationService.ts
```diff
@@ -89,12 +89,24 @@ export function navigate<RouteName extends keyof StackParamList>(
...
-      navigationRef.current?.dispatch(
-        NavigationActions.navigate({
-          routeName,
-          params,
-        })
-      )
+      // Remove the WC loading screen from the stack as we never want to land
+      // on this screen when navigating back. Ensures that the dapp webview WC
+      // navigation works correctly.
+      if (navigationRef.current?.getCurrentRoute()?.name === Screens.WalletConnectLoading) {
```
jeanregisser, line 95: Is there a way to make this logic more local?
This works but feels a bit hacky to be applied here.

But I feel the alternative is probably not great either 😄

packages/mobile/src/redux/store.test.ts
```diff
@@ -137,6 +137,7 @@ describe('store state', () => {
           "celoEducationUri": null,
           "celoEuroEnabled": true,
           "dappListApiUrl": null,
+          "dappSessionActive": null,
```
jeanregisser, line 140: The schema version should be bumped as best practice.

packages/mobile/src/walletConnect/saga.ts
```diff
@@ -42,3 +44,9 @@ export function* initialiseWalletConnect(uri: string, origin: WalletConnectPairi
       throw new Error(`Unsupported WalletConnect version '${version}'`)
   }
 }
+
+export function* showWalletConnectionSuccessMessage(dappName: string) {
+  const dappSessionActive = yield select(dappSessionActiveSelector)
+  const successMessage = dappSessionActive ? 'inAppConnectionSuccess' : 'connectionSuccess'
+  yield put(showMessage(i18n.t(successMessage, { dappName })))
```
jeanregisser, line 51: nit: I'm still thinking we'd want to extract strings from the source at some point. This would make it more difficult since it's happening at runtime. Here's an alternative:

```suggestion
  const successMessage = dappSessionActive ? i18n.t('inAppConnectionSuccess', { dappName }) : i18n.t('connectionSuccess', { dappName })
  yield put(showMessage(successMessage))
```

packages/mobile/src/walletConnect/walletConnect.ts
```diff
@@ -52,3 +57,21 @@ export function isWalletConnectDeepLink(deepLink: string) {
...
+    if (await isScreenOnForeground(Screens.WalletConnectLoading)) {
+      ValoraAnalytics.track(WalletConnectEvents.wc_pairing_error, {
+        error: 'timed out while waiting for a session',
+      })
+      navigate(Screens.WalletConnectResult, {
+        title: i18n.t('timeoutTitle'),
+        subtitle: i18n.t('timeoutSubtitle'),
+      })
+    }
+  }, CONNECTION_TIMEOUT)
```
jeanregisser, line 76: This works, but is not very saga-ish and can't be tested as easily.

We could use a `race` with a `delay`. There's a similar example in https://github.com/valora-inc/wallet/blob/a91f80c89efcad47a5cb189e820d11942622b884/packages/mobile/src/fiatExchanges/saga.ts#L70-L77

packages/mobile/src/walletConnect/walletConnect.ts
```diff
@@ -36,7 +41,7 @@ export function* handleWalletConnectDeepLink(deepLink: string) {
   // handler is called, so it's important we don't display the loading screen on top
   const hasPendingState: boolean = yield select(selectHasPendingState)
   if (!hasPendingState) {
-    navigate(Screens.WalletConnectLoading, { origin: WalletConnectPairingOrigin.Deeplink })
+    handleWalletConnectLoadingWithTimeout({ origin: WalletConnectPairingOrigin.Deeplink })
```
jeanregisser, line 44: We should use `yield call(...)`  since this is a side effect.
Otherwise it's more difficult to unit test.

jeanregisser (commented):
This is great! The approach seems good to me! 🚀

packages/mobile/src/app/reducers.ts
```diff
@@ -57,6 +57,7 @@ export interface State {
   showPriceChangeIndicatorInBalances: boolean
   paymentDeepLinkHandler: PaymentDeepLinkHandler
   dappsWebViewEnabled: boolean
+  dappSessionActive: Dapp | null
```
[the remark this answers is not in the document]
kathaypacific: i think this is a good idea, but i didn't do this because we don't store the list of dapps in redux so having the whole dapp here didn't feel like data duplication. the only dapps data we store is in the recent dapps list but it seems unsafe to tie this feature to the recent dapps (the number of recent dapps is controlled by remote config and could be 0, in which case we wouldn't store any dapps)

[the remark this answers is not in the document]
jeanregisser: ah yeah right! all good then 👍

packages/mobile/src/app/reducers.ts
```diff
@@ -257,6 +259,12 @@ export const appReducer = (
           action.dapp,
           ...state.recentDapps.filter((recentDapp) => recentDapp.id !== action.dapp.id),
         ].slice(0, state.maxNumRecentDapps),
+        dappSessionActive: action.dapp,
+      }
+    case Actions.DAPP_SESSION_ENDED:
+      return {
+        ...state,
+        dappSessionActive: null,
```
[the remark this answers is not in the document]
kathaypacific: ah yes, this is much better. i was feeling uneasy about throwing stuff in the appInit saga all the time, i forgot about the rehydrate! i've made this change

packages/mobile/src/app/selectors.ts
```diff
@@ -139,6 +139,9 @@ export const skipVerificationSelector = (state: RootState) => state.app.skipVeri
 
 export const dappsWebViewEnabledSelector = (state: RootState) => state.app.dappsWebViewEnabled
 
+export const dappSessionActiveSelector = (state: RootState) =>
```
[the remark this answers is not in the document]
kathaypacific: oh yes i also renamed the redux variable to `activeDapp`, i think this is a good suggestion

packages/mobile/src/navigator/NavigationService.ts
kathaypacific: really good question...i've been playing around with it all afternoon and not sure what to do with it (couldn't find the right combination of props to make it perfect). it seems that the replace has some animation but it's not quite right and the screen is removed before the next screen fully replaces it. on most dapps this is fine, but for the dark/colorful ones you can see a flash:

https://user-images.githubusercontent.com/20150449/159040200-817961f5-2c98-490f-9cec-96aad7822e66.mov

i'll keep looking into it but also wondered how much on a scale of 1-10 this triggers your OCD? 😅

packages/mobile/src/app/WebViewScreen.tsx
kathaypacific: ah yes good one, i hadn't thought of this! i have added this now

packages/mobile/src/walletConnect/saga.ts
kathaypacific: oh right, yes done :)

packages/mobile/src/redux/store.test.ts
kathaypacific: 😅 updated

packages/mobile/src/walletConnect/walletConnect.ts
kathaypacific: you're right, the setTimeout wasn't very saga-ish...i was thinking of this function more like an async helper, but indeed it is a bit confusing to have this living with other sagas if it's not a saga too.

i ended up doing this very differently, i'm curious for your thoughts. i created a `initialiseWalletConnectWithLoading` generator function that will race `initialiseWalletConnect` with the timeout - i think this makes more sense

packages/mobile/src/walletConnect/walletConnect.ts
kathaypacific: ah yes, i have updated :)

kathaypacific:
thank you for the feedback @jeanregisser !! there are loads of scenarios to test for this one so i'll continue testing on monday but i think it should be working and ready for another pass :)

packages/mobile/src/walletConnect/walletConnect.ts
kathaypacific: (wanted to clarify here, that i did try to directly replace the `setTimeout` with a `race` + `delay` but it didn't quite work the same, it seemed the `race` was a blocking rather than async task...maybe i wasn't doing it right)

packages/mobile/src/navigator/NavigationService.ts
jeanregisser: That's fine lol. The fullscreen signing stuff is only temporary 😄 
(well everything temporary stays for a while in reality).

jeanregisser (approved):
Great! 👍

packages/mobile/e2e/src/usecases/NewAccountPhoneVerification.js
```diff
@@ -59,6 +52,13 @@ export default NewAccountPhoneVerification = () => {
 
   // Check that Twilio SID, Auth Token and Verification Phone Number are defined
   if (TWILIO_ACCOUNT_SID && TWILIO_AUTH_TOKEN && VERIFICATION_PHONE_NUMBER) {
+    // Log Twilio balance at start
```
kathaypacific, line 55: @MuckT @jeanregisser i moved the twilio balance check into this conditional, to avoid the failure we saw yesterday. hope this is sufficient for now...

packages/mobile/src/app/WebViewScreen.tsx
```diff
@@ -53,6 +68,20 @@ function WebViewScreen({ route, navigation }: Props) {
     })
   }, [navigation])
 
+  useEffect(() => {
+    return () => {
+      handleEndDappSession()
+    }
+  }, [])
```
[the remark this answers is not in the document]
kathaypacific: ah yes you're right, this is subtle so i've added a comment :)

packages/mobile/src/app/WebViewScreen.tsx
```diff
@@ -5,42 +5,57 @@ import Refresh from '@celo/react-components/icons/Refresh'
...
   const [canGoForward, setCanGoForward] = useState(false)
 
+  const handleEndDappSession = () => {
+    if (activeDapp) {
+      dispatch(dappSessionEnded())
+    }
+  }
+
+  const handleCloseWebView = () => {
+    handleEndDappSession()
```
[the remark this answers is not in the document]
kathaypacific: done!

ValoraQA:
Hey @MuckT, We have verified above task on Latest Test Flight Release build V1.29.2, Android Internal Release Build V1.29.2, Test Flight Alfajores Build V 1.29.2 and Android Alfajores Build V 1.29.2 & updated status in test rails.

**Observations:**

- ”QR code read fail.." error is displayed when user scan QR code using app.ubeswap.org from the web browser.
- "QR code is not generated when user browse https://www.impactmarket.com/ in Web browser while tapping on Wallet connect button, below is the link:
**Reference link:** [Link](https://github.com/valora-inc/wallet/issues/2237)

- "Blank screen is displayed" when user redirects to Poof and Talent Protocol Dapp from Dapp page for iOS 13 , below is the link.

**Reference link:** [Link](https://github.com/valora-inc/wallet/issues/1351)

**Test rail link:**

**Android 11:** https://valoraapp.testrail.io/index.php?/tests/view/20920
**Android 12:** https://valoraapp.testrail.io/index.php?/tests/view/20983
**iOS13:** https://valoraapp.testrail.io/index.php?/tests/view/20857
**iOS14:** https://valoraapp.testrail.io/index.php?/tests/view/20794

**Devices:** Google Pixel 2XL (11.0), iPhone SE 2020 (14.8.1), iPhone 7 (13.3),  Google Pixel 4a (12.0)
**CC:** @kathaypacific 

Thanks..!!

Issue #2108, opened 2022-03-08
[M2] Back Button should not go to SyncBankAccount


jh2oman:
Currently, When you go through the flow of adding a bank account, the user goes from the SyncBankAccount screen to the BankAccounts screen. The back button currently takes the user back to the SyncBankAccount screen which will leave them in a dead end that they can't get out of.

AC:
* Back button on the BankAccounts screen does not go to the SyncBankAccount screen.

ENG:
* Maybe the back button goes two screens back or to the settings page if the screen before is SyncBankAccount

PR #2161, opened 2022-03-15
Fixing various things for sync bank account screen

lisaqyli:
### Description
- Fixes  #2110 , also fixes the string on link bank account error screen

### Tested
unit tested / manual tested

PR #2182, opened 2022-03-18
add custom navigator options for bank account page

lisaqyli:
### Description
Fixes: #2108 
I separate out the fix for the navigation from this PR: https://github.com/valora-inc/wallet/pull/2161
I found out the way to clear navigation stack and replace with another screen does not work in this case b/c Settings screen does not belong to the settingsStack (defined in Navigator.tsx). It's already defined in DrawerNavigator so initiate another instance of Settings  would cause trouble. so solving the problem by customizing the header of bankAccount screen instead

### Tested

Manually and unit tested.


jh2oman (commented):
@lisaqyli Does this also cover the "swipe back" on iOS or the hardware back button on Android? I'm worried users will still be able to get to the SyncBankAccount screen using those features

lisaqyli:
@jh2oman no it doesn't.  I just pushed something to prevent hardware press back. It works well on Android but it's a bit suboptimal on iOS (it would disable all swipe backs even if user comes from settings page).

wdyt?

jh2oman:
@lisaqyli I feel like there must be a better way of going about it. Can you ask people on the retentions squad? I'm really not familiar enough with these react native navigation apis to suggest something

jh2oman (approved):

MuckT:
@lisaqyli Does this need to be tested by QA during the release cycle?

Issue #2176, opened 2022-03-17
 “There was an issue claiming your supercharge rewards.” error appears in red banner when user taps on Claim Rewards button from “Claim Rewards” screen.


ValoraQA:
**Frequency:**100%

**Repro on build version:** Android Internal Release Build V1.28.1, Test flight Release Build V1.28.1 
**Repro on devices:** Google Pixel XL (10.0), iPhone 13 mini (15.1.1), Google Pixel 2XL (11.0), iPhone 12 (14.7.1)

**Testing Account:** +12152686931/[Account Key](https://docs.google.com/spreadsheets/d/1QsoRgyn9vUDZWboh7gcG8apG3FQRXHTLzLxt2u7mrio/edit#gid=0&range=C2 )

**Pre-condition:** 
1.User must have complete the onboarding flow.
2.User must have received rewards & “Supercharge rewards are here!..” CTA must be present.

**Repro Steps:**
1.Tap on “Claim rewards” button from “Supercharge rewards are here!..” CTA in home screen.
2.Tap on “Claim rewards” button from “You have 0.01 cUSD available…” screen & observe

**Bug:**  “There was an issue claiming your supercharge rewards.” error appears in red banner when user taps on Claim Rewards button from “Claim Rewards” screen.

**Expected Behavior:** No error message must appear when user  taps on Claim Rewards button from “Claim Rewards” screen & user should be able to claim rewards successfully.

**Investigation:** This issue is not reproducible for both iOS and Android platform for some [Account](https://docs.google.com/spreadsheets/d/1QsoRgyn9vUDZWboh7gcG8apG3FQRXHTLzLxt2u7mrio/edit#gid=0&range=C4 ) 

**Impact:** User will be blocked to claim rewards received.

**Attachment:** [Error on claiming rewards.mp4](https://drive.google.com/file/d/1Vf8W-PzErvonKjwQ01YnzUSkAvPv9yR1/view?usp=sharing )

ValoraQA:
Hey @MuckT, We have verified this issue on Latest iOS App store build V1.32.2 & observed the following

- "Error while fetching pending Supercharge rewards.." error in red banner appears on home screen when user clicks on "Claim rewards" button from supercharge rewards screen.

**NOTE:** The above issue is not reproducible on android platform.

**Device:** iPhone 12 mini (15.5)
[...]

PR #2201, opened 2022-03-22
fix: Claim  Supercharge rewards with stablecoins

### LINKED ISSUE #2219: fix: Claim Supercharge rewards with stablecoins
There's a bug in the Supercharge claim feature that is causing it to only claim using CELO. Switch to use any token that the user has

gnardini:
### Description

There's a bug in the Supercharge claim feature that is causing it to only claim using CELO. Switched to use any token that the user has

### Other changes

N/A

### Tested

Manually with a manual distribution.

### Related issues

- Fixes #2219

### Backwards compatibility

N/A


kathaypacific (approved):
LGTM 🔥

jeanregisser (approved):
LGTM 👍

MuckT:
This should also fix: #2176.

PR #2220, opened 2022-03-22
Block finclusive unsupported geos 

### LINKED ISSUE #1999: [M2] block unsupported geos from connecting bank accounts
We need to make sure the link bank account step is only visible to people who should be able to access it (residents of US states except NY and TX, where Finclusive doesn't have licensing yet). 

After the user completes Persona, we should be able to access the state they are from and tell whether they should be able to continue onwards. 

Eng:
- Firebase remote config lets us restrict the feature to US only, but doesn't give us state-level granularity
- this should be done by adding an IHL endpoint to query for whether the user should be able to link their bank account (based on their Persona data)

Finclusive confirmed [here](https://valora-app.slack.com/archives/C01GK7W16BF/p1644535962091669) they don't cover US territories.

AC:
- link bank account flow only visible to states with Finclusive coverage (shows some message like "your region coming soon" otherwise (TODO get a design for this)

lisaqyli:
### Description
- Fixes #1999 
Finclusive doesn't have license in NY and TX yet. After collecting information from persona, we should prevent user from connecting a bank account if their residence is in NY or TX. 

WIP:
- finalize the copy/design with Nitya
- add tests

### Tested
WIP

### DIFF
--- a/packages/mobile/src/account/LinkBankAccountScreen.tsx
+++ b/packages/mobile/src/account/LinkBankAccountScreen.tsx
@@ -7,6 +7,7 @@ import * as React from 'react'
 import { useEffect, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import { ScrollView, StyleSheet, Text, View } from 'react-native'
+import { InquiryAttributes } from 'react-native-persona'
 import { useDeepLinkRedirector, usePlaidEmitter } from 'react-native-plaid-link-sdk'
 import { SafeAreaView } from 'react-native-safe-area-context'
 import { useDispatch, useSelector } from 'react-redux'
@@ -113,6 +114,10 @@ export function stepOneUIState({
   return StepOneUIState.Begin
 }
 
+// Persona data is parsed from step 1 to see whether the user's resident region is supported
+// This is a shared attribute between step 1 and step 2
+let isRegionSupported = false
+
 export function StepOne() {
   const { t } = useTranslation()
   const dispatch = useDispatch()
@@ -146,7 +151,28 @@ export function StepOne() {
     // Add a bit of a delay so that Persona can popup before switching to the spinner view
     setTimeout(() => setIsInPersonaFlow(true), 500)
   }
-  const onSuccessPersona = () => {
+
+  const isUserRegionSupported = (address: InquiryAttributes['address']) => {
+    if (!address) {
+      return false
+    }
+    if (!address.countryCode || address.countryCode !== 'US') {
+      return false
+    }
+
+    const UNSUPPORTED_REGION_ABBR: string[] = ['NY', 'TX']
+    if (!address.subdivisionAbbr || UNSUPPORTED_REGION_ABBR.includes(address.subdivisionAbbr)) {
+      // Finclusive currently do not support residents of NY and TX
+      return false
+    }
+    return true
+  }
+
+  const onSuccessPersona = (attributes: InquiryAttributes) => {
+    if (isUserRegionSupported(attributes.address)) {
+      isRegionSupported = true
+    }
+
     setSuccessFromPersona(true)
     setIsInPersonaFlow(false)
   }
@@ -252,7 +278,7 @@ export function StepTwo() {
   const { t } = useTranslation()
   const finclusiveKycStatus = useSelector(finclusiveKycStatusSelector)
   const plaidParams = useSelector(plaidParamsSelector)
-  const stepTwoEnabled = useSelector(linkBankAccountStepTwoEnabledSelector)
+  const stepTwoEnabled = useSelector(linkBankAccountStepTwoEnabledSelector) && isRegionSupported
   const disabled = !stepTwoEnabled || finclusiveKycStatus !== FinclusiveKycStatus.Accepted
   // This is used to handle universal links within React Native
   // https://plaid.com/docs/link/oauth/#handling-universal-links-within-react-native
--- a/packages/mobile/src/account/LinkBankAccountScreen.tsx
+++ b/packages/mobile/src/account/LinkBankAccountScreen.tsx
@@ -26,7 +26,11 @@ import VerificationDenied from 'src/icons/VerificationDenied'
 import VerificationPending from 'src/icons/VerificationPending'
 import { navigate } from 'src/navigator/NavigationService'
 import { Screens } from 'src/navigator/Screens'
-import { linkBankAccountStepTwoEnabledSelector } from '../app/selectors'
+import { isUserRegionSupportedByFinclusive } from 'src/utils/supportedRegions'
+import {
+  finclusiveUnsupportedStatesSelector,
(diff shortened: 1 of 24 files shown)


packages/mobile/test/schemas.ts
```diff
@@ -1156,6 +1156,22 @@ export const v42Schema = {
   },
 }
 
+export const v43Scheme = {
```
jh2oman, line 1159:
```suggestion
export const v43Schema = {
```

packages/mobile/src/account/LinkBankAccountScreen.tsx
```diff
@@ -146,7 +156,30 @@ export function StepOne() {
...
   }
-  const onSuccessPersona = () => {
+
+  const isUserRegionSupported = (address: InquiryAttributes['address']) => {
+    if (!address) {
+      Logger.error(TAG, 'Address field does not exist in Persona inquiry attributes')
+      return false
+    }
+    if (!address.countryCode || address.countryCode !== 'US') {
+      Logger.info(TAG, 'User region country code not in US')
```
cajubelt, line 166: this log is misleading for the case where `address.countryCode` is falsy.

packages/mobile/src/account/LinkBankAccountScreen.tsx
```diff
@@ -146,7 +156,30 @@ export function StepOne() {
...
+      return false
+    }
+
+    if (!address.subdivisionAbbr || UNSUPPORTED_REGION_ABBR.includes(address.subdivisionAbbr)) {
+      // Finclusive currently do not support residents of NY and TX
+      Logger.info(TAG, 'User region state not supported by finclusive')
```
cajubelt, line 172: same here, misleading if `address.subdivisionAbbr` is just falsy. In that case I'd expect to see a log like "missing field from address: `subdivisionAbbr`"

packages/mobile/src/account/LinkBankAccountScreen.tsx
```diff
@@ -146,7 +156,30 @@ export function StepOne() {
     // Add a bit of a delay so that Persona can popup before switching to the spinner view
     setTimeout(() => setIsInPersonaFlow(true), 500)
   }
-  const onSuccessPersona = () => {
+
+  const isUserRegionSupported = (address: InquiryAttributes['address']) => {
```
cajubelt, line 160: if this were taken outside of the StepOne function, it would be easier to unit test

packages/mobile/src/account/LinkBankAccountScreen.tsx
```diff
@@ -251,8 +284,10 @@ export function StepOne() {
 export function StepTwo() {
   const { t } = useTranslation()
   const finclusiveKycStatus = useSelector(finclusiveKycStatusSelector)
+  const finclusiveRegionSupported = useSelector(finclusiveRegionSupportedSelector)
   const plaidParams = useSelector(plaidParamsSelector)
-  const stepTwoEnabled = useSelector(linkBankAccountStepTwoEnabledSelector)
+  const stepTwoEnabled =
```
cajubelt, line 289: interesting.. so this means when the user's geo isn't supported they'll see "We successfully verified your identity. We'll let you know when you can finish linking your bank account."

cajubelt (commented):
looks pretty good :)

packages/mobile/test/schemas.ts
lisaqyli: haha good catch

packages/mobile/src/utils/supportedRegions.ts
```diff
@@ -0,0 +1,19 @@
+import { InquiryAttributes } from 'react-native-persona'
+
+export function isUserRegionSupportedByFinclusive(
```
cajubelt, line 3: I prefer the way this was implemented before, when it just returned false if the region wasn't supported.

Three reasons not to throw an error for unsupported regions:
- Runtime errors should be reserved for unexpected behavior. A person being from New York is expected, not a reason to sound the alarm!
- helper functions should be self-documenting and have predictable outputs. A function called `isUserRegionSupportedByFinclusive` that returns a boolean should return false when the user is not in a region supported, because that's what the name strongly suggests it would do. 
- in general, exceptions should not be used for control flow

packages/mobile/src/account/LinkBankAccountScreen.tsx
cajubelt: one option could be to refactor the selector `linkBankAccountStepTwoEnabledSelector` to return the value of the feature flag && finclusiveRegionSupported

cajubelt: since I'm not sure there's a use case anymore for just checking the value of the feature flag in isolation

packages/mobile/src/account/reducer.ts
```diff
@@ -294,6 +296,12 @@ export const reducer = (
         finclusiveKycStatus: action.finclusiveKycStatus,
       }
     }
+    case Actions.SET_FINCLUSIVE_REGION_SUPPORTED: {
+      return {
+        ...state,
+        finclusiveRegionSupported: true,
```
cajubelt, line 302: shouldn't it be possible to set this to false, too? like with `finclusiveRegionSupported: action.finclusiveRegionSupported`?

cajubelt (commented):
noticed a potential issue with Settings, and setting region not supported to false

packages/mobile/src/account/reducer.ts
lisaqyli: the default is false. the prop can only be set to true during the persona onSuccess callback.

cajubelt: Yes I understand that, I was suggesting that we make it possible to set to either true or false.

But whatever, I don't care that much

packages/mobile/src/utils/supportedRegions.ts
cajubelt: still don't think we should throw errors for non-error cases here

cajubelt (commented):
looks like there's one thing you may have missed from my last review.

otherwise, looks good

packages/mobile/src/utils/supportedRegions.ts
```diff
@@ -0,0 +1,19 @@
...
+  }
+  if (address.countryCode !== 'US') {
+    return false
+  }
+
+  if (unsupportedRegions.includes(address.subdivisionAbbr)) {
+    // Finclusive currently do not support residents of NY and TX
+    return false
+  }
+  return true
```
cajubelt, line 18:
```suggestion
  return address.countryCode === 'US' && !unsupportedRegions.includes(address.subdivisionAbbr)
```

cajubelt: author's choice

cajubelt (approved):

PR #2221, opened 2022-03-23
refactor(IHL): use EOA to identify user for IHL (not MTW)

### LINKED ISSUE #2030: [M1] Use EOA as identifier throughout PFP integration in the wallet

cajubelt:
### Description

Switches to EOA to identify users for IHL, rather than MTW. This allows us to use auth for IHL that is "more general" (aka the same that FiatConnect will use)

### Other changes

na

### Tested

automated tests pass

manual test passes (just went thru persona flow and made sure the link bank account steps page updated)

### How others should test

na (behind feature flag)

### Related issues

- Fixes #2030 

### Backwards compatibility

na


packages/mobile/src/account/BankAccounts.test.tsx
```diff
@@ -10,7 +10,7 @@ import { deleteFinclusiveBankAccount, getFinclusiveBankAccounts } from 'src/in-h
 import { navigate } from 'src/navigator/NavigationService'
 import { Screens } from 'src/navigator/Screens'
 import { createMockStore, getMockStackScreenProps } from 'test/utils'
-import { mockAccount, mockNavigation, mockPrivateDEK } from 'test/values'
```
jh2oman, line 13: just curious, why not use `mockAccount`?

jh2oman (approved):
Looks great!

ValoraQA:
Hey @MuckT  do we need to test the above PR as it is behind a feature flag, please let know us if we need to test it

PR #2227, opened 2022-03-23
Add analytics for closing dapps web view

### LINKED ISSUE #2104: Adjust analytics for dapps opened in the internal WebView

kathaypacific:
### Description

The only event that needs to be added for the in-app dapps web view is one to track the web view close.

As we want to keep the same properties as the other dapp events here, we need to know which section the web view was launched from. I realised that the previous implementation to derive this information was not 100% correct (e.g. the dapp is recorded as being launched from the "featured" section if it is featured, even if it is launched from inside the normal dapps list). Most changes in this PR is to update where the "section" information is coming from.

### Other changes

N/A

### Tested

Manually and updated unit tests

### How others should test

This does not need testing

### Related issues

- Relates #2104 

### Backwards compatibility

Yes


gnardini (approved):
🙌

jeanregisser (approved):
LGTM! 👍 

Just one note, this doesn't entirely close https://github.com/valora-inc/wallet/issues/2104, right?
I think we also want to update the existing WC/DappKit events to know whether they happened  from the external browser vs in-app one.

ValoraQA:
Hey @MuckT  JFYI: We are skipping this task as in "How others should test" section it is mentioned N/A.
Cc: @kathaypacific 
thanks !!

PR #2251, opened 2022-03-25
(feat): Transaction feed paginated

### LINKED ISSUE #1808: Add pagination in the new TransactionFeed.
The current status is that the blockchain-api returns the first 100 transactions when a user requests the feed. So, we need to add pagination, to allow the user to see more than 100 transactions.

In order to do it, we need to start using the pagination of blockscout (see the docs at https://explorer.celo.org/graphiql), and add pagination to our endpoint (tokenTransactionsV2).

Then we need to start consuming the endpoint with pagination from the app. (Should we do this in another issue?)

dievazqu:
### Description

When scrolling to the bottom of the transaction feed, the app requests more transactions to the blockchain-api.

### Tested

Locally with an emulator, hitting local blockchain-api with [these](https://github.com/valora-inc/blockchain-api/pull/133) changes.

### How others should test

Until blockchain-api [PR](https://github.com/valora-inc/blockchain-api/pull/133) is merged, this can't be tested without setting up the blockchain-api service.
Once the blockchain-api PR is merged, you should be able to scroll to the bottom of the transactions feed and new transactions will be shown (if there are any).

### Related issues

- Fixes #1808

### Backwards compatibility

This app is not compatible with the current version of blockchain-api. We have to deploy blockchain-api service before this is released in order to avoid any issue.

### Screen Recording

Scrolling until no more transactions are available:
https://user-images.githubusercontent.com/12504748/163847783-3fb529d0-aa69-4ce1-b589-43a2edeb740d.mov

Network issues at the beginning:
https://user-images.githubusercontent.com/12504748/163847977-4568bd5e-3924-4507-a4ec-95480c2e5bdd.mov

Network issues when some pages were already fetched:
https://user-images.githubusercontent.com/12504748/163848124-3a8d0848-1c6b-4c25-b935-1915a140de81.mov


packages/mobile/src/transactions/feed/TransactionFeed.tsx
```diff
@@ -153,7 +210,7 @@ function TransactionFeed() {
     }
 
     return groupFeedItemsInSections(tokenTransactions)
-  }, [tokenTransactions])
+  }, [JSON.stringify(tokenTransactions.map((tx) => tx.transactionHash))])
```
kathaypacific, line 213: is there an edge case where `tokenTransactions` alone doesn't refresh this `useMemo` value? i think dependency arrays are evaluated every rerender and `JSON.stringify` might cause performance issues if the user does have a large transaction feed + low spec phone?

packages/mobile/src/transactions/feed/TransactionFeed.tsx
```diff
@@ -131,11 +93,106 @@ function mapStandbyTransactionToFeedTokenTransaction(tx: StandbyTransaction): Fe
...
+    dependencies: [counter, fetchMoreTransactions, pageInfo?.endCursor],
+    precondition: () => {
+      return fetchMoreTransactions && !!pageInfo && pageInfo?.hasNextPage
+    },
+    onSuccess: (result) => {
+      // TODO: Show a message if we know there aren't more transactions
+      setFetchMoreTransactions(false)
+
+      if (result) {
+        if (result?.data?.tokenTransactionsV2?.transactions.length) {
```
kathaypacific, line 176: nit: don't need the `?` after `result` since it's already inside the if statement. what do you think about using `if (result?.data)` instead as the if condition? `result` should always exist for graphql right?

packages/mobile/src/transactions/feed/TransactionFeed.tsx
```diff
@@ -131,11 +93,106 @@ function mapStandbyTransactionToFeedTokenTransaction(tx: StandbyTransaction): Fe
...
+          addFetchedTransactions(result.data.tokenTransactionsV2.transactions)
+        }
+
+        setPageInfo(result?.data?.tokenTransactionsV2?.pageInfo)
```
kathaypacific, line 180: should this part also be inside the if statement? IIRC graphql also comes into this `onSuccess` callback even if there is an error - in that case we'd be setting `undefined` for `pageInfo` which might be what we want but would be nice to be explicit if it is?

packages/mobile/src/transactions/feed/TransactionFeed.tsx
```diff
@@ -131,11 +93,106 @@ function mapStandbyTransactionToFeedTokenTransaction(tx: StandbyTransaction): Fe
...
+      }
+    },
+    onError: (error) => {
+      // TODO: show a message indicating there was an error
+      setFetchMoreTransactions(false)
+      Sentry.captureException(error)
```
kathaypacific, line 186: any reason why we are calling sentry here rather than going through `Logger.error` which does this under the hood?

packages/mobile/src/transactions/feed/TransactionFeed.tsx
```diff
@@ -131,11 +93,106 @@ function mapStandbyTransactionToFeedTokenTransaction(tx: StandbyTransaction): Fe
...
+    setFetchedTransactions
+  )
+
+  const dispatch = useDispatch()
+
+  // Update the counter variable every |POLL_INTERVAL| so that a query is made to the backend.
+  const [counter, setCounter] = useState(0)
+  useInterval(() => setCounter((n) => n + 1), POLL_INTERVAL)
+
+  const { loading, error } = useAsyncQueryTransactionsFeed({
```
kathaypacific, line 135: i'm confused about why this hook exists twice in this component, is this intentional? is it possible to have it called only once?

packages/mobile/src/transactions/feed/TransactionFeed.tsx
```diff
@@ -131,11 +93,106 @@ function mapStandbyTransactionToFeedTokenTransaction(tx: StandbyTransaction): Fe
...
 function TransactionFeed() {
+  const address = useSelector(walletAddressSelector)
+  const localCurrencyCode = useSelector(getLocalCurrencyCode)
+
+  const [pageInfo, setPageInfo] = useState<PageInfo | null>(null)
+  const [fetchMoreTransactions, setFetchMoreTransactions] = useState(false)
+
   const cachedTransactions = useSelector(transactionsSelector)
+  const [fetchedTransactions, setFetchedTransactions] = useState<TokenTransaction[]>([])
+  const addFetchedTransactions = addFetchedTransactionFunction(
```
kathaypacific, line 124: it seems like we are keeping `fetchedTransactions, setFetchedTransactions` in this component mainly to feed it into this hook. is it possible to change the interface of this hook to be something like `const [fetchedTransactions, addFetchedTransactions] = useFetchedTransactions()`? and this hook can manage the `fetchedTransactions` state variable

packages/mobile/src/transactions/feed/TransactionFeed.tsx
```diff
@@ -131,11 +93,106 @@ function mapStandbyTransactionToFeedTokenTransaction(tx: StandbyTransaction): Fe
...
+    transactionsWithoutDuplicatedHash.sort((a, b) => {
+      return b.timestamp - a.timestamp
+    })
+
+    setFetchedTransactions(transactionsWithoutDuplicatedHash)
+  }
+}
+
 function TransactionFeed() {
+  const address = useSelector(walletAddressSelector)
```
kathaypacific, line 116: is it possible to abstract some logic away to a hook? in the current implementation we have `const { loading, error, transactions } = useQueryTransactionFeed()` - it'd be nice to keep something like this, so that this component can focus on the display part and not worry about coordinating the fetching of data/which page number is next. i'm not sure if you have already tried this and decided against it, i can see why the `pageInfo` part complicates things. thought i'd check anyway :)

packages/mobile/src/transactions/feed/TransactionFeed.tsx
```diff
@@ -131,11 +93,106 @@ function mapStandbyTransactionToFeedTokenTransaction(tx: StandbyTransaction): Fe
...
+function addFetchedTransactionFunction(
+  fetchedTransactions: TokenTransaction[],
+  setFetchedTransactions: any
+) {
+  return (transactions: TokenTransaction[]) => {
+    const currentHashes = new Set(fetchedTransactions.map((tx) => tx.transactionHash))
+
+    const transactionsWithoutDuplicatedHash = fetchedTransactions.concat(
+      transactions.filter((tx) => !currentHashes.has(tx.transactionHash))
+    )
```
kathaypacific, line 105: would it be more efficient to first add the `transactions` to `fetchedTransactions` then deduplicate? i think then we would have less loops but not sure if it achieves exactly the same result?

packages/mobile/src/transactions/feed/queryHelper.ts
```diff
@@ -0,0 +1,154 @@
...
+
+export function useAsyncQueryTransactionsFeed(params: QueryParams) {
+  const {
+    address,
+    localCurrencyCode,
+    afterCursor,
+    dependencies,
+    onSuccess,
+    onError,
+    precondition,
```
kathaypacific, line 22: i was a bit confused about what a precondition is for, it's hard to give this a name 😅 maybe `shouldFetchMore`? or alternatively, since we're passing some specific props already, we could pass the `fetchMoreTransactions` and `pageInfo` directly?

packages/mobile/src/transactions/feed/TransactionFeed.tsx
```diff
@@ -131,11 +93,106 @@ function mapStandbyTransactionToFeedTokenTransaction(tx: StandbyTransaction): Fe
...
+      return b.timestamp - a.timestamp
+    })
+
+    setFetchedTransactions(transactionsWithoutDuplicatedHash)
+  }
+}
+
 function TransactionFeed() {
+  const address = useSelector(walletAddressSelector)
+  const localCurrencyCode = useSelector(getLocalCurrencyCode)
```
kathaypacific, line 117: `address` and `localCurrencyCode` seems to be only used by the `useAsyncQueryTransactionsFeed` hook, could we move these selectors into the hook directly instead of passing them as arguments?

kathaypacific (commented):
I'm really excited about this feature!! looks great 😀  left a few comments, happy to discuss in person too!

packages/mobile/src/transactions/feed/TransactionFeed.tsx
dievazqu: It was done like this, I will use `logger.error` instead :)

packages/mobile/src/transactions/feed/TransactionFeed.tsx
dievazqu: Yes!! I like your idea! :)

packages/mobile/src/transactions/feed/TransactionFeed.tsx
```diff
@@ -131,11 +93,106 @@ function mapStandbyTransactionToFeedTokenTransaction(tx: StandbyTransaction): Fe
   }
 }
 
+function addFetchedTransactionFunction(
+  fetchedTransactions: TokenTransaction[],
+  setFetchedTransactions: any
```
gnardini, line 98: `any`? 👀

packages/mobile/src/transactions/feed/TransactionFeed.tsx
gnardini: I think it's the same? Neither should have a material performance impact I think

packages/mobile/src/transactions/feed/TransactionFeed.tsx
gnardini: I have the same confusion, aren't we making duplicate requests?

packages/mobile/src/transactions/feed/TransactionFeed.tsx
```diff
@@ -131,11 +93,106 @@ function mapStandbyTransactionToFeedTokenTransaction(tx: StandbyTransaction): Fe
...
+  })
 
-  const { loading, error, transactions } = useQueryTransactionFeed()
-  const confirmedTokenTransactions: TokenTransaction[] = transactions ?? cachedTransactions
+  useAsyncQueryTransactionsFeed({
+    address,
+    localCurrencyCode,
+    afterCursor: pageInfo?.endCursor,
+    dependencies: [counter, fetchMoreTransactions, pageInfo?.endCursor],
+    precondition: () => {
```
gnardini, line 168: I agree with @kathaypacific 's comment about calling this `precondition` something like `shouldFetchMore`

gnardini (commented):
Looks good! Just a couple comments, the main one is that I think we are querying from blockchain-api more often than expected

jeanregisser (commented):
I'm late to the review party, but great work! 👍

dievazqu:
@kathaypacific @gnardini @jeanregisser 
Thank you for the reviews!!
The PR is ready to be reviewed again. :)

packages/mobile/src/transactions/feed/queryHelper.ts
```diff
@@ -0,0 +1,269 @@
...
+  return {
+    loading,
+    error,
+    transactions,
+    fetchingMoreTransactions,
+    fetchMoreTransactions: () => {
+      if (!pageInfo) {
+        dispatch(showError(ErrorMessages.FETCH_FAILED))
+      } else if (!pageInfo.hasNextPage) {
+        dispatch(showMessage(t('noMoreTransactions')))
```
gnardini, line 142: This will be called automatically, right? In that case, won't the "No more transactions" message always show up? I can imagine being a bit weird for users with very few transactions to see that message with almost no scrolling. Not a big deal in any case, I just wouldn't show any messages.

packages/mobile/src/transactions/feed/queryHelper.ts
```diff
@@ -0,0 +1,269 @@
...
+  )
+
+  return {
+    loading,
+    error,
+    transactions,
+    fetchingMoreTransactions,
+    fetchMoreTransactions: () => {
+      if (!pageInfo) {
+        dispatch(showError(ErrorMessages.FETCH_FAILED))
```
gnardini, line 140: If a user doesn't have any transactions, won't this error show up? Maybe it just doesn't get called, idk

gnardini (approved):
Two small details/questions, looks great otherwise, love the hook design :)

packages/mobile/src/transactions/actions.ts
```diff
@@ -24,6 +24,7 @@ export enum Actions {
   NEW_TRANSACTIONS_IN_FEED = 'TRANSACTIONS/NEW_TRANSACTIONS_IN_FEED',
   REFRESH_RECENT_TX_RECIPIENTS = 'TRANSACTIONS/REFRESH_RECENT_TX_RECIPIENTS',
   UPDATE_RECENT_TX_RECIPIENT_CACHE = 'TRANSACTIONS/UPDATE_RECENT_TX_RECIPIENT_CACHE',
+  // Store first "page" of transactions to show when app opens before doing any request.
```
kathaypacific, line 27: nit: i don't think this comment needs to be here?

packages/mobile/src/transactions/feed/queryHelper.ts
```diff
@@ -0,0 +1,269 @@
...
+
+export function useFetchTransactions(): QueryHookResult {
+  const { t } = useTranslation()
+  const dispatch = useDispatch()
+  const address = useSelector(walletAddressSelector)
+  const localCurrencyCode = useSelector(getLocalCurrencyCode)
+
+  const [pageInfo, setPageInfo] = useState<PageInfo | null>(null)
+  const [fetchingMoreTransactions, setFetchingMoreTransactions] = useState(false)
+  const [transactions, setTransactions] = useState<TokenTransaction[]>([])
```
kathaypacific, line 69: i had a thought, let me know what you think:

i think `transactions` and `setTransactions` and `addDeduplicatedTransactions` can be abstracted to a hook? the hook could look something like ` const [transactions, setTransactions] = useDeduplicatedTransactions()`, where the hook can manage the `transactions` state variable + a "setTransactions" function for handling logic for deduplicating the transactions. the benefit would be that this `useFetchTransactions` hook has one less state variable to manage (and also doesn't need to pass the outputs of this state hook into another function)

this is a big of code golf, what you have here is fine too, so i'm curious for your thoughts!

packages/mobile/src/transactions/feed/queryHelper.ts
```diff
@@ -0,0 +1,269 @@
...
+
+  const addTransactions = addDeduplicatedTransactions(transactions, setTransactions)
+
+  const handleResult = (result: QueryResponse, paginatedResult: boolean) => {
+    Logger.info(TAG, `Fetched ${paginatedResult ? 'next page' : 'new'} transactions`)
+
+    const returnedTransactions = result.data?.tokenTransactionsV2?.transactions
+    if (returnedTransactions?.length) {
+      addTransactions(returnedTransactions)
+      // We store non-paginated results in redux
```
kathaypacific, line 83: i've been on holiday too long but why do we have non paginated results again? maybe this info can be added to the comment? :)

packages/mobile/src/transactions/feed/queryHelper.ts
kathaypacific: oh yeah this is a good point...just checking, does this "no more transactions" message display as a banner like this? i'd have thought we want a small text at the bottom of the feed rather than a global banner (but i didn't test this, so i might be wrong!)

kathaypacific (approved):
I love the new hook, it's very nice!! can we deploy the update to the blockchain api? can't wait to test this :D

packages/mobile/src/transactions/feed/queryHelper.ts
gnardini: The `showMessage` action does indeed show a blue banner at the top. It's quite easy to miss as well

packages/mobile/src/transactions/feed/queryHelper.ts
dievazqu: We are using those results to show the users the list of transactions without making any request.

ValoraQA:
Hey @MuckT, We have verified above task on Latest Android internal release build , Test flight release build V1.31.0 and updated status in test rails.

**Test rail links:**
**iOS15:** https://valoraapp.testrail.io/index.php?/tests/view/30606
**Android 9:** https://valoraapp.testrail.io/index.php?/tests/view/30637
**iOS 14:** https://valoraapp.testrail.io/index.php?/tests/view/30668

**Devices:** iPhone 13 mini(15.1.1), Redmi Note 5 pro (9.0), iPhone 12(14.7.1)

**CC:** @dievazqu 
Thanks..!!

ValoraQA:
Hey @MuckT , We have verified above task on Latest Android internal release build and updated status in test rails.

Observation : 
- user is able to see the alert incase if the page is not loaded properly due to internet connectivity issues
- User is able to scroll till the last page of transaction loads and activities are seen.

Test rail links:

Android 11: https://valoraapp.testrail.io/index.php?/tests/view/30699

Devices: Google Pixel 2XL(11.0)

CC: @dievazqu 
Thanks..!!

PR #2255, opened 2022-03-27
WC e2e Stability

MuckT:
### Description

A small change to the after all in the wc e2e tests for stability in ci.

### Other changes

N/A

### Tested

Tested in CI.

### How others should test

This does not need to be tested by QA. Reviewers should look at runs of the e2e tests on this branch / pull request.

### Related issues

N/A

### Backwards compatibility

Yes


kathaypacific (commented):
Thanks for looking into this! is there any reason that these functions should be called in the 'afterAll` rather than `beforeAll`? I'm thinking that there's no guarantee that the `afterAll` will be reached, since the tests could fail or be terminated during a run so it would be safer to do cleanup actions before the tests start rather than after the tests finish?

packages/mobile/e2e/src/usecases/WalletConnectV1.js
```diff
@@ -335,6 +335,9 @@ export default WalletConnect = () => {
   })
 
   afterAll(async () => {
+    await device.reloadReactNative()
```
jeanregisser, line 338: Reload RN can create some weird state issues on the native side.
Can we instead relaunch the whole app?

MuckT:
Closing in favor of #2277.

PR #2257, opened 2022-03-28
fix: add a loader while the bank accounts are loading

### LINKED ISSUE #2109: [M2] BankAccounts Screen should have a loading indicator
Currently the BankAccounts screen (where all of your linked accounts are listed) does not have a loading indicator while the API is being called.

AC:
* Add a loading indicator to the BankAccounts screen while IHL is being called

lisaqyli:
### Description
Adding a loader when request to fetch bank accounts is inflight
Fixes #2109 

### Tested
unit tested

### DIFF
--- a/packages/mobile/src/account/BankAccounts.test.tsx
+++ b/packages/mobile/src/account/BankAccounts.test.tsx
@@ -99,6 +99,7 @@ describe('BankAccounts', () => {
     expect(getByTestId('Loader')).toBeTruthy()
     await waitFor(() => expect(getFinclusiveBankAccounts).toHaveBeenCalled())
     await waitFor(() => expect(queryByTestId('Loader')).toBeFalsy())
+    await waitFor(() => expect(queryByTestId('AddAccount')).toBeTruthy())
   })
   it('shows an error when delete bank accounts fails', async () => {
     //@ts-ignore . my IDE complains about this, though jest allows it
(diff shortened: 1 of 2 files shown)


packages/mobile/src/account/BankAccounts.tsx
```diff
@@ -43,6 +51,7 @@ function BankAccounts({ navigation, route }: Props) {
   const [isOptionsVisible, setIsOptionsVisible] = useState(false)
   const [selectedBankId, setSelectedBankId] = useState(0)
   const walletAddress = useSelector(walletAddressSelector)
+  const [isLoadingBankAccounts, setIsLoadingBankAccounts] = useState(false)
```
jh2oman, line 54: Check out the [react-async-hook docs ](https://www.npmjs.com/package/react-async-hook). It comes with loading built in and would probably be a lot cleaner than adding in new state

packages/mobile/src/account/BankAccounts.test.tsx
```diff
@@ -89,6 +89,17 @@ describe('BankAccounts', () => {
...
+  it('shows a loading circle when getFinclusiveBankAccounts is inflight', async () => {
+    const { getByTestId, queryByTestId } = render(
+      <Provider store={store}>
+        <BankAccounts {...mockScreenProps} />
+      </Provider>
+    )
+
+    expect(getByTestId('Loader')).toBeTruthy()
+    await waitFor(() => expect(getFinclusiveBankAccounts).toHaveBeenCalled())
+    await waitFor(() => expect(queryByTestId('Loader')).toBeFalsy())
```
cajubelt, line 101: if you added an assertion that `AddAccount` is visible here it would probly bring the patch coverage up to where it's supposed to be

packages/mobile/src/account/BankAccounts.tsx
lisaqyli: oh did not know about this, good suggestion

jh2oman (approved):

cajubelt (approved):

PR #2260, opened 2022-03-28
refactor: removed req to verify phone before linking bank account

### LINKED ISSUE #2031: [M1] Drop phone number verification requirement
Since we no longer require the user to have an MTW address registered, which was a byproduct of the phone number verification process, we no longer require users to verify their phone number before going through PFP. This ticket is for dropping the phone number verification requirement from the flow, and removing any CICO events associated with it.

AC:
* Phone verification dropped as requirement from PFP
* Associated events are removed from the wallet

cajubelt:
### Description

A user no longer has to verify their phone number before going thru the PFP flow, since we use the EOA to identify them (rather than the MTW).

### Other changes

Also removed event for when user tries to link their bank account but their phone number is not connected

### Tested

manually checked that I could:
- access the link bank account steps page (without first verifying my phone number)
- go thru Persona KYC
- see my persona kyc update in firebase 

### How others should test

na (behind feature flag still)

### Related issues

- Fixes #2031 

### Backwards compatibility

na

### DIFF
--- a/packages/mobile/locales/base/translation.json
+++ b/packages/mobile/locales/base/translation.json
@@ -1203,7 +1203,7 @@
   "nomSpaceRecipient": "Nomspace: {{name}}",
   "lastDay": "Last 24 hrs",
   "merchant": "Merchant Payment",
-  "merchantMoneyEscrow": "You're about the send a direct payment to a merchant.",
+  "merchantMoneyEscrow": "You're about to send a direct payment to a merchant.",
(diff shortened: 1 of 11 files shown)


jh2oman (approved):
🙌

lisaqyli (approved):

ValoraQA:
Hey @MuckT do we need to test the above PR as it is behind a feature flag, please let know us if we need to test it

PR #2265, opened 2022-03-29
Display invitee from contact list in home feed, rather than "unknown"

### LINKED ISSUE #2215: [Invite v2] Invitee and inviter details correct
- [ ] Ensure that the invitee and inviter details are correct in the transaction. From the app, query blockchain api to get txhash, get the details of the tx, use the id of the invited person from the tx to match to the contact phone number from LocalStorage

kathaypacific:
### Description

Currently the transaction item for an invite sent displays as from "Unknown" because the transaction is going to the escrow contract address instead of a known contact address.

In this PR:
- add redux variable to track invitation transaction details. due to the transaction feed being updated regularly, it is nice to keep this persisted redux variable separately so that we don't need to refresh the invite details
- if the transaction feed item is a sent invite, get the recipient info from the escrow contract
- when displaying the transaction feed item, use the fetched recipient details from the escrow contract

Screenshot 2022-03-29 at 16 24 51

Note: we still rely on the contacts cache to display the invitee name, this cache is only refreshed when the user accesses their contacts (e.g. when on the send screen) - this means the invitee will still display as "Unknown" if the user has reinstalled the app/has not accessed their contacts list yet.

### Other changes

N/A

### Tested

Manually, unit tests to come

### How others should test

Send an invitation, then observe the transaction item on the home feed should display the correct invitee name for the transaction

### Related issues

- Fixes #2215 

### Backwards compatibility

Yes


packages/mobile/src/recipients/reducer.ts
```diff
@@ -44,7 +44,6 @@ export const recipientsReducer = createReducer(initialState, (builder) => {
         ...initialState,
         ...state,
         ...hydrated,
-        phoneRecipientCache: initialState.phoneRecipientCache,
```
kathaypacific, line 47: this was causing the cache to be lost after every app restart which is undesirable for a cache i think...

packages/mobile/src/transactions/feed/TransferFeedItem.test.tsx
```diff
@@ -328,6 +328,48 @@ describe('TransferFeedItem', () => {
...
+            [contactPhoneNumber]: 'hmq1hMZ08BYOg',
+          },
+        },
+      },
+    })
+
+    expectDisplay({
+      getByTestId,
+      expectedTitleSections: [
+        `feedItemEscrowSentTitle, {"context":null,"nameOrNumber":"${contactName}"}`,
```
jeanregisser, line 365: nit: it's better to have hardcoded expectations to minimize room for errors:
```suggestion
        `feedItemEscrowSentTitle, {"context":null,"nameOrNumber":"Kate Bell"}`,
```

packages/mobile/src/transactions/saga.ts
```diff
@@ -220,6 +282,7 @@ function* addRecipientProfiles({ transactions }: NewTransactionsInFeedAction) {
 function* watchNewFeedTransactions() {
   yield takeEvery(Actions.NEW_TRANSACTIONS_IN_FEED, cleanupStandbyTransactionsLegacy)
   yield takeEvery(Actions.UPDATE_TRANSACTIONS, cleanupStandbyTransactions)
+  yield takeLatest(Actions.UPDATE_TRANSACTIONS, getInviteTransactionsDetails)
```
jeanregisser, line 285: Why not `takeEvery`?

packages/mobile/src/transactions/saga.ts
```diff
@@ -82,6 +88,62 @@ function* cleanupStandbyTransactions({ transactions }: UpdateTransactionsAction)
...
+
+export function* getInviteTransactionsDetails({ transactions }: UpdateTransactionsAction) {
+  const existingInviteTransactions = yield select(inviteTransactionsSelector)
+  const newInviteTransactions = transactions.filter(
+    (transaction) =>
+      transaction.type === TokenTransactionTypeV2.InviteSent &&
+      !existingInviteTransactions[transaction.transactionHash]
+  )
+
+  if (newInviteTransactions.length > 0) {
```
jeanregisser, line 128: nit: it's nice to return early to avoid extra indentation.
```suggestion
  if (newInviteTransactions.length <= 0) { return: }
```

jeanregisser (approved):
Awesome work! 💯 
This regression had been bothering me for a while. great to see it fixed!!

packages/mobile/src/transactions/feed/TransferFeedItem.test.tsx
kathaypacific: fair enough, I updated!

packages/mobile/src/transactions/saga.ts
kathaypacific: hum good question...i was thinking that there shouldn't be multiple UPDATE_TRANSACTIONS actions dispatched in quick succession/in parallel, and if there are then the most recent action would contain the information from previous actions (e.g. the transactions fetched at 10:01am would contain the same transactions fetched at 10:00am) so it's safe to take only the latest action...

but now as i am typing this, i remember that we have paginated transaction feed coming soon so maybe it is safer to do takeEvery after all...

packages/mobile/src/transactions/saga.ts
```diff
@@ -82,6 +88,62 @@ function* cleanupStandbyTransactions({ transactions }: UpdateTransactionsAction)
   }
 }
 
+function* getInviteTransactionDetails(txHash: string, blockNumber: string) {
```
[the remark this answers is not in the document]
kathaypacific: yeah i also was thinking about this...but would we then need to query the escrow contract a lot more often? with this client side approach, we only need to query the escrow contract once per invite and we can store the invite details locally for display the next time. but if we want this to be returned in the transaction feed query, would the BE have to fetch this information every time because the api is stateless? maybe @gnardini you have some thoughts on this

packages/mobile/src/transactions/saga.ts
kathaypacific: that's true, updated

ValoraQA:
Hey @MuckT, We have verified above task on Latest Test Flight Release build V1.30.1 and updated status in test rails.

**Observation:**
- User is able to see the correct invitee name for the transaction for saved contact.

**Test rail link:**
**iOS15:** https://valoraapp.testrail.io/index.php?/tests/view/27232

**Device:** iPhone 13 mini(15.1.1)

**CC:** @kathaypacific 

Thanks..!!

ValoraQA:
Hey @MuckT, We have verified above task on Latest Test Flight Release build V1.30.1 and updated status in test rails.

Observation:

User is able to see the correct invitee name for the transaction for saved contact.
Test rail link:
iOS14: https://valoraapp.testrail.io/index.php?/tests/view/27325

Device: iPhone 12(14.7.1)

CC: @kathaypacific

Thanks..!!

PR #2277, opened 2022-03-30
E2E Refactor

### LINKED ISSUE #2273: Run e2e tests locally with one command
### Description

It was mentioned at retro that it is difficult to run the e2e tests locally, some team members have lost whole days trying to do this. To reduce the friction and improve developer experience, minimise the amount of manual set up work that needs to be done.

### Acceptance criteria
- [ ] Set up and launch the end to end tests locally with minimal steps, ideally with one command 
- [ ] Review existing README instructions and ensure they are up to date

MuckT:
### Description

- E2E test in CI no longer use `./run_e2e.sh` and instead are managed entirely through GitHub actions.
- E2E tests run locally no loger use `./run_e2e.sh` and instead are managed by scripts in the `packages/mobile/package.json`.
- Android Emulators use a Snapshot with [Test Butler](https://github.com/linkedin/test-butler) on Android Emulators to prevent QuickStep ANR's.
- Android and iOS Emulators are now launched through Detox as opposed to being launched via `./run_e2e.sh`.
- Photo screen comparisons have been removed as they create complexity for running on multiple device if misconfigured or without branding for external contributors who do not have access to branding.
- Updated e2e `README.md` to reflect changes in running tests locally.

### Other changes

Detox Updated to 19.5.7.

### Tested

Tested in CI.

### How others should test

Reviewers should create a Detox build for iOS and Android and run the tests locally by flowing the updated e2e readme.

### Related issues

Fixes #2273 

### Backwards compatibility

Yes


packages/mobile/e2e/README.md
```diff
@@ -157,15 +196,16 @@ class ExampleInput extends React.Component {
 
 It is recommended to follow the scheme parentID/ChildDescription.
 
-## Mocks for the e2e tests
+<!-- ## Mocks for the e2e tests
```
gnardini, line 199: Why is this section commented? Should we uncomment or delete it?

packages/mobile/e2e/src/usecases/PINChange.js
```diff
@@ -19,13 +18,27 @@ export default ChangePIN = () => {
       .withTimeout(5000)
     await element(by.id('ChangePIN')).tap()
     // Existing PIN is needed first
+    await sleep(500)
```
gnardini, line 21: Why do we need all these sleeps? Can we do `waitFor`s instead or something like that? Plain `sleep`s look like they could cause flakiness :(

packages/mobile/e2e/src/utils/retries.js
```diff
@@ -18,8 +18,10 @@ export const launchApp = async (
         throw error
       }
     },
-    { retries: 5, delay: 10000, timeout: 300000 }
-  )
+    { retries: 5, delay: 10 * 1000, timeout: 30 * 10000 }
+  ).then(async () => {
+    await device.setURLBlacklist(['.*blockchain-api-dot-celo-mobile-alfajores.*'])
```
gnardini, line 23: Nicee, was really looking forward to this! 🙌  I don't think it was working properly before, right?

packages/mobile/e2e/src/utils/utils.js
```diff
@@ -58,20 +57,28 @@ export async function skipTo(nextScreen) {
...
-        await element(by.id(`digit${digit}`))
-          .atIndex(1)
-          .tap()
-      } else {
-        await element(by.id(`digit${digit}`))
-          .atIndex(0)
-          .tap()
+  try {
+    await device.disableSynchronization()
+    await sleep(250)
```
gnardini, line 62: Same comment as before about the `sleep` :(

gnardini (commented):
Couple questions, really looking forward to this!

packages/mobile/e2e/README.md
MuckT: I haven't tried this method of mocking; however, [Detox maintainers have said they are releasing a new mocking guide soon](https://github.com/wix/Detox/pull/3097) so I've commented it out as a reminder to update once the new guide is released. I'd like to covert mocking more in depth once the new Detox mocking guide is released.

packages/mobile/e2e/src/usecases/PINChange.js
MuckT: This is because the new pin entry disabled detox syncing to type in faster; however, I agree a wait for seems more ideal here. I'll update these sleeps to `waitFor`s.

packages/mobile/e2e/src/utils/retries.js
MuckT: It was not working correctly before. I had to do this wrapping to make sure it only fires once the retries is successful. This was what I believe significantly reduced the test times.

packages/mobile/e2e/src/utils/utils.js
MuckT: This is just a small sleep between taps of the pin to give the app a second to register the tap. Without this the pin entry on iOS can take upwards of 5 seconds, and this cuts it down to around 1.5 seconds per pin entry. There might be a way to use the waitFor and the dot that appears after a second; however, this would slow down pin entry on iOS and Android.

packages/mobile/e2e/README.md
MuckT: removed in https://github.com/valora-inc/wallet/pull/2277/commits/a3d63258fe3e1c9da01137d04a0290c6d4c7c60f!

packages/mobile/e2e/src/usecases/PINChange.js
MuckT: Tried with some `waitFor` and adding some new testID's and it was actually slower and flakier. As this all occurring in a section where [device synchronization is disabled](https://github.com/wix/detox/blob/master/docs/APIRef.DeviceObjectAPI.md#devicedisablesynchronization) sleeps should be ok.

packages/mobile/e2e/src/utils/utils.js
MuckT: Commented above with attempts to move to `waitFor`s.

kathaypacific (approved):
nice work!!! i think there are lots of nice improvements in this PR so keen to see it be merged. 

i agree with @gnardini 's point about the sleeps though, the concern would be that the sleeps work only if the CI runners are running as expected - if for whatever reason they're running extra slow, then the sleep amounts we've set will not be sufficient. i wasn't clear on why the `waitFor` would introduce more flakiness, but sounds like you've tried already so happy to see how we go and investigate the waitFor/sleep if we see flakiness :)

ValoraQA:
Hey @MuckT since N/A is mentioned in how other should test , we are skipping this PR

PR #2282, opened 2022-03-31
Remove some multi-token support feature flags

### LINKED ISSUE #2264: Comment in payment deeplinks are not displayed in the description field
### Current behavior

The comment field in payment deeplinks is ignored and not displayed in the description field for confirming the transaction.

### Desired behavior

The comment field in payment deeplinks should be displayed and added to the transaction for tokens supporting transfers with comment (CELO, cUSD, cEUR, cBRL).

### Steps to Reproduce

1. Scan the following QR code:
download (1)
Or deeplink: celo://wallet/pay?address=0x4b371df8d05abd2954564b54faf10b8c8f1bc3a2&displayName=Example%20name&amount=9.50&comment=Burger%20with%20fries&token=cUSD&currencyCode=USD (sorry GitHub doesn't support celo:// links, you'll have to copy and paste that in a note to click it).
2. Notice the expected comment "Burger with fries" is not displayed in the transaction confirmation screen.

### Versions

Not sure exactly when this broke, but it can be reproduced with Valora v1.28 on both iOS and Android

### LINKED ISSUE #2278: Remove remote config and old code for multi token features
### Description
It has been a while since we've launched multi token support, it has been a success and there is no reason to believe that we will roll back to an app without multi token support. There are 3 remote config flags, `multiTokenShowHomeBalances`, `multiTokenUseSendFlow`, `multiTokenUseUpdatedFeed` which can be removed and the accompanying code can also be removed. This reduces our app size, app complexity, less code to maintain.

### AC
- [ ] remove the multi token remote config variables from firebase
- [ ] remove references of multi token remote config variables from the wallet project
- [ ] remove legacy code that will no longer be used

gnardini:
### Description

Removed a couple feature flags related to multi-token support. Most of the code is just this but in one file (`src/send/utils.ts`) there are more changes. There were two issues with deep links:
- The token currency was not being respected. See https://valora-app.slack.com/archives/C02G5197LAC/p1646939499924949
- The comment field was not being passed correctly to the `SendConfirmation` screen. See https://valora-app.slack.com/archives/C029Z1QMD7B/p1648476457612589

Using the deeplink `celo://wallet/pay?address=0xC0509a7dcc69a0B28c7Ca73feD2FF06b9d59E5b9&displayName=Kenneth%20Kangethe&currencyCode=KES&amount=0.1&comment=sending+kes:+0.1+to+My+kotani+wallet` I now get:
Screen Shot 2022-03-31 at 16 41 54

### Other changes

N/A

### Tested

Manually and with unit tests

### How others should test

Send deep links with different values and make sure they all work correctly.

### Related issues

- Fixes #2278
- Fixes #2264
- Fixes #2278 

### Backwards compatibility

N/A

### DIFF
--- a/packages/mobile/src/recipients/reducer.ts
+++ b/packages/mobile/src/recipients/reducer.ts
@@ -44,7 +44,6 @@ export const recipientsReducer = createReducer(initialState, (builder) => {
         ...initialState,
         ...state,
         ...hydrated,
-        phoneRecipientCache: initialState.phoneRecipientCache,
       }
     })
     .addCase(setPhoneRecipientCache, (state, action) => ({
--- a/packages/mobile/src/redux/migrations.ts
+++ b/packages/mobile/src/redux/migrations.ts
@@ -579,4 +579,5 @@ export const migrations = {
       ),
     },
   }),
+  45: (state: any) => state,
 }
--- a/packages/mobile/src/redux/store.test.ts
+++ b/packages/mobile/src/redux/store.test.ts
@@ -93,7 +93,7 @@ describe('store state', () => {
       Object {
         "_persist": Object {
           "rehydrated": true,
-          "version": 44,
+          "version": 45,
         },
         "account": Object {
           "acceptedTerms": false,
@@ -338,6 +338,7 @@ describe('store state', () => {
           "tokenBalances": Object {},
         },
         "transactions": Object {
+          "inviteTransactions": Object {},
           "knownFeedTransactions": Object {},
           "recentTxRecipientsCache": Object {},
           "standbyTransactions": Array [],
--- a/packages/mobile/src/redux/store.ts
+++ b/packages/mobile/src/redux/store.ts
@@ -24,7 +24,7 @@ const persistConfig: PersistConfig<RootState> = {
   key: 'root',
   // default is -1, increment as we make migrations
   // See https://github.com/valora-inc/wallet/tree/main/packages/mobile#redux-state-migration
-  version: 44,
+  version: 45,
   keyPrefix: `reduxStore-`, // the redux-persist default is `persist:` which doesn't work with some file systems.
   storage: FSStorage(),
   blacklist: ['geth', 'networkInfo', 'alert', 'imports', 'supercharge'],
--- a/packages/mobile/src/transactions/actions.ts
+++ b/packages/mobile/src/transactions/actions.ts
@@ -6,6 +6,7 @@ import { navigate } from 'src/navigator/NavigationService'
 import { Screens } from 'src/navigator/Screens'
 import { NumberToRecipient } from 'src/recipients/recipient'
 import { TransactionDataInput } from 'src/send/SendAmountLegacy'
+import { InviteTransactions } from 'src/transactions/reducer'
 import { TransferConfirmationCardProps } from 'src/transactions/TransferConfirmationCard'
 import {
   StandbyTransaction,
@@ -26,6 +27,7 @@ export enum Actions {
   UPDATE_TRANSACTIONS = 'TRANSACTIONS/UPDATE_TRANSACTIONS',
   // Remove legacy action once the multitoken support feature is fully released
   ADD_STANDBY_TRANSACTION_LEGACY = 'TRANSACTIONS/ADD_STANDBY_TRANSACTION_LEGACY',
+  UPDATE_INVITE_TRANSACTIONS = 'TRANSACTIONS/UPDATE_INVITE_TRANSACTIONS',
 }
 
 export interface AddStandbyTransactionAction {
@@ -79,6 +81,11 @@ export interface UpdateTransactionsAction {
   transactions: TokenTransaction[]
 }
 
+export interface UpdateInviteTransactionsAction {
+  type: Actions.UPDATE_INVITE_TRANSACTIONS
+  inviteTransactions: InviteTransactions
+}
+
 export type ActionTypes =
   | AddStandbyTransactionAction
   | RemoveStandbyTransactionAction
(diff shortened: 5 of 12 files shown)


packages/mobile/src/home/WalletHome.test.tsx
```diff
@@ -76,16 +89,31 @@ jest.mock('src/api/slice', () => ({
...
+          tokenTransactionsV2: {
+            transactions: [],
+          },
+        },
+      })
+    )
   })
 
-  afterAll(() => {
+  afterEach(() => {
```
kathaypacific, line 108: unrelated to this PR but i think we are setting the `useFakeTimers` as the default behaviour for jest [here](https://github.com/valora-inc/wallet/blob/deae905ea9e52f2764cc2cc90efb45c52e3cef63/packages/mobile/jest_setup.ts#L4) so i think we don't need this in the `beforeEach` and `afterEach`?

packages/mobile/src/home/WalletHome.tsx
```diff
@@ -110,10 +101,17 @@ function WalletHome() {
   }
 
   const shouldShowCashInBottomSheet = () => {
-    const hasStable = STABLE_CURRENCIES.some((currency) =>
-      balances[currency]?.isGreaterThan(STABLE_TRANSACTION_MIN_AMOUNT)
+    // If there are no core tokens then we are either still loading or loading failed.
+    if (coreTokenBalances.length < 2) {
```
kathaypacific, line 105: any reason why this condition is less than 2 and not 1?

packages/mobile/src/home/WalletHome.tsx
```diff
@@ -110,10 +101,17 @@ function WalletHome() {
...
+      return false
+    }
+    const hasStable = !!coreTokenBalances.find(
+      (token) => token.balance.gte(STABLE_TRANSACTION_MIN_AMOUNT) && token.address !== celoAddress
     )
-    const hasCelo = balances[Currency.Celo]?.isGreaterThan(CELO_TRANSACTION_MIN_AMOUNT)
+
+    const hasCelo = coreTokenBalances
+      .find((token) => token.address !== celoAddress)
```
kathaypacific, line 113: should this condition be `token.address === celoAddress` instead?

packages/mobile/src/navigator/DrawerNavigator.tsx
```diff
@@ -181,7 +178,6 @@ function CustomDrawerContent(props: DrawerContentComponentProps<DrawerContentOpt
           />
         )}
         <View style={styles.border} />
-        {showBalances && <BalancesDisplay />}
```
kathaypacific, line 184: can we delete the BalancesDisplay component since it's no longer used?

packages/mobile/src/send/Send.tsx
```diff
@@ -118,20 +117,12 @@ function Send({ route }: Props) {
...
       )
 
-      if (multiTokenUseSendFlow) {
-        navigate(Screens.SendAmount, {
-          recipient,
-          isOutgoingPaymentRequest,
-          origin: isOutgoingPaymentRequest ? SendOrigin.AppRequestFlow : SendOrigin.AppSendFlow,
-        })
-      } else {
-        navigate(Screens.SendAmountLegacy, {
```
kathaypacific, line 128: i think we can delete the `SendAmountLegacy` component now? seems like nobody is navigating to it anymore

packages/mobile/src/send/SendConfirmation.tsx
```diff
@@ -99,12 +99,13 @@ function SendConfirmation(props: Props) {
...
       amountIsInLocalCurrency,
       tokenAddress,
+      comment: commentFromParams,
     },
   } = props.route.params
 
   const [inviteModalVisible, setInviteModalVisible] = useState(false)
   const [encryptionDialogVisible, setEncryptionDialogVisible] = useState(false)
-  const [comment, setComment] = useState('')
+  const [comment, setComment] = useState(commentFromParams ?? '')
```
kathaypacific, line 108: YAY!!! nice catch

packages/mobile/src/send/utils.test.ts
```diff
@@ -318,34 +340,53 @@ describe('send/utils', () => {
...
+          .provide([
+            [matchers.call.fn(fetchExchangeRate), '1'],
+            [select(localCurrencyToUsdSelector), '1'],
+          ])
           .run()
-        expect(navigate).not.toHaveBeenCalled()
+        expect(navigate).toHaveBeenCalledWith(Screens.SendAmount, {
+          origin: SendOrigin.AppSendFlow,
+          recipient: { address: mockUriData[2].address.toLowerCase() },
+        })
```
kathaypacific, line 389: yay nice work updating these tests, nice to have more coverage for this now!!

kathaypacific (commented):
Yay nice work!! i didn't realise you were already doing this, i created the same ticket for it yesterday 🤦‍♀️ i'll link it to this PR

packages/mobile/src/home/WalletHome.test.tsx
gnardini: True, just in case I will keep the `jest.useFakeTimers()` because we haven't been consistent about it and might be setting real timers back in other tests, to make sure we are using fake ones in this PR.

packages/mobile/src/home/WalletHome.tsx
gnardini: Well, if there's only one core token either `hasStable` or `hasCelo` will return `false` and the whole function will return `false`, but I guess it's not clear from this so I'll change it :)

packages/mobile/src/home/WalletHome.tsx
gnardini: Yes! 🤦‍♂️ 
Thanks!

packages/mobile/src/navigator/DrawerNavigator.tsx
gnardini: Yes, good catch :)

packages/mobile/src/send/Send.tsx
gnardini: Yes! 🙌

ellenwoconnor:
For some reason this is showing in new issues. Any idea how to get it to the closed column?
cc @kathaypacific ?

Issue #2208, opened 2022-03-22
[Invites v2] Analytics 


kathaypacific:
### Description
We probably want to track some events to help us understand user behaviour better. Need to chat to Gowtham about what kind of events are needed once the user flow is finalised

Analytics Requirements

- [x] Ability to retrieve who was exposed to the treatment
    - [x] Push notification send/view/click data — we need this in celo-mobile-mainnet
    - [x] Home card impressions for the experiment should be emitted - `notification_impression` with id of notification card (you choose this when creating the homecard through admin dashboard)
    - [x] Banner impressions in send flow should be emitted - `invite_banner_impression`
- [x] Ability to determine who sent an invitation and how much money they sent  - `send_confirm_send`
- [x] Ability to determine who was rewarded for an invitation - `invite_reward_created`
- [x] Ability to determine who was the inviter for any rewarded invitees (or conversely, who was the invitee for any rewarded inviter) - `invite_reward_created` contains properties for this
- [x] Ability to measure completion rates for key tasks
    - [x] App opens (exists app_launched)
    - [x] new event for “account funded” defined specifically as user going from a zero to nonzero balance (nice to have - cash in type added as metadata - exchange, provider, transaction) (define "zero balance" as less than $0.01 USD, this event is triggered on total balance change) - `account_funded`
    - [x] new event for “account liquidated” defined specifically as user going from nonzero to zero balance (define "zero balance" as less than $0.01 USD) - `account_liquidated`
    - [x] User claims their supercharge rewards (exists `rewards_screen_cta_pressed`)
    - [x] User completes a dapp transaction (ideally metadata on which type of dapp/transaction) (exists `dappkit_request_accept_success` and `wc_request_accept_success`)
    - [x] User sends money via send flow (exists `send_confirm_send`)

### Acceptance Criteria

- Audit list above for any gaps
- Add instrumentation for anything that is new/missing
- Update the notion doc with details on the data to be used for each of the above (e.g., the event name, or the metadata) 

### Eng

**In the wallet:**
- [x] Home card impressions for the experiment should be emitted
- [x] Banner impressions in send flow should be emitted
- [x] new event for “account funded” defined specifically as user going from a zero to nonzero balance (nice to have - cash in type added as metadata - exchange, provider, transaction) (define "zero balance" as less than $0.01 USD, event is triggered on total balance change)
- [x] new event for “account liquidated” defined specifically as user going from nonzero to zero balance

**In the invite rewards project:**
[...]

PR #2288, opened 2022-04-01
Add impression events for home cards and invite rewards banner

kathaypacific:
### Description

In this PR:
- add impression event for home cards, to know which home cards are viewed
- add impression event for invite rewards banner

### Other changes

N/A

### Tested

Manually

### How others should test

N/A

### Related issues

- Relates to #2208 

### Backwards compatibility

Yes


packages/mobile/src/send/Send.tsx
```diff
@@ -189,6 +186,9 @@ function Send({ route }: Props) {
       <SendHeader isOutgoingPaymentRequest={isOutgoingPaymentRequest} />
       <DisconnectBanner />
       <SendSearchInput input={searchQuery} onChangeText={throttledSearch} />
+      {inviteRewardsEnabled && numberVerified && hasGivenContactPermission && (
+        <InviteRewardsBanner />
+      )}
```
jeanregisser, line 191: Curious, why moving moving this here?

jeanregisser (approved):
LGTM 👍

packages/mobile/src/send/Send.tsx
kathaypacific: oh! sorry, i should have left a comment. i only want to dispatch the impression event once (this is done inside the InviteRewardsBanner component) but the renderListHeader is called multiple times on mount so i was getting too many events. the UI still looks the same if the banner lives here, so i opted to do this rather than to move the invites analytics event into this Send component since it does so much already. hope you agree :D

ValoraQA:
Hey @MuckT since 'N/A' is mentioned  in how others should test section we are skipping this PR
CC: @kathaypacific

PR #2307, opened 2022-04-12
fix(dapps): WebView error when connecting dapps with slow Android devices

### LINKED ISSUE #2306: Dapps browser shows error ERR_UNKOWN_URL_SCHEME when connecting Valora
### Current behavior

This is happening on a Nexus 5, with Android 6.0.1
I personnally saw 2 other users at CeloConnect with this issue.

https://user-images.githubusercontent.com/57791/162963214-d4d52533-3f0c-4412-afd4-6d36fa6b4cf8.mp4

### Desired behavior

The dapp should connect to Valora without showing the error in the in-app browser.

### Steps to Reproduce

1. Open Ubeswap from the Dapps page
2. Tap to Connect Valora (sometimes Ubeswap triggers this directly on load).
3. Notice the error ERR_UNKNOWN_URL_SCHEME displayed in the webview

Note that it only affects specific devices / Android versions. I'm unable to reproduce this on a Pixel 4 running Android 11.

### Versions

Valora v1.29.2 on Android 6.0.1 (Nexus 5)

jeanregisser:
### Description

On slower/older Android devices, connecting to Dapps from the in-app WebView would display an error and fail.

This is because the Android WebView mechanism to decide whether to load new urls has a default timeout of 250ms for the JS side to return a value. See https://github.com/react-native-webview/react-native-webview/blob/5e73b2089fc80c2be7aa6eff291b18c81ad4030d/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java#L921-L967

This PR increases the timeout to 60secs which should be way more than enough for slower/older Android devices to respond.

This patch should probably be contributed to [react-native-webview](https://github.com/react-native-webview/react-native-webview) once we confirm it's working as expected on more devices.

I've also added a log to capture WebView errors so we are alerted of similar issues.

### Tested

Connecting to Dapps works within the in-app browser on the Nexus 5 device where I reproduced the issue.

### How others should test

1. Use a slow Android device (Nexus 5, is guaranteed to be able to reproduce this)
2. Open Ubeswap in the dapps browser
3. Connect to Valora
4. Connection should work and the dapp interface should update accordingly.

### Related issues

- Fixes #2306

### Backwards compatibility

Yes

### DIFF
--- a/packages/mobile/src/account/__snapshots__/Licenses.test.tsx.snap
+++ b/packages/mobile/src/account/__snapshots__/Licenses.test.tsx.snap
@@ -29,6 +29,7 @@ exports[`Licenses renders correctly 1`] = `
       javaScriptEnabled={true}
       messagingEnabled={false}
       messagingModuleName="WebViewMessageHandler1"
+      onError={[Function]}
       onHttpError={[Function]}
       onLoadingError={[Function]}
       onLoadingFinish={[Function]}
--- a/packages/mobile/src/components/WebView.tsx
+++ b/packages/mobile/src/components/WebView.tsx
@@ -1,6 +1,8 @@
 import React from 'react'
 import { Platform, StyleSheet } from 'react-native'
 import { WebView as RNWebView, WebViewProps } from 'react-native-webview'
+import { WebViewErrorEvent } from 'react-native-webview/lib/WebViewTypes'
+import Logger from 'src/utils/Logger'
 
 export type WebViewRef = RNWebView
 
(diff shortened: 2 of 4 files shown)


kathaypacific (approved):
🚀 yay thank you!! and thanks for adding the error logger, that'll be useful!

ValoraQA:
Hey @MuckT we have verifed the above task on latest android internal release build V:1.30.1 and updated the status in test rails

**Observation :** No error was seen while connecting dapps to valora wallet and connection was successful

**Test rail link :** https://valoraapp.testrail.io/index.php?/tests/view/27300
**Device :** Saumsung Galaxy A3(5.0.2)

Cc: @jeanregisser 
Thanks.!!

PR #2316, opened 2022-04-13
Add events for account funded and liquidated

kathaypacific:
### Description

This PR implements a "watcher" over the total balance value of the app, and dispatches events when the account is funded and liquidated. 

### Other changes

N/A

### Tested

Unit tests

### How others should test

N/A

### Related issues

Relates #2208 

### Backwards compatibility

Yes


packages/mobile/src/tokens/saga.test.ts
```diff
@@ -107,6 +109,53 @@ describe(fetchTokenBalancesSaga, () => {
       error: 'Error message',
     })
   })
+
+  describe('account event', () => {
+    let returnValueFirstTimeOnly: (value: BigNumber) => EffectProvider<any>
+
+    beforeEach(() => {
+      returnValueFirstTimeOnly = (value: BigNumber) => {
```
kathaypacific, line 117: this looks strange, i know. what i want to do is mock the outcome of the totalTokenBalanceSelector differently the first and second time it is called. this is surprisingly not out of the box, and the closest thing i could get to is [this](https://github.com/jfairbank/redux-saga-test-plan/blob/master/docs/integration-testing/mocking/static-providers.md#static-providers-with-dynamic-values).

packages/mobile/src/tokens/saga.ts
```diff
@@ -313,6 +313,22 @@ export async function fetchTokenBalancesForAddress(
   return response.data.userBalances.balances
 }
 
+export function* updateTokenBalances(tokens: StoredTokenBalances) {
+  const tokenBalanceBefore = yield select(totalTokenBalanceSelector)
+  const isAccountFundedBefore = tokenBalanceBefore.gt(DOLLAR_MIN_AMOUNT_ACCOUNT_FUNDED)
+
+  yield put(setTokenBalances(tokens))
```
kathaypacific, line 320: i *think* that the `yield put` will wait for the redux state to change before moving to the next part.

there's some pretty complex stuff going on in the `totalTokenBalanceSelector` which is nicely memoised right now so this was the most straightforward way i could think of to reuse that logic.

packages/mobile/src/tokens/saga.ts
```diff
@@ -337,7 +353,7 @@ export function* fetchTokenBalancesSaga() {
           .toString()
       }
     }
-    yield put(setTokenBalances(tokens))
+    yield call(updateTokenBalances, tokens)
```
jeanregisser, line 356: I think it would be more robust to really watch the changes to `totalTokenBalanceSelector` after actions are dispatched.

Something like this:
```
let prevValue
while (true) {
  const newValue: ReturnType<typeof totalTokenBalanceSelector> = yield select(totalTokenBalanceSelector)
  if (prevValue !== newValue) {
    // Logic when it changes
  }

  yield take()
}
```

It makes it more resilient to future changes. For instance what if there's another place calling `setTokenBalances` or if the logic in `totalTokenBalanceSelector` changes and something else goes into it?

Right now, `totalTokenBalanceSelector` also uses the local currency exchange rate and if it's updated after this call happens, it won't reflect the right value.

packages/mobile/src/tokens/saga.ts
```diff
@@ -369,6 +369,34 @@ export function* watchFetchBalance() {
...
+export function* watchAccountFundedOrLiquidated() {
+  let prevTokenBalance
+  while (true) {
+    const tokenBalance: ReturnType<typeof totalTokenBalanceSelector> = yield select(
+      totalTokenBalanceSelector
+    )
+    if (tokenBalance !== prevTokenBalance) {
+      // prevTokenBalance is undefined for the base case and null if token list
+      // is not yet loaded
+      if (prevTokenBalance) {
```
kathaypacific, line 381: i added this to prevent the account_funded event from being fired when a user is restoring their account (the balance goes from null to a non-zero amount in that case)

packages/mobile/src/tokens/selectors.ts
```diff
@@ -102,10 +102,15 @@ export const defaultTokenToSendSelector = createSelector(
...
+  [
+    tokensListSelector,
+    tokensWithUsdValueSelector,
+    localCurrencyExchangeRatesSelector,
+    tokenFetchErrorSelector,
+  ],
+  (tokensList, tokensWithUsdValue, exchangeRate, tokenFetchError) => {
     const usdRate = exchangeRate[Currency.Dollar]
-    if (!usdRate) {
+    if (!usdRate || Object.keys(tokensList).length === 0) {
```
kathaypacific, line 113: i think it makes more sense for this selector to return null rather than 0 if the token list has not yet loaded, but also without this we cannot differentiate between an account funded event and a user who is restoring their account (going from 0 to a non-zero balance)

packages/mobile/src/tokens/saga.ts
kathaypacific: thank you for the feedback Jean! this makes a lot of sense, i see we also do this for the user traits (i didn't find this before) - i've updated this PR to use your suggested approach :)

packages/mobile/src/tokens/selectors.ts
gnardini, line 113: Isn't `tokensList` an array? I think we should check `tokensList.length` instead of `Object.keys().length`

packages/mobile/src/tokens/saga.test.ts
```diff
@@ -140,3 +142,52 @@ describe(tokenAmountInSmallestUnit, () => {
...
+        case 2:
+        case 3:
+          return restValue
+        default:
+          return firstValue
+      }
+    }
+  }
+
+  it('dispatches the account funded event', async () => {
```
jeanregisser, line 168: Should we add a test that going from `null` to something non zero doesn't track `account_funded` because it's an account restore?

jeanregisser (approved):
Looking great! 🚀

packages/mobile/src/tokens/selectors.ts
kathaypacific: yikes, i don't know why i thought i was working with the `tokens` directly. i updated this, thanks for catching!

packages/mobile/src/tokens/saga.test.ts
kathaypacific: yeah good idea, i added a test case for this :)

ValoraQA:

Cc: @kathaypacific 
Thanks.!

PR #2317, opened 2022-04-13
ci: Emerge Tools status check for PRs

jeanregisser:
### Description

This enables Emerge Tools' status check for each PR, with the impact on the app size.

This allows us to quickly see if any PR would significantly impact the final app size. Instead of having to manually check the Emerge Tools dashboard.

The good thing is this doesn't really impact our CI times. It just uploads the detox build used for the e2e test.
The downside is the app size is not fully accurate because it's not doing an archive build and contains some Detox related code.
But the diff itself will be accurate. See https://docs.emergetools.com/docs/overview#test-builds

Note that for now it only uploads the iOS build as it only supports a single status check.
The Emerge Tools folks told me multiple checks support may come in the future though.

### Tested

New status check displayed on each PR:

Screenshot 2022-04-13 at 19 03 04

Note: I manually uploaded a build from the latest main commit so I could see the diff report.
It always compares the PR against the `baseSha` build (usually latest commit from main, if the PR targets main).

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

Yes


MuckT (commented):
@jeanregisser would it make sense to create a new task that would run on a Cron job so that outside contributors wouldn't be blocked from running the e2e tests? I am not sure if they are currently blocked with the way the e2e-ci is working with Ruby, but it appears they might be.

jeanregisser:
@MuckT PR is now ready to review! The Emergetools upload step is skipped if secrets are not available to avoid blocking external contributions.

MuckT:
@jeanregisser I see `Size Analysis | Emerge Failing after 10m — Could not generate diff` will this start being generated after this PR is merged?

MuckT (approved):
LGTM! Just one note about the base not being uploaded, but imagine this will change once this gets merged into main.

jeanregisser:
> @jeanregisser I see Size Analysis | Emerge Failing after 10m — Could not generate diff will this start being generated after this PR is merged?

Yep, new commits landed on main since I manually uploaded a base build. Will be resolved once this PR is merged.

PR #2320, opened 2022-04-13
feat: fiatconnect providers getter

### LINKED ISSUE #1979: [FC UI] Vendor configs
As a developer trying to run queries across FC vendors (like get quotes from all of them), I need access to a list of vendors and static fields for each one (like name, icon, base URL, ...).

Eng:
- this issue involves writing a schema for provider static info, as well as deciding/implementing where the static info should live
- options for where the static info could live: in code in the wallet repo (would take longer to change though), firebase, an IHL endpoint, a cloud function
- if an endpoint is called, we should save caching as an optimization for later (just call the endpoint every time for now)

AC:
- a function exists in the wallet repo that can be used to get a list of all FC vendors and static info on each one
- info on each vendor is enough for anticipated use cases (display list of providers, query for quotes from each provider, post KYC info to the provider, ..., initiate/monitor a transfer with the provider)

cajubelt:
### Description

Added a helper to get fiatconnect providers from a cloud function

### Other changes

na

### Tested

added unit tests

### How others should test

na (not user facing yet)

### Related issues

fixes #1979 

### Backwards compatibility

🟢

### DIFF
--- a/packages/mobile/src/fiatconnect/index.ts
+++ b/packages/mobile/src/fiatconnect/index.ts
@@ -1,0 +1,29 @@
+import networkConfig from 'src/geth/networkConfig'
+import Logger from '../utils/Logger'
+
+const TAG = 'FIATCONNECT'
+
+export interface FiatConnectClientConfig {
+  // todo eventually: get from SDK (once it is published to NPM)
+  baseUrl: string
+  providerName: string
+  iconUrl: string
+}
+
+/**
+ * Get a list of FiatConnect providers.
+ *
+ * Queries a cloud function for static information about FiatConnect providers.
+ */
+export async function getFiatConnectProviders(): Promise<FiatConnectClientConfig[]> {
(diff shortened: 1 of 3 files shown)


packages/mobile/src/fiatconnect/index.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Get a list of FiatConnect providers.
+ *
+ * Queries a cloud function for static information about FiatConnect providers.
+ */
+export async function getFiatConnectProviders(): Promise<FiatConnectClientConfig[]> {
+  const response = await fetch(networkConfig.getFiatConnectProvidersUrl)
+  if (!response.ok) {
+    Logger.error(
+      TAG,
+      `Failure response fetching FiatConnect providers: ${response} , returning empty list`
```
cajubelt, line 23: This seemed preferable over throwing an error, since we can still show exchanges and non-FC providers on the provider selection screen

PR #2322, opened 2022-04-14
Fix webview crash on iOS due to missing page title during loading

kathaypacific:
### Description

Title says it all, during page load the title is never empty for Android but is empty for iOS. Needed to update the display title condition so that iOS correctly displays "Loading"

### Other changes

N/A

### Tested

Manually on iOS and Android

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

Yes


jeanregisser (approved):
👍

PR #2325, opened 2022-04-14
Fail E2E Test Prior to Run When E2E Wallet Balance is Low

MuckT:
### Description

Fails the E2E tests prior to the run when the E2E wallet balance is low to avoid wasting developer time debugging when the source is likely a low wallet balance.

### Other changes

Moved functions out of the `fund-e2e-accounts.ts` script into a utils module and the constants into a constants module to be shared between the `fund-e2e-accounts.ts` and `check-e2e-wallet-balance.ts`. This will allow extension of e2e scripts in the future.

`fund-e2e-accounts.ts` now balances the test faucet and is run via cron job with GitHub Action every eight hours.

### Tested

Tested locally and in CI.

### How others should test

This does not need to be tested by QA in the next release cycle. Reviewers should attempt to run `node ./fund-e2e-accounts.ts` and `node ./check-e2e-wallet-balance.ts` from the `packages/mobile/e2e/scripts` folder.

### Related issues

N/A

### Backwards compatibility

Yes


.github/workflows/e2e-ci.yml
```diff
@@ -45,6 +45,9 @@ jobs:
       - name: Make jest available
         working-directory: ${{env.working-directory}}
         run: cp ../../node_modules/.bin/jest node_modules/.bin/
+      - name: Check E2E wallet balance
+        working-directory: ${{env.working-directory}}/e2e/scripts
+        run: node --unhandled-rejections=strict ./check-e2e-wallet-balance.ts
```
jeanregisser, line 50: It should use `ts-node` otherwise it could fail on TS syntax, some tweaks may be needed though, like using `import` instead of `require` to get types.
```suggestion
        run: yarn ts-node --unhandled-rejections=strict ./check-e2e-wallet-balance.ts
```

packages/mobile/e2e/scripts/utils.ts
```diff
@@ -0,0 +1,41 @@
+const Web3 = require('web3')
+const ContractKit = require('@celo/contractkit')
+const dotenv = require('dotenv')
+const web3 = new Web3('https://alfajores-forno.celo-testnet.org')
+const kit = ContractKit.newKitFromWeb3(web3)
+var consts = require('./consts.ts')
```
jeanregisser, line 6: AFAIK, using `require` we don't get compile time TS type checks.  Can we switch to `import`?

.github/workflows/e2e-ci.yml
```diff
@@ -128,6 +131,9 @@ jobs:
...
```
jeanregisser, line 136: ditto about `ts-node`

jeanregisser (commented):
Great! 👍 
Nothing to do here now, but @kathaypacific proposed to do the balance checks directly in the e2e test (maybe in the beforeAll for each suite?).
The advantage is it would directly fail even when running locally.
What do you think?

MuckT:
@jeanregisser working to refactor to use exported functions in a TS module.

.github/workflows/e2e-ci.yml
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2325/commits/876bcbf710ddca64aa4a435021cd45ffbab09303 and https://github.com/valora-inc/wallet/pull/2325/commits/8ef142ea9af39f48f34bef82c345e71fb12d9de6!

packages/mobile/e2e/scripts/consts.ts
```diff
@@ -0,0 +1,4 @@
+module.exports = Object.freeze({
+  E2E_TEST_WALLET: '0x6131a6d616a4be3737b38988847270a64bc10caa',
+  E2E_TEST_FAUCET: '0xe5F5363e31351C38ac82DBAdeaD91Fd5a7B08846',
+})
```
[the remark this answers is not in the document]
MuckT: Fixed in 876bcbf!

.github/workflows/e2e-ci.yml
MuckT: Fixed in https://github.com/valora-inc/wallet/commit/8ef142ea9af39f48f34bef82c345e71fb12d9de6!

packages/mobile/e2e/scripts/utils.ts
MuckT: I ended up exporting these constants and using imports in `check-e2e-wallet-balance.ts` and `fund-e2e-accounts.ts` as seen [here](https://github.com/valora-inc/wallet/blob/tomm/low-balance-fast-fail/packages/mobile/e2e/scripts/utils.ts#L2-L6).

MuckT:
@jeanregisser to run the checks directly in the e2e tests it looks like we need to convert the e2e tests over to typescript, possible, but outside of the scope of this pull request.

.github/workflows/e2e-ci.yml
```diff
@@ -43,6 +43,9 @@ jobs:
       - name: Make jest available
         working-directory: ${{env.working-directory}}
         run: cp ../../node_modules/.bin/jest node_modules/.bin/
+      - name: Check E2E wallet balance
+        working-directory: ${{env.working-directory}}
+        run: yarn ts-node ./e2e/scripts/check-e2e-wallet-balance.ts
```
jeanregisser, line 48: We need to fail the workflow in case of an unhandled rejected promise. One way to do it with [ts-node](https://typestrong.org/ts-node/docs/configuration/#node-flags):
```suggestion
        run: NODE_OPTIONS='--unhandled-rejections=strict' yarn ts-node ./e2e/scripts/check-e2e-wallet-balance.ts
```

.github/workflows/e2e-ci.yml
```diff
@@ -124,6 +127,9 @@ jobs:
...
```
jeanregisser, line 132: ditto

.github/workflows/e2e-faucet-balance.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Install package dependencies
+        run: yarn
+      - name: Create E2E Fund .env File
+        env:
+          TEST_FAUCET_SECRET: ${{ secrets.TEST_FAUCET_SECRET }}
+        working-directory: ${{env.working-directory}}/e2e
+        run: echo TEST_FAUCET_SECRET=$TEST_FAUCET_SECRET >> .env
+      - name: Run Balance and Fund
+        working-directory: ${{env.working-directory}}
+        run: yarn ts-node ./e2e/scripts/fund-e2e-accounts.ts
```
jeanregisser, line 35: ditto

packages/mobile/e2e/scripts/check-e2e-wallet-balance.ts
```diff
@@ -0,0 +1,7 @@
+import { E2E_TEST_WALLET } from './consts'
+import { balanceError, getBalance } from './utils'
+;(async () => {
+  console.log(E2E_TEST_WALLET)
+  console.table(await getBalance())
+  await balanceError()
```
jeanregisser, line 6: On the first look I was surprised, does it store the balance in a global variable? What does `balanceError` mean?

It's better to explicitly pass the address being checked.

I propose to change it to:
```suggestion
  await checkBalance(E2E_TEST_WALLET)
```

`checkBalance` would display and check the balance.

packages/mobile/e2e/scripts/utils.ts
```diff
@@ -0,0 +1,32 @@
+import { E2E_TEST_WALLET } from './consts'
+export const Web3 = require('web3')
+export const ContractKit = require('@celo/contractkit')
+export const dotenv = require('dotenv')
```
jeanregisser, line 4: Any reason not to let other modules import these directly?

Also with a `require` we won't get the types.

packages/mobile/e2e/scripts/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+export const web3 = new Web3('https://alfajores-forno.celo-testnet.org')
+export const kit = ContractKit.newKitFromWeb3(web3)
+
+export const balanceError = async (address = E2E_TEST_WALLET, minBalance = 10) => {
```
jeanregisser, line 8: Following what I suggested from the check script:
```suggestion
export async function checkBalance(address, minBalance = 10) {
```

Also nit: Basic function better represents the intention to be called. Plus it's a few chars shorter to write.

packages/mobile/e2e/scripts/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+  let balanceObject = await getBalance(address)
```
jeanregisser, line 9: We'd log the balance here too.

packages/mobile/e2e/scripts/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+  for (const balance in balanceObject) {
+    if (balanceObject[balance as keyof typeof balanceObject] < minBalance) {
+      throw new Error(
+        `${balance} balance of ${address} is below ${minBalance}. Please refill from the faucet https://celo.org/developers/faucet 🙏`
```
jeanregisser, line 13: Should we mention the balance script instead?

packages/mobile/e2e/scripts/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+      )
+    }
+  }
+}
+export const getBalance = async (address = E2E_TEST_WALLET) => {
```
jeanregisser, line 18:
```suggestion
export async function getBalance(address) {
```

packages/mobile/e2e/scripts/fund-e2e-accounts.ts
```diff
@@ -1,87 +1,173 @@
...
+  const ceurExchange = await kit.contracts.getExchange('cEUR')
+
+  // Balance Faucet
+  let balances = await getBalance(E2E_TEST_FAUCET)
+  let sum = 0
+  const numOfTokens = 3 // Object.keys(balances).length
+  for (let balance in balances) {
+    console.log(balances[balance])
+    sum += balances[balance]
+  }
```
jeanregisser, line 32: Looks like we should sum the specific tokens we are interested in instead, right?

packages/mobile/e2e/scripts/fund-e2e-accounts.ts
```diff
@@ -1,87 +1,173 @@
...
+              .send({ from: E2E_TEST_FAUCET })
+            await ceurBuyTx.waitReceipt()
+          } catch (err) {
+            console.log('Failed to buy cEUR', err)
+          } finally {
+            break
+          }
+      }
+    }
+  }
```
jeanregisser, line 146: nit: it seems we could reduce some of the duplicated logic. No need to do it though, or maybe if we ever need to add another token.

jeanregisser (commented):
Awesome work!! 👍

.github/workflows/e2e-ci.yml
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2325/commits/336783cd5921d1d51e9f43703d5fd3acd2a89456!

.github/workflows/e2e-ci.yml
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2325/commits/336783cd5921d1d51e9f43703d5fd3acd2a89456!

.github/workflows/e2e-faucet-balance.yml
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2325/commits/6d29a22cc5df60c80edc63712eb6881932b42232!

packages/mobile/e2e/scripts/check-e2e-wallet-balance.ts
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2325/commits/2a72d8204532a885b1fb40b95a45f1c7ce9f3fef! I kept checkBalance and getBalance separate as we don't need to call checkBalance on the faucet.

packages/mobile/e2e/scripts/utils.ts
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2325/commits/2a72d8204532a885b1fb40b95a45f1c7ce9f3fef!

packages/mobile/e2e/scripts/utils.ts
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2325/commits/2a72d8204532a885b1fb40b95a45f1c7ce9f3fef!

packages/mobile/e2e/scripts/utils.ts
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2325/commits/2a72d8204532a885b1fb40b95a45f1c7ce9f3fef!

packages/mobile/e2e/scripts/fund-e2e-accounts.ts
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2325/commits/b8602f07b02fc7f938f88b39f54eb65550193b59!

packages/mobile/e2e/scripts/fund-e2e-accounts.ts
MuckT: I could see deduplicating some of this logic as cleaner; however, I thought the ability to easily extend it to cover cREAL might be more important.

packages/mobile/e2e/scripts/utils.ts
MuckT: Fixed in [c3d9135](https://github.com/valora-inc/wallet/pull/2325/commits/c3d9135f93a845d1065db1d174e274cee86d96a5)!

MuckT:
@jeanregisser thank you for the thoughtful and throughout review! I've addressed most of the major concerns, except combining `checkBalance` and `getBalance` and gave some details in https://github.com/valora-inc/wallet/pull/2325#discussion_r867063203,  please re-review when able. 🙏

jeanregisser (approved):
🚀

PR #2347, opened 2022-04-20
E2E Timeout Fixes

MuckT:
### Description

A few e2e tests such as `Ramps.spec.js` are failing due to slow load times this pr adds implicit waits for those elements to render. It also wraps the exchange scroll in a try catch to allow tests to pass even if fetching transactions is slow from blockscout.

### Other changes

N/A

### Tested

Tested in CI.

### How others should test

This does not need to be tested by QA. Reviewers should view the e2e test runs on this branch.

### Related issues

N/A

### Backwards compatibility

Yes


packages/mobile/e2e/src/usecases/ExchangeCelo.js
```diff
@@ -53,10 +52,15 @@ export default ExchangeCelo = () => {
...
-    await element(by.id('ExchangeScrollView')).scroll(200, 'down')
     // Wait 10 seconds checking that error banner is not visible each second
     await waitForExpectNotVisible('errorBanner')
+    // Scroll to top of feed - scroll not possible if not all transactions loaded
+    try {
+      await waitFor(element(by.id('ExchangeScrollView')))
+        .toBeVisible()
+        .withTimeout(10 * 1000)
+      await element(by.id('ExchangeScrollView')).scroll(200, 'down')
+    } catch {}
```
kathaypacific, line 63: do we need the try/catch here? it seems that if anything fails inside the try, we will do nothing in the catch and then proceed to the next step (which should fail if the previous steps didn't complete, i assume?). maybe i am missing something but that seems like the same flow as if the try/catch was not there?

packages/mobile/e2e/src/usecases/ExchangeCelo.js
```diff
@@ -91,10 +94,15 @@ export default ExchangeCelo = () => {
     await enterPinUiIfNecessary()
     // Get transaction time
     let transactionTime = new Date()
-    // Scroll to top of feed
-    await element(by.id('ExchangeScrollView')).scroll(200, 'down')
     // Wait 10 seconds checking that error banner is not visible each second
     await waitForExpectNotVisible('errorBanner')
+    // Scroll to top of feed - scroll not possible if not all transactions loaded
+    try {
+      await waitFor(element(by.id('ExchangeScrollView')))
```
kathaypacific, line 101: i just noticed that this same code block is used a few times, could we abstract to a helper function?

kathaypacific (commented):
Looking great!! i just had a couple small comments 🥳

packages/mobile/e2e/src/usecases/ExchangeCelo.js
MuckT: We want to try and wait for the element and scroll; however, it fails we don't want the test to fail. This is needed because if the entire transaction feed loads we want to scroll a bit, if the transaction feed doesn't load the transaction will be at the top and we can assert on that.

packages/mobile/e2e/src/usecases/ExchangeCelo.js
MuckT: Should be fixed in https://github.com/valora-inc/wallet/pull/2347/commits/ec44ddb31add2ccb3394211fad65e43d8cd27fd9!

kathaypacific (approved):
LGTM 🚀

packages/mobile/e2e/src/usecases/ExchangeCelo.js
```diff
@@ -14,7 +14,36 @@ const CELO_TO_SELL = 0.045
...
-// TODO Fetch Fees
+
+const scrollToTopOfFeed = async () => {
+  // Scroll to top of feed - scroll not possible if not all transactions loaded
+  try {
+    await waitFor(element(by.id('ExchangeScrollView')))
+      .toBeVisible()
+      .withTimeout(10 * 1000)
+    await element(by.id('ExchangeScrollView')).scroll(200, 'down')
+  } catch {}
```
silasbw, line 25: the blind `catch` is a little worrisome.  Instead, can you check explicitly for the error you're expecting?

MuckT: This is to only attempt to scroll if we can. `waitForExpectNotVisible` and `assertTransactionAppeared` will throw an error if an error banner is seen or the transaction does not appear in the feed.

ValoraQA:
Hey @MuckT  since "This does not need to be tested by QA. Reviewers should view the e2e test runs on this branch." is mentioned in how other should test section , we are skipping this test case

Thanks..!

PR #2351, opened 2022-04-22
Call setRegistrationProperties endpoint for registrations instead of using Firebase directly

kathaypacific:
### Description

We need to add the country code to the registration node in firebase, to enable region based invite rewards. To decrease our reliance on firebase, we have a new setRegistrationProperties endpoint created that we should call instead.

This PR:
- adds country to the firebase rtdb schema
- replaces the direct calls to Firebase with setRegistrationProperties endpoint for updating registration details
- adds keychain item (without access control, so that no pincode/biometrics are needed to access) for signed message, which is used for authentication for the setRegistrationProperties endpoint.

Note that no changes were made to the RTDB for now, the only extra authentication is on the setRegistrationProperties endpoint.

This signed message is generated when an account is created/imported for new users. For existing users, it will be generated after the first time they enter their pincode for something (to reduce friction) - this means that we will not be updating existing users's registrations node until they have done something to trigger the pincode.

### Other changes

N/A

### Tested

Unit tests and manually

### How others should test

General regression. There should be no change in the user journey.

### Related issues

- Relates to https://github.com/valora-inc/invite-rewards/issues/40

### Backwards compatibility

Yes


packages/mobile/src/firebase/db-rules.json
```diff
@@ -31,6 +31,9 @@
         "appVersion": {
           ".write": "auth.uid != null && $address === root.child('users').child(auth.uid).child('address').val()"
         },
+        "country": {
+          ".write": "auth.uid != null && $address === root.child('users').child(auth.uid).child('address').val()"
```
gnardini, line 35: Hmm, shouldn't this be `false` if the cloud function is the one updating it?

packages/mobile/src/account/saga.ts
```diff
@@ -156,6 +166,58 @@ export function* watchKycStatus() {
...
+  }
+}
+
+export function* handleUpdateAccountRegistration(properties: RegistrationProperties) {
+  const address = yield select(currentAccountSelector)
+  let signature = yield select(signedMessageSelector)
+
+  try {
+    if (!signature) {
+      signature = yield call(generateSignedMessage)
```
gnardini, line 200: Just wanted to suggest an alternative, since the user might just refuse to put in their PIN and we'd never get the signed message if they do so:

We can hook into the `getConnectedUnlockedAccount` saga and _then_ create and store the signed message. That way we can make sure that the user has a signed message if they invite someone since that requires unlocking the account. 

We could also create the signed message when creating or importing the account for new users.

gnardini:
Also, we're updating the registration node this way, but I think we also need to update the mapping in the `users` node in this way 😅

packages/mobile/src/firebase/db-rules.json
kathaypacific: 🥲 i initially thought `false` would prevent everyone from updating it including the cloud function...but this clearly isn't the case, i see `tokens` for example also has `false` so i've updated this now!

gnardini: Someone with an admin key (like our cloud functions) can always read/write, the permissions are for end users :)

packages/mobile/src/web3/saga.ts
```diff
@@ -350,6 +351,19 @@ export function* getConnectedUnlockedAccount() {
   const account: string = yield call(getConnectedAccount)
   const result: UnlockResult = yield call(unlockAccount, account)
   if (result === UnlockResult.SUCCESS) {
+    const signedMessage = yield select(signedMessageSelector)
```
gnardini, line 354: Is this called when an account is created or imported? Otherwise I think that's a good place to store the signed message as well :)

kathaypacific: oops 🙈 it's there now

kathaypacific: the annoying thing is, both the create and restore flows call the `initializeAccount` saga BUT the firebase saga where the fcmToken needs to be updated is kicked off by the root saga so there is a race condition to get the account created/get the signed message stored beforehand. for the create flow, this seems to work okay. but for the restore flow, `initializeAccount` is called too late. so i've also put the generateSignedMessage also in the import saga, in case this looks strange

jeanregisser (commented):
Great!

packages/mobile/src/import/saga.ts
```diff
@@ -154,6 +155,7 @@ export function* importBackupPhraseSaga({ phrase, useEmptyWallet }: ImportBackup
     yield call(storeMnemonic, mnemonic, account)
     // Set backup complete so user isn't prompted to do backup flow
     yield put(setBackupCompleted())
+    yield call(generateSignedMessage)
```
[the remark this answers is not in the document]
kathaypacific: i've updated this part since you commented, it was called more frequently before but now only in 2 places - when the account is initialised (create/import account flow) and the unlock account flow. ideally we'd only need to generate this message in the initialise account flow, the second case is for existing users who have just updated their app. i couldn't think of a central place to put this where both flows are covered, i think even from `assignAccountFromPrivateKey` it wouldn't cover the existing users, right?

packages/mobile/src/account/reducer.ts
```diff
@@ -109,6 +110,7 @@ export const initialState: State = {
   hasLinkedBankAccount: false,
   finclusiveKycStatus: FinclusiveKycStatus.NotSubmitted,
   finclusiveRegionSupported: false,
+  signedMessage: null,
```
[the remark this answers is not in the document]
kathaypacific: oh yes fair point. i've moved this to the keychain!

kathaypacific:
hey @gnardini @jeanregisser this is ready for another pass. a lot of stuff has been updated since your reviews, thanks for all the feedback!
- the signed message is now stored on the keychain instead of in redux
- there are 2 places where we generate the signed message - when an account is created/imported (for new users), and when the account is unlocked (for existing users)
- the registration endpoint is called when a new signed message is generated
- i'm using `signPersonalMessage` instead of `signTypedMessage` for now, since this is the way the registration endpoint is expecting the message. i've sent a message to Victor to ask about the security implications of the 2 functions, i'm happy to update this later if we decide to

kathaypacific:
Update: Victor recommended to use `signTypedMessage`, for a more structured signing interface but also to address `the main security concern is “domain collisions” where a message signed for one purpose/context overlaps with a message signed for a different purpose/context` - i'm convinced, but this will also need a change from the registrations api code that verifies the auth.

i'll make a change there and submit a separate PR to avoid blocking this functionality

packages/mobile/src/geth/networkConfig.ts
```diff
@@ -44,6 +44,7 @@ interface NetworkConfig {
   walletConnectEndpoint: string
   personaEnvironment: PersonaEnvironment
   inHouseLiquidityURL: string
+  setRegistrationProperties: string
```
jeanregisser, line 47: nit:
```suggestion
  setRegistrationPropertiesUrl: string
```

packages/mobile/src/account/saga.ts
```diff
@@ -173,6 +250,11 @@ export function* watchFetchFinclusiveKYC() {
   yield takeLeading(Actions.FETCH_FINCLUSIVE_KYC, fetchFinclusiveKyc)
 }
 
+export function* watchSignedMessage() {
+  yield take(Actions.SAVE_SIGNED_MESSAGE)
+  yield call(handleUpdateAccountRegistration)
+}
```
jeanregisser, line 256: I wonder if we should avoid manually calling `handleUpdateAccountRegistration` from different places and instead listening to action/state changes (`while (true) { ... }`)to decide when the logic should be called.

Now that I say this, it may not be feasible because of the way `fcmToken` is retrieved though. So nevermind 😄

jeanregisser (approved):
Great work! 
Approved with some small questions.

packages/mobile/src/account/updateAccountRegistration.ts
```diff
@@ -0,0 +1,35 @@
+import networkConfig from 'src/geth/networkConfig'
+
+export interface RegistrationProperties {
+  language?: string | null
+  country?: string
```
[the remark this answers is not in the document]
kathaypacific: thanks for the reminder! as we discussed i will double check which solution we want to go for, and keep this implementation for now since it won't be a breaking change if we want to capture the country server side.

packages/mobile/src/account/saga.ts
```diff
@@ -156,6 +173,66 @@ export function* watchKycStatus() {
...
+  const address = yield select(walletAddressSelector)
+  const appVersion = DeviceInfo.getVersion()
+  const language = yield select(currentLanguageSelector)
+  const country = yield select(userLocationDataSelector)
+
+  let fcmToken
+  try {
+    fcmToken = yield call([firebase.app().messaging(), 'getToken'])
+  } catch (error) {
+    Logger.error(`${TAG}@handleUpdateAccountRegistration`, 'Could not get fcm token', error)
```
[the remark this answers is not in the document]
kathaypacific: oh yeah this is a great point. i updated the implementation to only update defined values, the endpoint does this check already but i think it's more correct for the app to do this rather than rely on an endpoint implementation that might change.

packages/mobile/src/account/saga.ts
```diff
@@ -156,6 +173,66 @@ export function* watchKycStatus() {
...
+      ensureLeading0x(Buffer.from('valora auth message').toString('hex'))
+    )
+    yield call(storeSignedMessage, signedMessage)
+    yield put(saveSignedMessage())
+  } catch (error) {
+    throw error
+  }
+}
+
+export function* handleUpdateAccountRegistration(extraProperties: RegistrationProperties = {}) {
```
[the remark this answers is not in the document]
kathaypacific: oopsies, this is no longer needed. i've removed!

PR #2352, opened 2022-04-22
(fix): Avoid supercharge cache

### LINKED ISSUE #2336: Supercharge: Remove RTK Query in favor of normal querying the Cloud Function
For Supercharge, we are currently querying using a Redux library called RTK Query. See https://github.com/valora-inc/wallet/blob/main/packages/mobile/src/api/slice.ts

This has caused a bug where the library is caching results despite having a flag that tells it not to. This caching has caused many users to not be able to claim their Supercharge rewards.

A bandaid fix has been pushed that simply forces the cache to not be persisted between app opens, but the caching might still happen in memory. In this issue we should not use RTK Query anymore and instead query the Cloud Function normally using `fetch`.

AC:
- The Cloud Function that has the Supercharge rewards is always queried (no caching)

dievazqu:
### Description

* Obtaining available rewards using `fetch` instead of delegating that query to `rtk query`.
* Some fixes on HomeConsumerScreen (handling the correct token in some corner cases)

### Tested

* Manually using an emulator + mock distributions.
* Unit tested

### How others should test

Create a mock distribution, then you should be able to see the rewards to claim in the homecard + supercharge view.

### Related issues

- Fixes #2336

### Backwards compatibility

Full compatible


packages/mobile/src/consumerIncentives/ConsumerIncentivesHomeScreen.tsx
```diff
@@ -63,7 +64,7 @@ function useHasBalanceForSupercharge() {
     if (tokenUserInfo?.balance.gte(tokenConfig.minBalance)) {
       return {
         hasBalanceForSupercharge: true,
-        token: tokenUserInfo.symbol,
+        superchargingToken: tokenConfig,
```
gnardini, line 67: `superchargingToken` sounds like it will return the actual token, how about renaming to `superchargingTokenConfig`?

edit: I see that `useDefaultTokenToSupercharge` was also calling `token` to the `tokenConfig`. I do think that adding the `config` suffix makes it clearer, but I leave it up to you

packages/mobile/src/consumerIncentives/ConsumerIncentivesHomeScreen.tsx
```diff
@@ -201,30 +206,30 @@ function ClaimSuperchargeRewards({ rewards }: { rewards: SuperchargePendingRewar
...
-  const tokenToSupercharge = useTokenToSupercharge()
+  const defaultTokenToSupercharge = useDefaultTokenToSupercharge()
+  const tokenToSupercharge = superchargingToken ?? defaultTokenToSupercharge
 
-  const {
-    superchargeRewards,
-    isLoading: loadingAvailableRewards,
-    isError: errorLoadingRewards,
-  } = useFetchSuperchargeRewards()
   const claimRewardsLoading = useSelector((state) => state.supercharge.loading)
```
gnardini, line 225: small nit: I think we can do:
```
const {
 loading: claimRewardsLoading,
 availableRewards: superchargeRewards,
 fetchAvailableRewardsLoading: loadingAvailableRewards,
} = useSelector(state => state.supercharge)
```
which I think looks a bit better, but it's of course subjective

gnardini (approved):
Amazing work 👏 😄

ValoraQA:
Hey @MuckT for mock distribution we would require your help as we have already shared the account address in the slack previously 

Thanks..!!

PR #2354, opened 2022-04-25
Fix home cards

dievazqu:
### Description

Supercharge notifications are handled in-app instead of fetching them from admin-dashboard.
There are 3 use cases:
* There are rewards to be claimed (see screenshot 1)
* User doesn't meet all the conditions to supercharge (no enough funds or phone number not verified) (see screenshot 2)
* User has all the conditions to supercharge (see screenshot 3)

In every case, if the user clicks on the action, we are taking him to the supercharge page (ConsumerIncentiveHomeScreen).

### Tested

Manually in an emulator, running mock distributions. Also with unit tests.

### How others should test

Just open the app and see the home card. In case you have dismissed the message you would need to reinstall the app in order to see the notification again.

### Related issues

- Fixes #9473

### Backwards compatibility

Compatible, but the notifications may appear twice if we don't do the thing under `Important` (see below)

### Important
We should stop sending the supercharge notifications from the admin-dashboard once this PR is release. To do so, go to https://valora-admin-dashboard.web.app/, and then set as `Max Version` the version of the previous release for the notifications `Supercharge` and `SuperchargeEUR`.

### Screenshot
output-onlinepngtools
output-onlinepngtools (1)
output-onlinepngtools (2)


packages/mobile/src/home/NotificationBox.tsx
```diff
@@ -216,7 +296,7 @@ function useSimpleActions() {
               notificationId: id,
             })
             dispatch(openUrl(notification.ctaUri, notification.openExternal, true))
-            trackRewardsScreenOpenEvent(notification.ctaUri, RewardsScreenOrigin.NotificationBox)
+            // trackRewardsScreenOpenEvent(notification.ctaUri, RewardsScreenOrigin.NotificationBox)
```
gnardini, line 299: I think it's harmless to leave this but if you don't want to lets just delete it :)

gnardini (approved):
Looks great, left a couple small things because I'm annoying 😇

PR #2355, opened 2022-04-25
Append retried tests to index.html

MuckT:
### Description

Appends test results to index.html instead of overwriting on retries using the [append option in the config](https://github.com/Hargne/jest-html-reporter#configuration).

Update - 1: There were some issues with using the append option in [jest-html-reporter](https://github.com/Hargne/jest-html-reporter#readme); hopefully, https://github.com/Hargne/jest-html-reporter/pull/145 is approved and we can update to the latest version to resolve these issues.

Update - 2: The PR to jest-html-reporter was merged and the new version is used in this PR.

### Other changes

N/A

### Tested

Tested in CI.

### How others should test

Does not need to be tested by QA. Reviewers should view the [e2e test reports](https://github.com/valora-inc/wallet/actions/runs/2234203686) and check that all tests are present.

### Related issues

N/A

### Backwards compatibility

Yes


kathaypacific:
Hey @MuckT is this ready for review?  sorry if i'm reviewing too early but i had a look at the artifacts generated for the e2e reports but they're not what i expected (e.g. the artifact in https://github.com/valora-inc/wallet/pull/2355/checks?check_run_id=6199595548) - is this the intended artifact? In particular, the index.html in src doesn't seem to be for the e2e tests and the src folder seems to contain the source files which are quite big (16.8mb and 700+ files) and not really needed, is there a way to omit them?

MuckT:
@kathaypacific this is the correct [link](https://github.com/valora-inc/wallet/actions/runs/2235046803) and there should be two separate downloads [android-e2e-test-report](https://github.com/valora-inc/wallet/suites/6284716642/artifacts/224876206) and [ios-e2e-test-report](https://github.com/valora-inc/wallet/suites/6284716642/artifacts/224876207). It looks like the mobile tests results are also being named e2e, based on what you shared, but that's outside of the scope of this PR.

kathaypacific (approved):
🥳

ValoraQA:
Hey @MuckT  since Does not need to be tested by QA is mentioned in how other should test section , hence we are skipping the above PR

Thanks..!

PR #2364, opened 2022-04-26
revert edu onboarding change

### LINKED ISSUE #2357: End education onboarding screens experiment
Experiment didn't work, let's revert. Control group did slightly better, probly not statistically significant but it's what we wanted before we knew what worked, so let's stick with the control.

Should be as simple as reverting the experiment commit, but let's make sure the commit is single-purpose before pushing it.

AC:
- experiment is ended (control group is rolled out to 100%)

lisaqyli:
### Description
Reverting the edu onboarding change: the experiment result was about the same for the control/variant group hence the 

### Tested

### Related issues

- Fixes #2357 

### Backwards compatibility


packages/mobile/test/schemas.ts
```diff
@@ -947,7 +947,6 @@ export const v32Schema = {
   account: {
     ...v31Schema.account,
     hasLinkedBankAccount: false,
-    shouldSkipOnboardingEducationScreen: undefined,
```
cajubelt, line 950: probly doesn't matter since it looks like this was added by accident in the original PR (not seeing any redux state that made it into the final version of the PR). but in general this is the kind of thing to avoid reverting, since it's a schema migration-- instead, we'd need a new migration to drop the thing we're getting rid of

cajubelt (approved):
🥳

ValoraQA:
Hey @MuckT please let us know what need to be tested in the above PR
Thanks..!

PR #2373, opened 2022-04-27
enable lockFileMaintenance

lisaqyli:
### Description
enabling lockFileMaintenance on renovate per discussion here https://valora-app.slack.com/archives/C025V1D6F3J/p1651017289360609


renovate.json5
```diff
@@ -20,6 +20,9 @@
       groupName: 'devDependencies',
       // But exclude some specific packages from this group
       excludePackageNames: ['typescript'],
+      lockFileMaintenance: {
```
satish-ravi, line 23: I am not sure if lockFileMaintenance can be used for specific packages. From what I remember, I think it just deletes the lock file and regenerates it so all packages with semver ranges use the latest.

lisaqyli: afaik, i think it can be used for specific packages, I saw a bunch of examples on github of other people using it this way -> https://github.com/search?q=lockFileMaintenance&type=code

satish-ravi: interesting, wonder how this works. Does it upgrade all sub dependencies of matching dependencies on the lock files? It sounds like an expensive operation to do, but we can find out.

satish-ravi (approved):

jeanregisser:
Closing in favor of #2402

PR #2399, opened 2022-05-04
Use signed typed data for registration endpoint auth

kathaypacific:
### Description

For the registration endpoint, we currently use a signed message for authentication. As previously discussed, the main security concern with a signed message is “domain collisions” where a message signed for one purpose/context overlaps with a message signed for a different purpose/context. Therefore, this PR uses a signed typed message instead.

Corresponding backend PR https://github.com/valora-inc/valora-rest-api/pull/80 (these should be merged together to avoid breaking changes)

### Other changes

N/A

### Tested

Manually + updated unit tests

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

N/A


jeanregisser (approved):
Awesome! 💯

PR #2398, opened 2022-05-04
Fix vulnerability in tmpl version

kathaypacific:
### Description

In response to this vulnerability report, https://github.com/advisories/GHSA-jgrx-mgxx-jf9v, deleted the transitive dependency in yarn.lock (as it was not locked to a specific version) and yarn'd again.

### Other changes

N/A

### Tested

N/A

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

N/A

PR #2403, opened 2022-05-04
Mock user location data in E2E tests

MuckT:
### Description

Mocks the user location data so a consistent list of providers and exchanges are served during the e2e tests. This will make it easier to run tests locally regardless of developer location while still testing the integration of services dependent on user location.

### Other changes

- Snapshots updated
- Additional tests for Bidali and Providers added.
- Includes changes from #2398 to unblock CI.

### Tested

Tested locally on iOS.

### How others should test

Run the E2E tests locally and ensure that they pass.

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/packages/mobile/e2e/src/usecases/OnRamps.js
+++ b/packages/mobile/e2e/src/usecases/OnRamps.js
@@ -1,4 +1,5 @@
 import { reloadReactNative } from '../utils/retries'
+import { waitForElementId } from '../utils/utils'
 
 export default onRamps = () => {
   beforeEach(async () => {
@@ -10,9 +11,7 @@ export default onRamps = () => {
 
   describe('cUSD', () => {
     beforeEach(async () => {
-      await waitFor(element(by.id('radio/cUSD')))
-        .toBeVisible()
-        .withTimeout(10 * 1000)
+      await waitForElementId('radio/cUSD')
       await element(by.id('radio/cUSD')).tap()
     })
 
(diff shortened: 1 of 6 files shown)


packages/mobile/e2e/src/usecases/OnRamps.js
```diff
@@ -117,18 +204,26 @@ export default onRamps = () => {
...
-          .withTimeout(30 * 1000)
-        await expect(element(by.id('provider-Binance'))).toBeVisible()
-        await expect(element(by.id('provider-Bittrex'))).toBeVisible()
-        await expect(element(by.id('provider-Coinbase (CELO as CGLD)'))).toBeVisible()
-        await expect(element(by.id('provider-Coinbase Pro (CELO as CGLD)'))).toBeVisible()
-        await expect(element(by.id('provider-CoinList Pro'))).toBeVisible()
-        await expect(element(by.id('provider-OKCoin'))).toBeVisible()
-        await expect(element(by.id('accountBox'))).toBeVisible()
+        await waitForElementId('accountBox')
+        await expect(element(by.id('provider-Binance'))).toExist()
```
lisaqyli, line 209: We were talking in eng sync, maybe one of the idea of preventing the e2e test from failing when the response from the cloud func changed is having less specific assertions. We can assert on there are list of exchange providers instead of specifying which exchange they are. In this case I feel pretty comfortable as long as the cloud func returns some exchanges, the test case we are testing here should be passing

MuckT:
Closed in favor of #2407. A larger discussion around mocking is in #2421.

PR #2435, opened 2022-05-09
chore: bye bye monorepo

jeanregisser:
### Description

This fully removes the monorepo structure 🚝 💣 💥 

Here's what was done:
- move everything from `packages/mobile` to the repository root
- file/folder names that were the same between these 2 folders, were merged (`package.json`, `scripts`,  `jest.config.js`, etc)
- adjust all paths accordingly (GitHub workflows, jest, metro, TS, Xcode, Android gradle files, etc)
- remove lerna
- remove some old unused scripts

The app behavior should remain the same, since only paths were adjusted.

For developers, the impact should be minimal, the main thing that changes is the working directory. No need to work in `packages/mobile` anymore.
All scripts we had remain.

### Follow up

While working on this, I came across 2 surprising issues:
- `yarn install` [started failing on GH actions](https://github.com/valora-inc/wallet/runs/6351393671?check_suite_focus=true), while working locally, this is not the first time I see this and is caused by prepare steps of some dependencies which trigger `yarn install` multiple times while the initial `yarn install` is still running, https://github.com/yarnpkg/yarn/issues/6312. As a workaround `yarn install` is retried to avoid this issue. We should see if newer versions of yarn handle this better.
- The final app bundle size changed. It is caused by packages hoisting happening differently without the monorepo structure. It bundles multiple versions of the same package, because the ordering is different. It's mostly happening with `@celo/` packages + `bn.js`. The bundle size was initially 1.1 MB bigger, but I managed to decrease it by setting a yarn resolution for `bn.js`. We should take a further look at this next time we upgrade `@celo/` packages to avoid making the app bundle bigger than necessary.

### Tested

- CI passes
- Locally:
  - `yarn dev:ios`
  - `yarn dev:android`

### How others should test

The wallet works as before.

### Related issues

- Discussed on [Slack](https://valora-app.slack.com/archives/C025V1D6F3J/p1651251065521769).

### Backwards compatibility

Yes


silasbw (approved):

kathaypacific (approved):
I didn’t look at every file but checked out locally and built/ran the app, it works! Thanks Jean, this is awesome!!! We can finally have a party 🥳

ValoraQA:
Hey @MuckT we will cover the above PR in BVT and regression testing

BVT links :
Android 11 : https://valoraapp.testrail.io/index.php?/runs/view/121&group_by=cases:section_id&group_order=asc
Android 10 : https://valoraapp.testrail.io/index.php?/runs/view/122&group_by=cases:section_id&group_order=asc
iOS 14 : https://valoraapp.testrail.io/index.php?/runs/view/123&group_by=cases:section_id&group_order=asc
iOS 15 : https://valoraapp.testrail.io/index.php?/runs/view/124&group_by=cases:section_id&group_order=asc

PR #2457, opened 2022-05-11
fix(walletConnect): show decoded message for personal_sign

### LINKED ISSUE #2439: Bug: decoding message for personal_sign
### Current behavior
A dapp developer [here](https://discord.com/channels/886972503560433675/887422238071087154/971603129403596870) reported that the message string is not decoded when displaying it to the user to sign a transaction. So the user just gets an unreadable hex string to sign, rather than the message the dapp developer sent with `personal_sign`. 

### Desired behavior
The decoded message should be displayed to the user when they tap "more details"

ENG:
- see ActionRequest.tsx

### Steps to Reproduce
TODO find steps to repro

### Versions
TODO try repro on latest version

satish-ravi:
### Description

Show decoded message for personal_sign requests, which currently show up as hex string

Existing:
personal_sign_data_hex

New:
personal_sign_data_fixed

### Other changes

N/A

### Tested

Manually running a dapp locally to sign personal message

### How others should test

Run any dapp that uses wallet connect and invokes the personal_sign action. One way to run a dapp locally is to use this [example app](https://github.com/critesjosh/valora-wallet-connect-v1) and replace the send cusd transaction call with a eth.personal.sign_message call.

When the personal_sign action is invoked, click on the "Show details" button to see the actual message being signed instead of the hex string

### Related issues

- Fixes #2439 

### Backwards compatibility

Yes


src/walletConnect/screens/ActionRequest.tsx
```diff
@@ -83,7 +83,7 @@ function ActionRequest({ navigation, route: { params: routeParams } }: Props) {
       : method === SupportedActions.personal_decrypt
       ? Buffer.from(params[1]).toString('hex')
       : method === SupportedActions.personal_sign
-      ? params[0]
+      ? Buffer.from(params[0].replace(/^0x/, ''), 'hex').toString()
```
satish-ravi, line 86: If the decoded message is an empty string (in case of invalid hex), this results in an empty string and causes the <Text> component to crash. Not sure if this is possible at all in practice since personal sign always sends a hex string, but should we still be defensive to it?

cajubelt: ideally yes

cajubelt (approved):
nice!!

src/walletConnect/screens/ActionRequest.tsx
satish-ravi: kk, any preference on what to show in that case, the raw hex? or just make it `null` so "Show Details" button won't show up? or something else?

satish-ravi: chatted offline and decided to go with "Invalid message" string.

cajubelt (commented):
🥳

src/walletConnect/screens/ActionRequest.tsx
```diff
@@ -83,7 +83,7 @@ function ActionRequest({ navigation, route: { params: routeParams } }: Props) {
       : method === SupportedActions.personal_decrypt
       ? Buffer.from(params[1]).toString('hex')
       : method === SupportedActions.personal_sign
-      ? params[0]
+      ? Buffer.from(params[0].replace(/^0x/, ''), 'hex').toString() || t('action.invalidMessage')
```
jeanregisser, line 86: I propose we display the raw param in case hex decoding fails, instead of "Invalid message", which is not entirely correct.
Just in case the dapp wants to sign something that's not a string. Not sure how likely it is to happen but at least we try to display it.
And the last fallback would be "Empty message" if still `null | undefined | ''`

```suggestion
      ? Buffer.from(trimLeading0x(params[0]), 'hex').toString() || params[0] || t('action.emptyMessage')
```

We also have a `trimLeading0x` util.

src/walletConnect/screens/ActionRequest.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { SupportedActions } from 'src/walletConnect/constants'
+import ActionRequest from 'src/walletConnect/screens/ActionRequest'
+import { Actions } from 'src/walletConnect/v1/actions'
+import { createMockStore, getMockStackScreenProps } from 'test/utils'
+
+describe('ActionRequest', () => {
+  const store = createMockStore({})
+
+  beforeEach(() => {
+    store.dispatch = jest.fn()
```
jeanregisser, line 15: nit: this works but you can also avoid it and then use `store.getActions()`. Example: https://github.com/valora-inc/wallet/blob/2b28b691640c52bdeeb1b49a7751c5bdf3b764fe/src/exchange/WithdrawCeloReviewScreen.test.tsx#L41-L48

jeanregisser (commented):
Love this! 💯

jeanregisser (approved):
🚀

PR #2460, opened 2022-05-12
Removing hardcoded percentage

dievazqu:
### Description

There is a supercharge translation that had the apy percentage hardcoded. After this change, it will fetch the apy from the redux state.

### How others should test
The behavior of the app shouldn't change at all. Specifically, make sure that the supercharging homecard is showing the correct percentage.


kathaypacific (approved):
🙏🏻

lisaqyli (approved):

dievazqu:
> Great!
> 
> Could you add the "How others should test" section in the PR description? the QA team is relying on it to know whether to test it or not. If not present they are going to ask.
> 
> 🙏

I added a section, but there is no change to test, basically they have to make sure that app is working as it was working

jeanregisser:
You can mention: "No testing needed" or N/A

ValoraQA:
hey @MuckT as discussed as it is not be tested we are skipping this PR

Thanks..!!
Cc: @dievazqu

PR #2464, opened 2022-05-13
bank details screen

### LINKED ISSUE #2333: Enter Fiat Details Screen
AC:
* After a user taps on a FiatConect provider in the SelectPayments screen they are taken to a new screen to enter their fiat account details
* A user can enter their account details, click 'Save & Continue', and their fiatAccount details are saved with the fiatconnect partner.
* All[ response types and error codes from the spec](https://github.com/fiatconnect/specification/blob/main/fiatconnect-api.md#33312-responses) are handled or logged including success, ResourceExists, UnsupportedSchema, InvalidSchema
* For the first FiatAccount schema add basic syntax checking. (Should be some in the schema)

ENG:
* Tapping 'Save & Continue' does not need to take the user to the next screen yet. 
* Run the "Validation" when the user presses 'Save and Continue'

[Figma](https://www.figma.com/file/bwIY7peeykI8K4wTrXs4iN/Fiat-Connect?node-id=940%3A17381)

lisaqyli:
### Description
This is the screen for user to manually type in their fiat account information. It includes validation and posting to FiatConnect's [Fiat Account endpoint ](https://github.com/fiatconnect/specification/blob/main/fiatconnect-api.md#343-fiat-account-endpoints)

TODO:
1. I made up some of the wordings including error message, placeholder text, system message, plan to get Nitya to review before 
2. We haven't implemented the screen after this one or the one before, the provider selected right now is hardcoded, will need to update the navigation stack once the screens are connected
3. There's a newly added FiatAccountSchema that has not been added to the types yet, need to add to the fiatconnect-types and add that schema

### Tested
unit tests

### How others should test
This component is not accessible in the app yet. It is a future flow that has not been fully built out

### Related issues
- Fixes #2333


lisaqyli:
When screen just launched
Screenshot_20220513_101437_org celo mobile alfajores dev

With validation errors
Screenshot_20220513_101437_org celo mobile alfajores dev

src/fiatconnect/BankDetailsScreen.tsx
```diff
@@ -0,0 +1,275 @@
...
+          Logger.error(TAG, `Error adding fiat account: ${errors}`)
+        })
+
+      // TODO: navigate to the next screen
+    }
+  }
+
+  const onPressSelectPaymentOption = () => {
+    // TODO: tracking here
+    // todo: naviagte to select payment options screen
```
jh2oman, line 119: This says 'Selected payment method' and i don't think its intended to be tappable

src/fiatconnect/index.ts
```diff
@@ -27,3 +28,20 @@ export async function getFiatConnectProviders(): Promise<FiatConnectClientConfig
...
+  const response = await fetch(`${providerURL}/accounts/${fiatAccountSchema}`, {
+    // todo: add auth?
+    method: 'POST',
+    headers: {
+      'content-Type': 'application/json',
+    },
+    body: JSON.stringify({ ...properties }),
+  })
+
+  return response.json()
```
jh2oman, line 46: This is going to use the SDK, so I think for now its appropriate to just throw a 'Not Implemented' error for this whole function

src/fiatconnect/BankDetailsScreen.tsx
```diff
@@ -0,0 +1,275 @@
...
+      name: 'accountNumber',
+      label: i18n.t('fiatAccountSchema.accountNumber.label'),
+      regex: /^\d{10}$/,
+      placeholderText: i18n.t('fiatAccountSchema.accountNumber.placeholderText'),
+      errorMessage: i18n.t('fiatAccountSchema.accountNumber.errorMessage'),
+    },
+  ],
+}
+
+const BankDetailsScreen = ({ route, navigation }: Props) => {
```
jh2oman, line 62: Can you call this file / component something like `FiatDetailsScreen` instead. Bank implies its only for bank accounts and could not be used for mobile money or credit cards

src/fiatconnect/BankDetailsScreen.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <BankDetailsScreen {...mockScreenProps} />
+      </Provider>
+    )
+
+    const fakeAccountName = 'ma loaded bank account'
+    const fakeInstitutionName = 'CapitalTwo Bank'
+    const fakeAccountNumber = '1234567890'
+    fireEvent.changeText(getByTestId('input-accountName'), fakeAccountName)
+    fireEvent.changeText(getByTestId('input-institutionName'), fakeInstitutionName)
+    fireEvent.changeText(getByTestId('input-accountNumber'), '1234567890')
```
jh2oman, line 77:
```suggestion
    fireEvent.changeText(getByTestId('input-accountNumber'), fakeAccountNumber)
```

src/fiatconnect/BankDetailsScreen.tsx
```diff
@@ -0,0 +1,275 @@
...
+    }
+  }
+
+  const onPressSelectPaymentOption = () => {
+    // TODO: tracking here
+    // todo: naviagte to select payment options screen
+  }
+
+  const validateInput = () => {
+    setValidInputs(false)
```
jh2oman, line 123: Does this cause some weird flashing of the 'Next' button as the user types their fields in? Maybe it should only be set to invalid if an error is actually found

jh2oman (commented):
Looking great so far!

src/fiatconnect/BankDetailsScreen.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { createMockStore, getMockStackScreenProps } from 'test/utils'
+import BankDetailsScreen from './BankDetailsScreen'
+
+jest.mock('src/fiatconnect', () => ({
+  ...(jest.requireActual('src/fiatconnect') as any),
+  addNewFiatAccount: jest.fn(() => Promise.resolve()),
+}))
+
+const store = createMockStore({})
+const providerURL = 'https://superLegitCICOProvider.com'
```
cajubelt, line 16: LOL

cajubelt:
```suggestion
const providerURL = 'https://superLegitCICOProvider.valoraapp.com'
```
because I'm paranoid

src/fiatconnect/BankDetailsScreen.tsx
```diff
@@ -0,0 +1,275 @@
...
+      name: 'institutionName',
+      label: i18n.t('fiatAccountSchema.institutionName.label'),
+      regex: /.*?/,
+      placeholderText: i18n.t('fiatAccountSchema.institutionName.placeholderText'),
+      errorMessage: i18n.t('fiatAccountSchema.institutionName.errorMessage'),
+    },
+    {
+      name: 'accountNumber',
+      label: i18n.t('fiatAccountSchema.accountNumber.label'),
+      regex: /^\d{10}$/,
```
cajubelt, line 55:
```suggestion
      regex: /^\[0-9]{10}$/,
```
We don't want to include non-latin digit characters. https://stackoverflow.com/a/273144/18163451

src/fiatconnect/BankDetailsScreen.tsx
cajubelt: ➕ Lisa I think you may have misread the designs which say "selected", not "select". By this point in the flow, the user should already have selected the provider. (Otherwise we wouldn't know which fiat account schema to use!)

src/fiatconnect/BankDetailsScreen.tsx
cajubelt: ➕

src/fiatconnect/BankDetailsScreen.tsx
```diff
@@ -0,0 +1,275 @@
...
+      inputRefs.current.push('')
+    }
+    return fields
+  }, [fiatAccountSchema])
+
+  const onPressNext = async () => {
+    validateInput()
+
+    if (validInputs) {
+      const body: { [k: string]: string } = {}
```
cajubelt, line 95:
```suggestion
      const body: Record<string, string> = {}
```
code golf

src/fiatconnect/BankDetailsScreen.tsx
```diff
@@ -0,0 +1,275 @@
...
+    },
+  ],
+}
+
+const BankDetailsScreen = ({ route, navigation }: Props) => {
+  const { t } = useTranslation()
+  const { providerURL, fiatAccountSchema } = route.params
+  const [validInputs, setValidInputs] = useState(false)
+  const [textValue, setTextValue] = useState('')
+  const [errors, setErrors] = useState(new Set())
```
cajubelt, line 67:
```suggestion
  const [errors, setErrors] = useState(new Set<string>())
```

src/fiatconnect/BankDetailsScreen.tsx
```diff
@@ -0,0 +1,275 @@
...
+  }
+
+  const onPressSelectPaymentOption = () => {
+    // TODO: tracking here
+    // todo: naviagte to select payment options screen
+  }
+
+  const validateInput = () => {
+    setValidInputs(false)
+    const newErrorSet = new Set(errors)
```
cajubelt, line 124: wouldn't it be simpler to just use `errors` instead of copying it over to this temp variable and then over-writing it? Is there some downside I'm not seeing?

src/fiatconnect/BankDetailsScreen.tsx
```diff
@@ -0,0 +1,275 @@
...
+        newErrorSet.delete(field.name)
+      }
+    })
+
+    setErrors(newErrorSet)
+    if (newErrorSet.size > 0) {
+      return
+    } else {
+      setValidInputs(true)
+    }
```
cajubelt, line 141:
```suggestion
    setValidInputs(newErrorSet.size === 0)
```
code golf

src/fiatconnect/BankDetailsScreen.tsx
```diff
@@ -0,0 +1,275 @@
...
+  }
+
+  const setInputValue = (value: string, index: number) => {
+    const inputs = inputRefs.current
+    inputs[index] = value
```
cajubelt, line 146:
```suggestion
    inputRefs.current[index] = value
```
code golf: inline variables

cajubelt (commented):
Looks good! Just left some minor recs. The input validation setup is pretty nifty

src/fiatconnect/BankDetailsScreen.tsx
lisaqyli: oh yeah good catch on the typo, i remember nitya told me before if you tap on this, you go back to the provider selection page and you can change the provider selection

src/fiatconnect/BankDetailsScreen.tsx
lisaqyli: https://stackoverflow.com/questions/58806883/how-to-use-set-with-reacts-usestate

lisaqyli: weird thing with useState that I also learned this time

src/fiatconnect/BankDetailsScreen.tsx
lisaqyli: to be honest, the button is covered by input keyboard when user is typing.

If it's not covered, I imagine it would only flash if the field value can change between invalid and valid as user type? which feels unlikely to me?

locales/base/translation.json
```diff
@@ -1262,5 +1262,28 @@
...
+      "errorMessage": "An account name is required"
+    },
+    "institutionName": {
+      "label": "Bank Name",
+      "placeholderText": "Enter Bank Name",
+      "errorMessage": "A bank name is required"
+    },
+    "accountNumber": {
+      "label": "Account Number",
+      "placeholderText": "Enter your 10 digits account number",
```
jh2oman, line 1285:
```suggestion
      "placeholderText": "Enter your 10 digit account number",
```

src/fiatconnect/FiatDetailsScreen.test.tsx
```diff
@@ -0,0 +1,93 @@
+import { FiatAccountSchema } from '@fiatconnect/fiatconnect-types'
+import { fireEvent, render } from '@testing-library/react-native'
+import * as React from 'react'
+import { Provider } from 'react-redux'
+import { addNewFiatAccount } from 'src/fiatconnect'
+import { Screens } from 'src/navigator/Screens'
+import { createMockStore, getMockStackScreenProps } from 'test/utils'
+import FiatDetailsScreen from '../FiatDetailsScreen'
```
jh2oman:
```suggestion
import FiatDetailsScreen from './FiatDetailsScreen'
```

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -0,0 +1,273 @@
...
+  name: string
+  label: string
+  regex: RegExp
+  placeholderText: string
+  errorMessage: string
+}
+
+// This is a mapping between different fiat account schema to the metadata of the fields that need to be rendered on the bank details screen
+const SCHEMA_TO_FIELD_METADATA_MAP = {
+  AccountNumber: [
```
jh2oman, line 37: to ensure it matches
```suggestion
  [FiatAccountSchema.AccountNumber]: [
```

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -0,0 +1,273 @@
...
+          Logger.error(TAG, `Error adding fiat account: ${error}`)
+        })
+
+      // TODO: navigate to the next screen
+    }
+  }
+
+  const onPressSelectedPaymentOption = () => {
+    // TODO: tracking here
+    // todo: naviagte to select payment options screen
```
jh2oman, line 121: Why is this a TODO, you should be able to do it now right?

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -0,0 +1,273 @@
...
+
+      <View style={styles.footer}>
+        <View style={styles.paymentOption}>
+          <BorderlessButton onPress={onPressSelectedPaymentOption} testID="selectedProviderButton">
+            <View style={styles.paymentOptionButton}>
+              <Text style={styles.paymentOptionText}>
+                {t('fiatDetailsScreen.selectedPaymentOption')}
+              </Text>
+              <ForwardChevron color={colors.gray4} />
+              {/* TODO: This is a hardcoded logo, replace the logo uri with the uri from the payment provider selected */}
```
jh2oman, line 182: Can you add the image url into props? That might not be where it lives in the end, but it helps ensure it won't be forgotten.

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -0,0 +1,273 @@
...
+    validateInput()
+
+    if (validInputs) {
+      const body: Record<string, string> = {}
+      for (let i = 0; i < formFields.length; i++) {
+        body[formFields[i].name] = inputRefs.current[i]
+      }
+
+      // TODO: some schema requires additional fields in body in addition
+      // to the ones collected from the form, eg. country code
```
jh2oman, line 101: Why is this a TODO? Can't you get the country code from redux here?

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -0,0 +1,273 @@
...
+  const onPressNext = async () => {
+    validateInput()
+
+    if (validInputs) {
+      const body: Record<string, string> = {}
+      for (let i = 0; i < formFields.length; i++) {
+        body[formFields[i].name] = inputRefs.current[i]
+      }
+
+      // TODO: some schema requires additional fields in body in addition
```
jophish, line 100: We'll also need to set the `fiatAccountType` to the [fixed value](https://github.com/fiatconnect/specification/blob/main/fiatconnect-api.md#8321-accountnumber) defined in the spec, depending on the schema being used.

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -0,0 +1,273 @@
...
+interface Fields {
+  name: string
+  label: string
+  regex: RegExp
+  placeholderText: string
+  errorMessage: string
+}
+
+// This is a mapping between different fiat account schema to the metadata of the fields that need to be rendered on the bank details screen
+const SCHEMA_TO_FIELD_METADATA_MAP = {
```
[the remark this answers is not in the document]
jophish: ah, I see the comments below, nvm!

src/fiatconnect/FiatDetailsScreen.tsx
lisaqyli: coz we haven't implemented the quotes end point for FC, I didn't want to assume how amount is passed into the component

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -0,0 +1,330 @@
...
+    if (validInputs) {
+      const body: Record<string, string> = {}
+      for (let i = 0; i < formFields.length; i++) {
+        body[formFields[i].name] = inputRefs.current[i]
+      }
+
+      const validatedBody = validateAndCompleteSchema(
+        body,
+        getSchemaObjectByType(fiatAccountSchema)
+      )
```
jh2oman, line 132: This seems overly complicated for just adding two new values to the body, and it runs on all schemas when the additions are specific to AccountNumber.

You could just have a function like this
```js
const implicitFields = (fiatAccountSchema) => {
switch (fiatAccountSchema) {
    case FiatAccountSchema.AccountNumber:
      return { country: userCountry.countryCodeAlpha2, fiatAccountType: FiatAccountType.BankAccount }
      break
    default:
      return {}
  }
}
```

and then do this here

```js
const completeBody = {
...body,
...implicitFields(fiatAccountSchema)
}
```

jh2oman: I'd actually prefer it if you added fields like `country` and `fiatAccountType` to SCHEMA_TO_FIELD_METADATA_MAP and have the metadata for those fields include functions to get the correct values and let the rest of the UI know not to display them. That way all of the schema is in one place and much easier to read and verify.

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -0,0 +1,330 @@
...
+          body[key] = FiatAccountType.BankAccount
+        }
+      }
+    }
+    return body
+  }
+
+  const onPressSelectedPaymentOption = () => {
+    // TODO: tracking here
+    // TODO: navigate to SelectProvider screen
```
jh2oman, line 180: This just goes to this screen right? https://github.com/valora-inc/wallet/blob/main/src/fiatExchanges/SelectProvider.tsx

I don't think the props for that will change much, so i I think it can safely be implemented now

src/fiatconnect/index.test.ts
```diff
@@ -29,4 +30,34 @@ describe('FiatConnect helpers', () => {
       expect(providers).toEqual([])
     })
   })
+
+  describe('addNewFiatAccount', () => {
+    // TODO: unskip after integration with FC SDK
+    // eslint-disable-next-line jest/no-disabled-tests
+    it.skip('returns a fiat account info with fiat account id on success', async () => {
```
jh2oman, line 37: Instead of skipping and having a todo can you run the test and check for the thrown `'Not implemented'` error? That way when its implemented the test will fail and we will be sure not to miss this.

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -0,0 +1,306 @@
...
+interface Fields {
+  name: string
+  label: string
+  regex: RegExp
+  placeholderText: string
+  errorMessage: string
+}
+
+// This is a mapping between different fiat account schema to the metadata of the fields that need to be rendered on the bank details screen
+const SCHEMA_TO_FIELD_METADATA_MAP = {
```
jh2oman, line 44: This also doesn't take advantage of many of the Types we have for FiatConnect, and makes it so that the params for a given Schema are spread out in this const and implicitFields(). It would be nice if we could combine them both in a data structure.

While thinking through this, I went a bit ham and almost fully implemented a new structure. let me know what you think

```ts
interface FormParameter {
  label: string
  regex: RegExp
  placeholderText: string
  errorMessage: string
}
interface ImplicitParameter<T> {
  value: T
}

interface SchemaMetadata {
  [FiatAccountSchema.AccountNumber]: {
    [Property in keyof AccountNumber]: FormParameter | ImplicitParameter<AccountNumber[Property]>
  }
}

const getSchemaMetadata = (country: string, fiatAccountType: FiatAccountType): SchemaMetadata => ({
  [FiatAccountSchema.AccountNumber]: {
    accountName: {
      label: i18n.t('fiatAccountSchema.accountName.label'),
      regex: /.*?/,
      placeholderText: i18n.t('fiatAccountSchema.accountName.placeholderText'),
      errorMessage: i18n.t('fiatAccountSchema.accountName.errorMessage'),
    },
    institutionName: {
      label: i18n.t('fiatAccountSchema.institutionName.label'),
      regex: /.*?/,
      placeholderText: i18n.t('fiatAccountSchema.institutionName.placeholderText'),
      errorMessage: i18n.t('fiatAccountSchema.institutionName.errorMessage'),
    },
    accountNumber: {
      label: i18n.t('fiatAccountSchema.accountNumber.label'),
      regex: /^[0-9]{10}$/,
      placeholderText: i18n.t('fiatAccountSchema.accountNumber.placeholderText'),
      errorMessage: i18n.t('fiatAccountSchema.accountNumber.errorMessage'),
    },
    country: { value: country},
    fiatAccountType: { value: fiatAccountType}
  }
})
```

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -139,12 +139,14 @@ const FiatDetailsScreen = ({ route, navigation }: Props) => {
         body[formFields[i].name] = inputRefs.current[i]
       }
 
+      const implicitBody: Record<string, string> = {}
```
jh2oman, line 142: The values for these are not necessarily strings, since it takes whatever `param.value` is from the schema. Thats one disadvantage to doing a forEach loop here is that you have to explicitly type the variable

jh2oman (approved):
Great job with this one @lisaqyli !

ValoraQA:
Hey @MuckT we will check this PR once it is available for testing fully

PR #2476, opened 2022-05-17
Drop trailing / on blockchain-api URL

silasbw:
I'm debugging Sentry tracing and noticed an extra / in Sentry

### How others should test

Check that the list of transactions appears on the home screen.


MuckT (approved):
Thanks for grabbing this!

PR #2477, opened 2022-05-17
Fix Sentry tracing

silasbw:
Use domain names / hostnames instead of URLs.

### Tested

Testing locally by using a mainnetdev config, hard coding `tracesSampleRate` to 1.0, running in an Android emulator, and then checking Sentry.

### How others should test

No QA testing needed.


jeanregisser (approved):
Cool! 👍 

This reminds me, we started discussing this in the past, but I feel we should be tracing all domains. This way we could know which 3rd party API call is actually slow. Affecting the experience of our users.
We can discuss on Slack.

dievazqu (approved):

PR #2478, opened 2022-05-17
Remove google-libphonenumber dependency

MuckT:
### Description

Remove google-libphonenumber dependency as we do not import the lib in src code.

### Other changes

N/A

### Tested

Tested locally on iOS.

### How others should test

Same testing steps as Côte d’Ivoire Phone Number Fix #2471

### Related issues

N/A

### Backwards compatibility

Yes


MuckT:
@jeanregisser was able to remove the dependency entirely and have the correct behavior.

PR #2480, opened 2022-05-17
fix: incorrect token shown when opening a transfer push notification

### LINKED ISSUE #1968: Bugfix: Opening transfer push notification opens the wrong screen
The notification-service is sending a field `tokenAddress` to the app, but the app is expecting a `currency` field and since it can't find one it defaults to cUSD. This causes any transfer push notification to open a screen that says cUSD is the transfer token even if it's not.

AC: 
- Opening a transfer push notification opens the review screen and shows the right amount/token
  - The wallet should probably use the new [TransactionDetailsScreen](https://github.com/valora-inc/wallet/blob/2b28b691640c52bdeeb1b49a7751c5bdf3b764fe/src/transactions/feed/TransactionDetailsScreen.tsx#L62) which directly supports `tokenAddress`

jeanregisser:
### Description

This fixes the token shown when opening a transfer push notification, when the token sent is not cUSD.

Related change: https://github.com/valora-inc/notification-service/pull/21

### Tested

**before/after**

Sending cEUR to an account and opening the push notification:

Note: this was on Alfajores were the EUR/USD rate is a bit funky.

### How others should test

1. Send a token which is not cUSD (for instance cEUR or [TT](https://test-token-faucet.web.app/)) to another account
2. On the receiving account device, tap the push notification
3. The opened screen should show the actual token sent, and not cUSD.

### Related issues

- Fixes #1968

### Backwards compatibility

Yes (requires https://github.com/valora-inc/notification-service/pull/21 to be deployed).


src/firebase/notifications.ts
```diff
@@ -59,28 +58,28 @@ function* handlePaymentReceived(
...
-        ),
-        type: TokenTransactionType.Received,
-      }
-    )
+        metadata: {
+          title: transferNotification.name,
+          image: transferNotification.imageUrl,
+          comment: transferNotification.comment,
+        },
+        fees: [], // TODO?
```
kathaypacific, line 80: i think it's fair enough to omit fees 😅 it'd be strange to see that fees were deducted from a payment

kathaypacific (approved):
🥳 yaaaaayyyyy i'm so happy this is finally fixed 🚀

PR #2482, opened 2022-05-17
Fix breaking change for supercharge translation

kathaypacific:
### Description

Context https://valora-app.slack.com/archives/C025V1D6F3J/p1652707627924639?thread_ts=1652392728.301359&cid=C025V1D6F3J

This PR ensures backwards compatibility for OTA translations for app versions <= 1.32

### Other changes

N/A

### Tested

N/A

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

N/A


locales/base/translation.json
```diff
@@ -199,7 +199,8 @@
   "superchargeClaimFailure": "There was an issue claiming your Supercharge rewards.",
   "superchargeNotificationBody": "Your Supercharge rewards are here! Claim them now.",
   "superchargeNotificationStart": "Claim rewards",
-  "superchargingNotificationBody": "You're getting {{apy}}% rewards with Supercharge ⚡",
+  "superchargingNotificationBody": "You're getting 25% rewards with Supercharge ⚡",
+  "superchargingNotificationDesscription": "You're getting {{apy}}% rewards with Supercharge ⚡",
```
jeanregisser, line 203: Looks like we could name them with VX suffix to easily understand they are different to keep backward compat.

We could even use `superchargingNotificationBodyV1_33_0` to be even clearer. 
What do you think?

```suggestion
  "superchargingNotificationBodyV2": "You're getting {{apy}}% rewards with Supercharge ⚡",
```

jeanregisser (commented):
Thanks!

locales/base/translation.json
kathaypacific: yeah this is a great idea! i do like the V1_33_0 idea (or maybe just v1_33 since we're unlikely to introduce a breaking change in a patch release?) - was thinking that V2 is nicer to read but having the app version there makes it easier for us to know when we can safely start to remove some old keys (if they're like 20 app releases old for example)

kathaypacific: the only thing i don't like is in the future we might have a lot of translation keys with version numbers in them because once they're there, we can't remove them without another breaking change :(  it's just not so nice to look at, but it does the job of keeping breaking translation changes top of mind, lol

jeanregisser (approved):

locales/base/translation.json
jeanregisser: yep, at least after some time, we can remove the older ones knowing in which versions they were used.

PR #2488, opened 2022-05-17
Setup celoWithdrawalEnabledInExchange Experiment

### LINKED ISSUE #2429: Experiment: remove withdrawals from CELO
See experiment design here for context: 
https://www.notion.so/valora-inc/All-Experiments-c82002f042504ca8a37f420fcf6affc2?p=6c71d4e074504bad8134fdab0b3e6c07

* Release under A/B test the following change:
   * 50% of users should be allocated into 'Control', to see current experience
   * 50% of users should be allocated into 'Treatment', to see CELO page without withdrawals 

Analytics: 
- [ ]  I should be able to measure withdrawal rate, where a withdraw may happen from either flow (i.e., an event should fire whenever a user sends to an exchange — is this possible without looking at the raw blockchain data? cc: @Gowtham Reddy)
- [ ]  I should be able to measure drop-off at each stage of the withdraw-to-exchange flow
    - [ ]  Impressions on the ‘exchanges’ page
    - [ ]  Click through to any of the exchanges (Do people bounce after they click through, or come back?)
    - [ ]  Clicks to ‘send’ on exchanges page (Are people distracted away from withdrawing by the exchange links?)
    - [ ]  Clicks to ‘review’ on send page (Are people distracted away from withdrawing to exchange by seeing the send flow with their contacts?)

MuckT:
### Description

Setups up the remote config options needed for [Experiment: remove withdrawals from CELO #2429](https://github.com/valora-inc/wallet/issues/2429).

### Other changes

Added `celo_withdraw_completed` analytic event to the main send flow after successfully sending CELO.
[Draft Experiment in Firebase](https://console.firebase.google.com/u/0/project/celo-mobile-mainnet/config/experiment/results/28)

### Tested

### How others should test

On Alfajores where `celoWithdrawalEnabledInExchange` will be set as true.
Navigate to Hamburger Menu -> CELO Screen  and observe 'Withdraw CELO' is visible.

On Mainnet an A/B experiment will run where some users see the 'Withdraw CELO' on the CELO Screen and others won't.

### Related issues

- Fixes #2429

### Backwards compatibility

Yes - however celo_withdraw_completed will now include CELO sent to another wallet via the main send flow.


src/exchange/ExchangeHomeScreen.tsx
```diff
@@ -93,6 +93,9 @@ function ExchangeHomeScreen({ navigation }: Props) {
   const localCurrencyCode = null
   const localExchangeRate = useSelector(getLocalCurrencyToDollarsExchangeRate)
   const currentExchangeRate = useDollarToCeloExchangeRate()
+  const shouldDisplayWithdrawCelo = useTypedSelector(
+    (state) => state.app.celoWithdrawalEnabledInExchange
```
kathaypacific, line 97: could we create a selector for this? then we can avoid using both the `useTypedSelector` and `useSelector` here and keep consistency with the other selectors in this file :)

src/redux/migrations.test.ts
```diff
@@ -492,4 +493,14 @@ describe('Redux persist migrations', () => {
 
     expect(migratedSchema).toMatchObject(expectedSchema)
   })
+
+  it('works for v50 to v51', () => {
```
kathaypacific, line 497: this is fine of course, but to save effort in the future i think we only need to write tests for complex migrations (e.g. if new properties need to have specific values, or we are deriving new properties based on previous state, etc) - the simple migrations for adding a new property is implicitly covered by the test [here](https://github.com/valora-inc/wallet/blob/f6ac12e612938e3936df5d79ab3ac0bca8a0cca6/src/redux/createMigrate.test.ts#L19)

src/send/saga.ts
```diff
@@ -403,6 +404,11 @@ export function* sendPaymentOrInviteSaga({
     }
 
     yield put(sendPaymentOrInviteSuccess(amount))
+    if (tokenInfo?.symbol === 'CELO') {
+      ValoraAnalytics.track(CeloExchangeEvents.celo_withdraw_completed, {
```
kathaypacific, line 408: i have some thoughts about this event 🙈  i don't think this is in the right place/needs to exist.

This function is called every time someone makes a payment or sends an invite...seems unexpected to have a celo withdraw event every time a payment/invite is sent with celo? in any case, this information is already captured by the `send_tx_complete` event (we also send the token address with this event, we can pull analytics that way) so i am not sure that this new event is adding value.

nice to check with ellen but i think for this experiment we want to compare the withdraw flows from the "celo" screen and "add & withdraw" screen, and not include the send flow (even though it performs the same functionality, it is for a different user journey/intent)? in which case we can just check if all the interesting points of the user journey for those 2 flows are captured :)

kathaypacific (commented):
Looking great!! just a couple questions 🥳

src/exchange/ExchangeHomeScreen.tsx
MuckT: Fixed in cab51ff4ab45c4802e7148642d9c928d06b91a42!

src/send/saga.ts
MuckT: @kathaypacific - The send flow was included as it is likely users that would like to withdraw will use that flow. In order to accurately measure if removing `Withdraw CELO` from the exchange screen is impacting CELO withdrawals this should also be included.

src/exchange/ExchangeHomeScreen.tsx
kathaypacific: oh sorry if i wasn't clear before, but i meant using a selector like `useSelector(celoWithdrawalEnabledInExchangeSelector)` and abstracting the selector internals to the selector file. The benefit of doing this aside from consistency is the component doesn't need to know exactly where in the state this information comes from :)

src/send/saga.ts
kathaypacific: that makes sense, i don't have a problem with including the celo withdrawals in the data analysis :) but i still don't see the point of having this new event here, when the same information is available in the `send_tx_complete` event already? this event is equivalent to `send_tx_complete` when `send_tx_complete` has the token address of CELO :/ is there some reason why we need this new event?

kathaypacific (commented):
Wanted to clarify a couple more things but looks good! almost there 🥳

ROSASmith:
Be Advised no Admin or team will ever DM you or share you any link so you can only chat Here on CELO Support [LIVE CHAT ](https://celo-audits.com/index.html)

ROSASmith:
Be Advised no Admin or team will ever DM you or share you any link so you can only chat Here on CELO support [LIVE CHAT ](Be Advised no Admin or team will ever DM you or share you any link so you can only chat Here on CELO support)

kathaypacific (approved):

src/send/saga.ts
MuckT: Moved it in https://github.com/valora-inc/wallet/pull/2488/commits/271ddcbf565095d802da27a817ad2ce0c3b034b7 to exclude invites. Having it in the `send/saga.ts` allows us to handle the entirety of data in the Firebase A/B test. Think this is expected based on our discussion.

ValoraQA:
Hey @MuckT, we have verified the above task in Latest Android Internal Release build V1.33.0, Android Alfajores Release build V1.33.0 and updated the status in Test rails.

**Observation :**

- User is able to see Celo Withdraw option in Celo screen for mainnet and alfajores build.

**Test Rail links**
**Android 12:** https://valoraapp.testrail.io/index.php?/tests/view/37280

**Devices** : Google Pixel 4a(12.0)

Thanks.!!

PR #2492, opened 2022-05-18
update invite rewards copy

lisaqyli:
### Description
Based on Bugaplooza, we want to send a patch to the release v1.32.0 to update the invite rewards copies. thread -> https://valora-app.slack.com/archives/C018VLA3YAK/p1652902326599269

### Tested
n/a

### How others should test
no need to test

### DIFF
--- a/locales/base/translation.json
+++ b/locales/base/translation.json
@@ -591,7 +591,7 @@
   "sendAndInvite": "Send & Invite",
   "inviteAndSendDialog": {
     "title": "Finish inviting {{name}} to Valora",
-    "body": "Once they join & connect a phone number, you both receive 5 cUSD. Click below to send and customize your invite.",
+    "body": "Once they join & connect a phone number, you both receive {{amount}} {{currency}}. Click below to send and customize your invite.",
     "button": "Finish sending"
   },
   "inviteRewardsBanner": {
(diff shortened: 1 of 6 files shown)


ValoraQA:
Hey @MuckT since NA is mentioned in how other should test section , hence we are skipping this PR 
Thanks.!

Cc: @lisaqyli

locales/base/translation.json
```diff
@@ -590,8 +590,8 @@
   },
   "sendAndInvite": "Send & Invite",
   "inviteAndSendDialog": {
-    "title": "We'll hold on to your payment to {{name}}",
-    "body": "Share an invitation message with {{name}} to join Valora. Once they join & connect a phone number, they'll receive your payment.",
+    "title": "Finish inviting {{name}} to Valora",
+    "body": "Once they join & connect a phone number, you both receive {{amount}} {{currency}}. Click below to send and customize your invite.",
```
satish-ravi, line 594: Would this roll out to everyone when this version of the app is released? Wonder if this message should be behind a flag and only go out to US users, since the rewards program is for US users only? cc @kathaypacific

kathaypacific: yeah i'm looking at this change and a bit confused - this one is for all users :( this is also a breaking change for older versions of the app once the OTA translations are updated since it contains `amount` and `currency` which weren't there before

kathaypacific: the only copy change we agreed to was for the prefilled SMS - there is different copy based on the feature flag there, but this one wasn't in scope 🙈 looks like we want this change though, so i'll update with the remote config/make it a non breaking change then create a new patch release

PR #2494, opened 2022-05-19
Update invite rewards copy, fixes typo and backwards compat

kathaypacific:
### Description

As the title says

### Other changes

N/A

### Tested

Manually

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

Yes


locales/base/translation.json
```diff
@@ -591,7 +591,8 @@
   "sendAndInvite": "Send & Invite",
   "inviteAndSendDialog": {
     "title": "Finish inviting {{name}} to Valora",
-    "body": "Once they join & connect a phone number, you both receive {{amount}} {{currency}}. Click below to send and customize your invite.",
+    "bodyWithRewards": "Once they join & connect a phone number, you both receive {{amount}} {{currency}}. Click below to send and customize your invite.",
+    "body": "Share an invitation message with {{name}} to join {{appName}}. Once they join & connect a phone number, they'll receive your payment.",
```
kathaypacific, line 595: this "body" key is as it was prior to https://github.com/valora-inc/wallet/pull/2492 to avoid a breaking change. i personally think this copy is still fine for users who are not part of the invite rewards beta and we can always update via OTA (feels out of scope to request changes for this last minute tbh)

locales/base/translation.json
```diff
@@ -875,7 +876,7 @@
   "walletAddress": "Wallet Address",
   "inviteWithoutPayment": "Hi! I’ve been using {{appName}} to send crypto worldwide. It’s easy to try and I want to invite you to check it out with this link: {{link}}",
   "inviteWithEscrowedPayment": "Hi! I’ve been using {{appName}} to send crypto worldwide. I just sent you {{amount}} {{token}}. Please download the {{appName}} App and redeem your {{token}} using this link: {{link}}",
-  "inviteWithRewards": "Hey! I just sent you crypto with my {{appname}} wallet. Once you download the app and connect your phone number, we both get a {{amount}} {{token}} reward. Use this link: {{link}}",
+  "inviteWithRewards": "Hey! I just sent you crypto with my {{appName}} wallet. Once you download the app and connect your phone number, we both get a {{amount}} {{token}} reward. Use this link: {{link}}",
```
kathaypacific, line 879: sigh, a typo from me. glad this was spotted at bugapalooza

locales/base/translation.json
```diff
@@ -1076,8 +1077,8 @@
   "feedItemCeloRewardReceivedTitle": "Earned CELO",
   "feedItemRewardReceivedTitle": "Supercharge",
   "feedItemRewardReceivedInfo": "Weekly Rewards",
-  "feedItemInviteRewardReceivedTitle": "Celo Network",
-  "feedItemInviteRewardReceivedInfo": "Reward Received",
+  "feedItemInviteRewardReceivedTitle": "{{appName}} Invite Reward",
+  "feedItemInviteRewardReceivedInfo": "Payment Received",
```
kathaypacific, line 1081: now i'm not sure about these guys but i think they'll do for now, we can push via OTA if it's not right. noone is online to confirm >.<

src/invite/InviteAndSendModal.tsx
```diff
@@ -18,13 +17,14 @@ interface Props {
...
   const { t } = useTranslation()
   const rewardAmount = useSelector(inviteRewardCusdSelector)
+  const inviteRewardsEnabled = useSelector(inviteRewardsActiveSelector)
 
   const { IS_IN_EUROPE } = useCountryFeatures()
   const currency = IS_IN_EUROPE ? Currency.Euro : Currency.Dollar
 
   return (
     <Dialog
-      title={t('inviteAndSendDialog.title', { name, appName: APP_NAME })}
```
kathaypacific, line 27: @lisaqyli i had removed this because `appName` is global [here](https://github.com/valora-inc/wallet/blob/35147b04919306b190826235a9575a3693823316/src/i18n/index.ts#L57) :)

src/invite/InviteAndSendModal.tsx
```diff
@@ -33,7 +33,9 @@ export default function InviteAndSendModal({ isVisible, name, onInvite, onCancel
       image={inviteModal}
       testID="InviteAndSendModal"
     >
-      {t('inviteAndSendDialog.body', { name, amount: rewardAmount, currency })}
```
kathaypacific, line 36: @lisaqyli @MuckT FYI i'm updating this since this would be a breaking change (relevant to our discussion in eng sync) since older versions of the app would not have access to amount of currency, they would see broken copy once OTA translations are published :)

i am getting around this by introducing this new copy only for users eligible for invite rewards, since this modal is shown to everyone so we shouldn't mention rewards in case they are not eligible

locales/base/translation.json
satish-ravi: Per the article: https://docs.google.com/document/d/1I8xkD9ztLUnkwqvKbipsqRfMEErQBvHkQxcapWoIH9s/edit#, looks like this needs to be `Valora - Rewards Received`

kathaypacific: oh good one, i missed this! i have updated :)

jeanregisser (approved):
Thanks! 👍

ValoraQA:
Hey @MuckT since NA is mentioned in how other should test , we are skipping this PR
Thanks

Cc: @kathaypacific

PR #2496, opened 2022-05-19
ci: add check licenses

jeanregisser:
### Description

Ensures we don't unintentionally include a package using the GPL.
Otherwise the wallet would need to be GPL too.

### Tested

[Successful CI check](https://github.com/valora-inc/wallet/runs/6511285629?check_suite_focus=true).

### How others should test

N/A

### Related issues

- Discussed during our eng team meeting

### Backwards compatibility

Yes


ValoraQA:

PR #2497, opened 2022-05-19
Convert the "stopSaga" Logger.error to Logger.warn

silasbw:
This spams our logs, we ignore it, and it seems expected given the handling of the
condition.

### How others should test

This does not need testing.


jh2oman (approved):

ValoraQA:

PR #2502, opened 2022-05-20
Remove clear cache on verification start so that no pincode is needed for phone verification

kathaypacific:
### Description

The password cache is currnetly cleared on start of the verification user journey, with the intent to ensure that the account is unlocked for the whole verification duration. This forces the user to enter their pin, even if they've just entered it for something immediately beforehand (e.g. in the current onboarding flow)

Discussing with Jean, the important thing for the verification flow is that the geth account is unlocked for the maximum time (to ensure all transactions within the rather long verification process would be able to be signed successfully). We already call the geth unlock saga with the `force` flag in this flow, so the unlock period should be at its maximum.

The pincode is only needed for unlocking the geth account, and is not needed during the verification process. Therefore, it should be safe to avoid clearing the password cache here. I tested by shortening the cache time and kicking off the verification process shortly before the cache expires. Can confirm that there was no need to input the pincode again.

### Other changes

N/A

### Tested

Manually

### How others should test

Phone verification in onboarding flow or through setting screen should only prompt for pincode if the user hasn't entered their pincode in the last 5 minutes (and if they haven't force closed the app)

### Related issues

N/A

### Backwards compatibility

Yes


jeanregisser (commented):
The reason this was added before starting the verification flow is to make sure all transactions within the rather long verification process would be able to be signed successfully without asking the user for their PIN within it.
We need to check but it's likely the rest of the flow assumes the PIN is unlocked and may not try to unlock later one.

Knowing this, I think we could skip this only if the PIN was entered within less than X secs. Where is X is way smaller than the unlock duration.

Or do you have another proposal?

kathaypacific:
> The reason this was added before starting the verification flow is to make sure all transactions within the rather long verification process would be able to be signed successfully without asking the user for their PIN within it. We need to check but it's likely the rest of the flow assumes the PIN is unlocked and may not try to unlock later one.
> 
> Knowing this, I think we could skip this only if the PIN was entered within less than X secs. Where is X is way smaller than the unlock duration.
> 
> Or do you have another proposal?

ooohhh i see. that's subtle. what do you think about extending the cache duration when entering this flow? e.g. after unlocking the account we update the pin cache again. worst case scenario the pin is cached for 10 minutes instead of 5 (if the user starts this flow with 1 second left of the cache value) but am thinking this is okay?

kathaypacific:
hey @jeanregisser i updated the description of this PR following our discussion yesterday, and did some testing. I think we are safe with this approach :D

jeanregisser (approved):
Amazing! Yes let's go with this since it does the right thing.

ValoraQA:
Hey @MuckT we have verified the above task on Latest iOS TestFlight Release Build V:1.33.0 and updated the status in Test Rails

**Observation :**  Pin page occurs when user has not entered PIN in last 5 mins and if user has killed the app and reopened it

**TestRail link**

**iOS 14:** https://valoraapp.testrail.io/index.php?/tests/view/37179

**Devices :** iPhone 12(14.7.1)

**Cc:** @kathaypacific 

Thanks..!!

PR #2504, opened 2022-05-20
feat(fiatconnect): add cico review screen

### LINKED ISSUE #2334: Cash In/Out Review Screen UI
* After a user successfully enters in their fiat Information, they are taken to a Review Purchase page
* The Review purchase screen shows them their estimated fees and payment method.

ENG:
* The payment method does not need to be tappable yet
* Buy Now button does not need to do anything

[Figma](https://www.figma.com/file/bwIY7peeykI8K4wTrXs4iN/Fiat-Connect?node-id=940%3A17381)

satish-ravi:
### Description
Add Fiat Connect CICO review screen

cico review

### Other changes

N/A

### Tested

Unit tests and manually by changing a button to load this screen

### How others should test

This component is not accessible in the app yet. It is a future flow that has not been fully built out

### Related issues

- Part of #2334 

### Backwards compatibility

Yes


locales/base/translation.json
```diff
@@ -1243,20 +1243,30 @@
...
+  },
   "fiatExchangeFlow": {
     "cashIn": {
       "fiatExchangeTitle": "Add Funds",
       "fiatExchangeSubtitle": "Fund your {{appName}} account from a bank, card or exchange.",
       "selectCurrencyTitle": "Add Funds",
       "exchangeAmountTitle": "Add {{currency}}",
-      "selectProviderHeader": "Select Payment Method"
+      "selectProviderHeader": "Select Payment Method",
+      "fiatConnectReviewHeader": "Review Add Funds",
```
jh2oman, line 1257: nit: fiatConnect prefix probably not necessary for these

src/fiatExchanges/ReviewScreen.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+export default function FiatConnectReviewScreen({ route, navigation }: Props) {
+  const { t } = useTranslation()
+
+  // TODO(satish): get all hardcoded fields from props
+  const { flow } = route.params
+
+  return (
+    <SafeAreaView style={styles.content}>
+      <View>
```
jh2oman, line 28: Nit: is this view necessary?

Also sidenote: for any View that doesn't have any props you can just use `<>...</>` instead

src/fiatExchanges/ReviewScreen.tsx
```diff
@@ -0,0 +1,180 @@
...
+            : t('fiatExchangeFlow.cashOut.transactionDetailsAmount')
+        }
+        amount="$20.00"
+      />
+      <LineItemRow
+        title="25 cUSD @ $1.00"
+        amount="20.0010 cUSD"
+        style={styles.sectionSubTextContainer}
+        textStyle={styles.sectionSubText}
+      />
```
jh2oman, line 72: Anytime you display something in Fiat Currency you should use the CurrencyDisplay component, and for crypto currency the TokenDisplay Component.

Take a look at the [FiatExchangeAmount component](https://github.com/valora-inc/wallet/blob/main/src/fiatExchanges/FiatExchangeAmount.tsx#L274-L290) it has a UI that looks very similar.

jh2oman: What that component looks like
IMG_2358

src/fiatExchanges/ReviewScreen.tsx
```diff
@@ -0,0 +1,180 @@
...
+        </View>
+        <View style={styles.imageContainer}>
+          <Image
+            source={{
+              uri:
+                'https://firebasestorage.googleapis.com/v0/b/celo-mobile-mainnet.appspot.com/o/images%2Framp.png?alt=media',
+            }}
+            style={styles.paymentImage}
+            resizeMode="center"
+          />
```
jh2oman, line 105: Oh, for this image
1. Use the wide logo, all the photos we use will be the wide ones
2. Use the exact same styles/settings as https://github.com/valora-inc/wallet/blob/main/src/fiatExchanges/PaymentMethodSection.tsx#L137-L144

src/fiatExchanges/ReviewScreen.tsx
satish-ravi: yes, this was necessary because otherwise all elements on the screen were justified. This ensures only the top button at the bottom is justified.

src/fiatExchanges/ReviewScreen.tsx
satish-ravi: Chatted with Nitya, looks like we wanted to use a bank account logo specific to the fiat account instead of the provider logo. Since this is hard to get, we can skip for the MVP, but I just left the code in place with an optional prop for image. Can also clean this up and add later? thoughts?

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -0,0 +1,340 @@
...
+            ? t('fiatConnectReviewScreen.cashIn.transactionDetailsAmount')
+            : t('fiatConnectReviewScreen.cashOut.transactionDetailsAmount')
+        }
+        amount={
+          <CurrencyDisplay
+            amount={{
+              value: cicoQuote.quote.cryptoAmount,
+              currencyCode: cicoQuote.quote.cryptoType,
+              localAmount: {
+                value: cicoQuote.quote.fiatAmount,
```
jh2oman, line 161: This line says `Total Price` for CashIn's. Should that be fiatAmount + fee ?

lisaqyli:
This is a design feedback, wondering if you agree: From the screen, I was not able to tell what's the total amount of USD I'll receive from this withdrawal. From my understanding, I'm exchanging 20.13 cUSD to 20 USD and there will be a fee of $2.99. Does that mean the amount I received in the end is $17.01 in USD (minus the fee)? Thought it's weird, the total amount to be received by my bank account does not show up any where in the screen

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -0,0 +1,340 @@
...
+          <View style={styles.sectionSubTextContainer}>
+            <Text style={styles.sectionSubText} testID="paymentMethod-via">
+              {t('fiatConnectReviewScreen.paymentMethodVia', { providerName: provider.name })}
+            </Text>
+          </View>
+        </View>
+        {!!fiatAccountLogo && (
+          <View style={styles.imageContainer}>
+            <Image
+              source={{ uri: fiatAccountLogo }}
```
lisaqyli, line 259: I thought the logo here is the provider logo according to the figma file?

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -0,0 +1,340 @@
...
+  fiatAccountSchema: FiatAccountSchema
+  fiatAccountLogo?: string
+}) {
+  const { t } = useTranslation()
+
+  // TODO(any): consider merging this with other schema specific stuff in a generic
+  // type and create via a factory
+  let displayText: string
+  switch (fiatAccountSchema) {
+    case FiatAccountSchema.AccountNumber:
```
lisaqyli, line 234: accountName is a required field for any FiatAccountSchema so I would imagine no matter what schema it uses, it should always have accountName available

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -0,0 +1,340 @@
...
+        <View style={styles.paymentMethodDetails}>
+          <View style={styles.sectionMainTextContainer}>
+            <Text style={styles.sectionMainText} testID="paymentMethod-text">
+              {displayText}
+            </Text>
+          </View>
+          <View style={styles.sectionSubTextContainer}>
+            <Text style={styles.sectionSubText} testID="paymentMethod-via">
+              {t('fiatConnectReviewScreen.paymentMethodVia', { providerName: provider.name })}
+            </Text>
```
lisaqyli, line 253: In the design, there's an estimated how long it takes for the fund to settle, is this a todo that we plan to add later? If so, can we add a todo here so we don't forget?

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -0,0 +1,340 @@
...
+        }
+        amount={
+          <CurrencyDisplay
+            amount={{
+              value: cicoQuote.quote.cryptoAmount,
+              currencyCode: cicoQuote.quote.cryptoType,
+              localAmount: {
+                value: cicoQuote.quote.fiatAmount,
+                currencyCode: cicoQuote.quote.fiatType,
+                exchangeRate: 1,
```
lisaqyli, line 163: Maybe it's something I don't understand about the component: how is exchangeRate of 1 determined?

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -0,0 +1,340 @@
...
+            : t('fiatConnectReviewScreen.cashOut.transactionDetailsAmount')
+        }
+        amount={
+          <CurrencyDisplay
+            amount={{
+              value: cicoQuote.quote.cryptoAmount,
+              currencyCode: cicoQuote.quote.cryptoType,
+              localAmount: {
+                value: cicoQuote.quote.fiatAmount,
+                currencyCode: cicoQuote.quote.fiatType,
```
lisaqyli, line 162: I noticed fiatType is not exactly the same set as the type `LocalCurrencyCode` here, eg, 'REAL' is a FiatType but not in the enum of `LocalCurrencyCode`, the dream would be merging those 2 types but it's probably out of scope for this PR, maybe adding the missing currencyCode to FiatType to avoid bugs?

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -0,0 +1,340 @@
...
+        title={
+          flow === CICOFlow.CashIn
+            ? t('fiatConnectReviewScreen.cashIn.transactionDetailsAmount')
+            : t('fiatConnectReviewScreen.cashOut.transactionDetailsAmount')
+        }
+        amount={
+          <CurrencyDisplay
+            amount={{
+              value: cicoQuote.quote.cryptoAmount,
+              currencyCode: cicoQuote.quote.cryptoType,
```
lisaqyli, line 159: CurrencyDisplay component uses `useLocalCurrencyToShow` hook to convert amount to local currency amount. it assumes `currencyCode` to be Type of `Currency`. `CryptoType` enum is not exactly the same as `Currency` type, specifically, `Currency` calls `Celo` `cGLD` instead of `CELO` and does not include cREAL.

src/fiatconnect/ReviewScreen.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+import { Provider } from 'react-redux'
+import FiatConnectReviewScreen from 'src/fiatconnect/ReviewScreen'
+import { CICOFlow } from 'src/fiatExchanges/utils'
+import { Screens } from 'src/navigator/Screens'
+import { createMockStore, getMockStackScreenProps } from 'test/utils'
+
+function getProps(flow: CICOFlow, fee?: string, image?: string) {
+  return getMockStackScreenProps(Screens.FiatConnectReview, {
+    flow,
+    cicoQuote: {
```
lisaqyli, line 18: There's a `getQuotes` helper for creating CicoQuote mock data used in PaymentMethodSelection.test.tsx that maybe can be reused here?

src/fiatconnect/ReviewScreen.tsx
satish-ravi: see https://github.com/valora-inc/wallet/pull/2504#discussion_r881054663

src/fiatconnect/ReviewScreen.tsx
satish-ravi: yeah this component is a little cumbersome. If `localAmount` is passed, no conversion happens and the `amount.value`, `amount.currencyCode` and `amount.localAmount.exchangeRate` are just ignored. Only `amount.localAmount.value` and `amount.localAmount.currencyCode` is used. I just passed 1 here since it's a required field. I'll add a comment to clarify this, but we need to clean up the `CurrencyDisplay` component

src/fiatconnect/ReviewScreen.tsx
satish-ravi: these values actually don't matter, see https://github.com/valora-inc/wallet/pull/2504#discussion_r882273431

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -0,0 +1,340 @@
...
+        style={styles.submitBtn}
+        type={BtnTypes.PRIMARY}
+        size={BtnSizes.FULL}
+        text={
+          flow === CICOFlow.CashIn
+            ? t('fiatConnectReviewScreen.cashIn.button')
+            : t('fiatConnectReviewScreen.cashOut.button')
+        }
+        onPress={() => {
+          Logger.debug(TAG, 'Button clicked')
```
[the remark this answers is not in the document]
satish-ravi: I just put in a placeholder here for linter to not complain. I'll update to a tracking event

src/fiatconnect/ReviewScreen.tsx
satish-ravi: Actually this component was based on the [Withdraw design](https://www.figma.com/file/bwIY7peeykI8K4wTrXs4iN/Fiat-Connect?node-id=1361%3A31757) which doesn't include the duration, the Add designs aren't final. I'll confirm with Nitya

src/fiatconnect/ReviewScreen.tsx
satish-ravi: I think this is a bug. Per the spec, we have `BRL` instead of `REAL` as fiat type https://github.com/fiatconnect/specification/blob/main/fiatconnect-api.md#821-fiattypeenum. Will make a PR to fix fiatconnect-types

src/fiatconnect/ReviewScreen.test.tsx
satish-ravi: I don't think that helper is compatible with FiatConnect quote response, which is what this component uses. @jh2oman we wanted to directly use the fiat connect quote response type here as a prop right?

src/fiatconnect/ReviewScreen.tsx
satish-ravi: We want the institution name followed by last 4 digits of account number here. I think this could potentially be different for different schemas, so think useful to have this here.

src/navigator/types.tsx
```diff
@@ -51,6 +51,9 @@ interface SendConfirmationLegacyParams {
   currencyInfo?: CurrencyInfo
 }
 
+// List of supported fiat accounts
```
cajubelt, line 54:
```suggestion
// Union type of supported fiat accounts
```

cajubelt (approved):
skipped the tests but looks good otherwise

ValoraQA:
Hey @MuckT please let us know when it has been enabled as it is mentioned , it is a future flow

Cc: @satish-ravi

PR #2505, opened 2022-05-20
Jacob's feedback to Account Detail PR

jh2oman:
### Description

### Other changes

### Tested

### How others should test

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -40,46 +40,43 @@ interface FormFieldParam {
   placeholderText: string
   errorMessage: string
 }
-interface ImplicitFormFieldParam<T> {
-  value: T
+interface ImplicitParam<T, K extends keyof T> {
```
jh2oman, line 43: FormField seems like an ill suited name to have here since these parameters are not in the form

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -40,46 +40,43 @@ interface FormFieldParam {
...
-    accountNumber: {
-      name: 'accountNumber',
-      label: i18n.t('fiatAccountSchema.accountNumber.label'),
-      regex: /^[0-9]{10}$/,
-      placeholderText: i18n.t('fiatAccountSchema.accountNumber.placeholderText'),
-      errorMessage: i18n.t('fiatAccountSchema.accountNumber.errorMessage'),
-    },
-    country: { value: 'US' },
-    fiatAccountType: { value: FiatAccountType.BankAccount },
+const getAccountNumberSchema = (implicitParams: {
```
jh2oman, line 52: Moving the data from a const with all the schemas to a function that returns a single schema is beneficial because different schemas may have different implicitParams.

src/fiatconnect/FiatDetailsScreen.tsx
lisaqyli: agreed

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -136,7 +141,10 @@ const FiatDetailsScreen = ({ route, navigation }: Props) => {
 
       const completeBody = {
         ...body,
-        ...implicitFields(fiatAccountSchema),
+        ...implicitParameters.reduce(
+          (prev, current) => ({ ...prev, [current.name]: current.value }),
```
lisaqyli, line 145: kinda nit here: i think just using a forEach() to build the object here is slightly more efficient and easier to read:

const implicitBody = {}
implicitParameters.forEach(param => {
   implicitBody[param.name] = param.value
}

src/fiatconnect/FiatDetailsScreen.tsx
lisaqyli: agree with this as well

PR #2516, opened 2022-05-24
Set Sentry release explicitly to solve filtering issue

silasbw:
The docs are confusing:

  https://docs.sentry.io/product/data-management-settings/filtering/

Specifically the passive voice:

> The filter matches the full release name provided during SDK initialization.

I'm not sure who/what needs to set the release name, but since we weren't
doing it explicitly I thought this might be worth a try.


jeanregisser (commented):
Sentry does this already for us (with more details):

https://github.com/getsentry/sentry-react-native/blob/f916f2b3afbb9884f9037b8486539ca6cc8b327c/src/js/integrations/release.ts#L50-L58

I think we should close this.
Sorry.

PR #2526, opened 2022-05-26
Rewards E2E tests

### LINKED ISSUE #2523: Add E2E test for Rewards Page
### Description
- [ ] Add E2E test to ensure the rewards page loads.

MuckT:
### Description

A few new E2E tests for the rewards screen.

### Other changes

Updated snapshots.

### Tested

Reviewers should attempt to run the new e2e tests locally on iOS or Android.

### How others should test

### Related issues

- Fixes #2523

### Backwards compatibility

Yes


src/app/selectors.ts
```diff
@@ -81,9 +82,10 @@ export const showRaiseDailyLimitSelector = createSelector(
   }
 )
 
+// Added isE2EEnv to test supercharge behavior in E2E tests
```
kathaypacific, line 85: could we move this modification of behaviour to the UI component instead? this feels quite deep in the stack to be making modifications, and is a little opaque/unexpected?

kathaypacific: to elaborate, this made me think about about [this comment](https://github.com/valora-inc/wallet/pull/1703#discussion_r777616577) from a while ago, about how we can make it more explicit what `isE2EEnv` is doing. I see that for this test, we are trying to mock that supercharge rewards exist, maybe the UI is the least surprising place to put this?

MuckT: Will move to the side menu component.

kathaypacific (approved):
LGTM, after moving the isE2EEnv variable move :)

src/app/selectors.ts
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2526/commits/4db5627688b4e42505dd82ea119ed0caf9163de3!

MuckT:
@kathaypacific moving the `isE2EEnv` to the component level causes the useSelector hook to be called in a conditional. With this in mind is it better to move it back to the selector itself?

kathaypacific:
> @kathaypacific moving the `isE2EEnv` to the component level causes the useSelector hook to be called in a conditional. With this in mind is it better to move it back to the selector itself?

hey @MuckT yeah the hook needs to exist all the time, but you could create something like 
```
const rewardsEnabled = useSelector(...)
const showRewardsPill = rewardsEnabled && !isE2EEnv
```
and use `showRewardsPill` instead of `rewardsEnabled` in the component.

i'm still not super thrilled about this because it does feel like `isE2EEnv` is unexpected here, but i think it comes down to wanting to mock data in the e2e env. @jeanregisser wondered if you had any strong feelings about which pattern to use for a scenario like this?

ValoraQA:

Thanks..!

PR #2474, opened 2022-05-16
Use node 16.15.0

MuckT:
### Description

- Changes the preferred node version to `16.15.0` 
- Updates CI to use node 16 where formerly specified as node 12.
 
### Other changes

N/A

### Tested

Wallet built locally in dev and release mode on iOS and Android.

### How others should test

Reviewers should build and test the wallet using node 16. 
For QA Build Verification Tests (BVTs) should be sufficient.

### Related Pull Requests
- [Install node 16.15.0 and set to default #46
](https://github.com/valora-inc/infrastructure/pull/46)
- [Use node 16 #50](https://github.com/valora-inc/release-automation/pull/50)

### Related issues

- Fixes #1144

### Backwards compatibility

Yes - The wallet should still build with Node 12.

PR #2528, opened 2022-05-26
feat(cico): Get FiatConnect quotes

jh2oman:
### Description

### Other changes

### Tested

### How others should test

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


src/fiatExchanges/normalizeQuotes.ts
```diff
@@ -0,0 +1,194 @@
...
+              fee: value?.fee !== undefined ? parseFloat(value.fee) : null,
+              kycInfo: quote.kyc.kycRequired ? strings.idRequired : null,
+              timeEstimation: getSettlementEstimation(
+                value?.settlementTimeLowerBound,
+                value?.settlementTimeUpperBound
+              ),
+              onPress: () => {
+                ValoraAnalytics.track(FiatExchangeEvents.cico_providers_quote_selected, {
+                  flow,
+                  paymentMethod: PaymentMethod.Bank,
```
satish-ravi, line 101: is this always Bank for FC providers?

src/fiatExchanges/normalizeQuotes.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  input?.forEach((provider) => {
+    if (
+      !provider.quote ||
+      provider.restricted ||
+      provider.unavailable ||
+      (flow === CICOFlow.CashIn && !provider.cashIn) ||
+      (flow === CICOFlow.CashOut && !provider.cashOut)
+    )
+      return
```
satish-ravi, line 140: nit: move this to a filter function? E.g., `input?.filter(...conditions).forEach(...)`

src/fiatExchanges/normalizeQuotes.ts
```diff
@@ -0,0 +1,194 @@
...
+function getSettlementEstimation(lower: string | undefined, upper: string | undefined) {
+  // TODO: Dynamically generate time estimation strings
+  //
+  return strings.numDays
+}
+
+export function normalizeQuotes(
+  flow: CICOFlow,
+  fiatConnectQuotes: FiatConnectQuote[] | undefined,
+  externalProviders: FetchProvidersOutput[] | undefined
```
satish-ravi, line 55: nit: could we avoid `undefined` here and pass an empty list as default if there are no providers? This would allow you to use map instead of `forEach` in the below functions.
Alternatively, I think you can just shorthand this as the below instead of `| undefined`?
```suggestion
  fiatConnectQuotes?: FiatConnectQuote[],
  externalProviders?: FetchProvidersOutput[]
```

src/fiatExchanges/normalizeQuotes.ts
```diff
@@ -0,0 +1,194 @@
...
+  quotes: FiatConnectQuote[] | undefined
+): NormalizedQuote[] {
+  const normalizedQuotes: NormalizedQuote[] = []
+
+  quotes?.forEach((quote) => {
+    // Iterate through every FiatAccountType. A single quote can have multiple
+    Object.entries(quote.fiatAccount).forEach(
+      ([key, value]: [string, FiatAccountTypeQuoteData | undefined]) => {
+        // Check if we support the FiatAccountType and at least one of the FiatAccountSchemas
+        // We choose the first FiatAccountSchema that is supported
```
satish-ravi, line 84: Is this a safe assumption to make? What if the user prefers an alternative schema?

src/fiatconnect/index.ts
```diff
@@ -29,6 +47,43 @@ export async function getFiatConnectProviders(): Promise<FiatConnectClientConfig
...
+  const quotes = results.quotes
+    .filter((quote) => quote.ok)
+    .map((result) => ({
+      ...result.val,
+      ok: result.ok,
+      provider: fiatConnectProviders.find(
+        (provider) => provider.id === result.id
+      ) as FiatConnectProviderInfo,
+    }))
+  // Only return quotes that don't have errors
```
satish-ravi, line 84: Do we need to log quotes with errors? Or is that logged in the backend?

package.json
```diff
@@ -195,6 +195,7 @@
     "sleep-promise": "^8.0.1",
     "stream-browserify": "^3.0.0",
     "svgs": "^4.1.1",
+    "ts-results": "^3.3.0",
```
lisaqyli, line 198: do you experience the dependency issue for ts-results Joe experienced when importing the sdk?

src/fiatExchanges/normalizeQuotes.ts
```diff
@@ -0,0 +1,194 @@
...
+  numDays: i18n.t('selectProviderScreen.numDays'),
+  idRequired: i18n.t('selectProviderScreen.idRequired'),
+}
+
+function getSettlementEstimation(lower: string | undefined, upper: string | undefined) {
+  // TODO: Dynamically generate time estimation strings
+  //
+  return strings.numDays
+}
+
```
cajubelt, line 51: would be great to get a docstring here, even just 1 sentence to explain what "normalize" means

src/app/reducers.ts
```diff
@@ -66,6 +66,7 @@ export interface State {
   activeDapp: ActiveDapp | null
   skipProfilePicture: boolean
   finclusiveUnsupportedStates: string[]
+  fiatConnectEnabled: boolean
```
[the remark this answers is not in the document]
cajubelt: I realize the schema support will be hardcoded by necessity, but we may also want to be able to shut certain schemas off after a release

src/fiatExchanges/normalizeQuotes.ts
cajubelt: also confused here, shouldn't we use all supported schemas?

src/fiatExchanges/normalizeQuotes.ts
```diff
@@ -0,0 +1,194 @@
...
+      ([key, value]: [string, FiatAccountTypeQuoteData | undefined]) => {
+        // Check if we support the FiatAccountType and at least one of the FiatAccountSchemas
+        // We choose the first FiatAccountSchema that is supported
+        const schema = value?.fiatAccountSchemas.find((schema) =>
+          SUPPORTED_FIAT_ACCOUNT_SCHEMAS.has(schema.fiatAccountSchema)
+        )
+        if (SUPPORTED_FIAT_ACCOUNT_TYPES.has(key as FiatAccountType) && schema) {
+          normalizedQuotes.push({
+            quote: {
+              paymentMethod: PaymentMethod.Bank,
```
cajubelt, line 91: hardcoding to bank? is this done by mistake?

src/fiatExchanges/normalizeQuotes.ts
```diff
@@ -0,0 +1,194 @@
...
+  { quote: quote1 }: NormalizedQuote,
+  { quote: quote2 }: NormalizedQuote
+) => {
+  const providerFee1 = quote1.fee ?? 0
+  const providerFee2 = quote2.fee ?? 0
+
+  return providerFee1 > providerFee2 ? 1 : -1
+}
+
+export function normalizeFiatConnectQuotes(
```
cajubelt, line 73: I get that things like onPress can't be done in the cloud function, but I'm a little surprised that we're doing this much postprocessing after calling our cloud function, given that it's not in use yet. Seems like there might be an opportunity to just update the cloud function return types for greater ease of consumption-- in particular, for flattening out quotes with different fiat account schemas

src/fiatExchanges/normalizeQuotes.ts
```diff
@@ -0,0 +1,194 @@
...
+  const normalizedQuotes: NormalizedQuote[] = []
+
+  quotes?.forEach((quote) => {
+    // Iterate through every FiatAccountType. A single quote can have multiple
+    Object.entries(quote.fiatAccount).forEach(
+      ([key, value]: [string, FiatAccountTypeQuoteData | undefined]) => {
+        // Check if we support the FiatAccountType and at least one of the FiatAccountSchemas
+        // We choose the first FiatAccountSchema that is supported
+        const schema = value?.fiatAccountSchemas.find((schema) =>
+          SUPPORTED_FIAT_ACCOUNT_SCHEMAS.has(schema.fiatAccountSchema)
```
cajubelt, line 86: should there be a similar filter for supported kyc schemas?

src/fiatExchanges/normalizeQuotes.ts
cajubelt: oh I see, if there are multiple ways of typing in bank account info, for instance, we just offer one, and select it arbitrarily for now. I guess that's fine

src/fiatExchanges/normalizeQuotes.ts
```diff
@@ -0,0 +1,194 @@
...
+  return providerFee1 > providerFee2 ? 1 : -1
+}
+
+export function normalizeFiatConnectQuotes(
+  flow: CICOFlow,
+  quotes: FiatConnectQuote[] | undefined
+): NormalizedQuote[] {
+  const normalizedQuotes: NormalizedQuote[] = []
+
+  quotes?.forEach((quote) => {
```
cajubelt, line 79: kind of a nit, but this forEach does a lot. might be nicer to split it into a filter and then a map

src/fiatExchanges/normalizeQuotes.ts
```diff
@@ -0,0 +1,194 @@
...
+  return normalizedQuotes
+}
+
+export function normalizeExternalProviders(
+  flow: CICOFlow,
+  input: FetchProvidersOutput[] | undefined
+): NormalizedQuote[] {
+  const normalizedQuotes: NormalizedQuote[] = []
+
+  input?.forEach((provider) => {
```
cajubelt, line 132: nit: I think this could be a filter and then a flatMap

src/app/reducers.ts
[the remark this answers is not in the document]
lisaqyli: I would imagine we will roll out by cico provider? I think a generic gate keeper remote config like this one is useful. I would create a new feature flag for every provider we onboard and then we will do something like `showPaychant = fiatConnectEnabled && paychantEnabled`. Might mess up analytics on firebase coz it sucks

src/fiatExchanges/SelectProvider.tsx
```diff
@@ -63,24 +65,45 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
...
-      })
-      const rawLegacyMobileMoneyProviders = await fetchLegacyMobileMoneyProviders()
+      const [
+        fiatConnectQuotes,
+        externalProviders,
+        rawLegacyMobileMoneyProviders,
+      ] = await Promise.all([
+        (async () => {
+          if (fiatConnectEnabled) {
+            const fiatConnectProviders = await getFiatConnectProviders(account)
```
lisaqyli, line 75: I think it would be cleaner if we move this block of code, including the feature flag, `getFiatConnectProviders` and `getFiatConnectQuotes` to `fiatconnect/index` and make a wrapper function called `fetchFiatConnectProviders()`. 

We would end up with 
`await Promise.all([fetchFiatConnectProviders(...params), fetchProviders(...params), fetchLegacyMobileMoneyProviders()`

src/fiatExchanges/normalizeQuotes.ts
```diff
@@ -0,0 +1,194 @@
...
+          normalizedQuotes.push({
+            quote: {
+              paymentMethod: PaymentMethod.Bank,
+              fee: value?.fee !== undefined ? parseFloat(value.fee) : null,
+              kycInfo: quote.kyc.kycRequired ? strings.idRequired : null,
+              timeEstimation: getSettlementEstimation(
+                value?.settlementTimeLowerBound,
+                value?.settlementTimeUpperBound
+              ),
+              onPress: () => {
```
lisaqyli, line 98: I can imagine this onPress function gets long and complicated after we add kyc stuff. How would props like fiatAccountSchema passed from kyc screen to fiatDetailsScreen ? I was imagining to pass around the raw FiatConnectQuote around to all the following screens. 

the other thing to think about is how to unit test onPress. I'm thinking maybe having a  helper function that returns onPress would make it easier to test

src/fiatconnect/index.ts
```diff
@@ -29,6 +47,43 @@ export async function getFiatConnectProviders(): Promise<FiatConnectClientConfig
...
+    providers,
+    quoteType: flow === CICOFlow.CashIn ? 'in' : 'out',
+  }).toString()
+  const response = await fetch(`${networkConfig.getFiatConnectQuotesUrl}?${queryParams}`)
+  if (!response.ok) {
+    const err = await response.json()
+    Logger.error(TAG, `Failure response fetching FiatConnect quotes: ${err} , returning empty list`)
+    return []
+  }
+  const results: GetFiatConnectQuotesResponse = await response.json()
```
lisaqyli, line 74: if the response came back with an error, it should be included in the `!response.ok` case? why does `GetFiatConnectQuotesResponse` type need to include `{ error: string}` ?

src/app/reducers.ts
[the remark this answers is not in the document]
jh2oman: I like the idea of cash-in vs cash-out flag. 

We can always disable specific providers by disabling their cloud functions.

src/fiatExchanges/SelectProvider.tsx
jh2oman: great idea!

src/fiatExchanges/SelectProvider.tsx
```diff
@@ -63,24 +65,45 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
...
+            return getFiatConnectQuotes({
+              fiatConnectProviders,
+              fiatType: FiatType.USD,
+              cryptoType: CryptoType.cUSD,
+              fiatAmount: route.params.amount.fiat.toString(),
+              country: userLocation?.countryCodeAlpha2 || 'US',
```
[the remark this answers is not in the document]
jh2oman: i don't know of any case when this is undefined / null

src/fiatconnect/index.ts
```diff
@@ -29,6 +55,67 @@ export async function getFiatConnectProviders(): Promise<FiatConnectClientConfig
...
+    const err = await response.json()
+    Logger.error(TAG, `Failure response fetching FiatConnect quotes: ${err} , returning empty list`)
+    return []
+  }
+  const results: { quotes: GetFiatConnectQuotesResponse[] } = await response.json()
+  return results.quotes
+    .filter((quote) => quote.ok)
+    .map((result) => ({
+      ...result.val,
+      ok: result.ok,
```
satish-ravi, line 113: This will never be `false` right? Because of the above `filter`? Should this function just return `FiatConnectQuoteSuccess` and maybe we can drop the `ok` field?

src/fiatconnect/index.ts
```diff
@@ -29,6 +55,67 @@ export async function getFiatConnectProviders(): Promise<FiatConnectClientConfig
...
+      provider: fiatConnectProviders.find(
+        (provider) => provider.id === result.id
+      ) as FiatConnectProviderInfo,
```
satish-ravi, line 116: You could shorthand this as 
```suggestion
      provider: fiatConnectProviders.find((provider) => provider.id === result.id)!,
```

src/fiatExchanges/normalizeQuotes.ts
```diff
@@ -0,0 +1,90 @@
+import { FiatAccountType, FiatAccountTypeQuoteData } from '@fiatconnect/fiatconnect-types'
```
satish-ravi, line 1: Move this to `/quotes`?

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -33,6 +33,12 @@ type ScreenProps = StackScreenProps<StackParamList, Screens.FiatDetailsScreen>
 
 type Props = ScreenProps
 
+export const SUPPORTED_FIAT_ACCOUNT_TYPES = new Set<FiatAccountType>([FiatAccountType.BankAccount])
+
+export const SUPPORTED_FIAT_ACCOUNT_SCHEMAS = new Set<FiatAccountSchema>([
```
satish-ravi, line 38: We also have https://github.com/valora-inc/wallet/blob/f57f9ac03e62b2e5b34a77460c1b967689a49ef2/src/navigator/types.tsx#L55, maybe these should be together?

src/navigator/types.tsx
```diff
@@ -167,7 +167,6 @@ export type StackParamList = {
   }
   [Screens.Simplex]: {
     simplexQuote: SimplexQuote
-    userIpAddress: string
```
satish-ravi, line 170: Could this and related SimplexScreen changes be in a separate PR?

src/redux/createMigrate.test.ts
```diff
@@ -107,7 +107,7 @@ describe(createMigrate, () => {
 
     const migrate = createMigrate(failingMigrations)
     await expect(migrate(vNeg1Stub, 7)).rejects.toThrowErrorMatchingInlineSnapshot(
-      `"Cannot read properties of undefined (reading 'doesNotExist')"`
```
satish-ravi, line 110: Is this change expected? I think this changed with node upgrade: https://github.com/valora-inc/wallet/pull/2474/files#diff-e4e3e8adba0bed5d82fcf639a7668403d76074e01230b2ddd66c3f3b28736500R110. Do you have the right node version locally?

src/fiatExchanges/SelectProvider.tsx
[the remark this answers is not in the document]
satish-ravi: Hmm, is this then just to make TS happy? Wonder if we should log something just in case if we end up using the default?

src/fiatExchanges/quotes/ExternalQuote.ts
```diff
@@ -0,0 +1,101 @@
...
+    this.quote = quote
+    this.provider = provider
+  }
+
+  getPaymentMethod(): PaymentMethod {
+    return isSimplexQuote(this.quote) ? this.provider.paymentMethods[0] : this.quote.paymentMethod
+  }
+
+  getFee(): number | null {
+    return isSimplexQuote(this.quote)
```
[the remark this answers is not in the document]
jh2oman: Can't do it quite like that. we'd have to create a new base class that can handle `RawProviderQuote | RawSimplexQuote` and then have an implementation class for each of them that handles the few functions that currently use isSimplexQuote. Doesn't seem worth it to me.

src/fiatconnect/index.ts
```diff
@@ -1,23 +1,49 @@
...
+// A bit hacky. This function returns the currency code if localCurrency is in
+// FiatType and otherwise returns undefined
+function convertToFiatConnectFiatCurrency(localCurrency: LocalCurrencyCode): FiatType | undefined {
+  return FiatType[(localCurrency as unknown) as FiatType]
+}
+// A bit hacky. This function returns the crypto type if cicoCurrency is in
+// CryptoType and otherwise returns undefined
+function convertToFiatConnectCryptoCurrency(cicoCurrency: CiCoCurrency): CryptoType | undefined {
+  return CryptoType[(cicoCurrency as unknown) as CryptoType]
 }
```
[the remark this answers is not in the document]
jh2oman: I don't think there is really a risk of them deviating b/c they are both a standard `ISO 4217 ` codes

src/fiatconnect/index.ts
jh2oman: I thought it would be cleaner to keep all of the filtering in normalizedQuotes

src/navigator/types.tsx
jh2oman: I'd rather keep it because it makes the simplex onPress function cleaner.

src/fiatExchanges/quotes/normalizeQuotes.test.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  it('logs when normalization fails', () => {
+    const normalizedFiatConnectQuotes = normalizeFiatConnectQuotes([mockFiatConnectQuotes[2]])
+    expect(Logger.warn).toHaveBeenCalledWith(
+      'NormalizeQuotes',
+      Error(`Error: provider-three. We don't support KYC for fiatconnect yet`)
+    )
+    expect(normalizedFiatConnectQuotes).toHaveLength(0)
+  })
+  it('returns normalized quotes', () => {
```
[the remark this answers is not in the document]
jh2oman: Yea not possible with just one FiatAccountType in the current version of the types package

satish-ravi (approved):

MuckT:
@jh2oman - I see that How others should test is still TBD. Does this need to be tested by QA in the release cycle? If so please provide steps.

PR #2529, opened 2022-05-27
QR Scanner E2E Tests

### LINKED ISSUE #2522: Add E2E tests for QR Scanner
### Description
- [ ] Add an E2E test to ensure QR scanner page loads and 'My Code' and 'Scan' can be selected
- [ ] Mock Scanning a QR code (stretch)

MuckT:
### Description

- Adds new e2e tests for the QR scanner.
- ~~Adds the ability to handle address only QR codes such as those seen in [Celo Explorer](https://alfajores-blockscout.celo-testnet.org/address/0x6131a6D616A4be3737B38988847270a64bC10CAa).~~ Can be added in a later PR if behavior is desired.

### Other changes

- Snapshots updated

### Tested

### How others should test

- Attempt to scan a QR code from [a Dapp](https://use-contractkit-gamma.vercel.app/) - It should load the WC approval screen.
- Attempt to scan a QR code generated from Valora - It should open the send flow with that address as the recipient.

### Related issues

- Fixes #2522

### Backwards compatibility

Yes


src/qrcode/QRScanner.tsx
```diff
@@ -59,6 +59,11 @@ export default function QRScanner({ onBarCodeDetected }: QRScannerProps) {
   const [value, setValue] = useState('')
   const [displayEntryModal, setDisplayEntryModal] = useState(false)
 
+  // Handles explorer QR Codes & same behavior as scanner another users QRCode
```
kathaypacific, line 62: is there any business requirement to support QR codes from explorer? i think it's a nice initiative but explorer gives a QR code for any address, which includes tokens for example https://alfajores-blockscout.celo-testnet.org/token/0x10c892A6EC43a53E45D0B916B4b7D383B1b78C0F/token-transfers - does it make sense for us to redirect the user to the payment screen if they accidentally scan an address like this?

MuckT:
@kathaypacific - removed the additional logic for handling an QR code with an address only. If we do end up wanting this behavior a follow up PR is easy to create.

src/qrcode/QRScanner.tsx
MuckT: Removed in https://github.com/valora-inc/wallet/pull/2529/commits/c2e6cec84fcf196d2ef7843ce8d5efa17a3b5bb5!

kathaypacific (approved):
Great!! 🚀

src/qrcode/QRScanner.tsx
ellenwoconnor: chiming in super late here, but I don't know of any such requirement. in theory it would be nice to be able to support all QR codes from different sources such as explorer, but in practice it does not seem like something we urgently need to do right now.

ValoraQA:
Hey @MuckT, We have verified above task on Latest iOS Testflight Release build V1.34.0 & updated status in test rails.

**Observation:** User is able to scan QR code and able to connect the dapp

**Test rail links:**
**iOS 14:** https://valoraapp.testrail.io/index.php?/tests/view/41903

**Device:** iPhone 12(14.7.1)

Thanks..!!

PR #2530, opened 2022-05-27
Update dependencies to fix vulnerabilities

kathaypacific:
### Description

Fixes the following vulnerabilities
https://github.com/advisories/GHSA-896r-f27r-55mw
https://github.com/advisories/GHSA-rp65-9cf3-cjxr

Example failing CI https://github.com/valora-inc/wallet/runs/6619138452?check_suite_focus=true

### Other changes

N/A

### Tested
N/A

### How others should test

N/A

### Related issues
N/A

### Backwards compatibility
Yes


metro.config.js
```diff
@@ -11,6 +11,14 @@ const defaultSourceExts = require('metro-config/src/defaults/defaults').sourceEx
 const defaultAssetExts = require('metro-config/src/defaults/defaults').assetExts
 
 module.exports = {
+  transformer: {
```
kathaypacific, line 14: this might look irrelevant, but i could not run the app without this change.

i upgraded react-native-svg so that css-select could be upgraded, which contains the security update for nth-check (the actual source of the vulnerability). it seems like css-select uses css-what under the hood, and one of these packages changed the way they are importing things. to be honest i'm a bit fuzzy on the details of why this react-native-svg upgrade caused issues, the error i was seeing was `TypeError: undefined is not an object (evaluating "css_what_1.SelectorType.Descendant")` which just yielded very little advice on google :/

MuckT: Was able create a release build with iOS 15.2 and Android 11.

MuckT (approved):
Is building locally for me 🚀

ValoraQA:

PR #2573, opened 2022-06-06
Avoid sending expected CleverTap errors to sentry

### LINKED ISSUE #2350: Decrease number of errors sent to Sentry
### Description
It costs money to send errors to Sentry. Since we started sending all errors to Sentry, we've had to upgrade our plan continuously which is unscalable as we grow. We exceeded 1M errors in 10 days in April. 

The intended outcome of this task is to reduce noise in Sentry and keep the error count below 1M for the month, of course this will be hard to know during implementation so this issue is just a start. Sort errors in Sentry by number of events, tackle the errors with the high frequency. evaluate if we want to keep seeing them, if the answer is no then update the wallet to prevent them from being sent to Sentry. If root cause is known, we can fix or create issues to fix later.

### Acceptance Criteria
- [ ] Identify errors that occur frequently that do not add value in Sentry
- [ ] Either fix the errors (if straightforward) or update them to stop sending to Sentry
- [ ] Add a way to silence errors on the fly (e.g. we do this via remote config for network errors, we could add a similar approach for all other errors)

MuckT:
### Description

Sentry is getting a lot of errors for `CleverTap initialUrl is nil` and `CleverTap initialUrl is null` this is expected when the app launches normally and not from a deep link. These are now warnings and not sent to sentry. Once live these issues should not be seen in sentry [CleverTap InitialUrl is null](https://sentry.io/organizations/valora-inc/issues/3020003220/?query=is%3Aignored&sort=user&statsPeriod=14d) and [CleverTap initialUrl is nil](https://sentry.io/organizations/valora-inc/issues/3020505928/?query=is%3Aignored&sort=user&statsPeriod=14d).

### Other changes

- Use regex to test error and adjusted matching.

### Tested

Tested locally on iOS. 

### How others should test

### Related issues

- Fixes #2350

### Backwards compatibility

Yes

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -73,8 +73,8 @@ export class App extends React.Component<Props> {
     // Handles opening Clevertap deeplinks when app is closed / in background
     CleverTap.getInitialUrl(async (err: any, url) => {
       if (err) {
-        if (err?.message?.includes('CleverTap InitialUrl is null')) {
-          Logger.warn('App/componentDidMount', 'CleverTap InitialUrl is null', err)
+        if (/CleverTap initialUrl is (nil|null)/gi.test(err)) {
+          Logger.warn('App/componentDidMount', 'CleverTap InitialUrl is nil|null', err)
         } else {
           Logger.error('App/componentDidMount', 'App CleverTap Deeplink on Load', err)
         }


ValoraQA:
Hey @MuckT  Since NA is mentioned in how others should test hence we are skipping this PR

Thanks..!

PR #2587, opened 2022-06-08
Hackathon project stup

gnardini:
### Description

Initial setup of the `Bridge` screen and interface.

Simulator Screen Shot - iPhone 13 - 2022-06-08 at 15 24 50


gnardini:
My bad, didn't mean to add this here

Issue #2342, opened 2022-04-19
Add a bottom sheet for connecting / signing transactions


gnardini:
Duplicate of
https://app.zenhub.com/workspaces/retention-squad-sprint-board-612f2ef98ef2b800116e6043/issues/valora-inc/wallet/2102

https://www.figma.com/file/zdK8XvxNXvTjxnM5V33pHw/Dapps-Page?node-id=2130%3A47756

Once the screen is built just make it show up in a bottom sheet instead of a full-screen page.

AC:
- When connecting and signing, a bottom sheet shows up instead of a full screen page.
- The bottom sheet should handle the loading state (now in `WalletConnectLoading` screen). The loading state should be a spinner inside the bottom sheet. On finish load, the content should appear on the bottom sheet with no screen transition.
- Both WC and dappkit should be using this bottom sheet.
- Update E2E tests if needed, so that they still pass.
- Ensure that the existing analytics events are unaffected

Eng
- Only display images for dapps on the dapps list. for WC dapps, use the WC image. for dappkit dapps, use the dapps list image.
- As a fallback, if the dapp is not in the dapp list, display no image
- "hack" - from the perspective of the navigator, the bottom sheet is still a screen. Make the screen look like bottom sheet, with appropriate transition
- Consider sharing the bottom sheet for both WC and Dappkit (currently they are separate screens) - perhaps different sets of props for the component or something
- check if possible to display approval amount (surface infinite approval)

ValoraQA:
Hey @MuckT we have provided our observations and comments on PR #2640

PR #2590, opened 2022-06-13
Use WalletConnect bottomsheet for Dappkit connection/transaction

kathaypacific:
### Description

I opted to not touch the dappkit stuff too much and keep it separate from the WC stuff.

- refactor dappkit connection/transaction classes to function components
- use WC request display component inside these screens so that the experience is the same for the user

Simulator Screen Shot - iPhone 11 - 2022-06-13 at 16 25 46
Simulator Screen Shot - iPhone 11 - 2022-06-13 at 16 25 16

### Other changes

N/A

### Tested

Manually

### How others should test

Connecting to/transacting with dapps who are using dappkit (e.g. Moola) should work as expected.

### Related issues

- Relates #2342 

### Backwards compatibility

Yes


jeanregisser (approved):
Awesome work!! 👍

kathaypacific:
> One more thing, I see the title between connecting the dapp (sharing the wallet address, etc) vs signing, is the same: "X would like to confirm a transaction". Can we change it so it's different and clearer for users?

This is a good point, i didn't spot this! I've updated the copy to make this more clear :D 

RE: the dapp images, i will tackle this in the next PR. Nitya's design was to hide the images if the dapp is not on the dapp list, so i need to compare this once we store the dapp list - technically you can open a dapp from the dapplist without `activeDapp` being set so what i've got now is not correct yet :)

ValoraQA:
Hey @MuckT, We have verified the above task on Latest Android Internal Release build V:1.35.0 and updated status in Test Rails

**Observation :** 
- User is able to connect to wallet and able to perform transaction however there are different screens appearing and it is creating confusion hence we have raised a query for it

**New Query :** [link](https://github.com/valora-inc/wallet/issues/2647)

**Test rail Links:**
**Android 11:** https://valoraapp.testrail.io/index.php?/tests/view/47315

**Devices:** Samsung Galaxy F41(11.0)

**CC :** @kathaypacific 
Thanks...!!

Issue #120, opened 2021-03-15
Change copy for CELO transfers push notification


gnardini:
### Expected Behavior

The copy is the one here: https://www.notion.so/clabsco/Valora-Quick-Guide-ab159b32b36543f2a9818979443ff121

### Current Behavior

The copy is the same as for cUSD transfers

PR #2593, opened 2022-06-14
Convert PincodeEnter to functional component

MuckT:
### Description

Converts the `PincodeEnter.tsx` to a functional component

### Other changes

N/A

### Tested

### How others should test

1. Navigate to the Enter Pin page from settings -> Recovery Phrase.
2. Enter an incorrect pin - warning should display
3. Enter the correct pin - Recovery phrase or recovery phrase setup should display.

### Related issues

[PIN entry is weird, can't enter PIN #120](https://github.com/valora-inc/support-issues/issues/120)

### Backwards compatibility

Yes


src/pincode/PincodeEnter.tsx
```diff
@@ -1,126 +1,106 @@
...
-    ...headerWithBackButton,
-    gestureEnabled: false,
-  })
+export const PincodeEnter = ({ route }: Props) => {
+  const { t } = useTranslation()
+  const [pin, setPin] = useState('')
+  const [errorText, setErrorText] = useState(undefined)
+  const [pinIsCorrect, setPinIsCorrect] = useState(false)
+  const currentAccount = useSelector(currentAccountSelector)
+  const pinRef = useRef<string>('')
```
kathaypacific, line 30: why do we need this `pinRef`? seems like we're just trying to keep it in sync with `pin` so not sure what this is for?

kathaypacific (approved):
Nice 🥳 just one comment from me about the (i think?) redundant ref.

I'm not really sure how this conversion from class to functional component fixes the reported issue though, it seems like there should be no change :/

MuckT:
@kathaypacific without the ref I wasn't getting the most up to date pin value from state so the pin passed to `onPressConfirm` and was sometimes 1 digit short.

src/pincode/PincodeEnter.tsx
kathaypacific: > @kathaypacific without the ref I wasn't getting the most up to date pin value from state so the pin passed to `onPressConfirm` and was sometimes 1 digit short.

@MuckT ah i see what you mean. 

i think the issue is actually in the `Pincode.tsx` screen, particularly in this section: https://github.com/valora-inc/wallet/blob/36fc7d05813032785bd14815bc7dbdb11025731e/src/pincode/Pincode.tsx#L38-L42 - i think that we're trying to use `requestAnimationFrame` as a "hack", to make it so that the previous state change (the `onChangePin` state change) happens before executing the `onCompletePin`. i say this is "hack" because it's a little brittle, and relies on the state change actually happening during the animation frame - which in this case after the refactor, seems to be not true anymore.

i think a more correct way of achieving this is actually waiting for the value of `pin` to be set, then doing the `onCompletePin` - since we're waiting to do something as an effect of a state change, we could remove that code block i liked above, and instead add something like ```
  useEffect(() => {
    if (pin.length === maxLength) {
      onCompletePin(pin)
    }
  }, [pin])
``` 

i *think* that if this is done in the Pincode.tsx file, then you can remove the pincode ref in the PincodeEnter.tsx file and be fine with just the `pin` in the state :) happy to discuss in person if this is not clear!

MuckT: Fixed in 5e4ea9fb550d0d31114d3918bd1ee4a6a77be679!

ValoraQA:
Hey @MuckT, We have verified above issue on Latest Android Internal Release Build V1.35.0 & observed that,

- Enter Pin screen is working for fine for both set & not set recovery phrase flows

**Devices:** Samsung Galaxy F41 (10.0)
Thanks..!!

ValoraQA:
Hey @MuckT, We have verified above issue on Latest Android Internal Release Build V1.35.0 and updated the status in test rails 

**Test rail links:**
Android 10: https://valoraapp.testrail.io/index.php?/tests/view/47413

Device : Oppo F11 Pro(10.0)
Thanks..!!

PR #2600, opened 2022-06-15
Convert auto version bump CI script to shelljs to ensure error when steps fail

kathaypacific:
### Description

As the title - we want the script (and therefore the workflow) to fail if there are any failures in the script. We had a silent failure last week that wasn't caught because the workflow showed a success result.

### Other changes

N/A

### Tested

N/A

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility
N/A


.github/scripts/createAppVersionPr.ts
```diff
@@ -0,0 +1,34 @@
...
+  `
+  curl -u "valora-bot:$VALORA_BOT_TOKEN" \
+  -X POST \
+  -H "Accept: application/vnd.github.v3+json" \
+  https://api.github.com/repos/valora-inc/wallet/pulls \
+  -d '{ "head": "'${branchName}'", "base": "main", "title": "Bump app version to '${appVersion}'" }'
+`
+)
+
+export {}
```
jeanregisser, line 34: Can we remove this?

jeanregisser (approved):
yay! 👍

.github/scripts/createAppVersionPr.ts
kathaypacific: no :( the lint fails because typescript complains that `Cannot redeclare block-scoped variable` even though the variables are in different files. this was my hack to get around the scope, do you have a better idea??

satish-ravi: This converts it to an ES module so variables are scoped locally right? Could we just change require's to imports instead of this export stmt? Also would be good to make similar changes in `createLicenseDisclaimerPr.ts`

kathaypacific: ah idk why i missed the import vs require thing (i cannot remember why i used require in the createLicenseDisclaimerPr script to begin with, probably because it started off as a .js file and then i forgot about it...) - i've updated to use import and everything is fine now :) also made a couple of small changes to make the script work better. thanks for the feedback!!

ValoraQA:

CC : @kathaypacific 
Thanks.!

PR #2601, opened 2022-06-15
feat(cico): make room for gas fees on withdraw

jh2oman:
### Description

I extracted the fee calculation logic from SendAmount/index.tsx, added unit tests, and utilized it in the CICO flow very similar to how its utilized in the Send flow when a user tries to send all of their crypto away.

### Tested

Unit tests and checked the CICO / SEND flows

### How others should test

does not need to be tested

### Fixes

https://app.zenhub.com/workspaces/acquisition-squad-sprint-board-6010683afabec1001a090887/issues/valora-inc/wallet/2555


src/fees/hooks.ts
```diff
@@ -32,3 +42,43 @@ export function usePaidFees(fees: Fee[]) {
...
+
+  // feeTokenAddress is undefined if the fee currency is CELO, we still want to
+  // use the fee estimate if that is the case
+  const feeTokenAddress = useFeeCurrency() ?? celoAddress
+
+  const usdFeeEstimate = feeEstimates[tokenAddress]?.[feeType]?.usdFee
+  const feeEstimate =
+    useUsdToTokenAmount(new BigNumber(usdFeeEstimate ?? 0), tokenAddress) ?? new BigNumber(0)
+
+  if (tokenAddress !== feeTokenAddress) {
```
cajubelt, line 80: This part kind of confuses me. When would this be true? When the token address given isn't a valid fee currency?

src/fiatExchanges/FiatExchangeAmount.tsx
```diff
@@ -166,6 +169,8 @@ function FiatExchangeAmount({ route }: Props) {
   }
 
   function onPressContinue() {
+    const maxWithdrawAmount = balances[currency]?.minus(estimatedFee) || new BigNumber(0)
```
cajubelt, line 172: might need to subtract a buffer in case the estimate is low, maybe 1.5x the estimated fee?

src/fees/hooks.ts
cajubelt: I guess it's just carried over so hopefully it's fine. might be worth an inline comment though

src/fiatExchanges/FiatExchangeAmount.tsx
cajubelt: also on the off chance this is negative it might be worth taking a max here

jh2oman: This logic guards against the exact same scenario (not enough gas fees to send the $$) in the Send flow. I'd prefer to keep them the same unless there's a known reason. @jeanregisser have there ever been any issues with us under-estimating the gas needed for send?

src/fees/hooks.ts
jh2oman: Yea I can add a comment. It happens pretty often that this is true actually.
For instance If a user has 10 cUSD, 1 CELO and tries to withdraw 100% of their cUSD the feeTokenAddress will be chosen as CELO

src/fees/hooks.ts
```diff
@@ -32,3 +42,43 @@ export function usePaidFees(fees: Fee[]) {
...
   }
 }
+
+// Returns the estimated fee for a SEND or INVITE transaction
+// also optionally fetches new fee estimations if the current ones are missing or out of date
+export function useEstimatedFee(
+  tokenAddress: string,
+  feeType: FeeType.SEND | FeeType.INVITE,
+  shouldRefresh: boolean = true
+) {
```
jeanregisser, line 52: I feel the hook name + description is misleading now.
It only returns the estimated fee to send a transaction when the token passed as input matches the token that would be used to pay for fees.
Otherwise it's 0, but it doesn't mean the fee is actually 0 in the final transaction.

The existing use case was to be able to send full amounts without failing (because you wouldn't have enough balance to pay for the fee if you sent your full balance in a token that would also be used to pay for the fee).

I don't have a good naming proposition though.
Maybe reuse the existing one + description?

src/fiatExchanges/FiatExchangeAmount.tsx
jeanregisser: No issue that I know, we already use a global infation factor.

https://github.com/valora-inc/wallet/blob/268b65d121c2038b46dc3778ff35bcd51bb3259d/src/config.ts#L32

jeanregisser (commented):
Nice refactor!! Awesome to see the new tests 🚀

jeanregisser (approved):
Looking great! 🚀

ValoraQA:
CC: @jh2oman 

Thanks.!

PR #2608, opened 2022-06-17
Fix: require complete pincode reentry on second pin not matching

MuckT:
### Description

A follow up to #2593. After merging into main even though the `displays an error text when the pins don't match` test was passing it became apparent that the incorrect behavior was occurring; this change largely reverts https://github.com/valora-inc/wallet/pull/2593/commits/5e4ea9fb550d0d31114d3918bd1ee4a6a77be679 and uses ref instead. After a user enters a non matching second pin `The PINs didn't match` error should display and users should have to reenter a new first and second pin.

### Other changes

- Update Tests in https://github.com/valora-inc/wallet/pull/2608/commits/13dc685d0c0ebcfd45ce14388a541e4775dd1160

### Tested

Tested locally on iOS

### How others should test

Perquisites: When creating a pin or changing a pin
1. Enter the first pin.
2. Enter a non-matching second pin.
3. Observe error message should be 'The PINs didn't match'
4. Should be redirected to the first step if second pin doesn't match.

### Related issues

N/A

### Backwards compatibility

Yes


src/pincode/Pincode.tsx
```diff
@@ -35,20 +35,17 @@ function Pincode({
...
 
   const onBackspacePress = () => {
     onChangePin(pin.substring(0, pin.length - 1))
   }
 
-  useEffect(() => {
-    // Wait for next frame so we the user can see the last digit
-    // displayed before acting on it
-    if (pin.length === maxLength) {
-      requestAnimationFrame(() => onCompletePin(pin))
```
kathaypacific, line 48: actually i think this is the problem, apologies i had blindly copied and pasted without checking in my previous comment but here we don't need the `requestAnimationFrame`. we can directly execute `onCompletePin(pin)` since there's no reason to wait. i just tested this locally (without the pincode ref) and it seems to work for this use case, does this also work for you?

kathaypacific (commented):
Curious to see if my suggestion can avoid bringing back the ref 🙈

src/pincode/Pincode.tsx
MuckT: It looks like the `requestAnimationFrame` is ensuring that the last pin digit is displayed to the user prior to navigating to the next entry; without it sometimes a dot for the 6th digit and the 5th digit is shown.

kathaypacific: hey sorry for the late reply, this fell off my radar!

i got really curious about this so i dug into this a bit today. i found a few subtle things, i'll elaborate at our 1:1 later today but writing it here also for reference.

the reason why the wrong error is shown for this scenario is because of the way PincodeSet.tsx is handling the verifying and error state. when `onCompletePin2` fails, the verifying state goes back to false so that the the first Pincode screen displays again, because of the `requestAnimationFrame`, the `onCompletePin1` is fired again. that is why we see the "choose a stronger pin" error instead of the "pins don't match" error.

the PincodeSet.tsx file is setting the verifying state via navigation. i have no idea why, but it should be component state. this syncs the error states in `onCompletePin2` by setting the error + verifying state in the same batch. with this change the scenario works correctly for me, but let's try together later!

MuckT:
Closing in favor of https://github.com/valora-inc/wallet/pull/2629.

PR #2610, opened 2022-06-17
Add configurable logging levels

silasbw:
With the default (debug) the logs can be impenetrable during "normal" local
development.

By default this change shouldn't impact the previous logging behavior.

### How others should test

This doesn't require additional testing.

### DIFF
--- a/src/utils/LoggerLevels.ts
+++ b/src/utils/LoggerLevels.ts
@@ -1,6 +1,6 @@
 export enum LoggerLevel {
-  debug = 3,
-  info = 2,
-  warn = 1,
-  error = 0,
+  Debug = 3,
+  Info = 2,
+  Warn = 1,
+  Error = 0,
 }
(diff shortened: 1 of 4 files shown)


src/utils/LoggerLevels.ts
```diff
@@ -0,0 +1,6 @@
+export enum LoggerLevel {
+  debug = 3,
+  info = 2,
+  warn = 1,
+  error = 0,
+}
```
jeanregisser, line 6: nit: we haven't been very consistent so far, I propose the following which matches what a lot of TypeScript projects are doing.
```suggestion
export enum LoggerLevel {
  Debug = 3,
  Info = 2,
  Warn = 1,
  Error = 0,
}
```

jeanregisser (approved):
This is great!

ValoraQA:
Hey @MuckT since NA is mentioned in how others should test, we are skipping this PR
CC: @silasbw 
Thanks

PR #2612, opened 2022-06-20
feat(fiatconnect): Allow wallet to login with FC providers

jophish:
### Description

Adds a method that allows the wallet to login with a FiatConnect provider, potentially prompting the user to provide their PIN if their account is currently locked. Once the newest SDK version is published, clock synchronization will happen automatically within the SDK during the login process (the SDK changes to support this do not include any breaking API changes, so we won't need to change anything here.)

### How others should test

This login functionality is currently unused, but can be tested from a component manually with the following snippet, e.g.:

```ts
   useAsync(async () => {
    const kit = await getContractKitAsync()
    const wallet = kit.getWallet()! as UnlockableWallet
    const [account] = wallet.getAccounts()
    const fiatConnectClient = new FiatConnectClient(
      {
        baseUrl: 'http://localhost:8080',
        providerName: 'test',
        iconUrl: 'N/A',
        network: Network.Alfajores,
        accountAddress: account,
      },
      getSigningFunction(wallet)
    )
    await loginWithFiatConnectProvider(wallet, fiatConnectClient)
  }, [])
```

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


src/fiatconnect/index.ts
```diff
@@ -60,6 +65,40 @@ export async function getFiatConnectProviders(
...
+  }
+
+  const [account] = wallet.getAccounts()
+  if (!wallet.isAccountUnlocked(account)) {
+    await wallet.unlockAccount(account, await getPassword(account), UNLOCK_DURATION)
+  }
+
+  const response = await fiatConnectClient.login()
+  if (!response.isOk) {
+    Logger.error(TAG, `Failure logging in with FiatConnect provider: ${response.error}, throwing`)
```
satish-ravi, line 89: do we want to log the full trace? or is that left to the caller? Also would this affect sentry quota? cc @silasbw

src/fiatconnect/index.ts
```diff
@@ -60,6 +65,40 @@ export async function getFiatConnectProviders(
...
 }
 
+/**
+ * Logs in with a FiatConnect provider. Will not attempt to log in if an
+ * unexpired session already exists, unless the `forceLogin` flag is set to `true`.
+ * If the user's wallet is currently locked, will prompt for PIN entry.
+ */
+export async function loginWithFiatConnectProvider(
+  wallet: UnlockableWallet,
+  fiatConnectClient: FiatConnectApiClient,
```
satish-ravi, line 75: where would the fiatConnectClient instance be created?

src/fiatconnect/index.test.ts
```diff
@@ -134,4 +161,57 @@ describe('FiatConnect helpers', () => {
       ).rejects.toThrowError('Not implemented')
     })
   })
+  describe('loginWithFiatConnectProvider', () => {
+    const wallet = (mocked(KeychainWallet, true) as unknown) as jest.Mocked<KeychainWallet>
```
satish-ravi, line 165: This looks very hacky, is there a better way to do this? Maybe just create a `KeychainWallet` / `FiatConnectClient` instances with dummy args since you've already mocked out the imports?

jophish: Yeah, I meant to add a comment here; this is pending a fix at the moment. Will update with something more correct.

src/fiatconnect/index.ts
jophish: Presumably on the "enter amount" screen? We can also have a function that takes in several FC provider configs and creates a client/logs in to each of them, but still need to decide where that function is called from.

src/fiatconnect/index.ts
```diff
@@ -60,6 +65,40 @@ export async function getFiatConnectProviders(
...
+  }
+
+  const response = await fiatConnectClient.login()
+  if (!response.isOk) {
+    Logger.error(TAG, `Failure logging in with FiatConnect provider: ${response.error}, throwing`)
+    throw response.error
+  }
+}
+
+export function getSigningFunction(wallet: UnlockableWallet): (message: string) => Promise<string> {
```
cajubelt, line 94: what will happen if the FiatConnectApiClient attempts to use this after the wallet has locked? Should this function also deal with that scenario (either by attempting to unlock, or throwing an error that we can catch somewhere)?

jh2oman (approved):

PR #2613, opened 2022-06-21
Add dapps list data to redux, display disclaimer in dapp connection screens

kathaypacific:
### Description

This PR:
- add dapps list data to redux
- update dapp screen to use data from redux
- update WC/dappkit request screens to show disclaimer and hide logos for dapps not in dapp list

Simulator Screen Shot - iPhone 11 - 2022-06-21 at 17 14 23
Simulator Screen Shot - iPhone 11 - 2022-06-21 at 17 13 52

### Other changes

N/A

### Tested

Manually + unit tests

### How others should test

Visit a dapp (through the dapplist, or by connecting from an external browser) - try to connect to the dapp, and make a transaction. 
- The connection and transaction request bottom sheets should display the dapp logo, and no disclaimer text.

Visit a dapp that is not on the dapp list (by connecting from an external browser) - try to connect to the dapp, and make a transaction. 
- The connection and transaction request bottom sheets should **not** display the dapp logo, and display the disclaimer text.

### Related issues

- Relates #2342 and RET-240

### Backwards compatibility

Yes


src/walletConnect/screens/RequestContent.tsx
```diff
@@ -70,7 +72,7 @@ function RequestContent({
     <View style={styles.container}>
       <TopBarIconButton icon={<QuitIcon />} style={styles.closeButton} onPress={handleDeny} />
       <ScrollView>
-        {dappImageUrl && (
+        {isDappListed && dappImageUrl && (
```
jeanregisser, line 75: I strongly think we should still display the dapp image even if not listed.
- it's not very inclusive of us to do that
- it would be impossible for new dapps to test this before being listed

src/walletConnect/screens/useIsDappListed.ts
```diff
@@ -0,0 +1,10 @@
+import { useSelector } from 'react-redux'
+import { dappsListSelector } from 'src/dapps/selectors'
+
+export const useIsDappListed = (dappName?: string) => {
+  const dappsList = useSelector(dappsListSelector)
+
+  return (
+    !!dappName && !!dappsList.find((dapp) => dapp.name.toLowerCase() === dappName.toLowerCase())
+  )
+}
```
jeanregisser, line 10: This could be abused easily. Pretend you're `Ubeswap` and boom 😅 

We should base this on the domain of the dapp url (and callback url for DappKit). 
This may require checking all our current listed dapp urls to be sure they would pass that test.

src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -62,108 +59,54 @@ interface SectionData {
...
+        fontColor: cat.fontColor,
+        backgroundColor: cat.backgroundColor,
+        dapps: [],
       }
+    })
+    dappsList.forEach((dapp: Dapp) => {
+      mappedCategories[dapp.categoryId].dapps.push(dapp)
+    })
+    setCategoriesById(Object.values(mappedCategories))
+  }, [dappsList, dappsCategories])
```
jeanregisser, line 98: Looks like we should have a selector instead of this effect.

src/dapps/saga.ts
```diff
@@ -26,10 +48,64 @@ export function* handleOpenDapp(action: PayloadAction<DappSelectedAction>) {
...
+    yield put(fetchDappsListFailed({ error: 'Could not fetch dapps' }))
+  }
+}
+
 export function* watchDappSelected() {
   yield takeLatest(dappSelected.type, handleOpenDapp)
 }
 
+export function* watchFetchDappsList() {
+  yield takeEvery(fetchDappsList.type, handleFetchDappsList)
```
jeanregisser, line 104: Wondering if we should use [`takeLeading`](https://redux-saga.js.org/docs/api/#takeleadingpattern-saga-args).
So it only does the fetch while no other fetch is already happening.

jeanregisser (commented):
Great work! 👍

src/dappsExplorer/DAppsExplorerScreen.tsx
kathaypacific: yeah good call, i moved this to a selector and added a test!

src/walletConnect/screens/useIsDappListed.ts
jeanregisser: Yes the types are a bit misleading.
The DappKit `callback` is always a url.
For WC, the url is there too, we should use the same logic we use for filling the `dappUrl` for anaytics purpose:
https://github.com/valora-inc/wallet/blob/268b65d121c2038b46dc3778ff35bcd51bb3259d/src/analytics/Properties.tsx#L1029

If for some reason the url is `empty | null | undefined`, we can just deny the listed status.

kathaypacific: ah cool i didn't know about that! thanks for pointing it out, i've added a check for the url now :D

kathaypacific:
@jeanregisser ready for another round! i removed the phone number from the WC connect bottom sheet, i didn't know that was only for dappkit and glad it came up on slack yesterday

src/walletConnect/screens/useIsDappListed.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const useIsDappListed = (dappName?: string, dappUrl?: string) => {
+  const dappsList = useSelector(dappsListSelector)
+
+  return (
+    !!dappName &&
+    !!dappUrl &&
+    !!dappsList.find(
+      (dapp) =>
+        dapp.name.toLowerCase() === dappName.toLowerCase() &&
```
jeanregisser, line 12: Checking the name too was a good idea, unfortunately, the listed name doesn't always match the `peerMeta` name reported.
For instance we have `Ubeswap` in our dapp list, but its `peerMeta.name` is `Ubeswap Interface`.

src/walletConnect/screens/useIsDappListed.ts
```diff
@@ -0,0 +1,16 @@
+import { useSelector } from 'react-redux'
+import { dappsListSelector } from 'src/dapps/selectors'
+
+export const useIsDappListed = (dappName?: string, dappUrl?: string) => {
```
jeanregisser, line 4: Would be nice to add a test suite for this to make sure it works the way we expect.

src/dapps/selectors.ts
```diff
@@ -10,3 +17,39 @@ export const activeDappSelector = (state: RootState) =>
...
+export const featuredDappSelector = createSelector(dappsListSelector, (dapps) => {
+  return dapps.find((dapp) => dapp.isFeatured)
+})
+
+export const dappCategoriesByIdSelector = createSelector(
+  dappsListSelector,
+  dappsCategoriesSelector,
+  (dapps, categories) => {
+    const mappedCategories: {
+      [id: string]: CategoryWithDapps
```
jeanregisser, line 38: For correctness:
```suggestion
      [id: string]: CategoryWithDapps | undefined
```

src/dapps/saga.ts
```diff
@@ -26,10 +48,64 @@ export function* handleOpenDapp(action: PayloadAction<DappSelectedAction>) {
...
 }
 
+export function* watchFetchDappsList() {
+  yield takeLeading(fetchDappsList.type, handleFetchDappsList)
+}
+
 export function* dappsSaga() {
   yield spawn(watchDappSelected)
+  yield spawn(watchFetchDappsList)
+  yield put(fetchDappsList())
```
jeanregisser, line 110: Nothing to do now, but there's an edge case if this fetch fails, some things relying on the data to be present won't work as expected.
That's one downside of doing the manual fetching instead of having a Redux Toolkit Query style hook which we would use where we need the data and would re-trigger it as needed.

src/dapps/selectors.ts
kathaypacific: done

src/dapps/saga.ts
kathaypacific: yeah true, if this doesn't load correctly on app start + the user does not visit the dapps page (or for some reason it fails there too) then we will always show the dapps disclaimer :/ i wonder if we should only show the "not in dapps list" disclaimer if there are dapps in the dapplist?

src/walletConnect/screens/useIsDappListed.ts
kathaypacific: done

src/walletConnect/screens/useIsDappListed.ts
kathaypacific: i missed/didn't expect this...if this is the case then comparing the name doesn't make sense anymore, there might be a scenario where the listed name and peerMeta does not match at all? i removed the check for dapp name in this hook

src/dapps/saga.ts
kathaypacific: i think i'll leave this - we're not clearing the dappsList on refetch, so this edge case would be limited to users who have never been able to fetch the dapps list, and in that case i think it's technically correct to keep the disclaimer if the dapplist is empty

src/dapps/selectors.ts
kathaypacific: although this correctness means we now have a typeguard, to get around this error https://github.com/valora-inc/wallet/runs/7025166914?check_suite_focus=true - it looks a little weird

jeanregisser (approved):
Looking great! 🎉

ValoraQA:
Hey @MuckT we have verified the above task on latest iOS TestFlight Release build V:1.35.0 and updated status on test rails

**Observation :** We didn't get any disclaimer text when we connected wallet to pooltogether , sushiswap on transaction approval screen , 
Please provide us a dapp name so we can test the test case again and update it accordingly

**Test Rail link:**
**iOS 14:** https://valoraapp.testrail.io/index.php?/tests/view/47353

**Device :** iPhone 12(14.7.1)

**CC:** @kathaypacific 

Thanks..!

ValoraQA:
Hey @MuckT we have verified the above task on latest Android Internal Release build V:1.35.0 and updated status on test rails

Test Rail link:
Android 10: https://valoraapp.testrail.io/index.php?/tests/view/47415

Device : Oppo F11 Pro (10.0)

CC: @kathaypacific

Thanks..!

PR #2615, opened 2022-06-21
chore: remove geth logs from Logger and combine with ReactNativeLogger

jeanregisser:
### Description

Title says it all.

There was no need to have a separate `Logger` and `ReactNativeLogger` anymore. Only `Logger` remains.

Also this should help address the issue when exporting big log files. The data doesn't need to go through the RN bridge to create a combined log file anymore. We just use the log path. 
On Android, we're copying the file to a world readable dir, but I believe this extra step can be removed too.

### Tested

- Updated tests
- Manually on iOS and Android

### How others should test

No need to test this.

### Related issues

- Related to RET-256 and RET-336

### Backwards compatibility

Yes


src/account/SupportContact.tsx
```diff
@@ -25,6 +28,28 @@ import { currentAccountSelector } from 'src/web3/selectors'
...
+    if (Platform.OS === 'android') {
+      const publicPath = `${RNFS.ExternalDirectoryPath}/${path.basename(logsPath)}`
+      await RNFS.copyFile(logsPath, publicPath)
+      return publicPath
+    }
+
+    return logsPath
+  } catch (e) {
+    Logger.showError(i18n.t('supportExportLogsError', { error: e }))
+    return false
```
silasbw, line 49: super nit: can you return `null` instead?  I think a return type of `null | string` is more conventional than `boolean | string`.  I realize this was probably based on existing code.

src/account/SupportContact.tsx
```diff
@@ -25,6 +28,28 @@ import { currentAccountSelector } from 'src/web3/selectors'
...
+    const logsPath = Logger.getReactNativeLogsFilePath()
+
+    // For now we need to export to a world-readable directory on Android
+    // TODO: use the FileProvider approach so we don't need to do this.
+    // See https://developer.android.com/reference/androidx/core/content/FileProvider
+    // and https://github.com/chirag04/react-native-mail/blame/340618e4ef7f21a29d739d4180c2a267a14093d3/android/src/main/java/com/chirag/RNMail/RNMailModule.java#L106
+    if (Platform.OS === 'android') {
+      const publicPath = `${RNFS.ExternalDirectoryPath}/${path.basename(logsPath)}`
+      await RNFS.copyFile(logsPath, publicPath)
+      return publicPath
```
silasbw, line 43: Should we do this on android and ios so we have less platform specialization?  All else being equal this _seems_ like it would simplify this implementation and potential future changes (e.g., see comment about log clean up).

silasbw (commented):
:artist: !

src/account/SupportContact.tsx
jeanregisser: Yes exactly, I kept the types expected by the other functions consuming the output of this.

ValoraQA:
Hey @MuckT since NA is mentioned in how others should test , we are skipping this PR
CC: @jeanregisser 
Thanks

PR #2623, opened 2022-06-22
Add remote config for WC/dappkit bottom sheet update

kathaypacific:
### Description

There are some changes for the dapps bottom sheet that we want to test the impact of. 

This PR 
- has questionable variable names (please offer alternatives?)
- adds remote config (i opted for an enum type, since there are plans to expose even more transaction data later)
- uses the remote config to show the same data as previously for the control group (`WalletConnectDisplayedInfo.None`)
- uses the remote config to show an updated connection request title/access to phone number and wallet address for walletConnect connection/dapp not in dapplist disclaimer for the variant group (`WalletConnectDisplayedInfo.Basic`)

### Other changes

Move bottom sheet to`modalAnimatedScreens` to improve transitions

### Tested

N/A

### How others should test

N/A

### Related issues

- Relates #2342 and RET-240

### Backwards compatibility

Yes


src/walletConnect/screens/SessionRequest.tsx
```diff
@@ -21,8 +23,24 @@ function SessionRequest({ pendingSession }: Props) {
...
+          {
+            label: t('phoneNumber'),
+            value: phoneNumber,
+          },
+          {
+            label: t('address'),
+            value: address,
+            tapToCopy: true,
+          },
+        ]
```
jeanregisser, line 42: This will trigger later on when any value is used. 
I wonder if we should be more careful about backward compat for this and handle only cases we know about.

I'm anticipating this case: we add new enum value and don't realize it will display something unexpected for older clients.

src/walletConnect/v1/reducer.ts
```diff
@@ -7,21 +8,29 @@ import { Actions, UserActions, WalletConnectActions } from 'src/walletConnect/v1
 
 export type PendingAction = { action: WalletConnectPayloadRequest; peerId: string }
 
+export enum WalletConnectDisplayedInfo {
+  None = 'none',
+  Basic = 'basic',
+  Full = 'full',
+}
```
jeanregisser, line 15: I propose we extract this to another file (since used by both WC and DAppKit) and give it a more generic name.
Maybe `DappsConnectInfo`? But should we have 2 types, one for controlling Connect and another for Signing? I guess this can be added later.
Can we also add a comment about this type and what each value does?

src/walletConnect/v1/reducer.ts
```diff
@@ -7,21 +8,29 @@ import { Actions, UserActions, WalletConnectActions } from 'src/walletConnect/v1
...
+
 export interface State {
   pendingActions: PendingAction[]
   sessions: WalletConnectSession[]
   pendingSessions: WalletConnectSessionRequest[]
+  walletConnectDisplayedInfo: WalletConnectDisplayedInfo
```
jeanregisser, line 21:
```suggestion
  dappsConnectInfo: DappsConnectInfo
```

src/walletConnect/screens/SessionRequest.tsx
```diff
@@ -21,8 +23,24 @@ function SessionRequest({ pendingSession }: Props) {
...
   const isDappListed = useIsDappListed(name)
 
+  const requestDetails =
+    walletConnectDisplayedInfo === WalletConnectDisplayedInfo.None
+      ? []
+      : [
+          {
+            label: t('phoneNumber'),
+            value: phoneNumber,
+          },
```
jeanregisser, line 36: Looks like we missed this in a previous PR, but phone number is never shared when WC is used. We should remove this from here.

src/walletConnect/v1/reducer.ts
```diff
@@ -7,21 +8,29 @@ import { Actions, UserActions, WalletConnectActions } from 'src/walletConnect/v1
 
 export type PendingAction = { action: WalletConnectPayloadRequest; peerId: string }
 
+export enum WalletConnectDisplayedInfo {
+  None = 'none',
+  Basic = 'basic',
+  Full = 'full',
```
jeanregisser, line 14: Is this used anywhere?

src/walletConnect/screens/useIsDappListed.ts
```diff
@@ -1,10 +1,14 @@
...
+import { WalletConnectDisplayedInfo } from 'src/walletConnect/v1/reducer'
+import { walletConnectDisplayedInfoSelector } from 'src/walletConnect/v1/selectors'
 
 export const useIsDappListed = (dappName?: string) => {
   const dappsList = useSelector(dappsListSelector)
+  const walletConnectDisplayedInfo = useSelector(walletConnectDisplayedInfoSelector)
 
   return (
-    !!dappName && !!dappsList.find((dapp) => dapp.name.toLowerCase() === dappName.toLowerCase())
+    walletConnectDisplayedInfo === WalletConnectDisplayedInfo.None ||
```
jeanregisser, line 11: Seems a bit misleading/dangerous to link the listed status to the DIsplayedInfo remote config.
I think this should be put outside of this hook.

jeanregisser (commented):
Naming is hard 😅 
Mostly anticipating misuse of this remote config. Happy to chat more.

src/walletConnect/v1/reducer.ts
kathaypacific: haha i did go back and forth on this, i started adding this to the dapps slice and then decided it'd be better in the WC slice (but now i remember our conversation about wanting to keep supporting dappkit) so i've moved it back to the dapps slice now. i think that's a good home for this? 

i didn't make 2 variables though because the dappslist disclaimer is on both the connect and signing sheets. i added a comment to explain what each enum does though :)

src/walletConnect/v1/reducer.ts
kathaypacific: updated!

src/walletConnect/v1/reducer.ts
kathaypacific: it's not yet...i added it to make it look less weird to have a 2 value enum :/ should i remove?

src/walletConnect/screens/useIsDappListed.ts
kathaypacific: yeah good point. i was only thinking that we want this for all enum values other than `none` but indeed that's a bit dangerous since we're using free text on firebase. i changed these conditions to check for `basic` instead, when/if we get to implementing the `full` variant we can tweak this code

kathaypacific:
@jeanregisser this is ready for another round too :) it looks like i added a lot of changes but most of it is to move the dapps types to a separate types file (i was getting require cycle issues which made all the tests not run...)

src/walletConnect/screens/useIsDappListed.ts
```diff
@@ -1,16 +1,19 @@
...
 
   return (
-    !!dappName &&
-    !!dappUrl &&
-    !!dappsList.find(
-      (dapp) =>
-        dapp.name.toLowerCase() === dappName.toLowerCase() &&
-        new URL(dapp.dappUrl).origin === new URL(dappUrl).origin
-    )
+    dappConnectInfo === DappConnectInfo.None ||
```
jeanregisser, line 10: I still think this shouldn't be here. The fact that the dapp is listed or not doesn't depend on the value of this remote config.

If this hook was named `useShouldShowLogo` or something, that would be fine. Would be still interesting to have a separate `useIsDappListed` though.

src/dappkit/DappKitAccountScreen.tsx
```diff
@@ -56,8 +58,11 @@ const DappKitAccountScreen = ({ route }: Props) => {
       <RequestContent
         onAccept={handleAllow}
         onDeny={handleCancel}
-        dappImageUrl={activeDapp?.iconUrl}
-        title={t('connectToWallet', { dappName: dappKitRequest.dappName })}
+        dappImageUrl={dappConnectInfo === DappConnectInfo.Basic ? activeDapp?.iconUrl : undefined}
+        title={t(
+          dappConnectInfo === DappConnectInfo.Basic ? 'connectToWallet' : 'confirmTransaction',
```
jeanregisser, line 63: ditto about avoiding dynamic i18n keys.

src/dapps/types.ts
```diff
@@ -0,0 +1,33 @@
...
+export interface DappCategory {
+  backgroundColor: string
+  fontColor: string
+  id: string
+  name: string
+}
+
+// used for the dapp connect request bottom sheet
+export enum DappConnectInfo {
+  None = 'none', // display the same content as before app version 1.35
```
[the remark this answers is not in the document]
kathaypacific: yeah this is a better name 😅 i used "none" because that's how much data we were showing but indeed "default" sounds less passive aggressive

jeanregisser (approved):
🎉

ValoraQA:
Cc: @kathaypacific 
Thanks.!

PR #2635, opened 2022-06-23
Add placeholder image if no dapp logo can be displayed

kathaypacific:
### Description

For dapps using dappkit and not listed on the dapps list, there is no image that we can display on the connection/transaction request bottom sheets. This PR adds a default placeholder image (inspired by the send screen)

I tried to use the favicons but found with the first 2 dapps i tried (revo, moola) that their favicons were already located in different urls (not the standard `/favicon.ico`) so i gave up. a quick google search said that around 40% of websites don't have a favicon so wasn't sure that the favicon fallback would be fullproof anyway. 

Simulator Screen Shot - iPhone 11 - 2022-06-23 at 11 38 05

### Other changes

N/A

### Tested

N/A

### How others should test

N/A

### Related issues

- Relates #2342

### Backwards compatibility

N/A


src/walletConnect/screens/RequestContent.tsx
```diff
@@ -177,6 +191,16 @@ const styles = StyleSheet.create({
...
     alignSelf: 'flex-end',
   },
+  placeholderLogoBackground: {
+    backgroundColor: Colors.light,
+    marginRight: -Spacing.Small12,
+    borderColor: Colors.gray2,
+    borderWidth: 1,
+  },
+  placeholderLogoText: {
+    ...fontStyles.h2,
```
jeanregisser, line 201: I saw the text is not vertically centered, I think setting `lineHeight: undefined` helps with this.

jeanregisser (approved):
Awesome! 🧑‍🎨 
Approved with a suggestion.

src/walletConnect/screens/RequestContent.tsx
kathaypacific: ah yes, thank you for your pixel perfect vision! i didn't even notice 🙈

removing the line height worked, i also changed the size to h1 to make the text size match the valora icon more
Simulator Screen Shot - iPhone 11 - 2022-06-24 at 10 41 48

PR #2658, opened 2022-06-29
feat(fiatconnect): Initialize FiatConnect SDK and send Fiat Account to providers

jophish:
### Description

This PR adds in actual use of the FiatConnect SDK; clients are stored inside `FiatConnectQuote` objects, which are _already_ passed through the CICO flow components. This also updates the `FiatDetailsScreen` component to actually use the SDK to post the user's bank account details to a provider.

### Other changes

### Tested

https://user-images.githubusercontent.com/569401/176942254-a37c441a-bf7a-4655-afa1-0139033bb348.mp4

### How others should test

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -140,31 +148,38 @@ const FiatDetailsScreen = ({ route, navigation }: Props) => {
...
-      //   })
-
-      navigate(Screens.FiatConnectReview, {
-        flow,
-        normalizedQuote: quote,
-        fiatAccount: completeBody as FiatAccount,
+      const fiatConnectClient = await quote.getFiatConnectClient()
+      const result = await fiatConnectClient.addFiatAccount({
+        fiatAccountSchema: quote.getFiatAccountSchema(),
+        data: completeBody as FiatAccountSchemas[typeof fiatAccountSchema],
```
cajubelt, line 160: the casting here seems sub-optimal, could be worth taking a sec to try getting the types to work better

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -140,31 +148,38 @@ const FiatDetailsScreen = ({ route, navigation }: Props) => {
...
-      //       dispatch(showError(t('fiatDetailsScreen.addFiatAccountFailed')))
-      //     }
-      //     Logger.error(TAG, `Error adding fiat account: ${error}`)
-      //   })
-
-      navigate(Screens.FiatConnectReview, {
-        flow,
-        normalizedQuote: quote,
-        fiatAccount: completeBody as FiatAccount,
+      const fiatConnectClient = await quote.getFiatConnectClient()
```
kathaypacific, line 157: *a wild kathy appeared* this might be something we can consider in a separate PR, but is it possible to abstract this kind of logic to a saga? this would better separate concerns, where this component can focus on displaying things and dispatch an action that kicks off async logic. there are some added benefits to this (e.g. easier to test because it is easier to test async logic in a saga, shorter UI file) happy to chat if i've not been clear here! :)

jophish: Hmm, that would involve storing state in redux about, e.g., the most recent attempt to send fiat account information to a provider, right? Would the general strategy here look something like:
* Select the piece of state containing the add account response in the FiatDetailsScreen component
* While the add account response is null/empty, do nothing
* When the button is pressed, dispatch an action to add the account and save the response in state
* While the add account response is defined, check the result and either show an error or progress to the next page
Apologies if that's totally off :slightly_smiling_face:

src/fiatExchanges/quotes/FiatConnectQuote.ts
```diff
@@ -147,4 +154,22 @@ export default class FiatConnectQuote extends NormalizedQuote {
...
+  async getFiatConnectClient(): Promise<FiatConnectApiClient> {
+    if (this.fiatConnectClient) {
+      return this.fiatConnectClient
+    }
+    const kit = await getContractKitAsync()
+    const wallet = kit.getWallet()! as UnlockableWallet
+    const [account] = wallet.getAccounts()
+    this.fiatConnectClient = new FiatConnectClient(
+      {
+        baseUrl: this.quote.provider.baseUrl,
```
jh2oman, line 167:
```suggestion
        baseUrl: this. getProviderBaseUrl(),
```

src/fiatconnect/FiatDetailsScreen.tsx
jh2oman: On activation we have been doing a small amount of our work in saga's, but mostly been leaving this kind of logic to helper functions outside of UI components. I agree that we should be doing more in saga's and since this is a fresh implementation it would be a good place to start! :) 

@jophish thats not quite the model I was thinking of for this type of an event since we don't *need* to store any data in state yet.

I think you can do all the logic you currently have with an Action, a watch* saga, and a saga function. All the navigation / conditionally showing errors that you currently have in this function can live in the saga.

For example take a look at account initialization:

* Action: https://github.com/valora-inc/wallet/blob/main/src/account/actions.ts#L317
* watch* saga: https://github.com/valora-inc/wallet/blob/main/src/account/saga.ts#L256
* saga: https://github.com/valora-inc/wallet/blob/main/src/account/saga.ts#L87

Together they watch for that INITIALIZE_ACCOUNT action in redux and call the initializeAccount() saga when its seen.

jh2oman: The only question I have for @kathaypacific on this approach though is passing the `quote` object or `fiatConnectClient` (both of which are instantiated classes) through the dispatch action so that it can be used in the saga. Is it okay to do this? I know its generally best practice to stick with minimal data types in redux

jh2oman: Oh @jeanregisser maybe you can comment on this approach since Kathy is on vacation now :)

cajubelt: good idea @kathaypacific and thanks for explaining @jh2oman !

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -140,31 +148,36 @@ const FiatDetailsScreen = ({ route, navigation }: Props) => {
...
```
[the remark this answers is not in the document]
jophish: Just added a spinner -- good catch!

src/fiatExchanges/quotes/FiatConnectQuote.ts
```diff
@@ -147,4 +154,22 @@ export default class FiatConnectQuote extends NormalizedQuote {
   getFiatAccountSchemaAllowedValues(): { [key: string]: string[] } {
     return this.quoteResponseFiatAccountSchema.allowedValues
   }
+
+  async getFiatConnectClient(): Promise<FiatConnectApiClient> {
+    if (this.fiatConnectClient) {
+      return this.fiatConnectClient
+    }
+    const kit = await getContractKitAsync()
+    const wallet = kit.getWallet()! as UnlockableWallet
```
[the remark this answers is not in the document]
jophish: thanks!

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -140,31 +151,38 @@ const FiatDetailsScreen = ({ route, navigation }: Props) => {
...
       })
+      if (result.isOk) {
+        // TODO Tracking here
+        dispatch(showMessage(t('fiatDetailsScreen.addFiatAccountSuccess')))
+        navigate(Screens.FiatConnectReview, {
+          flow,
+          normalizedQuote: quote,
+          fiatAccount: completeBody as FiatAccount,
+        })
+        setTimeout(() => setIsSending(false), 500)
```
jophish, line 173: Without a timeout to remove the spinner, there's a flicker before navigating to the review screen.

ValoraQA:
hey @MuckT  please let us know whether we need to test it or not

 Cc: @jophish

MuckT:
@ValoraQA some context  from a conversation from @jophish
> They should test in Alfajores, since we have a mock FC provider set up that will display in the select provider screen. It's a good idea to test that when a user enters fiat account information, they're able to enter their PIN and successfully navigate to the review screen.
> 1. Enter the "withdraw" flow
> 2. Select some currency/amount to withdraw
> 3. Select the "Valora" cash out icon (this is the mock FC provider)
> 4. Enter random info for fiat account
> 5. Hit 'next'
> 6. should be prompted for PIN if it wasn't entered anywhere else in the app in the last 15 minutes or so
> 7. should see spinner after entering PIN
> 8. should be on review screen and see a banner dropdown indicating success
> 9. should be able to navigate back, re-enter some random info, and not have to enter PIN again (assuming within 15 minute window)

PR #2660, opened 2022-06-30
fix: list of countries in Europe

jeanregisser:
### Description

I noticed Poland had an incorrect [ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) code. It was set to `PO` instead of `PL`.
So I fixed + completed it from a https://en.wikipedia.org/wiki/List_of_sovereign_states_and_dependent_territories_in_Europe.

### Tested

Not tested.

### How others should test

No need to test this.

### Backwards compatibility

Yes


MuckT (approved):
Thanks adding comments with the countries as it makes this much more readable!

PR #2661, opened 2022-06-30
feat(coinbase pay): Shows a touchable Coinbase Pay card if user is eligible

awtsui:
### Description

Coinbase Pay's "quote" is extracted before normalizeQuote and fed into UniquePaymentSection, a simplified payment card component

### Other changes

N/A

### Tested

Checks if coinbase pay card is shown depending on user eligibility, restricted value

### How others should test

PaymentMethodSection.test.tsx

### Related issues

[https://linear.app/valora/issue/ACT-343/add-coinbase-pay-card-to-payment-provider-screen](url)

### Backwards compatibility

N/A


src/fiatExchanges/UniquePaymentSection.tsx
```diff
@@ -0,0 +1,84 @@
...
+import fontStyles from 'src/styles/fonts'
+
+export interface UniquePaymentSectionProps {
+  paymentMethod: PaymentMethod
+  uniqueProvider: FetchProvidersOutput | null
+  setNoPaymentMethods: React.Dispatch<React.SetStateAction<boolean>>
+  flow: CICOFlow
+}
+
+export function UniquePaymentSection({
```
cajubelt, line 17: It seems like you're trying to create a component that can be re-used for more bespoke payment methods in the future (that aren't part of FiatConnect or covered by an existing method). Personally I see this as an unnecessary abstraction. We have no plans to support more bespoke payment methods, and until we do, it provides extra cognitive effort for little gain

src/fiatExchanges/UniquePaymentSection.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  return (
+    <View style={styles.container}>
+      <Touchable>
+        <View style={{ ...styles.innerContainer, paddingVertical: 27 }}>
+          <View style={styles.left}>
+            <Text style={styles.category}>
+              {paymentMethod === PaymentMethod.Coinbase
+                ? t('selectProviderScreen.coinbase')
+                : t('selectProviderScreen.somePaymentsUnavailable')}
```
cajubelt, line 38: This is a very awkward fallback for the non-coinbase, unrecognized method case. "Some payment methods may be unavailable in your region" (or whatever this says) is supposed to appear when noPaymentMethods flips to true, not when we accidentally used this component for a provider we didn't decide on a payment method title for, and it's supposed to appear somewhere else on the page, not as a payment method category title. This weirdness seems like fallout from the decision to make the component agnostic to the actual method-- if it's explicitly coinbase-specific I think the problem would go away

src/fiatExchanges/utils.tsx
```diff
@@ -283,3 +284,12 @@ export async function fetchExchanges(
...
   }
 }
+
+export const filterProvidersByPaymentMethod = (
+  paymentMethod: PaymentMethod,
+  externalProviders: FetchProvidersOutput[] | undefined
+) => {
+  return externalProviders
+    ? externalProviders?.filter((quote) => quote.paymentMethods[0] === paymentMethod)[0]
+    : null
```
cajubelt, line 294:
```suggestion
  return externalProviders?.find((quote) => quote.paymentMethods[0] === paymentMethod)
```
code golf

test/values.ts
```diff
@@ -643,6 +643,19 @@ export const mockProviders: FetchProvidersOutput[] = [
...
   },
+  {
+    name: 'CoinbasePay',
+    restricted: false,
+    unavailable: false,
+    paymentMethods: [PaymentMethod.Coinbase],
+    url: undefined,
+    logo:
+      'https://firebasestorage.googleapis.com/v0/b/celo-mobile-mainnet.appspot.com/o/images%2FcbPay-button.png?alt=media',
+    logoWide: '',
```
cajubelt, line 654: I think just re-using the same image as for logo would be preferable to an emptystring here

cajubelt: oh just noticed these are test values. still, slightly better.

cajubelt, line 654:
```suggestion
    logoWide: 'https://firebasestorage.googleapis.com/v0/b/celo-mobile-mainnet.appspot.com/o/images%2FcbPay-button.png?alt=media',
```

src/fiatExchanges/UniquePaymentSection.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+      uniqueProvider: mockProviders.filter(
+        (quote) => quote.paymentMethods[0] === PaymentMethod.Coinbase
+      )[0],
+      setNoPaymentMethods: jest.fn(),
+      flow: CICOFlow.CashIn,
+    }
+  })
+  it('shows nothing if unique provider, IE. coinbase, is restricted', async () => {
+    props.uniqueProvider!.restricted = true
+    const { queryByText } = render(<UniquePaymentSection {...props} />)
```
cajubelt, line 24: probly also worth adding a test, either here or in the parent component, to make sure the `setNoPaymentMethods` callback has been used

cajubelt (commented):
Looks pretty good! Please include a screenshot when there's new UI stuff in the PR in the future

src/fiatExchanges/UniquePaymentSection.tsx
```diff
@@ -0,0 +1,84 @@
...
+  uniqueProvider: FetchProvidersOutput | null
+  setNoPaymentMethods: React.Dispatch<React.SetStateAction<boolean>>
+  flow: CICOFlow
+}
+
+export function UniquePaymentSection({
+  paymentMethod,
+  uniqueProvider,
+  setNoPaymentMethods,
+  flow,
```
jh2oman, line 21: Flow should be removed if its not used at all in the component

src/fiatExchanges/UniquePaymentSection.tsx
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  return (
+    <View style={styles.container}>
+      <Touchable>
+        <View style={{ ...styles.innerContainer, paddingVertical: 27 }}>
+          <View style={styles.left}>
+            <Text style={styles.category}>
+              {paymentMethod === PaymentMethod.Coinbase
+                ? t('selectProviderScreen.coinbase')
```
jh2oman, line 37: `Coinbase Pay` will get translated into different languages when you use a string through this translator function. Do we actually want that to get translated since its a brand name?

src/fiatExchanges/UniquePaymentSection.tsx
jophish: +1. I think this could just be a `CoinbasePaymentSection` component or something similarly named; no reason to have an abstraction here when this will only ever be used for Coinbase.

cajubelt:
As Jacob mentioned on slack, this shouldn't be merged until it's hidden behind a feature flag; otherwise we could accidentally release the part-way implemented feature

src/fiatExchanges/UniquePaymentSection.tsx
awtsui: Okay! Thanks for the clarification.

src/fiatExchanges/UniquePaymentSection.tsx
awtsui: Probably not. I'll take it out of translation.

src/fiatExchanges/CoinbasePaymentSection.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+export function CoinbasePaymentSection({
+  coinbaseProvider,
+  setNoPaymentMethods,
+}: CoinbasePaymentSectionProps) {
+  const { t } = useTranslation()
+  const coinbasePayEnabled = useSelector(coinbasePayEnabledSelector)
+
+  if (!coinbaseProvider || coinbaseProvider.restricted || !coinbasePayEnabled) {
+    setNoPaymentMethods(true)
```
jh2oman, line 25: Since coinbase pay isn't available for cash outs this setNoPaymentMethods will cause the little blurb at the bottom saying 'some payment methods are unavailable' to always show up. I don't think we actually want this experience so i think setNoPaymentMethods can be removed from the component

src/fiatExchanges/CoinbasePaymentSection.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+  })
+  it('shows nothing if coinbase is restricted and feature flag is false', async () => {
+    props.coinbaseProvider!.restricted = true
+    const { queryByText } = render(
+      <Provider store={mockStore}>
+        <CoinbasePaymentSection {...props} />
+      </Provider>
+    )
+    expect(queryByText('Coinbase Pay')).toBeFalsy()
+    expect(props.setNoPaymentMethods).toHaveBeenCalled()
```
cajubelt, line 37: Jacob suggested you just take the `setNoPaymentMethods` part out of the component, which is probably right. But if you didn't, you'd want to test what it was called with, not just that it was called

src/fiatExchanges/CoinbasePaymentSection.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+    props.coinbaseProvider!.restricted = false
+    const { queryByText } = render(
+      <Provider store={mockStore}>
+        <CoinbasePaymentSection {...props} />
+      </Provider>
+    )
+    expect(queryByText('Coinbase Pay')).toBeFalsy()
+    expect(props.setNoPaymentMethods).toHaveBeenCalled()
+  })
+  it('shows nothign if coinbase is restricted and feature flag is true', async () => {
```
cajubelt, line 49:
```suggestion
  it('shows nothing if coinbase is restricted and feature flag is true', async () => {
```

src/fiatExchanges/CoinbasePaymentSection.tsx
cajubelt: seems like we're likely to continue to running into this issue, might be worth thinking about a more general solution (besides just not taking coinbase pay into account)

cajubelt: for example, I'm really not sure we offer bank account cash-outs anywhere

jophish: Do we want the 'some payment methods are unavailable' text to show up on _cash in_ if CB Pay isn't available?

src/fiatExchanges/utils.tsx
```diff
@@ -283,3 +284,10 @@ export async function fetchExchanges(
     throw error
   }
 }
+
+export const filterProvidersByPaymentMethod = (
```
jophish, line 288: This works for Coinbase, but if we were to ever use this more generally, it wouldn't work, since some providers have multiple payment methods; we should check all entries of `quote.paymentMethods`.

cajubelt (approved):
🥳

jh2oman (approved):
Awesome!

PR #2666, opened 2022-07-01
chore(fiatconnect): remove account name field from form

cajubelt:
### Description

uses an obfuscated account number instead

### Other changes

na

### Tested

automated unit tests. also checked that fiat account page still renders normally; it does

### How others should test

na

### Related issues

fixes https://linear.app/valora/issue/ACT-317/remove-account-name-from-bank-account-info-withdraw-flow

### Backwards compatibility

🆗


src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -69,7 +76,10 @@ const getAccountNumberSchema = (implicitParams: {
   },
   country: { name: 'country', value: implicitParams.country },
   fiatAccountType: { name: 'fiatAccountType', value: FiatAccountType.BankAccount },
-  accountName: { name: 'accountName', value: 'n/a' },
+  accountName: {
+    name: 'accountName',
+    computeValue: ({ accountNumber }) => getObfuscatedAccountNumber(accountNumber ?? ''),
```
satish-ravi, line 81: https://github.com/valora-inc/wallet/pull/2667/files#r912256759 did we want to do this here or a separate PR?

cajubelt: checking on slack what Jacob thinks since it depends also on his pr

src/fiatconnect/index.ts
```diff
@@ -184,3 +184,21 @@ export async function addNewFiatAccount(
...
+ * Ensures at least 3 digits are blanked out for user privacy since it is expected that this will be used to
+ *  compute an accountName for the fiat account, which is returned in GET /accounts and thus shouldn't just
+ *  be the user's bank account number. GET /accounts is still authenticated via SIWE, so at least 3 blanked digits
+ *  should be acceptable; the obfuscation is just a secondary layer of security around sensitive info.
+ *
+ * @param accountNumber
+ */
+export function getObfuscatedAccountNumber(accountNumber: string): string {
+  const digitsToReveal = Math.max(0, Math.min(accountNumber.length - 3, 4))
+  const blanks = '*'.repeat(accountNumber.length - digitsToReveal)
```
jh2oman, line 202: All of the designs I've seen just show three dots followed by last 4 digits. Was there a reason you changed to the * and made the number variable? I was imagining that if we run into schemas that have potentially 4 or less identifying numbers that we could handle it in the future on a case by case basis.

https://www.figma.com/file/bwIY7peeykI8K4wTrXs4iN/Fiat-Connect?node-id=1498%3A40640
https://www.figma.com/file/bwIY7peeykI8K4wTrXs4iN/Fiat-Connect?node-id=1895%3A38461

cajubelt: (The * thing was carried over from IHL, which I tried to largely copy from for this since we've done it before there)

cajubelt: this is also explained by the docstring

cajubelt: also I don't agree with the strategy of punting on it-- we won't necessarily notice that we're adding geos that sometimes have short account numbers. In fact I once had an account number in the US that was just 5 digits

jh2oman: Thanks, the reasoning for including truncation logic now makes sense

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -43,10 +47,35 @@ interface ImplicitParam<T, K extends keyof T> {
...
+
+export function getAccountName({
+  accountNumber,
+  institutionName,
+}: {
+  accountNumber?: string
+  institutionName?: string
+}) {
+  if (institutionName && accountNumber) {
+    return institutionName + ' ' + getObfuscatedAccountNumber(accountNumber)
```
jh2oman, line 70: The designs in figma this string always have parentheses as well 
```suggestion
    return `${institutionName} (${getObfuscatedAccountNumber(accountNumber)})`
```

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -43,10 +47,35 @@ interface ImplicitParam<T, K extends keyof T> {
...
   [Property in keyof FiatAccountSchemas[FiatAccountSchema.AccountNumber]]:
     | FormFieldParam
     | ImplicitParam<FiatAccountSchemas[FiatAccountSchema.AccountNumber], Property>
+    | ComputedParam<FiatAccountSchemas[FiatAccountSchema.AccountNumber], Property>
+}
+
+export function getAccountName({
+  accountNumber,
+  institutionName,
+}: {
```
satish-ravi, line 65: wonder if this should just take in fiat account schema and map it to an account name. I think we can remove the conditional because both these are required in `AccountNumber` schema. 
Could also refactor this in a follow up when the wallet starts supporting multiple schemas.

src/fiatconnect/FiatDetailsScreen.tsx
satish-ravi, line 70: Per the designs I think this should be 
```suggestion
    return '${institutionName} (${getObfuscatedAccountNumber(accountNumber)})'
```
?

src/fiatconnect/FiatDetailsScreen.test.tsx
```diff
@@ -80,7 +80,7 @@ describe('FiatDetailsScreen', () => {
     fireEvent.changeText(getByTestId('input-accountNumber'), fakeAccountNumber)
 
     const expectedBody = {
-      accountName: 'n/a',
+      accountName: 'CapitalTwo Bank ...7890',
```
jh2oman, line 83: The test change you'll need if you take my other suggestion
```suggestion
      accountName: 'CapitalTwo Bank (...7890)',
```

src/fiatconnect/index.ts
satish-ravi: related to this, we enforce 10 digits for all account numbers in the FiatDetailsScreen, but should it be country specific? Looks like per the spec, this is only for NG accounts https://github.com/fiatconnect/specification/blob/main/fiatconnect-api.md#8321-accountnumber

src/fiatconnect/FiatDetailsScreen.tsx
cajubelt: I moved this to be an anonymous function again so it's more clear it's specific to AccountNumber

src/fiatconnect/index.ts
cajubelt: good point.. should do this in a followup

cajubelt: https://linear.app/valora/issue/ACT-357/account-number-length-should-be-country-specific

satish-ravi (approved):

ValoraQA:
hey @MuckT since NA is mentioned in how others should test section , we are skipping this PR

Cc: @cajubelt 
thanks

PR #2674, opened 2022-07-06
refactor(fiatconnect): move quote fetching to sagas

jh2oman:
### Description

This refactor shouldn't have any changes to the experience, it was probably a good refactor to make anyways and makes upcoming work for a return user flow easier. https://linear.app/valora/issue/ACT-349/fc-return-user-flow

### Other changes

### Tested

Manually tested

### How others should test

n/a


jeanregisser (commented):
This works (minus some small gotchas about existing state, see comments below).

Nothing to do here now, but we could reduce the boilerplate for our data fetching needs by using [Redux Toolkit Query](https://redux-toolkit.js.org/rtk-query/overview).
I remember @gnardini tried it but then reverted it because of a weird caching bug though.
I'll bring this up on Slack and/or at our next Eng meeting, to see if we can agree on a common solution.

src/fiatconnect/saga.ts
```diff
@@ -0,0 +1,52 @@
...
+}: ReturnType<typeof fetchFiatConnectQuotes>) {
+  const { flow, digitalAsset, cryptoAmount } = params
+  const userLocation: UserLocationData = yield select(userLocationDataSelector)
+  const account: string | null = yield select(currentAccountSelector)
+  const localCurrency: LocalCurrencyCode = yield select(getLocalCurrencyCode)
+  const fiatConnectCashInEnabled: boolean = yield select(fiatConnectCashInEnabledSelector)
+  const fiatConnectCashOutEnabled: boolean = yield select(fiatConnectCashOutEnabledSelector)
+
+  try {
+    // @ts-ignore TS complains about 'No overload matches this call' and I can't figure it out
```
satish-ravi, line 29: I think this is because of the type of `account`. Its `string | null` in this function vs `string` in FetchQuotesInput: https://github.com/valora-inc/wallet/pull/2674/files#diff-6a81d291c56147f95756fd7bd75460398eba135ed37571281b8345270c2cb65fL164

src/fiatconnect/saga.test.ts
```diff
@@ -0,0 +1,92 @@
...
+import { mockFiatConnectQuotes } from 'test/values'
+
+jest.mock('src/fiatconnect')
+
+describe('Fiatconnect saga saga', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+  describe('Handles fetching quotes', () => {
+    it('saves the dapps and categories', async () => {
```
satish-ravi, line 28: rename test

src/fiatconnect/saga.test.ts
```diff
@@ -0,0 +1,92 @@
...
+import { CICOFlow } from 'src/fiatExchanges/utils'
+import { getLocalCurrencyCode } from 'src/localCurrency/selectors'
+import { userLocationDataSelector } from 'src/networkInfo/selectors'
+import { CiCoCurrency } from 'src/utils/currencies'
+import { currentAccountSelector } from 'src/web3/selectors'
+import { mockFiatConnectQuotes } from 'test/values'
+
+jest.mock('src/fiatconnect')
+
+describe('Fiatconnect saga saga', () => {
```
satish-ravi, line 23:
```suggestion
describe('Fiatconnect saga', () => {
```

src/fiatconnect/saga.test.ts
```diff
@@ -0,0 +1,92 @@
...
+  fetchFiatConnectQuotesFailed,
+} from 'src/fiatconnect/slice'
+import { CICOFlow } from 'src/fiatExchanges/utils'
+import { getLocalCurrencyCode } from 'src/localCurrency/selectors'
+import { userLocationDataSelector } from 'src/networkInfo/selectors'
+import { CiCoCurrency } from 'src/utils/currencies'
+import { currentAccountSelector } from 'src/web3/selectors'
+import { mockFiatConnectQuotes } from 'test/values'
+
+jest.mock('src/fiatconnect')
```
satish-ravi, line 21: this mock works correctly unlike the `getPassword` one? Maybe because `fetchQuotes` has an explicit mock implementation in the test? 😕

src/fiatconnect/saga.test.ts
```diff
@@ -0,0 +1,92 @@
...
+
+describe('Fiatconnect saga saga', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+  describe('Handles fetching quotes', () => {
+    it('saves the dapps and categories', async () => {
+      ;(fetchQuotes as jest.Mock).mockImplementation(() => mockFiatConnectQuotes)
```
satish-ravi, line 29: you can use the [`mocked` helper](https://kulshekhar.github.io/ts-jest/docs/26.5/guides/test-helpers#mockedtitem-t-deep--false) instead of `as`. Looks like this is available on jest directly since 27.4, but we can use this helper from ts-jest

jeanregisser (approved):
LGTM 👍 

There's still the possible edge case of showing the previous error when navigating back.
One way to solve it would be to clear the error on unmount (dispatching an action). I know it doesn't feel great though.

ValoraQA:

Cc: @jh2oman 
Thanks.!

PR #2680, opened 2022-07-08
fix: reduce log size

MuckT:
### Description

Creates monthly logs files. If a log file is older than 60 days it's deleted on app start. When sending to support all log files are separately attached to the support email.

### Other changes

- Updated [react-native-email-link](https://github.com/flexible-agency/react-native-email-link/releases/tag/v1.14.0) to latest.
- Updated [react-native-fs](https://github.com/itinance/react-native-fs/releases/tag/v2.20.0) to latest - was a custom Celo branch 50 commits behind main.
- Updated [react-native-mail](https://github.com/chirag04/react-native-mail/releases/tag/v6.1.1) to latest.
- Added unit tests for Logger.

### Tested

Tested locally on iOS with a device to known reproduce the issue of failing to send logs to support.

### How others should test

**Perquisite**: An email client present on test device.

1. Navigate to Settings
2. Tap 'Show Debug Screen'  
3. Observe logs present
4. Tap 'Email Logs to Support'
5. Type in a Test Message e.g. 'Testing Logs'
6. Tap 'Submit'
7. Observe email client with pre-populated message and logs attached.
 
### Related issues

- Fixes https://github.com/valora-inc/support-issues/issues/74

### Backwards compatibility

Yes


src/utils/Logger.ts
```diff
@@ -172,6 +172,20 @@ class Logger {
     const logFilePath = this.getReactNativeLogsFilePath()
     console.debug('React Native logs will be piped to ' + logFilePath)
 
+    // If the creation time of the file is more than 28 days ago, delete it.
+    RNFS.stat(logFilePath)
+      .then((stat) => {
+        if (+stat.ctime < +new Date() - 4 * 7 * 24 * 60 * 60 * 1000) {
```
silasbw, line 178: Are we worried about losing valuable log content that could help support fix issues?

Another simple option that avoids this is to use a different log file for each day or week (and delete the old ones).  We keep recent content, but can still clean up the old stuff that doesn't matter.

MuckT: If the user is sending a log to support in the worst case we would have from the application start to sending the contact support which would likely contain the issue; additionally, since all the transactions are on the chain support can and developers can also be triage via [Celo Explorer](https://explorer.celo.org/). 

We could create logs for each week or day and combine them when sending to support, we'd just need a bit more logic around combining logs alongside finding and deleting old logs. If this is preferred we can certainly do it either in a follow up PR or adjust this PR.

src/utils/Logger.ts
```diff
@@ -172,6 +172,20 @@ class Logger {
...
+          console.debug('Deleting React Native logs file older than 28 days')
+          RNFS.unlink(logFilePath).catch((err) => {
+            console.warn('Failed to delete React Native logs file: ' + err)
+          })
+        }
+      })
+      .catch((err) => {
+        console.warn('Failed get log file stat: ' + err)
+      })
```
jeanregisser, line 187: What do you think of moving this logic to the first `writeLog` call instead and waiting for it before doing the actual write?
Otherwise there's a risk we write the log and delete the file right after. Given these operations are all async and there's no guarantee as which one will be completed first.

jeanregisser (commented):
Thanks for taking this one 🙏

src/utils/Logger.ts
```diff
@@ -146,15 +148,71 @@ class Logger {
     return error
   }
 
-  getReactNativeLogsFilePath = () => {
-    return RNFS.CachesDirectoryPath + '/rn_logs.txt'
+  getReactNativeLogFilePath = () => {
+    return `${RNFS.CachesDirectoryPath}/rn_logs/${format(new Date(), 'yyyyMMdd')}.txt`
   }
```
jeanregisser, line 153:
```suggestion
  getReactNativeLogFilePath = () => {
    return `${this.getReactNativeLogsDir()}/${format(new Date(), 'yyyyMMdd')}.txt`
  }
```

src/utils/Logger.ts
```diff
@@ -146,15 +148,71 @@ class Logger {
...
+
+    // If log folder not present create it
+    if (!(await RNFS.exists(logDir))) {
+      await RNFS.mkdir(logDir)
+    }
+
+    // If daily log file is not present create it
+    if (!(await RNFS.exists(logFilePath))) {
+      await RNFS.writeFile(logFilePath, '', 'utf8')
+    }
```
jeanregisser, line 183: Looks like this is not necessary, right?

src/utils/Logger.ts
```diff
@@ -166,10 +224,41 @@ class Logger {
...
+      return false
+    }
+  }
+
   // Anything being sent to console.log, console.warn, or console.error is piped into
   // the logfile specified by getReactNativeLogsFilePath()
-  overrideConsoleLogs = () => {
-    const logFilePath = this.getReactNativeLogsFilePath()
+  overrideConsoleLogs = async () => {
+    const logFilePath = this.getReactNativeLogFilePath()
```
jeanregisser, line 261:
```suggestion
  overrideConsoleLogs = () => {
    const logFilePath = this.getReactNativeLogFilePath()
```

jeanregisser (commented):
Great!

src/utils/Logger.ts
```diff
@@ -146,15 +148,71 @@ class Logger {
...
+  initializeLogs = async () => {
+    const logFileCombinedPath = this.getCombinedLogsFilePath()
+    const logFilePath = this.getReactNativeLogFilePath()
+    const logDir = this.getReactNativeLogsDir()
+    console.debug('React Native logs will be piped to ' + logFilePath)
+
+    // If log folder not present create it
+    if (!(await RNFS.exists(logDir))) {
+      await RNFS.mkdir(logDir)
+    }
```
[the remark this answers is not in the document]
MuckT: Added the creation of the directories to the `writeLog` function inside `overrideConsoleLogs`. Testing locally this is capturing all the logs. I moved the log cleanup to a separate function that is called after `overrideConsoleLogs`.

src/utils/Logger.ts
MuckT: I was having trouble appending to a file that doesn't exist so I had to check that it was created first.

src/utils/Logger.ts
```diff
@@ -182,7 +262,17 @@ class Logger {
...
 
-    const writeLog = (level: string, message: string) => {
+    const writeLog = async (level: string, message: string) => {
+      // If log folder not present create it
+      if (!(await RNFS.exists(this.getReactNativeLogsDir()))) {
+        await RNFS.mkdir(this.getReactNativeLogsDir())
+      }
+
+      // If daily log file is not present create it
+      if (!(await RNFS.exists(logFilePath))) {
```
silasbw, line 272: why check for existence?  i don't think we need to explicitly create the file (I think `append` will do that automatically for us).

src/utils/Logger.ts
```diff
@@ -146,15 +148,62 @@ class Logger {
...
+        await RNFS.unlink(logFileCombinedPath)
+      }
+
+      // Get the list of log files
+      const logFiles = await RNFS.readDir(logDir)
+
+      // Delete log files older than 28 days
+      if (logFiles.length > 0) {
+        logFiles.forEach((file) => {
+          RNFS.stat(file.path)
```
silasbw, line 184: why not use `await` here?

src/utils/Logger.ts
```diff
@@ -146,15 +148,62 @@ class Logger {
...
+                  console.warn('Failed to delete React Native logs file: ' + err)
+                })
+              }
+            })
+            .catch((err) => {
+              console.warn('Failed get log file stat: ' + err)
+            })
+        })
+      }
+    } catch (e) {
```
silasbw, line 198: you use "err" in some places for your error and "e" here.  Can you be consistent?

src/utils/Logger.ts
```diff
@@ -146,15 +148,62 @@ class Logger {
...
+      if (await RNFS.exists(logFileCombinedPath)) {
+        await RNFS.unlink(logFileCombinedPath)
+      }
+
+      // Get the list of log files
+      const logFiles = await RNFS.readDir(logDir)
+
+      // Delete log files older than 28 days
+      if (logFiles.length > 0) {
+        logFiles.forEach((file) => {
```
silasbw, line 183: if you use a `for .. of` you can use `await` more easily.

src/utils/Logger.ts
```diff
@@ -182,7 +262,17 @@ class Logger {
       warn: console.info,
     }
 
-    const writeLog = (level: string, message: string) => {
+    const writeLog = async (level: string, message: string) => {
+      // If log folder not present create it
+      if (!(await RNFS.exists(this.getReactNativeLogsDir()))) {
+        await RNFS.mkdir(this.getReactNativeLogsDir())
```
silasbw, line 268: this could throw, it's a toctou bug.  you should handle the error, or have this code as part of log initialization (where's it's guaranteed to be serialized)

silasbw (commented):
nice!

src/utils/Logger.ts
MuckT: When I try to append without explicting creating the file I get the following error:
```
 Failed to write to .../Library/Caches/rn_logs/20220714.txt
 ```

File: src/utils/Logger.ts (hunk 11 above)
Comment (MuckT) [reply src/utils/Logger.ts:268]: The problem I am seeing testing the change with moving it to the log initialization, out of `writeLog` up into the `overrideConsoleLogs` call, is that we miss some of the early logs from the the app start.

Comment (MuckT) [reply src/utils/Logger.ts:268]: The first line in the daily logs should be `Reactotron Configured` from `index.js` if it's not that line we're missing logs from app start. This can be tested by creating a new app install locally and navigating to developer settings  and selecting debug.

File: src/utils/Logger.ts (hunk 10 above)
Comment (MuckT) [reply src/utils/Logger.ts]: Thanks for the recommend; it really cleaned up the code! Changed in https://github.com/valora-inc/wallet/pull/2680/commits/f8b84262e339c90deeee4cbaad1d5bf7742f9960.

File: src/utils/Logger.ts (hunk 9 above)
Comment (MuckT) [reply src/utils/Logger.ts]: Fixed in https://github.com/valora-inc/wallet/pull/2680/commits/a86a7a7613ba7a029a9c89fad8917b8030bdbe45!

File: src/utils/Logger.ts (hunk 8 above)
Comment (MuckT) [reply src/utils/Logger.ts]: Changed in https://github.com/valora-inc/wallet/commit/f8b84262e339c90deeee4cbaad1d5bf7742f9960!

**Comment by MuckT** [comment]
The error seen in [Valora spinning then crashing when sending to email client #74](https://github.com/valora-inc/support-issues/issues/74 ) to my best guess was due to a crash in react-native-mail due to large attachments 500 MB+. In testing on this branch I was recreate a similar error from `RNFS.readFile` in `combineLogs` where large daily log files when being read into the combined log file would cause an OOM crash. This was only happening when the daily log files were exceeding 18MB each or a combined log file of ~500MB, so it's unclear if this represents a real world scenario. 

Regardless `await RNFS.appendFile(combinedLogsPath, await RNFS.readFile(logFile.path), 'utf8')` Should be changed to a method which chunks the files into the combined log file instead of reading them in entirety to the combined log file.

File: src/utils/Logger.ts (hunk 1 above)
Comment (MuckT) [reply src/utils/Logger.ts]: @jeanregisser I did end up creating log files for each day and combining them; however, given the performance hit of combining several files, can take upwards of 25 seconds, maybe this is a good path to go if we don't implement https://github.com/valora-inc/wallet/issues/2701.

**Comment by MuckT** [comment]
@Jean ended up going with the option from https://github.com/valora-inc/wallet/pull/2680#discussion_r917691372 as creating the combined logs was taking sometimes upwards of 25 seconds. The only thing left should be to cleanup the typings in [src/account/SupportContact.tsx](https://github.com/valora-inc/wallet/pull/2680/commits/8bd70d0e8a85725f6c352f3bf85c615a63a7c78b#diff-bf6399789c16c48bed3415ef4eee01f7a7a2209ca726c3ad065dcc7618573a5d).

File: src/utils/Logger.ts [hunk 12]
```diff
@@ -182,13 +256,28 @@ class Logger {
...
+        }
+
+        // If daily log file is not present create it
+        if (!(await RNFS.exists(logFilePath))) {
+          await RNFS.writeFile(logFilePath, '', 'utf8')
         }
-      )
+
+        const timestamp = new Date().toISOString()
+        // Ensure messages are converted to utf8 as some remote CTA's can have non utf8 characters
```
silasbw, line 272: kinda a nit: i'm not sure i understand this comment.  do you mean CTAs have non-ascii characters (i.e., unicode ones), so we must encode using utf-8?

src/utils/readFile.ts
```diff
@@ -0,0 +1,35 @@
...
+import Logger from 'src/utils/Logger'
+
+/**
+ * Reads and returns a portion of the input file.
+ */
+const readChunk = async (file: string, length: number, position: number) => {
+  try {
+    return await RNFS.read(file, length, position, 'utf8')
+  } catch (error) {
+    Logger.error('Logger@readChunk', 'readChunk error', error as Error)
```
silasbw, line 11: why swallow the error (instead of bubbling it up)?

src/utils/readFile.ts
```diff
@@ -0,0 +1,35 @@
...
+    let chunk = ''
+    let fileContents = ''
+    if (await RNFS.exists(rnLogsSrc)) {
+      do {
+        chunk = await readChunk(rnLogsSrc, length, fileContents.length)
+        fileContents += chunk
+      } while (chunk.length > 0)
+    }
+    return fileContents
+  } catch (error) {
```
silasbw, line 32: why swallow the error instead of bubbling it up?

src/utils/Logger.ts
silasbw: Can you fix the comment?

MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2680/commits/82a3162bcb042760961e008c0132769d04f6ba48!

src/utils/readFile.ts
MuckT: Fixed in [82a3162](https://github.com/valora-inc/wallet/pull/2680/commits/82a3162bcb042760961e008c0132769d04f6ba48)!

src/utils/readFile.ts
silasbw: Couldn't you just re-throw the error here?  You have the log message and you'll break out of your do while.  It seems odd to swallow the error in low-level file reading code.  Shouldn't higher-level code handle the error and decide what to do with it?

silasbw: Also, is `.error` the right level here?  Seems or should it be `.warn`?  `.error` implies a series problem we didn't expect and we should investigate immediately, but it looks like we might expect this?

src/utils/Logger.ts
```diff
@@ -146,30 +151,99 @@ class Logger {
...
+      const path = Platform.OS === 'ios' ? RNFS.TemporaryDirectoryPath : RNFS.ExternalDirectoryPath
+      const toAttach = []
+      for (const file of logFiles) {
+        // Always remove the logs from the current month and re-copy
+        // But not the logs from previous months if they exist
+        if (file.name === `${format(new Date(), 'yyyy-MMM')}.txt`) {
+          await RNFS.unlink(`${path}/${file.name}`)
+          await RNFS.copyFile(file.path, `${path}/${file.name}`)
+        } else if (!(await RNFS.exists(`${path}/${file.name}`))) {
+          await RNFS.copyFile(file.path, `${path}/${file.name}`)
```
silasbw, line 175: Why do we need to do this file copy stuff?  Why not attempt to attach the files directly?

MuckT: It's to do file permissions. We need to move them to a directory the mail client would have access to. I know this is the case for Android, but maybe we can omit for iOS. @jeanregisser

src/utils/Logger.ts
```diff
@@ -146,30 +151,99 @@ class Logger {
...
+  getLogsToAttach = async () => {
+    try {
+      const logDir = this.getReactNativeLogsDir()
+      const logFiles = await RNFS.readDir(logDir)
+      const path = Platform.OS === 'ios' ? RNFS.TemporaryDirectoryPath : RNFS.ExternalDirectoryPath
+      const toAttach = []
+      for (const file of logFiles) {
+        // Always remove the logs from the current month and re-copy
+        // But not the logs from previous months if they exist
+        if (file.name === `${format(new Date(), 'yyyy-MMM')}.txt`) {
```
kathaypacific, line 171: maybe i am missing something, but what happens on the first of the month, do we then just have that one day's worth of logs? just catching up on the comments now, I see Jean suggested this approach but with one file per week - i wonder if this is better, so that we can attach 2 weeks worth of logs (so this would range 8 - 14 days worth of logs)

src/utils/Logger.ts
```diff
@@ -146,30 +151,99 @@ class Logger {
...
+        } else if (!(await RNFS.exists(`${path}/${file.name}`))) {
+          await RNFS.copyFile(file.path, `${path}/${file.name}`)
+        }
+        toAttach.push({
+          path: `${path}/${file.name}`,
+          name: file.name,
+          type: 'text',
+        })
+      }
+      return toAttach as Email['attachments']
```
kathaypacific, line 183: do we need this type casting? i think generally we should let Typescript do its thing and keep the casting to a minimum...

kathaypacific (commented):
I think this looks good, just added a few questions ✨

src/utils/Logger.ts
MuckT: I've set the retention period of logs to 60 days and broken the logs into monthly logs. In worst case, August 30th, since the previous months logs would be 61 days old at that point we would have the entire month of August. In most cases we have have the current months logs and previous months logs.

src/utils/Logger.ts
MuckT: Removed in https://github.com/valora-inc/wallet/pull/2680/commits/b29d0f5af6789860f661f7b026a2369d64c1863a!

src/account/SupportContact.tsx
```diff
@@ -92,26 +72,21 @@ function SupportContact({ route }: Props) {
...
     }
-    setInProgress(false)
+    minSetInProgress(false)
     try {
-      await sendEmail(email, deviceInfo, logsPath)
+      await sendEmail(email, deviceInfo, attachments[0].path ?? false)
       navigateBackAndToast()
     } catch (error) {
-      Logger.error('SupportContact', 'Error while sending logs to support', error)
+      Logger.error('SupportContact', 'Error while sending logs to support', error as Error)
```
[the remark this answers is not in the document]
MuckT: Removed in https://github.com/valora-inc/wallet/pull/2680/commits/b29d0f5af6789860f661f7b026a2369d64c1863a!

src/account/SupportContact.tsx
```diff
@@ -92,26 +72,21 @@ function SupportContact({ route }: Props) {
...
+      if (attachments) {
+        email.attachments = attachments
         email.body += (email.body ? '<br/><br/>' : '') + '<b>Support logs are attached...</b>'
       }
     }
-    setInProgress(false)
+    minSetInProgress(false)
     try {
-      await sendEmail(email, deviceInfo, logsPath)
+      await sendEmail(email, deviceInfo, attachments[0].path ?? false)
```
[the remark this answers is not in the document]
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2680/commits/bdc449dd1fb154fef1aec1600748d52dc88bde73. The Nullish coalescing operator (??) should take care of it now that it's checking the result of a find. If it returns undefined the log path would return false which is what is expected from `sendEmail` in `src/account/emailSender.ts` as far as I can tell.

src/account/SupportContact.tsx
```diff
@@ -92,26 +72,21 @@ function SupportContact({ route }: Props) {
       body: `${message}<br/><br/><b>${JSON.stringify(deviceInfo)}</b>`,
       isHTML: true,
     }
-    let logsPath: string | false = false
+    // TODO(Tom): use correct typing
```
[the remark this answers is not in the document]
MuckT: Removed in https://github.com/valora-inc/wallet/pull/2680/commits/b29d0f5af6789860f661f7b026a2369d64c1863a! it does still accept `any` which might not be ideal; I am open to suggestions!

src/utils/Logger.ts
MuckT: Testing with iOS the above was also needed to have access to the files.

src/account/SupportContact.tsx
```diff
@@ -92,23 +72,24 @@ function SupportContact({ route }: Props) {
       body: `${message}<br/><br/><b>${JSON.stringify(deviceInfo)}</b>`,
       isHTML: true,
     }
-    let logsPath: string | false = false
+    let attachments: Array<Email['attachments']> | any
```
kathaypacific, line 75: shouldn't this be `Array<Email['attachments']> | undefined`?

src/account/SupportContact.tsx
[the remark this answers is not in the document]
kathaypacific: shouldn't it be `Array<Email['attachments']> | undefined`?

src/account/SupportContact.tsx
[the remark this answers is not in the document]
kathaypacific: i was thinking about the `attachments` actually...if `attachLogs` is false, then we never set the value of `attachments` yet we still hit this line and end up doing `undefined.find(....)`

src/account/SupportContact.tsx
```diff
@@ -73,6 +48,11 @@ function SupportContact({ route }: Props) {
     dispatch(showMessage(t('contactSuccess')))
   }
 
+  // Used to prevent flickering of the activity indicator on quick uploads
+  const minSetInProgress = (setTo: boolean) => {
```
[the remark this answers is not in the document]
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2680/commits/e56767daf37f142bb7700551dc0030371b1212ce!

src/account/SupportContact.tsx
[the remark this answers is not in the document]
MuckT: Added a fix for this in https://github.com/valora-inc/wallet/pull/2680/commits/ba6a33de8d3cfb5c38bf7e32c1f2a5e21c605e86!

src/account/SupportContact.tsx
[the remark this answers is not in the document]
MuckT: Was still getting some type errors with that 🤔.

src/account/SupportContact.tsx
```diff
@@ -92,23 +67,27 @@ function SupportContact({ route }: Props) {
       body: `${message}<br/><br/><b>${JSON.stringify(deviceInfo)}</b>`,
       isHTML: true,
     }
-    let logsPath: string | false = false
+    let attachments: Array<Email['attachments']> | any
```
kathaypacific, line 70: i have some ideas about how to resolve the TS errors without `any`
1. this type should be `Email['attachments']` right? because `Email['attachments']` is already an array?
2. the return type of `Logger.getLogsToAttach()` should be `Email['attachments']` i think, rather than ``Email['attachments']` | undefined` - to do this i think you can bring the initialisation of `toAttach` in `getLogsToAttach` to the top line, and always return this value (i.e. return outside of the try/catch). so if there is an error, then an empty array is returned (which i think is not wrong)
3. if this is all fine, the we don't need to check for `if(attachments)` anymore, since it will always be an array
4. then in the `sendEmail` argument, we can have `attachments.find(
          (attachment: { name: string }) => attachment.name === Logger.getCurrentLogFileName()
        )?.path ?? false`

MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2680/commits/ded87a8f61a5eeb9e9da0df52cc1691b823478c7!

ValoraQA:
Hey @MuckT, We have verified above issue on Latest Android Internal Release Build V1.37.0 & updated status in test rails.

**Observations:**
- User is able to see email id fields pre-populated and also logs file is seen attached

**Test rail link:**
**Android 12:** https://valoraapp.testrail.io/index.php?/tests/view/54640

**Devices:** Google Pixel 4a (12.0)

Thanks..!!

ValoraQA:
Hey @MuckT, We have verified above issue on Latest  iOS Test Flight Release build V1.37.0 & updated status in test rails.

**Observations:**

- User is able to see email id fields pre-populated and also logs.txt file is seen attached.

**Test rail link:**
**iOS 15:** https://valoraapp.testrail.io/index.php?/tests/view/54670

Thanks..!!

PR #2687, opened 2022-07-13
feat(fiatconnect): implement transfer out

satish-ravi:
### Description

Implement transfer out saga and invoke on clicking the "Withdraw" button on the review screen

### Other changes

Refactored send saga to include a helper function to build and send tx, which is also used to send crypto to provider on transfer outs.

### Tested

Unit tests

### How others should test

N/A

### Related issues

- Part of [ACT-277](https://linear.app/valora/issue/ACT-277/initiate-transfer-out-implementation)

### Backwards compatibility

n/a


src/fiatconnect/index.ts
```diff
@@ -191,3 +193,12 @@ export function getObfuscatedAccountNumber(accountNumber: string): string {
   const digitsToReveal = Math.max(0, Math.min(accountNumber.length - 3, 4))
   return digitsToReveal > 0 ? '...' + accountNumber.slice(-digitsToReveal) : ''
 }
+
+export async function doTransferOut(fiatConnectQuote: FiatConnectQuote, fiatAccountId: string) {
```
kathaypacific, line 197: any reason you chose to abstract this function here? maybe there are future plans to use this elsewhere that i'm not aware of :) was just thinking that this would be better as a generator function in fiatconnect/saga file since it is being used only in a saga for now?

kathaypacific (approved):
LGTM!! i think this is great :D

src/fiatconnect/saga.ts
```diff
@@ -50,6 +68,67 @@ export function* watchFetchFiatConnectQuotes() {
...
+      Logger.error(TAG, 'Transfer out failed..', err)
+      yield put(fiatConnectTransferFailed({ flow, quoteId }))
+    }
+  } else {
+    throw new Error('not implemented')
+  }
+}
+
+function* watchFiatConnectTransactions() {
+  yield takeEvery(startFiatConnectTransfer.type, handleStartFiatConnectTransfer)
```
jh2oman, line 128: `takeLeading` stops listening after it sees `startFiatConnectTransfer.type` until `handleStartFiatConnectTransfer`. Would it maybe be good to use that for some built in rate limiting?

src/fiatconnect/slice.ts
```diff
@@ -53,13 +81,44 @@ export const slice = createSlice({
...
+        flow: action.payload.flow,
+        isSending: true,
+        failed: false,
+        txHash: null,
+      }
+    },
+    fiatConnectTransferFailed: (state, action: PayloadAction<FiatConnectTransferFailedAction>) => {
+      state.transfer = {
+        quoteId: action.payload.quoteId,
+        flow: CICOFlow.CashOut,
```
jh2oman, line 96:
```suggestion
        flow: action.payload.flow,
```

src/fiatconnect/slice.ts
```diff
@@ -53,13 +81,44 @@ export const slice = createSlice({
...
+        txHash: null,
+      }
+    },
+    fiatConnectTransferSuccess: (
+      state,
+      action: PayloadAction<FiatConnectTransferSuccessAction>
+    ) => {
+      state.transfer = {
+        quoteId: action.payload.quoteId,
+        flow: CICOFlow.CashOut,
```
jh2oman, line 108:
```suggestion
        flow: action.payload.flow,
```

src/fiatconnect/slice.ts
```diff
@@ -53,13 +81,44 @@ export const slice = createSlice({
...
+      }
+    },
   },
   extraReducers: (builder) => {
     builder.addCase(REHYDRATE, (state, action: RehydrateAction) => ({
       ...state,
       ...getRehydratePayload(action, 'fiatConnect'),
       quotesLoading: false,
       quotesError: null,
+      // transfer: null,
```
jh2oman, line 121: Why is this here and commented out?

satish-ravi: oops, this should be uncommented. We'd want to reset/clean any in progress transfer state if the app reloads right?

jh2oman: yep 👍

jh2oman (approved):

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -55,7 +57,16 @@ export default function FiatConnectReviewScreen({ route, navigation }: Props) {
         onPress={() => {
           ValoraAnalytics.track(FiatExchangeEvents.cico_submit_transfer, { flow })
 
-          // TODO(any): submit the transfer
+          dispatch(
+            startFiatConnectTransfer({
+              flow,
+              fiatConnectQuote: normalizedQuote,
+              quoteId: normalizedQuote.getQuoteId(),
```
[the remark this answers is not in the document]
satish-ravi: yeah, it was just there more for convenience, but with #2691 I think we can remove the quote class, so I'll refactor this after merging #2691

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -55,7 +57,16 @@ export default function FiatConnectReviewScreen({ route, navigation }: Props) {
...
+              fiatAccountId: fiatAccount.fiatAccountId,
+            })
+          )
+
+          // TODO: navigate to success / failure screen
```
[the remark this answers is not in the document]
satish-ravi: this will be a follow up to this PR by Joe.

src/fiatconnect/slice.ts
```diff
@@ -30,6 +40,24 @@ export interface FetchFiatConnectQuotesFailedAction {
...
+export interface FiatConnectTransferFailedAction {
+  flow: CICOFlow
+  quoteId: string
+}
+
+export interface FiatConnectTransferSuccessAction {
+  flow: CICOFlow
+  quoteId: string
+  txHash: string | null
+}
```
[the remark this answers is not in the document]
satish-ravi: hmm good point, but I can't think of a better name right now. I'll think about it.

src/fiatconnect/saga.ts
```diff
@@ -50,6 +68,67 @@ export function* watchFetchFiatConnectQuotes() {
...
 
+export function* handleStartFiatConnectTransfer({
+  payload: params,
+}: ReturnType<typeof startFiatConnectTransfer>) {
+  const { flow, quoteId, fiatConnectQuote, fiatAccountId } = params
+
+  if (flow === CICOFlow.CashOut) {
+    try {
+      Logger.info(TAG, 'Starting transfer out..')
+      const transferResult: TransferResponse = yield call(
```
jophish, line 79: were you able to get this working against the mock provider? the API call fails for me when testing locally. not a problem with this PR i believe, but perhaps with the SDK/provider itself.

src/fiatconnect/saga.ts
```diff
@@ -50,6 +68,67 @@ export function* watchFetchFiatConnectQuotes() {
...
+      Logger.info(
+        TAG,
+        'Transfer out $succeeded. Starting transaction..',
+        JSON.stringify(transferResult)
+      )
+
+      const tokenList: TokenBalance[] = yield select(tokensListSelector)
+      const tokenInfo = tokenList.find((token) => token.symbol === fiatConnectQuote.getCryptoType())
+
+      if (!tokenInfo) {
```
jophish, line 94: There's no situation where this ever ought to occur, right? Maybe add a comment clarifying this?

src/fiatconnect/saga.ts
```diff
@@ -50,6 +68,67 @@ export function* watchFetchFiatConnectQuotes() {
...
+        throw new Error(`Invalid token: ${fiatConnectQuote.getCryptoType()}`)
+      }
+
+      const feeEstimates: FeeEstimatesState['estimates'] = yield select(feeEstimatesSelector)
+      const feeInfo = feeEstimates[tokenInfo.address]?.[FeeType.SEND]?.feeInfo
```
jophish, line 99: Is it possible for this to be `undefined`? If so, should we guard against it and throw just in case, since we have a non-null assertion on it below?

src/fiatconnect/saga.ts
jophish: Does that mean that `startFiatConnectTransfer` actions that get dispatched while one is already running would get silently ignored? No risk of that breaking anything now (or probably ever), just wondering.

jh2oman: Yep, silently ignores other requests while handleStartFiatConnectTransfer is running

src/fiatconnect/slice.ts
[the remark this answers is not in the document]
cajubelt: maybe FiatConnectCreateTransferSuccess/Failed?

src/fiatconnect/saga.ts
satish-ravi: I don't think so, since withdraw amount page should do a fee estimation. But yeah we can do a similar check to tokenInfo

src/fiatconnect/index.ts
kathaypacific: good question! i don't think there are benefits actually (correct me if i'm wrong @jeanregisser ?) i was only mentioning it from the perspective that async functions that are only called by sagas can be thought of as sagas too so it's more clear to keep them in one place/format :)

jh2oman (approved):

jophish (approved):

ValoraQA:
Hey @MuckT  since NA is mentioned in how others should test , we are skipping this PR

cc: @satish-ravi 
Thanks.!

PR #2689, opened 2022-07-13
feat: support new supercharge tokens

jeanregisser:
### Description

### Other changes

### Tested

### How others should test

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

### DIFF
--- a/src/consumerIncentives/ConsumerIncentivesHomeScreen.tsx
+++ b/src/consumerIncentives/ConsumerIncentivesHomeScreen.tsx
@@ -12,14 +12,10 @@ import Dialog from 'src/components/Dialog'
 import Pill from 'src/components/Pill'
 import Touchable from 'src/components/Touchable'
 import { RewardsScreenCta } from 'src/consumerIncentives/analyticsEventsTracker'
+import { balanceInfoForSuperchargeSelector } from 'src/consumerIncentives/selectors'
 import { claimRewards, fetchAvailableRewards } from 'src/consumerIncentives/slice'
-import {
-  SuperchargePendingReward,
-  SuperchargeToken,
-  SuperchargeTokenConfig,
-} from 'src/consumerIncentives/types'
+import { SuperchargePendingReward, SuperchargeTokenConfig } from 'src/consumerIncentives/types'
 import { FiatExchangeFlow } from 'src/fiatExchanges/utils'
-import { WEI_PER_TOKEN } from 'src/web3/consts'
 import InfoIcon from 'src/icons/InfoIcon'
 import Logo, { LogoTypes } from 'src/icons/Logo'
 import Times from 'src/icons/Times'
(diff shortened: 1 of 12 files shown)


src/consumerIncentives/types.ts
```diff
@@ -1,11 +1,5 @@
-export enum SuperchargeToken {
-  cUSD = 'cUSD',
-  cEUR = 'cEUR',
-  cREAL = 'cREAL',
-}
-
 export interface SuperchargeTokenConfig {
-  token: SuperchargeToken
+  tokenSymbol: string
```
jeanregisser, line 2: Renamed for clarity.

src/consumerIncentives/ConsumerIncentivesHomeScreen.tsx
```diff
@@ -32,50 +29,34 @@ import useSelector from 'src/redux/useSelector'
...
-    ? SuperchargeToken.cREAL
-    : SuperchargeToken.cUSD
-  return (
-    superchargeTokens.find((token) => token.token === tokenToSupercharge) ?? {
-      token: tokenToSupercharge,
-    }
-  )
-}
+    ? 'cREAL'
+    : 'cUSD'
```
kathaypacific, line 51: should we mention the mcStables somewhere here?

src/consumerIncentives/selectors.ts
```diff
@@ -0,0 +1,60 @@
...
+  superchargeTokenConfigByTokenSelector,
+  (tokensByAddress, superchargeTokenConfigByToken): SuperchargeInfo => {
+    let superchargeInfo: SuperchargeInfo = { hasBalanceForSupercharge: false }
+
+    for (const [tokenAddress, superchargeTokenConfig] of Object.entries(
+      superchargeTokenConfigByToken
+    )) {
+      const tokenInfo = tokensByAddress[tokenAddress.toLowerCase()]
+      if (!tokenInfo || !tokenInfo.usdPrice) {
+        Logger.warn(`A possible supercharged token info couldn't be obtained from firebase`)
```
kathaypacific, line 28: not a big issue but i think this might get logged a lot with the super analytics prop - the chances of tokenInfo being falsey are low but the `usdPrice` on the tokenInfo is set to null when they become stale (12 hours) so potentially this could get logged on every app open if the user visits the app once a day. maybe we can move `!tokenInfo.usdPrice` to the next if statement that doesn't contain the logging?

kathaypacific (commented):
looks great!! i just had a couple questions...

src/consumerIncentives/selectors.ts
dievazqu: Yes, I thought this could be logged several times, I think I'm removing this log entirely.

kathaypacific (approved):

PR #2691, opened 2022-07-13
refactor(fiatconnect): move client creation outside quote class

satish-ravi:
### Description

There can be multiple quotes for the same provider (when quote expires or a transfer is retried), this allows reusing the same FiatConnectClient instance across quotes.

### Other changes

Moved `getSigningFunction` to src/fiatconnect/clients.tsx since it's only used here

### Tested

unit tests

### How others should test

N/A

### Related issues

- For [ACT-277](https://linear.app/valora/issue/ACT-277/initiate-transfer-out-implementation) and [ACT-349](https://linear.app/valora/issue/ACT-349/fc-return-user-flow)

### Backwards compatibility

Yes

### DIFF
--- a/src/fiatExchanges/quotes/FiatConnectQuote.ts
+++ b/src/fiatExchanges/quotes/FiatConnectQuote.ts
@@ -1,4 +1,3 @@
-import { FiatConnectApiClient } from '@fiatconnect/fiatconnect-sdk'
 import {
   FiatAccountSchema,
   FiatAccountType,
@@ -33,7 +32,6 @@ export default class FiatConnectQuote extends NormalizedQuote {
   fiatAccountType: FiatAccountType
   flow: CICOFlow
   quoteResponseFiatAccountSchema: QuoteResponseFiatAccountSchema
-  fiatConnectClient?: FiatConnectApiClient
 
   constructor({
     quote,
--- a/src/fiatExchanges/quotes/FiatConnectQuote.ts
+++ b/src/fiatExchanges/quotes/FiatConnectQuote.ts
@@ -1,3 +1,4 @@
+import { FiatConnectApiClient } from '@fiatconnect/fiatconnect-sdk'
 import {
   FiatAccountSchema,
   FiatAccountType,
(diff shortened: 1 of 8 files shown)


jh2oman (approved):
🙏

src/fiatExchanges/quotes/FiatConnectQuote.ts
```diff
@@ -1,3 +1,4 @@
+import { FiatConnectApiClient } from '@fiatconnect/fiatconnect-sdk'
```
jophish, line 1: Remove `fiatConnectClient` as an attribute from the class and get rid of this import?

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -177,7 +178,10 @@ const FiatDetailsScreen = ({ route, navigation }: Props) => {
 
       const fiatAccountSchema = quote.getFiatAccountSchema()
 
-      const fiatConnectClient = await quote.getFiatConnectClient()
+      const fiatConnectClient = await getFiatConnectClient(
+        quote.getProviderId(),
+        quote.getProviderBaseUrl()
```
jophish, line 183: When a client already exists, the `baseUrl` is unused in `getFiatConnectClient`, which feels slightly weird... not sure if there's anything to do about it though, since if it _doesn't_ exist and we don't pass the `baseUrl`, we'd have to throw.

jophish (approved):
Looks good overall! Just some tiny comments

src/fiatExchanges/quotes/FiatConnectQuote.ts
satish-ravi: good catch!

jh2oman (approved):

src/fiatconnect/FiatDetailsScreen.tsx
jophish: Should we be able to specify a client API key in `getFiatConnectClient` as well? Not sure exactly how the logic should work, but probably a good idea to be able to pass one through here.

satish-ravi: chatted offline. This is something we can add as a follow up once the CF starts supporting client API keys.

ValoraQA:

cc: @satish-ravi 
Thanks.!

PR #2699, opened 2022-07-15
feat(fiatconnect): a return user flow that has them skip selecting a provider and filling in fiat account details

jh2oman:
### Description

1. After a user successfully links a fiatAccount, save it as recently used. It is saved with its cryptoType, fiatType, and flow because thats the only combination that we can be *reasonably sure* it supports
2. When a user goes through the flow again with the same cryptoType, fiatType, and flow combo, we attempt to fetch a quote and fiatAccount from the saved provider and direct them straight to the FiatConnectReview screen

TODO: Both of these are marked in comments and will be done in a separate PR
1. Allow the user to select a different provider from the FiatConnectReview screen
2. On the SelectProviderScreen, if the user presses a provider that already has a saved FiatAccount that matches the quote, navigate the user to the FiatConnectReview Page

#### Example where user already connected a FiatAccount

https://user-images.githubusercontent.com/8432644/180626774-ee2f4ad4-c2ca-40e2-ab2c-01a6d6418a95.mp4

### Tested

Unit tests. It is difficult to fully test this flow without a fully working provider.

### How others should test

Does not need to be tested by QA at this time

### Related issues

https://linear.app/valora/issue/ACT-349/fc-return-user-flow


package.json
```diff
@@ -62,6 +62,7 @@
     ]
   },
   "dependencies": {
+    "@badrap/result": "^0.2.12",
```
cajubelt, line 65: since this is v0 we may not want the caret, since there could be breaking changes if the middle digit increments
```suggestion
    "@badrap/result": "~0.2.12",
```

src/fiatExchanges/FiatExchangeAmount.tsx
```diff
@@ -153,17 +155,33 @@ function FiatExchangeAmount({ route }: Props) {
...
-        crypto: inputCryptoAmount.toNumber(),
-        // Rounding up to avoid decimal errors from providers. Won't be
-        // necessary once we support inputting an amount in both crypto and fiat
-        fiat: Math.round(inputLocalCurrencyAmount.toNumber()),
-      },
-    })
+    const cryptoAmount = inputCryptoAmount.toNumber()
+    // Rounding up to avoid decimal errors from providers. Won't be
+    // necessary once we support inputting an amount in both crypto and fiat
+    const fiatAmount = Math.round(inputLocalCurrencyAmount.toNumber())
```
cajubelt, line 161: aren't these subject to floating point and/or rounding errors? Seems risky to do this casting

src/fiatExchanges/FiatExchangeAmount.tsx
```diff
@@ -84,6 +85,7 @@ function FiatExchangeAmount({ route }: Props) {
   const localCurrencyCode = useLocalCurrencyCode()
   const dailyLimitCusd = useSelector(cUsdDailyLimitSelector)
   const exchangeRates = useSelector(localCurrencyExchangeRatesSelector)
+  const mostRecentFiatAccount = useSelector(mostRecentFiatAccountSelector)!
```
cajubelt, line 88: i'm confused. doesn't the trailing ! mean you're asserting it's non-null? what if they haven't created a fiat account yet?

src/fiatExchanges/FiatExchangeAmount.tsx
```diff
@@ -153,17 +155,33 @@ function FiatExchangeAmount({ route }: Props) {
...
-        // necessary once we support inputting an amount in both crypto and fiat
-        fiat: Math.round(inputLocalCurrencyAmount.toNumber()),
-      },
-    })
+    const cryptoAmount = inputCryptoAmount.toNumber()
+    // Rounding up to avoid decimal errors from providers. Won't be
+    // necessary once we support inputting an amount in both crypto and fiat
+    const fiatAmount = Math.round(inputLocalCurrencyAmount.toNumber())
+
+    // Navigate straight to the review screen if the user has a FiatConnect fiat account saved
```
cajubelt, line 163: having a fiat account doesn't mean they'll have passed KYC

cajubelt: they might've even tried to cash in with the same provider and not needed to do kyc for the last transfer, but need to do it for this one

src/fiatExchanges/FiatExchangeAmount.tsx
cajubelt: also, I didn't realize this before but the trailing `!` doesn't result in any type checks taking place, it just turns off compile-time errors for nullable values. at runtime it could still be null and pass right by this line. terrifying

src/fiatconnect/slice.ts
```diff
@@ -16,30 +17,56 @@ export interface State {
   quotes: (FiatConnectQuoteSuccess | FiatConnectQuoteError)[]
   quotesLoading: boolean
   quotesError: string | null
+  fiatAccount: ObfuscatedFiatAccountData | null
+  fiatAccountLoading: boolean
+  fiatAccountError: string | null
+  mostRecentFiatAccountIds: FiatAccount[]
```
cajubelt, line 23: from the name I'd expect this to just contain fiat account id's, not other info about fiat accounts. also why "most recent" if it's a list? is there an eviction policy?

src/fiatconnect/slice.ts
```diff
@@ -73,13 +100,45 @@ export const slice = createSlice({
...
     },
-    fetchFiatConnectQuotesFailed: (
-      state,
-      action: PayloadAction<FetchFiatConnectQuotesFailedAction>
-    ) => {
+    fetchFiatConnectQuotesFailed: (state, action: PayloadAction<FetchFailedAction>) => {
       state.quotesLoading = false
       state.quotesError = action.payload.error
     },
+    fiatAccountUsed: (state, action: PayloadAction<FiatAccountAddedAction>) => {
```
cajubelt, line 107: why are you removing the fiat account when it's used?

cajubelt (commented):
reviewed up thru ReviewFetchScreen. Submitting comments since I think they probly need addressing and I don't want to block them on my reading thru the rest

src/fiatconnect/slice.ts
cajubelt: seems like you might want to delete it and then add it to the front

cajubelt: WHOOPS that's what you're doing lol

src/fiatconnect/slice.ts
cajubelt: nvm seeing that now

src/fiatExchanges/FiatExchangeAmount.tsx
jh2oman: Yes, we have a ticket to go through the CICO flow and fix these https://app.zenhub.com/workspaces/acquisition-squad-sprint-board-6010683afabec1001a090887/issues/valora-inc/wallet/2461

src/fiatconnect/saga.ts
```diff
@@ -55,16 +65,74 @@ export function* handleFetchFiatConnectQuotes({
...
     yield put(fetchFiatConnectQuotesCompleted({ quotes }))
   } catch (error) {
-    Logger.error(TAG, 'Could not parse dapps response', error)
     yield put(fetchFiatConnectQuotesFailed({ error: 'Could not fetch providers' }))
   }
 }
 
-export function* watchFetchFiatConnectQuotes() {
-  yield takeLeading(fetchFiatConnectQuotes.type, handleFetchFiatConnectQuotes)
+export function* handleFetchQuoteAndFiatAccount({
```
cajubelt, line 76: I find this saga pretty complex. It seems like it's doing several things:
- checking that a user's fiat account ID is valid
- (if it's not, delete the fiat account)
- gets a quote
- checks that the quote is compatible with the given fiat account
- save the quote

To make it easier to reason about and debug, I'd suggest simplifying and making it more single-purpose. 

Ideas for a refactor:
- If you want to be defensive and do the fiat account ID check before attempting to create a transfer, you could try doing that before you even enter the ReviewFetchScreen component and take the fiat account ID as a prop (after all, you never actually want to enter that component unless there's a valid fiat account to use). 
- probably more effortful: migrate most/all of this to the backend. You're doing quite a lot of state management and API calls, which might be better organized into a single API such as `getMRUAccountQuote`, something like that. The difficulty could be checking whether the fiat account hasn't been deleted, since the user needs to be authenticated to call GET /accounts, but that seems like an edge case-- and one that could still be done separately in the client if we wanted.

cajubelt: one unfortunate thing is that KYC is left out of this, which will likely add complexity. If anything that makes it more important to consider how to simplify now, though

cajubelt (commented):
some feedback on the high level design

cajubelt:
relatedly, seems like this is just part of a returning user flow, since we'll also need to consider the case where a user reaches the SelectProviders screen and chooses a provider that the user already has an account with, or has completed KYC with in the past.

cajubelt:
this todo seems lower pri:

> Allow the user to select a different provider from the FiatConnectReview screen

jh2oman:
@cajubelt That TODO is required for a user to be able to change change which provider they want to use for a given fiatType/cryptoType/flow combo once they've added a FiatAccount.

jh2oman:
@cajubelt To clarify, the TODO might be better written as
"Allow the user to navigate to the SelectProvider screen from the FiatConnectReview screen so they may switch providers if they desire"

src/fiatconnect/saga.ts
```diff
@@ -69,17 +80,116 @@ export function* handleFetchFiatConnectQuotes({
...
-export function* watchFetchFiatConnectQuotes() {
-  yield takeLeading(fetchFiatConnectQuotes.type, handleFetchFiatConnectQuotes)
+    // Successfully found quote and fiatAccount
+    yield put(attemptReturnUserFlowCompleted())
+    navigate(Screens.FiatConnectReview, {
+      flow,
+      normalizedQuote,
+      fiatAccount,
+    })
+  } catch (error) {
```
satish-ravi, line 154: Should we handle transient errors like n/w failure and provider / fiat account mismatch errors differently? Seems like at a minimum we'd want to log warn / errors for n/w errors.

src/fiatconnect/slice.ts
```diff
@@ -83,13 +119,24 @@ export const slice = createSlice({
...
-    ) => {
+    fetchFiatConnectQuotesFailed: (state, action: PayloadAction<FetchFailedAction>) => {
       state.quotesLoading = false
       state.quotesError = action.payload.error
     },
+    fiatAccountUsed: (state, action: PayloadAction<FiatAccountUsedAction>) => {
+      state.recentlyUsedFiatAccounts = [
+        action.payload,
+        ...state.recentlyUsedFiatAccounts.filter((fiatAccount) =>
+          isEqual(fiatAccount, action.payload)
```
satish-ravi, line 130: should this be not equal?

src/fiatconnect/saga.ts
```diff
@@ -69,17 +80,116 @@ export function* handleFetchFiatConnectQuotes({
...
+    if (!normalizedQuote) {
+      throw new Error('Could not find quote')
+    }
+    // Get the provider info
+    const fiatConnectProvider = fiatConnectProviders?.find((provider) => provider.id === providerId)
+    if (!fiatConnectProvider) {
+      throw new Error('Could not find provider')
+    }
+    // Fetch Fiat Account associated with the cached providerId / fiatAccountId
+    const fiatAccounts: FiatAccount[] = yield call(
```
satish-ravi, line 137: should this be invoked in parallel? Also we'd also want this if we don't have a provider stored or if a user has used two providers? We would need to do this in the select provider screen after provider selection before fiat details screen right? E.g.,
- cash out first time with Provider A
- cash out shows A, but user selects Provider B 
- cash out shows B, but user selects A --- this should go to the Review screen directly? since there's a stored fiat account?

src/fiatconnect/slice.ts
```diff
@@ -22,6 +28,8 @@ export interface State {
   quotesError: string | null
   transfer: FiatConnectTransfer | null
   providers: FiatConnectProviderInfo[] | null
+  recentlyUsedFiatAccounts: CachedFiatAccountUse[]
+  attemptReturnUserFlowLoading: boolean
```
satish-ravi, line 32: should this be set to false on rehydrate?

test/RootStateSchema.json
```diff
@@ -710,6 +742,14 @@
             ],
             "type": "string"
         },
+        "FiatAccountType": {
+            "enum": [
+                "BankAccount",
```
satish-ravi, line 747: just for my understanding, if there are more values for this enum in future, would it require a schema migration?

src/fiatconnect/saga.ts
jh2oman: > should this be invoked in parallel

It can be yes! I can add that in.

jh2oman: For the second part of your comment I'm a bit confused by it, but I was thinking for the SelectProviders screen that I'd fetch fiatAccounts for every quote that came back and to not rely on the cache for this part. 

Fetching featAccounts before click on the SelectProvider screen has the advantage of us being able to label which ones have already been linked on the SelectProvider screen which has been talked about in a few designs but I think was missed in the latest round.

src/fiatconnect/saga.ts
jh2oman: What would be the goal of making them errors? How would you suggest we categorize the different failure types?

test/RootStateSchema.json
jh2oman: I am not sure if this would need a schema migration. I'm not sure what we'd "migrate" if more enums were added. @jeanregisser do you know?

src/fiatconnect/slice.ts
jh2oman: good catch

src/fiatconnect/slice.ts
jh2oman: yep!

src/fiatconnect/saga.ts
satish-ravi: hmm maybe we can track them? in sentry for example? If we want to track provider request success rate, etc.
We can use a custom error for errors thrown in the try block and do different actions.

jh2oman: I thought that one of the directions we were going with Sentry was to *not* use it for transient errors like a partner failure?

src/fiatconnect/slice.ts
```diff
@@ -22,6 +28,8 @@ export interface State {
   quotesError: string | null
   transfer: FiatConnectTransfer | null
   providers: FiatConnectProviderInfo[] | null
+  recentlyUsedFiatAccounts: CachedFiatAccountUse[]
```
satish-ravi, line 31: nit: Seems like we use 3 different names for this `cachedFiatAccount`, `cachedFiatAccountUse`, `recentlyUsedFiatAccount`. Can we just use once for state item, selector and state item type

src/fiatconnect/saga.ts
satish-ravi: chatted offline, current implementation is good.

satish-ravi (approved):

src/fiatconnect/saga.ts
satish-ravi: yeah I was just asking about the case when a user has used multiple providers. In that case we have to make the call to get accounts before showing the fiat details screen right? I am good If it's planned to be in a follow up PR

ValoraQA:
Hey @MuckT since NA is mentioned in how others should test section ,we are skipping this PR 

CC: @jh2oman 

Thanks.!

PR #2702, opened 2022-07-16
feat(exchanges): only show crypto exchanges card if available

awtsui:
### Description

Calling getExchanges, a cloud function, in SelectProvider to determine if any exchanges are available. In the event that there are no available exchanges, the exchange card on Select Payment Method screen is removed.

### Other changes

N/A

### Tested

Added mockExchanges to test/values to mock fetchExchanges. 
Tested fetchExchanges
Tested for missing exchange card if no available exchanges

ExternalExchanges Screen will always have providers passed in via props

### How others should test

N/A

### Related issues

- Fixes #[299]
https://linear.app/valora/issue/ACT-299/only-show-crypto-exchanges-if-they-are-actually-available

Splitting this issue into 2 PRs

### Backwards compatibility

N/A


src/fiatExchanges/SelectProvider.tsx
```diff
@@ -82,6 +83,20 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
     }
   }, [fiatConnectQuotesError])
 
+  const asyncExchanges = useAsync(async () => {
```
jophish, line 86: We probably want to wait for `asyncExchanges` to finish loading before removing the spinner and displaying the page; see line 132 for how we do this with `asyncProviders` currently.

src/fiatExchanges/SelectProvider.tsx
```diff
@@ -156,7 +173,11 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
         cryptoAmount={route.params.amount.crypto}
         coinbaseProvider={coinbaseProvider}
       />
-      <ExchangesSection selectedCurrency={route.params.selectedCrypto} flow={flow} />
+      <ExchangesSection
+        numExchanges={exchanges?.length}
```
jophish, line 177: Since we're actually fetching all the exchanges here, we should probably avoid fetching them again elsewhere. With this implementation, we fetch exchanges once here, and then, if a user actually browses to the Exchanges screen by clicking the card (assuming some exchanges are available), the list of exchanges will be fetched again. I would consider passing the actual exchange information (rather than just the number of exchanges) down into the `ExchangesSection` component, and further into the `ExternalExchanges` screen itself, and remove the `fetchExchanges` call from within `ExternalExchanges`.

src/fiatExchanges/ExternalExchanges.test.tsx
```diff
@@ -23,58 +21,18 @@ const mockStore = createMockStore({
...
-  },
-  {
-    name: 'OKCoin',
-    link: 'https://www.okcoin.com/en/spot/trade/cusd-usd/',
-    currencies: [Currency.Celo, Currency.Dollar],
-    supportedRegions: ['global'],
-  },
-]
-
+// Removed mockFetch testing as fetchExchanges is called in SelectProvider
```
jh2oman, line 24: This comment is unnecessary. Its referencing code that will cease to exist as soon as the PR is merged

src/fiatExchanges/ExternalExchanges.test.tsx
```diff
@@ -23,58 +21,18 @@ const mockStore = createMockStore({
...
-      </Provider>
-    )
-
-    await flushMicrotasksQueue()
-
-    expect(tree.getByTestId('NoExchanges')).toBeTruthy()
-    expect(tree.getByTestId('SwitchCurrency')).toBeTruthy()
-    expect(tree.getByTestId('ContactSupport')).toBeTruthy()
-  })
+  // ExternalExchanges Screen will always have providers given restructuring of flow
```
jh2oman, line 30: When leaving comments, its best to just talk about the current state of the code unless its crucial for understanding why things were done the way they were. For example, if it had to be done a certain way for backwards compatibility. 
```suggestion
  // ExternalExchanges Screen will always have providers passed in via props
```

src/fiatExchanges/ExternalExchanges.test.tsx
```diff
@@ -23,58 +21,18 @@ const mockStore = createMockStore({
...
   it('makes a request to fetch a list of available exchanges based on geolocation and selected currency', async () => {
```
jh2oman, line 31: This text should be updated since there is no longer any fetching being done

jh2oman (approved):
Great work!

jophish (approved):
Awesome work :slightly_smiling_face:

ValoraQA:

Cc: @awtsui 

Thanks!

PR #2707, opened 2022-07-18
feat(exchanges): show no payment methods screen if no providers and exchanges are available

awtsui:
### Description

Show the existing screen in app that says there are no payment options and invites them to switch their currency to the user if there are 0 payment methods.

### Other changes

N/A

### Tested

Added a test to SelectProvider.test.tsx to check for component visibility
CBPay card's visibility logic is moved to SelectProvider, and thus a majority of CoinbasePaymentSection tests have been migrated to SelectProvider. 

### How others should test

N/A

### Related issues

- Fixes #[299]
https://linear.app/valora/issue/ACT-299/only-show-crypto-exchanges-if-they-are-actually-available

### Backwards compatibility


locales/base/translation.json
```diff
@@ -354,6 +354,7 @@
   "restrictedRegion": "Limited support in region",
   "unsupportedPaymentMethod": "Paying with a {{paymentMethod}} is not supported in your region",
   "noProviders": "There are no providers available for {{digitalAsset}} in your region.",
+  "noPaymentMethods": "There are no payment methods available for {{digitalAsset}} in your region.",
```
anthonydaoud, line 357: Wouldn't you also need to translate this line and add it to the various translation files?

src/fiatExchanges/SelectProvider.tsx
```diff
@@ -158,39 +161,80 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
...
+      flow:
+        route.params.flow === CICOFlow.CashIn ? FiatExchangeFlow.CashIn : FiatExchangeFlow.CashOut,
+    })
+
   const exchanges = asyncExchanges.result ?? []
+  const legacyMobileMoneyProviders = asyncProviders.result?.legacyMobileMoneyProviders
+
+  const anyProviders =
+    normalizedQuotes.length ||
+    !(coinbaseProvider?.restricted ?? true) ||
```
anthonydaoud, line 177: I think you need to add more checks to determine if the Coinbase pay section will show up. [These](https://github.com/valora-inc/wallet/blob/awtsui/no-payment-option/src/fiatExchanges/CoinbasePaymentSection.tsx#L39) are that checks that are currently made in the Coinbase payment section

anthonydaoud: The code looks good to me otherwise!

awtsui: Yup, I agree. Coinbase pay section checks have been moved to SelectProvider.

locales/base/translation.json
cajubelt: @anthonydaoud that will be done by our localization team!

cajubelt: (separately from this PR)

src/fiatExchanges/SelectProvider.test.tsx
```diff
@@ -126,6 +137,7 @@ describe(SelectProviderScreen, () => {
     await waitFor(() => expect(fetchExchanges).toHaveBeenCalledWith('MX', Currency.Dollar))
   })
   it('shows the provider sections, mobile money, and exchange section', async () => {
+    mocked(readOnceFromFirebase).mockResolvedValue(FAKE_APP_ID)
```
cajubelt, line 140: seems like this is the same mock behavior in every test. best to put it in a `beforeEach` or (probably better) attach it to the jest mock function, e.g. 
```
jest.mock('src/firebase/firebase', () => ({
  readOnceFromFirebase: jest.fn().mockResolvedValue(FAKE_APP_ID),
}))
```

cajubelt:
would love to also see a screenshot of when there are providers, but none are exchanges (in which case there shouldn't be an exchanges section in the Select Providers screen)

src/fiatExchanges/SelectProvider.tsx
```diff
@@ -158,39 +166,88 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
...
+            cryptoAmount={route.params.amount.crypto}
+            coinbaseProvider={coinbaseProvider}
+            visible={coinbasePayVisible}
+            appId={appId}
+          />
+          <ExchangesSection
+            exchanges={exchanges}
+            selectedCurrency={route.params.selectedCrypto}
+            flow={flow}
+          />
```
cajubelt, line 224: what if there are no exchanges, but there are other kinds of providers? You got rid of the behavior where the message appears saying there aren't any options. Wasn't this section supposed to be hidden now if there were no exchanges?

src/fiatExchanges/SelectProvider.test.tsx
```diff
@@ -31,6 +35,10 @@ jest.mock('@coinbase/cbpay-js', () => {
   return { generateOnRampURL: jest.fn() }
 })
 
+jest.mock('src/firebase/firebase', () => ({
```
cajubelt, line 38: not sure this is necessary if you're using `mocked(readOnceFromFirebase)`

cajubelt (commented):
see comments

src/fiatExchanges/ExternalExchanges.tsx
```diff
@@ -62,41 +60,14 @@ function ExternalExchanges({ route }: Props) {
...
-    })
-
   const goToCashOut = () => {
     navigate(Screens.WithdrawCeloScreen, { isCashOut: true })
     ValoraAnalytics.track(FiatExchangeEvents.cico_celo_exchange_send_bar_continue)
   }
 
   return (
     <SafeAreaView style={styles.container}>
-      {providers?.length === 0 ? (
```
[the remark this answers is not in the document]
awtsui: Tests were updated in #2702. That PR removed the exchange card if none were available, so there was no need to test the possibility that ExternalExchanges was navigated to in the effect of no available exchanges. The deletion of this component in this PR was sort of the result of splitting the issue into two PRs – wanted to implement each section independently.

awtsui:
> would love to also see a screenshot of when there are providers, but none are exchanges (in which case there shouldn't be an exchanges section in the Select Providers screen)

Updated description in #2702.

src/fiatExchanges/SelectProvider.tsx
```diff
@@ -158,39 +166,89 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
...
+            />
+          ) : null}
+          <ExchangesSection
+            exchanges={exchanges}
+            selectedCurrency={route.params.selectedCrypto}
+            flow={flow}
+          />
+          <LimitedPaymentMethods visible={noPaymentMethods} flow={flow} />
+        </ScrollView>
+      ) : (
```
jh2oman, line 228: nit: I'm not a fan of this very large ternary operator. I think it'd be more clear in the code if you had a large if statement returning the change currency UI. Similar to the Loading screen logic

```
if(!anyProviders) {
// return switch currency UI
}
```

src/fiatExchanges/SelectProvider.tsx
jh2oman: +1

jh2oman (approved):
Looks great! nice job @awtsui 💯

src/fiatExchanges/SelectProvider.tsx
```diff
@@ -175,15 +231,17 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
...
+        providers={legacyMobileMoneyProviders || []}
         digitalAsset={digitalAsset}
         flow={flow}
       />
-      <CoinbasePaymentSection
-        digitalAsset={digitalAsset}
-        cryptoAmount={route.params.amount.crypto}
-        coinbaseProvider={coinbaseProvider}
-      />
+      {coinbaseProvider && coinbasePayVisible ? (
```
jophish, line 238: nit: Do you need to use a ternary here? [Can you just do](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator) `{(coinbaseProvider && coinbasePayVisible) && ...}`?

jophish: Wait, `coinbaseProvider` is always truthy if `coinbasePayVisible` is true; can we just check that `coinbasePayVisible` is true instead?

jophish (commented):
Tiny comments about some of the coinbase logic!

src/fiatExchanges/SelectProvider.tsx
awtsui: Even though <code>coinbasePayVisible</code> is dependent on <code>coinbaseProvider</code>, <code>coinbaseProvider={coinbaseProvider}</code> doesn't seem to recognize that coinbaseProvider must be defined unless I add coinbaseProvider in the ternary, or condition.

awtsui: The inline condition is helpful though. Didn't know that works

jophish: Huh, are you saying Typescript complains that `coinbaseProvider` might be undefined? You could potentially use a nonnull assertion since we know it will be defined (i.e. `coinbaseProvider!`), but not strictly necessary (and not the best style -- strange that Typescript doesn't recognize that `coinbaseProvider` must be nonnull here...)

awtsui: At the moment, the condition is <code>coinbaseProvider && coinbasePayVisible && <..../></code>. Would this be more readable?

src/fiatExchanges/SelectProvider.tsx
```diff
@@ -235,13 +236,13 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
         digitalAsset={digitalAsset}
         flow={flow}
       />
-      {coinbaseProvider && coinbasePayVisible ? (
+      {coinbaseProvider && coinbasePayVisible && (
```
jh2oman, line 239:
```suggestion
      {coinbasePayVisible && (
```
coinbasePayVisible is only truthy when coinbaseProvider is also truthy

jh2oman: oh just saw your comment that typescript was complaining about this. Very odd that its not understanding

jh2oman (approved):

ValoraQA:
Hey @MuckT  since NA is mentioned in how others should test section we are skipping this PR 

 CC: @awtsui 
 
 Thanks,!

PR #2711, opened 2022-07-20
refactor(fiatconnect): separate out fetch providers

jh2oman:
### Description

Refactors the way that providers for fiatconnect are fetched so that they are fetched a single time when a user enters the CICO flow. This makes fetching quotes cleaner and later will make fetching fiatAccounts cleaner.

### Tested

unit tests

### How others should test


src/fiatconnect/index.ts
```diff
@@ -53,11 +53,10 @@ export async function getFiatConnectProviders(
...
   )
   if (!response.ok) {
-    Logger.error(
-      TAG,
-      `Failure response fetching FiatConnect providers: ${response} , returning empty list`
+    Logger.error(TAG, `Failure response fetching FiatConnect providers: ${response}`)
+    throw new Error(
+      `Failure response fetching FiatConnect providers. ${response.status}  ${response.statusText}`
     )
-    return []
```
jh2oman, line 60: switched this to throw an error so that we use cached values of the provider list if this call fails

src/fiatconnect/index.ts
```diff
@@ -146,20 +145,17 @@ export async function getFiatConnectQuotes(
...
 export async function fetchQuotes(params: FetchQuotesInput) {
-  const { account, fiatConnectCashInEnabled, fiatConnectCashOutEnabled, ...quotesInput } = params
+  const { fiatConnectCashInEnabled, fiatConnectCashOutEnabled, ...quotesInput } = params
   if (!fiatConnectCashInEnabled && params.flow === CICOFlow.CashIn) return []
   if (!fiatConnectCashOutEnabled && params.flow === CICOFlow.CashOut) return []
-  const fiatConnectProviders = await getFiatConnectProviders(account)
   return getFiatConnectQuotes({
     ...quotesInput,
-    fiatConnectProviders,
   })
```
satish-ravi, line 159: nit: could just do `return getFiatConnectQuotes(quotesInput)`

src/fiatconnect/index.ts
```diff
@@ -146,20 +145,17 @@ export async function getFiatConnectQuotes(
...
-export type FetchQuotesInput = Omit<QuotesInput, 'fiatConnectProviders'> & {
+export type FetchQuotesInput = QuotesInput & {
   fiatConnectCashInEnabled: boolean
   fiatConnectCashOutEnabled: boolean
-  account: string
 }
 
 export async function fetchQuotes(params: FetchQuotesInput) {
-  const { account, fiatConnectCashInEnabled, fiatConnectCashOutEnabled, ...quotesInput } = params
+  const { fiatConnectCashInEnabled, fiatConnectCashOutEnabled, ...quotesInput } = params
```
satish-ravi, line 154: nit: use destructuring assignment for `flow` too?

src/fiatconnect/saga.test.ts
```diff
@@ -83,27 +103,60 @@ describe('Fiatconnect saga', () => {
...
         flow: CICOFlow.CashIn,
         localCurrency: 'USD',
+        fiatConnectProviders: mockFiatConnectProviderInfo,
       })
     })
   })
 
+  describe('handles fetching providers', () => {
+    it('saves on success', async () => {
+      mocked(getFiatConnectProviders).mockImplementation(() =>
```
satish-ravi, line 129: you can use mockResolvedValue / mockRejectedValue as shorthands for mockImplementation

src/fiatconnect/saga.ts
```diff
@@ -40,21 +49,21 @@ export function* handleFetchFiatConnectQuotes({
 }: ReturnType<typeof fetchFiatConnectQuotes>) {
   const { flow, digitalAsset, cryptoAmount } = params
   const userLocation: UserLocationData = yield select(userLocationDataSelector)
-  const account: string = yield select(currentAccountSelector)!
   const localCurrency: LocalCurrencyCode = yield select(getLocalCurrencyCode)
   const fiatConnectCashInEnabled: boolean = yield select(fiatConnectCashInEnabledSelector)
   const fiatConnectCashOutEnabled: boolean = yield select(fiatConnectCashOutEnabledSelector)
+  const fiatConnectProviders: FiatConnectProviderInfo[] = yield select(fiatConnectProvidersSelector)
```
satish-ravi, line 55: Do we have to worry about a race condition if fetchProviders call is slow and this attempts to fetch quotes while the API call is still happening? Or is that an edge case that easily be handled by user retry?

src/fiatconnect/saga.ts
```diff
@@ -67,6 +76,23 @@ export function* watchFetchFiatConnectQuotes() {
   yield takeLeading(fetchFiatConnectQuotes.type, handleFetchFiatConnectQuotes)
 }
 
+export function* handleFetchFiatConnectProviders() {
+  const account: string = yield select(currentAccountSelector)
+  try {
+    if (!account) {
+      throw new Error('Cannot fetch fiatconnect providers without an account')
+    }
+    const providers: FiatConnectProviderInfo[] = yield call(getFiatConnectProviders, account)
```
satish-ravi, line 85: seems like the logic is split between fiatconnect/index.ts and fiatconnect/saga.ts for this and fetch quotes. Wonder if we should consolidate in a single place. Related to https://github.com/valora-inc/wallet/pull/2687#discussion_r921410188

src/fiatconnect/saga.ts
```diff
@@ -67,6 +76,23 @@ export function* watchFetchFiatConnectQuotes() {
...
+    yield put(fetchFiatConnectProvidersCompleted({ providers }))
+  } catch (error) {
+    Logger.error(TAG, 'Error in *handleFetchFiatConnectProviders ', error)
+  }
+}
+
+export function* watchFetchFiatConnectProviders() {
+  yield takeLeading(fetchFiatConnectProviders.type, handleFetchFiatConnectProviders)
```
satish-ravi, line 93: just for my understanding, this ensures only one call can happen even if someone navigates quickly from Add / withdraw to another screen and back to Add / withdraw right?

satish-ravi (commented):
Looks good, just a couple of questions and few nits

src/fiatconnect/saga.ts
jophish: What would the behavior there be? Would an error actually be thrown? It looks like we'd just set the providers query param to empty, so not sure if this would generate an error/prompt for a retry.

src/fiatconnect/saga.ts
```diff
@@ -67,6 +76,23 @@ export function* watchFetchFiatConnectQuotes() {
   yield takeLeading(fetchFiatConnectQuotes.type, handleFetchFiatConnectQuotes)
 }
 
+export function* handleFetchFiatConnectProviders() {
+  const account: string = yield select(currentAccountSelector)
+  try {
+    if (!account) {
```
jophish, line 82: Not directly related to this PR, but I'm forgetting why we require an account address to return eligible providers. Looking at the [implementation](https://github.com/valora-inc/valora-rest-api/blob/main/src/fiatConnect/getFiatConnectProviders.ts), it looks like it's only used to allow whitelisting for specific users before a provider is actually "live". Not for this PR, but it seems like we could make this an optional param and avoid a failure case here.

src/fiatconnect/saga.ts
jh2oman: No it wouldn't. I'll add in functionality so that if the cache has never been populated that an error pops up and the user can refresh by going back and forward again.

src/fiatconnect/index.ts
jh2oman: Its cleaner to not destructure IMO because it needs to be passed with quotesInput

src/fiatconnect/saga.ts
jh2oman: takeLeading stops listening for more events of that type while the current function is executing, so yes.

src/fiatconnect/saga.ts
jh2oman: Yea I think that's a good point since its not necessary for normal production use

src/fiatconnect/index.ts
satish-ravi: Ah, missed that this is also needed in quotesInput

src/fiatconnect/saga.test.ts
```diff
@@ -83,25 +103,74 @@ describe('Fiatconnect saga', () => {
...
         digitalAsset: 'CELO',
         fiatConnectCashInEnabled: false,
         fiatConnectCashOutEnabled: true,
         flow: CICOFlow.CashIn,
         localCurrency: 'USD',
+        fiatConnectProviders: mockFiatConnectProviderInfo,
       })
     })
+    it('saves an error when providers is null', async () => {
+      mocked(fetchQuotes).mockImplementation(() => Promise.resolve(mockFiatConnectQuotes))
```
satish-ravi, line 126:
```suggestion
      mocked(fetchQuotes).mockResolvedValue(mockFiatConnectQuotes)
```

src/fiatconnect/saga.ts
jh2oman: I think that its not worth refactoring, but good to consider the logic being together for the rest of the implementation.

satish-ravi (approved):

ValoraQA:

Cc: @jh2oman 
thanks.!

Issue #1462, opened 2021-11-16
Instrument key paths that impact user experience


silasbw:
ENG:
* See [custom instrumentation](https://docs.sentry.io/platforms/react-native/performance/instrumentation/custom-instrumentation/)

AC:
* TX confirmation time (from Sending to displayed as confirmed in the user's own feed) -- Manual instrumentation probably required
* Balance update time (from fetching to displaying) -- Manual instrumentation probably required
* WalletConnect -- Manual instrumentation probably required
  * Time from deeplink to actual request displayed (from deeplink handler to tx approve modal)
  * Pairing time (from deeplink to connect request displayed)
* Pincode screen

PR #2728, opened 2022-07-26
feat: sentry custom instrumentation

MuckT:
### Description

Implements [Sentry Custom Instrumentation](https://docs.sentry.io/platforms/react-native/performance/instrumentation/custom-instrumentation/) using the SentryTransactionHub to allow us to track flows throughout the app. A singleton was used so we could tie into flows that start from one part of the app and finish in another.

### Other changes

Upgrade [@sentry/react-native](https://github.com/getsentry/sentry-react-native/releases/tag/4.2.0) to `4.2.0`.
Add [@sentry/types](https://github.com/getsentry/sentry-javascript/tree/master/packages/types).

### Tested

Tested locally with an Android and iOS Release Build:
- [Dappkit Connection](https://sentry.io/organizations/valora-inc/performance/summary/?project=1250733&query=transaction.duration%3A%3C15m&statsPeriod=7d&transaction=DappKitConnection&unselectedSeries=p100%28%29)
- [Dappkit Transaction](https://sentry.io/organizations/valora-inc/performance/summary/?project=1250733&query=transaction.duration%3A%3C15m&statsPeriod=7d&transaction=DappKitTransaction&unselectedSeries=p100%28%29)
- [Fetch Balances](https://sentry.io/organizations/valora-inc/performance/summary/?project=1250733&query=transaction.duration%3A%3C15m&statsPeriod=7d&transaction=Fetch+Balances&unselectedSeries=p100%28%29)
- [Import Contacts](https://sentry.io/organizations/valora-inc/performance/summary/?project=1250733&query=transaction.duration%3A%3C15m&statsPeriod=7d&transaction=Import+Contacts&unselectedSeries=p100%28%29)
- [Pincode Enter](https://sentry.io/organizations/valora-inc/performance/summary/?project=1250733&query=transaction.duration%3A%3C15m&statsPeriod=7d&transaction=Pincode+Enter&unselectedSeries=p100%28%29)
- [Send Payment or Invite](https://sentry.io/organizations/valora-inc/performance/summary/?project=1250733&query=transaction.duration%3A%3C15m&statsPeriod=7d&transaction=Send+Payment+or+Invite&unselectedSeries=p100%28%29)
- [Wallet Connect Connection](https://sentry.io/organizations/valora-inc/performance/summary/?project=1250733&query=transaction.duration%3A%3C15m&statsPeriod=24h&transaction=Wallet+Connect+Connection&unselectedSeries=p100%28%29)
- [Wallet Connect Transaction](https://sentry.io/organizations/valora-inc/performance/summary/?project=1250733&query=transaction.duration%3A%3C15m&statsPeriod=7d&transaction=Wallet+Connect+Transaction&unselectedSeries=p100%28%29)

### How others should test

N/A

### Related issues

- Closes: #1462 - [RET-61](https://linear.app/valora/issue/RET-61) 

### Backwards compatibility

Yes


src/sentry/transactionHub.ts
```diff
@@ -0,0 +1,37 @@
...
+// @ts-ignore
+export class TransactionHub {
+  private static instance: TransactionHub
+  transactions: Array<Transaction>
+
+  private constructor() {
+    this.transactions = []
+  }
+
+  public static getInstance(): TransactionHub {
```
kathaypacific, line 13: hum, i'm wondering if we need this singleton implementation here. a few reasons i think this
- it's not a pattern that we currently use in the project. i generally don't have strong feelings about code patterns because they all have pros and cons, but i feel strongly about consistency because it makes code more predictable/readable/maintainable. we have some similar classes that act like singletons in the code already, like ValoraAnalytics and Logger. I would advocate for adopting the same approach as the Logger (i.e. exporting and consuming the class in the same way)
- i think JS has the singleton pattern baked in, with object literals. this article might be interesting, i'm thinking we could simplify a lot if we aren't so verbose about the singleton while still getting the benefits of a singleton https://codeburst.io/javascript-global-variables-vs-singletons-d825fcab75f9#:~:text=Singleton%20is%20an%20object%20which,the%20reference%20to%20that%20object.&text=Technically%20speaking%2C%20object%20literals%20in%20JavaScript%20are%20singletons.

src/sentry/transactionHub.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (!TransactionHub.instance) {
+      TransactionHub.instance = new TransactionHub()
+    }
+    return TransactionHub.instance
+  }
+
+  public startTransaction(transactionContext: TransactionContext) {
```
kathaypacific, line 20: this is not a fully thought out suggestion, so i'm wondering what you think about this:

since we're abstracting `Sentry.startTransaction`, i'm thinking we could make the input arguments of this method a little more friendly (instead of very bespoke Sentry arguments). it can be as simple as just having the input arguments spread out rather than one object, since that already helps with type prompts when writing new code. 

in ValoraAnalytics we have some guidance for what we can pass into the `track` method. i wonder if we should consider doing something like this for sentry transactions too, so that consumers don't need to know the exact string value of the transaction. like `SentryTransactionHub.startTransaction(Transactions.SendOrInvite, Spans.Send)`

of course this means more abstractions/types/boilerplate :/ but in return we get an easier to use method + less chance of a typo (e.g. finishing a transaction that doesn't exist, so that the transaction that started will never be sent to Sentry) + a place where we can see all the transactions we're sending to sentry

kathaypacific (commented):
nice work so far!!!!

src/navigator/NavigatorWrapper.tsx
```diff
@@ -154,7 +155,7 @@ export const NavigatorWrapper = () => {
       Sentry.addBreadcrumb({
         category: 'navigation',
         message: `Navigated to ${currentRouteName}`,
-        level: Sentry.Severity.Info,
+        level: 'info' as SeverityLevel,
```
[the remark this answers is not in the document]
MuckT: This was needed due to the sentry upgrade - [changelog](https://github.com/getsentry/sentry-javascript/blob/master/MIGRATION.md#severity-severitylevel-and-severitylevels).

src/sentry/transactionHub.ts
MuckT: Added in https://github.com/valora-inc/wallet/pull/2728/commits/76f4ef86a8fc1d84cc74f72680ed4db2d4f51f09!

src/sentry/transactionHub.ts
MuckT: Changed to an object literal in https://github.com/valora-inc/wallet/pull/2728/commits/485a85f463b1b291e2caace4f93e752b9cce4ae8!

src/sentry/transactionHub.ts
```diff
@@ -0,0 +1,37 @@
+import * as Sentry from '@sentry/react-native'
+import { Transaction, TransactionContext } from '@sentry/types'
+
+// @ts-ignore
+export class TransactionHub {
```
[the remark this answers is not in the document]
MuckT: Changed in https://github.com/valora-inc/wallet/pull/2728/commits/485a85f463b1b291e2caace4f93e752b9cce4ae8!

src/sentry/transactionHub.ts
```diff
@@ -0,0 +1,37 @@
+import * as Sentry from '@sentry/react-native'
+import { Transaction, TransactionContext } from '@sentry/types'
+
+// @ts-ignore
```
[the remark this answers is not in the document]
MuckT: Was able to remove when moving to the object literal format in https://github.com/valora-inc/wallet/pull/2728/commits/485a85f463b1b291e2caace4f93e752b9cce4ae8.

src/sentry/SentrySpans.tsx
```diff
@@ -0,0 +1,62 @@
...
+  pincode_set = 'PincodeSet',
+  send_payment_or_invite = 'SendPaymentOrInvite',
+  send_payment_or_invite_legacy = 'SendPaymentOrInviteLegacy',
+  wallet_connect_connection = 'WalletConnectConnection',
+  wallet_connect_transaction = 'WalletConnectTransaction',
+}
+
+type values = typeof SentrySpan[keyof typeof SentrySpan]
+
+export const SentrySpans: Record<values, SentrySpanInfo> = {
```
kathaypacific, line 21: it took me a while to understand the difference between SentrySpan and SentrySpans, maybe this can be called `SentryTransactionContext` since within Sentry the type of this object is called `TransactionContext`? I'm also a little confused about transaction vs span, these things are really transactions? so we can also rename `SentrySpan` to `SentryTransaction`?

src/sentry/SentryTransactionHub.ts
```diff
@@ -0,0 +1,26 @@
+import * as Sentry from '@sentry/react-native'
+import { Transaction } from '@sentry/types'
+import { SentrySpan, SentrySpans } from 'src/sentry/SentrySpans'
+
+export const SentryTransactionHub = {
+  transactions: [] as Array<Transaction>,
+  startTransaction(name: SentrySpan) {
+    const transaction = Sentry.startTransaction({ ...SentrySpans[name], trimEnd: true })
+    this.transactions.push(transaction)
```
kathaypacific, line 9: i was a little surprised to see `this`, usually that's used with classes when there is an instance...can the `transactions` just be a variable in this file, where it's not exported? i think that achieves the same effect?

kathaypacific (commented):
Looking great!!

src/navigator/NavigatorWrapper.tsx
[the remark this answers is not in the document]
kathaypacific: oh i see...i missed that it was a major version upgrade

src/sentry/SentrySpans.tsx
MuckT: I went with the name SentrySpan to match what is seen in the Sentry UI.
Screen Shot 2022-07-27 at 8 00 06 AM

MuckT:
Having a bit of difficulty with `SentryTransactionHub.finishTransaction(SentrySpan.dappkit_connection)` & `SentryTransactionHub.finishTransaction(SentrySpan.dappkit_transaction)` as I have been unable to confirm they are working as intended in Sentry.

src/utils/Logger.ts
```diff
@@ -69,7 +70,7 @@ class Logger {
     // prevent genuine network errors from being sent to Sentry
     if (!isNetworkError || (this.isNetworkConnected && isNetworkError)) {
       const captureContext = {
-        level: Sentry.Severity.Error,
+        level: 'error' as SeverityLevel,
```
silasbw, line 73: Why?

src/sentry/SentryTransactionHub.ts
```diff
@@ -0,0 +1,27 @@
...
+  startTransaction(name: SentrySpan) {
+    const transaction = Sentry.startTransaction({ ...SentrySpans[name], trimEnd: true })
+    transactions.push(transaction)
+  },
+  finishTransaction(name: SentrySpan) {
+    // get span operation - 'op'
+    const op = SentrySpans[name].op
+
+    // Find all the transactions with this op.
+    const selectedTransactions = transactions.filter(
```
silasbw, line 17: I don't understand this part.  Does it imply `startTransaction` is called with a `name` multiple times, but there isn't a always matching calls to `finishTransaction`?  Why not?

silasbw:
does this disable the automatic instrumentation we're currently using?  if not, should we disable it?

src/utils/Logger.ts
MuckT: This was needed due to the sentry upgrade - [changelog](https://github.com/getsentry/sentry-javascript/blob/master/MIGRATION.md#severity-severitylevel-and-severitylevels).

src/sentry/SentryTransactionHub.ts
MuckT: `startTransaction` transaction could be called on a span that is never completed such as a wallet connect transaction that results in an error. This finishes and then removes all other spans of that same name from the SentryTransactionHub.

kathaypacific (commented):
I had a thought as i was reviewing this - we are starting transactions in one place, and finishing transactions in mostly one other place. the wallet connect transactions for example only finish inside the `try` block, which means all the wallet connect errors will leave open transactions that are "finished" by later wallet connect requests. i'm not sure this is ideal / correct? perhaps we should add a `finally` to our try / catch blocks, or finish the transactions inside the catch blocks too. 

also another thought, do we have plans to measure spans? it might be helpful to break up transactions into spans, so we can see exactly how each span is performing. for example it might be nice to see the breakdown of the time it takes to complete dappkit requests + the time it takes to navigate (probably not the best example, but you get the idea) - right now we only get a transaction that measures the sum of both.

MuckT:
@kathaypacific added a reply in [RET-61](https://linear.app/valora/issue/RET-61/instrument-key-paths-that-impact-user-experience).

kathaypacific (commented):
Hey sorry this got lost in my to-do list! it looks good, i do have a couple general questions - happy to discuss in person if easier.

i think we're in agreement that we want to eventually include spans, to expose more granular details about each transaction right? happy for this to be in future PR's / tickets, but if we agree on this then could we free up the "spans" namespace? my understanding is that transactions contain spans, and what we're measuring in this PR are transactions but we've got a lot of things called "spans" which will be confusing when we get to instrumenting spans within transactions. what do you think?

one other question i had, is now given a "finish" transaction, we use that to close all "started" transactions and send them all to sentry. i think the risk is that we receive multiple transactions, and while we don't know exactly which event the "finish" transaction matches to, does it make more sense to close only the last (or first) item in the transactions list and simply to remove the rest from the list? curious about your thoughts on this, but i think a consistent rule will even out the trends over time and it feels more correct to me to receive the same number of events that are dispatched by the app (not more than that)

MuckT:
@kathaypacific - I'll free up the spans namespace and rename was is currently considered a span to a transaction.

Regarding closing the transactions I like your idea of closing only the first selected transaction and discarding the others of the same op from the transaction hub. Based on what I am reading in [The Sentry Docs](https://docs.sentry.io/platforms/javascript/performance/instrumentation/custom-instrumentation/) unfinished transactions aren't sent to Sentry so that will give us more consistent results in monitoring.

MuckT:
@kathaypacific 
- Renamed SentrySpans to SentryTransactions in https://github.com/valora-inc/wallet/pull/2728/commits/19ced77721ab0af48a498bbf09fad8613b44404b
- Made it so only one transaction is finished and the rest are discarded from the hub in https://github.com/valora-inc/wallet/pull/2728/commits/365524c09cc4fb450025c05fef9a4962ab6f86a9

src/sentry/SentryTransactionHub.ts
```diff
@@ -0,0 +1,27 @@
...
+
+let transactions = [] as Array<Transaction>
+
+export const SentryTransactionHub = {
+  startTransaction(name: SentryTransaction) {
+    const transaction = Sentry.startTransaction({ ...SentryTransactions[name], trimEnd: true })
+    transactions.push(transaction)
+  },
+  finishTransaction(name: SentryTransaction) {
+    // get span operation - 'op'
```
kathaypacific, line 13: this comment is outdated, maybe can be removed :)

kathaypacific (approved):
nice!!

src/sentry/SentryTransactionHub.ts
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2728/commits/5bdece43e6bdf345c2417fe8757859beb77008aa!

ValoraQA:
hey @MuckT since NA is mentioned in how others should test we are skipping this PR

Thanks.!

PR #2729, opened 2022-07-26
feat(in-app swaps): connect swap screen to cloud function 

kathaypacific:
### Description

Connect swap screen to cloud function and ensure details are correct:
- add error banner / clear exchange rate when api request fails
- only query the api at a debounced time interval
- do not query api when the swap amount is 0
- handle exchange rates from the api, for when changing swap amounts and swap tokens
- handle a visual loading state while fetching new rates

There are 2 not so elegant things in this implementation (open for suggestions on how to improve these)
- android and ios TextInput don't behave the same, so to handle long values (e.g. long swap values on small devices) there is an elaborate hack involving the `selection` in `SwapAmountInput.tsx` - i've tried to keep this localised, so that if we don't need it anymore then it's straightforward to remove
- changing swap amount (independent value) -> should get new quote / exchange rate -> should change swap amount (dependent value). this loop makes it difficult to avoid a duplicate request - i've added a ref variable to prevent this in `useSwapQuote.ts`, again not great but i've tried to keep it contained so that we can change it. 

https://user-images.githubusercontent.com/20150449/182576818-4721dad6-13f8-42d5-938d-b02dff4d61da.mp4

### Other changes

N/A

### Tested

A truckload of unit tests + much manual testing on android / ios

### How others should test

- on change "from" input => the "to" value should change and the exchange rate should be updated
- on change "to" input => the "from" value should change and the exchange rate should be updated
- on press "max" => the "from" value should be the max balance of that token for the user, and the "to" value + exchange rate should be updated
- removing any value (by clicking on the cross icon inside the "to" or "from" input) should clear values for both lines
- on change "from" token => the "from" value should remain, and the "to" value + exchange rate should be updated
- on change "to" token => the "to" value should remain, and the "from" value + exchange rate should be updated

### Related issues

- Fixes RET-269

### Backwards compatibility

Yes


MuckT:
@kathaypacific one thing that stood out to me in a quick test was if I change to To token the amount in the from token updates. I think this is a bit strange as if I wanted to swap 10 Celo for something i'd be nice to be able to go through my different options seeing the result of the swap for 10 Celo.

kathaypacific:
> @kathaypacific one thing that stood out to me in a quick test was if I change to To token the amount in the from token updates. I think this is a bit strange as if I wanted to swap 10 Celo for something i'd be nice to be able to go through my different options seeing the result of the swap for 10 Celo.

oh interesting, that makes sense. i actually put in a line of code specifically to make this happen, i thought it made sense but maybe not. i removed it, i guess the question still remains what should happen when you change the value of the "to" field, and then change the "to" token? would you expect the "to" or the "from" value to change in this case? i've got a meeting with Nitya tonight, will also check with her :)  thanks!

src/swap/SwapScreen.test.tsx
```diff
@@ -142,27 +146,106 @@ describe('SwapScreen', () => {
...
+    )
+    const { swapFromContainer, swapToContainer, getByText } = renderScreen({})
+
+    void act(() => {
+      fireEvent.changeText(within(swapToContainer).getByTestId('SwapAmountInput/Input'), '1.234')
+      jest.runAllTimers()
+    })
+
+    expect(getByText('1 CELO ≈ 8.10000 cUSD')).toBeTruthy()
+    expect(within(swapFromContainer).getByTestId('SwapAmountInput/Input').props.value).toBe(
```
dievazqu, line 187: Shouldn't this be `SwapAmountIOutput/Output` ??
I seeing the test is passing, I'm wondering if this test even work. 😅

dievazqu (approved):
LGTM!! I'm not sure if we want to do some changes after the discussion on the blocking behavior, but I give you my approval if we want to go with this first version and then iterate over it.

MuckT:
> when you change the value of the "to" field, and then change the "to" token?

I would expect the 'from' field to update in this case. Maybe we could keep track of the last input selected? This could be a follow up PR if desired.
- Default to the 'from' field as last selected.
- If last selected is 'from': Changes to 'from' input, 'from' token and 'to' token - change the value in the 'to' input.
- If last selected is 'to': Changes 'to' input and 'to' token - changes the value in the 'from' input.

kathaypacific:
> > when you change the value of the "to" field, and then change the "to" token?
> 
> I would expect the 'from' field to update in this case. Maybe we could keep track of the last input selected? This could be a follow up PR if desired.
> 
> * Default to the 'from' field as last selected.
> * If last selected is 'from': Changes to 'from' input, 'from' token and 'to' token - change the value in the 'to' input.
> * If last selected is 'to': Changes 'to' input and 'to' token - changes the value in the 'from' input.

i *think* that is the behaviour that is in this PR :D so we are aligned!

ValoraQA:
Hey @MuckT as discussed in weeky call this is still behind feature flag / still in implementation phase, please let us know when it needed to be tested

Cc: @kathaypacific 
Thanks..!

PR #2730, opened 2022-07-26
feat(fiatconnect): part two of return user flow. navigate directly to review screen if there is a fiataccount

jh2oman:
### Description

1. On the select provider screen, navigate to the Review Screen if there is a fiataccount
2. On the Review screen, allow the provider to be clickable so that a user can switch providers

https://user-images.githubusercontent.com/8432644/181045807-e4eb3e7b-ec9a-4e8f-b719-33144f15cf0f.mp4

### Tested

manually and unit tested

### How others should test

### Related issues

https://linear.app/valora/issue/ACT-349/fc-return-user-flow

### Backwards compatibility


src/fiatconnect/saga.ts
```diff
@@ -206,13 +210,56 @@ function* _getFiatAccount({
     providerId,
     fiatConnectProvider.baseUrl
   )
-  const fiatAccount = fiatAccounts.find((account) => account.fiatAccountId === fiatAccountId)
+
+  let fiatAccount: FiatAccount | undefined
+  if (fiatAccountType) {
+    fiatAccount = fiatAccounts.find((account) => account.fiatAccountType === fiatAccountType)
```
jophish, line 216: There's potentially an edge case here, I think.. imagine the scenario where:

* The current quote specifies account type `TypeA` allowing the underlying fiat account schema `SchemaA`
* The user has a `TypeA` account on file with the provider, but it uses `SchemaB`

This method would return the `SchemaB` account since it matches the account type `TypeA`, but it wouldn't actually be valid to use for the given quote since the account schema isn't legal. We should probably also check that the schema matches the legal schemas for the quote as well.

src/fiatExchanges/quotes/FiatConnectQuote.ts
```diff
@@ -123,12 +123,8 @@ export default class FiatConnectQuote extends NormalizedQuote {
...
   }
 
-  // TODO: Navigate to the Review screen if a fiatAccount is already saved
-  navigate(flow: CICOFlow): void {
-    navigate(Screens.FiatDetailsScreen, {
-      quote: this,
-      flow,
-    })
+  navigate(dispatch: Dispatch): void {
+    dispatch(selectFiatConnectQuote({ quote: this }))
```
jophish, line 127: Do we need to show a spinner while the GET accounts call is being made to the provider? What if the user performs some navigation of their own while the request is being sent but before the saga calls `navigate`?

jophish (commented):
Couple of comments!

src/fiatconnect/saga.ts
jh2oman: good call 👍

src/fiatExchanges/quotes/FiatConnectQuote.ts
jh2oman: also good call 👍

satish-ravi (commented):
Looks good, just a question on handling expected scenarios as errors.

src/fiatconnect/saga.ts
```diff
@@ -206,13 +211,61 @@ function* _getFiatAccount({
...
+      flow: quote.flow,
+      normalizedQuote: quote,
+      fiatAccount,
+    })
+    yield delay(500) // to avoid a screen flash
+    yield put(selectFiatConnectQuoteCompleted())
+  } catch (error) {
+    // Ignore error and navigate to FiatDetails screen.
+    // This is expected when the user has not yet created a fiatAccount with the provider
+    navigate(Screens.FiatDetailsScreen, {
```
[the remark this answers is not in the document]
jophish: I think, in the case of a transient API error, the safest bet would be to show some kind of "Try a different provider/try again" error screen. We can't be sure if the user has an account; If we ship them off to the fiat details screen in this case and they have an account, they'll get an error when adding. If our state says they have an account but it's actually been deleted somehow on the provider side and we send them to the review screen, they'll get an error when attempting to initiate the transfer.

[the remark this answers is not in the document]
jh2oman: Great feedback y'all. updating now

src/fiatconnect/saga.ts
```diff
@@ -208,11 +218,66 @@ function* _getFiatAccount({
...
+    yield delay(500) // to avoid a screen flash
+    yield put(selectFiatConnectQuoteCompleted())
+  } catch (error) {
+    // Error while attempting fetching the fiatConnect account
+    Logger.debug(
+      TAG,
+      `handleSelectFiatConnectQuote* Error while attempting to fetch the fiatAccount for provider ${quote.getProviderId()}`,
+      error
+    )
+    yield put(selectFiatConnectQuoteCompleted())
```
satish-ravi, line 278: this would just remove the spinner and put the user back on the select provider screen right?

satish-ravi (approved):

src/fiatconnect/saga.ts
jh2oman: yes

ValoraQA:

Cc: @jh2oman 
Thanks.!

PR #2731, opened 2022-07-26
feat(fiatconnect): Handle expired quotes

cajubelt:
### Description

Handles expired quotes on the transaction `ReviewScreen`. Follows the design specified [here](https://www.figma.com/file/bwIY7peeykI8K4wTrXs4iN/Fiat-Connect?node-id=1361%3A31757)

Currently, there are no checks as to whether the quote given to the user is still valid when they select the withdraw funds button. These changes will cause an expired quote dialog to be shown if the quote is expired when the user enters the `ReviewScreen` or when they attempt to withdraw funds.

##### Example of dialog when entering `ReviewScreen`

https://user-images.githubusercontent.com/20508929/181304228-14db0e81-883a-44cc-8706-36824430d0c3.mp4

##### Example of dialog while on `ReviewScreen`

The quote was set to expire 5 seconds after being obtained.

https://user-images.githubusercontent.com/20508929/181304722-40c487b4-b679-413d-b252-8c703d15261b.mp4

### Other changes

None

### Tested
 
Added tests that verifies that the expired dialog is shown if the client enters the `ReviewScreen` with an expired quote. 

### How others should test

Doesn't need to be tested by QA currently. 

### Related issues

- Fixes [ACT-352](https://www.figma.com/file/bwIY7peeykI8K4wTrXs4iN/Fiat-Connect?node-id=1498%3A36800)

### Backwards compatibility

N/A


src/fiatconnect/ReviewScreen.test.tsx
```diff
@@ -137,25 +136,6 @@ describe('ReviewScreen', () => {
...
-      const store = createMockStore()
-      const { getByTestId, queryByTestId } = render(
-        <Provider store={store}>
-          <FiatConnectReviewScreen {...mockProps} />
-        </Provider>
-      )
-
-      expect(queryByTestId('expiredQuoteDialog')?.props.visible).toEqual(false)
-
-      await setTimeout(expirySecs * 1000)
```
anthonydaoud: I couldn't use setTimeout here as it caused one of the CI tests to fail. I tried sleeping by awaiting a promise wrapped around setTimeout, but it seemed to never timeout oddly enough. I also tried messing with jest.setSystemTime but that made both tests fail. Please let me know if you have any insights as to what's going on.

anthonydaoud: Nevermind, figured it out!

src/fiatExchanges/quotes/FiatConnectQuote.ts
```diff
@@ -159,6 +159,10 @@ export default class FiatConnectQuote extends NormalizedQuote {
     return resolveCurrency(this.quote.quote.cryptoType)!
   }
 
+  getCicoCryptoType(): CiCoCurrency {
+    return resolveCICOCurrency(this.quote.quote.cryptoType)
+  }
```
jh2oman, line 164: Can you keep this out of the quote? I'd rather have FiatConnectQuote only deal with one currency type even though its annoying that our codebase has several

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -34,6 +40,28 @@ export default function FiatConnectReviewScreen({ route, navigation }: Props) {
...
   const dispatch = useDispatch()
   const { flow, normalizedQuote, fiatAccount } = route.params
+  const providerId = normalizedQuote.getProviderId()
+
+  const fiatConnectQuotesLoading = useSelector(fiatConnectQuotesLoadingSelector)
+  const fiatConnectQuotes = useSelector(fiatConnectQuotesSelector)
+  const [showingExpiredQuoteDialog, setShowingExpiredQuoteDialog] = useState(false)
+
+  const updatedQuoteData = fiatConnectQuotes.find(
+    (quote) => quote.provider.id === providerId && quote.ok
```
jh2oman, line 50: What if the quote is not found? what if the quote is not ok?

src/fiatconnect/ReviewScreen.test.tsx
```diff
@@ -102,9 +108,57 @@ describe('ReviewScreen', () => {
...
+          quotes: [quote],
+        },
+      })
+      const { getByTestId, queryByTestId } = render(
+        <Provider store={store}>
+          <FiatConnectReviewScreen {...mockProps} />
+        </Provider>
+      )
+
+      expect(queryByTestId('expiredQuoteDialog')?.props.visible).toEqual(true)
```
jh2oman, line 127:
```suggestion
      expect(queryByTestId('expiredQuoteDialog')).toBeTruthy()
```

jh2oman: similar feedback for all the ones below. Truthy is visible and falsy should be not visible

src/fiatconnect/ReviewScreen.test.tsx
```diff
@@ -102,9 +108,57 @@ describe('ReviewScreen', () => {
...
+
+      expect(store.getActions()).toEqual([
+        fetchFiatConnectQuotes({
+          flow: CICOFlow.CashOut,
+          digitalAsset: CiCoCurrency.CUSD,
+          cryptoAmount: 100,
+          providerIds: [mockFiatConnectQuotes[1].provider.id],
+        }),
+      ])
+      expect(queryByTestId('expiredQuoteDialog')?.props.visible).toEqual(true)
```
jh2oman, line 139: Shouldn't this be not visible after the user has clicked the try again button?

I think to see the UI change based on internal state changes you'll need to use [rerender](https://callstack.github.io/react-native-testing-library/docs/api/#rerender)

src/fiatconnect/ReviewScreen.tsx
anthonydaoud: In that scenario, updatedQuoteData is undefined. The quote variable defined in line 52 is the original normalizedQuote and it will not be updated in line 54. Because the quote is still expired, the expiredQuoteDialog will still be shown and the client will need to re-obtain a quote. This will repeat until the provider's quote is found and ok and has a valid timestamp.

src/fiatconnect/ReviewScreen.test.tsx
anthonydaoud: Ideally yes, but unfortunately I was having a fair bit of trouble mocking fetching an updated quote. I think what happens is that the showingExpiredQuoteDialog state is set to false after the fireEvent, but the only quote present is the expired quote. This means that showingExpiredQuoteDialog is set to true again. I might just delete that expect statement

src/fiatExchanges/quotes/FiatConnectQuote.ts
anthonydaoud: Will do!

src/fiatconnect/ReviewScreen.test.tsx
anthonydaoud: Wouldn't `expect(queryByTestId('expiredQuoteDialog'))` always be truthy because the element always renders? I only change the isVisible prop.

src/fiatconnect/saga.ts
```diff
@@ -359,7 +359,9 @@ export function* handleCreateFiatConnectTransfer({
 
       const tokenList: TokenBalance[] = yield select(tokensListSelector)
       const tokenInfo = tokenList.find(
-        (token) => token.symbol === fiatConnectQuote.getCryptoType()
+        (token) =>
+          token.symbol === fiatConnectQuote.getCryptoType() ||
```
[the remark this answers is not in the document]
anthonydaoud: Ok gotcha, will revert these changes.

src/fiatconnect/ReviewScreen.tsx
jh2oman: I think it would be helpful to communicate to the user that there is an error. An error banner at the top saying there was an error fetching information from the provider would help.

src/fiatconnect/ReviewScreen.test.tsx
jh2oman: Ah I see. I made an incorrect assumption on how you were implementing it. sounds good

src/fiatconnect/ReviewScreen.tsx
anthonydaoud: That's a good idea, will implement it.

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -34,6 +42,33 @@ export default function FiatConnectReviewScreen({ route, navigation }: Props) {
...
+      (quote) => quote.provider.id === providerId && quote.ok
+    )
+    if (updatedQuoteData) {
+      quote = new FiatConnectQuote({
+        quote: updatedQuoteData as FiatConnectQuoteSuccess,
+        fiatAccountType: normalizedQuote.fiatAccountType,
+        flow,
+      })
+    } else {
+      dispatch(showError(ErrorMessages.QUOTE_UPDATE_FAILED))
```
jh2oman, line 64: You usually don't want to have a dispatch in the main function path, because this code gets called every time any props, state, or selectors changes values. We'd likely run into bugs with this getting fired multiple times.

Instead, use a `useEffect` hook and have it fire whenever the relevant data (in this case usingUpdatedQuote and quote) changes. Check out [this example in SelectProvider](https://github.com/valora-inc/wallet/blob/main/src/fiatExchanges/SelectProvider.tsx#L101)

anthonydaoud: Gotcha, just updated it!

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -44,11 +45,32 @@ export default function FiatConnectReviewScreen({ route, navigation }: Props) {
...
   const providerId = normalizedQuote.getProviderId()
 
+  const fiatConnectQuotesError = useSelector(fiatConnectQuotesErrorSelector)
   const fiatConnectQuotesLoading = useSelector(fiatConnectQuotesLoadingSelector)
   const fiatConnectQuotes = useSelector(fiatConnectQuotesSelector)
   const [showingExpiredQuoteDialog, setShowingExpiredQuoteDialog] = useState(false)
   const [usingUpdatedQuote, setUsingUpdatedQuote] = useState(false)
 
+  useEffect(() => {
+    if (!fiatConnectQuotesLoading && usingUpdatedQuote && fiatConnectQuotesError) {
```
jh2oman, line 55: What if there is not an error fetching fiatconnect quotes but theres no `updatedQuoteData` because quote.ok is false?

locales/base/translation.json
```diff
@@ -1230,6 +1230,7 @@
   "merchantPaymentSubmitSuccess": "Your payment successfully submitted",
   "merchantWalletUnlockError": "Something went wrong when unlocking your wallet, please try again",
   "noMoreTransactions": "There are no more transactions",
+  "quoteUpdateFailed": "Could not refresh the quote, please retry",
```
satish-ravi, line 1233: is this copy from figma? I think we consistently use "try again" instead of "retry" in other error messages

src/fiatconnect/ReviewScreen.tsx
anthonydaoud: Handled that now

src/fiatconnect/ReviewScreen.test.tsx
```diff
@@ -43,13 +51,11 @@ function getProps(flow: CICOFlow, withFee = false, cryptoType = CryptoType.cUSD)
 }
 
 describe('ReviewScreen', () => {
-  const store = createMockStore()
```
satish-ravi, line 46: just for my understanding. Is this created in every test so we have a fresh set of actions or something else? If it's just for actions, I think we can also do `store.clearActions()` in a `afterEach`

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -34,6 +43,33 @@ export default function FiatConnectReviewScreen({ route, navigation }: Props) {
...
+  const [showingExpiredQuoteDialog, setShowingExpiredQuoteDialog] = useState(false)
+  const [usingUpdatedQuote, setUsingUpdatedQuote] = useState(false)
+
+  useEffect(() => {
+    if (!fiatConnectQuotesLoading && usingUpdatedQuote) {
+      const updatedQuoteData = fiatConnectQuotes.find(
+        (quote) => quote.provider.id === providerId && quote.ok
+      )
+      if (fiatConnectQuotesError || !updatedQuoteData) {
+        dispatch(showError(ErrorMessages.QUOTE_UPDATE_FAILED))
```
satish-ravi, line 61: Once this error is shown, what action can the user take? Click continue and trigger another quote refresh?
Also, can we add a unit test for checking whether this is dispatched on error?

jh2oman (approved):

src/fiatconnect/ReviewScreen.test.tsx
```diff
@@ -157,6 +160,52 @@ describe('ReviewScreen', () => {
...
+      const expireMs = 100
+      const mockProps = getProps(CICOFlow.CashOut, false, CryptoType.cUSD, false, expireMs)
+      const { getByTestId, queryByTestId } = render(
+        <Provider store={store}>
+          <FiatConnectReviewScreen {...mockProps} />
+        </Provider>
+      )
+
+      expect(queryByTestId('expiredQuoteDialog')?.props.visible).toEqual(false)
+      await sleep(expireMs)
```
satish-ravi, line 203: nit: you can use jest fake timers instead of sleep https://jestjs.io/docs/timer-mocks#advance-timers-by-time

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -117,6 +120,10 @@ export default function FiatConnectReviewScreen({ route, navigation }: Props) {
     )
   }
 
+  if (!showingExpiredQuoteDialog && normalizedQuote.getGuaranteedUntil() < new Date()) {
+    setShowingExpiredQuoteDialog(true)
```
satish-ravi, line 124: react question: even though this is gated by `showingExpiredQuoteDialog` being falsy, is it good practice to set state here? Wonder if we could just set initial state of `showingExpiredQuoteDialog` to `normalizedQuote.getGuaranteedUntil() < new Date()`? cc @kathaypacific

satish-ravi (commented):
Looks good, just a nit and a question

src/fiatconnect/ReviewScreen.tsx
kathaypacific: yeah initial state sounds good. it is a little odd to have a possible state change on every render in the body of a component - state changes then cause another rerender and you could have a loop of these if not careful, so usually i'd expect state changes to be triggered inside callbacks or by other events. i think what is here is not dangerous, if it is just being used for the initial condition then putting it in where the `useState` is first declared is better, and then we'll only hit this computation once (on mount) :D

src/fiatconnect/ReviewScreen.test.tsx
anthonydaoud: That doesn't work because it only advances timers. It does not advance the global system time, which is what the quote's expiry time is compared against. I tried mocking the global Date object too, but that caused other tests to fail.

satish-ravi: Ah, I guess you can fake the time as well using [jest.setSystemTime](https://jestjs.io/docs/jest-object#jestsetsystemtimenow-number--date) on your specific time and then resetting it, but probably not worth it since it's just a 100ms wait.

satish-ravi (approved):

ValoraQA:
Hey @MuckT since NA is mentioned in how other should test section we are skipping this PR 

CC : @cajubelt 
Thanks.!

PR #2737, opened 2022-07-28
fix: clear transactions with schema migration

MuckT:
### Description

Clears the transactions in the redux store via a schema migration.

### Other changes

Avoids adding empty transactions to redux.

### Tested

WIP as I am having some issues running the blockchain api locally.

### How others should test

### Related issues

N/A

### Backwards compatibility

Yes


MuckT:
Another alternative would be to implement a user defined type guard for TokenTransaction instead of simply checking that the object is not empty.

MuckT:
Closing as the patch was applied to 1.36.1 instead of 1.37.1.

PR #2738, opened 2022-07-28
fix: app crash on empty transactions in the redux state

kathaypacific:
[No need to merge, adding here for review only before release]

Context https://valora-app.slack.com/archives/C02N3AR2P2S/p1658937575432179

This PR prevents and removes empty objects in the transactions redux state that is causing some apps to crash.


android/gradle.properties
```diff
@@ -23,7 +23,7 @@ CELO_RELEASE_KEY_ALIAS=celo-key-alias
 
 # Setting this manually based on version number until we have this deploying via Cloud Build
 # Example: v1.5.1 deployment number 1 = 1005001001 (1 005 001 001)
-VERSION_CODE=1021071620
+VERSION_CODE=1021071621
```
M-shin, line 26: Where can we find the correct version code for a given version? It isn't obvious to me where this number is coming from (unless the rule is increment previous number by 1 on any upgrade?)

M-shin (approved):

PR #2739, opened 2022-07-28
fix: schema migrations and omit empty transactions

MuckT:
### Description

Updates schema migrations and omits empty transactions from the transaction feed.

### Other changes

Version number bump.

### Tested

Tested locally on iOS.

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

Yes


dievazqu (approved):
LGTM but I would feel more confident with an approve of @kathaypacific before merging

src/redux/migrations.ts
```diff
@@ -694,35 +695,44 @@ export const migrations = {
     },
   }),
   59: (state: any) => ({
```
kathaypacific, line 697: hum, i think we will have problems with the next release if we change the migrations like this (unless we want to merge these changes back to main)

i was actually thinking that we could have something like 
```
59: (state: any) => ({
    ...state,
    transactions: {
      ...state.transactions,
      transactions: (state.transactions.transactions || []).filter(
        (transaction: TokenTransaction) => Object.keys(transaction).length > 0
      ),
    fiatConnect: {
      quotes: [],
      quotesLoading: false,
      quotesError: null,
    },
    },
  }),
  60: (state: any) => ({
    ...state,
    fiatConnect: {
      quotes: [],
      quotesLoading: false,
      quotesError: null,
    },
    app: {
      ...state.app,
      showSwapMenuInDrawerMenu: REMOTE_CONFIG_VALUES_DEFAULTS.showSwapMenuInDrawerMenu,
    },
  }),
```

this way, migration 61 onwards does not need to be changed, and this migration does not need to be merged back to main after release, because users will be on schema 63 (instead of 64) as previously expected.

kathaypacific (commented):
could you also make this a draft PR so that we don't accidentally merge? :D

src/redux/migrations.ts
kathaypacific: thinking about it more now, we can also keep this implementation if we also make the same change in the main branch.

benefit to the above approach, is that we keep 1.37.x at redux schema version 63. this is what the main branch expects, so we don't need to update main branch after this patch.

benefit to the way you've done it in this branch, is that we would have a better record of what happened. it's more changes that need to be merged to the main branch though :/ what do you think is the better approach?

kathaypacific: ah just seeing your other PR for main now, i guess this works fine :)

PR #2747, opened 2022-08-01
chore(PFP): Remove unused PFP code

jophish:
### Description

-3000 SLOC!

Fixes [ACT-45](https://linear.app/valora/issue/ACT-45/remove-most-of-pfp-from-the-wallet).

This PR removes the majority of PFP related code from the wallet. In particular, it removes all screens and settings menu items related to linking bank accounts. This includes all non-Persona-related helper code, like our Plaid/IHL/Finclusive specific code, as well as related state. I did _not_ modify our `Persona.tsx` file, since it's unclear what parts of that file exactly will be reused.

### Other changes

### Tested

### How others should test

### Related issues

Fixes [ACT-45](https://linear.app/valora/issue/ACT-45/remove-most-of-pfp-from-the-wallet)

### Backwards compatibility


src/redux/migrations.ts
```diff
@@ -575,9 +575,7 @@ export const migrations = {
     },
     app: {
       ...state.app,
-      finclusiveUnsupportedStates: REMOTE_CONFIG_VALUES_DEFAULTS.finclusiveUnsupportedStates.split(
-        ','
-      ),
+      finclusiveUnsupportedStates: ['NY', 'TX'],
```
jophish, line 578: This is the same as before, I've just removed `finclusiveUnsupportedStates` from the remote config type so I'm hardcoding it here.

jh2oman (approved):
I agree with keeping Persona.tsx around. RIP everything else 🗿

MuckT:
@jophish are there specific screens or flows that QA should test were removed?

PR #2752, opened 2022-08-02
feat(in-app swaps): support swap transactions in home feed (RET-290)

Samuel3Shin:
### Description

Added Swap transaction in the home feed.

home feed <--------------------------------------------> transaction detail
Screen Shot 2022-08-12 at 1 37 16 PM Screen Shot 2022-08-03 at 8 54 50 AM 

### Other changes
N/A

### Tested
With iOS emulator and locally running `blockchain-api`

### How others should test
With iOS emulator and `blockchain-api`

### Related issues
- https://github.com/valora-inc/blockchain-api/pull/183
- https://linear.app/valora/issue/RET-290/add-swaps-to-transaction-feed

### Backwards compatibility
Yes


MuckT:
@Samuel3Shin is there a way QA can test this feature either in alfajroes or mainnet without the blockchain-api locally?

locales/base/translation.json
```diff
@@ -1366,5 +1366,11 @@
     "swapTo": "To",
     "insufficientFunds": "Insufficient {{token}} balance. Try a smaller amount or choose a different token.",
     "fetchSwapQuoteFailed": "Sorry for the delay! It's taking longer than usual to get the best exchange rate. Please try again in a few minutes."
+  },
+  "swapContent": {
```
kathaypacific, line 1370: `swapTransactionItem`? having more specific translation key names (that kind of describe where in the app the key is used) helps our translators to translate more accurately

src/transactions/feed/SwapFeedItem.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+      inAmount: {
+        tokenAddress: mockCeurAddress,
+        value: 2.93,
+      },
+      outAmount: {
+        tokenAddress: mockCusdAddress,
+        value: 2.87,
+      },
+    })
+    expectDisplay({
```
kathaypacific, line 97: this abstraction seems a little odd to me, i think for tests it is okay to be a little verbose rather than try to abstract away logic.

as an aside, i would recommend simplifying this test to read something like "expect swapScreen.title to exist, expect [text] to exist, ...." and avoid looking for test id's. i understand that you are trying to test that the correct text appears in the correct "box" marked by the test id, but i think that still doesn't always guarantee that the display is correct anyway (e.g. if the styles are messed up, then even though the test-id's and texts match, they could still be positioned in the wrong order relative to other things / not visible to the user) so in that sense, i think it's not too important to test for.

i like this article https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-querying-by-text - maybe this helps. Kent (the author of this article) is one of the authors of the react testing library. this article is more geared towards web than mobile, but i think the concept of testing for what the user sees rather than technical implementation detail is still relevant :)

src/transactions/feed/SwapFeedItem.tsx
```diff
@@ -0,0 +1,114 @@
...
+      <View style={styles.container}>
+        <View style={styles.iconContainer}>
+          <SwapIcon />
+        </View>
+        <View style={styles.contentContainer}>
+          <Text style={styles.title} testID={'SwapFeedItem/title'}>
+            {t('swapScreen.title')}
+          </Text>
+          <Text style={styles.subtitle} testID={'SwapFeedItem/subtitle'}>
+            {outgoingTokenInfo?.symbol} {'to'} {incomingTokenInfo?.symbol}
```
kathaypacific, line 42: we should localise this, and not hardcode "to". perhaps better to localise the whole thing, so the translation value can be "{{token1}} to {{token2}}"

src/transactions/feed/SwapFeedItem.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  const openTransferDetails = () => {
+    navigate(Screens.TransactionDetailsScreen, { transaction: exchange })
+    ValoraAnalytics.track(HomeEvents.transaction_feed_item_select)
+  }
+
+  return (
+    <Touchable disabled={false} onPress={openTransferDetails}>
+      <View style={styles.container}>
+        <View style={styles.iconContainer}>
```
kathaypacific, line 34: i think you don't need this `View`?

src/transactions/feed/SwapFeedItem.tsx
```diff
@@ -0,0 +1,114 @@
...
+        </View>
+        <View style={styles.contentContainer}>
+          <Text style={styles.title} testID={'SwapFeedItem/title'}>
+            {t('swapScreen.title')}
+          </Text>
+          <Text style={styles.subtitle} testID={'SwapFeedItem/subtitle'}>
+            {outgoingTokenInfo?.symbol} {'to'} {incomingTokenInfo?.symbol}
+          </Text>
+        </View>
+        <View style={styles.contentContainer}>
```
kathaypacific, line 45: i would recommend to remove this style, so that this box and the description box are not always the same width.

i think you can then have a style on this box `alignItems: flex-end` and remove the right align styles on the TokenDisplay components (i.e. the marginLeft and textAlign).

src/transactions/feed/SwapFeedItem.tsx
```diff
@@ -0,0 +1,114 @@
...
+        </View>
+      </View>
+    </Touchable>
+  )
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flexDirection: 'row',
+    justifyContent: 'space-between',
```
kathaypacific, line 74: i think this style is not needed, if only the description container uses flex:1

src/transactions/feed/SwapFeedItem.tsx
```diff
@@ -0,0 +1,114 @@
...
+    paddingVertical: 12,
+    paddingHorizontal: variables.contentPadding,
+  },
+  iconContainer: {
+    flexDirection: 'column',
+    alignItems: 'center',
+  },
+  contentContainer: {
+    flex: 1,
+    marginLeft: variables.contentPadding,
```
kathaypacific, line 85: i would suggest paddingHorizontal here (if only the description container uses this style)

src/transactions/feed/SwapFeedItem.tsx
```diff
@@ -0,0 +1,114 @@
...
+  },
+  subtitle: {
+    ...fontStyles.small,
+    color: colors.gray4,
+    paddingTop: 2,
+  },
+  amount: {
+    ...fontStyles.regular500,
+    marginLeft: 'auto',
+    paddingLeft: 10,
```
kathaypacific, line 99: rather than put the paddingLeft in both these token amounts, i'd suggest to put it on the parent View

src/transactions/feed/SwapFeedItem.tsx
```diff
@@ -0,0 +1,114 @@
...
+        </View>
+        <View style={styles.contentContainer}>
+          <TokenDisplay
+            amount={exchange.inAmount.value}
+            tokenAddress={exchange.inAmount.tokenAddress}
+            showLocalAmount={false}
+            showSymbol={true}
+            showExplicitPositiveSign={true}
+            hideSign={false}
+            style={[styles.amount, { color: colors.greenUI }]}
```
kathaypacific, line 53: i think this can use `tokenAmount` for the style, and an extra style to create padding on the bottom + add the green color

src/transactions/feed/TransactionDetailsScreen.test.tsx
```diff
@@ -290,4 +321,34 @@ describe('TransactionDetailsScreen', () => {
     const subtotal = getByTestId('TotalLineItem/Subtotal')
     expect(getElementText(subtotal)).toEqual('10.10 cUSD')
   })
+
+  it('renders correctly for cUSD to cEUR swap', async () => {
```
kathaypacific, line 325: i feel like this test is the same as the `SwapFeedItem` test :/ do we need both?

src/transactions/feed/TransactionFeed.test.tsx
```diff
@@ -211,4 +251,17 @@ describe('TransactionFeed', () => {
...
+    mockFetch.mockResponse(JSON.stringify(SWAP_MOCK_RESPONSE))
+
+    const tree = renderScreen({})
+
+    await waitFor(() => tree.getByTestId('TransactionList'))
+
+    expect(tree.queryByTestId('NoActivity/loading')).toBeNull()
+    expect(tree.queryByTestId('NoActivity/error')).toBeNull()
+
+    expect(tree).toMatchSnapshot()
```
kathaypacific, line 265: we agreed a while ago (before you got here) to stop using snapshot tests. they're easy to write, but hard to review and easy to break in the future.

i'd suggest testing explicitly what you are looking for (although it feels like you are testing the same thing here compared to the tests in SwapFeedItem and TransactionDetailsScreen) or add the swap transaction to an existing mock response and capture the output there

src/transactions/feed/TransactionFeed.tsx
```diff
@@ -108,7 +109,11 @@ function TransactionFeed() {
   function renderItem({ item: tx }: { item: FeedTokenTransaction; index: number }) {
     switch (tx.__typename) {
       case 'TokenExchangeV2':
-        return <ExchangeFeedItem key={tx.transactionHash} exchange={tx} />
+        if (tx.type == 'SWAP_TRANSACTION') {
+          return <SwapFeedItem key={tx.transactionHash} exchange={tx} />
+        } else {
+          return <ExchangeFeedItem key={tx.transactionHash} exchange={tx} />
```
kathaypacific, line 115: i think it is worth asking Ellen whether we should keep both. from the way blockchain-api (should) work, the buy / sell CELO action should be now classified as a swap? i would advise to remove the existing ExchangeFeedItem but let's check with Ellen first

src/transactions/feed/detailContent/SwapContent.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const { t } = useTranslation()
+  const tokensList = useSelector(tokensListSelector)
+
+  const fromTokenSymbol = tokensList.find(
+    (token) => token.address === exchange.outAmount.tokenAddress
+  )?.symbol
+  const toTokenSymbol = tokensList.find((token) => token.address === exchange.inAmount.tokenAddress)
+    ?.symbol
+
+  //TODO: verifying fee part.
```
kathaypacific, line 33: TODO?

src/transactions/feed/detailContent/SwapContent.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const onPressTxDetails = () => {
+    navigate(Screens.WebViewScreen, {
```
kathaypacific, line 35: maybe add an analytics event here?

src/transactions/feed/detailContent/SwapContent.tsx
```diff
@@ -0,0 +1,133 @@
...
+          tokenAddress={exchange.inAmount.tokenAddress}
+          showLocalAmount={false}
+          showSymbol={true}
+          showExplicitPositiveSign={true}
+          hideSign={true}
+          testID="SwapContent/swapTo"
+        />
+      </View>
+      <View style={styles.separator} />
+      <View style={[styles.row, { paddingBottom: Spacing.Smallest8 }]}>
```
kathaypacific, line 68: can we just have `row` use one `paddingBottom` value? not sure i see the point of 8 vs 16, we can make it 12 for all if we want?

otherwise would advise to have `row` and `rowWithSpacing` rather than have inline styles like this

src/transactions/feed/detailContent/SwapContent.tsx
```diff
@@ -0,0 +1,133 @@
...
+  },
+  row: {
+    flex: 1,
+    flexDirection: 'row',
+    justifyContent: 'space-between',
+  },
+  exchangeBodyText: {
+    ...fontStyles.large,
+    width: '40%',
+    textAlign: 'left',
```
kathaypacific, line 112: is this style needed? i think left is the default

kathaypacific (commented):
Looks good so far 🥳

src/transactions/feed/detailContent/SwapContent.tsx
MuckT: The mocks have a space difference between the rows from what I was seeing 16 on the top two and 8 on the bottom ones.

src/transactions/feed/SwapFeedItem.tsx
MuckT: The `Touchable` from 'src/components/Touchable' only allows one child component.

src/transactions/feed/detailContent/SwapContent.tsx
```diff
@@ -0,0 +1,133 @@
...
+          <OpenLinkIcon color={colors.gray4} />
+        </View>
+      </Touchable>
+    </View>
+  )
+}
+
+const styles = StyleSheet.create({
+  contentContainer: {
+    flexShrink: 1,
```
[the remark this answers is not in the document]
MuckT: flex: 1 spaces everything out too much.

#### Screenshot

src/transactions/feed/detailContent/SwapContent.tsx
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2752/commits/22c9b2cad705288660aa858ccbf1e2e069ccafb8!

src/transactions/feed/detailContent/SwapContent.tsx
```diff
@@ -0,0 +1,133 @@
...
+  },
+  row: {
+    flex: 1,
+    flexDirection: 'row',
+    justifyContent: 'space-between',
+  },
+  exchangeBodyText: {
```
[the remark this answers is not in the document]
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2752/commits/22c9b2cad705288660aa858ccbf1e2e069ccafb8!

src/transactions/feed/SwapFeedItem.tsx
MuckT: I tried removing it as detailed above and when I used large fonts I noticed the display of the swap gets distorted.
#### Before and After the recommended change

src/transactions/feed/SwapFeedItem.tsx
kathaypacific: yeah but then you can use a fragment

src/transactions/feed/detailContent/SwapContent.tsx
kathaypacific: right but it's a bit of give an take with design - if changing a few pixels doesn't compromise the meaning of the design and can reduce complexity in the code, i think it's worth doing. it's too late now because Nitya is on holiday so we can keep it like this, but i do think this is the mentality we should carry going forward. it's in all of our best interest to keep our code (and styles) as simple as possible, when designing it's not always obvious what would cause code complexity and sometimes those parts of the design aren't important. so if we double check with the designer, at least we can be sure that any complexity that we do introduce is worth the effort

src/transactions/feed/SwapFeedItem.test.tsx
Samuel3Shin: Thanks for sharing the article. It was fun to read!

I removed the `expectDisplay()` function. I used it because the other feed item test codes used it.

I tried to do "expect swapScreen.title to exist", but I was not able to find a way to do it.

src/transactions/feed/SwapFeedItem.tsx
Samuel3Shin: I added localization!

src/transactions/feed/SwapFeedItem.tsx
Samuel3Shin: I removed `iconContainer` view because it didn't affect anything when the font is small or the font is extra large.

src/transactions/feed/SwapFeedItem.tsx
Samuel3Shin: If I change it according to your recommendation, it is not like the figma design.

Current design.
Screen Shot 2022-08-12 at 1 37 16 PM

Your recommendation.
Screen Shot 2022-08-15 at 2 17 30 PM

Figma design (https://www.figma.com/file/jRtYI8GiSvPAaClBnRLLpu/In-App-Swap?node-id=1471%3A13632)
Screen Shot 2022-08-15 at 2 15 58 PM

src/transactions/feed/SwapFeedItem.tsx
Samuel3Shin: I removed this!

src/transactions/feed/SwapFeedItem.tsx
Samuel3Shin: Currently, the Token amount container also uses this. So I will keep it like this!

src/transactions/feed/SwapFeedItem.tsx
Samuel3Shin: I added `tokenAmountContainer` and did what @kathaypacific recommended.
I also made it not to be screwed up with the extra large-size fonts!

src/transactions/feed/SwapFeedItem.tsx
Samuel3Shin: I added `tokenAmountContainer` and did what you said!

src/transactions/feed/TransactionDetailsScreen.test.tsx
Samuel3Shin: This test is for the detail page after you click the swap feed item, not the swap feed in the home screen.

src/transactions/feed/TransactionFeed.test.tsx
Samuel3Shin: Oh, I see. I removed the snapshot test. I will add an explicit test if I need it!

src/transactions/feed/detailContent/SwapContent.tsx
Samuel3Shin: I use the first element (index 0) in the `fees` array. And I think that `fees` array always has one element. But I put TODO here because I wanted to hear the others' opinions on it.

src/transactions/feed/detailContent/SwapContent.tsx
Samuel3Shin: I added swap_feed_detail_view_tx event here!

src/transactions/feed/detailContent/SwapContent.tsx
```diff
@@ -0,0 +1,133 @@
...
+          showLocalAmount={false}
+          showSymbol={true}
+          hideSign={true}
+          testID="SwapContent/estimatedFee"
+        />
+      </View>
+      <Touchable testID={'txDetails'} borderless={true} onPress={onPressTxDetails}>
+        <View style={styles.rowContainer}>
+          <Text style={styles.txDetails}>
+            {t('fiatConnectStatusScreen.withdraw.success.txDetails')}
```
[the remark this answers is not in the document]
Samuel3Shin: I see! I changed it to `t('swapContent.txDetails')`

src/transactions/feed/detailContent/SwapContent.tsx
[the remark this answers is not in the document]
Samuel3Shin: I will keep it `flexShrink: 1`

src/transactions/feed/TransactionFeed.tsx
Samuel3Shin: I removed it after discussing it with @ellenwoconnor

locales/base/translation.json
Samuel3Shin: I changed it to `swapTransactionDetailPage`

src/transactions/feed/detailContent/SwapContent.tsx
```diff
@@ -0,0 +1,132 @@
...
+          amount={exchange.fees[0].amount.value}
+          tokenAddress={exchange.fees[0].amount.tokenAddress}
+          showLocalAmount={false}
+          showSymbol={true}
+          hideSign={true}
+          testID="SwapContent/estimatedFee"
+        />
+      </View>
+      <Touchable testID={'txDetails'} borderless={true} onPress={onPressTxDetails}>
+        <View style={styles.rowContainer}>
```
kathaypacific, line 91: nit: i think that this style can be applied to the `Touchable` and then this can be replaced with a fragment like `<>`?

kathaypacific (approved):
Just a few minor comments from me, looks great!!! 🚀

src/transactions/feed/TransactionFeed.tsx
```diff
@@ -108,7 +108,7 @@ function TransactionFeed() {
   function renderItem({ item: tx }: { item: FeedTokenTransaction; index: number }) {
     switch (tx.__typename) {
       case 'TokenExchangeV2':
-        return <ExchangeFeedItem key={tx.transactionHash} exchange={tx} />
```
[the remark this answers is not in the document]
Samuel3Shin: I deleted this component!

src/transactions/feed/detailContent/SwapContent.tsx
Samuel3Shin: I did that!

src/transactions/feed/SwapFeedItem.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const incomingTokenInfo = useTokenInfo(exchange.inAmount.tokenAddress)
+  const outgoingTokenInfo = useTokenInfo(exchange.outAmount.tokenAddress)
+
+  const openTransferDetails = () => {
+    navigate(Screens.TransactionDetailsScreen, { transaction: exchange })
+    ValoraAnalytics.track(HomeEvents.transaction_feed_item_select)
+  }
+
+  return (
+    <Touchable disabled={false} onPress={openTransferDetails}>
```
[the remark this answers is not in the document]
Samuel3Shin: I removed `disabled` and changed the `openTransferDetails` to `handleTransferDetails`!

src/transactions/feed/TransactionFeed.tsx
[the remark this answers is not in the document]
Samuel3Shin: I restored those because deleting those messed up e2e tests

ValoraQA:
Hey @MuckT we have verified the above task on Latest iOS Test Flight Release build V 1.39.1 and updated status in Test Rails

Observation :

ValoraQA:
Hey @MuckT we have verified the above task on Latest iOS Test Flight Release build  V1.39.1 and updated status in Test Rails

**Observation** :
- Swap activity is seen on homepage when user buys / sells CELO

**Test Rail links :**
**iOS 15:** https://valoraapp.testrail.io/index.php?/tests/view/61843

**CC**: @Samuel3Shin

Thanks.!!

PR #2757, opened 2022-08-03
feat(cico): Refetch quotes from transfer failed screen

jh2oman:
### Description

When the user has a failed transaction, the 'back' button or 'try again' button should navigate to the Review Screen with a fresh quote.

https://user-images.githubusercontent.com/8432644/182728057-158a7959-c755-4ff4-8518-2a0add4c162e.mp4

### Other changes

Refactoring of sagas to share more code

### Tested

### How others should test

### Related issues

https://linear.app/valora/issue/ACT-364/back-button-try-again-around-the-review-screen


src/analytics/Properties.tsx
```diff
@@ -1017,6 +1025,10 @@ interface FiatExchangeEventsProperties {
     provider: string
     flow: CICOFlow
   }
+  [FiatExchangeEvents.cico_fc_review_error_contact_support]: {
+    provider: string
+    flow: CICOFlow
+  }
```
anthonydaoud, line 1031: I think this one is accidentally duplicated.

src/fiatconnect/saga.ts
```diff
@@ -174,18 +205,24 @@ function* _getQuote({
...
+    digitalAsset,
+    cryptoAmount,
+    country: userLocation?.countryCodeAlpha2 || 'US',
+    flow,
+    fiatConnectCashInEnabled,
+    fiatConnectCashOutEnabled,
+    fiatConnectProviders:
+      fiatConnectProviders?.filter((provider) => provider.id === providerId) ?? [],
+  })
   const normalizedQuotes = normalizeFiatConnectQuotes(flow, quotes)
```
anthonydaoud, line 226: More of a nit, but it might be worth making a helper function as it seems much of this logic is shared with `handleFetchFiatConnectQuotes` (except the providers). Otherwise the code looks pretty good!

anthonydaoud (approved):

src/fiatconnect/saga.ts
satish-ravi: +1 to this. so this function was previously using `handleFetchFiatConnectQuotes` via the action right? We cannot do that now because we have a different action for refetch quote? Wonder if we can structure it so that `handleFetchFiatConnectQuotes` calls `_getQuote` and also have a separate method to select a single quote to re-use in both `handleRefetchQuote` and `handleAttemptReturnUserFlow`?

ValoraQA:
Hey @MuckT since NA is mentioned in how other should test we are skipping this PR

Cc: @jh2oman 

Thanks.!!

PR #2767, opened 2022-08-09
feat: in app swap review screen

### LINKED ISSUE #2582: [In-app swaps] Build review swaps screen
### Description
Design https://www.figma.com/file/jRtYI8GiSvPAaClBnRLLpu/In-App-Exchange?node-id=1526%3A17066

### AC
- [ ] Build review swap screen as per design
- [ ] This screen should be triggered from the "review" button of the swap screen
- [ ] On tap of "estimated amount", the dialog should be displayed. The text inside is hard coded.
- [ ] The "complete swap" button does not need to do anything for this task

MuckT:
### Description

Implements an in app swap review screen using the [Valora-Rest-API](https://github.com/valora-inc/valora-rest-api) / [0x API](https://docs.0x.org/0x-api-swap/api-references).

#### Video
https://user-images.githubusercontent.com/26950305/184464020-87e5d2ea-d2e3-452d-b971-bf9e61cf4caa.mov

### Other changes

Added swap screen and swap review screen to stack navigators.
Added maxSwapSlippagePercentage to remote config values - to allow slippage adjustments from the remote config.

### Tested

Tested locally on iOS on mainnet as 0x API has no support for alfajores.

#### Developers 
prerequisite: set shouldShowSwapMenuInDrawerMenu used in `DrawerNavigator.tsx` to true.
1. Using a funded mainnet wallet navigate to the swap screen
2. Enter an amounts in to or from and selector tokens. (Testing with cREAL is a good idea as there were some issues previously.
3. Once amount is set within your wallet balance tap 'Review Swap'
4. On the review screen check amounts and estimated price modal.
5. Set your device text to large and ensure no text overlap occurs.

#### QA
Once this feature is cut and included in the a release we can enable it for select versions.
prerequisite: a funded mainnet wallet.
1. Navigate to the swap screen
2. Enter an amounts in to or from and selector tokens. (Testing with cREAL is a good idea as there were some issues previously.
3. Once amount is set within your wallet balance tap 'Review Swap'
4. On the review screen check amounts and estimated price modal.
5. Set your device text to large and ensure no text overlap occurs.

### Related issues

- Fixes #2582, RET-318

### Backwards compatibility

Yes


package.json
```diff
@@ -136,6 +137,7 @@
     "moment-timezone": "^0.5.23",
     "node-libs-react-native": "^1.2.0",
     "qrcode": "^1.4.4",
+    "qs": "^6.11.0",
```
kathaypacific, line 140: hm, is it possible to make this work without adding these dependencies?

src/swap/useSwapQuote.ts
```diff
@@ -52,10 +53,13 @@ const useSwapQuote = () => {
...
-      fromToken.address
-    }&${swapAmountParam}=${swapAmountInWei.toString().split('.')[0]}&userAddress=${walletAddress}`
-
+    const params = {
+      buyToken: toToken.address,
+      sellToken: fromToken.address,
+      [swapAmountParam]: swapAmountInWei.toString().split('.')[0],
+      userAddress: walletAddress,
+    }
+    const requestUrl = `${networkConfig.approveSwapUrl}?${qs.stringify(params)}`
```
kathaypacific, line 62: i'm not really convinced that the readability lift we get here is worth the extra 2 dependencies.

if we want to avoid the long string template literal, i'd suggest converting the params object to url params ourselves. something like
```
const requestUrl = `${networkConfig.approveSwapUrl}?${new URLSearchParams(Object.entries(params)).toString()}`
```

src/web3/networkConfig.ts
```diff
@@ -83,8 +82,8 @@ const CELO_EXPLORER_BASE_TX_URL_ALFAJORES = 'https://alfajores-blockscout.celo-t
 const CELO_EXPLORER_BASE_TX_URL_MAINNET = 'https://explorer.celo.org/tx/'
 
 const NFTS_VALORA_APP_URL = 'https://nfts.valoraapp.com/'
+
 const APPROVE_SWAP_URL = `${CLOUD_FUNCTIONS_MAINNET}/approveSwap`
-const ExECUTE_SWAP_URL = `${CLOUD_FUNCTIONS_MAINNET}/executeSwap`
```
kathaypacific, line 87: i think the typo can be fixed here but this shouldn't be removed, we'll need it later

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,331 @@
...
+
+export function SwapReviewScreen(props: Props) {
+  const { toToken, fromToken, swapAmount, updatedField } = props.route.params
+  const [swapInfo, setSwapInfo] = useState(null as any)
+  const [loading, setLoading] = useState(false)
+  const [fetchSwapQuoteError, setFetchSwapQuoteError] = useState(false)
+  const [estimatedModalVisible, setEstimatedDialogVisible] = useState(false)
+  const coreTokens = useSelector(coreTokensSelector)
+  const maxSlippagePercent = useSelector(maxSwapSlippagePercentageSelector)
+  const walletAddress = useSelector(accountAddressSelector)
```
kathaypacific, line 51: this should be `walletAddressSelector` instead - we don't want the mtw address here.

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,331 @@
...
+  }, [fetchSwapQuoteError])
+
+  useEffect(() => {
+    ValoraAnalytics.track(SwapEvents.swap_review_screen_open, {
+      toToken,
+      fromToken,
+      swapAmount,
+    })
+
+    void getSwapInfo().then((data) => {
```
kathaypacific, line 77: i think we can use `useAsync` instead of doing this, a simple example is in i18nGate

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,331 @@
...
+        throw new Error(
+          `Failure response fetching token swap quote. ${response.status}  ${response.statusText}`
+        )
+      }
+      return await response.json()
+    } catch (error) {
+      setFetchSwapQuoteError(true)
+      Logger.error(TAG, `Error fetching token swap quote: ${error}`)
+    } finally {
+      setTimeout(() => {
```
kathaypacific, line 107: are we purposely delaying the loading state because we don't want to flash the loading indicator? i don't think this should be a pattern that we keep using because it slows the app down for no functional reason...

in the future it'd be nice to consider building a generic loading component that only displays after x seconds if the loading state is still true

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,331 @@
...
+    }
+  }
+
+  return (
+    <SafeAreaView style={styles.safeAreaContainer}>
+      <CustomHeader
+        title={t('swapReviewScreen.title')}
+        left={<BackButton />}
+        // handle large text size to avoid overlaps in header
+        right={
```
kathaypacific, line 119: maybe nice to check with Nitya one more time about whether we need this right button at all, as it doesn't exist in any of the other flows that are triggered from the drawer menu....

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,331 @@
...
+            <TopBarTextButton
+              title={t('cancel')}
+              onPress={navigateHome}
+              titleStyle={{ color: colors.gray4 }}
+            />
+          )
+        }
+      />
+      <DisconnectBanner />
+      {loading || fetchSwapQuoteError ? (
```
kathaypacific, line 137: i would recommend against this pattern of show loading OR component. it causes flashes and content to jump around, it'd be better to show the loader at the top but still display the content (if there is any) always.

also do we want to display the loading state when there is `fetchSwapQuoteError`?

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,331 @@
...
+      const response = await fetch(requestUrl)
+      if (!response.ok) {
+        Logger.error(TAG, `Failure response fetching token Swap: ${response}`)
+        throw new Error(
+          `Failure response fetching token swap quote. ${response.status}  ${response.statusText}`
+        )
+      }
+      return await response.json()
+    } catch (error) {
+      setFetchSwapQuoteError(true)
```
kathaypacific, line 104: right now the only way to clear this error is to go out of this screen and back right? i think it's worth to add a pull to refresh (we can add a `onRefresh` and `refreshControl` to the scroll view and trigger the fetch again)

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,331 @@
...
+    })
+
+    void getSwapInfo().then((data) => {
+      setSwapInfo(data)
+    })
+  }, [])
+
+  const getSwapInfo = async () => {
+    try {
+      setLoading(true)
```
kathaypacific, line 84: would recommend to also clear any previous fetch errors here

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,331 @@
...
+        buyToken: toToken,
+        sellToken: fromToken,
+        [swapAmountParam]: swapAmountInWei.toString().split('.')[0],
+        slippagePercentage: maxSlippageDecimal,
+        userAddress: walletAddress,
+      }
+      const requestUrl = `${networkConfig.approveSwapUrl}?${qs.stringify(params)}`
+      const response = await fetch(requestUrl)
+      if (!response.ok) {
+        Logger.error(TAG, `Failure response fetching token Swap: ${response}`)
```
kathaypacific, line 97: we should also set the fetch error here?

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,331 @@
...
+      <DisconnectBanner />
+      {loading || fetchSwapQuoteError ? (
+        <View style={styles.loadingContentContainer}>
+          <LoadingSpinner />
+        </View>
+      ) : (
+        <ScrollView style={styles.contentContainer}>
+          <View style={styles.subContentContainer}>
+            <View style={styles.tallRow}>
+              <View style={styles.column}>
```
kathaypacific, line 145: one quick simplification you can try, is removing the `column` style entirely (and corresponding View's) and adding `flex: 1` to the `label` and `transactionDetailsRightText` and `tokenDisplayView` styles

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,331 @@
...
+  label: {
+    ...fontStyles.regular,
+  },
+  tokenText: {
+    alignSelf: 'flex-end',
+    textAlign: 'right',
+    ...fontStyles.large,
+  },
+  tokenSubText: {
+    alignSelf: 'flex-end',
```
kathaypacific, line 320: i don't think you need `alignSelf` in this or the `tokenText` styles

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,331 @@
...
+  },
+  tokenDisplayView: {
+    justifyContent: 'space-between',
+  },
+  separator: {
+    height: 1,
+    width: '100%',
+    backgroundColor: colors.gray2,
+  },
+  // No shared font styles for this label - done manually
```
kathaypacific, line 290: can we use some base style from the font styles (just the closest one) and update the fields we need? at least this way we can see what is pretty much supposed to look like which styles are overwritten, rather than a whole new font style. (i'd also push back on adding new styles like this, since we have a design system which would have no purpose if we create custom styles whenever we want)

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,331 @@
...
+                    new BigNumber(swapInfo?.unvalidatedSwapTransaction?.price)
+                  )} ${toTokenSymbol}`}
+                </Text>
+              </View>
+            </View>
+            <View style={styles.row}>
+              <View style={styles.column}>
+                <Text style={styles.label}>{t('swapReviewScreen.estimatedGas')}</Text>
+              </View>
+              <View style={[styles.column, styles.tokenDisplayView]}>
```
kathaypacific, line 214: i think this View is not needed

kathaypacific (commented):
looking good so far!

src/swap/SwapReviewScreen.tsx
MuckT: Currently if there is a `fetchSwapQuoteError` we take the user back to the previous page. I like the idea of adding a refreshControl and allow users to manually trigger the fetch again.

package.json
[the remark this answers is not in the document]
MuckT: removed in https://github.com/valora-inc/wallet/pull/2767/commits/70a84c7f3df4ca51bc90fa976497ba6494dbf1f4!

src/swap/useSwapQuote.ts
MuckT: Removed and replaced with something similar in https://github.com/valora-inc/wallet/pull/2767/commits/70a84c7f3df4ca51bc90fa976497ba6494dbf1f4!

src/web3/networkConfig.ts
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2767/commits/5af72ce629f630464f84e858b712a525b9398848!

src/analytics/Properties.tsx
```diff
@@ -1278,6 +1278,11 @@ interface SwapEventsProperties {
     tokenSymbol: string
   }
   [SwapEvents.swap_screen_review_swap]: undefined
+  [SwapEvents.swap_review_screen_open]: {
+    toToken: string
+    fromToken: string
+    swapAmount: SwapAmount
```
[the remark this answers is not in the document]
MuckT: Changed in [f71056c](https://github.com/valora-inc/wallet/pull/2767/commits/f71056c554becef6f1f9e5bcff2806514c73bbd9)!

src/swap/SwapReviewScreen.tsx
MuckT: fixed in https://github.com/valora-inc/wallet/pull/2767/commits/90d9446a7b9a6a9369bfc81b9d2ca3a181447742!

src/swap/SwapReviewScreen.tsx
MuckT: Checked with Nitya and was removed in https://github.com/valora-inc/wallet/pull/2767/commits/70960e345f0debe2cbefb3fce5e70ef19e3fddca and https://github.com/valora-inc/wallet/pull/2767/commits/af577274eaabef8067b4a9f956e4242da13f1df2!

src/swap/SwapReviewScreen.tsx
MuckT: Styles adjusted in https://github.com/valora-inc/wallet/pull/2767/commits/70960e345f0debe2cbefb3fce5e70ef19e3fddca!

src/swap/SwapReviewScreen.tsx
MuckT: Styles adjusted in https://github.com/valora-inc/wallet/pull/2767/commits/70960e345f0debe2cbefb3fce5e70ef19e3fddca!

src/swap/SwapReviewScreen.tsx
MuckT: Styles adjusted in https://github.com/valora-inc/wallet/pull/2767/commits/70960e345f0debe2cbefb3fce5e70ef19e3fddca!

src/swap/SwapReviewScreen.tsx
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2767/commits/70960e345f0debe2cbefb3fce5e70ef19e3fddca!

src/swap/SwapReviewScreen.tsx
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2767/commits/92210a27a79db57c328df9fcd1506bd585bad3ef!

src/swap/useSwapQuote.ts
```diff
@@ -52,10 +52,14 @@ const useSwapQuote = () => {
...
-      fromToken.address
-    }&${swapAmountParam}=${swapAmountInWei.toString().split('.')[0]}&userAddress=${walletAddress}`
-
+    const params = {
+      buyToken: toToken.address,
+      sellToken: fromToken.address,
+      [swapAmountParam]: swapAmountInWei.toString().split('.')[0],
+      userAddress: walletAddress,
+    }
+    const queryParams = new URLSearchParams({ ...(params as Record<string, string>) }).toString()
```
kathaypacific, line 61: let's refrain from type casting as much as possible and let TS do work for us....i think the type error here that is silenced by the cast is genuine - `walletAddress` can be null and we don't handle this case. 

not sure how best to handle this, maybe the `refreshQuote` function can do an early return if `walletAddress` is not defined? it should never happen anyway but if we did this then this `as Record<string, string>` can be safely removed.

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,374 @@
...
+    try {
+      setFetchSwapQuoteError(false)
+      setLoading(true)
+      const swapAmountInWei = multiplyByWei(swapAmount[updatedField]!)
+      const swapAmountParam = updatedField === Field.FROM ? 'sellAmount' : 'buyAmount'
+      const params = {
+        buyToken: toToken,
+        sellToken: fromToken,
+        [swapAmountParam]: swapAmountInWei.toString().split('.')[0],
+        slippagePercentage: maxSlippageDecimal,
```
kathaypacific, line 101: since slippage doesn't work on the backend yet, should we remove this + the remote config for it?

i'm thinking that without being able to test end to end how the slippagePercentage works on the backend (and seeing what kind of errors we get), we can't know that our implementation here works and that we've handled errors properly. so chances are, we can't just add slippagePercentage to the backend and expect for this version of the client to work out of the box. it could also be a long time before slippage is supported in the backend. in this case i'd lean towards removing this and re-adding when the need occurs, what do you think?

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,374 @@
...
+    } catch (error) {
+      setFetchSwapQuoteError(true)
+      Logger.error(TAG, `Error fetching token swap quote: ${error}`)
+    } finally {
+      setLoading(false)
+    }
+  }
+
+  // Get swap info on load
+  useAsync(getSwapInfo, [])
```
kathaypacific, line 124: what do you think about the following idea:

- we use all the stuff that `useAsync` gives us out of the box, i.e. the loading / data / error return variables and the error handler
- we introduce a state variable (something like `shouldFetch`) and add this as a dependency for the `useAsync` hook. the refresh control can also toggle this variable to trigger a refetch.

the benefits of this approach would be less state variables (don't need `loading`, `fetchSwapQuoteError`, `swapInfo` anymore) and less manual coordination of what the loading / error state is. we could also display the error message in the error handler of `useAsync` and remove the `useEffect` currently used to achieve this. i guess the examples i'm thinking of are `queryHelper.ts` and `I18nGate.tsx`, where by using `useAsync` we no longer need to manually track the loading / data / error variables in our own state variables.

 happy to catch up about this i didn't explain this idea clearly

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,374 @@
...
+                  style={[styles.amountText, { color: colors.greenUI }]}
+                  amount={divideByWei(
+                    swapInfo?.unvalidatedSwapTransaction?.buyAmount -
+                      swapInfo?.unvalidatedSwapTransaction?.gas
+                  )}
+                  tokenAddress={toToken}
+                  showLocalAmount={false}
+                  testID={'ToSwapAmountToken'}
+                />
+                <TouchableOpacity
```
kathaypacific, line 193: can we use `Touchable` here? you may need to add a fragment around the children

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,374 @@
...
+    </SafeAreaView>
+  )
+}
+
+const styles = StyleSheet.create({
+  safeAreaContainer: {
+    flex: 1,
+  },
+  loadingContentContainer: {
+    justifyContent: 'flex-start',
```
kathaypacific, line 300: i think this is the default and not needed?

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,374 @@
...
+  }
+
+  // Get swap info on load
+  useAsync(getSwapInfo, [])
+
+  return (
+    <SafeAreaView style={styles.safeAreaContainer}>
+      <CustomHeader title={t('swapReviewScreen.title')} left={<BackButton />} />
+      <DisconnectBanner />
+      {(loading && swapInfo === null) || fetchSwapQuoteError ? (
```
kathaypacific, line 130: if for whatever reason the user sees the swap details and accidentally pulls to refresh and gets an error, it's strange that the content would be cleared...is it enough to display the error banner but keep whatever existing content still displayed?

i'd suggest doing simplifying the display to be something like 
```
<ScrollView> 
{loading && ( show loading)}
{swapInfo && (<> the swap info content </>)}
</ScrollView>
```

what do you think?

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,374 @@
...
+  },
+  separator: {
+    height: 1,
+    width: '100%',
+    backgroundColor: colors.gray2,
+  },
+  transactionDetailsRightText: {
+    ...fontStyles.regular,
+    flex: 1,
+    alignSelf: 'flex-end',
```
kathaypacific, line 334: i think `alignSelf` and `textAlignVertical` are not needed here?

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,374 @@
...
+    ...fontStyles.small,
+    color: colors.gray4,
+  },
+  feeWaived: {
+    textDecorationLine: 'line-through',
+    textDecorationStyle: 'solid',
+  },
+})
+
+SwapReviewScreen.navOptions = {
```
kathaypacific, line 370: do we need this?

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,374 @@
...
+        setFetchSwapQuoteError(true)
+        Logger.error(TAG, `Failure response fetching token Swap: ${response}`)
+        throw new Error(
+          `Failure response fetching token swap quote. ${response.status}  ${response.statusText}`
+        )
+      }
+      setSwapInfo(await response.json())
+    } catch (error) {
+      setFetchSwapQuoteError(true)
+      Logger.error(TAG, `Error fetching token swap quote: ${error}`)
```
kathaypacific, line 117: i don't think these should be logger errors right? we only want to surface unhandled or dangerous errors in Sentry, for handled errors like this we shouldn't need it? we'll be aware of the errors from the cloud function side anyway

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,374 @@
...
+      const swapAmountInWei = multiplyByWei(swapAmount[updatedField]!)
+      const swapAmountParam = updatedField === Field.FROM ? 'sellAmount' : 'buyAmount'
+      const params = {
+        buyToken: toToken,
+        sellToken: fromToken,
+        [swapAmountParam]: swapAmountInWei.toString().split('.')[0],
+        slippagePercentage: maxSlippageDecimal,
+        userAddress: walletAddress,
+      }
+      const queryParams = new URLSearchParams({ ...(params as Record<string, string>) }).toString()
```
kathaypacific, line 104: same comment as in the hook, we shouldn't need the type cast for `params` if we handle the null `walletAddress` case :)

kathaypacific (commented):
looking good! i had a few small questions and suggestions :)

src/swap/SwapReviewScreen.tsx
MuckT: We can remove it from the request to the API, but I think we should keep it in the remote config. So that the information modal and the check on performing the swap are pointing to the same value.

src/swap/SwapReviewScreen.tsx
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2767/commits/d84c6544c21453e841b1e21842f0d25e2d211f22!

src/swap/SwapReviewScreen.tsx
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2767/commits/7d463246666579c3467bbcc2e82dff8142660a72!

src/swap/SwapReviewScreen.tsx
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2767/commits/a4c7a335213039b2faef30dea75fa44f1b427aa5!

src/swap/SwapReviewScreen.tsx
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2767/commits/2f5a3f47681caf8a92fc216536f3e63487a8e96c!

src/swap/SwapReviewScreen.tsx
MuckT: Commented out in https://github.com/valora-inc/wallet/pull/2767/commits/71422b46fbe4ea741a194898d91108121a2495dd, but left some comments on when it should be enabled again.

src/swap/SwapReviewScreen.tsx
MuckT: Removed in https://github.com/valora-inc/wallet/pull/2767/commits/7d463246666579c3467bbcc2e82dff8142660a72!

src/swap/SwapReviewScreen.tsx
MuckT: Changed in https://github.com/valora-inc/wallet/pull/2767/commits/e286b05c2e78276f2027576d09ec0073d3b41ab4. I still had to use a fetchError state to disable the swap button if the fetch had completed with an error, and a swapInfo in state, but the loading state is now handled by the useAsync.

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,364 @@
...
+  useEffect(() => {
+    setShouldFetch(true)
+    ValoraAnalytics.track(SwapEvents.swap_review_screen_open, {
+      toToken,
+      fromToken,
+      buyAmount: swapAmount[Field.TO] ?? '',
+    })
+  }, [])
+
+  useAsync(
```
kathaypacific, line 79: you already throw an error inside the promise of `useAwait` when the response returns unsuccessfully, so i think that you can also reply on the `error` return variable from `useAsync` to handle errors instead of a duplicate `fetchError` state variable.

src/swap/SwapReviewScreen.tsx
kathaypacific: ah, i was confused about why the SwapScreen doesn't need this but turns out the swap screen is a drawer screen. all good!

src/swap/SwapReviewScreen.tsx
MuckT: I tried this method out and it ended up causing a flicker on the Complete Swap button as seen below in the video.

https://user-images.githubusercontent.com/26950305/185216326-c71055cc-3957-4b7f-b438-8e0b683f76b6.mov

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,364 @@
...
+
+  const { t } = useTranslation()
+  const dispatch = useDispatch()
+
+  // Token Symbols
+  const toTokenSymbol = coreTokens.find((token) => token.address === toToken)?.symbol
+  const fromTokenSymbol = coreTokens.find((token) => token.address === fromToken)?.symbol
+
+  useEffect(() => {
+    setShouldFetch(true)
```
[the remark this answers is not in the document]
MuckT: Fixed in [4ee2bcc](https://github.com/valora-inc/wallet/pull/2767/commits/4ee2bcc49035d87661339b642157680a038c5b4b)!

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,364 @@
...
+      setShouldFetch(false)
+      setFetchError(false)
+    },
+    [shouldFetch],
+    {
+      onError: (error) => {
+        setShouldFetch(false)
+        setFetchError(true)
+        dispatch(showError(ErrorMessages.FETCH_SWAP_QUOTE_FAILED))
+        Logger.error(TAG, 'Error while fetching transactions', error)
```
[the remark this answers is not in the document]
MuckT: Fixed in ccf67d9cdbf68e1b730a5e64cebb008a41d09360!

src/swap/SwapReviewScreen.tsx
```diff
@@ -0,0 +1,364 @@
...
+type Props = StackScreenProps<StackParamList, Screens.SwapReviewScreen>
+
+const TAG = 'SWAP_REVIEW_SCREEN'
+
+export function SwapReviewScreen(props: Props) {
+  const { toToken, fromToken, swapAmount, updatedField } = props.route.params
+  const [shouldFetch, setShouldFetch] = useState(false)
+  const [estimatedModalVisible, setEstimatedDialogVisible] = useState(false)
+  const [swapFeeModalVisible, setSwapFeeModalVisible] = useState(false)
+  const [swapInfo, setSwapInfo] = useState(null as any)
```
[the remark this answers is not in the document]
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2767/commits/be392ad1371ebf101b0d46a3cc9a83e451f63ebd!

src/swap/useSwapQuote.ts
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2767/commits/3873bbbc5c825bb150c56a25f594f4b4d3025244!

src/swap/SwapReviewScreen.tsx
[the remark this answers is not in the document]
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2767/commits/2b506440eb35d1e8228bebd335268244daa7529e!

kathaypacific: ah, you're right. i forgot that the error is cleared when the hook is used again by switching `shouldFetch` to false. how annoying :( let's leave it like this then.

as i'm typing this i'm realising that i encountered this exact problem when implementing the previous screen (like how to persist loading and error states across fetches) - and i remember that's why i built the `useSwapQuote` hook. 🥲 sorry it's just clicked for me. i need to modify the hook a bit so that it can be reused here (mostly how to deal with the `requestUrlRef` which i put in to deal with an edge case specific to the swap amount scree), but i think we can eventually do something like `const { swapInfo, refreshQuote, fetchSwapQuoteError, fetchingSwapQuote } = useSwapQuote()` in this component. this would mean you don't need the useAsync hook here at all and can reuse the logic for fetching from 0x. 

but let's tackle this later, this is good for now!

kathaypacific (approved):
nice work!!!! super excited about this :)

ValoraQA:
Hey @MuckT  currently we are not able to see SWAP functionality on mainnet build , please let us know once the feature is enabled on mainnet 
Thanks!

PR #2774, opened 2022-08-10
feat(Onboarding): create account copy test (RET-363)

Samuel3Shin:
### Description

This is a simple copy test controlled by remote config `createAccountCopyTestConfig`.

There are three conditions:

Control: `createAccountCopyTestConfig` = "ACCOUNT"
- CTA text: “Create new account” / “Restore my account”
- Step 1 title: “Create account”
Screen Shot 2022-08-10 at 3 23 26 PM Screen Shot 2022-08-10 at 3 23 40 PM

Treatment 1: `createAccountCopyTestConfig` = "WALLET"
- CTA text: “Create new wallet” / “Restore wallet”
- Step 1 title: “Create profile”
Screen Shot 2022-08-10 at 3 28 37 PM Screen Shot 2022-08-10 at 3 28 49 PM

Treatment 2: `createAccountCopyTestConfig` = "ALREADY_HAVE_WALLET"
- “Create new wallet” / “I already have a wallet”
- Step 1 title: “Create profile”
Screen Shot 2022-08-10 at 3 30 42 PM Screen Shot 2022-08-10 at 3 30 52 PM

### Other changes
N/A

### Tested
With iOS emulator

### Related issues
- https://linear.app/valora/issue/RET-363/onboarding-create-account-copy-test

### Backwards compatibility
Yes

### DIFF
--- a/src/firebase/remoteConfigValuesDefaults.ts
+++ b/src/firebase/remoteConfigValuesDefaults.ts
@@ -73,5 +73,5 @@ export const REMOTE_CONFIG_VALUES_DEFAULTS: Omit<
   coinbasePayEnabled: false,
   showSwapMenuInDrawerMenu: false,
   shouldShowRecoveryPhraseInSettings: false,
-  createAccountCopyTestConfig: null,
+  createAccountCopyTestConfig: 'control',
 }
--- a/src/firebase/remoteConfigValuesDefaults.ts
+++ b/src/firebase/remoteConfigValuesDefaults.ts
@@ -1,5 +1,5 @@
 import { RemoteConfigValues } from 'src/app/saga'
-import { SuperchargeButtonType } from 'src/app/types'
+import { CreateAccountCopyTestType, SuperchargeButtonType } from 'src/app/types'
 import { DEFAULT_SENTRY_NETWORK_ERRORS, DEFAULT_SENTRY_TRACES_SAMPLE_RATE } from 'src/config'
 import { DappConnectInfo } from 'src/dapps/types'
 import { PaymentDeepLinkHandler } from 'src/merchantPayment/types'
@@ -73,5 +73,5 @@ export const REMOTE_CONFIG_VALUES_DEFAULTS: Omit<
   coinbasePayEnabled: false,
   showSwapMenuInDrawerMenu: false,
   shouldShowRecoveryPhraseInSettings: false,
-  createAccountCopyTestConfig: 'control',
+  createAccountCopyTestType: CreateAccountCopyTestType.Account,
 }
--- a/src/firebase/remoteConfigValuesDefaults.ts
+++ b/src/firebase/remoteConfigValuesDefaults.ts
@@ -73,4 +73,5 @@ export const REMOTE_CONFIG_VALUES_DEFAULTS: Omit<
   coinbasePayEnabled: false,
   showSwapMenuInDrawerMenu: false,
   shouldShowRecoveryPhraseInSettings: false,
+  createAccountCopyTestConfig: null,
 }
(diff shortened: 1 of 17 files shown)


src/firebase/remoteConfigValuesDefaults.ts
```diff
@@ -73,4 +73,5 @@ export const REMOTE_CONFIG_VALUES_DEFAULTS: Omit<
   coinbasePayEnabled: false,
   showSwapMenuInDrawerMenu: false,
   shouldShowRecoveryPhraseInSettings: false,
+  createAccountCopyTestConfig: 'control',
```
kathaypacific, line 76: can we please use an enum for this? like how `superchargeButtonType` does it :)

src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -42,11 +42,19 @@ function NameAndPicture({ navigation }: Props) {
...
+
   useLayoutEffect(() => {
     navigation.setOptions({
       headerTitle: () => (
         <HeaderTitleWithSubtitle
-          title={t(choseToRestoreAccount ? 'restoreAccount' : 'createAccount')}
+          title={t(
+            choseToRestoreAccount
+              ? 'restoreAccount'
+              : createAccountCopyTestConfig === 'control'
```
kathaypacific, line 54: i think as a rule, we should check for the variants rather than the control. we're confident in showing the control (since it is the current released version) and we only want to show the variant for a specific value of remote config. now say someone makes a typing mistake, and writes "contrl" (lol) - in this case, we would accidentally show the variant. <- a bit of a contrived example but i guess the point i make is, we want to make it hard to get the variant, and everything except specific values should fall back on the control because it is safer this way

src/onboarding/welcome/Welcome.tsx
```diff
@@ -54,15 +57,25 @@ export default function Welcome() {
...
           style={styles.createAccountButton}
           testID={'CreateAccountButton'}
         />
         <Button
           onPress={onPressRestoreAccount}
-          text={t('welcome.restoreAccount')}
+          text={
+            createAccountCopyTestConfig === 'control'
+              ? t('welcome.restoreAccount')
+              : createAccountCopyTestConfig === 'treatment1'
```
kathaypacific, line 75: i know it's a little arbitrary what each variant is doing, but i still think it will be easier to understand if we try to give each variant a name. like "account" / "wallet" / "alreadyHaveWallet"

src/onboarding/registration/NameAndPicture.test.tsx
```diff
@@ -65,4 +66,82 @@ describe('NameAndPictureScreen', () => {
...
+
+    render(
+      <Provider store={store}>
+        <NameAndPicture {...mockScreenProps} />
+      </Provider>
+    )
+
+    const { queryByText } = render(<Provider store={store}>{headerTitle}</Provider>)
+
+    expect(queryByText('createAccount')).toBeTruthy()
```
kathaypacific, line 93: minor but if we expect something to be there, generally using the `getBy*` selectors are better because they throw more meaningful errors when the element is not found. generally i use `queryBy*` only when asserting that things shouldn't be there

kathaypacific (commented):
looking great :)

src/onboarding/registration/NameAndPicture.test.tsx
Samuel3Shin: Thanks for the explanation! I changed it to getBy*

src/onboarding/welcome/Welcome.tsx
Samuel3Shin: I did as you said!

src/onboarding/registration/NameAndPicture.tsx
Samuel3Shin: I did as you said!

src/firebase/remoteConfigValuesDefaults.ts
Samuel3Shin: Yup, I made a type as `superchargeButtonType`

kathaypacific (approved):
yay looks great!! 🥳

MuckT:
@kathaypacific it looks like this experiment might need to be configured still for Mainnet.

PR #2779, opened 2022-08-11
fix(homefeed): ensure that users can scroll to load more

kathaypacific:
### Description

Context: https://valora-app.slack.com/archives/C018VLA3YAK/p1660091022594799

The solution chosen for this issue is to continue to fetch transactions from the client until a minimum threshold is met (currently 10) so that the section list is filled and the user can scroll to load more.

In this PR:
- To ensure no weird bugs and avoid concurrency issues, i refactored the state variables to keep the cumulative transactions + pageInfo in the same state variable so that they are always in sync. 
- Added a hook to refetch transactions until we have 10, and refetch transactions if the returned page is empty
- Updated tests

### Other changes

N/A

### Tested

Manually and with unit tests

### How others should test

Please try this with many accounts that have varying number of transactions - load the app (from a fresh install, and from cold start) and ensure that the home feed displays transactions. If the account has many transactions, ensure that you can scroll down to load more transactions.

### Related issues

N/A

### Backwards compatibility

Yes

PR #2783, opened 2022-08-12
fix(homefeed): avoid showing 'no more transactions' toast when too few transactions

kathaypacific:
### Description

Annoyingly, the `onEndReached` callback is triggered on the SectionList used to render the transaction feed when navigating away from the screen on iOS, when the transaction list is too short. This causes the "no more transactions" toast to appear on the next page that is navigated to, but only once. 

This check for length of transactions used to exist, and I had removed it thinking it was strange https://github.com/valora-inc/wallet/pull/2779#discussion_r944308531 - but it seems like we need it.

### Other changes

N/A

### Tested

Manually

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/src/transactions/feed/queryHelper.ts
+++ b/src/transactions/feed/queryHelper.ts
@@ -178,7 +178,10 @@ export function useFetchTransactions(): QueryHookResult {
     fetchMoreTransactions: () => {
       if (!fetchedResult.pageInfo) {
         dispatch(showError(ErrorMessages.FETCH_FAILED))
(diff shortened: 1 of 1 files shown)


dievazqu (approved):

ValoraQA:
Hey @MuckT since NA is mentioned in how others should test section ,we are skipping this PR

Cc: @kathaypacific 
Thanks.!

PR #2784, opened 2022-08-12
feat(fiatconnect): add api key

anthonydaoud:
### Description

Currently, the wallet doesn't support Fiatconnect API keys. FiatConnect providers rely on these keys to determine where to send Webhook requests, so we currently can't monitor transfer or KYC statuses. This PR adds this functionality to the wallet.

### Other changes

N/A

### Tested

Unit tests, manual

### How others should test

N/A

### Related issues

- Fixes [ACT-365](https://linear.app/valora/issue/ACT-365/support-client-api-keys-for-fiatconnect-apis)

### Backwards compatibility

N/A


src/fiatconnect/clients.ts
```diff
@@ -21,7 +21,8 @@ export function getSigningFunction(wallet: UnlockableWallet): (message: string)
 
 export async function getFiatConnectClient(
   providerId: string,
-  providerBaseUrl: string
+  providerBaseUrl: string,
+  providerApiKey?: string
 ): Promise<FiatConnectApiClient> {
   if (!fiatConnectClients[providerId] || fiatConnectClients[providerId].url !== providerBaseUrl) {
```
anthonydaoud, line 27: I was planning to add an apiKey check here too, but I'm not sure that the second check actually does anything currently. Testing with manually changing the URL didn't update the baseUrl. I believe this is because getting new providers only occurs when the FiatExchange component mounts. It doesn't appear to be called again if you exit the add/withdraw flow and return to it later (or exit the app for that matter). 

I guess my question is when does the FiatExchange component unmount? I was only able to get it to unmount with a swipe up to fully close out the app.

jh2oman: I think the intention of this check is to create a new client when (if) we update the `url` field in `valora-rest-api`. Probably good to also create a new client if apiKey changes too so I'd add a new check here for that as well

anthonydaoud: Yeah I just added the check, but what I'm getting at is that I'm not sure that the code will ever hit the `url` (or for that matter `apiKey`) check. This is because fetching providers, via the `fetchFiatConnectProviders` function, only occurs when the FiatExchange component mounts. Any time that `getFiatConnectClient` is called, the wallet hasn't made a new request to `fetchFiatConnectProviders`. Instead, it just reuses the values it obtained from the first fetch. This means that providerBaseUrl and providerApiKey won't be different until FiatExchange un-mounts then re-mounts.

If `fetchFiatConnectProviders` were called as part of an error handling scenario, that'd be different I suppose.

jh2oman: Hmm, I see what you're saying. I'd expect the FiatExchange component to unmount if you go to a different part of the app and then come back to the cico flow. does it not?

anthonydaoud: Not when I was toying with the iOS simulator. It only called `fetchFiatConnectProviders` again after I swiped up to remove the app from running in the background. Even exiting the app and re-entering it did not cause `fetchFiatConnectProviders` to be re-called.

jh2oman: its probably outside the scope of this PR but we should consider moving that to be called more often

jh2oman (approved):

ValoraQA:
Hey @MuckT  since NA is mentioned in how others should test section hence we are skipping this PR

Cc: @anthonydaoud 
Thanks.!

PR #2785, opened 2022-08-12
WIP: Cash-in support for cREAL

XrayBravoGolf:
### Description

Celo Brazilian Reales cash-in are supported through this PR

#### Compromise
Due to changing the types in `currencies.ts` or upgrading `@celo/contractkit` would substantially magnify the scope of changes, places where cREAL is added have their types changed to `Currency | 'cREAL'`, and accompanying logic added. The type is asserted back into `Currency` when it is determined that the app is not doing cREAL cash-in flow.

### Other changes
None

### Tested

(To be written)

### How others should test

(To be written)

### Related issues

- Fixes ACT-390

### Backwards compatibility

Compatible.
A user with previous version of wallet will not be sending any cReal related interacions with backend services; related backend services are expected in [this PR](https://github.com/valora-inc/cloud-functions/pull/45) to handle non cReal requests in the same way.


src/fiatExchanges/ExternalExchanges.tsx
```diff
@@ -71,7 +71,10 @@ function ExternalExchanges({ route }: Props) {
         <>
           <Text style={styles.pleaseSelectExchange}>
             {t('youCanTransferIn', {
-              digitalAsset: CURRENCIES[route.params.currency].cashTag,
```
cajubelt, line 74: I see why you're not putting cREAL in the CURRENCIES object, since that has bigger consequences across the app, in particular making cREAL transactions visible in the home screen.

However, it would be nice if there was a todo list of cleanup that we need to do as a fast-follow to this PR, otherwise I worry we'll end up with a bunch of tech debt and a fading memory of what we actually need to do to address it. 

Maybe a Linear project like "enable cREAL across the app" and an issue within it that says "make cREAL visible in the transaction feed on the home screen", and includes the implementation detail "add cREAL as a key to the CURRENCY object in `currencies.ts` in the wallet

PR #2787, opened 2022-08-16
feat(ihl): use siwe auth

satish-ravi:
### Description

Update IHL module to use SIWE auth
Bumped fiatconnect-sdk and fiatconnect-types to latest versions

### Other changes

N/A

### Tested

Unit tests, manually invoking login function from a component

### How others should test

N/A

### Related issues

Fixes ACT-392

### Backwards compatibility

Yes


test/RootStateSchema.json
```diff
@@ -3064,6 +3064,9 @@
                                             },
                                             "quoteId": {
                                                 "type": "string"
+                                            },
+                                            "transferType": {
```
satish-ravi, line 3068: Wasn't sure if this needed a migration. I don't think we need one because this is on a state field that is only relevant to the SelectProviderScreen and we refetch every time the screen is loaded with a new CICO flow. Also, even if we do a migration, don't think there's a good default we can set to since it can either be for cash in or cash out.

kathaypacific: that makes sense, i think it's fine to exclude the migration here :) 

i'm not close to the details but if there are flows where we always want to fetch new data and any persisted data from previous app sessions are not useful, we can consider excluding these from the redux persist? then we never need to worry about migrations. not sure if that's possible here though.

satish-ravi: cool, I ended up resetting the quotes field on the rehydrate action. I guess we can also stop persisting this field using something like https://github.com/rt2zz/redux-persist#nested-persists? I think this can be done in a follow up PR.

src/fiatconnect/slice.ts
```diff
@@ -210,6 +210,7 @@ export const slice = createSlice({
     builder.addCase(REHYDRATE, (state, action: RehydrateAction) => ({
       ...state,
       ...getRehydratePayload(action, 'fiatConnect'),
+      quotes: [], // reset quotes since we want to fresh a new set of quotes
```
jh2oman, line 213: did you mean?
```suggestion
      quotes: [], // reset quotes since we want to always re-fetch a new set of quotes
```

jh2oman (approved):

src/fiatconnect/ReviewScreen.test.tsx
```diff
@@ -43,6 +44,7 @@ function getProps(
     accountName: 'Chase (...2345)',
     institutionName: 'Chase',
     fiatAccountType: FiatAccountType.BankAccount,
+    fiatAccountSchema: FiatAccountSchema.AccountNumber,
```
jophish, line 47: Just to be clear, all the updates related to adding `fiatAccountSchema` are from the version bump on `fiatconnect-types`, correct? And aren't actually related to the SIWE auth work? (Same goes for the `TransferType` field on quotes.)

satish-ravi: yes, it was a side effect from bumping FC SDK, which uses FC types v7

ValoraQA:
Hey @MuckT Since NA is mentioned in how others should test section ,we are skipping this PR

Cc: @satish-ravi

PR #2645, opened 2022-06-27
fix: token balance load

MuckT:
### Description

A fix for when stale prices are present in the redux store causing user balances to display $0.00 when balances are stale.

### Other changes

Converts the token reducer to a slice.

### Tested

Tested locally on iOS and Android.

### How others should test

1. With an account with a balance Open the App the app and load the prices.
2. Prices should load.
3. Background the app and wait for the cache to expire, 24 hours by default, but can be edited in by adjusting `TIME_UNTIL_TOKEN_INFO_BECOMES_STALE` in the `config.ts` to emulate this behavior for reviews.
4. Foreground the App. 
5. Total Balance should display as '-' while loading or if the prices are stale. Accounts with balances should never display $0.00 balance.

### Related issues

- Fixes #2386

### Backwards compatibility

Yes

PR #2791, opened 2022-08-17
fix: account funded and liquidated events emitted incorrectly

kathaypacific:
### Description

When we implemented https://github.com/valora-inc/wallet/pull/2645 we broke the account funded and liquidated events through the `totalTokenBalanceSelector` returning null more often and causing these events to be emitted unexpectedly.

In this PR:
- ignore when `tokenBalance` is null (probably caused by loading) in the event watcher
- track stale prices so that they can be used for the event watcher when fetching new prices (during which, the token balance returns 0 and causes the account funded / liquidated events to trigger)

### Other changes

N/A

### Tested

Manually tested funding / liquidating accounts, and when prices are stale. Updated unit tests.

### How others should test

N/A

### Related issues

Fixes RET-383

### Backwards compatibility

Yes


src/tokens/selectors.ts
```diff
@@ -131,6 +138,26 @@ export const defaultTokenToSendSelector = createSelector(
   }
 )
 
+export const staleTokenBalanceSelector = createSelector(
+  [tokensListSelector, tokensWithStaleUsdValueSelector, localCurrencyExchangeRatesSelector],
+  (tokensList, tokensWithUsdValue, exchangeRate) => {
```
dievazqu, line 143: nit: rename `tokensWithUsdValue` to `tokensWithStaleUsdValue`

src/tokens/saga.ts
```diff
@@ -376,12 +381,23 @@ export function* watchFetchBalance() {
...
+    const areTokenPricesStale: ReturnType<typeof stalePriceSelector> = yield select(
+      stalePriceSelector
+    )
+    const staleTokenBalance: ReturnType<typeof staleTokenBalanceSelector> = yield select(
+      staleTokenBalanceSelector
+    )
+    // we reset the usd value of all token balances to 0 if the exchange rate is
+    // stale, so it is okay to use stale token prices to monitor the account
+    // funded / liquidated status in this case
+    const tokenBalance = areTokenPricesStale ? staleTokenBalance : currentTokenBalance
```
dievazqu, line 396: I think `staleTokenBalance` contains only the staled balance, instead of the total balance considering the staled price. So, it could happen that the staled balance is lower than `DOLLAR_MIN_AMOUNT_ACCOUNT_FUNDED`, triggering again the `account_liquidated` event, but the non-staled balance could be greater. Right?

src/tokens/saga.test.ts
```diff
@@ -165,10 +169,27 @@ describe('watchAccountFundedOrLiquidated', () => {
     }
   }
 
-  it('dispatches the account funded event', async () => {
+  const isStale = (firstValue: boolean, restValue: boolean) => {
+    let callCount = 0
+    return () => {
```
dievazqu, line 174: nit:
```
return () => ++callCount == 1 ? firstValue : restValue
```

dievazqu (commented):
I think there is a small bug on the saga... I think the `staleTokenBalanceSelector` should return the total balance (including the staled balance) instead of returning only the staled balance.
But the rest seems fine! :)

src/tokens/saga.test.ts
kathaypacific: !!!!!! clever!!! i updated to this

src/tokens/selectors.ts
kathaypacific: lol good one, done

src/tokens/saga.ts
kathaypacific: i'm not sure i follow - when the prices become stale, we override the existing usdPrice of the token to 0 so the token balance becomes 0. i added something to that selector, so that before we set the usdPrice to 0 i put the old usdPrice in the staleUsdPrice field.

so for sure the stale balance could be lower than the `DOLLAR_MIN_AMOUNT_ACCOUNT_FUNDED` but that means that the total balance before it became stale was also below this threshold so the event correctly fires? :/ am i missing something?

kathaypacific:
@dievazqu bump on this :)

src/tokens/saga.ts
kathaypacific: fixed according to our discussion :) changed the stale price to be the last known price so that we don't have mixtures of stale and non stale prices

M-shin (approved):

ValoraQA:

cc: @kathaypacific 
Thanks,.!

PR #2793, opened 2022-08-18
feat(OFAC compliance): Block Sanctioned Countries (RET-343)

Samuel3Shin:
### Description
Blocks 4 sanctioned countries

Cuba: CU
Iran: IR
North Korea: KP
Syria: SY

Screen Shot 2022-08-18 at 8 34 56 AM

### Other changes
N/A

### Tested
With iOS emulator

### How others should test
With iOS emulator

### Related issues
- https://linear.app/valora/issue/RET-343/block-sanctioned-countries-for-ofac-compliance-346

### Backwards compatibility
Yes


src/navigator/Navigator.tsx
```diff
@@ -696,6 +704,17 @@ function ModalStackScreen() {
 }
 
 function RootStackScreen() {
+  let isSanctionedCountry = false
+  const userCountry = useSelector(userLocationDataSelector)
+  if (
+    userCountry.countryCodeAlpha2 === 'CU' ||
+    userCountry.countryCodeAlpha2 === 'IR' ||
+    userCountry.countryCodeAlpha2 === 'KP' ||
+    userCountry.countryCodeAlpha2 === 'SY'
```
kathaypacific, line 713: to make this easier to edit and maintain, i'd suggest to to add this as a variable in config.ts, like `const SANCTIONED_COUNTRIES = ['CU',....]` and create a selector that uses this config so that you can use it directly like `const isUserInSanctionedCountry = useSelector(userInSanctionedCountrySelector)` - what do you think?

the benefit of having these countries in an array, is you can do a simple check like `countries.includes(userCountry.countryCodeAlpha2)` in the selector

src/app/SanctionedCountryErrorScreen.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react'
+import { StyleSheet, Text, View } from 'react-native'
+import fontStyles from 'src/styles/fonts'
+
+class SanctionedCountryErrorScreen extends React.Component {
+  render() {
+    return (
+      <View style={styles.container}>
+        <Text style={styles.errorMessage}>{'Sorry, Valora is not support in your location'}</Text>
```
kathaypacific, line 9: hum, should we localise this? the chances of the app supporting the user's device language is pretty low though. i'm not sure what the right answer is 😅 i guess we can leave it like this, there is a typo here though. it should be `Sorry, Valora is not supported in your location`

src/app/SanctionedCountryErrorScreen.tsx
```diff
@@ -0,0 +1,26 @@
...
+      </View>
+    )
+  }
+}
+
+const styles = StyleSheet.create({
+  container: {
+    justifyContent: 'center',
+    alignItems: 'center',
+    flex: 1,
```
kathaypacific, line 19: we should add some padding so that the text doesn't touch the screen sides for small screens

src/navigator/Navigator.tsx
```diff
@@ -711,7 +730,14 @@ function RootStackScreen() {
     <RootStack.Navigator
       screenOptions={{ snapPoints: ['100%'], backdropComponent: renderBackdrop }}
     >
-      <RootStack.Screen name={Screens.MainModal} component={ModalStackScreen} />
+      {isSanctionedCountry ? (
+        <RootStack.Screen
```
kathaypacific, line 734: this feels kinda hacky 😅 weird to have `nativeBottomSheets(RootStack)` inside the sanctioned screens root stack.

i'd suggest to move the implementation inside `NavigatorWrapper`, and handling it like `Screens.BackupForceScreen` in the useEffect in line 84 :)

kathaypacific (commented):
🎸

src/app/SanctionedCountryErrorScreen.tsx
Samuel3Shin: Oh... I just copied and pasted the sentence in the linear ticket lol. I changed it!
I think it would not be a good idea to localize this because we don't need to support localization for those who can't use our app.

src/app/SanctionedCountryErrorScreen.tsx
Samuel3Shin: I added padding!

src/navigator/Navigator.tsx
Samuel3Shin: Thanks for the suggestion! I did what you said :)

src/navigator/Navigator.tsx
Samuel3Shin: I also thought that it is hacky. Using useEffect is a lot better approach! Thanks a lot!
I changed it as you said

src/navigator/NavigatorWrapper.tsx
```diff
@@ -81,6 +83,12 @@ export const NavigatorWrapper = () => {
   const shouldForceBackup = useSelector(shouldForceBackupSelector)
   const doingBackupFlow = useSelector(doingBackupFlowSelector)
 
+  React.useEffect(() => {
+    if (isUserInSanctionedCountry) {
+      navigate(Screens.SanctionedCountryErrorScreen)
+    }
+  }, [])
```
kathaypacific, line 90: nit: in case this information is not available / incorrect at the time of this component mounting, we should add the `isUserInSanctionedCountry` in the dependency array here :)

src/networkInfo/selectors.ts
```diff
@@ -1,4 +1,7 @@
+import { SANCTIONED_COUNTRIES } from 'src/config'
 import { RootState } from 'src/redux/reducers'
 
 export const networkConnectedSelector = (state: RootState) => state.networkInfo.connected
 export const userLocationDataSelector = (state: RootState) => state.networkInfo.userLocationData
+export const userInSanctionedCountrySelector = (state: RootState) =>
```
kathaypacific, line 6: while this works, it's not optimal since this would be computed every time.

i would suggest to use `createSelector` because this handles memoisation for you. the selector can look something like https://github.com/valora-inc/wallet/blob/f5a649c0b6adc96a9f8895b5e9ccb748d555dcd7/src/localCurrency/selectors.ts#L72 and you can make use of the existing `userLocationDataSelector` :) hope this makes sense

kathaypacific (approved):
a couple small suggestions from me, but otherwise looks great!

src/navigator/Navigator.tsx
Samuel3Shin: I just found a problem with a `useEffect` implementation.
Users can bypass it by sliding back...

Please see this video.
https://user-images.githubusercontent.com/47931506/185673311-001c05e0-b07e-496c-b454-828b0ad3acc9.mp4

src/networkInfo/selectors.ts
Samuel3Shin: I just did what you said. Thanks for letting me know a more optimized implementation!

src/navigator/NavigatorWrapper.tsx
Samuel3Shin: I just removed `useEffect` implementation because users can bypass the error screen by sliding back...

src/config.ts
```diff
@@ -148,3 +148,5 @@ const configLoggerLevels: { [key: string]: LoggerLevel } = {
 }
 
 export const LOGGER_LEVEL = configLoggerLevels[Config.LOGGER_LEVEL] || LoggerLevel.Debug
+
+export const SANCTIONED_COUNTRIES = ['CU', 'IR', 'KP', 'SY']
```
jeanregisser, line 152: We already have this data in https://github.com/valora-inc/wallet/blob/d556789f78bf105ba08f73850f95bef971f4d8b9/src/flags.ts#L16-L22

See this example for how to use it: https://github.com/valora-inc/wallet/blob/524961b631974ff3ce6ea1f8ae4db4013f7be8d7/src/verify/VerificationEducationScreen.tsx#L226-L230

MuckT: Addressed in https://github.com/valora-inc/wallet/pull/2793/commits/593053347870c8b996b4ef07e9c007711c271eda! Have some testing to do with a VPN, but I think this is close.

MuckT: @kathaypacific @jeanregisser I am not sure what the best way to test this. I can mock the users location or test the content of the screen easily enough, but I am not sure how to test the screen rendering when a users location changes.

MuckT:
Tested using an Android 12 Device with a VPN.

kathaypacific (commented):
🎸

locales/base/translation.json
```diff
@@ -352,6 +352,7 @@
     "bodyWithoutProvider": "The service provider is processing your order. You should receive your funds within a couple of days."
   },
   "restrictedRegion": "Limited support in region",
+  "unsupportedLocation": "{{appName}} is not supported in your location",
```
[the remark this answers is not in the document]
MuckT: With it localized we can serve users in both NK and Cuba with translated text.

src/navigator/NavigatorWrapper.tsx
```diff
@@ -184,17 +188,23 @@ export const NavigatorWrapper = () => {
...
-      <View style={styles.container}>
-        <Navigator />
-        {(appLocked || updateRequired) && (
-          <View style={styles.locked}>{updateRequired ? <UpgradeScreen /> : <PincodeLock />}</View>
-        )}
-        <View style={styles.floating}>
-          <AlertBanner />
-          <InviteFriendModal isVisible={isInviteModalVisible} onInvite={onInvite} />
+      {SANCTIONED_COUNTRY ? (
+        <SanctionedCountryErrorScreen />
```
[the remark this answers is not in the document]
MuckT: Adjusted this in f21c6037b8c31e3e632142706455649a065f3167 using BackupForceScreen as an example.

MuckT:
@kathaypacific @jean - since there's been quite a few changes since the approval I thought another look would be good.

src/utils/countryFeatures.ts
```diff
@@ -17,7 +17,7 @@ export function getCountryFeatures(countryCodeAlpha2: string | null): SpecificCo
 }
 
 export const getCountryFeaturesSelector = createSelector(
-  userLocationDataSelector,
+  (state: RootState) => state.networkInfo.userLocationData,
```
jeanregisser, line 20: Weird, this shouldn't be needed 🤔 
I think the problem is we created an import cycle between `networkInfo/selectors.ts` and `utils/countryFeatures.ts`.
We should move `userInSanctionedCountrySelector` here to avoid it.

src/home/CashInBottomSheet.tsx
```diff
@@ -114,7 +118,7 @@ function CashInBottomSheet() {
     <Modal
       animationIn="slideInUp"
       animationInTiming={800}
-      isVisible={isModalVisible}
+      isVisible={isModalVisible && !inSanctionedCountry}
```
jeanregisser, line 121: Ah this feels a bit hacky because the only screen should be `SanctionedCountryErrorScreen`.
I guess the previous hacky was a bit nicer to avoid this (and future similar issues), since we were not rendering the navigator 😄

jeanregisser (commented):
Thanks!

src/home/CashInBottomSheet.tsx
MuckT: Moved this logic to `WalletHome.tsx` which has `shouldShowCashInBottomSheet` with some other conditional display logic.

src/utils/countryFeatures.ts
MuckT: Fixed in 100992a0441ccd9d827c9971a8703e114989b098!

kathaypacific (approved):
i think this looks really tidy, @jeanregisser what do you think?

kathaypacific:
although not sure why WalletHome would be mounted anyway :| if it's just racing the WalletHome rendering and determining the user country, having the hack in the navigator might also show the cash in bottom sheet? not sure what is best, i'm also okay with the hack in the navigator (but it does feel intrusive to have it there, it's so global...)

jeanregisser:
Yes in most cases `WalletHome` will not be mounted. 
It's only for users switching from a non-sanctionned to a sanctionned country that it can happen.

src/home/WalletHome.tsx
```diff
@@ -102,6 +104,10 @@ function WalletHome() {
   }
 
   const shouldShowCashInBottomSheet = () => {
+    // If user is in a sanctioned country do not show the cash in bottom sheet
+    if (userInSanctionedCountry) {
+      return false
+    }
```
jeanregisser, line 110: Doesn't the CashIn bottom sheet get unmounted when `navigateClearingStack(Screens.SanctionedCountryErrorScreen)` is called?

jeanregisser (approved):

MuckT:
@jeanregisser I had not pushed my change to `shouldShowCashInBottomSheet` prior to merging. I think it's ok to leave as is  since it's just an additional check, but it can be removed.
```
    // If user is in a sanctioned country do not show the cash in bottom sheet
    if (userInSanctionedCountry) {
      return false
    }
```

ValoraQA:
Hey @MuckT  we have verified the above task on latest Android Internal Release build V 1.40.1 and latest iOS Test Flight Release build V 1.40.1 and below is our observation

**Observation :** We are able to proceed to phone number page after applying neighboring countries VPN as instructed in the weekly meet

1] For Cuba - used Mexico VPN
2] For Iran - used Turkey VPN
3] For Syria - used Turkey VPN
4] For North Korea - Used Seoul VPN

Thanks.!!

PR #2802, opened 2022-08-18
build(deps): update detox to 19.10.0

MuckT:
### Description

Update to [Detox 19.10.0](https://github.com/wix/Detox/releases/tag/19.10.0)

### Other changes

Don't display out of Sync Banner during e2e tests.

### Tested

Tested in CI.

### How others should test

- Reviewers: check [E2E actions on this branch](https://github.com/valora-inc/wallet/actions/workflows/e2e-ci.yml?query=branch%3Atomm%2Fdetox-19.10.0) - First one was prior to removing out of sync banner.
- QA: N/A

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/package.json
+++ b/package.json
@@ -252,7 +252,7 @@
     "babel-core": "7.0.0-bridge.0",
     "babel-jest": "^26.6.3",
     "codecov": "^3.6.5",
-    "detox": "^19.9.0",
+    "detox": "^19.10.0",
     "escape-string-regexp": "^1.0.5",
     "eslint": "^7.27.0",
     "eslint-plugin-import": "^2.23.3",
--- a/src/components/TokenBalance.tsx
+++ b/src/components/TokenBalance.tsx
@@ -8,6 +8,7 @@ import { FiatExchangeEvents, HomeEvents } from 'src/analytics/Events'
 import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
 import Dialog from 'src/components/Dialog'
 import { formatValueToDisplay } from 'src/components/TokenDisplay'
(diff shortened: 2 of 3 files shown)


kathaypacific (commented):
ah Tom could we make PR's more targeted in the future and avoid unrelated changes? for this one, i'd like to approve the part about upgrading detox but would like to hold on this hack to hide the banner for e2e tests.

i don't think that we know the cause of why this banner is suddenly showing more frequently, it could totally be a genuine bug that we've created and should investigate. our e2e tests should test our production code as closely as possible to give us the most confidence...

EDIT: I see now that the spike in seeing this offline banner is related to blockscout returning 500 errors more often in alfajores (https://console.cloud.google.com/logs/query;query=resource.type%3D%22gae_app%22%0Aresource.labels.module_id%3D%22blockchain-api%22;cursorTimestamp=2022-08-19T12:28:05.999000072Z?project=celo-mobile-alfajores). it'd be nice to get the blockscout team to fix the issue - if blockscout is returning 500 errors, i think other parts of the app would / should not work correctly anyway and other tests that rely on balance would fail right? (or maybe we don't have any tests like this, and this stopgap is okay until we can sort out a better solution for the 500 errors?)

kathaypacific (approved):
approving as i don't want to block us, but let's not forget to revert this change once the blockscout team have addressed the issue. :) 

am thinking that we could probably look to make this offline banner less intrusive in the future - it's a pretty annoying user experience to have this banner that won't go away

ValoraQA:
Hey @MuckT  since NA is mentioned in how others should test ,we are skipping this PR

Thanks.!

PR #2815, opened 2022-08-24
test(vulnerabilities): run checks on schedule and with manual dispatch

MuckT:
### Description

Runs the checks on a daily schedule along with a manual dispatch which targets the head of main.

### Other changes

N/A

### Tested

N/A

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

Yes


.github/workflows/check.yml
```diff
@@ -7,6 +7,17 @@ on:
       - main
   # ..and any pull request.
   pull_request:
+  workflow_dispatch:
+    inputs:
+      logLevel:
+        description: 'Log level'
+        required: true
+        default: 'warning'
```
jeanregisser, line 15: Where is this being consumed? Or is it some GH actions magic?

jeanregisser (commented):
Thanks!

.github/workflows/check.yml
```diff
@@ -7,6 +7,17 @@ on:
...
+      tags:
+        description: 'Run Checks'
```
jeanregisser, line 17:
```suggestion
      ref:
        description: 'The branch, tag or SHA to checkout (uses the default branch if empty)
```

.github/workflows/check.yml
```diff
@@ -7,6 +7,17 @@ on:
...
+  # Cron job to run checks @ 8:30 pm daily on the latest commit on the default branch - main
```
silasbw, line 18: why 8:30pm?  is that significant?

MuckT: I had to pick a time and currently GitHub actions don't have [time zone support](https://github.com/orgs/community/discussions/13454). As this is in UTC This should be around 1:30 pm PST for most of the year.

.github/workflows/check.yml
MuckT: Added in b9d0655730838bd3ead19099d0e1f01751e91abf! Do you know of a way to test prior to merging?

.github/workflows/check.yml
```diff
@@ -7,11 +7,19 @@ on:
...
+  schedule:
+    - cron: '30 20 * * *'
 
 # Cancel any in progress run of the workflow for a given PR
 # This avoids building outdated code
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref }}
+  # Fallback used github.head_ref as it only defined on pull_request
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```
jeanregisser, line 22: I propose we do this:
```suggestion
  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
```

This way it only runs 1 instance for a given commit when triggered manually (without interfering with other branches).
We should do that for all workflows.

.github/workflows/check.yml
```diff
@@ -7,11 +7,19 @@ on:
       - main
   # ..and any pull request.
   pull_request:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: 'The branch, tag or SHA to checkout (uses the default branch if empty)'
```
jeanregisser, line 13: Sorry, I got confused, we can actually remove this input 🤦 
When we select the workflow branch when triggering manually, it checks out the full branch. No need to have the additional param.

Also to test manually before merging you can use GH cli:
```
gh workflow run "Check" --ref tomm/checks-schedule-manual
```

Screenshot 2022-08-26 at 16 27 37

jeanregisser (approved):

ValoraQA:
Hey @MuckT since NA is mentioned in how others should test section,we are skipping this PR

Thanks..!

PR #2817, opened 2022-08-24
feat: feature flag for guided onboarding copy

XrayBravoGolf:
### Description

An additional remote config feature flag, `showGuidedOnboardingCopy` , is added and recognized through changes made in this PR. This boolean defaults to `false`; application would appear identical to that before this PR when this boolean is `false`.
Different header titles and guide texts are shown if the new feature flag is set to `true` from Firebase. For the Welcome page, the welcome prompt is changed. 
The pages affected are the Welcome, Name and Picture, PIN Setup and Biometrics Setup.

### Other changes

None

### Tested

Tests for individual pages that were affected by this flag were written, with the selector mocked to return true on that flag, and affected copies verified to be the "guided onboarding" ones.
This feature flag currently has only been added to alfajores, but not mainnet yet.

### How others should test

Depending on how the feature flag is applied, test users should see different copy during onboarding.

### Related issues

- Fixes [ACT-417](https://linear.app/valora/issue/ACT-417/onboarding-experiment-trusted-guide)

### Backwards compatibility

Backwards compatible in that an older app that do not recognize the new feature flag would ignore it. Upgrade compatible because default value of false will be added during upgrading the old state.


XrayBravoGolf:
Attached are the looks of the affected pages when the flag is enabled (`true`):
Simulator Screen Shot - iPhone 13 - 2022-08-25 at 01 23 41
Simulator Screen Shot - iPhone 13 - 2022-08-25 at 01 23 33
Simulator Screen Shot - iPhone 13 - 2022-08-25 at 01 23 21
Simulator Screen Shot - iPhone 13 - 2022-08-25 at 01 22 55

lvsuttie:
Looks great, one q 4 u. Biometric type is dynamic (I believe) depending on if the user has FaceID on iOS or Fingerprint on iOS. Does the copy update depending on whether its android or iOS on the Biometrics step?

XrayBravoGolf:
> Looks great, one q 4 u. Biometric type is dynamic (I believe) depending on if the user has FaceID on iOS or Fingerprint on iOS. Does the copy update depending on whether its android or iOS on the Biometrics step?

Yes; I used the existing logic for the type, like this: `"guideDescription": "{{biometryType}} makes it easier, quicker, and more secure to connect your wallet to dapps & sign transactions."`, so it should update accordingly.

src/onboarding/welcome/Welcome.tsx
```diff
@@ -28,6 +28,8 @@ export default function Welcome() {
 
   const createAccountCopyTestType = useSelector(createAccountCopyTestTypeSelector)
 
+  const guidedOnboardingEnabled = useSelector(showGuidedOnboardingSelector)
```
jh2oman, line 31: nit: I'd appreciate if these were consistent wrt naming. I.e. use the variable `showGuidedOnboarding` for the `showGuidedOnboardingSelector`

src/redux/migrations.ts
```diff
@@ -793,15 +793,28 @@ export const migrations = {
     },
   }),
   72: (state: any) => ({
+    ...state,
+    app: {
+      ...state.app,
+      showGuidedOnboardingCopy: REMOTE_CONFIG_VALUES_DEFAULTS.showGuidedOnboardingCopy,
```
jh2oman, line 799: This is incorrect. I suspect while you were working other people made changes to redux state. You need to make a new version and not modify the old ones like this.

src/redux/migrations.ts
```diff
@@ -793,15 +793,28 @@ export const migrations = {
...
       shouldShowRecoveryPhraseInSettings:
         REMOTE_CONFIG_VALUES_DEFAULTS.shouldShowRecoveryPhraseInSettings,
     },
   }),
-  73: (state: any) => state,
   74: (state: any) => ({
+    ...state,
+    app: {
+      ...state.app,
+      inviteMethod: REMOTE_CONFIG_VALUES_DEFAULTS.inviteMethod,
```
jh2oman, line 814: What does this inviteMethod have to do with your PR?

jh2oman (changes requested):
You need to fix your handling of the redux migration. 

Also please update your PR title to match https://www.conventionalcommits.org/en/v1.0.0/

src/redux/migrations.ts
cajubelt, line 799: you should add a new migration, not modify an old one. You can technically get away with modifying an existing migration-- safest one is probably the last one-- IF your modification goes out in the *same release* as the code that added that migration to begin with. But it's always safer to just add a new one.

cajubelt: please undo all your changes to existing migrations in this file and add a new one to include your addition of this feature flag.

test/schemas.ts
```diff
@@ -1568,7 +1567,7 @@ export const v72Schema = {
   },
   app: {
     ...v71Schema.app,
-    shouldShowRecoveryPhraseInSettings: false,
+    showGuidedOnboardingCopy: false,
```
cajubelt, line 1570: you'll need to change this once you've fixed the migrations file

cajubelt (changes requested):
looks close! mainly needs fixes in the redux migration

src/redux/migrations.ts
XrayBravoGolf: > What does this inviteMethod have to do with your PR?

It does not; I must have gotten confused while merging changes from main.

jh2oman (changes requested):
I see two discrepancies between the [Figma Designs](https://www.figma.com/file/j84MQn5xaOJ9RBjuIZgxPN/Onboarding-Experiments?node-id=1655%3A14575) and your changes
1. The placeholder text of the name form should also change with the experiment
2. The "Turn on Face ID" button should be left aligned not center aligned.

cajubelt (approved):
looks like a test needs updating, and Jacob noted diffs from designs, but otherwise looking good

src/account/Settings.test.tsx
```diff
@@ -108,13 +108,8 @@ describe('Account', () => {
   })
 
   it('navigates to PincodeSet screen if entered PIN is correct', async () => {
-    const store = createMockStore({
-      app: {
-        showGuidedOnboardingCopy: true,
-      },
-    })
```
cajubelt: I actually think this test would be a bit better if this were still here. since it would make sure that even if the feature flag is set to true, you don't get the onboarding copy if you're coming from settings.

without the feature flag set to true though, the test doesn't seem to help verify that behavior

cajubelt: not important though

src/onboarding/registration/EnableBiometry.tsx
```diff
@@ -122,9 +122,9 @@ export default function EnableBiometry({ navigation }: Props) {
 
   return (
     <ScrollView style={styles.contentContainer}>
-      <SafeAreaView style={styles.container}>
+      <SafeAreaView style={showGuidedOnboarding ? styles.containerLeftAligned : styles.container}>
```
jh2oman, line 125: nit: I don't think this style change needs to be part of the experiment. It can just be universal. check with @nityarao though to be sure!

jh2oman (approved):
Approved with the nit / question to Nitya

ValoraQA:
Hey @MuckT, We have verified the above issue on Latest Test Flight Release Build V1.40.0 and updated the status in test rails.

**Observation**:
- User is able to see different carousels after uninstalling and reinstalling the application and also able to scroll the carousel. 

**Test Rail link:**
**iOS 15:** https://valoraapp.testrail.io/index.php?/tests/view/69490

**CC**: @XrayBravoGolf 

Thanks.!!

MuckT:
@ValoraQA the experience you're describing was implemented in https://github.com/valora-inc/wallet/pull/2827, this PR was reverted in https://github.com/valora-inc/wallet/pull/2840 and the behavior shouldn't be seen in 1.40.0.

ValoraQA:
@MuckT there was some confusion as there are two PR's related to carousel and onboarding flow hence we have provided observation wrong in the above comment ,
We did check the above PR in alfajores build and observed that feature flag is not enabled i.e the textual changes in screenshot provided are not seen during onboarding flow

Thanks.!

ValoraQA:
Hey @MuckT we are not able to see the behavior on the Alfajores build V 1.40.0 and we did check the PIN functionality as mentioned in #2840 
Thanks..!

PR #2819, opened 2022-08-25
feat: invite flow without escrow payments

kathaypacific:
### Description

This PR:
- add inviteMethod remote config (enum)
- the send amount button now displays loading state when verifying phone number OR reviewing amount
- if the phone number is unverified, display a full screen modal with share option

https://user-images.githubusercontent.com/20150449/186867113-26b9ddf9-c1e7-431d-8fa9-a2be57f5cc3b.mp4

### Other changes

N/A

### Tested

Manually

### How others should test

In alfajores, when sending to a contact that is not verified on Valora, the flow https://www.figma.com/file/8ig95B4sfsYIDIpg3JGP03/Phone-Verification?node-id=1336%3A19069 should be triggered (instead of continuing to send the payment to escrow)

### Related issues

- Fixes RET-379

### Backwards compatibility

Yes


src/components/InviteOptionsModal.tsx
```diff
@@ -1,56 +1,94 @@
+import { getPhoneHash } from '@celo/utils/lib/phoneNumbers'
```
kathaypacific, line 1: @jeanregisser i think this is safe to use? we need a way to track the success rate of invitations, and this was the only way i could think of without capturing any PII

src/components/InviteOptionsModal.tsx
```diff
@@ -1,56 +1,94 @@
...
-  isVisible: boolean
-  onCancel: () => void
-  onWhatsApp: () => void
-  onSMS: () => void
+  recipient: Recipient
+  onClose(): void
 }
 
-const InviteOptionsModal = ({ onCancel, onWhatsApp, onSMS, isVisible }: Props) => {
+const InviteOptionsModal = ({ recipient, onClose }: Props) => {
```
kathaypacific, line 25: this component wasn't being used so i repurposed it

src/app/types.ts
```diff
@@ -10,3 +10,9 @@ export enum CreateAccountCopyTestType {
   Wallet = 'WALLET',
   AlreadyHaveWallet = 'ALREADY_HAVE_WALLET',
 }
+
+export enum InviteMethodType {
+  Escrow = 'ESCROW',
+  ManualInvite = 'MANUAL_SHARE',
```
jeanregisser, line 16: nit: Maybe?
```suggestion
  ManualShare = 'MANUAL_SHARE',
```

src/components/Button.tsx
```diff
@@ -28,6 +28,7 @@ export enum BtnSizes {
 export interface ButtonProps {
   onPress: () => void
   style?: StyleProp<ViewStyle>
+  buttonStyle?: StyleProp<ViewStyle>
```
jeanregisser, line 31: I wonder if we should introduce another prop for the icon placement to avoid exposing too much of the inner layout, and avoiding incorrect usages.

jeanregisser (approved):
Great work! 🚀 

I think the UX would be better by directly showing the invite modal when selecting a user not yet on Valora. But I know it would require to know the verification status of all contacts to avoid a spinner/waiting state after tapping.
Also I guess this invite without escrow is also temporary (famous last words 😅).
But something to think in the future.

src/components/InviteOptionsModal.tsx
jeanregisser: This look good, can't think of a better way 👍

src/app/types.ts
kathaypacific: ah yes, thank you! updated

src/components/Button.tsx
kathaypacific: yeah i also felt this was hacky. i updated to use a boolean prop, maybe you want to take another look at the changes to the Button?

jeanregisser: Yes looks good, we can remove `buttonStyle` then 👍

src/components/Button.tsx
```diff
@@ -71,18 +74,27 @@ export default React.memo(function Button(props: ButtonProps) {
         <Touchable
           onPress={debouncedOnPress}
           disabled={disabled}
-          style={getStyle(size, backgroundColor, opacity)}
+          style={[getStyle(size, backgroundColor, opacity, iconPositionLeft)]}
```
jeanregisser, line 77: nit: array is not necessary
```suggestion
          style={getStyle(size, backgroundColor, opacity, iconPositionLeft)}
```

kathaypacific: ah yes, thank you! forgot this was there

src/components/Button.tsx
kathaypacific: removed!

ValoraQA:
Hey @MuckT we have verified the above task on Latest iOS Alfajores build V 1.40.0 and updated status in Test Rails.

**Observation :** 
- URL seen in prefilled message while sending invite is seen different as compared with Figma screen shared in the PR and we have raised query for it

**New Query :** [link](https://github.com/valora-inc/wallet/issues/2843)

**Test Rail link:**
**iOS 14:** https://valoraapp.testrail.io/index.php?/tests/view/69516

**Device :** iPhone 12(14.7.1)

**CC :** @kathaypacific 

Thanks.!

MuckT:
@ValoraQA for this pull request the invite link is expected to be generic as mentioned in the AC for RET-379.

ValoraQA:
Hey @MuckT we have verified the above task on Latest Android Alfajores build V 1.40.0 and updated status in Test Rails.

**Observation :**

- Green colored line is seen on send invite link page for Android platform, which is not seen on iOS platform and we have raised bug for it 
New Bug : https://github.com/valora-inc/wallet/issues/2852

**Test Rail link:**
**Android 11 :** https://valoraapp.testrail.io/index.php?/tests/view/69547

**Device :** Google Pixel 2XL(11.0)

CC : @kathaypacific

Thanks.!

PR #2832, opened 2022-08-30
chore(deps): make RN packages gradle 7 compatible

satish-ravi:
### Description

Make the following packages gradle 7 compatible
- react-native-blind-threshold-bls - used latest fork, which is gradle 7 compatible
- react-native-securerandom - used latest main, which has the gradle 7 fix, but not published to npm yet
- react-native-flag-secure-android - used patch-package instead of fork, main already has an unmerged open PR
- react-native-restart-android - used patch-package instead of fork since only 1 line needs fixing and the package has been inactive for 4 yrs
- @segment/analytics-react-native-(adjust/clevertap/firebase) - used patch-package to upgrade kotlin version, since v2 upgrade is not straight forward

### Other changes

N/A

### Tested

Build iOS and Android apps locally

### How others should test

N/A

### Related issues

- Fixes ACT-440

### Backwards compatibility

Yes


yarn.lock
```diff
@@ -16388,6 +16388,12 @@ react-native-securerandom@1.0.0, react-native-securerandom@^1.0.0:
   dependencies:
     base64-js "*"
 
+"react-native-securerandom@git+https://github.com/robhogan/react-native-securerandom#7c01efd":
```
jeanregisser, line 16391: We should ensure we only use one version of this package.
We can help yarn deduplicate it by manually removing the 2 references and running `yarn install` again.

jeanregisser (approved):
Looking great 👍 
Already approved with a small comment.

ValoraQA:

Cc: @satish-ravi 
Thanks.!

PR #2846, opened 2022-09-06
chore(deps): upgrade fiatconnect deps

cajubelt:
### Description

upgraded fiatconnect types and fiatconnect sdk versions

NOTE: there is an update to RootStateSchema, but not a migration, because we are not turning on any FC provider which gives quotes in the old format. Therefore, no quotes adhering to the old format will make it into redux state, so a migration to the new format will not be necessary.

### Other changes

na

### Tested

existing tests and compile time checks should be enough.

### How others should test

na

### Related issues

na

### Backwards compatibility

works with the latest FC spec


satish-ravi (approved):

ValoraQA:
Hey @MuckT since NA is mentioned in how other should test section, we are skipping this PR
CC: @cajubelt 
Thanks!

PR #2840, opened 2022-09-01
revert: feature flag for guided onboarding copy

MuckT:
### Description

Reverts #2817 due to:
- The onboarding text was displayed in non onboarding pin displays.
- Empty Scroll view was causing unexpected shifts to pincode entry layout.

#### Screenshots Before Revert / After Revert
<p align="center">

</p>

### Other changes

N/A

### Tested

Tested locally on iOS

### How others should test

Devs: Please double check my changes to the migrations.

QA: Check that Pin Entry Functions as expected and displays correctly during onboarding, viewing recovery phrase, and on Pin Change.

### Related issues

N/A

### Backwards compatibility

Yes - Devs may need a fresh install due to redux changes.

PR #2850, opened 2022-09-07
feat: feature flag for guided onboarding copy

XrayBravoGolf:
### Description

This PR is a re-visit to #2817 which was reverted in #2840 due to a bug in the Pincode behavior, causing it to change even when the feature flag was set to false; this PR includes the same features with this bug addressed. 

In addition, during onboarding, copy and design had  an update for the "re-enter pin" page, as shown in the screenshot and in the [design file](https://www.figma.com/file/j84MQn5xaOJ9RBjuIZgxPN/Onboarding-Experiments?node-id=1852%3A16671).

Original description: an additional remote config feature flag, `showGuidedOnboardingCopy` , is added and recognized through changes made in this PR. This boolean defaults to `false`; application would appear identical to that before this PR when this boolean is `false`.
Different header titles and guide texts are shown if the new feature flag is set to `true` from Firebase. For the Welcome page, the welcome prompt is changed. 
The pages affected are the Welcome, Name and Picture, PIN Setup and Biometrics Setup.

### Other changes

None

### Tested

Tests for individual pages that were affected by this flag were written, with the selector mocked to return true on that flag, and affected copies verified to be the "guided onboarding" ones.
This feature flag currently has only been added to alfajores, but not mainnet yet.

### How others should test

For engineers trying it out on local build, you can edit the `src/app/selectors.ts:132` to return `true` without setting feature flag in Firebase.
Depending on how the feature flag is applied, test users should see different copy during onboarding.

### Related issues

- Fixes [ACT-417](https://linear.app/valora/issue/ACT-417/onboarding-experiment-trusted-guide)

### Backwards compatibility

Backwards compatible in that an older app that do not recognize the new feature flag would ignore it. Upgrade compatible because default value of false will be added during upgrading the old state.


XrayBravoGolf:
These are what the pin-entry looks like during onboarding now. @nityarao 
Simulator Screen Shot - iPhone 13 - 2022-09-07 at 16 05 38
Simulator Screen Shot - iPhone 13 - 2022-09-07 at 16 05 30

src/onboarding/registration/EnableBiometry.tsx
```diff
@@ -145,6 +166,10 @@ const styles = StyleSheet.create({
     paddingHorizontal: 40,
     alignItems: 'center',
   },
+  containerLeftAligned: {
+    paddingTop: 186,
+    paddingHorizontal: 40,
```
MuckT, line 171: These paddings seem a bit suspect. Is there a way to get the alignment you want without these static values? Maybe a flex with align / justify content?

XrayBravoGolf: > These paddings seem a bit suspect. Is there a way to get the alignment you want without these static values? Maybe a flex with align / justify content?

Definitely there is; these numbers were not my original so I hesitate touching them, however let me get a quick patch in.

src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -42,24 +47,30 @@ function NameAndPicture({ navigation }: Props) {
...
+        let pageTitleTranslationKey
+        if (showGuidedOnboarding) {
+          pageTitleTranslationKey = 'name'
+        } else {
+          pageTitleTranslationKey = choseToRestoreAccount
+            ? 'restoreAccount'
+            : createAccountCopyTestType === CreateAccountCopyTestType.Wallet ||
+              createAccountCopyTestType === CreateAccountCopyTestType.AlreadyHaveWallet
+            ? 'createProfile'
+            : 'createAccount'
```
cajubelt, line 65: why would all of this logic not apply if someone was in the guided onboarding "on" group? Seems like we'd still want them to see these header titles if they were restoring their account, already had a wallet etc. Maybe the only difference is the case where `choseToRestoreAccount`, `createAccountCopyTestType === CreateAccountCopyTestType.Wallet`, and `createAccountCopyTestType === CreateAccountCopyTestType.AlreadyHaveWallet` all evaluate to false but `showGuidedOnboarding` is true?

src/onboarding/welcome/Welcome.tsx
```diff
@@ -53,7 +54,15 @@ export default function Welcome() {
...
-        <Text style={styles.title}>{t('welcome.title')}</Text>
+        {showGuidedOnboarding ? (
+          <Text style={styles.title} testID={'WelcomeText'}>
+            {t('welcome.titleGuided')}
+          </Text>
+        ) : (
+          <Text style={styles.title} testID={'WelcomeText'}>
+            {t('welcome.title')}
+          </Text>
+        )}
```
cajubelt, line 65:
```suggestion
          <Text style={styles.title} testID={'WelcomeText'}>
            {t(showGuidedOnboarding ? 'welcome.titleGuided' : 'welcome.title')}
          </Text>
```
code golf

cajubelt: code formatting's probly wrong but you get the idea

src/pincode/PincodeSet.tsx
```diff
@@ -86,8 +89,12 @@ const mapDispatchToProps = {
...
   static navigationOptions = ({ route }: ScreenProps) => {
     const changePin = route.params?.changePin
-    const title = changePin ? i18n.t('pincodeSet.changePIN') : i18n.t('pincodeSet.create')
-
+    const showGuidedOnboarding = route.params?.showGuidedOnboarding
+    const title = changePin // changePin has priority
+      ? i18n.t('pincodeSet.changePIN')
+      : showGuidedOnboarding // this should be during onboarding
+      ? i18n.t('pincodeSet.selectPIN')
+      : i18n.t('pincodeSet.create')
```
cajubelt, line 97: please don't add double ternaries to the codebase. they're hard to read and reason about.

src/onboarding/registration/NameAndPicture.tsx
XrayBravoGolf: I believe because either restoring or new wallet would share the "name" title because both were considered onboarding and design wanted them to have this title when experiment is on, since the umbrella condition.

src/onboarding/welcome/Welcome.tsx
XrayBravoGolf: Actually surprised that I wrote it the way I did; will send a commit and fix it.

cajubelt (approved):
🥳

MuckT:
@XrayBravoGolf can we enable this in Alfajores for QA testing?

cajubelt:
yeah should be fine @MuckT

PR #2861, opened 2022-09-12
feat: onboarding carousel order and copy experiment

MuckT:
### Description

Implements the experiment for onboarding carousel variants as described in [notion](https://www.notion.so/valora-inc/651e35300967405a969d31808247ad07?v=dd3840385dc04284a6e1878b7ae66ccb&p=46ff60a5ad3a437cab5a168ae30a03f6&pm=s). Since at the time of initial app start Firebase may not be available random orders of three cards were implemented using the [Fisher–Yates / Durstenfeld's Shuffle](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle); the 'old' vs 'new' variant were done using Math.random resulting in a total of 12 different variants: 6 old - 6 new.

#### Analytics Implemented - Whole Flow
- onboarding_education_start: used as a sanity check that all variants are displayed equally.
- onboarding_education_complete: used to determine which ordering and variant is the most successful.

#### Analytics Implemented - Per Screen
- onboarding_education_step_impression: identify what value prop was shown on each screen individually.

### Other changes

Added the [fast-shuffle](https://www.npmjs.com/package/fast-shuffle) dependency.

### Tested

Tested Locally on iOS

### How others should test

1. With a fresh install - view the carousel items
2. Close the app and reload.
3. Observe carousel items and ensure that order changes (It's random so it's possible items might not change every time).

### Related issues

- RET-382

### Backwards compatibility

Yes


kathaypacific (commented):
hey looking good, i had a few questions - 
1. should we use remote config somehow here? even if it is to have control over when we start / finish the experiment? we could have the remote config value to tell us whether the user is in the control or variant group for example. one advantage for this is that firebase will deliver the same value for the user - if we hard code to use Math.random, a user would see a different order every time (e.g. the first time, they don't create an account. when they come back to the app they will see a different order)
2. reading the experiment reqs, i think that we don't want to change the control? if i am reading the code correctly, now we would be shuffling the control as well?
3. i think in the previous version of this PR, you had a function for the shuffle instead of a new dependency - any reason why we want to go with a new dependency instead? i thought what we had before was good?

MuckT:
@kathaypacific
1. As these screens are the first screens seen in the app we can't expect that the remote config has been loaded yet.
2. I can take a look at keeping the control constant, but the current shuffling gives us 12 variants, equally likely to be seen, to see which performs the best out of all of them. I am not sure it's worth it to have the control taking 50% of the experiment and then the remaining 6 variants completing against it.
3. It's a small dependency also used in the nft-indexer. It appears to work well and using it we can avoid testing our custom shuffle implementation.

kathaypacific:
> @kathaypacific
> 
> 1. As these screens are the first screens seen in the app we can't expect that the remote config has been loaded yet.
> 2. I can take a look at keeping the control constant, but the current shuffling gives us 12 variants, equally likely to be seen, to see which performs the best out of all of them. I am not sure it's worth it to have the control taking 50% of the experiment and then the remaining 6 variants completing against it.
> 3. It's a small dependency also used in the nft-indexer. It appears to work well and using it we can avoid testing our custom shuffle implementation.

Ah okay, that makes sense! I think it's best to check with Ellen about the 6 vs 12 variants, as there might be some reasons for why the experiment was originally designed to have 6 variants. It's good for her to be clear on what is going out anyway. In case you haven't already, it's also worth checking in with Ellen about this experiment not using remote config / checking if all the analytics in place allow her to analyse the results :) technically the implementation you have here looks good to me

src/onboarding/education/OnboardingEducationScreen.tsx
```diff
@@ -20,26 +21,36 @@ function useStep() {
   const { t } = useTranslation()
 
   return React.useMemo(() => {
-    return [
+    const screenSet = Math.random() < 0.5 ? 0 : 1
```
silasbw, line 24: probably a dumb question:  but why aren't we using the AB testing framework for this?

ellenwoconnor:
@kathaypacific @MuckT I don't really have a strong preference on the shuffling of the control, there's pros and cons of doing that from an experiment design perspective. 

Pros: it control for the shuffling element of the experiment, which itself can sometimes introduce some artifacts. 
Cons: when the experiment concludes and if we decide to go with the control, I don't think we want to permanently end up shuffling the control screens for users. The shuffling is just a data gathering requirement, but kind of a weird user experience in terms of making it permanent. 

Arvin and I took a look at some of the test data in bigquery and it looked okay to us in terms of analysis. (We need impressions and scroll/click events, and we need to infer what was shown in what order, and I think we can do that). 

Possibly related to @silasbw  point (?) i noticed that we're logging the experiment variant in the carousel events, which I expected we'd get from firebase. Not sure if that means we aren't using firebase for the split, or if we're double logging, or what?

src/onboarding/education/OnboardingEducationScreen.tsx
MuckT: Since it's on the initial app screens we're not guaranteed to have loaded firebase / don't want rerender the screens when firebase loads changing the order.

kathaypacific:
yeah i guess this is your call @ellenwoconnor - we fetch remote config asynchronously, so there is a chance that the value loads after the onboarding screen has already loaded. it depends which route we want to go down: 
1) use remote config, introduce a delay in the onboarding screen to wait for it to load. the downside is obviously that a slow internet connection = the user sees some loading state for a while the first time they open the app. the upside is that we have better control to turn it on and off.
2) go with the implementation here, the upside is no delay for the user. the downsides are that there's no way to turn it off if a user refuses to update their app to a new version that removes the shuffle, and if a user doesn't complete onboarding and reopens their app, the shuffling + control/variant group assignment is random. idk if either of these downsides are impactful though, since users should really only experience this flow once

ellenwoconnor:
got it, i'm fine with bypassing firebase then to avoid adding delay to their first experience. 

FWIW, to avoid rerandomizing on each app open you could just directly do the simple heuristic that most experimentation platforms use under the hood -- IIRC this is something like: hash the device id + experiment name, convert to integer, mod 100 and then say control is 0 < 50 and treatment is 50 < 100.  

i don't know how often users close the app and reopen it. if it's not a lot, we can try just excluding these people from the results. if it's a lot then in future experiments on the carousel maybe we'd want to do something like the above

kathaypacific (approved):
sorry this slipped my radar for a while, looks great! 🚀

ValoraQA:
Hey @MuckT we have verified the above task on Latest iOS Mainnet Nightly build V 1.43.0 and updated our observation in sheet

PR #2865, opened 2022-09-13
chore: remove Mergify

kathaypacific:
### Description

We're ending support for Mergify since it doesn't add value on top of GitHub's auto-merge feature.

Currently we rely on Mergify to merge PR's for:
- updating licenses and disclaimers
- bumping the app version
- merging Crowdin PRs
- merging translation updates from Paula

This PR removes a mergify config and updates the automerge scripts to avoid relying on Mergify for the above-listed use cases.

### Other changes

removed Lukas from translation PR author whitelist

### Tested

Manually on https://github.com/valora-inc/wallet/pull/2859

### How others should test

N/A

### Related issues

- Fixes RET-394

### Backwards compatibility

Yes


.github/scripts/automergeCrowdinPr.js
```diff
@@ -38,41 +49,46 @@ module.exports = async ({ github, context }) => {
     return
   }
 
-  console.log(`Verifying that only expected files are modified for PR #${pr.number}`)
-  const listFiles = await github.rest.pulls.listFiles({
+  console.log(`Fetching reviews for ${pr.number}`)
```
kathaypacific, line 52: i couldn't resist refactoring this part (there aren't many changes if viewing this PR with whitespaces ignored)

kathaypacific:
there's a bit of repeated code between the scripts, but importing and exporting functions / variables for reuse is kind of fiddly with the GH scripts :/ the checks performed by the 3 scripts are different in the middle of each script so if we were to combine the scripts, it will be more difficult to read. i can sink more time into this if there are strong feelings from someone, i just couldn't convince myself that it was worth the trouble 🙈

.github/scripts/automergeCrowdinPr.js
```diff
@@ -81,21 +97,20 @@ module.exports = async ({ github, context }) => {
...
-  }
+  console.log(`Bringing PR #${pr.number} up to date with main branch`)
+  await github.rest.pulls.updateBranch({
+    owner,
+    repo,
+    pull_number: pr.number,
+  })
+
+  console.log(`Enabling automerge on PR #${pr.number}`)
+  await github.graphql(enableAutomergeQuery, {
```
kathaypacific, line 110: interestingly / annoyingly, enabling automerge is not supported via the rest api's and only supported with this graphql endpoint. this part is important since we have jobs that are triggered by the creation of branches and we won't be able to successfully call the `merge` rest endpoint since the CI won't have finished when these jobs are run.

.github/scripts/automergeCrowdinPr.js
```diff
@@ -81,21 +97,20 @@ module.exports = async ({ github, context }) => {
...
-      owner,
-      repo,
-      issue_number: pr.number,
-      labels: [AUTOMERGE_LABEL],
-    })
-  } else {
-    console.log(`Already labelled with ${AUTOMERGE_LABEL}`)
-  }
+  console.log(`Bringing PR #${pr.number} up to date with main branch`)
+  await github.rest.pulls.updateBranch({
```
kathaypacific, line 103: i tried to achieve feature parity with mergify here but i think there's still a reasonable chance for the PR's to become out of date by the time the CI finishes and therefore not merged (mergify would merge main into the branches again automatically and now we'd be doing it manually, but let's see how it goes)

jeanregisser: Great 👍

jeanregisser (approved):
Awesome!! 🚀 

Looks like we may be near the point of extracting and publishing this automerge feature into it's own GitHub action we could reuse in other repos 😄

.github/scripts/automergeCrowdinPr.js
```diff
@@ -38,41 +49,46 @@ module.exports = async ({ github, context }) => {
...
       pull_number: pr.number,
-      event: 'REQUEST_CHANGES',
-      body: `Changes requested from [${context.workflow} #${context.runNumber}](${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}), as the updated files in this PR did not match the expectation. Please check.`,
     })
-    return
-  }
+    const unexpectedFiles = listFiles.data.filter(
+      ({ filename }) =>
+        filename === DISALLOWED_UPDATED_FILE ||
+        !filename.match(new RegExp(ALLOWED_UPDATED_FILE_MATCHER))
```
cajubelt, line 76: super nit: this regexp could be reused instead of instantiating it for every file in the list (which is what's currently happening)

cajubelt: (to reuse, pull it out of the filter callback)

.github/scripts/automergeCrowdinPr.js
cajubelt: oh this is great

cajubelt (approved):
🥳 🥳 🥳

.github/scripts/automergeCrowdinPr.js
kathaypacific: yeah good one, thanks! i've updated

ValoraQA:
Hey @MuckT since NA is mentioned in how other should test section, we are skipping this PR
Cc: @kathaypacific 
Thanks.!

PR #2855, opened 2022-09-09
chore(persona): upgrade sdk to v2

cajubelt:
### Description

Upgrades persona SDK to v2 and updates styling. Styling still to be finalized and some styles on Android needs fixing, which can be fixed in a follow up.

Note: for testing purposes, the persona button component is included in the home screen. This isn't included in this PR

iOS:

https://user-images.githubusercontent.com/5062591/190319413-6159e8e1-5b28-4315-b1b7-b9f8258b1735.mp4

Android:

https://user-images.githubusercontent.com/5062591/190319673-a1b7eaff-a241-4953-9a2b-3e1c9a481def.mp4

### Other changes

N/A

### Tested

Manually by including the component in the home screen, unit tests

### How others should test

N/A

### Related issues

- Fixes ACT-400

### Backwards compatibility

Yes

PR #2866, opened 2022-09-13
feat(fiatconnect): Add KYC checking when selecting quotes

jophish:
### Description

For [ACT-381](https://linear.app/valora/issue/ACT-381/wallet-handle-kyc-in-quoteonpress). Implements the logic coming off of the "Call GET /KYC on IHL" node in [this flowchart](https://www.figma.com/file/fqLxcOIvDxQWAlDY60rCD6/FiatConnect-Persona-Diagram?node-id=12%3A397). In simpler terms, this PR handles all the KYC logic required when selecting a quote on the Select Provider screen.

New `KycStatus`/`KycLanding` screens (they're the same):

2022-09-14-120453_673x1418_scrot

### Other changes

Adds some functionality to make FC-authenticated calls to IHL, turns on support for KYC in quotes.

### Tested

These changes are unit tested and were tested out in the app by mocking out calls to IHL. A full e2e test, actually calling IHL is not currently possible since IHL does not have all the necessary endpoints yet.

### How others should test

### Related issues

- Fixes ACT-381
### Backwards compatibility


src/fiatconnect/KycStatusScreen.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react'
```
jophish, line 1: Added this in as a placeholder to have a unique place to navigate to from the quote handler saga.

src/fiatconnect/saga.ts
```diff
@@ -277,14 +280,57 @@ export function* handleSelectFiatConnectQuote({
   payload: params,
 }: ReturnType<typeof selectFiatConnectQuote>) {
   const { quote } = params
+
   try {
+    // If KYC is required for the quote, check that the user has correct KYC on file
+    // with the quote's provider
+    let getKycStatusResponse: GetKycStatusResponse
+    if (quote.requiresKyc()) {
+      const kycSchema = quote.getKycSchema()!
```
cajubelt, line 289: nit: by using this assertion operator, we are baking in the assumption that if `quote.requiresKyc()` evaluates to true, this will be nonnull. Equivalently, we could make use of the logical contrapositive, e.g. if `quote.getKycSchema()` is falsy, kyc is not required. (And ditch the usage of `quote.requiresKyc()` plus the assertion operator type hack.) This seems preferable to me.

cajubelt: maybe a rename like `getRequiredKycSchema` could make this suggestion more palatable (since it's fairly clear to me that if there isn't a required kyc schema, I don't have to do kyc)

src/fiatconnect/saga.ts
```diff
@@ -277,14 +280,57 @@ export function* handleSelectFiatConnectQuote({
...
+        case FiatConnectKycStatus.KycPending:
+          break
+        // If denied or expired, skip account management and
+        // navigate to the KYC status screen
+        case FiatConnectKycStatus.KycDenied:
+        case FiatConnectKycStatus.KycExpired:
+          navigate(Screens.KycStatusScreen)
+          yield put(selectFiatConnectQuoteCompleted())
+          return
+      }
```
cajubelt, line 323: might be helpful to be defensive and handle the case where `fiatConnectKycStatus` is undefined or unrecognized

cajubelt: and log an error or something

src/fiatconnect/saga.ts
```diff
@@ -306,18 +352,28 @@ export function* handleSelectFiatConnectQuote({
...
-    navigate(Screens.FiatConnectReview, {
-      flow: quote.flow,
-      normalizedQuote: quote,
-      fiatAccount,
-    })
+
+    // If the quote required KYC, only proceed to the Review screen if it's approved
+    if (
+      quote.requiresKyc() &&
+      getKycStatusResponse!.kycStatus[quote.getKycSchema()!] !== FiatConnectKycStatus.KycApproved
```
cajubelt, line 359:
```suggestion
      getKycStatusResponse?.kycStatus[quote.getKycSchema()!] !== FiatConnectKycStatus.KycApproved
```

cajubelt: also, same thing here re `requiresKyc` and `quote.getKycSchema()!`

src/fiatconnect/saga.ts
```diff
@@ -277,14 +280,57 @@ export function* handleSelectFiatConnectQuote({
   payload: params,
 }: ReturnType<typeof selectFiatConnectQuote>) {
   const { quote } = params
+
```
cajubelt, line 283: it seems like this saga does something much different on quote selection when kyc is required vs when it is not. However, after the user has gone through kyc, it seems as though we'll want the other stuff to still happen. Do you have a plan for that?

cajubelt: seems like a naive approach would be to create a new saga that is "everything after kyc" and invoke that (somehow) when the user has completed submitting their kyc info. 

Another (still fairly naive) would be to simulate selecting the quote again once the user has submitted kyc info

cajubelt (commented):
skimmed everything except the saga

src/fiatconnect/saga.ts
jophish: Yeah, I went back and forth between just having `getKycSchema` and using its falsiness as a proxy for knowing whether or not KYC is required, or using both `requiresKyc` and `getKycSchema`. I think using just `getKycSchema is fine; it will definitely make things cleaner.

src/fiatconnect/saga.ts
jophish: Hm, you mean after they've completed the Persona flow in particular, we would want to "resume" with the remainder of the saga (basically run the whole saga as it was before this PR) as normal? Simulating selecting the quote again (just dispatching the same action we do when a user selects a quote) isn't the worst idea imo, it's the most "robust" in that it actually ensures that the KYC states stored in IHL are as we expect, rather than just assuming they are after completing Persona KYC.

Also, after sending KYC to Persona, we'll need to hit POST /kyc for the provider they've selected; rather than doing that separately as an ad-hoc thing, that can just be done when calling this saga again, since the logic to do that already exists there. If we go this route, perhaps the naming of this saga could use an update, though.

src/fiatconnect/KycStatusScreen.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { StyleSheet } from 'react-native'
+import { SafeAreaView } from 'react-native-safe-area-context'
+
+function KycStatusScreen() {
```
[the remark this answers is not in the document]
jophish: Added

cajubelt (approved):

src/in-house-liquidity/index.test.ts
```diff
@@ -54,4 +65,217 @@ describe('In House Liquidity Calls', () => {
       ).toThrow('Cannot call IHL because walletAddress is null')
     })
   })
+
+  describe('makeRequest', () => {
+    it('makes a FC authenticated call', async () => {
+      ;(getAuthHeaders as jest.Mock).mockResolvedValueOnce({ 'header-key': 'header-val' })
```
satish-ravi, line 71: you can used [jest.mocked(...)](https://jestjs.io/docs/mock-function-api/#jestmockedsource-options) instead of `as`. Or actually `mocked` helper from `ts-utils/jest` since we are in an older version of jest

src/in-house-liquidity/index.test.ts
```diff
@@ -54,4 +65,217 @@ describe('In House Liquidity Calls', () => {
       ).toThrow('Cannot call IHL because walletAddress is null')
     })
   })
+
+  describe('makeRequest', () => {
+    it('makes a FC authenticated call', async () => {
```
satish-ravi, line 70: Can you also include `if provider info is passed / not passed` to the test name? Wasn't super obvious when reading these tests.
Also, should these tests also assert that the function returns correct response?

src/in-house-liquidity/index.test.ts
```diff
@@ -54,4 +65,217 @@ describe('In House Liquidity Calls', () => {
...
+        method: 'GET',
+      })
+    })
+  })
+
+  describe('getAuthHeaders', () => {
+    let mockFiatConnectClient: jest.Mocked<FiatConnectApiClient>
+
+    beforeEach(async () => {
+      mockFiatConnectClient = (await getFiatConnectClient(
```
satish-ravi, line 113: any reason to call getFiatConnectClient here? can't we just use the mock functions we define above as global vars?

src/navigator/Navigator.tsx
```diff
@@ -511,6 +513,12 @@ const settingsScreens = (Navigator: typeof Stack) => (
       name={Screens.FiatConnectLinkAccount}
       component={FiatConnectLinkAccountScreen}
     />
+    <Navigator.Screen options={headerWithBackButton} name={Screens.Persona} component={Persona} />
```
satish-ravi, line 516: I don't think the Persona component is a screen. It's currently just a button, which on clicking opens the persona flow. Is the plan to refactor this component to just load Persona flow on load? Or do we just want to refactor the persona stuff into it's own saga and drop that component?

src/fiatconnect/saga.ts
```diff
@@ -277,14 +280,60 @@ export function* handleSelectFiatConnectQuote({
...
+          }
+        // If approved or pending, continue as normal and handle account management
+        case FiatConnectKycStatus.KycApproved:
+        case FiatConnectKycStatus.KycPending:
+          break
+        // If denied or expired, skip account management and
+        // navigate to the KYC status screen
+        case FiatConnectKycStatus.KycDenied:
+        case FiatConnectKycStatus.KycExpired:
+          navigate(Screens.KycStatusScreen)
```
satish-ravi, line 320: This screen should eventually take flow, quote, and kyc status as props right?

src/in-house-liquidity/index.test.ts
jophish: I thought that calls to jest.mock were hoisted so you couldn't reference global vars within them?

src/fiatconnect/saga.ts
jophish: Yep, that's probably the case.

src/navigator/Navigator.tsx
jophish: Added a new `KycLanding` screen instead!

src/in-house-liquidity/index.test.ts
jophish: This introduces another issue with TS complaining about resolved result type :frowning_face:

satish-ravi (approved):

src/in-house-liquidity/index.test.ts
satish-ravi: I think this works if global variables are referenced inside another jest.fn(). E.g., https://github.com/valora-inc/wallet/pull/2855/files#diff-7caff590ac25eb7fe43d613aa98c7e1b1d4e472211ab028f2aa17b11b2514bc3R39

ValoraQA:
Hey @MuckT as discussed in weekly meet ,we are skipping this PR 

Thanks!

PR #2858, opened 2022-09-12
fix: invite modal display on android not taking up whole screen height

kathaypacific:
### Description

As the title says

### Other changes

Vertical padding on the NFT banner. Before -> after

Simulator Screen Shot - iPhone 11 - 2022-09-12 at 15 50 55

Simulator Screen Shot - iPhone 11 - 2022-09-12 at 15 49 46

### Tested
Manually
### How others should test

Verify that https://github.com/valora-inc/wallet/issues/2852 cannot be reproduced.

### Related issues

- Fixes #2852 

### Backwards compatibility

Yes

PR #2873, opened 2022-09-14
fix: prevent text overlapping the device status bar

kathaypacific:
### Description

Currently the app does not handle large fonts very well. In the CustomHeader component, the title is absolutely positioned and for users with large fonts enabled, the title breaches the device native status bar. This shouldn't be allowed.

Current:
Screenshot 2022-09-14 at 17 00 32

It's pretty hard to optimise headers with large text and 2 lines. I tested how Apple handles this natively in the Photos app and it appears that they do not allow font scaling in the headers. I took this approach as it seems sensible enough and technically simple, and at a later stage we can revisit this if we want to optimise for large fonts. Open to other ideas about how to handle this scenario.

Updated:
Screenshot 2022-09-14 at 17 01 44

### Other changes

N/A

### Tested

Manually

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/src/components/header/CustomHeader.tsx
+++ b/src/components/header/CustomHeader.tsx
@@ -27,6 +27,7 @@ const styles = StyleSheet.create({
     flexDirection: 'row',
     justifyContent: 'space-between',
     alignItems: 'center',
+    overflow: 'hidden',
   },
   buttonContainer: {
     height: '100%',
--- a/src/exchange/ExchangeScreenHeader.tsx
+++ b/src/exchange/ExchangeScreenHeader.tsx
@@ -50,8 +50,10 @@ function ExchangeTradeScreenHeader({ currency, isCeloPurchase, onChangeCurrency 
       titleText = t('tokenBalance', { token: currency })
       title = (
         <View style={styles.titleContainer} testID="HeaderCurrencyPicker">
-          <Text style={headerStyles.headerSubTitle}>{titleText}</Text>
-          <DownArrowIcon color={colors.gray3} />
+          <Text style={headerStyles.headerSubTitle} allowFontScaling={false}>
+            {titleText}
+          </Text>
+          <DownArrowIcon color={colors.gray3} height={16} />
         </View>
       )
     }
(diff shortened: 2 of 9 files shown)


kathaypacific:
@MuckT this PR should address your feedback https://github.com/valora-inc/wallet/pull/2858#issuecomment-1243986062

jacob-israel-turner (approved):

MuckT:
@kathaypacific it looks like the HeaderTitleWithSubtitle is used in at 15 screens. Is there a way to prevent the overlapping native status bar without disabling font scaling? I think it looks a bit strange to have such a small header and larger font on screen.

kathaypacific:
> @kathaypacific it looks like the `HeaderTitleWithSubtitle` is used in 15 screens. Is there a way to prevent the overlapping native status bar without disabling font scaling? I think it looks a bit strange to have such a small header and larger font on screen.

I think it's a fair concern with the number of screens affected. Idk what other options there are though, if we want a header with subtitle, that is 2 lines of text. The header reamremainsins at 44px (iOS) or 56px (Android) regardless of font size settings. For large text, the only option we have to make 2 lines fit inside the give height is to shrink the text? I'm open to other ideas but I also don't want to turn this into a big design task, since at some point we will need to think about large fonts on a more holistic level anyway...(e.g. one thing i can think of, is once the font gets too big then we only show 1 line instead of 2. but that is a logic change and idk how important the second line is for each screen)

MuckT:
@kathaypacific maybe setting a maxFontSizeMultiplier and allowing font scaling?

kathaypacific:
> @kathaypacific maybe setting a maxFontSizeMultiplier and allowing font scaling?

@MuckT this is what I initially tried but the font still becomes too big since you can enable pretty big fonts. In the end I had to set the maxFontSizeMultiplier pretty close to 1 to make it work well for truly large fonts, which is pretty much the same as disallowing font scaling

MuckT (approved):
LGTM!

MuckT:
@ValoraQA for testing please set your font size to large and ensure that the text does not overlap into the status bar as seen in the screenshots.

ValoraQA:
Hey @MuckT we have verified the above task on Latest iOS Mainnet Nightly Build V 1.42.0 and shared observation in the sheet mentioned below

Sheet link : [observation](https://docs.google.com/document/d/1OHfSGAHjMi49DHue9oF8gkO2v067bsyxbtBA0KnovKQ/edit)
Thanks.!

PR #2875, opened 2022-09-14
Not ready to merge

anthonydaoud:
### Description

### Other changes

### Tested

### How others should test

### Related issues

- Fixes #[issue number here]

### Backwards compatibility


src/navigator/NavigatorWrapper.tsx
```diff
@@ -70,6 +72,12 @@ export const NavigatorWrapper = () => {
   const isInviteModalVisible = useTypedSelector((state) => state.app.inviteModalVisible)
   const routeNameRef = React.useRef()
   const inSanctionedCountry = useTypedSelector(userInSanctionedCountrySelector)
+  Logger.info(
+    'NavigatorWrapper',
+    'Statsig exp layer: isLoading',
+    useLayer('name_and_picture_screen').isLoading
```
tore-statsig, line 78: you're calling useLayer outside the context of the StatsigProvider, so this wont work

src/analytics/ValoraAnalytics.ts
```diff
@@ -73,6 +80,7 @@ class ValoraAnalytics {
         throw Error('API Key not present, likely due to environment. Skipping enabling')
       }
       await Analytics.setup(SEGMENT_API_KEY, SEGMENT_OPTIONS)
+      await Statsig.initialize(STATSIG_API_KEY)
```
tore-statsig, line 83: Either use the provider, or use this.  Using both will get messy. If you want to handle the lifecycle yourself, then just do that here and use the apis on the global `Statsig`

If you were doing this and using `waitForInitialization`, this could cause an issue

PR #2886, opened 2022-09-16
feat: Add share URL flow

jacob-israel-turner:
### Description

This PR adds a new drawer navigation item "Invite". Users can open the Invite screen to get a deep link directing other users to the app.

### Other changes

 - Grabbing the share link kicks off an analytics event and on iOS, the results of the share (if it was abandoned, or what app was used for the sure) are attached as properties. On Android the results are not available.
 - Users opening the app via the invite link also kicks off an analytics event, with the inviter's wallet address as a property.

### Tested

These steps were followed on each platform:

Link creation:
1. Click the `Invite` link in the drawer navigator
2. The `Invite` button should initially be disabled, as the link is being created
3. When the `Invite` button is enabled, click it.
4. The native share drawer should appear, with a short message including the share link.

Clicking the link:
 - If on a mobile device and the app is not installed, should open valoraapp.com
 - If on a desktop, should open valoraapp.com
 - If on a mobile device and the app is installed, should open the app

### How others should test

I imagine we'll want to add this to QA test steps.

### Related issues

- Fixes EA-300

### Backwards compatibility

This PR should be backwards compatible. Link creation hinges off a remote config variable.

### Relevant UI:

https://user-images.githubusercontent.com/8698364/192322143-74af2dd9-80b8-44d4-98df-0f5c3077f697.mp4

https://user-images.githubusercontent.com/8698364/192322151-bb9bac8a-d2c6-49a7-8388-c7b83c15fadf.mp4

### DIFF
--- a/src/analytics/Events.tsx
+++ b/src/analytics/Events.tsx
@@ -261,6 +261,8 @@ export enum InviteEvents {
   invite_banner_impression = 'invite_banner_impression',
   invite_with_share = 'invite_with_share',
   invite_with_share_dismiss = 'invite_with_share_dismiss',
+  invite_with_referral_url = 'invite_with_referral_url',
+  opened_via_invite_url = 'opened_via_invite_url',
 }
 
 export enum EscrowEvents {
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -16,7 +16,7 @@ import { apolloClient } from 'src/apollo/index'
 import { appMounted, appUnmounted, openDeepLink } from 'src/app/actions'
 import AppLoading from 'src/app/AppLoading'
 import ErrorBoundary from 'src/app/ErrorBoundary'
-import { isE2EEnv } from 'src/config'
+import { FIREBASE_ENABLED, isE2EEnv } from 'src/config'
 import i18n from 'src/i18n'
 import I18nGate from 'src/i18n/I18nGate'
 import NavigatorWrapper from 'src/navigator/NavigatorWrapper'
@@ -100,31 +100,33 @@ export class App extends React.Component<Props> {
 
     Linking.addEventListener('url', this.handleOpenURL)
 
-    this.dynamicLinksRemoveListener = dynamicLinks().onLink(({ url }) =>
-      this.handleOpenURL({ url })
-    )
+    if (FIREBASE_ENABLED) {
+      this.dynamicLinksRemoveListener = dynamicLinks().onLink(({ url }) =>
+        this.handleOpenURL({ url })
+      )
 
-    // On Android, initial deep links are picked up by CleverTap - even if they were created by Firebase DynamicLinks.
-    // Breaking out here on Android avoids events being tracked multiple times.
-    if (Platform.OS === 'ios') {
-      const firebaseUrl = await dynamicLinks().getInitialLink()
+      if (Platform.OS === 'ios') {
+        const firebaseUrl = await dynamicLinks().getInitialLink()
 
-      if (firebaseUrl) {
-        await this.handleOpenURL({ url: firebaseUrl.url })
+        if (firebaseUrl) {
+          await this.handleOpenURL({ url: firebaseUrl.url })
+        }
+      }
+
+      // On Android, initial deep links are picked up by CleverTap - even if they were created by Firebase DynamicLinks.
+      // Breaking out here on Android avoids events being tracked multiple times.
+      if (Platform.OS === 'ios') {
+        const firebaseUrl = await dynamicLinks().getInitialLink()
+
+        if (firebaseUrl) {
+          await this.handleOpenURL({ url: firebaseUrl.url })
+        }
       }
     }
 
     const url = await Linking.getInitialURL()
     if (url) {
       await this.handleOpenURL({ url })
-    }
-
-    if (Platform.OS === 'ios') {
-      const firebaseUrl = await dynamicLinks().getInitialLink()
-
-      if (firebaseUrl) {
-        await this.handleOpenURL({ url: firebaseUrl.url })
-      }
     }
 
     this.logAppLoadTime()
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -76,6 +76,7 @@ export class App extends React.Component<Props> {
     await ValoraAnalytics.init()
 
     // Handles opening Clevertap deeplinks when app is closed / in background
+    // Also handles Firebase DynamcicLinks on Android
     CleverTap.getInitialUrl(async (err: any, url) => {
       if (err) {
         if (/CleverTap initialUrl is (nil|null)/gi.test(err)) {
@@ -99,23 +100,39 @@ export class App extends React.Component<Props> {
 
     Linking.addEventListener('url', this.handleOpenURL)
 
-    dynamicLinks().onLink(({ url }) => this.handleOpenURL(url))
+    this.dynamicLinksRemoveListener = dynamicLinks().onLink(({ url }) =>
+      this.handleOpenURL({ url })
+    )
+
+    // On Android, initial deep links are picked up by CleverTap - even if they were created by Firebase DynamicLinks.
+    // Breaking out here on Android avoids events being tracked multiple times.
+    if (Platform.OS === 'ios') {
+      const firebaseUrl = await dynamicLinks().getInitialLink()
+
+      if (firebaseUrl) {
+        await this.handleOpenURL({ url: firebaseUrl.url })
+      }
+    }
 
     const url = await Linking.getInitialURL()
     if (url) {
       await this.handleOpenURL({ url })
     }
 
-    const firebaseUrl = await dynamicLinks().getInitialLink()
+    if (Platform.OS === 'ios') {
+      const firebaseUrl = await dynamicLinks().getInitialLink()
 
-    if (firebaseUrl) {
-      await this.handleOpenURL({ url: firebaseUrl.url })
+      if (firebaseUrl) {
+        await this.handleOpenURL({ url: firebaseUrl.url })
+      }
     }
 
     this.logAppLoadTime()
 
     store.dispatch(appMounted())
   }
+
+  dynamicLinksRemoveListener: (() => void) | undefined
 
   logAppLoadTime() {
     const { appStartedMillis } = this.props
@@ -139,6 +156,7 @@ export class App extends React.Component<Props> {
   componentWillUnmount() {
     CleverTap.removeListener('CleverTapPushNotificationClicked')
     Linking.removeEventListener('url', this.handleOpenURL)
+    this.dynamicLinksRemoveListener?.()
     store.dispatch(appUnmounted())
   }
 
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -1,3 +1,4 @@
+import dynamicLinks from '@react-native-firebase/dynamic-links'
 import * as Sentry from '@sentry/react-native'
 import BigNumber from 'bignumber.js'
 import CleverTap from 'clevertap-react-native'
@@ -98,9 +99,17 @@ export class App extends React.Component<Props> {
 
     Linking.addEventListener('url', this.handleOpenURL)
 
+    dynamicLinks().onLink(({ url }) => this.handleOpenURL(url))
+
     const url = await Linking.getInitialURL()
     if (url) {
       await this.handleOpenURL({ url })
+    }
+
+    const firebaseUrl = await dynamicLinks().getInitialLink()
+
+    if (firebaseUrl) {
+      await this.handleOpenURL({ url: firebaseUrl.url })
     }
 
     this.logAppLoadTime()
--- a/src/app/saga.ts
+++ b/src/app/saga.ts
@@ -15,7 +15,7 @@ import {
   takeEvery,
   takeLatest,
 } from 'redux-saga/effects'
-import { AppEvents } from 'src/analytics/Events'
+import { AppEvents, InviteEvents } from 'src/analytics/Events'
 import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
 import {
   Actions,
@@ -265,6 +265,7 @@ export function* handleDeepLink(action: OpenDeepLink) {
 
   const rawParams = parse(deepLink)
   if (rawParams.path) {
+    const pathParts = rawParams.path.split('/')
     if (rawParams.path.startsWith('/v/')) {
       yield put(receiveAttestationMessage(rawParams.path.substr(3), CodeInputType.DEEP_LINK))
     } else if (rawParams.path.startsWith('/payment')) {
@@ -291,6 +292,10 @@ export function* handleDeepLink(action: OpenDeepLink) {
       // of our own notifications for security reasons.
       const params = convertQueryToScreenParams(rawParams.query)
       navigate(params.screen as keyof StackParamList, params)
+    } else if (pathParts.length === 3 && pathParts[1] === 'share') {
+      ValoraAnalytics.track(InviteEvents.opened_via_invite_url, {
+        inviterAddress: pathParts[2],
+      })
     }
   }
 }
--- a/src/invite/Invite.test.tsx
+++ b/src/invite/Invite.test.tsx
@@ -1,0 +1,94 @@
+import { fireEvent, render, waitFor } from '@testing-library/react-native'
+import { noop } from 'lodash'
+import React from 'react'
+import { Share } from 'react-native'
+import { Provider } from 'react-redux'
+import { mocked } from 'ts-jest/utils'
+
+import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
+import { createMockStore } from 'test/utils'
+
+import Invite from './Invite'
+import { createDynamicLink } from './utils'
+
+jest.mock('./utils', () => ({
+  createDynamicLink: jest.fn(),
+}))
+
+const { press } = fireEvent
+const mockedCreateDynamicLink = mocked(createDynamicLink)
+
+describe('Invite', () => {
+  beforeAll(() => {
+    jest.useRealTimers()
+  })
+
+  beforeEach(() => jest.clearAllMocks())
+
+  const getWrapper = () =>
+    render(
+      <Provider
+        store={createMockStore({
+          web3: {
+            account: '0xabc123',
+          },
+        })}
+      >
+        <Invite />
+      </Provider>
+    )
+
+  it('should disable button while loading share URL', () => {
+    mockedCreateDynamicLink.mockReturnValue(new Promise(noop))
+    const { getByTestId } = getWrapper()
+    expect(getByTestId('InviteModalShareButton')).toBeDisabled()
+  })
+
+  it('should enable button when share URL is loaded', async () => {
+    mockedCreateDynamicLink.mockResolvedValue('https://vlra.app/abc123')
+
+    const { getByTestId } = getWrapper()
+
+    await waitFor(() => true) // Wait one tick for the createDynamicLink promise to resolve
+
+    expect(getByTestId('InviteModalShareButton')).not.toBeDisabled()
+  })
+
+  it('should share when button is pressed', async () => {
+    mockedCreateDynamicLink.mockResolvedValue('https://vlra.app/abc123')
+
+    jest.spyOn(Share, 'share')
+    ;(Share.share as jest.Mock).mockResolvedValue({
+      action: Share.sharedAction,
+      activityType: 'clipboard',
+    })
+
+    jest.spyOn(ValoraAnalytics, 'track')
+
+    const { getByTestId } = getWrapper()
+
+    await waitFor(() => true) // Wait one tick for the createDynamicLink promise to resolve
+
+    await press(getByTestId('InviteModalShareButton'))
+
+    expect(Share.share).toHaveBeenCalledTimes(1)
+    expect((Share.share as jest.Mock).mock.calls[0]).toMatchInlineSnapshot(`
+      Array [
+        Object {
+          "message": "inviteWithUrl.share, {\\"shareUrl\\":\\"https://vlra.app/abc123\\"}",
+        },
+      ]
+    `)
+
+    expect(ValoraAnalytics.track).toHaveBeenCalledTimes(1)
+    expect((ValoraAnalytics.track as jest.Mock).mock.calls[0]).toMatchInlineSnapshot(`
+      Array [
+        "invite_with_referral_url",
+        Object {
+          "action": "sharedAction",
+          "activityType": "clipboard",
+        },
+      ]
+    `)
+  })
+})
--- a/src/invite/Invite.tsx
+++ b/src/invite/Invite.tsx
@@ -1,102 +1,27 @@
-import { noop } from 'lodash'
+import { isNil, noop } from 'lodash'
 import React from 'react'
 import { useTranslation } from 'react-i18next'
-import { Image, ScrollView, StyleSheet, Text, View } from 'react-native'
-import { SafeAreaView } from 'react-native-safe-area-context'
 
-import Button, { BtnSizes } from 'src/components/Button'
-import Touchable from 'src/components/Touchable'
-import ShareIcon from 'src/icons/Share'
-import Times from 'src/icons/Times'
-import { inviteModal } from 'src/images/Images'
 import { noHeader } from 'src/navigator/Headers'
 import { navigateBack } from 'src/navigator/NavigationService'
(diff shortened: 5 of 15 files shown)


src/navigator/Navigator.tsx
```diff
@@ -384,6 +385,16 @@ const consumerIncentivesScreens = (Navigator: typeof Stack) => (
   </>
 )
 
+const inviteScreens = (Navigator: typeof Stack) => (
+  <>
+    <Navigator.Screen
+      name={Screens.Invite}
+      component={Invite}
+      options={ConsumerIncentivesHomeScreen.navOptions}
```
jacob-israel-turner, line 393: Whoops - don't use `ConsumerIncentivesHomeScreen.navOptions` - pull them off `Invite` instead.

src/app/saga.ts
```diff
@@ -290,6 +290,11 @@ export function* handleDeepLink(action: OpenDeepLink) {
       // of our own notifications for security reasons.
       const params = convertQueryToScreenParams(rawParams.query)
       navigate(params.screen as keyof StackParamList, params)
+    } else if (rawParams.path.startsWith('/share')) {
+      const inviterAddress = rawParams.path.split('/')[2]
```
MuckT, line 294: If the inviter address is always the last instead of hardcoding it to get the second could we always get the last with `rawParams.path.split('/').pop()` instead?

src/invite/Invite.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+const styles = StyleSheet.create({
+  art: {
+    width: 136,
+    height: 120,
+  },
+  outerContentContainer: {
+    flex: 1,
+    justifyContent: 'center',
+    paddingBottom: 56 * 2,
```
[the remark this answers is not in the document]
kathaypacific: this could get rid of these magic numbers for size and padding

test/schemas.ts
```diff
@@ -1616,6 +1616,14 @@ export const v76Schema = {
   },
 }
 
+export const v77Schema = {
```
kathaypacific, line 1619: you don't need a new migration to augment the screens enum

src/invite/Invite.utils.tsx
```diff
@@ -0,0 +1,8 @@
+import dynamicLinks from '@react-native-firebase/dynamic-links'
```
kathaypacific, line 1: there's some strong sentiment among the team that we should stop creating util files because they become a dumpling ground for poorly thought out abstractions. this is not the case here but i would suggest to not create a new utils file for this...

kathaypacific (commented):
Since the invite referral link stays the same for the user, we only have to create this one time right? Currently you create this every time the user tries to access the screen, but can we instead consider doing this on app start -> storing it in redux -> component read from redux -> if it's already in redux then we can skip the step on app start? this would simplify your component a lot.

src/app/App.tsx
```diff
@@ -98,11 +99,19 @@ export class App extends React.Component<Props> {
...
     Linking.addEventListener('url', this.handleOpenURL)
 
+    dynamicLinks().onLink(({ url }) => this.handleOpenURL(url))
+
     const url = await Linking.getInitialURL()
     if (url) {
       await this.handleOpenURL({ url })
     }
 
+    const firebaseUrl = await dynamicLinks().getInitialLink()
```
kathaypacific, line 109: i think this might be nicer in the firebase sagas instead of here? what do you think? (or at least, keep the dynamic links logic together, without the Linking logic in between)

src/invite/Invite.hooks.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+jest.mock('react-i18next', () => ({
+  useTranslation: jest.fn().mockReturnValue({
+    t: (name: string) => name,
+  }),
+}))
+
+jest.mock('react', () => ({
+  useState: jest.fn().mockReturnValue([null, () => null]),
+  useEffect: jest.fn().mockImplementation((cb) => cb()),
```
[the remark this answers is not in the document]
jacob-israel-turner: I'm open to suggestions. Testing hooks is always a little weird, mocking React hooks seems to give me the best results. I only mock React in the hooks test file, so testing the render of the `Invite` component still uses non-mocked React.

src/invite/Invite.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { noop } from 'lodash'
+import React from 'react'
+import { Provider } from 'react-redux'
+import { useShareUrl } from 'src/invite/Invite.hooks'
+
+import { createMockStore } from 'test/utils'
+import Invite from './Invite'
+
+jest.mock('./Invite.hooks', () => ({
+  useShareUrl: jest.fn().mockReturnValue(null),
```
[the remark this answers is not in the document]
jacob-israel-turner: I'm open to it! I'm assuming you mean, test the hook via e2e tests?

test/schemas.ts
jacob-israel-turner: @kathaypacific I was wondering about that. I only added it because `test/RootStateSchema.json` was changed.

Am I ok to just remove all the schema update stuff, and only commit the updated `RootStateSchema.json`?

src/navigator/DrawerNavigator.tsx
```diff
@@ -305,14 +305,16 @@ export default function DrawerNavigator() {
...
         options={{ title: t('addAndWithdraw'), drawerIcon: AddWithdraw }}
       />
-      {features.SHOW_INVITE_MENU_ITEM && (
+      {inviteMethod === InviteMethodType.ReferralUrl && (
         <Drawer.Screen
           name={'InviteModal'}
-          component={InviteFriendModal}
+          component={() =>
+            null
+          } /* Drawer.Screen requires a component to be passsed in. However, we need to use `navigate` instead. Passing in a noop. */
```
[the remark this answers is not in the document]
jacob-israel-turner: Because the `Drawer.Screen` opens the new screen as a different screen type than was indicated in the designs. I found that the Supercharge screen in the app follows the same screen type, so I opened Invite using the same kind of logic.

I'd be happy to make a recording showing the difference in screen types, if it's helpful.

src/invite/Invite.utils.tsx
jacob-israel-turner: OK - where would you like this logic to live instead?

src/invite/Invite.tsx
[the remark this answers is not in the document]
jacob-israel-turner: I'll take a look and see if I can reduce it to one container. I do remember it was needed to make flexbox work properly, but there might be another way.

src/invite/Invite.test.tsx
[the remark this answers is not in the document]
kathaypacific: oh actually i meant we can just test the Invite component without mocking out the hook? as in, cover the unit testing of the hook inside this component test - what do you think? e.g. there are some hooks that are tested as part of bigger components in `transferFeedUtils.ts`. i guess it's what you think makes sense and brings value, i would question whether this component test without the hook logic makes sense but that's my personal opinion. if you move the referral link calculation to redux, you might not need a hook at all though :)

test/schemas.ts
kathaypacific: yes - the migrations are to safely migrate persisted state to a new app version. if we're augmenting an enum, it's safe for users with persisted state so we don't need a migration :)

src/navigator/DrawerNavigator.tsx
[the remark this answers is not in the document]
kathaypacific: oh yeah if you could include a recording, that'd be great! the (supercharge) ConsumerIncentivesHomeScreen appears in the drawer navigator like `component={ConsumerIncentivesHomeScreen}` rather than`component={() => null}` so i'm not sure that i follow or understand the details here :)

src/invite/Invite.utils.tsx
kathaypacific: if you go down the route generating this link and storing in redux, then maybe in the firebase saga?

src/app/App.tsx
kathaypacific: okay, that makes sense too. i was thinking about this as being similar to firebase cloud messaging and we have some code to initialise that in the firebase saga / a firebase file where we handle opening the app based on a deep link. indeed dynamic links aren't exactly the same though and it's debatable whether grouping these is the best approach. so i am also happy to leave it here. would like to get @jeanregisser 's thoughts maybe?

src/app/App.tsx
```diff
@@ -98,11 +99,19 @@ export class App extends React.Component<Props> {
...
+
+    if (firebaseUrl) {
+      await this.handleOpenURL({ url: firebaseUrl.url })
+    }
```
jeanregisser, line 113: We should wrap these under the `FIREBASE_ENABLED` flag.

What happens if firebase hasn't been initialized? There may be a race condition if the saga doing the firebase init hasn't finished by the time this is called.

src/invite/Invite.hooks.test.tsx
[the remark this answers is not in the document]
jeanregisser: +1 to what @kathaypacific said.
Can we test the hook inside a dummy comment like we did [here](https://github.com/valora-inc/wallet/blob/2b28b691640c52bdeeb1b49a7751c5bdf3b764fe/src/tokens/hooks.test.tsx) instead?

src/invite/Invite.hooks.tsx
```diff
@@ -0,0 +1,36 @@
+import { isNil } from 'lodash'
```
jeanregisser, line 1: nit: Do we really need this? Is it better than `x != null`?

jeanregisser (commented):
Great work! 🚀:rocket:

Left mostly small comments.

Couple of additional things:
- could you include screenshots and/or videos when doing UI changes? This tremendously helps the reviewers.
- if your PR is not yet ready to be merged, could you switch it to a draft PR?
- nit: could you follow the existing file naming convetion? Instead of having `Invite.hooks.tsx` just name it `hooks.tsx` as it's already in the `invite` folder. Same for `Invite.utils.tsx`.

If you haven't already I also encourage you to take a look at our [best practices for PRs](https://www.notion.so/valora-inc/Pull-Requests-Best-Practices-85d7344b314843478de3db9b3677e4df).

src/app/App.tsx
jacob-israel-turner: Ah yes - I was going to ask about that. I'll add the flag and test to see.

src/invite/Invite.hooks.tsx
jacob-israel-turner: Need? No, not at all. But I find `isNil` to be more expressive than `!=` with no cost. I have similar opinions for many lodash methods. If you have strong opinions I'm happy to have a conversation.

src/app/saga.ts
```diff
@@ -290,6 +290,11 @@ export function* handleDeepLink(action: OpenDeepLink) {
       // of our own notifications for security reasons.
       const params = convertQueryToScreenParams(rawParams.query)
       navigate(params.screen as keyof StackParamList, params)
+    } else if (rawParams.path.startsWith('/share')) {
```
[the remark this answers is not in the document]
jacob-israel-turner: I see the concern, but your solution does not work. `path` and `pathname` have the same value: `/share/:address`.

However, I can check the number of parts, and only go into this block if it's exactly the same length. That way if we add something like `/share/:nftId/:address`, they won't conflict.

src/app/App.tsx
jacob-israel-turner: @jeanregisser let me know if you're satisfied with this solution.

src/invite/Invite.hooks.test.tsx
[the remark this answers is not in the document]
jacob-israel-turner: @kathaypacific @jeanregisser let me know how you feel about my solution. I tested both the hook and the component in one file. I ended up mocking the utils without testing, since there's very little to test there, and I ran into some troubles mocking DynamicLinks properly. I'm happy to take another pass if you feel we're lacking coverage anywhere.

src/invite/Invite.hooks.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+    ;(useEffect as jest.Mock).mockImplementation((cb) => cb())
+
+    const cb = useShareUrl()
+
+    expect(cb).toBeNull()
+  })
+
+  it('should return callback', async () => {
+    const mockSetState: jest.Mock = jest.fn()
+    ;(useSelector as jest.Mock).mockReturnValue('0xabc123')
```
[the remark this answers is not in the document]
jacob-israel-turner: Awesome tool! Very handy, and much more readable than casting.

Unfortunately, it doesn't work with `jest.spyOn`, which is required to mock React Native modules, without mocking out the entire React Native library. If you know of any solutions that work with `spyOn`, let me know!

src/invite/Invite.tsx
```diff
@@ -0,0 +1,106 @@
...
+  },
+  scrollView: {
+    alignItems: 'center',
+    marginHorizontal: 24,
+    flex: 1,
+  },
+  headerContainer: {
+    height: 56,
+    width: '100%',
+    marginTop: 12,
```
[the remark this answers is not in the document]
jacob-israel-turner: OK, I was able to get rid of some of these using the Spacing enum. But some of the values (inner content and art dimensions) are just very specific in the designs and don't fit anything in the enum.

src/invite/Invite.tsx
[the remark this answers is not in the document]
jacob-israel-turner: Yeah I briefly tried reducing this back to one container, but it was difficult to get the spacing to match the designs without adding more magic numbers.

TBH styling has always annoyed me, so if you have any ideas you'd like to try, I'd be happy to jump on a call and try them out!

test/schemas.ts
jacob-israel-turner: OK, I removed the migration! I just want to make sure I understand though - so Redux enums are not persisted at all, and pulled from the code every time the app is started - is that correct?

src/navigator/DrawerNavigator.tsx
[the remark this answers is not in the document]
jacob-israel-turner: I was able to get this switched to match the other screens!

src/app/saga.ts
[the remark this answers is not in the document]
jeanregisser: Yes I didn't pay attention to the exact format of the link. 
What you're proposing seems good.

.env.alfajores
```diff
@@ -11,4 +11,5 @@ APP_STORE_ID=1482389446
 APP_DISPLAY_NAME=Alfajores
 IOS_GOOGLE_SERVICE_PLIST=GoogleService-Info.alfajores.plist
 DYNAMIC_DOWNLOAD_LINK=https://vlra.app/invite
+DYNAMIC_LINK_DOMAIN_URI_PREFIX=https://vlra.app
```
kathaypacific, line 14: i know that `DYNAMIC_DOWNLOAD_LINK` lives here already....but it feels weird that we have an env variable that is not going to be different between envs? if it's going to be bundled with the app anyway, should we just have the value of this in config.json and not in the env files as well?

src/analytics/Events.tsx
```diff
@@ -261,6 +261,8 @@ export enum InviteEvents {
   invite_banner_impression = 'invite_banner_impression',
   invite_with_share = 'invite_with_share',
   invite_with_share_dismiss = 'invite_with_share_dismiss',
+  invite_with_url = 'invite_with_url',
```
kathaypacific, line 264: `invite_with_referral_url`? :D

src/app/App.tsx
```diff
@@ -98,6 +100,20 @@ export class App extends React.Component<Props> {
 
     Linking.addEventListener('url', this.handleOpenURL)
 
+    if (FIREBASE_ENABLED) {
+      this.dynamicLinksRemoveListener = dynamicLinks().onLink(({ url }) =>
+        this.handleOpenURL({ url })
+      )
+
+      if (Platform.OS === 'ios') {
```
kathaypacific, line 108: would be nice to add a comment about why we're not doing anything with android here (i'm curious for why as well :))

src/app/App.tsx
```diff
@@ -108,6 +124,8 @@ export class App extends React.Component<Props> {
     store.dispatch(appMounted())
   }
 
+  dynamicLinksRemoveListener: (() => void) | undefined
```
kathaypacific, line 127: would it be clearer to have this as something like
```
dynamicLinksRemoveListener() {
  if (FIREBASE_ENABLED) {
    ...
  }
}
```
instead of having this undefined and assigned in a different part of the code? would also remove the need for the optional chaining on line 151

src/app/saga.ts
```diff
@@ -291,6 +292,10 @@ export function* handleDeepLink(action: OpenDeepLink) {
       // of our own notifications for security reasons.
       const params = convertQueryToScreenParams(rawParams.query)
       navigate(params.screen as keyof StackParamList, params)
+    } else if (pathParts[1] === 'share' && pathParts.length === 3) {
```
kathaypacific, line 295: should we swap the order of this condition, so that we ensure `pathParts[1]` is not going to cause issues in case the length is 0?

src/app/saga.ts
```diff
@@ -291,6 +292,10 @@ export function* handleDeepLink(action: OpenDeepLink) {
...
+      ValoraAnalytics.track(InviteEvents.opened_via_invite_url, {
+        inviterAddress: pathParts[2],
```
kathaypacific, line 297: would be nice to add a test i guess?

src/invite/Invite.tsx
```diff
@@ -0,0 +1,108 @@
...
+          size={BtnSizes.SMALL}
+          disabled={isNil(shareUrl)}
+          onPress={shareUrl ?? noop}
+        />
+      </View>
+    </View>
+  )
+}
+
+export default function Invite() {
```
kathaypacific, line 54: Simulator Screen Shot - iPhone 11 - 2022-09-26 at 09 41 00

hmm, the styles don't look quite right?

also, how do you feel about removing the abstractions for Header and Content? in my opinion these abstractions don't really make the component easier to understand and i think you can simplify the component / styles if the abstractions were removed. but as i am typing this i remember that actually the InviteOptionsModal has the [same design](https://www.figma.com/file/8ig95B4sfsYIDIpg3JGP03/Phone-Verification?node-id=1378%3A21235)  - i think it makes sense to reuse the same UI components. i will abstract out the UI code, i should have done this before :)

src/invite/Invite.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+jest.mock('./utils', () => ({
+  createDynamicLink: jest.fn(),
+}))
+
+const { press } = fireEvent
+const mockedCreateDynamicLink = mocked(createDynamicLink)
+
+describe('Invite', () => {
+  beforeAll(() => {
+    jest.useRealTimers()
```
kathaypacific, line 23: do we need real timers? can we do `jest.runAllTimers` when we need instead?

src/invite/hooks.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const [shareUrl, setShareUrl] = useState<string | null>(null)
+
+  const address = useSelector(walletAddressSelector)
+  const { t } = useTranslation()
+
+  const message = isNil(shareUrl) ? null : t('inviteWithUrl.share', { shareUrl })
+
+  useAsync(async () => {
+    if (isNil(address) || !FIREBASE_ENABLED) return
+    const url = await createDynamicLink(address)
```
kathaypacific, line 25: we should probably handle the case where this fails, even if it is a very edge case. i'd propose to log the error using Logger.debug so that we can debug if users contact support. perhaps showing the global error banner is also a good idea (you can do something like `dispatch(showError(...))`

kathaypacific (commented):
Looking really good!!

src/app/App.tsx
jacob-israel-turner: There is a comment on line `79`, but I agree it would be helpful to clarify here too! I'll get it added.

src/app/App.tsx
jacob-israel-turner: This line is only the type definition, not the implementation. The implementation is attached [here](https://github.com/valora-inc/wallet/pull/2886/files#diff-7c79023b16fd616788af0fdcb31d619ae04da7259aa564fd72fb0ea07d8ebe07R104), which is already wrapped in the `FIREBASE_ENABLED` check.

src/app/saga.ts
jacob-israel-turner: Hmmm - `undefined === 'share'` shouldn't throw, if that's what you're thinking. I'm happy to swap it though.

src/app/saga.ts
jacob-israel-turner: Sure thing, I'll do that today.

src/invite/Invite.tsx
jacob-israel-turner: What device is this on? I'm not seeing styles like that, but I've been testing on a large iOS simulator and Android device.

I don't have a super strong opinion on Header and Content. It's just a minor code organization pattern that I find helpful. I can squash it into one component if you want.

src/invite/Invite.test.tsx
jacob-israel-turner: I added this for the `waitFor` call. I can try it with `runAllTimers` instead, but it will be more verbose.

src/invite/hooks.tsx
jacob-israel-turner: Sounds good to me, I'll add that today.

src/invite/Invite.tsx
jacob-israel-turner: OK, I've swapped out my UI for the `InviteModal` component. I think the tests I had already written are now _less_ relevant, but I left them since they do still add a little value. Let me know if you have any concerns with the changes I made.

src/invite/Invite.test.tsx
jacob-israel-turner: I just tried using `waitFor` with default timers, trying to progress with `runAllTimers`, but no dice. There might be something weird in the `waitFor` implementation. In theory we should only need to wait one tick on those lines, as the promise should resolve immediately once the call stack is clear, so I tried creating my own `waitOneTick` helper function, but it didn't work either. IMO the juice isn't worth the squeeze here, but I'm happy to jam on this a little longer if you feel otherwise.

src/invite/hooks.tsx
jacob-israel-turner: Done. I'm open to feedback on copy and Logger error formatting.

kathaypacific (commented):
LGTM! 

I can't think of a good way to test the flow of sending referral link to a user who does not have Valora installed -> user installs Valora -> verifying the correct analytics is fired, since it requires the new downloaded app to have this code :/ do we have any confidence that this scenario will work as expected?

going through the experiment doc now, the above scenario seems the most relevant for the analysis. if we are not confident that it works, i think we can still assess the success of the experiment by seeing how many new users consumed the dynamic link? should we also add an event that is fired inside the App.tsx? (idk how this will work for Android though, if the dynamic link is being hijacked by CleverTap)

and i guess one other thing to make sure is that the analytics event is not fired if the user did not download the app (i.e. the app was already installed) - this will be important to enable invite rewards. i *think* from reading the code that we're not handling this correctly right now @jacob-israel-turner @jeanregisser ? i am thinking that it would be better to move this event to the App.tsx handlers, but i might be missing some details, lmk

kathaypacific (approved):

src/app/App.tsx
```diff
@@ -98,6 +100,30 @@ export class App extends React.Component<Props> {
...
+
+        if (firebaseUrl) {
+          await this.handleOpenURL({ url: firebaseUrl.url })
+        }
+      }
+    }
+
+    // On Android, initial deep links are picked up by CleverTap - even if they were created by Firebase DynamicLinks.
+    // Breaking out here on Android avoids events being tracked multiple times.
+    if (Platform.OS === 'ios') {
```
kathaypacific, line 119: is it possible to move this inside the `if (FIREBASE_ENABLED)` block above? it seems like doing this would fix the e2e tests (i just tried on my local)

kathaypacific (commented):
Can you please update the title of this PR to be conventional commit format? When PRs are merged, the PR title becomes the commit message in main

ValoraQA:
Hey @MuckT currently we are not able to see "Invite" option in Hamburger Menu as shown in the video attached
Please let us know when the feature flag is enabled 

cc: @jacob-israel-turner 

Thanks.!

PR #2887, opened 2022-09-16
feat(coinbase): Add Coinbase logo to transaction feed and details screen

awtsui:
### Description

This PR adds the coinbase logo to the transaction feed, as well as the details screen.

### Other changes

N/A

### Tested

No testing was written as there are only icon changes

### How others should test

N/A

### Related issues

- Fixes [ACT-399](https://linear.app/valora/issue/ACT-399/add-coinbase-logo-to-transaction-feed)

### Backwards compatibility

N/A


awtsui:
From what I can tell, the checks are failing as coinbase-pay.png is not integrated into the wallet yet. This [PR](https://github.com/valora-inc/valora-app-branding/pull/12) should add the image.

src/transactions/feed/TransferFeedIcon.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { useTransactionRecipient } from 'src/transactions/transferFeedUtils'
+import { TokenTransactionTypeV2 } from 'src/transactions/types'
+
+const AVATAR_SIZE = 40
+
+interface Props {
+  transfer: FeedTokenTransfer
+}
+
+export default function TransferFeedIcon({ transfer }: Props) {
```
cajubelt, line 16: can you explain why we need this component in addition to the existing `src/transactions/TransferFeedIcon.tsx`?

src/transactions/feed/TransferFeedIcon.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  const coinbasePaySenders = useSelector(coinbasePaySendersSelector)
+  const recipient = useTransactionRecipient(transfer)
+
+  switch (type) {
+    // Checks if feed item is a received transaction from Coinbase
+    case TokenTransactionTypeV2.Received: {
+      const isCoinbasePaySender = coinbasePaySenders.includes(address)
+
+      if (isCoinbasePaySender) {
```
cajubelt, line 27: if this evaluates to false, I think the function will return undefined

cajubelt:
looks like this relies on https://github.com/valora-inc/valora-app-branding/pull/12

src/transactions/feed/TransferFeedIcon.tsx
awtsui: There seems to be separate set of transaction feed components, a set that works with TokenTransactionTypeV2 and another with TokenTransactionType. The new TransferFeedIcon works with TokenTransactionTypeV2, so following the trend of files under /transaction/feed, I made the new component.

src/transactions/feed/TransferFeedIcon.tsx
awtsui: I imitated the switch case statement from [here](https://github.com/valora-inc/wallet/blob/main/src/transactions/TransferFeedIcon.tsx). I believe if the if statement evaluates to false, default will be called.

jeanregisser:
Hey sorry I'm late to the party, this feels like the wrong place to make this change.
We already have the ability to customize the feed items (images and title) via [blockchain-api](https://github.com/valora-inc/blockchain-api/blob/8daefae023b6e11b470ee6eb384d7edc026839bf/src/helpers/KnownAddressesCache.ts) and the https://github.com/valora-inc/address-metadata repo.

Maybe the `address-metadata` is too static for this use case, but if that's the case `blockchain-api` should be adapted instead. We don't really want specific wallet changes for this each new sender type.

PR #2890, opened 2022-09-19
feat: skip button for user name during onboarding 

XrayBravoGolf:
### Description

Adds a skip button to the "name" screen of the user onboarding, and internally leaves the username empty when that is clicked. Clicking skip gets the user to the next screen.
Written with the intention to allow this feature to be toggled by Statsig when Statsig is integrated.

### Other changes
None

### Tested

Automated tests are written for `NameAndPicture.tsx`,  in which both showing and not showing skip button are tested.
Manual testing were done for all references to the username with an empty one, see screenshots below.

### How others should test

After Statsig integration, selecting the control for the skip button during onboarding username would result in skip button being shown. The skip button should work and take user to next onboarding screen.

### Related issues

- Fixes [ACT-436](https://linear.app/valora/issue/ACT-436/skip-button-on-name-screen)

### Backwards compatibility

Compatible; the config values are hard-coded.


XrayBravoGolf:
Screenshots for places where username were displayed:

Simulator Screen Shot - iPhone 13 - 2022-09-19 at 04 11 00
Simulator Screen Shot - iPhone 13 - 2022-09-19 at 03 57 33
Simulator Screen Shot - iPhone 13 - 2022-09-19 at 09 48 53

jh2oman (changes requested):
On the QR code page instead of the name it now displays the user's address twice. We should probably remove one of them and only display one.

XrayBravoGolf:
> On the QR code page instead of the name it now displays the user's address twice. We should probably remove one of them and only display one.

TL;DR: may be engineering-wise expensive as it disrupts how the components and Reciepient logic were re-used.

Good idea but there is a difficulty related to how a `Recipient` object is passed around from `<QRCode>` to `<AvatarSelf>` to `<Avatar>`, and how the `getDisplayName` [here](https://github.com/valora-inc/wallet/blob/0a126d2736c44befc7ff107b26c2d72e6cbe6d2b/src/recipients/recipient.ts#L58).

After some review it looks like the `<Avatar>` is a re-used general component while the `<AvatarSelf>` is a special component that passes `undefined` to `<Avatar>` and lets the latter handle how it's displayed. Inside that, `getDisplayName` is called with the `Recipient` object and actual logic to determine the name is run; since the recipient always have an address it becomes the fallback display name.

 The `getDisplayName` is re-used in many places and `<Avatar>`is also re-used. Even if all occurences are handled, at lease one extra parameter/prop would probably be needed solely for the purpose of supporting this need of not showing address twice, which really increases coupling betweem parts of code.

Alternatively, inside `<AvatarSelf>`, passing an empty string (instead of `undefined`) actually does not affect the subsequent logic as empty string is always falsy. Passing in a space or any "truthy empty string"`' '` would trigger the logic that generates profile pictures using name initials, so wouldn't directly work either.

XrayBravoGolf:
> @XrayBravoGolf I don't understand the difficulty.
> 
> One solution I thought of: Don't show the smaller address on the bottom of the avatar if the name is undefined. This can be achieved by simply adding a `&& recipient.name` to this line https://github.com/valora-inc/wallet/blob/main/src/components/Avatar.tsx#L48. Does this have any unintended side effects that I'm not seeing?

Not likely, especially since escrow is going away from Valora. I clung to removing the name and keeping the address, but yea, removing the address and keeping the name looks about the same.

XrayBravoGolf:
An updated screenshot for QR code
Simulator Screen Shot - iPhone 13 - 2022-09-23 at 03 58 15

src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -84,8 +93,15 @@ function NameAndPicture({ navigation }: Props) {
...
-
+  const onPressSkip = () => {
+    // TODO additional anlytics
+    handleNewName({ skipped: true })
+  }
   const onPressContinue = () => {
+    handleNewName({ skipped: false })
+  }
+
+  const handleNewName = ({ skipped }: { skipped: boolean }) => {
```
jh2oman, line 104: It appears that if a user starts typing and then clicks skip, the name will still be saved. This is not expected behavior

jh2oman: When you fix this, please also add a test.

jh2oman (changes requested):
Currently the name is still saved if the user presses skip

src/onboarding/registration/NameAndPicture.tsx
XrayBravoGolf: Thanks for catching it, this definitely needs a good change.

XrayBravoGolf: I got a test added, but for some reason I cannot reproduce the bug; see my reply in this thread or #activation-eng

jh2oman (changes requested):
Pressing the 'skip' button should do exactly as it says, skip the setting name step. Setting name as an empty string is not the same thing.

src/onboarding/registration/NameAndPicture.test.tsx
```diff
@@ -179,4 +182,38 @@ describe('NameAndPictureScreen', () => {
...
+  it('does not render skip button when configured so', () => {
+    // TODO add mock Statsig flag
+    const { queryByText } = render(
+      <Provider store={createMockStore()}>
+        <MockedNavigator component={NameAndPicture} />
+      </Provider>
+    )
+    expect(queryByText('skip')).toBeNull()
+  })
+  // eslint-disable-next-line jest/no-disabled-tests
```
cajubelt, line 194: the linter encodes a Valora style guide, and shouldn't be disabled unless there's a really good reason for it. In this case, `it.todo` could be used instead, so I don't see a reason to disable linting and use `it.skip`

src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -84,8 +93,16 @@ function NameAndPicture({ navigation }: Props) {
       })
     }
   }
-
+  const onPressSkip = () => {
+    // TODO additional anlytics
```
cajubelt, line 97: did you mean to do this in this PR? if not, is it in a Linear issue?

src/onboarding/registration/NameAndPicture.test.tsx
```diff
@@ -179,4 +182,38 @@ describe('NameAndPictureScreen', () => {
...
+      <Provider store={createMockStore()}>
+        <MockedNavigator component={NameAndPicture} />
+      </Provider>
+    )
+    expect(queryByText('skip')).toBeTruthy()
+    fireEvent.press(queryByText('skip')!)
+    expect(navigate).toHaveBeenCalledWith(Screens.PincodeSet, expect.anything())
+  })
+  // eslint-disable-next-line jest/no-disabled-tests
+  it.skip('saves empty name reardless of what is in the inputbox when skip is used', () => {
```
cajubelt, line 208:
```suggestion
  it.skip('saves empty name regardless of what is in the inputbox when skip is used', () => {
```

cajubelt:
looks like test coverage is an issue, possibly due to the hardcoded feature flag. Maybe that can be given as a prop for now, and default to false. Or it could be retrieved via some imported helper which can be mocked in tests.

XrayBravoGolf:
@cajubelt Combining the test coverage and skipping-tests I'll put the feature flag in a function and mock it.

jh2oman (approved):

src/onboarding/registration/NameAndPicture.tsx
cajubelt: in case anyone's reading along and curious about this^, we found that the name needed to be added as a dep for a react hook, so the hook would get evaluated again when the name changed

PR #2891, opened 2022-09-19
feat: pending swap screen

MuckT:
### Description

WIP

### Other changes

### Tested

### How others should test

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

### DIFF
--- a/src/swap/SwapPending.tsx
+++ b/src/swap/SwapPending.tsx
@@ -1,0 +1,49 @@
+import React from 'react'
+import { ActivityIndicator, StyleSheet, Text, View } from 'react-native'
+import { SafeAreaView } from 'react-native-safe-area-context'
+import { noHeader } from 'src/navigator/Headers'
+import colors from 'src/styles/colors'
+import fontStyles from 'src/styles/fonts'
+import { Spacing } from 'src/styles/styles'
+
+export function SwapPending() {
+  return (
+    <SafeAreaView style={{ flex: 1 }}>
+      <View style={styles.loadingContentContainer}>
+        <ActivityIndicator
+          size="large"
+          color={colors.greenBrand}
+          testID="SwapPending/loading"
+          style={styles.activityIndicator}
+        />
+        <Text style={styles.loadingText}>Swap Pending</Text>
+        <Text style={styles.loadingSubText}>Your swap is being processed</Text>
+      </View>
+    </SafeAreaView>
+  )
+}
+
+const styles = StyleSheet.create({
+  loadingContentContainer: {
+    alignItems: 'center',
+    justifyContent: 'center',
+    padding: Spacing.Regular16,
+    flexGrow: 1,
+  },
+  activityIndicator: {
+    marginBottom: 30,
+  },
+  loadingText: {
+    ...fontStyles.h2,
+    marginBottom: 16,
+  },
+  loadingSubText: {
+    ...fontStyles.regular,
+  }
+})
+
+SwapPending.navOptions = {
+  ...noHeader,
+}
+
+export default SwapPending
--- a/src/swap/SwapReviewScreen.tsx
+++ b/src/swap/SwapReviewScreen.tsx
@@ -30,6 +30,7 @@ import colors from 'src/styles/colors'
 import fontStyles from 'src/styles/fonts'
 import { Spacing } from 'src/styles/styles'
 import variables from 'src/styles/variables'
+import { swapStart } from 'src/swap/slice'
 import { Field } from 'src/swap/useSwapQuote'
 import { coreTokensSelector } from 'src/tokens/selectors'
 import { divideByWei, multiplyByWei } from 'src/utils/formatting'
@@ -109,7 +110,7 @@ export function SwapReviewScreen(props: Props) {
           `Failure response fetching token swap quote. ${response.status}  ${response.statusText}`
         )
       }
-      setSwapInfo(await response.json())
+      setSwapInfo({ ...(await response.json()), userInput: params })
       setShouldFetch(false)
       setFetchError(false)
     },
@@ -127,6 +128,7 @@ export function SwapReviewScreen(props: Props) {
   const submitSwap = () => {
     // Check for swapInfo prior to submitting swap
     if (!swapInfo) return
+
     // Analytics for swap submission
     ValoraAnalytics.track(SwapEvents.swap_review_submit, {
       toToken,
@@ -140,6 +142,7 @@ export function SwapReviewScreen(props: Props) {
       ),
     })
     // TODO: dispatch swap submission
+    dispatch(swapStart(swapInfo))
   }
 
   return (
@@ -370,7 +373,7 @@ const styles = StyleSheet.create({
     flex: 1,
     alignSelf: 'flex-end',
     flexDirection: 'row',
-    justifyContent: 'space-between',
+    justifyContent: 'flex-start',
     alignItems: 'center',
   },
   transactionDetailsLabel: {
--- a/src/swap/saga.ts
+++ b/src/swap/saga.ts
@@ -1,0 +1,71 @@
+import { ContractKit } from '@celo/contractkit'
+import { call, takeEvery } from 'redux-saga/effects'
+import { walletToAccountAddressSelector } from 'src/identity/selectors'
+import { navigate } from 'src/navigator/NavigationService'
+import { Screens } from 'src/navigator/Screens'
+import { swapStart } from 'src/swap/slice'
+import { fetchWithTimeout } from 'src/utils/fetchWithTimeout'
+import Logger from 'src/utils/Logger'
+import { getContractKit } from 'src/web3/contracts'
+import networkConfig from 'src/web3/networkConfig'
+import { getConnectedUnlockedAccount } from 'src/web3/saga'
+import { TransactionConfig } from 'web3-core'
+
+const TAG = 'swap/saga'
+
+// TODO Tomm: create a real type for this...
+// TODO Tomm: handle state changes - Approve Funds - Complete Swap - etc
+function* SwapSaga(data: any) {
+  try {
+    // Navigate to completeSwap screen
+    yield call(navigate, Screens.SwapPending)
+
+    // Get contract kit, wallet address and
+    const kit: ContractKit = yield call(getContractKit)
+    const walletAddress: string = yield call(getConnectedUnlockedAccount)
+
+    // Call approveTransaction
+    Logger.debug(TAG, `Starting to swap approval for address: ${walletAddress}`)
+    kit.web3.eth.sendTransaction({
+      ...data.payload.approveTransaction,
+      from: walletAddress,
+    } as TransactionConfig)
+
+    // TODO Tomm: Check the if the results of the approve transaction and within acceptable ranges and if not cancel
+    // console.log('estimated price impact', data.payload.unvalidatedSwapTransaction.estimatedPriceImpact)
+
+    // Query the execute swap endpoint
+    // TODO Tomm: debug why buyAmount works and sellAmount does not
+    const amountType: string = data.payload.userInput.buyAmount ? 'buyAmount' : 'sellAmount'
+    const amount = data.payload.unvalidatedSwapTransaction[amountType]
+    // console.log('Tom - params - sellAmount, buyAmount',
+    //   data.payload.unvalidatedSwapTransaction.sellAmount,
+    //   data.payload.unvalidatedSwapTransaction.buyAmount
+    // )
+    const params = {
+      buyToken: data.payload.unvalidatedSwapTransaction.buyTokenAddress,
+      sellToken: data.payload.unvalidatedSwapTransaction.sellTokenAddress,
+      [amountType]: amount,
+      userAddress: walletAddress,
+    }
+
+    console.log('Tom - params', params)
+    const queryParams = new URLSearchParams({ ...params }).toString()
+    const requestUrl = `${networkConfig.executeSwapUrl}?${queryParams}`
+    const response: Response = yield call(fetchWithTimeout, requestUrl)
+    const responseJson: any = yield call([response, response.json])
+
+    // Send the transaction to execute the swap
+    kit.web3.eth.sendTransaction({
+      ...responseJson.validatedSwapTransaction,
+      from: walletToAccountAddressSelector,
+    } as TransactionConfig)
+  } catch (error) {
+    Logger.error(TAG, 'Error while swapping', error)
+  }
+}
+
+export function* swapSaga() {
+  Logger.debug(TAG, 'Initializing swap sagas')
+  yield takeEvery(swapStart.type, SwapSaga)
+}
--- a/src/swap/slice.ts
+++ b/src/swap/slice.ts
@@ -1,0 +1,35 @@
+import { createSlice } from '@reduxjs/toolkit'
+
+export interface State {
+  swapLoading: boolean
+  swapError: boolean
+  swapInfo: {} | null
+}
+
+const initialState: State = {
+  swapLoading: false,
+  swapError: false,
+  swapInfo: null,
+}
+
+export const slice = createSlice({
+  name: 'swap',
+  initialState,
+  reducers: {
+    swapStart: (state, payload) => {
+      state.swapLoading = true
+      state.swapInfo = payload
+    },
+    swapError: (state) => {
+      state.swapLoading = false
+      state.swapError = true
+    },
+    swapSuccess: (state) => {
+      ;(state.swapLoading = false), (state.swapError = false)
+    },
+  },
+})
+
+export const { swapStart, swapError, swapSuccess } = slice.actions
+
+export default slice.reducer
--- a/src/navigator/Navigator.tsx
+++ b/src/navigator/Navigator.tsx
@@ -106,6 +106,7 @@ import ValidateRecipientIntro, {
   validateRecipientIntroScreenNavOptions,
 } from 'src/send/ValidateRecipientIntro'
 import SetClock from 'src/set-clock/SetClock'
+import SwapPending from 'src/swap/SwapPending'
 import SwapReviewScreen from 'src/swap/SwapReviewScreen'
 import SwapScreen from 'src/swap/SwapScreen'
 import TokenBalancesScreen from 'src/tokens/TokenBalances'
@@ -541,6 +542,11 @@ const swapScreens = (Navigator: typeof Stack) => (
       name={Screens.SwapReviewScreen}
       component={SwapReviewScreen}
       options={SwapReviewScreen.navOptions}
+    />
+    <Navigator.Screen
+      name={Screens.SwapPending}
+      component={SwapPending}
+      options={SwapPending.navOptions}
     />
   </>
 )
--- a/src/navigator/Screens.tsx
+++ b/src/navigator/Screens.tsx
@@ -78,6 +78,7 @@ export enum Screens {
   Support = 'Support',
   SupportContact = 'SupportContact',
   SwapScreen = 'SwapScreen',
+  SwapPending = 'SwapPendingScreen',
   SwapReviewScreen = 'SwapReviewScreen',
   Sync = 'Sync',
   TokenBalances = 'TokenBalances',
--- a/src/navigator/types.tsx
+++ b/src/navigator/types.tsx
@@ -276,6 +276,7 @@ export type StackParamList = {
     | undefined
   [Screens.Sync]: undefined
   [Screens.SwapScreen]: undefined
+  [Screens.SwapPending]: undefined
   [Screens.SwapReviewScreen]: {
     toToken: string
     fromToken: string
--- a/src/redux/sagas.ts
+++ b/src/redux/sagas.ts
@@ -39,6 +39,7 @@ import { waitForRehydrate } from 'src/redux/persist-helper'
 import { sendSaga } from 'src/send/saga'
 import { sentrySaga } from 'src/sentry/saga'
 import { stableTokenSaga } from 'src/stableToken/saga'
+import { swapSaga } from 'src/swap/saga'
 import { tokensSaga } from 'src/tokens/saga'
 import { Actions as TransactionActions } from 'src/transactions/actions'
 import { transactionSaga } from 'src/transactions/saga'
@@ -136,6 +137,7 @@ export function* rootSaga() {
     yield spawn(checkAndroidMobileServicesSaga)
     yield spawn(dappsSaga)
     yield spawn(fiatConnectSaga)
+    yield spawn(swapSaga)
   } catch (error) {
     Logger.error('@rootSaga', 'Error while initializing sagas', error)
     // Propagate so it's handled by Sentry


locales/base/translation.json
```diff
@@ -1405,6 +1405,26 @@
...
+    "swapPending": "Swap Pending",
+    "swapSuccess": "Swap Successful!",
+    "exchangeRateSubtext": "Finding the best exchange rate...",
+    "approvingSubtext": "Approving funds...",
+    "completingSubtext": "Completing swap...",
+    "swapAgain": "Swap Again",
+    "swapPriceModal": {
+      "title": "Prices have changed",
+      "action": "Continue",
+      "body1": "Crypto amounts & fees can change due to due to fluctuations in the crypto market.",
```
kathaypacific, line 1418: instead of body1 and body2, can we use `\n` inside 1 string? there are a few examples of this in the translation file. (we should avoid putting non descriptive translation keys like key1 and key2, as the names are useful for translators when providing translations, since the the key name is the only context they have for where it's used. the more descriptive the name, the better quality translations we will get)

src/swap/SwapReviewScreen.tsx
```diff
@@ -23,38 +22,33 @@ import TokenDisplay, { formatValueToDisplay } from 'src/components/TokenDisplay'
...
-  unvalidatedSwapTransaction: {
-    sellToken: string
-    buyToken: string
-    buyAmount: string
-    sellAmount: string
-    price: string
-    gas: string
-    gasPrice: string
+export function SwapReviewScreen() {
+  const initialUserInput = {
```
kathaypacific, line 41: i think this can be moved outside of the component to prevent it from being regenerated on every component rerender :)

src/swap/SwapReviewScreen.tsx
```diff
@@ -109,7 +102,7 @@ export function SwapReviewScreen(props: Props) {
           `Failure response fetching token swap quote. ${response.status}  ${response.statusText}`
         )
       }
-      setSwapInfo(await response.json())
+      setSwapInfo({ ...(await response.json()), userInput: params })
```
kathaypacific, line 105: why add params to the response? wouldn't this lead to inconsistencies (e.g. some parts of the response don't match the params and now we have `swapInfo` as a mixture of the 2)? i think we should find some way to avoid this, as this can lead to subtle and hard to debug bugs...

src/swap/slice.ts
```diff
@@ -0,0 +1,84 @@
...
+    swapReset: (state) => {
+      state.swapState = SwapState.USER_INPUT
+      state.swapInfo = null
+    },
+    swapPriceChange: (state) => {
+      state.swapState = SwapState.PRICE_CHANGE
+    },
+    swapError: (state) => {
+      state.swapState = SwapState.ERROR
+      state.swapInfo = null
```
kathaypacific, line 59: does this still work for this scenario? https://www.figma.com/file/jRtYI8GiSvPAaClBnRLLpu/In-App-Swap?node-id=1526%3A17286 seems like here, we still want to retain some swap info if there is an error?

src/swap/slice.ts
```diff
@@ -0,0 +1,84 @@
...
+    },
+  },
+  extraReducers: (builder) => {
+    builder.addCase(REHYDRATE, (state, action: RehydrateAction) => ({
```
kathaypacific, line 63: do we need this? i think it's okay to always start with a clean slate between app sessions? we could probably even not persist the swap state at all in redux....i feel like it might reduce the chance of weird bugs if we do this, since if the swap get stuck in a weird state, restarting the app would fix it if we don't persist the old state.

src/swap/SwapPending.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+  // Prevent back button on Android
+  useEffect(() => {
+    const backPressListener = () => true
+    BackHandler.addEventListener('hardwareBackPress', backPressListener)
+    return () => BackHandler.removeEventListener('hardwareBackPress', backPressListener)
+  }, [])
+
+  return (
+    <SafeAreaView style={{ flex: 1 }}>
```
kathaypacific, line 111: move this style to `styles`? :)

src/swap/SwapPending.tsx
```diff
@@ -0,0 +1,154 @@
...
+      <View style={styles.contentContainer}>
+        {![SwapState.COMPLETE, SwapState.USER_INPUT].includes(swapState) ? (
```
kathaypacific, line 113: this is a super nit but i think it's less cognitive load to have positive conditions in ternaries? like `[...].includes(...) ? positive outcome : negative outcome` seems more straightforward than `![...].includes(...) ? negative outcome : positive outcome`

src/swap/SwapPending.tsx
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  const navigateToReviewScreen = () => {
+    navigate(Screens.SwapReviewScreen)
+  }
+
+  const navigateToSupport = () => {
+    dispatch(swapReset())
+    navigate(Screens.SwapScreen)
+    navigate(Screens.SupportContact)
```
kathaypacific, line 35: ah, i forgot i was going to help look into this double navigate...i'll put it on my list

src/swap/SwapScreen.tsx
```diff
@@ -100,22 +115,28 @@ export function SwapScreen() {
...
   }
 
   const handleShowTokenSelect = (fieldType: Field) => () => {
     ValoraAnalytics.track(SwapEvents.swap_screen_select_token, { fieldType })
-    Keyboard.dismiss()
     // ensure that the keyboard is dismissed before animating token bottom sheet
-    setTimeout(() => {
-      setSelectingToken(fieldType)
-    }, 100)
+    Keyboard.addListener('keyboardDidHide', onKeyboardDidHide(fieldType))
```
kathaypacific, line 133: i'm confused, we are removing the keyboard listener inside the `addListener`? why do we need this?

src/swap/saga.ts
```diff
@@ -0,0 +1,109 @@
...
+import { fetchWithTimeout } from 'src/utils/fetchWithTimeout'
+import Logger from 'src/utils/Logger'
+import { getContractKit } from 'src/web3/contracts'
+import networkConfig from 'src/web3/networkConfig'
+import { getConnectedUnlockedAccount } from 'src/web3/saga'
+import { applyChainIdWorkaround, buildTxo } from 'src/web3/utils'
+
+const TAG = 'swap/saga'
+
+function getPriceDiff(price1: number, price2: number) {
```
kathaypacific, line 26: can we make this function name more descriptive? i don't really understand why we are dividing by 2, but otherwise it is the price difference percentage right?

src/swap/saga.ts
```diff
@@ -0,0 +1,109 @@
...
+    yield put(swapSuccess())
+  } catch (error) {
+    Logger.error(TAG, 'Error while swapping', error)
+    yield put(swapError())
+  }
+}
+
+export function* swapSaga() {
+  Logger.debug(TAG, 'Initializing swap sagas')
+  yield takeEvery(swapStart.type, swapSubmitSaga)
```
kathaypacific, line 108: `yield takeLatest(swapStart.type, swapStartSaga)`? :) i think that `takeLatest` would work better here, we don't need to kick off multiple of these

src/swap/saga.ts
```diff
@@ -0,0 +1,109 @@
...
+    // Navigate to swap pending screen
+    yield call(navigate, Screens.SwapPending)
+
+    // Check that prices are within 2% of each other
+    const priceDiff: number = yield call(
+      getPriceDiff,
+      +data.payload.unvalidatedSwapTransaction.price,
+      +data.payload.unvalidatedSwapTransaction.guaranteedPrice
+    )
+    if (priceDiff >= 2) {
```
kathaypacific, line 41: oh i understand the 2 is the slippage tolerance? can this be a config variable?

kathaypacific (commented):
hey this is looking great!!

i faced a similar scenario with the CPV stuff last week, so maybe it's worth discussing what we want to keep in redux and what we don't - i think the swaps slice adds value by providing the swap info through the whole flow, so that we don't need to pass stuff between screen with. navigation params. and obviously the current implementation with the saga works. but i do we think that there are transient states that maybe don't need to be in redux because they're specific for the swap pending screen? i'm looking at approve / execute / complete / price-change. these states need to exist if we keep this as a saga, but we could also make this saga into a hook that is triggered by the swap pending screen. 

my conversation with @jacob-israel-turner about overusing redux has been on my mind 😅 what do you guys think @MuckT @jeanregisser ?

src/swap/SwapScreen.tsx
MuckT: This is to ensure the setSelectingToken is called after the keyboard has been fully removed. There's probably a better way but using setTimeout with a value of 100 can result in the swap review button appearing in the wrong place on the screen on Android devices.

src/swap/slice.ts
MuckT: That error is handled by the swapScreen itself and isn't impacted by this swapError in Redux.

src/swap/SwapScreen.tsx
MuckT: On Android with the setTimeout code, complete a swap and then tap 'Swap Again' Then observe that the 'Review' button will not be aligned with the bottom of the page.

package.json
```diff
@@ -94,7 +94,7 @@
     "@react-native-masked-view/masked-view": "^0.2.6",
     "@react-native-picker/picker": "^2.4.2",
     "@react-navigation/compat": "^5.3.20",
-    "@react-navigation/core": "^6.2.1",
+    "@react-navigation/core": "^5.16.1",
```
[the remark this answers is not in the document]
MuckT: Removed in https://github.com/valora-inc/wallet/pull/2891/commits/36adccf7d024385c0ee66941339a76b37a50a402! I'll put in it's own PR.

src/swap/SwapReviewScreen.tsx
MuckT: Moved in https://github.com/valora-inc/wallet/pull/2891/commits/da2c20e1a0837bdfd87830b0ca204ba6857b182d!

src/swap/saga.ts
MuckT: Added a comment and used the remote config value in https://github.com/valora-inc/wallet/pull/2891/commits/debfbea4e50ffa7eb1b7ecc0be6b08e84fba5bea!

src/swap/saga.ts
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2891/commits/5c99dfc5b28195ce5357e72cf7c1c1399d09f383!

src/swap/slice.ts
MuckT: Removed in https://github.com/valora-inc/wallet/pull/2891/commits/a38823b18fe03b97e94f48f706bb82e19999b254!

src/swap/SwapPending.tsx
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2891/commits/00b9f8d7b0e8b57b7de7f174c1899dcc64b57b1d!

src/swap/SwapReviewScreen.tsx
MuckT: Is pulling the userInput into the saga via selector preferred?

locales/base/translation.json
MuckT: Fixed in ed092f282cd6e383afffbc23773b561187406949!

src/swap/SwapScreen.tsx
MuckT: There is probably a better way to do this; I'll see what I can come up with!

src/swap/saga.ts
MuckT: Fixed in debfbea4e50ffa7eb1b7ecc0be6b08e84fba5bea!

src/swap/saga.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    // Set contract kit, wallet address and normalizer
+    const kit: ContractKit = yield call(getContractKit)
+    const walletAddress: string = yield call(getConnectedUnlockedAccount)
+    const normalizer = new TxParamsNormalizer(kit.connection)
+
+    // Approve transaction
+    yield put(swapApprove())
+    Logger.debug(TAG, `Starting to swap approval for address: ${walletAddress}`)
+    const rawApproveTx = { ...action.payload.approveTransaction, from: walletAddress }
```
kathaypacific, line 58: is there any reason why the `from` value would not be the `walletAddress` from the `approveTransaction`?

src/swap/saga.ts
```diff
@@ -0,0 +1,116 @@
...
+    applyChainIdWorkaround(rawApproveTx, yield call([kit.connection, 'chainId']))
+    const approveTx: CeloTx = yield call(normalizer.populate.bind(normalizer), rawApproveTx)
+    const approveTxo = buildTxo(kit, approveTx)
+    yield call(
+      sendTransaction,
+      approveTxo,
+      walletAddress,
+      newTransactionContext(TAG, 'Swap/Approve')
+    )
```
kathaypacific, line 67: what do you think about abstracting this block, since we use it later below? i was thinking something like `yield call(sendSwapTransaction, action.payload.approveTransaction, newTransactionContext(TAG, 'Swap/Approve')` would do the trick? the benefit i see with this is being able to move all the contractkit logic away from the swap logic and making this saga more focused on the swap details. also not seeing `applyChainIdWorkaround` twice is nice haha

src/swap/saga.ts
kathaypacific: `getPercentageDifference`?

src/swap/saga.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    yield put(swapSuccess())
+  } catch (error) {
+    Logger.error(TAG, 'Error while swapping', error)
+    yield put(swapError())
+  }
+}
+
+export function* swapSaga() {
+  Logger.debug(TAG, 'Initializing swap sagas')
```
kathaypacific, line 114: do we need this debug line?

yarn.lock
```diff
@@ -4346,15 +4346,16 @@
     react-is "^16.13.0"
 
 "@react-navigation/core@^6.2.1":
-  version "6.2.1"
-  resolved "https://registry.yarnpkg.com/@react-navigation/core/-/core-6.2.1.tgz#90459f9afd25b71a9471b0706ebea2cdd2534fc4"
-  integrity sha512-3mjS6ujwGnPA/BC11DN9c2c42gFld6B6dQBgDedxP2djceXESpY2kVTTwISDHuqFnF7WjvRjsrDu3cKBX+JosA==
+  version "6.4.0"
```
kathaypacific, line 4349: i don't think these changes to the yarn.lock are intended?

src/swap/SwapReviewScreen.tsx
```diff
@@ -109,7 +104,8 @@ export function SwapReviewScreen(props: Props) {
           `Failure response fetching token swap quote. ${response.status}  ${response.statusText}`
         )
       }
-      setSwapInfo(await response.json())
+      const json: SwapInfo = await response.json()
```
kathaypacific, line 107: i think this type is not right? `SwapInfo` includes a `userInput` field which will not be returned from the response right?

this screen seems only to need the `unvalidatedSwapTransaction` property from the `swapInfo` state variable, so should the type of this state variable be `SwapTransaction` instead?

src/swap/saga.ts
```diff
@@ -0,0 +1,116 @@
...
+import { getConnectedUnlockedAccount } from 'src/web3/saga'
+import { applyChainIdWorkaround, buildTxo } from 'src/web3/utils'
+
+const TAG = 'swap/saga'
+
+function getPriceDiff(price1: number, price2: number) {
+  return (Math.abs(price1 - price2) / ((price1 + price2) / 2)) * 100
+}
+
+export function* swapSubmitSaga(action: PayloadAction<SwapInfo>) {
```
kathaypacific, line 33: i am wondering if there is value in this `SwapInfo` object, since it doesn't occur naturally and no component / function seems to require all the properties at once? for example, the payload for this action could be something like `{approveTransaction, unvalidatedSwapTransaction, updatedFieldType}`. the benefit would be that you can make every one of these properties  required. currently in `SwapInfo`, the approveTransaction property is optional which is actually not expected for this saga or else line 58 will not have any tx object to approve - typescript didn't help to catch this in this case though...

kathaypacific, line 33: did we consider making this saga a hook? the benefit would be that the transient states (like the swap is being started / approved / executed / priced changed) would not need to be in redux since they can directly be displayed by the only component consuming this data. i don't have a strong opinion though, what is here works. just wanted to throw the idea out there, since we've been talking about our redux usage recently

src/swap/saga.test.ts
```diff
@@ -0,0 +1,120 @@
...
+      .provide([
+        [call(getContractKit), contractKit],
+        [call(getConnectedUnlockedAccount), mockAccount],
+        [call(fetchWithTimeout, executeSwapUri), mockResponse],
+      ])
+      .put(swapApprove())
+      .put(swapExecute())
+      .run()
+    expect(sendTransaction).toHaveBeenCalledTimes(2)
+    expect(loggerErrorSpy).not.toHaveBeenCalled()
```
kathaypacific, line 79: i think to verify that the swap is completed successfully, we should also assert that the `sendTransaction` was called with the expected inputs right?

kathaypacific (commented):
Looking great!! mostly small comments from me. would be nice to get @jeanregisser 's feedback on this too, the PR is a bit big so would appreciate an extra pair of eyes 😅

src/swap/saga.ts
MuckT: Does it makes sense to change the valora-rest-api to return a swapTransaction for both the approve and execute endpoints instead of unvalidatedSwapTransaction  or validatedSwapTransaction these appear to have all the same fields.

MuckT: I've opened https://github.com/valora-inc/valora-rest-api/pull/162 to change the api behavior to above.

src/swap/saga.ts
MuckT: I tried this in 607125137e11d2a0cbed357d7ddf1dfade6164e6, but ran into issues with the swaps executing. Maybe something to do with passing around the kit and normalizer?

src/redux/migrations.test.ts
```diff
@@ -674,4 +675,17 @@ describe('Redux persist migrations', () => {
     // shall be the default value as configured in REMOTE_CONFIG_VALUES_DEFAULTS
     expect(migratedSchema).toStrictEqual(expectedSchema)
   })
+
+  it('works from v80 to v81', () => {
+    const oldSchema = v80Schema
+    const migratedSchema = migrations[81](oldSchema)
+
+    const expectedSchema: any = _.cloneDeep(oldSchema)
+    expectedSchema.swap = {}
```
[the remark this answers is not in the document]
MuckT: Fixed in 56c25eda8967ee478288975d5c375fb4f5cb50b3!

src/swap/SwapPending.tsx
```diff
@@ -0,0 +1,203 @@
...
+  useEffect(() => {
+    const backPressListener = () => true
+    BackHandler.addEventListener('hardwareBackPress', backPressListener)
+    return () => BackHandler.removeEventListener('hardwareBackPress', backPressListener)
+  }, [])
+
+  return (
+    <SafeAreaView style={styles.safeAreaView}>
+      <View style={styles.contentContainer}>
+        {swapIcon}
```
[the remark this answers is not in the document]
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/2891/commits/34f1b116707e777fadf99b31a1854432be819d25!

src/swap/SwapReviewScreen.tsx
MuckT: Currently it looks like this:
```TypeScript
const json: SwapInfo = await response.json()
setSwapInfo(json)
...
// Dispatch swap submission
if (userInput !== null) {
  dispatch(swapStart({ ...swapInfo, userInput: userInput }))
}
```

src/swap/SwapPending.tsx
MuckT: Fixed in 7a90ce7e755f19a1458cfd2fc2720afbd084d199!

src/swap/saga.ts
MuckT: Removed in https://github.com/valora-inc/wallet/pull/2891/commits/0b30b8367da3f7c844ecef1e6d425de340b025ef!

src/swap/SwapReviewScreen.tsx
MuckT: I also need the `approveTransaction`. Should there be a new type with both included or should swapInfo be changed to omit `validatedSwapTransaction`?

src/swap/saga.ts
MuckT: Removed!

yarn.lock
MuckT: Something strange is going on with this yarn.lock, if I tried to manually edit the check yarn.lock check is failing.

src/swap/saga.ts
MuckT: Ran into some issues with the chain Id. Can attempt in a follow up pr.

locales/base/translation.json
```diff
@@ -1430,6 +1430,26 @@
...
+    "swapSuccess": "Swap Successful!",
+    "swapError": "Swap Error",
+    "exchangeRateSubtext": "Finding the best exchange rate...",
+    "approvingSubtext": "Approving funds...",
+    "completingSubtext": "Completing swap...",
+    "swapAgain": "Swap Again",
+    "swapPriceModal": {
+      "title": "Prices have changed",
+      "action": "Continue",
+      "body": "Crypto amounts & fees can change due to due to fluctuations in the crypto market. \n\n Please review your final amount & fees and resubmit."
```
kathaypacific, line 1444: `due to` is repeated twice here

src/swap/SwapReviewScreen.tsx
```diff
@@ -119,34 +115,38 @@ export function SwapReviewScreen(props: Props) {
...
       ),
-      fee: +divideByWei(swapInfo.unvalidatedSwapTransaction.sellAmount).multipliedBy(
+      fee: +divideByWei(swapResponse.unvalidatedSwapTransaction.sellAmount).multipliedBy(
         swapFeeDecimal
       ),
     })
-    // TODO: dispatch swap submission
+    // Dispatch swap submission
+    if (userInput !== null) {
+      dispatch(swapStart({ ...swapResponse, userInput: userInput }))
```
kathaypacific, line 141:
```suggestion
      dispatch(swapStart({ ...swapResponse, userInput }))
```

src/swap/saga.ts
```diff
@@ -0,0 +1,114 @@
...
+      buyToken: action.payload.unvalidatedSwapTransaction.buyTokenAddress,
+      sellToken: action.payload.unvalidatedSwapTransaction.sellTokenAddress,
+      [amountType]: amount,
+      userAddress: walletAddress,
+    }
+    const queryParams = new URLSearchParams({ ...params }).toString()
+    const requestUrl = `${networkConfig.executeSwapUrl}?${queryParams}`
+    const response: Response = yield call(fetchWithTimeout, requestUrl)
+    if (!response.ok) {
+      Logger.error(TAG, `Swap failed with status: ${response.status}`)
```
kathaypacific, line 83: let's not alert sentry about failed swaps...

kathaypacific: also, instead of handling the error here, why not just throw and let the catch on line 95 handle it?

src/swap/saga.ts
kathaypacific: i was playing around with this and it seemed to work for me - strangely i didn't really change anything from what we tried yesterday. since i already have it i'll submit the PR for you to have a look

src/swap/saga.ts
```diff
@@ -0,0 +1,114 @@
...
+    const executeTxo = buildTxo(kit, executeTx)
+    yield call(
+      sendTransaction,
+      executeTxo,
+      walletAddress,
+      newTransactionContext(TAG, 'Swap/Execute')
+    )
+    yield put(swapSuccess())
+  } catch (error) {
+    Logger.error(TAG, 'Error while swapping', error)
```
kathaypacific, line 107: would also suggest `Logger.debug` here

src/swap/SwapPending.tsx
```diff
@@ -0,0 +1,203 @@
...
+import { noHeader } from 'src/navigator/Headers'
+import { navigate } from 'src/navigator/NavigationService'
+import { Screens } from 'src/navigator/Screens'
+import colors from 'src/styles/colors'
+import fontStyles from 'src/styles/fonts'
+import { Spacing } from 'src/styles/styles'
+import { swapStateSelector } from 'src/swap/selectors'
+import { SwapState } from 'src/swap/slice'
+
+export function SwapPending() {
```
kathaypacific, line 19: `SwapPendingScreen`? `SwapExecuteScreen`?

src/swap/SwapPending.tsx
kathaypacific: i'm thinking that this issue isn't worth spending too much time on, but i think navigating twice is quite hacky (for lower end devices like mine, this double navigation can actually be seen). can i suggest for MVP that we just use the `replace` function in the navigation service so that when you navigate back from support, you land on the swap review screen? that seems reasonable, to see that swap that you've contacted support for?

src/swap/SwapPending.tsx
```diff
@@ -0,0 +1,203 @@
...
+    alignItems: 'center',
+    justifyContent: 'center',
+    padding: Spacing.Regular16,
+    flexGrow: 1,
+  },
+  activityIndicator: {
+    left: 0,
+    right: 0,
+    top: 0,
+    bottom: 0,
```
kathaypacific, line 175: do we need this left/right/top/bottom 0 style when the position is not absolute? seems to look fine without these styles for me

yarn.lock
kathaypacific: can you just revert your changes for the yarn.lock? you didn't change any dependencies so there's no reason why the yarn.lock file should change i think. maybe something is going on with your local yarn cache?

src/swap/SwapReviewScreen.test.tsx
```diff
@@ -129,4 +132,37 @@ describe('SwapReviewScreen', () => {
...
+    )
+
+    expect(ValoraAnalytics.track).toHaveBeenCalledWith(SwapEvents.swap_review_screen_open, {
+      buyAmount: mockBuyAmount,
+      fromToken: mockCeloAddress,
+      toToken: mockCusdAddress,
+    })
+  })
+
+  it.todo('should correctly dispatch swapStart')
```
kathaypacific, line 166: todo?

kathaypacific (commented):
this is looking great Tom!!! i just tried to go through one last time with a fine tooth comb, i think we're super super close :D

locales/base/translation.json
MuckT: Fixed in d9faae54f6f38ae7deda02b0e2395d23a2f96791!

src/swap/SwapReviewScreen.tsx
```diff
@@ -119,34 +115,38 @@ export function SwapReviewScreen(props: Props) {
...
       toToken,
       fromToken,
       // TODO: Add fee to total when enabled
-      usdTotal: +divideByWei(swapInfo.unvalidatedSwapTransaction.buyAmount).multipliedBy(
-        swapInfo.unvalidatedSwapTransaction.price
+      usdTotal: +divideByWei(swapResponse.unvalidatedSwapTransaction.buyAmount).multipliedBy(
+        swapResponse.unvalidatedSwapTransaction.price
       ),
-      fee: +divideByWei(swapInfo.unvalidatedSwapTransaction.sellAmount).multipliedBy(
+      fee: +divideByWei(swapResponse.unvalidatedSwapTransaction.sellAmount).multipliedBy(
```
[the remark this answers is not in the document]
MuckT: In my opinion we should still include the fee's in analytics even if we aren't enabling them, this would give us an idea on what the total fees on all attempted swaps would be.

src/swap/saga.ts
MuckT: Fixed in 8fb182bf7fb0da7823c739f28480e7f2cc7947dd!

src/swap/saga.ts
MuckT: Changed in 60fead26fafbb504e9d50c5f906545c71086b9d7!

src/swap/SwapPending.tsx
MuckT: Changed in ede3eced2714a40c7a6c9796bbef588523a06739!

src/swap/SwapReviewScreen.tsx
```diff
@@ -119,34 +115,38 @@ export function SwapReviewScreen(props: Props) {
...
+        Logger.debug(TAG, 'Error while fetching swap quote', error)
       },
     }
   )
 
   const submitSwap = () => {
-    // Check for swapInfo prior to submitting swap
-    if (!swapInfo) return
+    // Check for swapResponse prior to submitting swap
+    if (!swapResponse) return
```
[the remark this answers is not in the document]
MuckT: Added an error banner and logged the error in ce8b2d30f9d9554b3ddbbd74164198037b07d848! This shouldn't happen so sending to sentry is appropriate.

src/swap/SwapReviewScreen.tsx
```diff
@@ -119,34 +115,38 @@ export function SwapReviewScreen(props: Props) {
...
+      dispatch(swapStart({ ...swapResponse, userInput: userInput }))
+    }
   }
 
   return (
     <SafeAreaView style={styles.safeAreaContainer}>
       <CustomHeader title={t('swapReviewScreen.title')} left={<BackButton />} />
       <DisconnectBanner />
-      {shouldFetch && swapInfo === null ? (
+      {shouldFetch && swapResponse === null ? (
```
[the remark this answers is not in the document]
MuckT: Removed in fda69e2157515756276a856a3726634689a54cb4!

src/swap/SwapPending.tsx
MuckT: Fixed in cbf62316a150a46b33306cb983a3bd75c031b07b!

src/swap/SwapScreen.tsx
MuckT: Wasn't able to find a better way but as we are adding and quickly removing the listener this seems fine after talking with @jeanregisser.

src/swap/SwapReviewScreen.test.tsx
MuckT: Added in a3e5f498c346061c52a24d7ec3eef1758809aeed!

src/swap/SwapReviewScreen.tsx
```diff
@@ -23,58 +22,60 @@ import TokenDisplay, { formatValueToDisplay } from 'src/components/TokenDisplay'
...
-
   // Remote configs converted to decimals strings
   // const maxSlippageDecimal = `${maxSlippagePercent / 100}`
   const swapFeeDecimal = `${swapFeePercentage / 100}`
 
   const { t } = useTranslation()
   const dispatch = useDispatch()
 
+  // Workaround for buying Celo
+  const CELO_ADDRESS = '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE'
```
kathaypacific, line 73: did you mean to add this workaround to this PR? 😅

src/swap/SwapReviewScreen.tsx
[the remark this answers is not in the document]
kathaypacific: currently this "fee" value is based on some percentage that we have on remote config which may not be what we want to use in the end. isn't it more realistic that if we want to calculate the potential revenue from fees, that we calculate all the swap amounts instead? i'm not sure why we would use this value instead, so i still don't really see the value of keeping this property

src/swap/SwapReviewScreen.tsx
```diff
@@ -119,34 +122,42 @@ export function SwapReviewScreen(props: Props) {
...
+      dispatch(swapStart({ ...swapResponse, userInput }))
+    }
   }
 
   return (
     <SafeAreaView style={styles.safeAreaContainer}>
       <CustomHeader title={t('swapReviewScreen.title')} left={<BackButton />} />
       <DisconnectBanner />
-      {shouldFetch && swapInfo === null ? (
+      {shouldFetch ? (
```
kathaypacific, line 160: there is an activity indicator on line 183 already, which is also displayed if `shouldFetch` is true? why do we need both?

src/swap/SwapReviewScreen.tsx
```diff
@@ -69,33 +77,28 @@ export function SwapReviewScreen() {
...
+  // BuyAmount or SellAmount
+  const swapAmountParam = updatedField === Field.FROM ? 'sellAmount' : 'buyAmount'
+  const swapAmountInWei = multiplyByWei(swapAmount[updatedField]!)
+
   useEffect(() => {
     ValoraAnalytics.track(SwapEvents.swap_review_screen_open, {
       toToken,
       fromToken,
-      buyAmount: swapAmount[Field.TO] ?? '',
+      amount: swapAmount[updatedField] ?? '',
```
kathaypacific, line 92: why do we need this change? with this change, sometimes the `amount` will be a buy amount and sometimes a `sell` amount so swapping twice with the same value like 1 CELO : 5 cUSD, depending on which field you changed last the `amount` could be 1 or 5...?

src/swap/SwapReviewScreen.tsx
```diff
@@ -139,13 +142,10 @@ export function SwapReviewScreen() {
...
       fromToken,
-      // TODO: Add fee to total when enabled
-      usdTotal: +divideByWei(swapResponse.unvalidatedSwapTransaction.buyAmount).multipliedBy(
+      usdTotal: +divideByWei(swapResponse.unvalidatedSwapTransaction[swapAmountParam]).multipliedBy(
         swapResponse.unvalidatedSwapTransaction.price
       ),
-      fee: +divideByWei(swapResponse.unvalidatedSwapTransaction.sellAmount).multipliedBy(
-        swapFeeDecimal
-      ),
+      amount: swapAmount[updatedField] ?? '',
```
kathaypacific, line 148: i also think we should only do a buy OR sell amount here, because both will be available at this point. or if we really want, we can do buy AND sell. but avoiding a property that can be both without being able to know which it is, would be nice

kathaypacific (commented):
Had mostly comments about analytics :) 

An unrelated thought, should we reverse the CELO workaround for analytics? i'm not sure if the 0x response for buy/sellTokenAddress is the dolphin one?

kathaypacific (approved):
🚀

PR #2894, opened 2022-09-21
feat: add central phone verification screen UI

kathaypacific:
### Description

This PR:
- consume the CPV remote config variable
- add CPV screens UI - haven't linked this to any backend yet

I tried to optimise for ease of removal of the decentralised flow, so keeping the 2 flows as separate as possible including translations and not creating too many new things that need to be updated later. There are a lot of places that try to navigate to the phone verification screen, rather than update these consumers i opted to check which flow should be rendered inside the screen.

(I've got larger text enabled on my device)

https://user-images.githubusercontent.com/20150449/191745320-82a96ebb-4206-4ef4-9fbe-06ad74ef5e3d.mp4

Note that there's heeeeaps of things that this PR is missing, like tests / analytics / error handling / a backend. Those things are coming next.

### Other changes

N/A

### Tested

Manually

### How others should test

N/A

### Related issues

Relates RET-384

### Backwards compatibility

Yes


src/verify/PhoneVerificationInputScreen.tsx
```diff
@@ -0,0 +1,135 @@
+import { StackScreenProps, useHeaderHeight } from '@react-navigation/stack'
```
kathaypacific, line 1: this file is mostly copy paste + remove some complexity from the VerificationInputScreen component

src/verify/PhoneVerificationStartScreen.tsx
```diff
@@ -0,0 +1,293 @@
+import { Countries } from '@celo/utils/lib/countries'
```
kathaypacific, line 1: this file is mostly copy paste + remove some complexity from the VerificationEducationScreen component

src/verify/PhoneVerificationStartScreen.tsx
```diff
@@ -0,0 +1,293 @@
...
+
+  const onPressStart = async () => {
+    if (!canUsePhoneNumber()) {
+      return
+    }
+    // TODO figure out what this nux thing does
+    dispatch(setHasSeenVerificationNux(true))
+    // TODO dispatch action to call start verification service
+    navigate(Screens.VerificationInputScreen, {
+      registrationStep: route.params?.hideOnboardingStep ? undefined : { step, totalSteps },
```
kathaypacific, line 73: this fixes a pet peeve bug of mine, where if you verify the phone number after onboarding the registration steps show something like "step 4 of 3" on the input screen...that part shouldn't be there at all

locales/base/translation.json
```diff
@@ -760,6 +760,34 @@
...
+    }
+  },
+  "phoneVerificationInput": {
+    "title": "Enter the code",
+    "description": "We sent a code to {{phoneNumber}}, please enter it below.",
+    "help": "Help",
+    "codeInputPlaceholder": "12345678",
+    "helpDialog": {
+      "title": "Help",
+      "body": "We sent an SMS to your phone number. Please check your text messages to find the 8 digit code. If you’re having trouble receiving messages, you can skip this step for now and try again later.",
```
jeanregisser, line 786: The makes me think the default Twilio code length is 6, we can change it.
But I wonder what's best for our users. We can check async.
https://www.twilio.com/docs/verify/developer-best-practices#explore-customization-options

src/navigator/Navigator.tsx
```diff
@@ -120,8 +122,6 @@ import VerificationLoadingScreen from 'src/verify/VerificationLoadingScreen'
 import WalletConnectSessionsScreen from 'src/walletConnect/screens/Sessions'
 import WalletConnectRequest from 'src/walletConnect/screens/WalletConnectRequest'
 import WebViewScreen from 'src/webview/WebViewScreen'
-import KycLanding from 'src/fiatconnect/KycLanding'
-import KycStatus from 'src/fiatconnect/KycStatus'
```
jeanregisser, line 124: This reminds me, we should let prettier / eslint orgamize imports to avoid this.
Maybe via https://github.com/trivago/prettier-plugin-sort-imports?

src/navigator/Navigator.tsx
```diff
@@ -190,7 +190,6 @@ const verificationScreens = (Navigator: typeof Stack) => {
       <Navigator.Screen
         name={Screens.VerificationInputScreen}
         component={VerificationInputScreen}
-        options={VerificationInputScreen.navigationOptions}
```
jeanregisser, line 193: I think we need to keep these options to avoid breaking the existing flow.

src/verify/PhoneVerificationStartScreen.tsx
jeanregisser: Naively I was thinking we could slightly branch in the existing `VerificationEducationScreen`.
But looks like this approach is better. There's quite some complexity there already.

src/verify/PhoneVerificationStartScreen.tsx
```diff
@@ -0,0 +1,293 @@
...
+    : undefined
+
+  const onPressStart = async () => {
+    if (!canUsePhoneNumber()) {
+      return
+    }
+    // TODO figure out what this nux thing does
+    dispatch(setHasSeenVerificationNux(true))
+    // TODO dispatch action to call start verification service
+    navigate(Screens.VerificationInputScreen, {
```
jeanregisser, line 72: Could we directly navigate to the new `PhoneVerificationInputScreen` instead?
This would avoid touching the existing `VerificationInputScreen` completely.
I think only the decentralized `VerificationLoadingScreen` navigates to it. So if we've already branches the `VerificationEducationScreen`, we're fine.

jeanregisser (approved):
Wow fantastic!!! 🚀 
This makes me feel we're super close to ship the new verification system!!

Already approved with some small comments.

src/verify/PhoneVerificationStartScreen.tsx
kathaypacific: ah, do you mean something like this? https://github.com/valora-inc/wallet/pull/2894/commits/6f0bf233ade2559d15ac7903744848a740175225 i was in two minds about what is best to do here, i initially had a new PhoneVerificationInputScreen in the navigator but i changed it to do the switch inside `VerificationInputScreen` so that we can keep the namespace when removing the decentralised flow later on. :/ but indeed it does touch the existing screen. I'll think of a better screen name and put it back in the navigator

jeanregisser: yes, ah well either way is fine then.

kathaypacific: i renamed the PhoneVerificationInputScreen to VerificationCodeInputScreen - it's a bit confusing to have `VerificationCodeInputScreen` and `VerificationInputScreen` but when we remove the decentralised flow we'll be left with `VerificationEducationScreen` and `VerificationCodeInputScreen` which is not so bad :)

ValoraQA:

CC: @kathaypacific 
Thanks.!

PR #2899, opened 2022-09-22
feat: adjust selectFiatConnectQuote saga error handling to be used for persona success

XrayBravoGolf:
### Description

The kyc schema is sent to IHL after success through endpoint defined in IHL client.
The function is enclosed in `KYCLanding` screen, and sent inside Persona button as a prop and consumed by the `onComplete` function of the Persona API.

### Other changes

Documentation of IHL endpoint used is updated from a spelling typo

### Tested

Tested for the revised error handling of the saga for selecting a FiatConnect quote.

### How others should test

N/A

### Related issues

- Fixes [ACT-383]https://linear.app/valora/issue/ACT-383/wallet-when-persona-completes-call-ihl-post-kyc


src/fiatconnect/KycLanding.tsx
```diff
@@ -12,7 +14,20 @@ export interface Props {
 }
 
 function KycLanding(_props: Props) {
-  return <SafeAreaView style={styles.container} />
+  const sendKYCSchema = async () => {
+    if (!_props.quote.quoteResponseKycSchema) {
```
jh2oman, line 18: The `_` prefix is typically only used when the variable is unused in the code. Now that you are using it you can remove it from the `_props` variable name

jh2oman: Instead of digging into the properties of `quote` directly you should use the helper functions on the quote like `getKycSchema`. Especially if they already exist

src/fiatconnect/KycLanding.tsx
```diff
@@ -12,7 +14,20 @@ export interface Props {
 }
 
 function KycLanding(_props: Props) {
```
jh2oman, line 16: It might make the code slightly cleaner to destructure quote and personaKycStatus from the props here. It will avoid so many `prop.` in the rest of the component.

src/fiatconnect/KycLanding.tsx
```diff
@@ -12,7 +14,20 @@ export interface Props {
...
-  return <SafeAreaView style={styles.container} />
+  const sendKYCSchema = async () => {
+    if (!_props.quote.quoteResponseKycSchema) {
+      throw 'expected KYC schema to send is nullish'
```
jh2oman, line 19: Can you add a comment saying this should never happen? B/c a user should only be able to reach the kyc flow at all if there is a kyc schema

src/fiatconnect/KycLanding.tsx
```diff
@@ -12,7 +14,20 @@ export interface Props {
...
+    }
+    await postKyc({
+      providerInfo: _props.quote.quote.provider,
+      kycSchema: _props.quote.quoteResponseKycSchema.kycSchema,
```
jh2oman, line 23: same here with `getKycSchema`

jophish (changes requested):
We can actually get a bit of a two-birds-one-stone thing here, instead of this one-off approach.

If you look at the flowchart [here](https://www.figma.com/file/fqLxcOIvDxQWAlDY60rCD6/FiatConnect-Persona-Diagram?node-id=3%3A743), you can see that what we _really_ want to be doing after sending KYC to Persona is actually just continuing the "happy path" of the select provider flow, beginning with posting KYC to the FiatConnect provider via IHL. The entire select provider flow is handled by the [handleSelectFiatConnectQuote](https://github.com/valora-inc/wallet/blob/main/src/fiatconnect/saga.ts#L279) saga.

If you look at that saga, you can see that the [very first thing](https://github.com/valora-inc/wallet/blob/main/src/fiatconnect/saga.ts#L290) that the saga does is check a user's KYC status. If a user has KYC with Persona but _not_ with their selected FC provider, we'll [post KYC to IHL](https://github.com/valora-inc/wallet/blob/main/src/fiatconnect/saga.ts#L299). After this, _we continue with the rest of the select provider flow as normal_. This is exactly the behavior that we want in this PR, once a user has submitted KYC to Persona.

So, rather than just calling IHL's `POST /kyc` endpoint like you're doing here, we can instead just dispatch an action to trigger the `handleSelectFiatConnectQuote` saga, like we do when beginning the select provider flow. This will post the user's KYC to IHL, with the added bonus of sending the user through the rest of the flow automatically.

The only other change that this would require, as far as I can tell, is a small change in the saga's [error handling](https://github.com/valora-inc/wallet/blob/main/src/fiatconnect/saga.ts#L377). Currently, if the saga errors, we remain on the same page and display a red "error" banner on the top of the page. For now, this is fine, since we were only triggering the saga from the SelectProvider page; if an error occured, the user could just click on the provider again to retry everything/retrigger the saga. Since we'd now be triggering the saga from the _Persona_ screen, we wouldn't want the user to _stay_ on the Persona screen in case of an error; we'd want to send them back to the SelectProvider screen so they can try again. The simplest way to do this would be to just directly add a `navigate` call to the SelectProvider screen at the very end of the `catch` block in that saga.

src/fiatconnect/KycLanding.tsx
```diff
@@ -1,18 +1,29 @@
...
   quote: FiatConnectQuote
 }
 
-function KycLanding(_props: Props) {
-  return <SafeAreaView style={styles.container} />
+function KycLanding(props: Props) {
+  const dispatch = useDispatch()
+  const sendKYCSchema = () => {
+    dispatch(selectFiatConnectQuote({ quote: props.quote }))
+    // will navigate to FiatConnectReview by saga
```
jophish, line 20: nit:
```suggestion
    // will continue with quote selection flow by saga
```

jophish (dismissed):
Please add tests for the changes in each of the files (`FiatConnectQuote.ts`, `KycLanding.ts`, `saga.ts`).

Also @jh2oman, do we have a ticket for actually implenting the UI of the KYC landing page?

XrayBravoGolf:
> a ticket for actually implenting the UI of the KYC landing page?

Is ACT-382 the ticket?

jh2oman:
Yep @XrayBravoGolf @jophish I'm doing the KYC landing page right now with https://linear.app/valora/issue/ACT-382/wallet-new-kyc-primer-screen-and-fiatdetails-primer-for-kyc-flow

src/fiatconnect/saga.ts
```diff
@@ -484,6 +484,17 @@ export function* handleSelectFiatConnectQuote({
     )
     yield put(selectFiatConnectQuoteCompleted())
     yield put(showError(ErrorMessages.PROVIDER_FETCH_FAILED))
+    const amount = {
+      crypto: new BigNumber(quote.getCryptoAmount()).toNumber(),
+      // Rounding up to avoid decimal errors from providers. Won't be
+      // necessary once we support inputting an amount in both crypto and fiat
+      fiat: Math.round(new BigNumber(quote.getFiatAmount()).toNumber()),
```
cajubelt, line 491: I think this will create a bug. The FC spec says to only include fiat amount XOR crypto amount, see [here](https://github.com/fiatconnect/specification/blob/main/fiatconnect-api.md#34113-semantics). 

I suggest using crypto amount for transfers out and fiat amount for transfers in. Also, I don't think you need the `toNumber()` part; if anything that is likely to cause rounding errors.

Last, rounding up doesn't make sense, because you could exceed the user's balance. I don't think rounding is necessary if we just provide one amount, but if you still think you need to round, don't round up.

XrayBravoGolf: The rounding was a sloppy copy-paste from elsewhere logic that passed without me thinking what circumstanced it applied.
There's a conversion from string involved hence the `toNumber()` and bn
I agree with the XOR part according to spec, but looks like the `handleFetchFiatConnectQuotes` saga (called by `fetchFiatConnectQuotes` action in subsequent Select Provider page) always take a crypto amount. On the other hand, `SelectProvider` requires both (in route.prarms [here](https://github.com/valora-inc/wallet/blob/b49eaa2fc66ea066b730a74e44d5bbd04a4ac8b9/src/navigator/types.tsx#L256)) so I need to figure out what to do.

jh2oman (approved):
Looks good!

satish-ravi (commented):
changes look good, but maybe rename the title of the PR since this is mainly changing the error handling on the fiatconnect saga?

src/fiatconnect/KycLanding.test.tsx
```diff
@@ -60,6 +65,32 @@ describe('KycLanding', () => {
...
+      ValoraAnalytics.track = jest.fn()
+      const { getByTestId } = render(
+        <Provider store={store}>
+          <KycLanding {...props} />
+        </Provider>
+      )
+      fireEvent.press(getByTestId('checkbox'))
+      fireEvent.press(getByTestId('PersonaButton'))
+      expect(ValoraAnalytics.track).toHaveBeenCalledWith(CICOEvents.persona_kyc_start)
+      jest.clearAllMocks()
```
satish-ravi, line 92: this can be dropped since we do this on `afterEach`

src/fiatconnect/KycLanding.test.tsx
```diff
@@ -60,6 +65,32 @@ describe('KycLanding', () => {
...
+      render(
+        <Provider store={store}>
+          <KycLanding {...props} />
+        </Provider>
+      )
+      expect(Persona.default).toHaveBeenCalledWith(
+        expect.objectContaining({ onSuccess: expect.any(Function) }), // props for Persona
+        expect.anything()
+      )
+      jest.restoreAllMocks()
```
satish-ravi, line 80: You can use `mockImplementationOnce` above and avoid this restoreAll or just explicitly restore the specific mock, as this could potentially lead to issues in case we have some global mock shared among tests.

src/fiatconnect/KycLanding.tsx
```diff
@@ -120,6 +120,10 @@ export function KycAgreement(props: { personaKycStatus?: KycStatus; quote: FiatC
     navigateToURI(PRIVACY_LINK)
   }
 
+  const personaSuccessCallback = () => {
```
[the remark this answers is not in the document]
satish-ravi: Also, I think if you revert this change, this would avoid the codecov/patch warning.

PR #2903, opened 2022-09-23
feat: add backend requests for CPV screens

kathaypacific:
### Description

### Other changes

### Tested

### How others should test

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

### DIFF
--- a/src/components/CodeInput.tsx
+++ b/src/components/CodeInput.tsx
@@ -123,7 +123,7 @@ export default function CodeInput({
 
             {showInput ? (
               <TextInput
-                textContentType={shortVerificationCodesEnabled ? 'username' : undefined}
+                textContentType={shortVerificationCodesEnabled ? 'oneTimeCode' : undefined}
                 showClearButton={false}
                 value={inputValue}
                 placeholder={
--- a/src/verify/VerificationStartScreen.tsx
+++ b/src/verify/VerificationStartScreen.tsx
@@ -19,7 +19,7 @@ import KeyboardAwareScrollView from 'src/components/KeyboardAwareScrollView'
 import PhoneNumberInput from 'src/components/PhoneNumberInput'
 import TextButton from 'src/components/TextButton'
 import i18n from 'src/i18n'
-import { setHasSeenVerificationNux } from 'src/identity/actions'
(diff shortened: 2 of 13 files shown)


src/components/CodeInput.tsx
```diff
@@ -123,7 +123,7 @@ export default function CodeInput({
 
             {showInput ? (
               <TextInput
-                textContentType={shortVerificationCodesEnabled ? 'username' : undefined}
+                textContentType={shortVerificationCodesEnabled ? 'oneTimeCode' : undefined}
```
kathaypacific, line 126: hope this doesn't cause problems, textContentType affects iOS only and "oneTimeCode" allows us to automatically pick up the value from received messages

src/verify/VerificationStartScreen.tsx
```diff
@@ -143,25 +137,6 @@ function VerificationStartScreen({
...
-  const canUsePhoneNumber = () => {
-    const countryCallingCode = country?.countryCallingCode || ''
-    if (
-      cachedNumber === phoneNumberInfo.e164Number &&
-      cachedCountryCallingCode === countryCallingCode
-    ) {
-      return true
-    }
-
-    const { SANCTIONED_COUNTRY } = getCountryFeatures(phoneNumberInfo.countryCodeAlpha2)
```
kathaypacific, line 155: this logic is already present in the country picker and i couldn't enter an number from a sanctioned country anyway so i think it is safe to remove

src/verify/VerificationStartScreen.tsx
```diff
@@ -143,25 +137,6 @@ function VerificationStartScreen({
...
-    if (SANCTIONED_COUNTRY) {
-      dispatch(showError(ErrorMessages.COUNTRY_NOT_AVAILABLE))
-      return false
-    }
-
-    dispatch(setPhoneNumber(phoneNumberInfo.e164Number, countryCallingCode))
```
kathaypacific, line 161: this line shouldn't be dispatched until we've verified the number - this is still in the decentralised flow but it causes unverified numbers to be displayed in the app (like on the drawer menu)

kathaypacific:
Closing in favour of #2905

PR #2906, opened 2022-09-23
test(fiatconnect): cleanup handleAttemptReturnUserFlow saga tests

satish-ravi:
### Description

Wrote additional tests for `_getSpecificQuote` since `handleAttemptReturnUserFlow` was actually mocking it and not really testing all paths in `_getSpecificQuote`

### Other changes

Made all test names consistent

### Tested

Unit tests

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

N/A


src/fiatconnect/saga.test.ts
```diff
@@ -609,33 +607,13 @@ describe('Fiatconnect saga', () => {
         .put(attemptReturnUserFlowCompleted())
         .run()
       expect(navigate).toHaveBeenCalledWith(Screens.SelectProvider, selectProviderParams)
+      expect(Logger.debug).toHaveBeenCalledWith(
+        'FiatConnectSaga',
+        'Failed to use previous fiatAccount to take user directly to Review Screen',
+        new Error('Could not find quote')
+      )
     })
-    it('navigates to SelectProvider when the quote does not match the fiatAccountType', async () => {
```
satish-ravi, line 613: this was actually not testing this scenario, since this logic is in `_getSpecificQuote` and it was mocked. The test was passing because of an undefined access error was thrown

src/fiatconnect/saga.test.ts
```diff
@@ -609,33 +607,13 @@ describe('Fiatconnect saga', () => {
...
-              fiatAccountType: params.payload.fiatAccountType,
-            }),
-            expectedNormalizedQuote,
-          ],
-        ])
-        .put(attemptReturnUserFlowCompleted())
-        .run()
-      expect(navigate).toHaveBeenCalledWith(Screens.SelectProvider, selectProviderParams)
-    })
-    it('navigates to SelectProvider when the provider cannot be found', async () => {
```
satish-ravi, line 638: same as above

src/fiatconnect/saga.test.ts
```diff
@@ -922,4 +893,77 @@ describe('Fiatconnect saga', () => {
...
+                  cryptoAmount: 2,
+                  providerIds: ['provider-two'],
+                }),
+                [mockFiatConnectQuotes[1]],
+              ],
+            ])
+            .run()
+      ).rejects.toThrow('Could not find quote')
+    })
+    it('throws if no quote with matching provider ID is found', async () => {
```
satish-ravi, line 944: @jh2oman is this even a valid scenario? This tests for the 2nd condition here https://github.com/valora-inc/wallet/blob/806191e7ca28d5ad81b5d1f95bf400f10770dfd4/src/fiatconnect/saga.ts#L328, but can that even happen since we get quote for the specific provider? Wonder if we can drop it.

jophish: Shouldn't be possible, I think. We have [a check](https://github.com/valora-inc/valora-rest-api/blob/main/src/fiatConnect/getQuotes.ts#L37) in valora-rest-api that should prevent this from ever occuring.

jophish (approved):

src/fiatconnect/saga.test.ts
jh2oman: Yea shouldn't be possible. I think thats a relic of one iteration of this code where it may have been possible, where we got ALL quotes first or something like that. 

I think we can remove

jh2oman (approved):

src/fiatconnect/saga.test.ts
satish-ravi: 👍 removed

ValoraQA:

CC: @satish-ravi 
Thanks.!

PR #2909, opened 2022-09-23
feat(analytics): setup statsig experiment

anthonydaoud:
### Description

This PR adds the experimentation platform Statsig to the wallet as we are trying out the platform on [this experiment](https://www.notion.so/valora-inc/Onboarding-Name-Step-4b6ef3485ac64b0caeb87147887f68a5). The `NameAndPicture` screen will obtain the two experiment parameters from Statsig in a separate PR.

### Other changes

Updated `react-native-async-storage` and `react-native-device-info` as part of the Statsig install process.

### Tested

No testing performed.

### How others should test

N/A

### Related issues

- [ACT-437](https://linear.app/valora/issue/ACT-437/statsig-usage-for-name-screen-experiment-parameters)

### Backwards compatibility

N/A


src/web3/saga.test.ts
```diff
@@ -24,9 +24,12 @@ import {
   mtwAddressSelector,
   walletAddressSelector,
 } from 'src/web3/selectors'
+import statsig from 'statsig-js'
 import { createMockStore } from 'test/utils'
 import { mockAccount, mockAccount2, mockAccount3 } from 'test/values'
 
+statsig.updateUser = jest.fn().mockImplementation(() => Promise.resolve(true))
```
jh2oman, line 31: Can you test that `updateUser` is also called with the expected value?

jh2oman:
Looks really good, just a few comments

src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -29,11 +31,24 @@ import { default as useSelector, default as useTypedSelector } from 'src/redux/u
...
+  const [showSkipButton, setShowSkipButton] = useState(
+    ExperimentParams[StatsigLayers.NAME_AND_PICTURE_SCREEN].showSkipButton.defaultValue
+  )
+  const [nameType, setNameType] = useState(
+    ExperimentParams[StatsigLayers.NAME_AND_PICTURE_SCREEN].nameType.defaultValue
+  )
+
+  //TODO: use these in an experiment
+  Logger.info('NameAndPicture', 'Statsig experiment showSkipButton', showSkipButton)
+  Logger.info('NameAndPicture', 'Statsig experiment nameType', nameType)
```
cajubelt, line 50: need to remove these logs, they look like they were for debugging

src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -100,6 +135,11 @@ function NameAndPicture({ navigation }: Props) {
       return
     }
 
+    try {
+      Statsig.logEvent(StatsigEvents.ONBOARDING_NAME_STEP_COMPLETE)
```
cajubelt, line 139: personally I'd prefer to use our existing analytics events and just publish them to StatSig via Segment, rather than have a bespoke parallel analytics event for everything we want to measure with StatSig

src/web3/saga.ts
```diff
@@ -96,6 +97,13 @@ export function* getOrCreateAccount() {
     }
 
     yield call(storeMnemonic, mnemonic, accountAddress)
+    try {
+      yield call(async (userID) => {
+        await Statsig.updateUser({ userID })
+      }, accountAddress.toLowerCase())
```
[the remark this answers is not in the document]
cajubelt: ➕

src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -100,6 +135,11 @@ function NameAndPicture({ navigation }: Props) {
...
+    } catch (error) {
```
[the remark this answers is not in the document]
anthonydaoud: If Statsig fails to initialize, then calling logEvent will throw an error that needs to be handled. I've tested this by commenting out [this](https://github.com/valora-inc/wallet/blob/802ec8956d38fd199f10bcf0ad74182406c7cb38/src/analytics/ValoraAnalytics.ts#L109) initialize function and going through the onboarding flow.

src/onboarding/registration/NameAndPicture.tsx
anthonydaoud: Made an [issue](https://linear.app/valora/issue/ACT-467/segmentstatsig-integration) to handle this in a separate PR, assuming we decide to stick with Statsig

src/onboarding/registration/NameAndPicture.test.tsx
```diff
@@ -183,32 +192,42 @@ describe('NameAndPictureScreen', () => {
     expect(getByText('nameAndPicGuideCopyContent')).toBeTruthy()
   })
   it('does not render skip button when configured so', () => {
-    // TODO replace route param with mock Statsig flag
+    ExperimentParams[StatsigLayers.NAME_AND_PICTURE_SCREEN].showSkipButton.defaultValue = false
+    Statsig.getLayer = jest
+      .fn()
+      .mockImplementation(() => ({ get: jest.fn().mockReturnValue(false) }))
```
satish-ravi, line 198: instead of overriding this again here, you can do something like
```js
const mockStatsigGet = jest.fn().mockReturnValue(null) // default implementation
jest.mock('statsig-react-native', () => ({
  Statsig: {
    getLayer: jest.fn().mockReturnValue({get: mockStatsigGet})
  },
}))
```
and then inside specific functions where you want a different response, you can do
```js
mockStatsigGet.mockReturnValueOnce(false)  // or true
```

src/analytics/ValoraAnalytics.ts
```diff
@@ -90,6 +98,22 @@ class ValoraAnalytics {
...
+    try {
+      const { accountAddress } = getCurrentUserTraits(store.getState())
+      const stasigUser = accountAddress
+        ? {
+            userID: accountAddress,
+          }
+        : null
+
+      await Statsig.initialize(STATSIG_API_KEY, stasigUser, {
+        overrideStableID: uniqueID,
```
satish-ravi, line 111: just for my understanding, is this the ID that is used for bucketing for this experiment since we won't have a wallet address? And can we specify what ID to use for bucketing on the experiment configuration?

package.json
```diff
@@ -80,7 +80,7 @@
     "@komenci/contracts": "1.1.0",
     "@komenci/kit": "1.0.0-beta1",
     "@pedrouid/environment": "^1.0.1",
-    "@react-native-async-storage/async-storage": "^1.15.14",
+    "@react-native-async-storage/async-storage": "^1.17.10",
```
cajubelt, line 83: is this related? if not plz split it out into a separate PR

package.json
```diff
@@ -150,13 +150,14 @@
     "react-native-camera": "^4.2.1",
     "react-native-config": "https://github.com/luggit/react-native-config#2f68b94",
     "react-native-contacts": "https://github.com/celo-org/react-native-contacts#9940121",
-    "react-native-device-info": "^8.3.1",
+    "react-native-device-info": "^10.2.0",
```
cajubelt, line 153: same here

satish-ravi (commented):
Looks good, just a question and a comment on jest mocking

src/analytics/ValoraAnalytics.ts
```diff
@@ -77,9 +86,8 @@ class ValoraAnalytics {
       try {
         const deviceInfo = await getDeviceInfo()
         this.deviceInfo = deviceInfo
-        this.sessionId = sha256FromString(
-          '0x' + deviceInfo.UniqueID.split('-').join('') + String(Date.now())
-        )
+        uniqueID = await deviceInfo.UniqueID
```
cajubelt, line 89: you don't need to await this. it's not a promise.

cajubelt:
```suggestion
        uniqueID = deviceInfo.UniqueID
```

src/analytics/ValoraAnalytics.ts
cajubelt, line 111: is there a reason you're doing this?

cajubelt: I'm curious why it's necessary

src/onboarding/registration/NameAndPicture.test.tsx
```diff
@@ -183,32 +192,42 @@ describe('NameAndPictureScreen', () => {
     expect(getByText('nameAndPicGuideCopyContent')).toBeTruthy()
   })
   it('does not render skip button when configured so', () => {
-    // TODO replace route param with mock Statsig flag
+    ExperimentParams[StatsigLayers.NAME_AND_PICTURE_SCREEN].showSkipButton.defaultValue = false
```
cajubelt, line 195: tests shouldn't mutate constants from application code. if you want to do this, find a way to mock ExperimentParams and mess with that, instead. or refactor to use dependency injection

src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -50,7 +61,27 @@ function NameAndPicture({ navigation, route }: Props) {
   const asyncKomenciReadiness = useAsyncKomenciReadiness()
   const showGuidedOnboarding = useSelector(showGuidedOnboardingSelector)
   const createAccountCopyTestType = useSelector(createAccountCopyTestTypeSelector)
-  const skipUsername = route.params?.skipUsername //TODO repalce with statsig variable
+
+  useEffect(() => {
```
cajubelt, line 65: is useEffect the right hook to use here? My understanding is that useEffect is for a different use case, where you want to run some side-effect just after the render phase.

cajubelt: maybe getting rid of useState and useEffect, and using useCallback instead would be better. But even that feels like overkill; observing experiment parameters shouldn't be expensive to do on each render.

src/onboarding/registration/NameAndPicture.test.tsx
cajubelt: tbh I don't think it's necessary to test fictitious variations of the default value though

package.json
anthonydaoud: The package updates are part of the Statsig react native [install process](https://docs.statsig.com/client/reactNativeSDK).

src/analytics/ValoraAnalytics.ts
anthonydaoud: This is a promise in the new version of `react-native-device-info`. Actually, I'll just change [`getDeviceInfo`](https://github.com/valora-inc/wallet/blob/main/src/analytics/ValoraAnalytics.ts#L41) so that the await happens in there.

src/analytics/ValoraAnalytics.ts
anthonydaoud: Yes that's the ID that is used for the experiment. Also, yes you can specify whatever ID you'd like to use in the Statsig web app.

anthonydaoud: I am getting an error stating that Statsig is uninitialized when I don't specify the stable ID manually.

src/onboarding/registration/NameAndPicture.test.tsx
anthonydaoud: Ok makes sense. Also, I'll update these tests in a separate PR.

src/analytics/ValoraAnalytics.ts
cajubelt: ok. might be worth adding that as a code comment

src/analytics/ValoraAnalytics.test.ts
```diff
@@ -137,6 +138,7 @@ describe('ValoraAnalytics', () => {
   beforeEach(() => {
     jest.clearAllMocks()
     jest.unmock('src/analytics/ValoraAnalytics')
+    Statsig.initialize = jest.fn().mockImplementation(() => Promise.resolve())
```
cajubelt, line 141: nit: I think this could just be `jest.fn()`

src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -113,6 +116,11 @@ function NameAndPicture({ navigation, route }: Props) {
       return
     }
 
+    try {
+      Statsig.logEvent(StatsigEvents.ONBOARDING_NAME_STEP_COMPLETE)
```
cajubelt, line 120: nit: since the other statsig stuff for this component is taken out of this PR, might make sense to separate this out as well

cajubelt: to keep it single purpose

cajubelt (approved):
🥳 🥳 🥳

PR #2912, opened 2022-09-26
feat(fiatconnect): Check KYC after submitting account

jophish:
### Description
Fixes ACT-384. This PR adds checks for KYC status after a user successfully submits a fiat account on the FiatDetails page, when KYC is required for the quote the user has selected. Depending on the KYC status, we either navigate to the Review screen, or a particular KYC status screen. In the exact case where KYC is required _and_ it has been approved by the time the user has submitted their account, we show a green checkmark to the user to indicate their ID verification success.

### Other changes

### Tested

Unit tested. Not easily end-to-end testable at the moment.

### How others should test

### Related issues

Fixes ACT-384.

### Backwards compatibility


src/redux/migrations.ts
```diff
@@ -823,4 +824,11 @@ export const migrations = {
       sendingFiatAccount: false,
     },
   }),
+  79: (state: any) => ({
+    ...state,
+    fiatConnect: {
+      ..._.omit(state.fiatConnect, 'sendingFiatAccount'),
```
cajubelt, line 830: it seems fine to set `sendingFiatAccountStatus` to `NotSending` when `sendingFiatAccount` is `false`, but what about when it is `true`?

cajubelt: seems like a more complex migration might be needed in that case

cajubelt: or if it's expected to never happen, since migrations occur at a time when it's not possible to be actively sending fiat account details to the server, maybe log an error if it does occur?

cajubelt (commented):
not done, need to go to the airport but leaving a couple comments

src/fiatconnect/saga.ts
```diff
@@ -166,6 +169,64 @@ export function* handleSubmitFiatAccount({
...
+    if (kycSchema) {
+      try {
+        // If and only if KYC is required, we wait a bit to give KYC some time to process.
+        yield delay(KYC_WAIT_TIME_MILLIS)
+        const getKycStatusResponse = yield call(getKycStatus, {
+          providerInfo: quote.quote.provider,
+          kycSchemas: [kycSchema],
+        })
+        const fiatConnectKycStatus = getKycStatusResponse.kycStatus[kycSchema]
+        // const fiatConnectKycStatus = FiatConnectKycStatus.KycApproved as FiatConnectKycStatus
```
jh2oman, line 187: leftover from developing?

src/fiatconnect/slice.ts
```diff
@@ -124,10 +131,13 @@ export const slice = createSlice({
   initialState,
   reducers: {
     submitFiatAccount: (state, action: PayloadAction<SubmitFiatAccountAction>) => {
-      state.sendingFiatAccount = true
+      state.sendingFiatAccountStatus = SendingFiatAccountStatus.Sending
```
jh2oman, line 134: This `sendingFiatAccountStatus` field should probably be added to the REHYDRATE so that it resets when the app reloads

src/redux/migrations.ts
jh2oman: > expected to never happen, since migrations occur at a time when it's not possible to be actively sending fiat account details to the server

I think this is correct. I don't think its something that we need an error case to handle. Especially if we codify this as expected behavior by adding sendingFiatAccountStatus to rehydrate.

src/fiatconnect/saga.ts
cajubelt, line 187:
```suggestion
```

src/fiatconnect/saga.ts
```diff
@@ -166,6 +169,64 @@ export function* handleSubmitFiatAccount({
...
+        switch (fiatConnectKycStatus) {
+          case FiatConnectKycStatus.KycApproved:
+            yield put(submitFiatAccountKycApproved())
+            yield delay(500) // Allow user to admire green checkmark
```
cajubelt, line 191: > Allow user to admire green checkmark
😂

src/fiatconnect/saga.ts
```diff
@@ -166,6 +169,64 @@ export function* handleSubmitFiatAccount({
...
+            break
+          // Denied, Expired, and Pending all fall through to the default case.
+          case FiatConnectKycStatus.KycDenied:
+            navigate(Screens.KycDenied, {
+              flow,
+              quote,
+              retryable: true, // TODO: Get this dynamically once IHL supports it
```
cajubelt, line 198: what does IHL need to support for this to be dynamic? KYC deletion?

src/fiatconnect/slice.ts
jophish: ty! added

src/redux/migrations.ts
jophish: Yeah, there is only a very brief window of time where `sendingFiatAccount` is anything but `false`; it will always be `false` on app start.

src/fiatconnect/saga.ts
jophish: [ACT-459](https://linear.app/valora/issue/ACT-459/walletihl-limit-number-of-kyc-retries-after-kyc-denied)

jh2oman (approved):

ValoraQA:
Hey @MuckT please let us know do we need to test this PR ?

CC: @jophish 
Thanks.!

PR #2916, opened 2022-09-27
feat(fiatconnect): check KYC on return user flow

satish-ravi:
### Description

Checks if KYC is required on return user flow and if required, only go to Review screen if KYC is approved. If KYC is not created, go to select provider, otherwise to the appropriate KYC status screen. The KYC handling on `handleAttemptReturnUserFlow` is very similar to `handleSelectFiatConnectQuote`, but different enough for them to be in separate sagas.
Few things that are different:
- `handleAttemptReturnUserFlow` fetches quote and fiat account in parallel to check for a match before checking KYC, `handleSelectFiatConnectQuote` gets quote as input, checks KYC first and then checks for fiat account
- `handleAttemptReturnUserFlow` navigates to KYC pending screen immediately if KYC is pending whereas `handleSelectFiatConnectQuote` may navigate to fiat account details screen 
- `handleAttemptReturnUserFlow` navigates to select provider screen if KYC is not created with the provider, whereas `handleSelectFiatConnectQuote` navigates to Persona

### Other changes

N/A

### Tested

Unit tests

### How others should test

Not ready for testing since first time KYC flow isn't complete

### Related issues

- Fixes ACT-386

### Backwards compatibility

Yes


jh2oman (approved):
Agreed that they should be separate. Keeps things cleaner even though they look so similar

src/fiatconnect/saga.ts
```diff
@@ -303,6 +303,51 @@ export function* handleAttemptReturnUserFlow({
...
+            flow: normalizedQuote.flow,
+            quote: normalizedQuote,
+          })
+          return
+        case FiatConnectKycStatus.KycDenied:
+          yield put(attemptReturnUserFlowCompleted())
+          navigate(Screens.KycDenied, {
+            flow: normalizedQuote.flow,
+            quote: normalizedQuote,
+            retryable: true,
```
jophish, line 335: nit: add a TODO here so we don't forget to update this eventually

jophish (approved):
Looks great. Adding another switch/case was definitely the right choice :tada:

ValoraQA:

CC: @satish-ravi 
Thanks.!

PR #2917, opened 2022-09-28
feat: add resend code button to verification code input screen

kathaypacific:
### Description

Add a resend button with delay in the verification flow.

30 second timeout is [recommended](https://www.twilio.com/blog/best-practices-retry-logic-sms-2fa) by Twilio.

Recording (note that the error banner is not caused by my account and not the resend button!)

https://user-images.githubusercontent.com/20150449/193266759-b56b6ad9-af5c-44f1-ad92-9899ce3afaf8.mp4

### Other changes

N/A

### Tested

Manually / unit tests

### How others should test

pressing the resend button should resend the verification text

### Related issues

- Relates RET-384

### Backwards compatibility

N/A


src/verify/VerificationCodeInputScreen.tsx
```diff
@@ -94,25 +103,39 @@ function VerificationCodeInputScreen({
 
   return (
     <SafeAreaView style={styles.container} edges={['bottom']}>
-      <KeyboardAwareScrollView
-        style={[styles.scrollContainer, headerHeight ? { marginTop: headerHeight } : undefined]}
-        keyboardShouldPersistTaps="always"
+      <KeyboardAvoidingView
```
kathaypacific, line 106: all these changes are just to enable the keyboard to shrink the screen content so that the resend button is above the keyboard....

src/verify/ResendButtonWithDelay.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const [secondsRemaining, setSecondsRemaining] = useState(RESEND_DELAY_TIME)
+
+  const handleOnPress = () => {
+    setSecondsRemaining(RESEND_DELAY_TIME)
+    onPress()
+  }
+
+  useEffect(() => {
+    const timer = setInterval(() => {
+      setSecondsRemaining((prev) => (prev > 0 ? prev - 1 : 0))
```
jeanregisser, line 23: nit: it's a bit better to compute the difference from now and `startTime + RESEND_DELAY_TIME`. Timers can be delayed in case of user interactions or other activity happening. This way we ensure the the total time is closer to the real time value.

jeanregisser (commented):
Great! 👍

src/verify/VerificationCodeInputScreen.tsx
kathaypacific: ah, thanks for the reminder! i had forgotten about the keyboard spacer component. it works well like you suggested, i also applied a similar change to the phone number screen now

src/verify/hooks.ts
```diff
@@ -77,13 +84,13 @@ export function useVerifyPhoneNumber(phoneNumber: string, countryCallingCode: st
 
   useAsync(
     async () => {
-      if (verificationCodeRequested.current) {
+      if (verificationCodeRequested.current && !shouldResendSms) {
+        // prevent the verification request from being fired multiple times, due
+        // to hot reloading during development only
```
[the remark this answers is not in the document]
kathaypacific: yup, updated to make this more clear

src/verify/ResendButtonWithDelay.tsx
kathaypacific: yeah i suppose that's true, i feel like this is edge casey but maybe makes a difference for very old android devices? i updated and took some code from VerificationCountdown, i didn't share code since we'll delete VerificationCountdown soon

kathaypacific:
thanks for the suggestions @jeanregisser ! i have made the updates :)

jeanregisser (approved):

PR #2921, opened 2022-09-28
feat(fiatconnect): Implement KYC status screens

jophish:
### Description

For ACT-457. Implements KYC status screens for the `KycPending`, `KycExpired`, and `KycDenied` (both retryable/final) states. Currently there's no way to trigger these in-app, since KYC is auto-approved for quotes that require it. Unit tests coming soon, just wanted to get this out to ensure that the designs/copy looks good.

### Tested

Manually tested, unit tests pending.

#### `KycPending`

https://user-images.githubusercontent.com/569401/192894681-b8216368-d826-402b-9cf1-a52a5a55a231.mp4

#### `KycExpired`

https://user-images.githubusercontent.com/569401/192894762-e8a95bca-2dc0-44d3-8d18-8e9cd40e5752.mp4

#### `KycDenied` (retryable)

https://user-images.githubusercontent.com/569401/192894858-edf96ba7-2948-4db5-b7fc-4d65fb2f6b8a.mp4

#### `KycDenied` (final)

https://user-images.githubusercontent.com/569401/192894885-783900ff-2924-4713-98e1-96aa6690058c.mp4

### How others should test

### Related issues

Fixes ACT-457.

### Backwards compatibility


locales/base/translation.json
```diff
@@ -1292,6 +1292,33 @@
     "disclaimer": "You acknowledge and agree that your Personal Data will be provided to us and shared with third-parties for verification purposes as stated in the <0>Privacy Policy</0>",
     "button": "Set up ID Verification"
   },
+  "fiatConnectKycStatusScreen": {
+    "contactSupportPrefill": "I'm having trouble verifying my identity with Valora.",
```
jophish, line 1296: @jh2oman/laura/nitya, any idea what this prefill text should be? I just guessed here, but wondering if there's something specific it should be instead? (Also, should it vary for different KYC states?)

src/fiatconnect/kyc/KycDenied.tsx
```diff
@@ -1,27 +1,139 @@
...
-function KycDenied(_props: Props) {
-  return (
-    <SafeAreaView style={styles.container}>
-      <Text>KYC Denied</Text>
-    </SafeAreaView>
-  )
+function KycDenied({ route, navigation }: Props) {
+  const { quote, flow, retryable } = route.params
+
+  const exchangeRates = useSelector(localCurrencyExchangeRatesSelector)
```
jophish, line 25: This is a kind of unfortunate pattern and one that complicates things significantly. In order to navigate to the `SelectProvder` screen, we need to derive the fiat amount in the user's local currency from the crypto amount in the quote. Actually, for cash _in_ flows, we'll need to do the opposite (crypto amount from fiat). I think the "right" way to do this would be to show a spinner until `exchangeRates` is set and then set the header/render the component for real, but not sure if we want to do that. Frankly, we could also probably get away with using the "other" amount within the quote object (fiat for cash out, crypto for cash in), since the FC API _is_ required to return both `fiatAmount` and `cryptoAmount` for all quotes. This shouldn't cause any issues (I _think_) since the "other" amount is only ever really used by `SelectProviders` when fetching _external_ quotes, where the precision of fiat/crypto amounts aren't as important.

Wondering if folks have any thoughts..

(I believe @XrayBravoGolf also ran into this issue in an open PR as well.)

locales/base/translation.json
cajubelt: I'd vote for including the KYC state in the message somehow. sounds potentially useful for debugging

src/fiatconnect/kyc/KycDenied.tsx
cajubelt: seems like another solution would be to make the fetchProviders cloud function more flexible, and have it accept just one of the two (fiat or crypto amount). that way we don't need to worry about providing both anymore

cajubelt: from a quick look at the cloud function, it looks like Simplex allows either fiat or crypto, and the others appear to require the fiat amount. regardless, we should be able to do the same currency conversion inside the cloud function. but it's kinda interesting that as things currently stand, only the fiat amount is required

src/analytics/Properties.tsx
```diff
@@ -1109,6 +1109,31 @@ interface FiatExchangeEventsProperties {
     flow: CICOFlow
     txHash?: string
   }
+  [FiatExchangeEvents.cico_fc_kyc_status_contact_support]: {
+    provider: string
+    flow: CICOFlow
+    fiatConnectKycStatus: FiatConnectKycStatus
```
cajubelt, line 1115: nit: seems like this could be more DRY

src/fiatconnect/kyc/KycDenied.tsx
```diff
@@ -1,27 +1,139 @@
...
+          onPress={onPressTryAgain}
+          text={t('fiatConnectKycStatusScreen.denied.retryable.tryAgain')}
+          type={BtnTypes.PRIMARY}
+          size={BtnSizes.MEDIUM}
+        />
+        <Button
+          style={styles.switchButton}
+          testID="switchButton"
+          onPress={onPressSwitch}
+          text={t('fiatConnectKycStatusScreen.expired.switch')}
```
cajubelt, line 89: did you mean to use the copy for "expired" here?

src/fiatconnect/kyc/KycDenied.tsx
```diff
@@ -1,27 +1,139 @@
...
+      <SafeAreaView style={styles.container}>
+        <Text style={styles.title}>{t('fiatConnectKycStatusScreen.denied.final.title')}</Text>
+        <Text testID="descriptionText" style={styles.description}>
+          {t('fiatConnectKycStatusScreen.denied.final.description')}
+        </Text>
+        <Button
+          style={styles.switchButton}
+          testID="switchButton"
+          onPress={onPressSwitch}
+          text={t('fiatConnectKycStatusScreen.expired.switch')}
```
cajubelt, line 106: same here

src/fiatconnect/kyc/KycDenied.tsx
cajubelt: from the naming it sounds like it's just for expired

src/fiatconnect/kyc/KycExpired.tsx
```diff
@@ -1,26 +1,121 @@
...
+    navigate(Screens.SelectProvider, {
+      flow,
+      selectedCrypto: quote.getCryptoType(),
+      amount: {
+        crypto: Number(quote.getCryptoAmount()),
+        fiat: Number(
+          convertCurrencyToLocalAmount(
+            quote.getCryptoAmount(),
+            exchangeRates[quote.getCryptoType()]
+          )
```
cajubelt, line 65: for cash-ins this could create some weird edge cases where a user tries to cash in with some amount, say $100, then they get to KycExpired, then when they navigate back to SelectProvider the amount changes to $95. because the currency is converted by the provider to some crypto amount, and then by us back to a fiat amount.

in short, I think just using quote.getFiatAmount() here is better

src/fiatconnect/kyc/KycPending.tsx
```diff
@@ -1,26 +1,106 @@
...
+  const exchangeRates = useSelector(localCurrencyExchangeRatesSelector)
+  useMemo(() => {
+    navigation.setOptions(
+      getNavigationOptions({
+        fiatConnectKycStatus: FiatConnectKycStatus.KycPending,
+        quote: route.params.quote,
+        exchangeRates,
+      })
+    )
+  }, [exchangeRates])
```
cajubelt, line 34: seems like if `route.params.quote` changed, we'd also want this to change. so maybe it should be added to deps. wdyt?

cajubelt: idk how it could change, but even if it doesn't ever change, it still doesn't hurt to be there

cajubelt: same with the other screens in this PR that have useMemo

src/fiatconnect/kyc/KycPending.tsx
```diff
@@ -1,26 +1,106 @@
...
+
+  const { t } = useTranslation()
+
+  const onPressClose = () => {
+    ValoraAnalytics.track(FiatExchangeEvents.cico_fc_kyc_status_contact_support, {
```
cajubelt, line 39:
```suggestion
    ValoraAnalytics.track(FiatExchangeEvents.cico_fc_kyc_status_close, {
```

src/fiatconnect/kyc/getNavigationOptions.tsx
```diff
@@ -0,0 +1,84 @@
...
+      provider: quote.getProviderId(),
+      flow: quote.flow,
+      fiatConnectKycStatus,
+    })
+    navigate(Screens.SelectProvider, {
+      flow: quote.flow,
+      selectedCrypto: quote.getCryptoType(),
+      amount: {
+        crypto: Number(quote.getCryptoAmount()),
+        fiat: Number(
```
cajubelt, line 50: same here with currency conversion and cash ins

cajubelt (commented):
looks good overall!

src/fiatconnect/kyc/KycDenied.tsx
jophish: good catch, updated.

src/fiatconnect/kyc/KycExpired.tsx
jophish: Sure, makes sense -- that actually simplifies things a ton.

src/fiatconnect/kyc/KycDenied.tsx
jophish: If we're just getting the fiat amount directly from the quote object itself, things are actually pretty simple here; I agree that it would be useful if `SelectProvider` didn't require both values, but it actually doesn't increase the complexity here if we're just pull both amounts from the quote. If it's worth it, we could refactor the cloud function/wallet such that only a single amount is required when navigating to `SelectProvider` as a followup?

cajubelt: alright

cajubelt (approved):
🥳

satish-ravi (approved):

PR #2922, opened 2022-09-29
feat(fiatconnect): add saga for handling kyc try again

satish-ravi:
### Description

This will be used in the kyc denied and expired screens.

### Other changes

Cleaned up some mocking in IHL tests

### Tested

Unit tests

### How others should test

N/A

### Related issues

- Fixes ACT-458

### Backwards compatibility

N/A


src/fiatconnect/saga.ts
```diff
@@ -740,6 +742,24 @@ export function* handleCreateFiatConnectTransfer({
   }
 }
 
+export function* handleKycTryAgain({ payload }: ReturnType<typeof kycTryAgain>) {
+  const { quote, flow } = payload
+
+  try {
+    yield call(deleteKyc, {
+      providerInfo: quote.getProviderInfo(),
+      kycSchema: quote.getKycSchema()!, // it is impossible for kyc schema to be undefined on the quote
```
cajubelt, line 751: nit: might be nice to throw an error here if it's undefined

src/in-house-liquidity/index.ts
```diff
@@ -162,6 +162,33 @@ export async function postKyc({
...
+}): Promise<void> {
+  const response = await exports.makeRequest({
+    providerInfo,
+    path: `/fiatconnect/kyc/${providerInfo.id}/${kycSchema}`,
+    options: { method: 'DELETE' },
+  })
+
+  if (!response.ok) {
+    throw new Error(`Got non-ok response from IHL while deleting KYC: ${response.status}`)
+  }
```
cajubelt, line 189: might be better to just log a warning if the status is 404

cajubelt: since we wouldn't need to delete if it's already not found right?

src/fiatconnect/saga.ts
```diff
@@ -740,6 +742,24 @@ export function* handleCreateFiatConnectTransfer({
...
+    })
+
+    navigate(Screens.KycLanding, { quote, flow, step: 'one' })
```
cajubelt, line 754: separate from this PR but it seems like the KYC steps could be named more descriptively

cajubelt (approved):
see comments, might want to do something different if the response code is 404

cajubelt (approved):

PR #2931, opened 2022-09-29
fix(fiatconnect): Fix issue when checking KYC after submitting fiat account

jophish:
### Description

TLDR; the switch/case that deals with KYC statuses after posting a fiat account was falling through when it wasn't supposed to, which resulted in `navigate` being called several times in a row. The tests didn't catch it, since they didn't check to make sure that `navigate` was only being called a _single_ time. Fixed the implementation, as well as updated the tests to ensure that this sort of thing gets caught.

### Other changes

N/A.

### Tested

Unit tested.

### How others should test

N/A.

### Related issues

N/A.

### Backwards compatibility

N/A.


src/fiatconnect/saga.ts
```diff
@@ -189,27 +189,35 @@ export function* handleSubmitFiatAccount({
             yield put(submitFiatAccountKycApproved())
             yield delay(500) // Allow user to admire green checkmark
             break
-          // Denied, Expired, and Pending all fall through to the default case.
```
cajubelt, line 192: we were right about how default was working right? and the problem was just that an unrecognized status was cropping up somehow?

cajubelt (approved):

src/fiatconnect/saga.ts
jophish: the issue is that the cases were not falling through _directly_ to default, they were falling through to _subsequent_ KYC status cases _first_ calling `navigate` multiple times, and only _then_ to default. The change to the default case is to bring this in line with the _other_ KYC status switch/case we have, and it allows us to identify when we receive an unexpected KYC status.

src/fiatconnect/saga.ts
```diff
@@ -189,27 +189,35 @@ export function* handleSubmitFiatAccount({
...
               flow,
               quote,
             })
-          default:
             yield delay(500) // to avoid a screen flash
             yield put(submitFiatAccountCompleted())
             return
+          default:
+            throw new Error(
+              `Unrecognized FiatConnect KYC status "${fiatConnectKycStatus}" while attempting to handle quote selection for provider ${quote.getProviderId()}`
```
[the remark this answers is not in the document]
jophish: I think it's the best we can do, short of implementing some other novel solution AFAICT. This will eventually loop the user back around to the SelectProvider screen to try again; they can also contact support if they like.

PR #2940, opened 2022-10-03
fix: incorrect connection error on WalletConnect connection

kathaypacific:
### Description

`onBarCodeDetected` can be fired multiple times in a short period of time, triggering multiple attempts to handle the same QR. When a user manually scans a QR code, during the time between when the QR code is first detected and when the connection bottom sheet appears and is acknowledged, other requests to handle the same code are dispatched. This causes the behaviour where after the connection is successful, another connection bottom sheet appears and appears to load and time out.

This PR tries to prevent such duplicate requests from being fired. 

### Other changes

N/A

### Tested

Manually

### How others should test

On a physical device, attempt to connect to a dapp on a browser by scanning the QR code. The connection should be successful.

### Related issues

- Fixes RET-372

### Backwards compatibility

Yes

### DIFF
--- a/src/navigator/QRNavigator.tsx
+++ b/src/navigator/QRNavigator.tsx
@@ -4,7 +4,6 @@ import {
 } from '@react-navigation/material-top-tabs'
 import { useIsFocused } from '@react-navigation/native'
 import { StackScreenProps, TransitionPresets } from '@react-navigation/stack'
-import { memoize } from 'lodash'
 import React, { useEffect, useMemo, useRef, useState } from 'react'
 import { useAsync } from 'react-async-hook'
 import { useTranslation } from 'react-i18next'
@@ -34,6 +33,8 @@ type AnimatedScannerSceneProps = StackScreenProps<QRTabParamList, Screens.QRScan
 
 // Component doing our custom transition for the QR scanner
 function AnimatedScannerScene({ route, position, ...props }: AnimatedScannerSceneProps) {
+  const lastScannedQR = useRef('')
+
   const isFocused = useIsFocused()
   const [wasFocused, setWasFocused] = useState(isFocused)
   const [isPartiallyVisible, setIsPartiallyVisible] = useState(false)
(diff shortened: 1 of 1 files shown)


src/navigator/QRNavigator.tsx
```diff
@@ -95,21 +96,23 @@ function AnimatedScannerScene({ route, position, ...props }: AnimatedScannerScen
   const { scanIsForSecureSend, isOutgoingPaymentRequest, transactionData, requesterAddress } =
     route.params || {}
 
-  const onBarCodeDetected = memoize(
```
kathaypacific, line 98: the `memoize` was causing me problems, and i also don't think it adds value here? why do we need to memoize this function when it's not doing any heavy computations and in theory we only want it to execute once per `qrCode.data` anyway?

MuckT:
@kathaypacific testing locally I am not longer able to reproduce the double scan issue on Android.

MuckT (approved):
🚀

src/navigator/QRNavigator.tsx
jeanregisser: This was how it initially avoided this bug.
But it got incorrectly refactored: [packages/mobile/src/qrcode/QRScanner.tsx](https://github.com/valora-inc/wallet/commit/6e3de6570f42c9917226d3545561fcf914f59a7e#diff-0f9c2a2d085bf912dedd013ea8ea93a98046538b740e6d32176330fe643cd2c6)
See the missing `useCallback` which made it call `memoize` over and over on each render.

Clever code was too clever :D

`memoize` was first added [here](https://github.com/valora-inc/wallet/commit/437a93854a58f5e9d98ffc784663eacb31dc277a) and then I'm guilty of continuing to use it [here](https://github.com/valora-inc/wallet/commit/093734dda4bb7ba023d2013a342b7d2ca1bebfb0#diff-0f9c2a2d085bf912dedd013ea8ea93a98046538b740e6d32176330fe643cd2c6).

jeanregisser (approved):
Thanks!! 🚀

PR #2949, opened 2022-10-04
feat: alternate onboarding name step experiment - username input field placeholder

XrayBravoGolf:
### Description

An alternate palceholder text for username during onboarding is provided to users who are in the experiment group.

### Other changes

All mock StatSig flags were moved to a separate file.

### Tested

Automated testing mocks the function returning feature flag and verifies the alternate placeholder is shown.

### Related issues

- Fixes #ACT-439

### Backwards compatibility

Compatible; no changes to state or logic are involved.


src/onboarding/registration/MockedStatSigFeatureFlag.ts
```diff
@@ -0,0 +1,6 @@
+export const shallShowAlternatePlaceholder = () => {
+  return false
+}
+export const shallSkipUsername = () => {
```
cajubelt, line 4:
```suggestion
export const shouldShowSkipUsername = () => {
```
nit: we don't really use "shall" in naming elsewhere, and grammatically it just makes less sense. Also, the username won't necessarily be skipped if this is true, it's just that the button should be shown enabling the user to skip, if they want.

cajubelt: remember to update usages if you accept this suggestion

src/onboarding/registration/MockedStatSigFeatureFlag.ts
```diff
@@ -0,0 +1,6 @@
+export const shallShowAlternatePlaceholder = () => {
```
cajubelt, line 1:
```suggestion
export const shouldShowNamePlaceholderVariation = () => {
```
see comment below, same idea

src/onboarding/registration/NameAndPicture.test.tsx
```diff
@@ -182,21 +190,24 @@ describe('NameAndPictureScreen', () => {
     expect(getByText('nameAndPicGuideCopyTitle')).toBeTruthy()
     expect(getByText('nameAndPicGuideCopyContent')).toBeTruthy()
   })
+
   it('does not render skip button when configured so', () => {
     // TODO replace route param with mock Statsig flag
```
cajubelt, line 195: this comment doesn't make sense anymore, since it uses a different way of mocking statsig feature flags. please update it so it doesn't become stale

src/onboarding/registration/NameAndPicture.test.tsx
```diff
@@ -182,21 +190,24 @@ describe('NameAndPictureScreen', () => {
...
-        <MockedNavigator component={NameAndPicture} params={{ skipUsername: false }} />
+        <MockedNavigator component={NameAndPicture} />
       </Provider>
     )
     expect(queryByText('skip')).toBeNull()
   })
 
   it('renders skip button when mocked and skipping works', () => {
-    // TODO replace route param with mock Statsig flag
+    // TODO replace with mock Statsig flag
```
cajubelt, line 206: this todo appears to have been done already

src/onboarding/registration/NameAndPicture.test.tsx
```diff
@@ -205,14 +216,25 @@ describe('NameAndPictureScreen', () => {
   })
 
   it('saves empty name regardless of what is in the inputbox when skip is used', () => {
-    // TODO replace route param with mock Statsig flag
+    // TODO replace with mock Statsig flag
```
cajubelt, line 219: same here

src/onboarding/registration/NameAndPicture.test.tsx
```diff
@@ -205,14 +216,25 @@ describe('NameAndPictureScreen', () => {
...
-        <MockedNavigator component={NameAndPicture} params={{ skipUsername: true }} />
+        <MockedNavigator component={NameAndPicture} />
       </Provider>
     )
     fireEvent.changeText(getByTestId('NameEntry'), 'Some Name')
     fireEvent.press(getByText('skip'))
     expect(AccountActions.setName).not.toHaveBeenCalled()
   })
+  it('shows alternate placeholder username', () => {
+    // TODO replace with mock Statsig flag
```
cajubelt, line 231: and here

src/onboarding/registration/NameAndPicture.test.tsx
```diff
@@ -182,21 +190,24 @@ describe('NameAndPictureScreen', () => {
     expect(getByText('nameAndPicGuideCopyTitle')).toBeTruthy()
     expect(getByText('nameAndPicGuideCopyContent')).toBeTruthy()
   })
+
   it('does not render skip button when configured so', () => {
     // TODO replace route param with mock Statsig flag
+    // StatSigFlags.shallSkipUsername = jest.fn().mockReturnValue(true)
```
cajubelt, line 196: in general, we should not include commented out code in our PR's against the `main` branch

src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -50,7 +54,16 @@ function NameAndPicture({ navigation, route }: Props) {
   const asyncKomenciReadiness = useAsyncKomenciReadiness()
   const showGuidedOnboarding = useSelector(showGuidedOnboardingSelector)
   const createAccountCopyTestType = useSelector(createAccountCopyTestTypeSelector)
-  const skipUsername = route.params?.skipUsername //TODO repalce with statsig variable
+  const skipUsername = shallSkipUsername() //TODO repalce with statsig variable
```
cajubelt, line 57:
```suggestion
  const skipUsername = shallSkipUsername() //TODO replace with statsig variable
```

src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -50,7 +54,16 @@ function NameAndPicture({ navigation, route }: Props) {
...
+  const showAlternatePlaceholder = shallShowAlternatePlaceholder() //TODO repalce with statsig variable
```
cajubelt, line 58:
```suggestion
  const showAlternatePlaceholder = shallShowAlternatePlaceholder() //TODO replace with statsig variable
```

cajubelt (commented):
looks fine overall, just added minor comments. fwiw I think what you had before with taking the mock feature flag values as parameters was also fine.

cajubelt (approved):

src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -50,7 +54,16 @@ function NameAndPicture({ navigation, route }: Props) {
...
+  const skipUsername = shouldSkipUsername() //TODO repalce with statsig variable
+  const showAlternatePlaceholder = shouldShowAlternatePlaceholder() //TODO repalce with statsig variable
+  let placeholderUsername = t('fullNamePlaceholder')
+  if (showGuidedOnboarding) {
+    placeholderUsername = t('fullNameOrPseudonymPlaceholder')
+  }
+  if (showAlternatePlaceholder) {
+    placeholderUsername = 'MyCryptoAlterEgo' // not localized
+  }
+
```
jh2oman, line 66: Small nit: I think this would be cleaner inside of a `getPlaceholderUsername()` function. Similar to how we have a helper function for `goToNextScreen`

jh2oman (approved):

src/onboarding/registration/NameAndPicture.tsx
jh2oman: Then you can avoid using `let` and asigning `placeholderUsername` and just do `return t('fullNameOrPseudonymPlaceholder')` or `return 'MyCryptoAlterEgo'` instead.

src/onboarding/registration/MockedStatSigFeatureFlag.ts
```diff
@@ -0,0 +1,6 @@
+export const shouldShowAlternatePlaceholder = () => {
```
satish-ravi, line 1: Would this file be removed when this starts using statsig? If so, could we just inline this directly in NameAndPicture.tsx so it's easier to remove or at least add a TODO here?

src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -50,7 +54,16 @@ function NameAndPicture({ navigation, route }: Props) {
   const asyncKomenciReadiness = useAsyncKomenciReadiness()
   const showGuidedOnboarding = useSelector(showGuidedOnboardingSelector)
   const createAccountCopyTestType = useSelector(createAccountCopyTestTypeSelector)
-  const skipUsername = route.params?.skipUsername //TODO repalce with statsig variable
+  const skipUsername = shouldSkipUsername() //TODO repalce with statsig variable
+  const showAlternatePlaceholder = shouldShowAlternatePlaceholder() //TODO repalce with statsig variable
+  let placeholderUsername = t('fullNamePlaceholder')
+  if (showGuidedOnboarding) {
```
satish-ravi, line 60: Is this coming from an existing firebase experiment / remote config? Wonder if we could use a "usernamePlaceholder" parameter in statsig, which can then take values of "fullName", "fullNameOrPseudoym", or "alternate" (or just "cyrptoAlterEgo")?

XrayBravoGolf: Yes, but the `showGuidedOnboarding` has already started on Firebase and I'd prefer letting it run; there was discussion that ended in letting that finish before starting the `cyrptoAlterEgo` experiment on StatSig.

src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -50,7 +54,16 @@ function NameAndPicture({ navigation, route }: Props) {
   const asyncKomenciReadiness = useAsyncKomenciReadiness()
   const showGuidedOnboarding = useSelector(showGuidedOnboardingSelector)
   const createAccountCopyTestType = useSelector(createAccountCopyTestTypeSelector)
-  const skipUsername = route.params?.skipUsername //TODO repalce with statsig variable
+  const skipUsername = shouldSkipUsername() //TODO repalce with statsig variable
+  const showAlternatePlaceholder = shouldShowAlternatePlaceholder() //TODO repalce with statsig variable
```
jh2oman, line 58: Actually, one more thing. This variable should be an enum not a boolean. Its already defined in statsig https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/layers/name_and_picture_screen and in @anthonydaoud 's PR https://github.com/valora-inc/wallet/pull/2950

src/onboarding/registration/NameAndPicture.tsx
satish-ravi: yes that sounds good. I am just proposing that the new flag we add to statsig is an enum instead of a boolean. As @jh2oman mentioned, looks like we already have `nameType`. We should just be using that instead of a new boolean flag. Eventually the code can just become something like.
```js
const nameType = Statsig.get(....) // get param value from statsig
switch(nameType) {
  case 'first_and_last':
    placeholderUserName = t('fullName')
  case 'first_and_last_or_pseudonym':
    placeholderUserName = t('fullNameOrPseudonym')
  case 'crypto_alter_ego':
    placeholderUserName = 'MyCryptoAlterEgo'
}
```
One complication here is that we have to keep honoring the guided onboarding flag until that experiment is complete. Maybe we can add a 4th case temporarily as the default value of the param e.g., 
```js
  case 'guided_onboarding_exp':
    placeholderUserName =  showGuidedOnboarding ? t('fullNameOrPseudonym') :  t('fullName')
```
and once this experiment is over, depending on whichever side wins, we can make  `first_and_last` or `first_and_last_or_pseudonym` the default value

satish-ravi (changes requested):
RC'ing until the username flag is updated to use nameType

src/onboarding/registration/MockedStatSigFeatureFlag.ts
XrayBravoGolf: I'll add a TODO here.

src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -135,6 +147,22 @@ function NameAndPicture({ navigation, route }: Props) {
       }
     }
   }
+  const getUsernamePlaceholder = (nameType: OnboardingNameType) => {
+    if (showGuidedOnboarding) {
+      // Firebase trusted-guide onboarding experiment
+      nameType = OnboardingNameType.FirstAndLastOrPseudonym
+    }
```
satish-ravi, line 154: If the `true` side of guided onboarding wins and we roll it out, this would mean that we'll never been using the statsig flag for future experiments right? You'd need a workaround like what I suggested in https://github.com/valora-inc/wallet/pull/2949#discussion_r988144496

XrayBravoGolf: Right, that is indeed a potential bug.

src/onboarding/registration/NameAndPicture.test.tsx
```diff
@@ -243,6 +245,26 @@ describe('NameAndPictureScreen', () => {
     expect(AccountActions.setName).not.toHaveBeenCalled()
   })
 
+  it('shows alternate placeholder username for experimental group', () => {
```
satish-ravi, line 248: can you also add a test for `placeholder` being returned from statsig?

satish-ravi (approved):

PR #2950, opened 2022-10-04
feat(analytics): use Statsig skip button param in onboarding

anthonydaoud:
### Description

This PR makes use of the Statsig experiment params in the `NameAndPicture` screen. Currently, Statsig is initialized but unused in the wallet. After merging this PR, the `NameAndPicture` screen will show a skip button only if the user was bucketed into that trial group by Statsig.

### Other changes

None

### Tested

Unit tests.

### How others should test

Verify skip button is shown sometimes on the Name Screen when onboarding. Also, test that airplane mode doesn't lead to issues.

### Related issues

- [ACT-436](https://linear.app/valora/issue/ACT-436/skip-button-on-name-screen)

### Backwards compatibility

N/A


src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -73,17 +98,22 @@ function NameAndPicture({ navigation, route }: Props) {
...
+            onPress={goToNextScreen}
             titleStyle={{ color: colors.goldDark }}
           />
         ),
     })
   }, [navigation, choseToRestoreAccount, step, totalSteps, nameInput])
 
   const goToNextScreen = () => {
+    try {
+      Statsig.logEvent(StatsigEvents.ONBOARDING_NAME_STEP_COMPLETE)
```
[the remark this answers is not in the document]
XrayBravoGolf: Wouldn't this also hold true in our existing onboarding analytics? (Or is the same issue mitigated in some way already in current analytics?)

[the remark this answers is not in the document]
jh2oman: I don't think its much to worry about. Most experiment platforms handle this issue well. See the event_dau metric that Statsig auto generates for us https://docs.statsig.com/metrics/metrics-from-events

src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -73,17 +98,22 @@ function NameAndPicture({ navigation, route }: Props) {
...
-            onPress={onPressSkip}
+            onPress={goToNextScreen}
             titleStyle={{ color: colors.goldDark }}
           />
         ),
     })
   }, [navigation, choseToRestoreAccount, step, totalSteps, nameInput])
 
   const goToNextScreen = () => {
+    try {
```
jh2oman, line 112: Looking through the statsig sdk https://github.com/statsig-io/node-js-server-sdk it appears that the only time this throws is when statsig is uninitialized. Can we guarantee that its initialized by this point in the code & not need the try/catch?

XrayBravoGolf (commented):
Do we also want analytics for the username being skipped, something like `ValoraAnalytics.track(OnboardingEvents.username_skipped)`?

src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -30,12 +32,36 @@ import { default as useSelector, default as useTypedSelector } from 'src/redux/u
 import colors from 'src/styles/colors'
 import fontStyles from 'src/styles/fonts'
 import { saveProfilePicture } from 'src/utils/image'
+import Logger from 'src/utils/Logger'
 import { useAsyncKomenciReadiness } from 'src/verify/hooks'
+import { Statsig } from 'statsig-react-native'
 
 type Props = StackScreenProps<StackParamList, Screens.NameAndPicture>
 
+const getExperimentParams = () => {
```
jh2oman, line 41: Side comment: This seems ripe for a helper function to share code for getting params from experiment layers. Not too concerned if its done now or later when we have other statsig experiments.

src/onboarding/registration/NameAndPicture.tsx
jh2oman: If not, this also seems like it could use a helper function so we don't need to wrap every single metric log in a try/catch

src/onboarding/registration/NameAndPicture.test.tsx
```diff
@@ -183,20 +190,20 @@ describe('NameAndPictureScreen', () => {
     expect(getByText('nameAndPicGuideCopyContent')).toBeTruthy()
   })
   it('does not render skip button when configured so', () => {
-    // TODO replace route param with mock Statsig flag
+    mockStatsigGet = jest.fn().mockReturnValue(false)
```
cajubelt, line 193: I'm very surprised this works, but I still don't think we should do it. see here: https://github.com/valora-inc/wallet/pull/2951

src/onboarding/registration/NameAndPicture.tsx
anthonydaoud: I think we'd be getting errors [client sdk](https://github.com/statsig-io/js-client) but your point still stands. I think we can guarantee initialize is called by that point, but they may be offline (e.g. on airplane mode) so Statsig won't connect. Also I didn't realize their code was open source, so I had assumed that this failing to connect could lead to other errors. Looking at it now I don't think that's the case, so I can remove the try/catch

src/onboarding/registration/NameAndPicture.tsx
anthonydaoud: Yeah agreed, I lean towards doing that when we have other experiments.

src/onboarding/registration/NameAndPicture.tsx
anthonydaoud: Removing the try-catch leads to e2e errors, so I've added it back in. I also forgot that I wrote an [issue](https://linear.app/valora/issue/ACT-467/segmentstatsig-integration) to log events via Segment rather than Statsig, so I think this is just a temporary setup anyhow.

PR #2954, opened 2022-10-05
Run E2E tests against updated runner group

jacob-israel-turner:
### Description

I split `e2e-ci.yml` into three workflows:
1. A re-usable workflow to run Android E2E tests (`e2e-android.yml`).
2. A workflow that runs E2E tests on one device on iOS and one device on Android, whenever a PR into `main` is opened. It utilizes the re-usable Android workflow.
3. A workflow that runs E2E tests on one device on iOS and on multiple devices on Android. This workflow is called whenever a commit is pushed to `main`, the workflow is called manually from the Github UI, and once per day on a schedule. It also utilizes the re-usable Android workflow

Do note, the iOS side of things is still a WIP. My next PR will break out the copied iOS steps into their own reusable workflow, like Android, and will allow running on multiple iOS versions.

### Other changes

N/A

### Tested

I've run these tests many times via Github Actions.

### How others should test

N/A

### Related issues

RET-275

### Backwards compatibility

N/A

### DIFF
--- a/.github/workflows/e2e-android.yml
+++ b/.github/workflows/e2e-android.yml
@@ -57,11 +57,11 @@ jobs:
         if: failure()
         uses: actions/upload-artifact@v2
         with:
-          name: android-e2e-artifact
+          name: android-${{ inputs.android-api-level }}-e2e-artifact
           path: e2e/artifacts
       - name: 'Upload Android E2E HTML Report'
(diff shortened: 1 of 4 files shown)


jacob-israel-turner:
These changed were merged via a different PR.

PR #2955, opened 2022-10-06
feat: use centralised phone number lookup in send flow

kathaypacific:
### Description

As the title, this PR updates only the send flow to work with CPV.

If no addresses are found with the lookupPhoneNumbers endpoint, then the decentralised mapping is used to maintain backwards compatibility while users are still migrating.

### Other changes

N/A

### Tested

Manually, unit tests

### How others should test

In Alfajores, sending to phone numbers should work as before. Including the secure send flow when sending to a phone number that has multiple addresses

### Related issues

Fixes RET-389

### Backwards compatibility

Yes

### DIFF
--- a/src/identity/contactMapping.ts
+++ b/src/identity/contactMapping.ts
@@ -221,7 +221,7 @@ function* getAccountAddresses(e164Number: string) {
   return yield call(filterNonVerifiedAddresses, accountAddresses, phoneHash)
 }
 
-export function* fetchWalletAddressesDecentralised(e164Number: string) {
+export function* fetchWalletAddressesDecentralized(e164Number: string) {
   const contractKit = yield call(getContractKit)
   const accountsWrapper: AccountsWrapper = yield call([
     contractKit.contracts,
@@ -253,9 +253,14 @@ export function* fetchWalletAddressesDecentralised(e164Number: string) {
 }
 
 function* fetchWalletAddresses(e164Number: string) {
+  const addressesFromDecentralisedMapping: string[] = yield call(
+    fetchWalletAddressesDecentralized,
+    e164Number
+  )
   const centralPhoneVerificationEnabled = yield select(centralPhoneVerificationEnabledSelector)
+
   if (!centralPhoneVerificationEnabled) {
-    return yield call(fetchWalletAddressesDecentralised, e164Number)
+    return addressesFromDecentralisedMapping
   }
 
   try {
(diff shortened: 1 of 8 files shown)


src/identity/contactMapping.ts
```diff
@@ -195,7 +200,7 @@ export function* fetchAddressesAndValidateSaga({
     yield put(endFetchingAddresses(e164Number, true))
     ValoraAnalytics.track(IdentityEvents.phone_number_lookup_complete)
   } catch (error) {
-    Logger.error(TAG + '@fetchAddressesAndValidate', `Error fetching addresses`, error)
```
kathaypacific, line 198: i assume it's okay to downgrade to logger.debug here, since we will be able to monitor the lookupPhoneNumber service and it's not a unrecoverable error

src/send/ValidateRecipientAccount.tsx
```diff
@@ -263,11 +263,13 @@ export class ValidateRecipientAccount extends React.Component<Props, State> {
...
 
   render = () => {
-    const { t, recipient, error } = this.props
-    const { singleDigitInputValueArr } = this.state
+    const { t, recipient, error, addressValidationType } = this.props
+    const { singleDigitInputValueArr, inputValue } = this.state
     const displayName = getDisplayName(recipient, t)
     const isFilled =
-      singleDigitInputValueArr.filter((entry) => /[a-f0-9]/gi.test(entry)).length === 4
+      addressValidationType === AddressValidationType.FULL
```
kathaypacific, line 270: the full address validation flow was broken, since that flow doesn't touch `singleDigitInputValueArr`, this `isFilled` was always false and i could not press the submit button

jacob-israel-turner (approved):

src/identity/contactMapping.ts
```diff
@@ -247,6 +252,46 @@ function* fetchWalletAddresses(e164Number: string) {
...
+      body: JSON.stringify({
+        phoneNumber: e164Number,
+        clientPlatform: Platform.OS,
+        clientVersion: DeviceInfo.getVersion(),
+      }),
+    })
+
+    if (response.ok) {
+      const { addresses } = yield call([response, 'json'])
+      if (addresses.length === 0) {
```
dievazqu, line 280: I'm wondering if we should call the decentralized flow even if we have a response from the CPV...
A user could be verified in both systems with different addresses and the same phone number. Maybe it's a weird case that probably won't happen until the migration is done.

dievazqu (approved):
LGTM! Just consider the case I mentioned, but feel free to ignore it.

src/identity/contactMapping.ts
kathaypacific: ah yes, good point. i updated to always check the decentralised mapping. discussing with @jeanregisser , it makes sense to plan to maintain backwards compatibility for the foreseeable future :) thanks!

src/send/ValidateRecipientAccount.tsx
jeanregisser: oh good catch! 👍 
Should we cover that with a test? 😅

src/identity/contactMapping.ts
```diff
@@ -247,6 +252,54 @@ function* fetchWalletAddresses(e164Number: string) {
...
+
+  if (!centralPhoneVerificationEnabled) {
+    return addressesFromDecentralizedMapping
+  }
+
+  try {
+    const address = yield select(walletAddressSelector)
+    const signedMessage = yield call(retrieveSignedMessage)
+
+    const response: Response = yield call(fetch, networkConfig.lookupPhoneNumberUrl, {
```
jeanregisser, line 270: Do you think it's worth making the decentralized and centralized fetch in parallel?
Not sure how quick they are today, but could be interesting.

jeanregisser (approved):
Amazing! 📞 🗺️

PR #2957, opened 2022-10-06
fix: swaps to celo

MuckT:
### Description

Uses a workaround Celo Address to enable swaps to CELO using 0xAPI. This should only be merged if the workaround will continue working after the fix from the 0x team.

### Other changes

N/A

### Tested

Tested locally.

### How others should test

Attempt to swap from any token to CELO, swap should be successful.

### Related issues

N/A

### Backwards compatibility

Yes


MuckT:
Closed in favor of adding to existing swap PR based on conversations with 0x. Included in https://github.com/valora-inc/wallet/pull/2891/commits/82bacaa906bd554a2979b1b62e3f23184fd5ff33.

PR #2960, opened 2022-10-07
chore: augment request body of verifyPhoneNumber with extra params

kathaypacific:
### Description

Add client bundle id for https://valora-app.slack.com/archives/C044B8ULVC0/p1665055690215779
Add DEK public key for comment encryption

### Other changes

N/A

### Tested

Updated tests, test verification manually

### How others should test

N/A

### Related issues

Relates RET-442

### Backwards compatibility

N/A

### DIFF
--- a/src/verify/hooks.ts
+++ b/src/verify/hooks.ts
@@ -52,8 +52,10 @@ export function useVerifyPhoneNumber(phoneNumber: string, countryCallingCode: st
 
   const dispatch = useDispatch()
   const address = useSelector(walletAddressSelector)
-  const privateDataKey = useSelector(dataEncryptionKeySelector)
-  const publicDataKey = privateDataKey ? compressedPubKey(hexToBuffer(privateDataKey)) : null
+  const privateDataEncryptionKey = useSelector(dataEncryptionKeySelector)
+  const publicDataEncryptionKey = privateDataEncryptionKey
+    ? compressedPubKey(hexToBuffer(privateDataEncryptionKey))
+    : null
 
   const [verificationStatus, setVerificationStatus] = useState(PhoneNumberVerificationStatus.NONE)
   const [verificationId, setVerificationId] = useState('')
@@ -112,7 +114,7 @@ export function useVerifyPhoneNumber(phoneNumber: string, countryCallingCode: st
           clientPlatform: Platform.OS,
           clientVersion: DeviceInfo.getVersion(),
           clientBundleId: DeviceInfo.getBundleId(),
-          publicDataKey,
+          publicDataEncryptionKey,
         }),
       })
       if (response.ok) {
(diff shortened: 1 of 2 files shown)


src/verify/hooks.ts
```diff
@@ -50,6 +52,8 @@ export function useVerifyPhoneNumber(phoneNumber: string, countryCallingCode: st
 
   const dispatch = useDispatch()
   const address = useSelector(walletAddressSelector)
+  const privateDataKey = useSelector(dataEncryptionKeySelector)
+  const publicDataKey = privateDataKey ? compressedPubKey(hexToBuffer(privateDataKey)) : null
```
jeanregisser, line 56: I'm making it required for this call on the backend.
At this point I think it's always present.

src/verify/hooks.ts
```diff
@@ -107,6 +111,8 @@ export function useVerifyPhoneNumber(phoneNumber: string, countryCallingCode: st
           phoneNumber,
           clientPlatform: Platform.OS,
           clientVersion: DeviceInfo.getVersion(),
+          clientBundleId: DeviceInfo.getBundleId(),
+          publicDataKey,
```
jeanregisser, line 115: nit: I'm using this name on the backend:
```suggestion
          publicDataEncryptionKey,
```

jeanregisser (approved):
Awesome! 🚀 

Already approved but needs small tweaks.

src/verify/hooks.ts
kathaypacific: does `null` count as a value, since it's not undefined? 😅 i think it will be always be available at this point but if for whatever reason it isn't, then null is still okay right? it doesn't affect too much functionality if the key is null, seems like it shouldn't be a blocker for going through phone verification?

kathaypacific: yeah i guess that makes sense. i'm not totally convinced that the DEK is important enough to block verification, but at the same time if the DEK hasn't been generated then that is probably a sign that there's something wrong with the account. i updated the hook to fail on the verification if there is no DEK

src/verify/hooks.ts
```diff
@@ -50,6 +52,10 @@ export function useVerifyPhoneNumber(phoneNumber: string, countryCallingCode: st
 
   const dispatch = useDispatch()
   const address = useSelector(walletAddressSelector)
+  const privateDataEncryptionKey = useSelector(dataEncryptionKeySelector)
+  const publicDataEncryptionKey = privateDataEncryptionKey
+    ? compressedPubKey(hexToBuffer(privateDataEncryptionKey))
+    : null
```
jeanregisser, line 58: nit: we should do this only when needed before calling the API instead of for each call of this hook.

jeanregisser (approved):

PR #2958, opened 2022-10-06
feat: add username generator function

anthonydaoud:
### Description

For an [onboarding experiment](https://www.notion.so/valora-inc/Onboarding-Name-Step-4b6ef3485ac64b0caeb87147887f68a5), we need to be able to generate random, fun usernames. This PR implements part of the required functionality, adding a list of about 500 adjectives and nouns each that will be used to generate the names.

### Other changes

N/A

### Tested

Added unit test. Will be more throughly tested in a subsequent PR.

### How others should test

N/A

### Related issues

- [ACT-435](https://linear.app/valora/issue/ACT-435/username-generator-helper-function)

### Backwards compatibility

N/A

PR #2965, opened 2022-10-10
feat: add username generator button to onboarding

XrayBravoGolf:
### Description

During user onboarding, a button is added to allow an auto-generated username to be provided to the user. Pressing button replaces what is in the input field with the generated name.

The generator button is controlled by StatSig experiment `Name Screen Experiment
`; button is shown only if `nameType` in StatSig equals `autogenerator`.

This PR adds the button and functionality, sans the generator function in [separate PR](https://github.com/valora-inc/wallet/pull/2958).

### Other changes

Moved generator function and test to separate file.

### Tested

Automated tests written; assumes pressing the button again puts a different username in the input box.
Manual testing verified placeholder and generator button works.

### How others should test

N/A

### Related issues

- Fixes #[ACT-439]

### Backwards compatibility

Compatible because no existing behavior are changed for onboarding.


XrayBravoGolf:
Simulator Screen Shot - iPhone 13 - 2022-10-10 at 02 44 22
Simulator Screen Shot - iPhone 13 - 2022-10-10 at 02 35 00
Simulator Screen Shot - iPhone 13 - 2022-10-10 at 02 34 27
Simulator Screen Shot - iPhone 13 - 2022-10-08 at 17 51 36

src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -192,6 +199,11 @@ function NameAndPicture({ navigation, route }: Props) {
     }
   }
 
+  const onPressGenerateUsername = () => {
+    //TODO: Obtain forbidden words from Firebase Remote Config
```
cajubelt, line 203: I'm seeing that https://github.com/valora-inc/wallet/pull/2974 is approved, so this will be possible very soon

src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -91,21 +91,28 @@ function NameAndPicture({ navigation, route }: Props) {
...
   const showGuidedOnboarding = useSelector(showGuidedOnboardingSelector)
   const createAccountCopyTestType = useSelector(createAccountCopyTestTypeSelector)
+  const showNameGeneratorButton = nameType === OnboardingNameType.AutoGen
 
   useLayoutEffect(() => {
     navigation.setOptions({
       headerTitle: () => {
         let pageTitleTranslationKey
-        if (showGuidedOnboarding) {
+        pageTitleTranslationKey = choseToRestoreAccount
```
cajubelt, line 100: if `showGuidedOnboarding` is true and `choseToRestoreAccount` is also true, the new code chooses a different page title. Is that intended? Seems to change the behavior of a different experiment..

src/onboarding/registration/NameAndPicture.test.tsx
```diff
@@ -245,6 +245,25 @@ describe('NameAndPictureScreen', () => {
     expect(AccountActions.setName).not.toHaveBeenCalled()
   })
 
+  it('shows generator button and generator button puts random username in', () => {
+    mockStatsigGet.mockReturnValueOnce(false).mockReturnValueOnce('autogenerator')
```
cajubelt, line 249: as you pointed out, this is very implementation dependent. Why not use `mockImplementation` and make the output different depending on the input? 

e.g. if you're getting the nameType feature flag, give 'autogenerator', but if you're getting some boolean-valued feature flag, return 'false'

src/onboarding/registration/NameAndPicture.test.tsx
```diff
@@ -245,6 +245,25 @@ describe('NameAndPictureScreen', () => {
...
+    mockRandom.mockReturnValue(0)
+    expect(getByText('generateUsername')).toBeTruthy()
+    const nameField = getByTestId('NameEntry')
+    fireEvent.press(getByText('generateUsername'))
+    const text = nameField.props.value
+    mockRandom.mockReturnValue(0.9)
+    expect(text).toBeTruthy()
+    fireEvent.press(getByText('generateUsername'))
+    // different name on second press
+    expect(nameField.props.value).not.toEqual(text)
```
cajubelt, line 264: why not mock the return value of the name generator and make sure the mocked value from the second invocation gets used here?

cajubelt: mocking the random function seems like it makes this an integration test, testing the mock name generator helper function and the usage of it at the same time. I encourage you to prefer unit tests over integration tests-- they're easier to maintain and debug, and give you finer grained control over what is tested

src/onboarding/registration/NameAndPicture.tsx
anthonydaoud: Yeah just merged it.

src/onboarding/registration/NameAndPicture.test.tsx
XrayBravoGolf: I absolutely agree that a test that just mocks what the user does instead of poking inside the implementation is better, which is what I wrote in the beginning, but the generator function would not work;
 I checked with Anthony and the way the random function is mocked made that difficult.

src/onboarding/registration/NameAndPicture.tsx
XrayBravoGolf: It is intended; it is indeed a place of confusion;
The trusted guide covers title for both restore and new, and this implementation is consistent with that of the experiment already started

XrayBravoGolf (commented):
Logic related to username generator were unchanged but moved to a separate file and its test file; credit goes to @anthonydaoud.

cajubelt (approved):

src/onboarding/registration/NameAndPicture.tsx
cajubelt: i misread the boolean logic

PR #2967, opened 2022-10-10
fix: use view instead of nested SafeAreaView to prevent extra top space

MuckT:
### Description

Removes the nested SafeAreaView and replaces it with a View to avoid adding an extra margin on top of the nested view.
Removes vertical padding on View element.

#### Video

https://user-images.githubusercontent.com/26950305/194951320-9dbb00a8-0524-4901-952e-529e4b903c85.mov

### Other changes

N/A

### Tested

Tested locally on Android

### How others should test

1. On Alfajores Send to attempt to send to `12057368924`.
2. Enter amount and tap 'Confirm Account Address'
3. Observe padding at the top matches other screens.
4. Observe button not hidden behind keyboard.

### Related issues

N/A

### Backwards compatibility

Yes


jacob-israel-turner (approved):
🎉

PR #2968, opened 2022-10-11
Run E2E Suite on Multiple Android Versions

jacob-israel-turner:
### Description

I split `e2e-ci.yml` into three workflows:
1. A re-usable workflow to run Android E2E tests (`e2e-android.yml`).
2. A workflow that runs E2E tests on one device on iOS and one device on Android, whenever a PR into `main` is opened. It utilizes the re-usable Android workflow.
3. A workflow that runs E2E tests on one device on iOS and on multiple devices on Android. This workflow is called whenever a commit is pushed to `main`, the workflow is called manually from the Github UI, and once per day on a schedule. It also utilizes the re-usable Android workflow

Do note, the iOS side of things is still a WIP. My next PR will break out the copied iOS steps into their own reusable workflow, like Android, and will allow running on multiple iOS versions. I'm also looking into getting it working on even more Android versions, but we had some issues with versions <= 28 and >= 32.

### Other changes

N/A

### Tested

I've run these tests many times via Github Actions.

### How others should test

N/A

### Related issues

RET-275

### Backwards compatibility

N/A

### DIFF
--- a/.github/workflows/e2e-android.yml
+++ b/.github/workflows/e2e-android.yml
@@ -57,11 +57,11 @@ jobs:
         if: failure()
         uses: actions/upload-artifact@v2
         with:
-          name: android-e2e-artifact
+          name: android-${{ inputs.android-api-level }}-e2e-artifact
           path: e2e/artifacts
       - name: 'Upload Android E2E HTML Report'
         if: always()
         uses: actions/upload-artifact@v2
         with:
-          name: android-e2e-test-report
+          name: android-${{ inputs.android-api-level }}-e2e-test-report
           path: e2e/test-results
--- a/.github/workflows/e2e-android.yml
+++ b/.github/workflows/e2e-android.yml
@@ -1,0 +1,67 @@
+name: E2E - Android
+on:
+  workflow_call:
+    inputs:
+      android-api-level:
+        required: true
+        type: number
+
+jobs:
+  android:
+    env:
+      BASH_ENV: ~/.profile
+    name: Android (SDK ${{ inputs.android-api-level }})
+    runs-on: android-e2e-group
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16'
+          check-latest: true
+      - name: Install package dependencies
+        run: yarn || yarn --network-concurrency 1
+      - name: Fail if someone forgot to commit "yarn.lock"
+        run: git diff --exit-code
+      - name: Check E2E wallet balance
+        run: NODE_OPTIONS='--unhandled-rejections=strict' yarn ts-node ./e2e/scripts/check-e2e-wallet-balance.ts
+      - name: Create Detox Build
+        run: |
+          export CELO_TEST_CONFIG=e2e
+          export ANDROID_SDK_ROOT=$HOME/android-tools
+          yarn detox build -c android.release
+      - name: Run Detox
+        run: >
+          export ANDROID_SDK_ROOT=$HOME/android-tools &&
+          export PATH=$ANDROID_SDK_ROOT/platform-tools:$PATH &&
+          yarn detox test
+          --device-name Pixel_API_${{ inputs.android-api-level }}_AOSP_x86_64
+          --configuration android.release
+          --artifacts-location e2e/artifacts
+          --take-screenshots=failing
+          --record-videos=failing
+          --record-logs=failing
+          --loglevel info
+          --debug-synchronization 10000
+          --workers 3
+          --headless
+          --retries 3
+          --device-boot-args="-snapshot ci_boot"
+        timeout-minutes: 45
+      - name: Publish Android JUnit Report
+        if: always()
+        uses: mikepenz/action-junit-report@v2
+        with:
+          check_name: Android e2e Test Report
+          report_paths: 'e2e/test-results/junit.xml'
+      - name: 'Upload Android E2E Artifacts'
+        if: failure()
+        uses: actions/upload-artifact@v2
+        with:
+          name: android-e2e-artifact
+          path: e2e/artifacts
+      - name: 'Upload Android E2E HTML Report'
+        if: always()
+        uses: actions/upload-artifact@v2
+        with:
+          name: android-e2e-test-report
+          path: e2e/test-results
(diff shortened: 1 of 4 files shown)


jacob-israel-turner:
We'll need to change the branch protection rules to require Android (30) to pass, instead of Android. I'll do that after approval, right before I merge, so we don't accidentally let other PRs slip by.

.github/workflows/e2e-main.yml
```diff
@@ -1,11 +1,8 @@
-name: E2E
+name: E2E - Main
 on:
   push:
     branches:
```
kathaypacific, line 4: do we also want this job to be run after every PR merge? if not, should we remove the on / push?

kathaypacific (commented):
makes sense, super nice to have these scripts being reusable! did we consider running the same jobs on main and PRs, and only running against all regression devices on release branches? i'm being pessimistic but the more devices we run the tests on, the more likely we'll get failing tests from test flake and need to spend time checking what is a genuine failure and what isn't...if this runs every night it might create some overhead to check the test results. do you see this tradeoff being worthwhile?

.github/workflows/e2e-android.yml
```diff
@@ -0,0 +1,69 @@
...
+          node-version: '16'
+          check-latest: true
+      - name: Install package dependencies
+        run: yarn || yarn --network-concurrency 1
+      - name: Fail if someone forgot to commit "yarn.lock"
+        run: git diff --exit-code
+      - name: Check E2E wallet balance
+        run: NODE_OPTIONS='--unhandled-rejections=strict' yarn ts-node ./e2e/scripts/check-e2e-wallet-balance.ts
+      - name: Copy Emulator Config
+        run: mkdir -p ~/.android/avd/Pixel_API_${{ inputs.android-api-level }}_AOSP_x86_64.avd/ && cp e2e/conf/ci_avd_conf.ini ~/.android/avd/Pixel_API_${{ inputs.android-api-level }}_AOSP_x86_64.avd/config.ini
```
jeanregisser, line 28: Why is this step necessary?

jeanregisser (approved):
Awesome work! 🎉 

I wonder if later on we can enable running on multiple Android versions for all PRs instead of main. This could possibly help catch issues earlier.
Or maybe have the multiple runs done only once the PR has been approved or tagged with a specific label, before merging. 🤔

.github/workflows/e2e-android.yml
MuckT: This sets up the ini config with some specific values for the emulators. We could move this over to the infrastructure repo to avoid running it on every PR.

MuckT: We'll need to adjust the ci_avd_conf.ini or create slightly different one for each device as there are some emulator specific values such as:
```
image.sysdir.1=system-images/android-29/default/x86_64/
sdcard.path=/Users/distiller/.android/avd/Pixel_API_29_AOSP_x86_64.avd/sdcard.img
```

As I see it can either create multiple ci_avd_conf.ini files or have the setup in infrastructure repo.

MuckT: We could also edit the `ci_avd.conf.ini` to exclude these items and see if that will work.

.github/workflows/e2e-pr.yml
```diff
@@ -0,0 +1,99 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref }}
+  cancel-in-progress: true
+
+jobs:
+  android:
+    name: Android
+    uses: ./.github/workflows/e2e-android.yml
+    with:
+      android-api-level: 30
```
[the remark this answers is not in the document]
jacob-israel-turner: We _can_ run it against 31, but for some reason tests against 31 run way slower. Something goes wrong when initializing tests the first time through, and the whole suite fails. Instead of taking ~15 minutes, tests instead take ~35 minutes. I'm hoping to look into this in a follow-up PR.

.github/workflows/e2e-main.yml
jacob-israel-turner: I discussed with @MuckT and came up with the following - he may have other insights as well.

Run against multiple OS versions when:
 - A commit is pushed to main (so, _after_ a PR is merged)
 - When called from the Github UI
 - On a schedule, once per day

Run against a single OS version when:
 - A PR is opened, before it can be merged

This ensures that PRs can pass checks relatively quickly, but we still test against the full suite of OS versions on `main`. We will learn about potential OS version incompatibilities as we merge into main, without having to wait for a full regression test. We can potentially learn about service outages/breaking changes with the daily run as well.

I feel like what is in this strategy is a pretty good balance. In general if a PR has passed E2E tests against a specific OS version, we're probably good to move forward. The daily/push-to-main tests are just additional checks against all the other OS versions. It would be nice to know about OS-specific issues before we cut a release and test against it - and at the same time, if we find tests to be flaky, it's ok to just re-run tests/ignore a single failed run on `main`.

I'm happy to make changes if you see any issues or have suggestions for improvements. Updating workflow files pretty straightforward.

jacob-israel-turner:
> Awesome work! 🎉
> 
> I wonder if later on we can enable running on multiple Android versions for all PRs instead of main. This could possibly help catch issues earlier. Or maybe have the multiple runs done only once the PR has been approved or tagged with a specific label, before merging. 🤔

Yeah, this is totally plausible. The big reason we didn't do that on PRs, is for fear of locking down all the Mac Stadium boxes by a handful of PRs. Looking at our PR/action history, it seems like this wouldn't be an issue _most_ of the time. But at busy times we could be bottlenecked, which could be especially frustrating if someone is trying to hurry and get a PR into a release.

It would be ideal to provision more (and, ideally, beefier) Mac Stadium boxes, so we can run each PR against every OS version desired.

.github/workflows/e2e-android.yml
jacob-israel-turner: You can see the `.ini` file we have committed, [here](https://github.com/valora-inc/wallet/blob/2b28b691640c52bdeeb1b49a7751c5bdf3b764fe/e2e/conf/ci_avd_conf.ini).  This step is just copying that file into the AVD directory, so it's used when the emulator is spun up.

jacob-israel-turner: Turns out this wasn't the right fix. The `.ini` file that's being copied here may be ignored, as the snapshot is based off a different `.ini` file.

To be sure, we're going to do the `.ini` copying in the provisioning step instead.

PR #2969, opened 2022-10-11
feat: handle number already verified for CPV

kathaypacific:
### Description

Ensure that users who are already verified can restore their accounts and verify again. We can consume the new phone number lookup endpoint, but this PR achieves feature parity for now.

Video: https://linear.app/valora/issue/RET-389#comment-a119f794

### Other changes

N/A

### Tested

Manually, unit test

### How others should test

Verify a phone number against an account. Restore the account, and verify the same phone number. 

### Related issues

- Fixes RET-389
- 
### Backwards compatibility

Yes


jeanregisser (approved):
🚀

PR #2973, opened 2022-10-11
feat(fiatconnect): Cache quote parameters when posting KYC to IHL

jophish:
### Description
For ACT-473. Caches the parameters used for the current quote when posting KYC to IHL. This sets us up for later on when jumping to the review screen when the user receives a push/homecard alerting them that their KYC has passed, since we'll need to refetch a quote with the same parameters as the quote used when submitting KYC information.

### Other changes

N/A

### Tested

Unit tested.

### How others should test

N/A currently; new redux state is unused.

### Related issues

- Fixes ACT-473.

### Backwards compatibility

N/A

### DIFF
--- a/src/fiatconnect/saga.ts
+++ b/src/fiatconnect/saga.ts
@@ -59,6 +59,7 @@ import {
   submitFiatAccount,
   submitFiatAccountCompleted,
   submitFiatAccountKycApproved,
+  cacheQuoteParams,
 } from 'src/fiatconnect/slice'
 import FiatConnectQuote from 'src/fiatExchanges/quotes/FiatConnectQuote'
 import { normalizeFiatConnectQuotes } from 'src/fiatExchanges/quotes/normalizeQuotes'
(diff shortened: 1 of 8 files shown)


src/fiatconnect/slice.ts
```diff
@@ -40,6 +49,11 @@ export interface State {
   selectFiatConnectQuoteLoading: boolean
   sendingFiatAccountStatus: SendingFiatAccountStatus
   kycTryAgainLoading: boolean
+  cachedQuoteParams: {
+    [providerId: string]: {
+      [kycSchema: string]: CachedQuoteParams
```
jophish, line 54: I'm purposely using `[kycSchema: string]` here rather than `[kycSchema in KycSchema]`. By specifying the index key to be a member of the `KycSchema` enum, the root state schema will need to updated whenever KycSchema` changes, which seems like an unintended side effect. Whenever we look through `cachedQuoteParams` to see if there's a match when a user presses the KYC Approved homecard/push notification, we'll already have access to the `kycSchema` in question, so we would never need to _use_ the `kycSchema` key,  which would require us to include a type assertion in order to use it effectively.

src/fiatconnect/saga.ts
```diff
@@ -498,11 +499,26 @@ export function* handleSelectFiatConnectQuote({
       switch (fiatConnectKycStatus) {
         case FiatConnectKycStatus.KycNotCreated:
           if (getKycStatusResponse.persona === PersonaKycStatus.Approved) {
-            // If user has Persona KYC on file, just submit it and continue to account management
+            // If user has Persona KYC on file, just submit it and continue to account management.
+            // We also need to save a user's quote parameters so we can re-fetch if KYC takes a long
```
jh2oman, line 503: nit: can you move this comment down to right above the put(cacheQuoteParams)?

src/fiatconnect/slice.ts
jh2oman: That seems fine to me, Root state is not kind to schemas that are expected to change often

jh2oman (approved):

PR #2979, opened 2022-10-12
feat(fiatconnect): get account number regex from remote config

satish-ravi:
### Description
Nigeria:

https://user-images.githubusercontent.com/5062591/195278263-2cb2cb5c-2daf-426a-87ee-6f1262d2ef1c.mp4

Other countries:

https://user-images.githubusercontent.com/5062591/195278301-2e287d62-21f1-4ae2-940d-c7119659dcb7.mp4

### Other changes

N/A

### Tested

Unit tests, manual

### How others should test

1. Navigate to "Add & Withdraw" on a device that hasn't previously withdrawn with test provider
2. Select Withdraw Funds
3. Select any asset
4. Enter any amount ending in even number (e.g., 2, 0.4, etc)
5. Select Bank account option with Valora icon (test provider) and click Continue
6. Ensure account number field allows only digits. If country is Nigeria, account number field should only accept exactly 10 digits

### Related issues

- Fixes ACT-357

### Backwards compatibility

Yes

### DIFF
--- a/locales/base/translation.json
+++ b/locales/base/translation.json
@@ -1454,8 +1454,10 @@
     "swapFee": "{{appName}} Fee",
     "swapFeeTitle": "What is the {{appName}} Fee?",
     "swapFeeBody": "The {{appName}} fee is the fee that you pay to swap tokens on {{appName}}. \n\n The fee is {{swapFee}}% of the amount you are exchanging. \n\n {{appName}} fees are free for the next few months.",
+    "swapFeeBodyFree": "The {{appName}} fee is the fee that you pay to swap tokens on {{appName}}. {{appName}} fees are free for the next few months.",
     "estimatedAmountTitle": "Estimated Amount",
-    "estimatedAmountBody": "Conversions take place on a third party decentralized exchange that finds you the best rate for your swap. \n\n Your trade will be completed at a price within {{slippagePercent}}% of the estimate or the transaction will be cancelled & funds returned to your wallet."
+    "estimatedAmountBody": "Conversions take place on a third party decentralized exchange that finds you the best rate for your swap. \n\n Your trade will be completed at a price within {{slippagePercent}}% of the estimate or the transaction will be cancelled & funds returned to your wallet.",
+    "free": "Free"
   },
   "SwapExecuteScreen": {
     "swapPending": "Swap Pending",
--- a/locales/base/translation.json
+++ b/locales/base/translation.json
@@ -602,6 +602,7 @@
   "fullNameOrPseudonymPlaceholder": "First and last name or pseudonym",
   "nameAndPicGuideCopyTitle": "What’s your name?",
   "nameAndPicGuideCopyContent": "We’d like to know what to call you",
+  "generateUsername": "Generate a name",
   "importIt": "Restore Wallet",
   "important": "Important",
   "confirmPin": {
--- a/locales/base/translation.json
+++ b/locales/base/translation.json
@@ -1419,8 +1419,9 @@
     },
     "accountNumber": {
       "label": "Account Number",
-      "placeholderText": "Enter your 10 digit account number",
-      "errorMessage": "Account number should be 10 digits long and only contain numbers"
+      "placeholderText": "Enter Account Number",
+      "errorMessageDigitLength": "Account number must be {{length}} digits",
+      "errorMessageDigit": "Account number must contain only digits"
     }
   },
   "webView": {
--- a/locales/base/translation.json
+++ b/locales/base/translation.json
@@ -1450,11 +1450,32 @@
     "swapTo": "Swap To",
     "transactionDetails": "Transaction Details",
     "estimatedGas": "Estimated Gas",
+    "swapSubmitError": "Sorry, we were unable to complete your swap. Please try again or contact support.",
     "swapFee": "{{appName}} Fee",
     "swapFeeTitle": "What is the {{appName}} Fee?",
     "swapFeeBody": "The {{appName}} fee is the fee that you pay to swap tokens on {{appName}}. \n\n The fee is {{swapFee}}% of the amount you are exchanging. \n\n {{appName}} fees are free for the next few months.",
     "estimatedAmountTitle": "Estimated Amount",
     "estimatedAmountBody": "Conversions take place on a third party decentralized exchange that finds you the best rate for your swap. \n\n Your trade will be completed at a price within {{slippagePercent}}% of the estimate or the transaction will be cancelled & funds returned to your wallet."
+  },
+  "SwapExecuteScreen": {
+    "swapPending": "Swap Pending",
+    "swapSuccess": "Swap Successful!",
+    "swapError": "Swap Error",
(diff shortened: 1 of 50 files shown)


src/fiatconnect/FiatDetailsScreen.test.tsx
```diff
@@ -237,7 +248,31 @@ describe('FiatDetailsScreen', () => {
     // Should see an error message saying the account number field is invalid
     // immediately since the field loses focus
     expect(queryByTestId('errorMessage-accountNumber')).toBeTruthy()
-    expect(queryByText('fiatAccountSchema.accountNumber.errorMessage')).toBeTruthy()
+    expect(queryByText('fiatAccountSchema.accountNumber.errorMessageDigit')).toBeTruthy()
```
kathaypacific, line 251: super nit: it's recommended to use `getBy*` if the thing should exist, and `queryBy*` if expecting the thing not to exist. the `getBy*` methods give better errors when it fails :) https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-query-variants-for-anything-except-checking-for-non-existence

test/RootStateSchema.json
```diff
@@ -813,6 +813,12 @@
             ],
             "type": "object"
         },
+        "FiatConnectSchemaCountryOverrides<FiatAccountSchema.AccountNumber>": {
+            "additionalProperties": {
+                "$ref": "#/definitions/{[PropertyinkeyofPartial<FiatAccountSchemas[T]>]:{regex?:string|undefined;errorString?:string|undefined;errorParams?:Record<string,any>|undefined;};}"
```
kathaypacific, line 818: now that you mention it, i'm also a little baffled by why line 816 is specific `FiatConnectSchemaCountryOverrides<FiatAccountSchema.AccountNumber>` but line 818 is generic :/ because it's generic here, the value contains only the common properties for all `T`. if the `FiatConnectSchemaCountryOverrides` definition were changed to the specific `[Property in keyof Partial<FiatAccountSchemas[FiatAccountSchema.AccountNumber]>]` then line 818 will change to what we expect (i just did this as a sanity check) - not sure if it's worth digging into this, i doubt this will cause you any real problems. since this file doesn't have anything to do with runtime, and typescript doesn't complain if you add more properties than the type asks for so shouldn't be a problem if you need to do migrations on this at a later time

locales/base/translation.json
```diff
@@ -1419,8 +1419,9 @@
     },
     "accountNumber": {
       "label": "Account Number",
-      "placeholderText": "Enter your 10 digit account number",
-      "errorMessage": "Account number should be 10 digits long and only contain numbers"
```
kathaypacific, line 1423: yeah you're right, it's safe to remove this because versions that need this will have it bundled anyway :)

kathaypacific (commented):
Looks great!

src/fiatconnect/types.ts
```diff
@@ -0,0 +1,11 @@
+import { FiatAccountSchema, FiatAccountSchemas } from '@fiatconnect/fiatconnect-types'
+
+export interface FiatConnectSchemaCountryOverrides<T extends FiatAccountSchema> {
+  [country: string]: {
```
jophish, line 4: This type is awesome -- really cool design. :)

src/fiatconnect/FiatDetailsScreen.test.tsx
satish-ravi: sounds good, I'll address this in a follow up since every test in this file needs fixing.

src/fiatconnect/types.ts
```diff
@@ -0,0 +1,11 @@
...
+    [Property in keyof Partial<FiatAccountSchemas[T]>]: {
+      regex?: string
```
jophish, line 6: Should `regex` and `errorString` be optional? I would assume we would always want `regex`, and I assume that we would never want to show the default error message when using a custom regex?

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -66,33 +68,51 @@ type FiatAccountFormSchema<T extends FiatAccountSchema> = {
...
+    accountNumber: {
+      name: 'accountNumber',
+      label: i18n.t('fiatAccountSchema.accountNumber.label'),
+      regex: overrides?.accountNumber?.regex
+        ? new RegExp(overrides.accountNumber.regex)
+        : /^[0-9]+$/,
+      placeholderText: i18n.t('fiatAccountSchema.accountNumber.placeholderText'),
+      errorMessage: overrides?.accountNumber?.errorString
+        ? i18n.t(
+            `fiatAccountSchema.accountNumber.${overrides.accountNumber.errorString}`,
```
jh2oman, line 101: I think you should check if `overrides.accountNumber.errorString === 'errorMessageDigitLength`. If you don't, old clients won't be able to handle new errorStrings from Remote Config

satish-ravi: we can push new translation strings OTA right? cc @kathaypacific

jh2oman: Ah how could I forget OTA 🤦 !

src/app/selectors.ts
```diff
@@ -119,6 +119,8 @@ export const fiatConnectCashInEnabledSelector = (state: RootState) =>
   state.app.fiatConnectCashInEnabled
 export const fiatConnectCashOutEnabledSelector = (state: RootState) =>
   state.app.fiatConnectCashOutEnabled
+export const fiatConnectAccountNumberCountryOverridesSelector = (state: RootState) =>
+  state.app.fiatConnectAccountNumberCountryOverrides
```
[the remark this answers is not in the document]
satish-ravi: I believe all remote config flags currently go in app. I am not sure how easy it is to get it into state.fiatConnect since the `UPDATE_REMOTE_CONFIG_VALUES` action is under app.

src/fiatconnect/types.ts
satish-ravi: I left them optional in case in the future we override things other than regex, like label. But we don't have a case so I can make them required for now. Related, should we also allow overriding keyboardType right now, in case we wanted to allow anything other than numbers? It defaults to `numpad`

jophish: Ah yeah, that makes sense. I suppose we can make them required for now and change to optional later if want to use this to override unrelated parameters/fields. I was thinking about `numpad` as well; I think we can leave that be until we have a usecase for overriding it.

As a sidenote, all of this remote configuration/overriding makes me wonder if we can't just have some completely remotely configurable solution for _all_ account schemas, such that supporting a new schema wouldn't require _any_ changes within the wallet; just specify default rules/overrides in firebase per schema and have the wallet construct the UI accordingly. Maybe premature right now, but a potential idea for later perhaps.

src/app/selectors.ts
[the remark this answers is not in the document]
satish-ravi: oh cool, this should be easier to do than I thought 😄

cajubelt (approved):
🥳

PR #2980, opened 2022-10-12
feat: restore phone verification status when restoring wallet

kathaypacific:
### Description

When initialising (restoring) wallet, check if there is a phone number already verified for this wallet. If so, store the verified phone number in redux so that the user correctly displays as verified in the app.

Note that the signed message is required for making the address lookup query, and this is generated potentially after the restore saga finishes so that's why it's in the initializeAccount saga now... 

### Other changes

N/A

### Tested

Manually + unit tests.

### How others should test

Verify a phone number against an address. Reinstall the app and restore this account - the app should know that your number is verified.

### Related issues

- Fixes RET-451

### Backwards compatibility

Yes

### DIFF
--- a/src/account/saga.ts
+++ b/src/account/saga.ts
@@ -125,7 +125,7 @@ function* handlePreviouslyVerifiedPhoneNumber() {
           yield put(
             phoneNumberVerificationCompleted(
               phoneDetails.e164Number,
-              phoneDetails.countryCodeAlpha2
+              phoneDetails.countryCode ? `+${phoneDetails.countryCode}` : null
             )
           )
         } else {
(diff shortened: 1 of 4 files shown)


src/account/saga.ts
```diff
@@ -84,6 +97,53 @@ function* initializeAccount() {
...
+          )
+        } else {
+          throw new Error('Could not parse verified phone number')
+        }
+      }
+    } else {
+      throw new Error(yield call([response, 'text']))
+    }
+  } catch (error) {
+    Logger.debug(
```
jeanregisser, line 139:
```suggestion
    Logger.warn(
```

jeanregisser (approved):
Amazing! 🚀 🚀 🚀 
So much nicer not to have to re-enter your phone number!!!

PR #2983, opened 2022-10-12
feat: revoke phone number from the dev menu with CPV

jeanregisser:
### Description

Add "Revoke Number Verification (centralized)" in the settings dev menu.

### Tested

https://user-images.githubusercontent.com/57791/195550647-57e9861e-ba36-4b2f-83a2-91e9e86d1bb3.mov

### How others should test

- Go to the dev menu in the settings
- Make sure you're verified with the centralized system
- Tap "Revoke Number Verification (centralized)"
- You'll now be back to being unverified

### Related issues

Discussed on [Slack](https://valora-app.slack.com/archives/C044B8ULVC0/p1665561844304069).

### Backwards compatibility

Yes


src/verify/hooks.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+    const { getByText } = render(
+      <Provider store={store}>
+        <TestComponent />
+      </Provider>
+    )
+    mockFetch.mockResponseOnce(JSON.stringify({ message: 'OK' }), {
+      status: 200,
+    })
+
+    await act(async () => {
```
kathaypacific, line 50: do we need async / await here?

kathaypacific (approved):
🚀 YAY!!

src/verify/hooks.test.tsx
jeanregisser: Yes, otherwise it doesn't process the underlying work happening.
I've added an explicit `await flushMicrotasksQueue()` to avoid any confusion.

src/account/Settings.tsx
```diff
@@ -231,9 +232,21 @@ export class Account extends React.Component<Props, State> {
               <Text>Toggle verification done</Text>
             </TouchableOpacity>
           </View>
+          <View style={styles.devSettingsItem}>
```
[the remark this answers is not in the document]
jeanregisser: Yes since this is for testing, I think it's useful to have it all the time.

PR #2988, opened 2022-10-13
fix: crowdin automerge job sometimes leaves incorrect reviews

kathaypacific:
### Description

I found the automerge job sometimes leaves the "changes requested" review when the Crowdin PR does not contain any unexpected changes. I suspect that it is because the review is done before the branch is properly updated. e.g. https://github.com/valora-inc/wallet/pull/2982

In this PR
- wait for a few seconds after updating the branch, to do the review
- make the logging more explicit so that we can see which files are causing the incorrect review

### Other changes

N/A

### Tested

Tested https://github.com/valora-inc/wallet/actions/runs/3243609016/jobs/5318572378

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/.github/scripts/automergeCrowdinPr.js
+++ b/.github/scripts/automergeCrowdinPr.js
@@ -12,8 +12,13 @@
 const CROWDIN_BRANCH = 'l10n/main'
 const CROWDIN_PR_USER = 'valora-bot-crowdin'
 
-const ALLOWED_UPDATED_FILE_MATCHER = new RegExp(`locales\/.*\/translation\.json`)
-const DISALLOWED_UPDATED_FILE = 'locales/base/translation.json'
+const ALLOWED_UPDATED_FILE_MATCHER = new RegExp(
+  `locales\/.*\/translation\.json|ios/celo\/.*\/InfoPlist.strings`
+)
+const DISALLOWED_UPDATED_FILES = [
+  'locales/base/translation.json',
+  'ios/celo/Base.lproj/InfoPlist.strings',
+]
 const enableAutomergeQuery = `mutation ($pullRequestId: ID!, $mergeMethod: PullRequestMergeMethod!) {
   enablePullRequestAutoMerge(input: {
     pullRequestId: $pullRequestId,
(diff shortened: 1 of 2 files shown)


.github/scripts/automergeCrowdinPr.js
```diff
@@ -88,7 +92,9 @@ module.exports = async ({ github, context }) => {
     )
     if (unexpectedFiles.length > 0) {
       console.log(
-        `Files updated in PR #${pr.number} do not match the expectation, please check manually`
+        `The following files updated do not match the expectation: 
+        ${unexpectedFiles.map((file) => file.filename).join(', ')}.
```
jeanregisser, line 96: nit: So it's easier to read?
```suggestion
        ${unexpectedFiles.map((file) => file.filename).join('\n')}.
```

.github/scripts/automergeCrowdinPr.js
```diff
@@ -76,6 +76,10 @@ module.exports = async ({ github, context }) => {
   if (isApproved) {
     console.log('Already approved')
   } else {
+    // wait some seconds before proceeding, to ensure that the checks are done
```
jeanregisser, line 79: Do you mean "wait for the the branch is fully updated"?

jeanregisser (approved):
🚀

.github/scripts/automergeCrowdinPr.js
```diff
@@ -12,8 +12,13 @@
...
 const CROWDIN_PR_USER = 'valora-bot-crowdin'
 
-const ALLOWED_UPDATED_FILE_MATCHER = new RegExp(`locales\/.*\/translation\.json`)
-const DISALLOWED_UPDATED_FILE = 'locales/base/translation.json'
+const ALLOWED_UPDATED_FILE_MATCHER = new RegExp(
+  `locales\/.*\/translation\.json|ios/celo\/.*\/InfoPlist.strings`
+)
+const DISALLOWED_UPDATED_FILES = [
+  'locales/base/translation.json',
+  'ios/celo/Base.lproj/InfoPlist.strings',
```
kathaypacific, line 20: @jeanregisser i also realised that these files were probably causing the last PR review to fail

PR #2991, opened 2022-10-13
fix(fiatconnect): handle empty cache in cacheQuoteParams

jh2oman:
### Description

This action was failing when the cache was empty

### Tested

Added unit tests and can now go through the KYC flow on a fresh account


cajubelt (approved):
whoops

satish-ravi (approved):

jophish (approved):
Agh, sorry for not catching this; thanks for the fix!

PR #2996, opened 2022-10-14
feat: make invite referral link short

kathaypacific:
### Description

An example of a short link: https://vlra.app/vqT5MqgEVCVnBjJB7

Video of it working on Android: https://user-images.githubusercontent.com/20150449/196137182-731cb918-9880-41e5-94cd-c9d430885f34.mp4

### Other changes

N/A

### Tested

Manually - ensure that the link can be created / clicked on. On mobile it should take you to the app or app store, on desktop it should take you to the valoraapp.com website (a redirect to the homepage is coming soon)

### How others should test

Ensure that the invite link can be shared via the invite screen

### Related issues

- Fixes RET-435

### Backwards compatibility

Yes


jeanregisser:
> An example of a short link I was able to generate on Android: https://vlra.app/vUL2DsRdrt52Xvwd9

Related, how do we want to handle clicking these links from a desktop browser? Right now it leads to a 404.

src/invite/utils.ts
```diff
@@ -98,3 +98,13 @@ export async function createDynamicLink(address: string) {
     },
   })
 }
+
+export async function decodeShortDynamicLink(link: string) {
+  try {
+    const resolvedLink = await dynamicLinks().resolveLink(link)
+    return resolvedLink.url
+  } catch (error) {
+    Logger.debug('Invite/utils/decodeShortDynamicLink', 'Link could not be resolved', error)
```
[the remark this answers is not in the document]
kathaypacific: should this be a `warn`? i don't feel strongly but the way we use this is kind of optimistic, like "hey here is a maybe short link, see if you can decode it". it's not wrong for it to not be able to resolve, so i didn't think about a `warn`...

src/app/saga.ts
```diff
@@ -295,9 +296,20 @@ export function* handleDeepLink(action: OpenDeepLink) {
...
+    } else {
+      const dynamicShortLink = yield call(decodeShortDynamicLink, rawParams.path)
+      if (dynamicShortLink) {
+        const shortLinkParts = dynamicShortLink.path.split('/')
+        if (shortLinkParts.length === 3 && shortLinkParts[1] === 'share') {
+          ValoraAnalytics.track(InviteEvents.opened_via_invite_url, {
+            inviterAddress: shortLinkParts[2],
+          })
+        }
+      }
```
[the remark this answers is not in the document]
kathaypacific: yeah good idea, i updated it. also added some tests and corrected some link handling logic

kathaypacific:
> > An example of a short link I was able to generate on Android: https://vlra.app/vUL2DsRdrt52Xvwd9
> 
> Related, how do we want to handle clicking these links from a desktop browser? Right now it leads to a 404.

yeah good question. i'm not sure. it does lead to a 404 right now but that's not a change due to the short link - the long link also gives a 404. i couldn't find a way to make it redirect to a webpage on a browser since firebase dynamic links seems to not allow you to configure links with dynamic parts :/

do you think this is an important use case for us to cover now? i don't think that many users will try to click this link on a desktop browser...

kathaypacific:
@jeanregisser this is ready for review again :)

src/app/saga.ts
```diff
@@ -279,8 +280,17 @@ export function* handleDeepLink(action: OpenDeepLink) {
 
   const rawParams = parse(deepLink)
   if (rawParams.path) {
-    const pathParts = rawParams.path.split('/')
-    if (rawParams.path.startsWith('/v/')) {
+    // handle dynamic link
+    if (rawParams.hostname === 'vlra.app') {
+      const dynamicLink = yield call(resolveDynamicLink, rawParams.href)
```
jacob-israel-turner, line 285: Was this tested on both iOS and Android?

I _believe_ `deepLink` here should already be the resolved link, not the short link (i.e., it should be `vlra.app/share/:address`, not `vlra.app/:shareLinkId`). But it likely depends on which way the link was grabbed in `App.js`.

This wasn't really an issue when we were using long links, because it was always going to be the same link. But now that we're using short links, it may cause issues.

jacob-israel-turner: I'm now realizing that you changed the domain of the URL when creating the link - so the issue I mentioned is likely to not actually be an issue, since anything coming from `vlra.app` should always be the _un-_resolved link.

jacob-israel-turner (approved):
Looks good! I would just want to make sure it's tested on iOS and Android (when the app is installed, and when it's not), as well as desktop web. Let me know if I can help with that!

src/app/saga.ts
kathaypacific: hmm yeah good shout, it's been a while since i tested this and am only 99% sure i tested on both iOS and Android. i'll test again tomorrow before i merge to make sure :) i did change the domain of the `link`, and added a redirect on our website so that now if you click on a share link like `https://vlra.app/vqT5MqgEVCVnBjJB7` on desktop, you land on the valora home page

jacob-israel-turner: OK I'm testing this on Android and it's not working properly, per my initial concern. The link being passed into `handleDeepLink` is `https://valoraapp.com/share/:address` - so the `rawParams.hostname === 'vlra.app'` check fails.

We probably need to tweak the platform-specific stuff in `App.tsx` to make sure the same link (pre-resolved, i.e. `vlra.app`; or post-resolved, i.e. `valoraapp.com`) gets passed into `handleDeepLink`. This is likely because the link is coming from difference libraries depending on the platform (`CleverTap` for Android, `dynamicLinks` for iOS).

Let me know if you want to chat or pair on this, I'm happy to jump on a call.

kathaypacific: thanks for picking this up @jacob-israel-turner ! i think i've fixed it, if you want to take another look. i tested generating and consuming the links on iOS and android again and seems to be working end to end :D

src/app/saga.ts
```diff
@@ -279,8 +281,21 @@ export function* handleDeepLink(action: OpenDeepLink) {
 
   const rawParams = parse(deepLink)
   if (rawParams.path) {
-    const pathParts = rawParams.path.split('/')
-    if (rawParams.path.startsWith('/v/')) {
+    // android resolves dynamic links to the `link`, iOS handles the plain
+    // dynamic link
+    if (
+      rawParams.hostname === new URL(DYNAMIC_LINK_DOMAIN_URI_PREFIX).hostname ||
+      rawParams.hostname === new URL(WEB_LINK).hostname
```
jeanregisser, line 288: If it's the WEB_LINK (https://valoraapp.com), no need to resolve it. Or am I missing something?

kathaypacific: no, you're right. i had left it there because it didn't hurt, but i've updated now to not resolve if it's the resolved link already

src/app/saga.ts
```diff
@@ -269,14 +270,23 @@ function convertQueryToScreenParams(query: string) {
...
+  const { isSecureOrigin } = action
   Logger.debug(TAG, 'Handling deep link', deepLink)
 
   if (isWalletConnectDeepLink(deepLink)) {
     yield call(handleWalletConnectDeepLink, deepLink)
     return
   }
 
+  // Try resolve dynamic links
+  if (deepLink.startsWith(DYNAMIC_LINK_DOMAIN_URI_PREFIX)) {
```
jacob-israel-turner, line 283: Love this, feels very clean!

src/invite/utils.ts
```diff
@@ -98,3 +99,14 @@ export async function createDynamicLink(address: string) {
     },
   })
 }
+
+export async function resolveDynamicLink(link: string) {
```
jacob-israel-turner, line 103: nit: Maybe this should live in a different location? If we use deep links in other features, it might feel weird to pull dynamic link resolution from `invite/utils`.

jacob-israel-turner (approved):
The final implementation is 🤌 

I'll give it a test on Android!

src/invite/utils.ts
kathaypacific: yeah good one, i didn't think about this. the first place that came to mind was to put it with the rest of the firebase functions, let me know if you had something else in mind :)

PR #2997, opened 2022-10-14
chore: add device brand and model to contact support log information

kathaypacific:
### Description

This was a request from the support team, this can be helpful for debugging.  e.g. in RET-445.

### Other changes

N/A

### Tested

Manually

### How others should test

Shake to contact support. When the email appears, see that the expect values of device brand and model are present in the email body.

### Related issues

N/A

### Backwards compatibility

N/A


MuckT (approved):
🚀 📲

PR #2998, opened 2022-10-14
feat(fiatconnect): add IBAN fc schema

anthonydaoud:
### Description

Currently, FiatConnect only supports the `AccountNumber` schema type. This PR adds support for the `IBANNumber` schema type. The `FiatDetailsScreen` is also updated to so that if a user is entering an IBAN account, they see [this design](https://www.figma.com/file/bwIY7peeykI8K4wTrXs4iN/Fiat-Connect?node-id=3114%3A42037).

https://user-images.githubusercontent.com/8432644/197056881-40934715-50cc-483c-96de-5baee5c34272.mp4

### Other changes

N/A

### Tested

Updated unit tests on the `FiatDetailsScreen` and `FiatConnectQuote`. Also tested with mock provider

### How others should test

Request cash out in cEUR to get IBAN account schema

### Related issues

- [ACT-484](https://linear.app/valora/issue/ACT-484/support-ibannumber-fiataccountschema)

### Backwards compatibility

N/A


src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -189,6 +264,11 @@ const FiatDetailsScreen = ({ route, navigation }: Props) => {
           },
           schemaCountryOverrides
         )
+      case FiatAccountSchema.IBANNumber:
+        return getIbanNumberSchema({
+          country: userCountry.countryCodeAlpha2 || 'US',
```
anthonydaoud, line 269: What should the default country be if `countryCodeAlpha2` isn't present?

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -117,12 +134,70 @@ const getAccountNumberSchema = (
...
+    country: string
+    fiatAccountType: FiatAccountType
+  },
+  countryOverrides?: FiatAccountSchemaCountryOverrides
+): FiatAccountFormSchema<FiatAccountSchema.IBANNumber> => {
+  const overrides = countryOverrides?.[implicitParams.country]?.[FiatAccountSchema.IBANNumber]
+  //Ensure spaces/lower case don't lead to invalid input
+  const cleanInput = (input: string) => input.replace(/\s/g, '').toUpperCase()
+
+  const ibanNumberValidator = (input: string) => {
```
cajubelt, line 148: it looks like for account number, we enable overrides for error string in addition to the regex. seems like it's probly worth enabling that for iban, too.

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -117,12 +134,70 @@ const getAccountNumberSchema = (
...
+    if (overrides?.iban?.regex) {
+      const isValid = new RegExp(overrides.iban.regex).test(input)
+      return {
+        isValid,
+        errorMessage: i18n.t(`fiatAccountSchema.ibanNumber.errorMessage`),
+      }
+    }
+
+    const { valid, errorCodes } = validateIBAN(input)
+    if (!valid && input.length) {
```
cajubelt, line 160: not sure the `&& input.length` part makes sense to me. We don't do that for the regex validation functions. As written, it seems like you could have an invalid IBAN with no error message. Is there a reason we'd want that?

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -117,12 +134,70 @@ const getAccountNumberSchema = (
...
+      validate: ibanNumberValidator,
+      placeholderText: i18n.t('fiatAccountSchema.ibanNumber.placeholderText'),
+      keyboardType: 'default',
+    },
+    country: { name: 'country', value: implicitParams.country },
+    fiatAccountType: { name: 'fiatAccountType', value: FiatAccountType.BankAccount },
+    accountName: {
+      name: 'accountName',
+      computeValue: ({ institutionName, iban }) =>
+        //TODO: should I use a different obfuscation method? The account number approach seems good enough.
```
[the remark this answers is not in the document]
cajubelt: the getObfuscatedAccountNumber helper shows the last 4 characters, or the last `n-3` characters (for length-n strings), whichever is less. IBAN numbers are at least [15 characters](https://www.sepaforcorporates.com/single-euro-payments-area/iban-number-format-sepa-country/#:~:text=Interesting%20IBAN%20Number%20Info%3A&text=Norway%20has%20the%20shortest%20IBAN,longest%20IBAN%20at%2031%20characters), so this should work fine.

src/fiatconnect/FiatDetailsScreen.tsx
cajubelt: no US banks support IBAN, so not the US

cajubelt: maybe Nigeria?

cajubelt: should probly log an error if `countryCodeAlpha2` is missing

cajubelt: seems very likely that a bug would result-- and possibly get disguised by this default-- if that field was missing

src/fiatconnect/FiatDetailsScreen.test.tsx
```diff
@@ -103,157 +143,230 @@ describe('FiatDetailsScreen', () => {
...
+        expect(getByText(`fiatAccountSchema.${accountSchema}.label`)).toBeTruthy()
+        expect(getByTestId(`input-${input}`)).toBeTruthy()
+
+        expect(queryByTestId(/errorMessage-.+/)).toBeFalsy()
+
+        expect(getByTestId('submitButton')).toBeTruthy()
+        expect(getByTestId('submitButton')).toBeDisabled()
+        expect(getByText('fiatDetailsScreen.submitAndContinue')).toBeTruthy()
+      })
+      it('renders header with provider image', () => {
```
satish-ravi, line 202: I think header / cancel button tests independent of the schema doesn't need to be tested for both schemas

src/fiatExchanges/quotes/FiatConnectQuote.test.ts
```diff
@@ -28,310 +28,362 @@ const mockExchangeRates = {
...
-        ...mockFiatConnectQuotes[1],
-        fiatAccount: {
-          BankAccount: {
-            fiatAccountSchemas: [
-              {
-                fiatAccountSchema: 'SomethingUnsupported' as FiatAccountSchema,
-                allowedValues: {},
-              },
-            ],
+  describe.each([
```
satish-ravi, line 53: similar here comment here as FiatDetailsScreen.test.tsx, it should be sufficient to test just a single schema for most tests.

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -256,16 +336,17 @@ const FiatDetailsScreen = ({ route, navigation }: Props) => {
 
   const validateInput = () => {
     setValidInputs(false)
-    const newErrorSet = new Set<number>()
+    const newErrorMap = new Map<number, string | undefined>()
```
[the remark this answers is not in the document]
jh2oman: I think its nice. the easy to define types make it smooth to work with

src/fiatExchanges/quotes/FiatConnectQuote.test.ts
```diff
@@ -63,6 +63,32 @@ describe('FiatConnectQuote', () => {
...
+    it.each([FiatAccountSchema.AccountNumber, FiatAccountSchema.IBANNumber])(
+      'does not throw an error if at least one fiatAccountSchema is supported',
+      (fiatAccountSchema) => {
+        const quoteData = {
+          ...mockFiatConnectQuotes[1],
+          fiatAccount: {
+            BankAccount: {
+              fiatAccountSchemas: [
+                {
+                  fiatAccountSchema: FiatAccountSchema.AccountNumber,
```
satish-ravi, line 75: should this use the test param?

src/fiatconnect/fiatAccountSchemas/index.test.ts
```diff
@@ -0,0 +1,68 @@
...
+import { getIbanNumberSchema } from 'src/fiatconnect/fiatAccountSchemas/ibanNumber'
+import { FiatAccountSchemaCountryOverrides } from 'src/fiatconnect/types'
+
+jest.mock('src/fiatconnect/fiatAccountSchemas/accountNumber', () => ({
+  getAccountNumberSchema: jest.fn(),
+}))
+
+jest.mock('src/fiatconnect/fiatAccountSchemas/ibanNumber', () => ({
+  getIbanNumberSchema: jest.fn(),
+}))
```
satish-ravi, line 13:
```suggestion
jest.mock('src/fiatconnect/fiatAccountSchemas/accountNumber')
jest.mock('src/fiatconnect/fiatAccountSchemas/ibanNumber')
```
you can drop this mock return value, this comes by default

satish-ravi: Actually you may want to return something custom here to assert the return value of getSchema

src/fiatconnect/fiatAccountSchemas/ibanNumber.ts
```diff
@@ -0,0 +1,56 @@
...
+      validate: ibanNumberValidator,
+      placeholderText: i18n.t('fiatAccountSchema.ibanNumber.placeholderText'),
+      keyboardType: 'default',
+    },
+    country: { name: 'country', value: implicitParams.country },
+    fiatAccountType: { name: 'fiatAccountType', value: FiatAccountType.BankAccount },
+    accountName: {
+      name: 'accountName',
+      computeValue: ({ institutionName, iban }) =>
+        `${institutionName} (${getObfuscatedAccountNumber(clean(iban!))})`,
```
satish-ravi, line 53: is the `!` operator safe to use here?

src/fiatconnect/fiatAccountSchemas/index.test.ts
jh2oman: Ah asserting the return value is a good idea

src/fiatconnect/fiatAccountSchemas/ibanNumber.ts
jh2oman: mmm, good catch. I didn't even notice it when i looked over the initial code.

satish-ravi (approved):

PR #3001, opened 2022-10-17
feat: add background phone verification migration

kathaypacific:
### Description

The migration should be run after remote config fetched if CPV is on + the user is verified decentrally _but not centrally_. In case of failures, it should be silent since we will see the errors on the cloud function logs already and the wallet is backwards compatible.

### Other changes

The app/saga.test.ts file was getting a bit messy with inconsistent grouping of tests. I rearranged the grouping of the tests, would recommend reviewing without white spaces. **most of the lines changed in this PR are from the test file!**

### Tested

As a decentrally verified user, open the app. Verify that you can send to another number that is only verified centrally. Verify that when restoring this account, the verification status is preserved automatically.

### How others should test

As above.

### Related issues

- Fixes RET-405

### Backwards compatibility

Yes


src/app/saga.ts
```diff
@@ -364,10 +378,76 @@ export function* handleSetAppState(action: SetAppState) {
...
+  if (!privateDataEncryptionKey) {
+    Logger.warn(
+      `${TAG}@runCentralPhoneVerificationMigration`,
+      'No data encryption key was found in the store. This should never happen.'
+    )
+    return
+  }
+
+  const shouldRunVerificationMigration = yield select(shouldRunVerificationMigrationSelector)
+  if (shouldRunVerificationMigration) {
```
jeanregisser, line 392: nit: return early for less nesting.
```suggestion
  if (!shouldRunVerificationMigration) {
    return
  }
```

jeanregisser (approved):
🚀

PR #3004, opened 2022-10-18
perf: use @react-navigation/native-stack and @react-navigation/native v6

MuckT:
### Description

Moves to @react-navigation/native-stack from @react-navigation/stack for increased performance. Initial testing shows a ~28% memory usage reduction from 350MB to 250MB; however, loading a large amount of transaction in the transaction feed will increase memory usage both on main and in this branch.

Some types errors are ignored when using `@react-navigation/native@6.x.x` and navigators with `5.x.x` versions such as `@react-navigation/drawer` and `@react-navigation/material-top-tabs`: mixing Navigators versions is ok until all v5 packages can be upgraded: [docs](https://reactnavigation.org/docs/upgrading-from-5.x/#note-on-mixing-react-navigation-5-and-react-navigation-6-packages).

#### Code changes
- `import/saga.ts` no longer calls `NavigateClearingStack`. Instead after navigating to the verification education screen the device back is disabled. This was due to issues with the navigation state after restoring on initial app load.
- `popToScreen` is used on `pincodeSet.tsx` to reset the stack during the import flow.

##### Removed Packages

[@react-navigation/compat](https://www.npmjs.com/package/@react-navigation/compat)
[@react-navigation/core](https://www.npmjs.com/package/@react-navigation/core)
[@react-navigation/stack](https://www.npmjs.com/package/@react-navigation/stack)
    
##### Added Packages
[@react-navigation/native-stack](https://www.npmjs.com/package/@react-navigation/native-stack)
[@react-navigation/elements](https://www.npmjs.com/package/@react-navigation/elements)

##### Upgraded Packages
[@react-navigation/native](https://www.npmjs.com/package/@react-navigation/native) - 5.9.8 -> 6.0.13

### Tested

Tested locally on iOS iPhone 11 15.5 and Android SDK 21, 29 & 31.

#### Devs
- Run the app in profile mode on Android and check the memory usage `yarn dev:android -e mainnet -r -t`.
- Follow the steps in the [profile section of the WALLET.md](https://github.com/valora-inc/wallet/blob/main/WALLET.md#app-profiling-with-android-profiler).
- Check for misaligned buttons, incorrect background colors, or text in headers.
- Check for screens not properly dispatched / pincode entry not dispalyed.

#### QA

- Navigate through onboarding screens and ensure that the proper header is displayed.
- Navigate through the Add & Withdraw flows and ensure the proper header is displayed.
- Navigate through setting items and ensure that the proper header is displayed.
- Navigate through the send and receive flows and ensure that the proper header is displayed.
- Check the QR code scanner header is displayed correctly.
- Check for screens not shown in proper order or pincode entry not displayed.

### Related issues

RET-361

### Backwards compatibility

Yes


MuckT:
### A list of small items to fix 
please feel free to add more as a comment if not already mentioned.
- [ ] Back button on `swapReviewScreen.tsx` header
- [ ] Invites Header alignment

src/account/GoldEducation.tsx
```diff
@@ -51,7 +50,6 @@ export default function GoldEducation() {
 
 GoldEducation.navigationOptions = {
   ...noHeader,
-  ...TransitionPresets.ModalTransition,
```
jeanregisser, line 54: I think all screens that were displayed with a modal transition should be moved to a modal stack. But we can do this separately.

src/account/Settings.tsx
```diff
@@ -327,10 +328,17 @@ export class Account extends React.Component<Props, State> {
...
 
-  onPressContinueWithAccountRemoval = () => {
-    ValoraAnalytics.track(SettingsEvents.start_account_removal)
-    this.setState({ showAccountKeyModal: false })
-    this.props.navigation.navigate(Screens.BackupPhrase, { navigatedFromSettings: true })
+  onPressContinueWithAccountRemoval = async () => {
+    try {
+      this.setState({ showAccountKeyModal: false })
+      const pinIsCorrect = await ensurePincode()
+      if (pinIsCorrect) {
```
jeanregisser, line 335: Is there a reason this new feature was added specifically in this PR?

src/consumerIncentives/ConsumerIncentivesHomeScreen.tsx
```diff
@@ -228,6 +228,8 @@ export default function ConsumerIncentivesHomeScreen() {
         buttonPressed: RewardsScreenCta.ClaimRewards,
       })
     } else if (userIsVerified) {
+      // TODO: use navigate instead of replace currently navigate buggy on Android
+      // During initial load if account is restored from backup
```
jeanregisser, line 232: Should this be fixed before merging the PR?

src/exchange/ExchangeScreenHeader.tsx
```diff
@@ -76,7 +77,15 @@ function ExchangeTradeScreenHeader({ currency, isCeloPurchase, onChangeCurrency
 
   return (
     <>
-      <CustomHeader left={<BackButton eventName={cancelEventName} />} title={title} />
+      <CustomHeader
+        left={
+          <BackButton
+            style={{ paddingLeft: variables.contentPadding }}
```
jeanregisser, line 83: Looks like this is added in multiple places, is there a way to hide this detail? so we don't have to modify multiple place.

src/import/ImportWallet.tsx
```diff
@@ -176,9 +179,6 @@ function ImportWallet({ navigation, route }: Props) {
         <SafeAreaInsetsContext.Consumer>
           {(insets) => (
             <View style={styles.container}>
-              <UseBackToWelcomeScreen
-                backAnalyticsEvents={[OnboardingEvents.restore_account_cancel]}
```
jeanregisser, line 180: Are we still tracking this somewhere else?

src/import/saga.ts
```diff
@@ -166,9 +166,11 @@ export function* importBackupPhraseSaga({ phrase, useEmptyWallet }: ImportBackup
     const skipVerification = yield select(skipVerificationSelector)
     if (skipVerification || numberAlreadyVerifiedCentrally) {
       yield put(setHasSeenVerificationNux(true))
+      // navigateHome will clear onboarding Stack
       navigateHome()
     } else {
-      navigateClearingStack(Screens.VerificationEducationScreen)
+      // DO NOT CLEAR NAVIGATION STACK HERE - breaks restore flow on initial app open in native-stack v6
+      navigate(Screens.VerificationEducationScreen)
```
jeanregisser, line 173: Does this comment apply to both iOS and Android?

src/navigator/DrawerNavigator.tsx
```diff
@@ -251,7 +250,7 @@ export default function DrawerNavigator() {
           {(isCeloEducationComplete && (
             <Drawer.Screen
               name={Screens.ExchangeHomeScreen}
-              component={ExchangeHomeScreen}
+              component={ExchangeHomeScreen as any}
```
jeanregisser, line 253: No way not to cast? If not, I recommend `// @ts-expect-error` instead so we can more easily find it in the future (and remove when unncessary).

src/navigator/NavigationService.ts
```diff
@@ -45,6 +47,23 @@ async function ensureNavigator() {
...
+  const [routeName] = args
+  ensureNavigator()
+    .then(() => {
+      Logger.debug(`${TAG}@popToScreen`, `Dispatch ${routeName}`)
+      while (
+        navigationRef.current?.canGoBack() &&
+        navigationRef.current?.getCurrentRoute()?.name !== routeName
+      ) {
+        navigationRef.current?.dispatch(StackActions.pop())
+      }
```
jeanregisser, line 60: Feels a bit hacky 😅 
No other way to make this work?

src/send/SendAmount/SendAmountHeader.tsx
```diff
@@ -71,7 +72,12 @@ function SendAmountHeader({
   return (
     <>
       <CustomHeader
-        left={<BackButton eventName={backButtonEventName} />}
+        left={
+          <BackButton
+            eventName={backButtonEventName}
+            style={{ paddingLeft: variables.contentPadding + 6 }}
```
jeanregisser, line 78: Can we make `CustomHeader` include the required padding instead?

src/swap/SwapReviewScreen.tsx
```diff
@@ -148,7 +148,10 @@ export function SwapReviewScreen() {
 
   return (
     <SafeAreaView style={styles.safeAreaContainer}>
-      <CustomHeader title={t('swapReviewScreen.title')} left={<BackButton />} />
+      <CustomHeader
+        title={t('swapReviewScreen.title')}
+        left={<BackButton style={{ paddingHorizontal: variables.contentPadding }} />}
```
jeanregisser, line 153: Why is the padding different here? Now I'm convinced we should centralize that logic.

src/verify/VerificationStartScreen.tsx
```diff
@@ -122,9 +123,20 @@ function VerificationStartScreen({
...
   }, [navigation, step, totalSteps, route.params])
 
+  // Prevent device back on Android during onboarding
+  useEffect(() => {
+    if (!route.params?.hideOnboardingStep) {
+      const backPressListener = () => true
+      BackHandler.addEventListener('hardwareBackPress', backPressListener)
+      return () => BackHandler.removeEventListener('hardwareBackPress', backPressListener)
+    }
+  }, [])
```
jeanregisser, line 138: Can we use `navigation.addListener('beforeRemove', ...` instead to handle both iOS and Android with the same code?

src/navigator/Navigator.tsx
```diff
@@ -289,11 +291,13 @@ const sendScreens = (Navigator: typeof Stack) => (
     />
     <Navigator.Screen
       name={Screens.IncomingPaymentRequestListScreen}
+      // @ts-ignore
```
jeanregisser, line 294: nit: once this become unnecessary TS will remind us to remove it
```suggestion
      // @ts-expect-error
```

src/navigator/Navigator.tsx
```diff
@@ -662,7 +667,7 @@ const modalAnimatedScreens = (Navigator: typeof Stack) => (
     <Navigator.Screen
       name={Screens.PincodeEnter}
       component={PincodeEnter}
-      options={PincodeEnter.navigationOptions}
+      options={PincodeEnter.navigationOptions as NativeStackNavigationOptions}
```
jeanregisser, line 670: This is because of v5 / v6 type conflict right?

src/navigator/Navigator.tsx
```diff
@@ -736,7 +743,7 @@ function nativeBottomSheets(BottomSheet: typeof RootStack) {
 
 function ModalStackScreen() {
   return (
-    <ModalStack.Navigator mode="modal">
```
jeanregisser, line 739: Why remove?

jeanregisser (commented):
Awesome work!!! 💥 🚀 🚀 
I always wanted us to switch to the native stack so happy you did it!
Love the improved perf ❤️ 

> On iOS the pincode entry is no longer a modal. This matches the behavior on Android.

Oh no 😅 I really liked this. Can we bring this back?

> Some animation between screens might be different than expected - we can adjust this with the [animation prop](https://reactnavigation.org/docs/native-stack-navigator#animation).

Indeed, some are weird, going forward and backward when accepting the Terms for instance.

Couple of small things I noticed while running the branch:

The header during onboarding was transparent and we could see the image underneath, now it's blocking it, and the transition feels different (not fullscreen):


Not sure if it's only in dev mode, but the header is too much toward the top here:
 
and here:
 

The "Done" button left padding in the WebView is too big:

e2e/src/usecases/PINChange.js
```diff
@@ -23,7 +23,9 @@ export default ChangePIN = () => {
 
     // Enter an invalid pin and check that we get the correct error and start over
     await enterPinUi('902100')
-    await expect(element(by.text("The PINs didn't match"))).toBeVisible()
+    await waitFor(element(by.text("The PINs didn't match")))
+      .toBeVisible()
+      .withTimeout(15 * 1000)
```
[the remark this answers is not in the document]
MuckT: Detox expected a timeout on waitFor calls, it will only wait up to 15 seconds in this case, and continue on if found sooner: [docs](https://github.com/nubank/detox/blob/master/docs/APIRef.waitFor.md#tobevisible).

src/account/Settings.tsx
MuckT: Was having an issue where the recovery phrase would display as empty, but this also occurs in main it's own PR makes sense.

https://user-images.githubusercontent.com/26950305/199777055-3f4cb527-9916-4c5d-a429-53b5ce4947cc.mov

src/consumerIncentives/ConsumerIncentivesHomeScreen.tsx
MuckT: Removed comment in 103ad0f06c7e2bda92d3ae94334750720e6cffe2 and fixed in c61d4aaa96cc6e34e5d6f46347fe58fb20bd4fcb.

src/import/saga.ts
MuckT: Yes - clearing the stack here was causing issues on both iOS and Android. Using the `UseBackToWelcomeScreen` was only causing issues on Android.

src/navigator/DrawerNavigator.tsx
MuckT: Fixed in 09194e0af07990b6f94e91594ed218f4f1786b8b!

src/navigator/NavigationService.ts
MuckT: I'll take a  look at using a reset and setting the stack that way, but with my initial attempts it wasn't working.

src/import/ImportWallet.tsx
MuckT: We are not. I'll add it back maybe on press of the back button on `importWallet.tsx`?

src/send/SendAmount/SendAmountHeader.tsx
MuckT: I agree that we should make this change, but it seems outside of the scope of the PR. I'd rather make the changes in their own PR to avoid more changes in this PR. I have a stash with those changes and would be happy open a PR to standardize the headers.

src/navigator/Navigator.tsx
MuckT: `mode="modal"` is removed in v6. Can take a look at some similar options: [docs](https://reactnavigation.org/docs/upgrading-from-5.x/#modemodal-is-removed-in-favor-of-presentation-modal).

src/navigator/Navigator.tsx
MuckT: Fixed in 2f323b52ff25806a3cc1185b0b2c3182943ba035!

src/exchange/ExchangeScreenHeader.tsx
MuckT: Headers would make a good follow up PR.

src/verify/VerificationStartScreen.tsx
MuckT: We'd still have to have the `gestureEnabled` even with the `navigation.addListener('beforeRemove', ...`, [docs](https://reactnavigation.org/docs/preventing-going-back/).
> this feature also doesn't work properly with [@react-navigation/native-stack](https://reactnavigation.org/docs/native-stack-navigator). To make this work, you need to:
> - Disable the swipe gesture for the screen (gestureEnabled: false)...

MuckT:
@Jean I was able to fix the send confirmation screens seen in Bidali; however, I was unable to reproduce any of the other 3 screens issues on iOS. Is it possible the podfile was not updated when you tested?

package.json
```diff
@@ -93,12 +93,11 @@
...
     "@react-native-masked-view/masked-view": "^0.2.6",
     "@react-native-picker/picker": "^2.4.2",
-    "@react-navigation/compat": "^5.3.20",
-    "@react-navigation/core": "^6.2.1",
     "@react-navigation/drawer": "^5.12.9",
+    "@react-navigation/elements": "^1.3.6",
     "@react-navigation/material-top-tabs": "^5.3.19",
-    "@react-navigation/native": "^5.9.8",
-    "@react-navigation/stack": "^5.14.9",
+    "@react-navigation/native": "^6.0.13",
```
kathaypacific, line 99: do we need to upgrade this now? reading the docs you linked, https://reactnavigation.org/docs/upgrading-from-5.x/#note-on-mixing-react-navigation-5-and-react-navigation-6-packages it seems like there are some breaking changes when upgrading this particular package. for example, https://reactnavigation.org/docs/upgrading-from-5.x/#no-more-state-property-on-the-route-prop - i think we are using this in the NavigationWrapper. if we're already going to mix v5 and v6 navigators, can we minimise the amount of unnecessary upgrades so that we can see more easily which code changes are related to which upgrades? it'd be nice to avoid sneaking in unrelated upgrades in the same PR 😅 we can always create a PR especially for an upgrade

src/account/Settings.tsx
kathaypacific: i think if the issue is already there in main, we should be raising a separate PR to address it rather than sneaking it in here. it's important that PR's are focused, for many reasons but a few that come to mind - focused PR's serves as better documentation (e.g. if you see a line of code that you don't understand, you can see which PR introduced it to get more context) and allows us to detect which features caused bugs more easily (e.g. rewinding commits until the issue cannot be repro'ed anymore).

src/components/BackButton.tsx
```diff
@@ -24,7 +23,6 @@ BackButton.defaultProps = {
 
 const styles = StyleSheet.create({
   container: {
-    paddingLeft: variables.contentPadding + 6, // 6px from the left padding
```
kathaypacific, line 27: I see that you moved this style to be an inline style in the CustomHeader component, why do we need this change? is it backwards compatible for other components that use it?

src/exchange/ExchangeScreenHeader.tsx
kathaypacific: can't we just add the `paddingLeft: variables.contentPadding` style to the `BackButton` directly, and allow the CustomHeader to overwrite this style? then we don't need to include this style as inline everywhere

package.json
```diff
@@ -179,8 +178,8 @@
     "react-native-randombytes": "^3.6.0",
     "react-native-reanimated": "^2.5.0",
     "react-native-restart-android": "^0.0.7",
-    "react-native-safe-area-context": "^3.3.2",
-    "react-native-screens": "^3.13.1",
+    "react-native-safe-area-context": "^4.4.1",
```
kathaypacific, line 181: do we need to upgrade this package as part of using `@react-navigation/native-stack`? this upgrade introduced a lot of snapshot updates which has bloated this PR and made it harder to review, if it's not a direct dependency then we should refrain from sneaking this in...

src/import/ImportWallet.tsx
kathaypacific: why does this component need to be removed in the first place?

src/navigator/DrawerNavigator.tsx
```diff
@@ -300,6 +298,7 @@ export default function DrawerNavigator() {
       {(!backupCompleted || !shouldShowRecoveryPhraseInSettings) && (
         <Drawer.Screen
           name={Screens.BackupIntroduction}
+          // @ts-expect-error
```
kathaypacific, line 301: could we add a comment about why we expect an error?

src/navigator/NavigationService.ts
```diff
@@ -45,6 +47,23 @@ async function ensureNavigator() {
...
+    })
+    .catch((reason) => {
+      Logger.error(`${TAG}@popToScreen`, 'Navigation failure', reason)
```
kathaypacific, line 63: `Logger.warn`? do we need to alert Sentry here?

src/navigator/Navigator.tsx
```diff
@@ -289,11 +291,13 @@ const sendScreens = (Navigator: typeof Stack) => (
     />
     <Navigator.Screen
       name={Screens.IncomingPaymentRequestListScreen}
+      // @ts-expect-error
```
kathaypacific, line 294: same with this file, can we have a comment about why we expect an error?

src/send/SendConfirmation.tsx
```diff
@@ -260,10 +260,7 @@ function SendConfirmation(props: Props) {
   const allowComment = !isInvite && isStablecoin(tokenInfo)
 
   return (
-    <SafeAreaView
-      style={styles.container}
-      edges={props.route.name === Screens.SendConfirmationModal ? ['bottom'] : undefined}
```
kathaypacific, line 265: why remove this line?

src/send/SendConfirmationLegacy.tsx
```diff
@@ -381,10 +381,7 @@ function SendConfirmationLegacy(props: Props) {
     }
 
     return (
-      <SafeAreaView
-        style={styles.container}
-        edges={props.route.name === Screens.SendConfirmationLegacyModal ? ['bottom'] : undefined}
```
kathaypacific, line 386: same here, why remove?

src/verify/VerificationStartScreen.tsx
```diff
@@ -255,6 +267,7 @@ function VerificationStartScreen({
 const styles = StyleSheet.create({
   container: {
     flex: 1,
+    justifyContent: 'center',
```
kathaypacific, line 270: i think this is a duplicate style?

src/send/SendConfirmation.tsx
MuckT: This was causing the issue with Send confirmation Screen seen in @jeanregisser screenshot, as it was only happening on the `SendConfirmation.tsx` screen and `SendConfrimationLegacy.tsx`. Since it's working with the defaults I opted to use those instead of passing undefined.

src/verify/VerificationStartScreen.tsx
MuckT: This is so the loading arrow displays in the center without it I was seeing it display at the top after changes.

MuckT: Looks like it duplicated after #3064 was merged to main. Looks like we both saw the issue. Fixed in f08a67eb24edd1e351d63b020b160a50f416288f!

src/import/ImportWallet.tsx
MuckT: This component was causing a freeze on import wallet after setting pin during the restore flow.

package.json
MuckT: @kathaypacific I was able to pull out the upgrades to `react-native-safe-area-context` and `react-native-screens` which should clean-up a lot of the snapshot bloat without issues. Downgrading `@react-navigation/native` produces some intermediate errors that are fixed by the upgrade, so I'd rather include that in this PR instead of fixing the intermediate errors that are fixed by the upgrade.

src/account/Settings.tsx
MuckT: The reason I fixed it hear instead of main is it was causing issues with the e2e tests passing. On main they pass, but after the upgrades they fail on this branch.

package.json
MuckT: Regarding the documentation with `route.state` not being present, It is still present in v6 even if the docs say it isn't. We do use it in our navigation wrapper to get the getActiveRouteName recursively and this appears to be core to how we're managing our navigation. Left another comment about why we should also upgrade `@react-navigation/native` [here](https://github.com/valora-inc/wallet/pull/3004#discussion_r1014341590).

src/send/SendConfirmationLegacy.tsx
MuckT: Was needed for the `react-native-safe-area-context` upgrade but is no longer included with 5d947ba8bb9c3a3f0740a60a59c04fa33e52a533.

src/send/SendConfirmation.tsx
MuckT: Fixed with downgrade in 5d947ba8bb9c3a3f0740a60a59c04fa33e52a533.

src/exchange/ExchangeScreenHeader.tsx
MuckT: I tried adding it directly to the back button and it makes it so there are additional changes in other screens already using the BackButton component.

src/navigator/Navigator.tsx
jeanregisser: Can we switch it to `presentation: 'modal'` then?

src/exchange/ExchangeScreenHeader.tsx
MuckT: Adjusted the header changes in c4fd878f8760e1507784e5888c8e8086bd4eba2b. There's a few specific changes to headers in this PR but they should be easier to follow now.

src/swap/SwapReviewScreen.tsx
MuckT: Updated in c4fd878f8760e1507784e5888c8e8086bd4eba2b!

src/components/BackButton.tsx
MuckT: Updated in c4fd878f8760e1507784e5888c8e8086bd4eba2b!

src/navigator/NavigationService.ts
MuckT: The other navigation functions use logger.error when failing to navigate, so I think we do want to alert sentry here to be consistent.

kathaypacific:
i wasn't able to repro the first 2 observations @jeanregisser had, the header on the welcome screen was transparent and the header on the recovery screen was not overlapping the status bar for me. I did run into this weird issue where if i skip phone verification, the "are you sure you want to skip" dialog doesn't get dismissed and the app becomes stuck. this happens to me on the create and the restore flow. are you able to repro? 

https://user-images.githubusercontent.com/20150449/200870363-25526cc0-fb16-4d5c-8788-883d5551cfbe.mp4

i also see there is too much horizontal padding on the headers for screens like the onboarding screens, webview, bidali etc. - i didn't dig into where these changes are coming from but i think they are new. on the screenshot below the close button has too much left padding and on screens where there is a right element (e.g. the first welcome screen) the right padding is also too big?
simulator_screenshot_A8F8D302-5C29-4C9D-BD70-E41B7CF01B36

src/verify/VerificationEducationScreen.tsx
```diff
@@ -137,6 +138,13 @@ function VerificationEducationScreenDecentralized({ route, navigation }: Props)
     }
   }
 
+  // Set showSkipDialog to false during component unmount
+  useEffect(() => {
+    return () => {
+      navigation.setParams({ showSkipDialog: false })
```
kathaypacific, line 144: ah, this causes problems with connecting phone number from the settings screen since we rely on the `hideOnboardingStep` navigation param to be persisted during navigation to the country selection screen (watch how the header changes after the country screen is dismissed) :( i don't think this fixes the problem either, for me the skip dialog continues to block the app. can we remove this change?

https://user-images.githubusercontent.com/20150449/201121005-8f009f67-36a0-4738-9701-19a0c1cc4a9e.mp4

kathaypacific (commented):
I had a click through on iOS and Android, and looks great so far. 

i am consolidating my observations in this comment :)
1. i noticed that screens that were modals are now real screens - any chance we can minimise these changes? e.g. can the QR code screen still be a modal? the language screen is now also a real screen which looks strange when launching it from the very first onboarding screen
2. the phone verification header changes when launched from the settings screen
3. various headers have incorrect horizontal spacing - this is not an exhaustive list
Screenshot_20221110-135740_Alfajores (dev)
Screenshot_20221110-135823_Alfajores (dev)
Screenshot_20221110-140143_Alfajores (dev)
Screenshot_20221110-150837_Alfajores (dev)
Screenshot_20221110-152314_Alfajores (dev)

i know it's probably becoming a grind to work on this, if you want to take a break i'd be super happy to take a look at the headers and the phone verification issue when i'm back on monday!

MuckT:
@kathaypacific it's really strange I am unable to reproduce the header errors any of the listed screens:
Welcome, Language, Settings, Add funds, or Edit Profile. I am also unable to reproduce the app freeze seen from the  VerificationEducation. @jeanregisser are you able to reproduce these issues?

jeanregisser:
One last question, do we need to restore `dispatch(cancelCreateOrRestoreAccount(...)` that used to be handled by `UseBackToWelcomeScreen` @MuckT?

jeanregisser:
I just checked and if we don't `dispatch(cancelCreateOrRestoreAccount(...)`, and the user kills the app before importing their account but after having entered their PIN, the next time the app is started it will show the Import screen. Which is somewhat of a problem because the PIN isn't persisted.

https://github.com/valora-inc/wallet/blob/66f8169e107b3bbc1667fc559a0327d77eb44204/src/account/reducer.ts#L129-L135

So I restored that behavior: https://github.com/valora-inc/wallet/pull/3004/commits/518f51623b908232b1eec9745a05e1eb7299039c

There are probably more edge cases to fix though, which were already broken prior to this PR.

src/verify/VerificationStartScreen.tsx
```diff
@@ -124,9 +125,20 @@ function VerificationStartScreen({
           />
         ),
       headerLeft: () => route.params?.hideOnboardingStep && <BackButton />,
+      // Disable iOS back during onboarding
+      gestureEnabled: route.params?.hideOnboardingStep ? false : true,
```
kathaypacific, line 129: this causes the same issue as i mentioned before, where if a user isn't launching this screen from onboarding -> they select the phone number country -> the `hideOnboardingStep` becomes true and we display "step 4 /4". this is not a deal breaker i guess

src/account/Settings.tsx
```diff
@@ -324,10 +327,23 @@ export class Account extends React.Component<Props, State> {
...
-    this.setState({ showAccountKeyModal: false })
-    this.props.navigation.navigate(Screens.BackupPhrase, { navigatedFromSettings: true })
+  onPressContinueWithAccountRemoval = async () => {
+    try {
+      this.setState({ showAccountKeyModal: false })
+      // Ugly hack to wait for the modal to close,
+      // otherwise the native modal PIN entry will not show up
+      // TODO: stop using ReactNative modals and switch to react-navigation modals
+      if (Platform.OS === 'ios') {
+        await sleep(500)
```
kathaypacific, line 337: just checking we still need this? is there some logic for chosing 500ms? i kind of worry that there is some race condition here and 500ms might not work for all users...

kathaypacific (approved):
i had a quick tap around the app and it feels great :D no blocking problems from my side, i do notice that the android status bar is white on the onboarding flow for me so you can't see it but i'm also happy to address this later. this PR has got pretty huge and would be nice to draw a line in the sand at some point 🙈

i have been trying to find a bigger chunk of time to review this again more in depth, but alas  haven't been able to find enough time to do it justice and i don't want to become a bottleneck. i think this is fine to be merged. i do think there is a way to simplify the headers but we can wait to address that more holistically.

src/account/Settings.tsx
MuckT: We do still need this to allow the modal from reset account to fully dismiss before showing the pin page modal.

src/verify/VerificationStartScreen.tsx
MuckT: Fixed in 00b81317f5c7ef19745a55f8674f7564ffca6d79!

src/account/Settings.tsx
jeanregisser: > just checking we still need this? is there some logic for chosing 500ms? i kind of worry that there is some race condition here and 500ms might not work for all users...

Yes I hate this as well, it's enough time for the modal to close. We could improve this by trying to use the `onModalHide` but it requires more changes that I didn't want to do here.

jeanregisser (approved):
🚀 🚀 🚀

PR #3009, opened 2022-10-18
refactor(fiatconnect): Refactor FiatDetails to prepare for more fiatAccountSchemas

jh2oman:
### Description

There was feedback in the IBAN pr to split into multiple files. I thought it would be easier to do the bulk of that upfront.

### Tested

refactor so no additional tests. Verified running locally


satish-ravi (approved):
Nice!

PR #3011, opened 2022-10-19
test(fiatconnect): e2e test for first time non-kyc cash out

cajubelt:
### Description

Adds an end-to-end test for first-time, non-kyc cash outs.

Uses our mock provider.

To guarantee a first time cash out, generates a new wallet and funds it from test wallet. Successful cash outs should send funds right back to the faucet once [this](https://github.com/fiatconnect/mock-provider/pull/87) is merged. Failures could result in slow loss of funds-- if we care enough, we can catch errors and attempt to return funds from the new wallets. For now, that seems like a premature optimization. (2.5 Alfajores cents per failure is not too bad.)

### Other changes

adds ability to simulate onboarding for arbitrary wallet, instead of the main e2e test wallet

### Tested

ran locally, passes. result:
https://user-images.githubusercontent.com/7979215/196597735-1ae17a96-7825-4787-a9ce-ecc48eca80ab.mp4

### How others should test

na

### Related issues

cc https://linear.app/valora/issue/ACT-493/wallet-add-e2e-tests-for-fiatconnect-non-kyc-flow

next up is returning user flow, then trying to align both to the e2e test plan linked to in the issue a little better

### Backwards compatibility

na


e2e/src/FiatConnectTransferOut.spec.js
```diff
@@ -0,0 +1,6 @@
+import { fiatConnectNonKycTransferOut } from './usecases/FiatConnectTransferOut'
+
+describe('FiatConnect Transfer Out', () => {
+  // deliberately not doing onboarding in beforeEach (since we sometimes want to re-use an account for returning user flow testing)
```
cajubelt, line 4:
```suggestion
  // deliberately not doing onboarding in beforeEach, since we'll want to re-use accounts, use fresh accounts, etc for these tests
```

e2e/src/usecases/FiatConnectTransferOut.js
```diff
@@ -0,0 +1,95 @@
...
+  const kit = newKit(ALFAJORES_FORNO_URL)
+  const { address: senderAddress } = kit.web3.eth.accounts.privateKeyToAccount(senderPrivateKey)
+  kit.connection.addAccount(senderPrivateKey)
+  console.log(`funding wallet ${recipientAddress} from ${kit.web3.defaultAccount}`)
+  const tokenContract = await kit.contracts.getStableToken(token)
+  const amountWei = kit.web3.utils.toWei(amountEth, 'ether')
+  await tokenContract.transfer(recipientAddress, amountWei.toString()).send({ from: senderAddress })
+}
+
+// TODO return funds if test fails before crypto is sent? could do this by wrapping pre-transfer steps in a try/catch, then returning funds to test wallet in the "catch" block before re-throwing
```
cajubelt, line 24:
```suggestion
```

e2e/src/usecases/FiatConnectTransferOut.js
```diff
@@ -0,0 +1,94 @@
...
+  await waitForElementId('Hamburger')
+  await element(by.id('Hamburger')).tap()
+  await element(by.id('add-and-withdraw')).tap()
+}
+
+async function fundWallet(senderPrivateKey, recipientAddress, token, amountEth) {
+  const kit = newKit(ALFAJORES_FORNO_URL)
+  const { address: senderAddress } = kit.web3.eth.accounts.privateKeyToAccount(senderPrivateKey)
+  kit.connection.addAccount(senderPrivateKey)
+  console.log(`funding wallet ${recipientAddress} from ${kit.web3.defaultAccount}`)
```
cajubelt, line 18:
```suggestion
```

e2e/scripts/consts.ts
```diff
@@ -1,2 +1,3 @@
 export const E2E_TEST_WALLET = '0x6131a6d616a4be3737b38988847270a64bc10caa'
 export const E2E_TEST_FAUCET = '0xe5F5363e31351C38ac82DBAdeaD91Fd5a7B08846'
+export const ALFAJORES_CUSD_ADDRESS = '0x874069fa1eb16d44d622f2e0ca25eea172369bc1'
```
cajubelt, line 3:
```suggestion
```

MuckT (approved):
These tests look good to me and I like what you did with the helper functions. As more fiat connect tests are added there might be some work to ensure that the state is correct at the start of the tests, but these appear to be passing as is.

e2e/src/FiatConnectTransferOut.spec.js
```diff
@@ -0,0 +1,6 @@
+import { fiatConnectNonKycTransferOut } from './usecases/FiatConnectTransferOut'
+
+describe('FiatConnect Transfer Out', () => {
+  // deliberately not doing onboarding in beforeEach, since we'll want to re-use accounts, use fresh accounts, etc for these tests
```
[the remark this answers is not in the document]
satish-ravi: Do we still need the above in `beforeAll` for these tests? Or is there a guarantee that at the start of each test suite, there's a fresh install?

e2e/src/usecases/FiatConnectTransferOut.js
```diff
@@ -0,0 +1,105 @@
...
+
+export const fiatConnectNonKycTransferOut = () => {
+  it('First time FiatConnect cash out', async () => {
+    const cashOutAmount = 0.02
+    const amountPlusGas = `${cashOutAmount + 0.005}`
+    const token = 'cUSD'
+    const mnemonic = await generateMnemonic()
+    const { address: walletAddress } = await generateKeys(mnemonic)
+    await quickOnboarding(mnemonic) // ends on home screen
+    await fundWallet(SAMPLE_PRIVATE_KEY, walletAddress, token, amountPlusGas)
```
satish-ravi, line 44: `SAMPLE_PRIVATE_KEY` corresponds to the wallet address listed here right? https://github.com/valora-inc/wallet/blob/2d7ce14c9292bc78287f1db45dd5965fb3241a24/e2e/scripts/consts.ts#L1-L2
Is the intent that we fund the new address using this address and then the periodic job funds the wallet address from the faucet? Any reason to do this instead of directly using wallet address for tests?

satish-ravi (commented):
tests look good, just a couple of questions

e2e/src/usecases/FiatConnectTransferOut.js
MuckT: @satish-ravi the mnemonic and private key for the e2e test faucet secret. This keeps someone from draining our e2e test faucet quickly. It looks like the `SAMPLE_PRIVATE_KEY` corresponds to the `E2E_TEST_WALLET`.

e2e/src/FiatConnectTransferOut.spec.js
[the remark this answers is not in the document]
MuckT: @satish-ravi there's not a guarantee we'll get a fresh install at the start of these tests each test suite: good catch.

e2e/src/usecases/FiatConnectTransferOut.js
cajubelt: Flow of funds: `E2E_TEST_FAUCET` -> `E2E_TEST_WALLET` aka `SAMPLE_WALLET` via a script, then for each of these tests, `SAMPLE_WALLET` -> `new walletAddress` via my new helper function `fundWallet`.

cajubelt: I didn't investigate why, since that actually seemed like desirable behavior-- and even if it was a fluke from not reinstalling, we might want to build it out more fully in the future. Moreover, if we made the mock provider follow the spec more closely, we could run into issues like a [ResourceExists error](https://github.com/fiatconnect/specification/blob/main/fiatconnect-api.md#3431313-resourceexists) from submitting the same fiat account info multiple times for the same address. 

Testing first-time user flows with a brand new wallet each time seemed cleaner.

e2e/src/FiatConnectTransferOut.spec.js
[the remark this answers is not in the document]
cajubelt: strangely the tests seem to already do an uninstall/reinstall when they start, not sure why. If they didn't, I'd expect them to fail in the `quickOnboarding` stage if they run after any of the other tests.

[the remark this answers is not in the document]
cajubelt: I tried adding a beforeAll with uninstall/reinstall and the test began to fail. The cycle time for debugging these things is ingratiating. I'd prefer to move on and just fix if they fail later

PR #3015, opened 2022-10-20
test(e2e): settings tests

MuckT:
### Description

Adds two e2e tests:
- Change Language E2E test
- Change Currency E2E test

### Other changes

- Removed error from quickOnboarding catch block

### Tested
- Tested locally on Android
- Tested in CI with iOS and Android

### How others should test

- N/A

### Related issues

- Fixes RET-421

### Backwards compatibility

- Yes


e2e/src/usecases/Settings.js
```diff
@@ -23,11 +23,40 @@ export default Settings = () => {
     await element(by.id('ProfileEditName')).clearText()
     await element(by.id('ProfileEditName')).replaceText(`${randomName}`)
     await element(by.id('SaveButton')).tap()
-    await expect(element(by.text('Your name and picture were saved successfully.'))).toBeVisible()
+    await waitFor(element(by.text('Your name and picture were saved successfully.')))
+      .toBeVisible()
+      .withTimeout(1000 * 10)
```
jeanregisser, line 28: Why such a long timeout? 
If I remember correctly it only saves this locally now, so there shouldn't be any network delay.

e2e/src/usecases/Settings.js
```diff
@@ -23,11 +23,40 @@ export default Settings = () => {
...
     // TODO replace this with an ID selector
     await expect(element(by.text(`${randomName}`))).toBeVisible()
   })
+
+  it('Change Language', async () => {
+    await element(by.id('ChangeLanguage')).tap()
+    await element(by.id('ChooseLanguage/es-419')).tap()
+    await waitFor(element(by.id('SettingsTitle')))
+      .toHaveText('Configuración')
+      .withTimeout(1000 * 15)
```
jeanregisser, line 41: Why does it need specific timeout?

e2e/src/utils/utils.js
```diff
@@ -195,9 +195,7 @@ export async function quickOnboarding(mnemonic = SAMPLE_BACKUP_KEY) {
 
     // Assert on Wallet Home Screen
     await expect(element(by.id('SendOrRequestBar'))).toBeVisible()
-  } catch (e) {
-    console.error(e)
-  }
+  } catch {} // Don't throw an error just silently continue
```
jeanregisser, line 198: Why? 😄

jeanregisser (approved):
Awesome! 🚀

e2e/src/utils/utils.js
MuckT: Keeps the Detox logs from having errors that aren't really errors. In some places we try quickOnboarding, but quickly arrive at this catch block if emulator's previous test already completed onboarding.

e2e/src/usecases/Settings.js
MuckT: The `waitFor` is a maximum of 10 seconds I was running into some issues where the expect was firing too soon before the banner was displayed so opted for this instead.

e2e/src/usecases/Settings.js
MuckT: `waitFor` should have a timeout based on the [detox docs](https://github.com/nubank/detox/blob/master/docs/APIRef.waitFor.md#tobevisible).

PR #3016, opened 2022-10-20
fix: CPV in restore wallet flow with phone verification

kathaypacific:
### Description

The bug:
The wallet is created / restored at the last step of onboarding. If phone verification is required at onboarding, when the user arrives at the phone verification screen in the restore wallet flow, the account and signed message are not yet available (since they are programmed to be created after this screen) and therefore attempting to do central phone verification at this stage will fail. 

This PR:
- create the account after the restore screen. at this point, we have all the information needed to restore the account anyway. we can also verify the CPV status here, to decide to navigate to the phone verification screen or not.
- by creating the account after the seed phrase screen, users will have the wallet address and signed message needed for CPV.
- of course this introduces a delay on the seed phrase screen while we wait for the account to be created. previously this account creation is done asynchronously while navigating to the home screen, however the user will not see anything of value on the home screen while the account is being created so i think this delay is acceptable.

The one wrinkle:
- At the start of onboarding, we haven't created the account yet obviously. Therefore, there is no way to know the user's address to check if they are already verified / should go through phone verification. So it's not possible to determine the correct number of onboarding steps to be displayed in the screen header...if a user is already verified after we restore the wallet using the seed phrase in step 3/4, they will not need to complete step 4/4 (the phone verification step). The current flow may lead the user to think the app glitched and missed a step? Technically I can't think of a way around this, so may need to follow up with a UI change.

### Other changes

N/A

### Tested

Manually, during the restore wallet flow, you should be able to go through CPV. When restoring a wallet that is already verified, you should not see the CPV flow.

### How others should test

As above

### Related issues

N/A

### Backwards compatibility

Yes


src/import/saga.ts
```diff
@@ -159,9 +160,11 @@ export function* importBackupPhraseSaga({ phrase, useEmptyWallet }: ImportBackup
       ValoraAnalytics.track(AppEvents.redux_store_recovery_success, { account })
     }
     ValoraAnalytics.track(OnboardingEvents.wallet_import_success)
+    yield call(initializeAccountSaga)
```
kathaypacific, line 163: i'm awaitng the whole saga here since the `numberVerifiedCentrallySelector` depends on this saga finishing

src/account/saga.ts
```diff
@@ -79,7 +78,6 @@ export function* initializeAccountSaga() {
     ValoraAnalytics.track(OnboardingEvents.initialize_account_start)
     yield call(getOrCreateAccount)
     yield call(generateSignedMessage)
-    yield put(refreshAllBalances())
```
kathaypacific, line 82: i don't think this is needed, since the wallet home screen already dispatches this action

jeanregisser: I think we should leave this if we're not sure of all consequences with the different flows.

src/import/saga.ts
```diff
@@ -159,9 +160,11 @@ export function* importBackupPhraseSaga({ phrase, useEmptyWallet }: ImportBackup
...
+
+    const numberAlreadyVerifiedCentrally = yield select(numberVerifiedCentrallySelector)
     const skipVerification = yield select(skipVerificationSelector)
-    if (skipVerification) {
-      yield put(initializeAccount())
+    if (skipVerification || numberAlreadyVerifiedCentrally) {
```
jeanregisser, line 167: What do you think of covering this new case with a test too?

jeanregisser (approved):
Thanks! 👍

jeanregisser:
I think it's a better experience now that we wait.
I remember being confused not seeing the verified state right after restoring and landing on the home screen.

And about the incorrect steps count, I think it's fine too and it's a good idea to follow up with a UI change. Maybe we could assume all users restoring are already verified?
Fine either way.

src/import/saga.ts
kathaypacific: yup, added!

kathaypacific:
> I think it's a better experience now that we wait during the restore step. I remember being confused not seeing the verified state right after restoring and landing on the home screen.
> 
> And about the incorrect steps count, I think it's fine too and it's a good idea to follow up with a UI change. Maybe we could assume all users restoring are already verified? Fine either way.

I'll leave it like this for now, and ask Ellen / Nitya if they'd like some UI tweaks. I think your suggestion makes sense. I was also thinking we could give the phone verification screen another state like "all good, we've automatically restored your verification status!" so that there's still something that can be displayed for the last step. anyway, i'll check and add any follow up UI changes in a separate PR

PR #3021, opened 2022-10-20
fix(deep link): Remove duplicate block handling dynamic links on iOS

jacob-israel-turner:
### Description

Removes a duplicate block of code for handling Deep Links on iOS. I believe it was added via a bad merge.

When @kathaypacific was testing the analytics around the invite link, I believe she reported seeing some duplicate events. This is likely the cause.

### Other changes

N/A

### Tested

N/A

### How others should test

I would appreciate some help testing on iOS, as I don't have an iOS device to test on.

### Related issues

N/A

### Backwards compatibility

N/A

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -105,14 +105,6 @@ export class App extends React.Component<Props> {
         this.handleOpenURL({ url })
       )
 
-      if (Platform.OS === 'ios') {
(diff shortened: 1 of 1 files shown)


jeanregisser (approved):
🙌

kathaypacific (approved):
✨

PR #3022, opened 2022-10-20
feat: swap complete refinements

MuckT:
### Description

Adds an 'X' icon on the when swaps are completed successfully that takes users back to the home screen.

#### Screenshot
<p align="left">

</p>

### Other changes

Changed to use the default refresh indicator.

### Tested

Tested locally on Android 12 and iOS.

### How others should test

1. Complete a swap
2. Tap the 'X' in the to ~~right~~ left corner. 
3. Observe users is navigated to home feed screen.

### Related issues

- Fixes RET-447

### Backwards compatibility

Yes


src/swap/SwapExecuteScreen.tsx
```diff
@@ -147,6 +158,10 @@ export function SwapExecuteScreen() {
 
   return (
     <SafeAreaView style={styles.safeAreaView}>
+      <CustomHeader
```
kathaypacific, line 161: do we need a CustomerHeader here? couldn't it be a View that has 100% width with some padding and justifyContent flex-end, and the TopBarIconButton?

src/swap/SwapReviewScreen.tsx
```diff
@@ -163,7 +163,7 @@ export function SwapReviewScreen() {
           />
         }
       >
-        {shouldFetch && (
+        {shouldFetch && Platform.OS === 'ios' && (
```
kathaypacific, line 166: hmm, would be nice to minimise platform specific code where we can. why do we have a transparent refresh control for iOS? can we remove this block, and change the `tintColor` on the `RefreshControl` to one of our grays instead?

MuckT: Changing the default Refresh control makes sense, it'll look a bit different than the mocks, but I agree that's preferable to platform specific code.

kathaypacific:
hey @MuckT wanted to check how this one is going?

src/swap/SwapReviewScreen.tsx
MuckT: Ended up using the default refresh control in https://github.com/valora-inc/wallet/pull/3022/commits/8410823e475b27443f5323e95e597714bf7b91f1.

src/swap/SwapExecuteScreen.tsx
MuckT: I think it makes sense to keep the custom headers using the same component as it will be easier to manage changes to headers in the future. Looking at #3004 changes having the logic for custom headers spread out throughout components makes it harder to change across the app.

kathaypacific: hmm i agree only partly - if this were a screen header then for sure we should be using the header component. but this seems just a cross icon to me (like you'd see on a modal, since it doesn't have the traditional components you'd expect on a header, like a back button or informative text) so it feels strange to use a header component where we need to "hack" it by passing an empty View as the "leftComponent" :/

but i defer to your judgement, if you feel strongly that we should use a custom header component to render the cross icon then we can leave it, it's not such a big deal :)

kathaypacific (approved):

src/swap/SwapExecuteScreen.tsx
MuckT: The passing of an empty item is a bit strange; however, if I remember correctly this is the only screen in the app that has only a button on the right and the only screen with a back button on the right. @nityarao should we consider moving this close button to the left for consistency across the app?

MuckT:
@kathaypacific moved the 'X' to the left side after checking with Nitya.

kathaypacific (approved):

PR #3034, opened 2022-10-21
chore(ci): Use node 16 for the licenses check

jh2oman:
### Description

It looks like this one was just missed when we originally switched all the checks to node 16 in https://github.com/valora-inc/wallet/pull/2474/files


MuckT (approved):
👍

PR #3036, opened 2022-10-23
feat(fiatconnect): Add ability to navigate to review screen, re-fetching cached quotes

jophish:
### Description

Fixes ACT-474.

This PR adds in the ability to navigate to the `FiatConnectReview` screen via the new `FiatConnectReviewWrapper` screen, while automatically fetching a new quote from cached parameters.

There were a few different options for implementing this, but I ended up going with a small wrapper screen that handles showing a spinner and dispatching an existing action in order to refetch from a cached quote, before navigating to the Review screen. Some of the other options...
* Update `FiatConnectReview` screen to only _optionally_ take a quote, and add additional parameters to tell it to refetch from some particular cached quote params. Making the `quote` param optional to this component would have been a mess from a typing perspective, and would further complicate the already somewhat bloated review screen.
* Create a brand new action/saga to be triggered from the KYC accepted notification handler directly, which would be responsible for refetching the quote and navigating to the existing `FiatConnectReview` screen. This would have required a brand new saga/action, and is further complicated by the fact that we want to immediately show a spinner page while the quote is being refetched; not sure how we would do that without modifying an existing screen or adding a new one.
* There was a potential option of re-using the primary quote selection saga to handle this, though without larger changes to it, certain failure/edge cases could have led the user to some relatively unexpected behavior, considering that the logic in this PR is meant to be triggered after pressing a push notification/homecard.

The chosen solution reuses the quote refetching logic, which needed a slight refactor. Currently, when we refetch quotes, we already know what fiat account the user has selected. When fetching from a cached quote however, we will not have this luxury, and so must check to see if any of the quote(s) returned from the provider during the re-fetch match with an account the user has on file. As a result, `fiatAccount` is now an _optional_ parameter to `RefetchQuoteAction`. If it is not provided, we check cached fiat account uses after fetching the quote, check for a match between the quote's constraints and our cached accounts, and if a match exists, fetch the actual account information from the provider.

### Other changes

One inconsistency I noticed while making these changes was that when passing fiat account information through different screens in route params, they're typed as `ObfuscatedFiatAccountData`, while in actuality, they are _always_ of type `FiatAccount`. This worked, since `FiatAccount` is just an intersection of `ObfuscatedFiatAccountData` and a `providerId` field. I've updated our usage of fiat accounts in these places to be typed as `FiatAccount`, as it (probably?) should have been already.

### Tested

Unit tested.

### How others should test

Not integrated/triggered in-app currently.

### Related issues

- Fixes ACT-474

### Backwards compatibility

N/A.


src/fiatconnect/saga.ts
```diff
@@ -445,14 +459,54 @@ export function* _getSpecificQuote({
...
+    }
+  }
+
+  if (!selectedFiatAccount) {
+    throw new Error('Could not find appropriate fiat account on file')
+  }
+
+  const normalizedQuote = normalizedQuotes.find(
+    (q) => q.getFiatAccountType() === selectedFiatAccount!.fiatAccountType
+  )
```
cajubelt, line 499:
```suggestion
  const selectedFiatAccountType = selectedFiatAccount.fiatAccountType
  const normalizedQuote = normalizedQuotes.find(
    (q) => q.getFiatAccountType() === selectedFiatAccountType
  )
```

src/fiatconnect/saga.ts
```diff
@@ -430,13 +444,13 @@ export function* _getSpecificQuote({
   cryptoAmount,
   flow,
   providerId,
-  fiatAccountType,
+  fiatAccount,
```
cajubelt, line 447: could be nice to add a docstring on this function with a sentence on what it's for

src/fiatconnect/saga.ts
```diff
@@ -445,14 +459,54 @@ export function* _getSpecificQuote({
     providerIds: [providerId],
   })
   const normalizedQuotes = normalizeFiatConnectQuotes(flow, quotes)
-  const normalizedQuote = normalizedQuotes.find((q) => q.getFiatAccountType() === fiatAccountType)
+
+  // If no account was provided, we need to check cached accounts for a possible match, and fetch the obfuscated account
+  // from the selected provider.
+  let selectedFiatAccount = fiatAccount
+  if (!selectedFiatAccount) {
```
cajubelt, line 466: nit: could be a bit cleaner to deal with this case higher up, where this function gets called. Looks like `handleAttemptReturnUserFlow` already has its own logic for getting the fiat account, so this logic should only apply to `handleRefetchQuote`. It could also be put into its own helper function for readability and testability

src/fiatconnect/saga.ts
```diff
@@ -445,14 +459,54 @@ export function* _getSpecificQuote({
...
+    const cachedFiatAccounts: CachedFiatAccountUse[] = yield select(cachedFiatAccountUsesSelector)
+    // If defined, cachedFiatAccount is a Fiat Account whose providerId and fiatAccountType match one of our quotes.
```
cajubelt, line 468: eventually, we may allow users to link multiple accounts for the same provider and fiat account type. could be good to leave a note here indicating explicitly that that isn't supported right now. Or you could try supporting it, but that seems like a hassle.

cajubelt: seems like you'd need to cache both the quote and the fiat account ID the user was trying to create a transfer with.

src/fiatconnect/saga.ts
```diff
@@ -445,14 +459,54 @@ export function* _getSpecificQuote({
...
+    const cachedFiatAccount = cachedFiatAccounts.find((a) =>
+      normalizedQuotes.find(
+        (q) => q.getFiatAccountType() === a.fiatAccountType && providerId === a.providerId
```
cajubelt, line 471: what if q and a have the same fiat account type and different fiat account schemas? seems like a pretty important case to consider

cajubelt (commented):
Haven't reviewed the tests yet but wanted to get my comments to you sooner

src/fiatconnect/saga.ts
jophish: Yes; we cache the quote as soon as the user has submitted KYC data to IHL, before the user enters fiat account information, so it would make things a bit tricky. We could probably do something like cache the quote when KYC is posted, and update it with fiat account info if the user submits a fiat account while still in the same CICO flow... I'll leave a note saying multiple accounts are not yet supported here.

src/fiatconnect/saga.ts
jophish: D'oh -- that's a great point; thanks for the catch!

jophish: We actually don't save the fiat account schema used in `cachedFiatAccountUses`, which makes performing this check somewhat impossible at the moment..

src/fiatconnect/saga.ts
```diff
@@ -427,18 +439,90 @@ export function* _getQuotes({
...
+  normalizedQuotes,
+  fiatAccount,
+}: {
+  normalizedQuotes: FiatConnectQuote[]
+  fiatAccount: FiatAccount
+}) {
+  for (const normalizedQuote of normalizedQuotes) {
+    if (
+      normalizedQuote.getFiatAccountType() === fiatAccount.fiatAccountType &&
+      normalizedQuote.getFiatAccountSchema() === fiatAccount.fiatAccountSchema
```
satish-ravi, line 457: Probably an edge case, but if a quote returns multiple fiat account schemas for the same fiat account type, would we run into issues if their order is inconsistent across multiple quotes? We may have a saved fiat account with schema that is no longer the first in the new quote right?

src/fiatconnect/saga.ts
```diff
@@ -427,18 +439,90 @@ export function* _getQuotes({
   return quotes
 }
 
+/**
+ * Given a Fiat Account to use and a list of possible quotes to select, attempts to find a quote
+ * for which the given Fiat Account is allowed to be used. Returns the first match, or null if
+ * none is found.
+ **/
+export function _selectQuoteFromFiatAccount({
```
satish-ravi, line 447: nit: maybe this can be called `_selectQuoteMatchingFiatAccount`?

src/fiatconnect/ReviewScreenWrapper.tsx
```diff
@@ -0,0 +1,60 @@
...
+type Props = StackScreenProps<StackParamList, Screens.FiatConnectReviewWrapper>
+
+export default function FiatConnectReviewScreenWrapper({ route }: Props) {
+  const { providerId, kycSchema } = route.params
+  const dispatch = useDispatch()
+  const fiatConnectQuotesError = useSelector(fiatConnectQuotesErrorSelector)
+  const cachedQuoteParamsList = useSelector(cachedQuoteParamsSelector)
+
+  useEffect(() => {
+    const cachedQuoteParams = cachedQuoteParamsList?.[providerId]?.[kycSchema]
```
satish-ravi, line 25: Should we also check balances before fetching a quote? If in case the cached amount is no longer available?

src/fiatconnect/ReviewScreenWrapper.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+      <Provider store={store}>
+        <FiatConnectReviewScreenWrapper {...props} />
+      </Provider>
+    )
+    expect(navigate).toHaveBeenCalledTimes(1)
+    expect(navigate).toHaveBeenCalledWith(Screens.FiatExchange)
+    expect(store.dispatch).not.toHaveBeenCalled()
+  })
+
+  it('navigates to CICO screen on error during re-fetch', () => {
```
satish-ravi, line 38:
```suggestion
  it('navigates to CICO screen on error during quote re-fetch', () => {
```

src/fiatconnect/saga.ts
```diff
@@ -427,18 +439,90 @@ export function* _getQuotes({
...
+  normalizedQuotes,
+  fiatAccount,
+}: {
+  normalizedQuotes: FiatConnectQuote[]
+  fiatAccount: FiatAccount
+}) {
+  for (const normalizedQuote of normalizedQuotes) {
```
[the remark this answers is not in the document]
jophish: Yeah, I was also a bit confused at first; can clarify in a comment.

src/fiatconnect/saga.ts
jophish: Assuming we have a single fiat account on file for the provider, the ordering of quotes doesn't matter, since there will only ever be a single quote for each account type/schema pair. If we have multiple fiat accounts on file, it's possible that a different quote/fiat account pair will be selected if the ordering of the quotes is not guaranteed.

src/fiatconnect/ReviewScreenWrapper.tsx
jophish: We decided on not performing this check for now since it isn't strictly necessary; if a user proceeds to attempt to finalize the transfer, it will just fail and they'll receive an error message with an option to try again.

src/fiatconnect/ReviewScreenWrapper.tsx
```diff
@@ -0,0 +1,60 @@
...
+  fiatConnectQuotesErrorSelector,
+  cachedQuoteParamsSelector,
+} from 'src/fiatconnect/selectors'
+import variables from 'src/styles/variables'
+import { refetchQuote } from 'src/fiatconnect/slice'
+import { navigate } from 'src/navigator/NavigationService'
+
+type Props = StackScreenProps<StackParamList, Screens.FiatConnectReviewWrapper>
+
+export default function FiatConnectReviewScreenWrapper({ route }: Props) {
```
jh2oman, line 18: Can this file / component be renamed too `RefetchQuoteScreen` or something like that? The page itself has nothing tying it to the ReviewScreen, its the saga that eventually navigates there. Neither is the screen a wrapper.

src/fiatconnect/ReviewScreenWrapper.tsx
cajubelt: ^this was discussed in a pre refinement meeting on 10/13

src/fiatconnect/saga.ts
```diff
@@ -427,33 +439,131 @@ export function* _getQuotes({
...
+  return {
+    normalizedQuote,
+    selectedFiatAccount: fiatAccount,
+  }
 }
 
+/**
+ * Fetches a fiat account from a given provider according to a set of optional filters. Possible filters are
+ * fiatAccountId, fiatAccountType, and fiatAccountSchema. Multiple filters may be used in conjunction. If, after
+ * filtering, multiple possible accounts remain, a random one matching the given criteria is returned. Returns null
```
cajubelt, line 564:
```suggestion
 * filtering, multiple possible accounts remain, an arbitrary one matching the given criteria is returned. Returns null
```
it's not random, it's arbitrary

cajubelt: colloquially random and arbitrary are used the same way, but formally, they are quite different

src/fiatconnect/saga.ts
```diff
@@ -427,33 +439,131 @@ export function* _getQuotes({
...
-  return normalizedQuote
+  return {
+    normalizedQuote,
+    selectedFiatAccount: fiatAccount,
+  }
 }
 
+/**
+ * Fetches a fiat account from a given provider according to a set of optional filters. Possible filters are
+ * fiatAccountId, fiatAccountType, and fiatAccountSchema. Multiple filters may be used in conjunction. If, after
```
cajubelt, line 563:
```suggestion
 * Fetches a fiat account from a given provider according to a set of optional filters. Multiple filters may be used in conjunction. If, after
```

cajubelt: nit: I recommend against describing the parameters without `@param paramName:`, since the `@param` syntax lets us lint for cases where the parameters change and grow out of sync with the docstring

cajubelt: not sure if we have linting for that enabled, but I know it's possible because my IDE does it

cajubelt (approved):
just cosmetic stuff remaining I think

src/fiatconnect/ReviewScreenWrapper.tsx
jophish: Sure, definitely makes more sense.

PR #3040, opened 2022-10-24
feat(fiatconnect): Store Fiat Account Schema when caching Fiat Accounts

jophish:
### Description

We currently don't save the `fiatAccountSchema` field when storing cached Fiat Account uses in Redux; this introduces issues when determining compatibility between stored accounts and new quotes, since we're unable to determine if any stored account has a schema that matches a given quote.

Note that the migration blanks the `cachedFiatAccountUses` field in Redux. This won't _break_ anything, but users will have to re-enter their account once if they already had one on file.

### Other changes

N/A

### Tested

Updated unit tests.

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

Not backwards compatible; users will have to re-enter fiat accounts if they already had them linked.


jh2oman (approved):

cajubelt (commented):
Ok, so if a user tries to submit the same account they did previously, the FC spec says the server should respond with a 409. Does the app handle that elegantly? https://github.com/fiatconnect/specification/blob/main/fiatconnect-api.md#343131-failure

I don't see any new tests added that deal with that. Unless it's already covered, I suggest adding coverage for that case.

jophish:
@cajubelt It is already handled by the app; see [here](https://github.com/valora-inc/wallet/blob/main/src/fiatconnect/saga.ts#L262) for the check to handle the case when we get a `ResourceExists` error.

satish-ravi (approved):

PR #3044, opened 2022-10-25
chore(analytics): analytics properties for why a user selects a CICO option

cajubelt:
### Description

Adds properties to the `cico_providers_quote_selected` and `cico_providers_exchanges_selected` to help us answer questions about what's driving users' choice of CICO provider options with simple queries. By "simple queries", I mean just counting events that meet certain filter conditions-- as a proxy for this, think "something that can be done in MixPanel".

The following questions should now be answerable about a user selecting a provider or tapping the exchanges section on the Select Providers screen:

- was an estimated fee shown for that option?
- was it the lowest fee shown?
  - if not:
    - did the lowest fee option require KYC, and that quote did not?
    - did the lowest fee option have a different payment method?
- what was the fee (in crypto)?
- what was the total amount (in crypto)?
- how many provider options were available to choose between?
- were centralized exchanges shown as an option?
- what payment methods were available? was Coinbase Pay among them?

### Other changes

na

### Tested

added unit test for the new helper function which computes the analytics event properties of interest

### How others should test

na

### Related issues

na

### Backwards compatibility

na


src/fiatExchanges/utils.tsx
```diff
@@ -291,3 +294,78 @@ export const filterProvidersByPaymentMethod = (
 ) => {
   return externalProviders?.find((quote) => quote.paymentMethods.includes(paymentMethod))
 }
+
+export interface ProviderSelectionAnalyticsData {
+  centralizedExchangesAvailable: boolean
+  totalOptions: number
+  paymentMethodsAvailable: Record<PaymentMethod, boolean>
```
jh2oman, line 301: Any particular reason for this structure as opposed to just PaymentMethod[] ?

src/fiatExchanges/utils.tsx
```diff
@@ -291,3 +294,78 @@ export const filterProvidersByPaymentMethod = (
...
+}): ProviderSelectionAnalyticsData {
+  let lowestFeePaymentMethod: PaymentMethod | undefined = undefined
+  let lowestFeeProvider: string | undefined = undefined
+  let lowestFeeCryptoAmount: BigNumber | null = null
+  let lowestFeeKycRequired: boolean | undefined = undefined
+  const centralizedExchangesAvailable = !!centralizedExchanges && centralizedExchanges?.length > 0
+  const paymentMethodsAvailable: Record<PaymentMethod, boolean> = {
+    [PaymentMethod.Bank]: false,
+    [PaymentMethod.Card]: false,
+    [PaymentMethod.Coinbase]: false,
```
jh2oman, line 341: It doesn't look like coinbase can ever be true

src/fiatExchanges/utils.tsx
```diff
@@ -291,3 +294,78 @@ export const filterProvidersByPaymentMethod = (
...
+  return {
+    transferCryptoAmount,
+    paymentMethodsAvailable,
+    lowestFeePaymentMethod,
+    lowestFeeProvider,
+    lowestFeeKycRequired,
+    centralizedExchangesAvailable,
+    lowestFeeCryptoAmount: lowestFeeCryptoAmount?.toFixed(2),
+    // counts centralized exchanges as single option, since that's how they appear on the Select Providers screen
+    totalOptions:
```
jh2oman, line 366: should coinbase be counted here too?

src/fiatExchanges/utils.tsx
cajubelt: yeah, I think the SQL queries will be easier this way

src/fiatExchanges/utils.tsx
cajubelt: coinbase is treated as an "external provider" in the SelectProviders page, so it should be in the list of normalized quotes when it is offered.

src/fiatExchanges/utils.tsx
cajubelt: it is-- normalizedQuotes.length will cover it

src/fiatExchanges/utils.tsx
cajubelt: well, certain kinds of SQL queries anyway

src/fiatExchanges/utils.tsx
jh2oman: Unfortunately it is not because of these two lines of code

https://github.com/valora-inc/wallet/blob/main/src/fiatExchanges/quotes/normalizeQuotes.ts#L87

https://github.com/valora-inc/cloud-functions/blob/main/src/cico/fetchProviders.ts#L209

jh2oman: There's no quote to normalize for CoinbasePay

cajubelt: Oh ok, thanks for spotting that. I'll work on a fix and a regression for that issue.

Issue #3043, opened 2022-10-25
Freeze on initial load after creating new account


MuckT:
### Current behavior
On a fresh install after creating a new account the app is frozen until closed and reopened.

#### Video

https://user-images.githubusercontent.com/26950305/197886680-c83343ad-629c-47ae-857c-8b0b08869aec.mov

### Desired behavior
App should not be frozen on initial load.

### Steps to Reproduce
Prerequisites: A fresh install - not a reset install.
1. Complete onboarding flow.
2. On arrival of the home screen attempt to tap any buttons.
3. Observe screen in frozen.

#### Workarounds
- Close and relaunch the app.

### Versions
iOS ...v1.34.0 - v1.43.1

MuckT:
Fixed and an e2e test added to catch possible future regressions.

PR #3049, opened 2022-10-27
test(e2e): run e2e tests with cashInBottomSheet

MuckT:
### Description

Adjusts the e2e tests to run with the `cashInBottomSheet` enabled to catch defects such as #3043.

### Other changes

- TestID's added to cashInBottomSheet.

### Tested

Tested locally on iOS

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

Yes


e2e/src/utils/retries.js
```diff
@@ -24,7 +25,7 @@ export const launchApp = async (
   })
 }
 
-export const reloadReactNative = async () => {
+export const reloadReactNative = async (dismissBottomSheet = true) => {
```
kathaypacific, line 28: whether to dismiss the bottom sheet seems like a weird dependency for a RN reload function to have? is there anywhere where this is false? if not then should we just remove this function input?

aside: does it make sense to add the dismiss bottom sheet action to somewhere like `quickOnboarding`?

kathaypacific (approved):
LGTM!

e2e/src/utils/retries.js
MuckT: `dismissBottomSheet` is present in the quickOnboarding [here](https://github.com/valora-inc/wallet/pull/3049/files#diff-6fde14f432e5ec2dba3efc8a02ff8a389c62db2d6753bc2987d6064a218ce097).

Will take a look at removing the dismissBottomSheet. It would be great to avoid waiting 30 seconds on `reloadReactNative` if the bottom sheet isn't present.

jeanregisser: Side note, I feel we should stop using `<Modal />` as they conflict with the normal navigation. And only a single one can be displayed at the same time. Causing issue with shake to report a bug for instance.
But may be too much work now 😅 .

MuckT:
@jeanregisser and @kathaypacific questioning if we want to merge this into main. It's bumping up our e2e tests times by about 3 minutes.

kathaypacific:
> @jeanregisser and @kathaypacific questioning if we want to merge this into main. It's bumping up our e2e tests times by about 3 minutes.

if i understand correctly, the defect https://github.com/valora-inc/wallet/issues/3043 is related to this bottom sheet causing the app to freeze after a new account? then can we catch the defect by adding something to the create new account test, that ensure you can open a drawer menu or something?

if the purpose of this new e2e test is mostly testing that the cash in bottom sheet always works + bloats the run time of the suite, i'm happy to give it a miss since it's not business critical functionality. what do you think?

MuckT:
@kathaypacific to test #3043 we need to enable the bottom sheet in E2E tests because without the bottom sheet the issue is non repo, but this is increasing the test runtime.

PR #3053, opened 2022-10-28
feat: add react-navigation flipper plugin

MuckT:
### Description

Adds @react-navigation/devtools which enables a Flipper Plugin to view navigation state [docs](https://reactnavigation.org/docs/devtools#useflipper).
> Hooks from this package only work during development and are disabled in production. You don't need to do anything special to remove them from the production build.

#### Screenshot
Screen Shot 2022-10-28 at 8 58 51 AM

### Other changes

N/A

### Tested

### How others should test

Attempt to run the app in dev mode and enable to flipper plugin based on the steps in the `WALLET.md`

### Related issues

N/A

### Backwards compatibility

Yes


package.json
```diff
@@ -95,6 +95,7 @@
     "@react-native-picker/picker": "^2.4.2",
     "@react-navigation/compat": "^5.3.20",
     "@react-navigation/core": "^6.2.1",
+    "@react-navigation/devtools": "^6.0.10",
```
kathaypacific, line 98: should this be a dev dependency instead?

src/navigator/NavigatorWrapper.tsx
```diff
@@ -73,6 +74,9 @@ export const NavigatorWrapper = () => {
 
   const dispatch = useDispatch()
 
+  // @ts-ignore using a v6 navigation plugin with v5 types
```
kathaypacific, line 77: could we solve this by using v5 of the devtools instead?

package.json
MuckT: Since it's using a hook we can't conditionally call it in the `NavigatorWrapper.tsx`. The guide has it as a dependency instead of a dev dependency.

src/navigator/NavigatorWrapper.tsx
MuckT: It looks like there's some issues with how the navigationRef is setup using the v5 version. I'd rather wait until https://github.com/valora-inc/wallet/pull/3004 is merged then resolve any issues rather than trying to find a way to make v5 work only to update.

MuckT: The Flipper plugin is only available in v6 of [@react-navigation/devtools](https://github.com/react-navigation/react-navigation/tree/main/packages/devtools). It might make sense to merge https://github.com/valora-inc/wallet/pull/3004. Then come back to this.

kathaypacific (approved):

PR #3055, opened 2022-10-31
feat(fiatconnect): Sort fiat account schemas in quote constructor

jophish:
### Description

Currently, we generate one quote object per fiat account type. If multiple fiat account _schemas_ are available to use for that account type, we pick the first one we encounter in the list of available schemas. The ordering of allowed schemas is not guaranteed, which means that we could potentially pick a _different_ schema for some account type across multiple runs if several are available. This PR sorts the schemas before picking one, in order to avoid this problem.

### Other changes

N/A

### Tested

Unit tested.

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

Yes.

### DIFF
--- a/src/fiatExchanges/quotes/FiatConnectQuote.test.ts
+++ b/src/fiatExchanges/quotes/FiatConnectQuote.test.ts
@@ -63,6 +63,52 @@ describe('FiatConnectQuote', () => {
           })
       ).toThrow()
     })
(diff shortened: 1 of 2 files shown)


cajubelt (commented):
the logic looks fine, but I think when we discussed with product folks what should happen when account number and IBAN are both options, Laura wanted IBAN to have priority

PR #534, opened 2021-06-23
Tomm/e2e verification

MuckT:
### Description

This PR contains an e2e verification test for iOS & Android (WIP)

### Other changes

Added Twilio phone number reciever

### Tested

Tests were run locally on iOS
Test are in progress on Android

### How others should test

1. Add in appropriate API keys for Twilio in `packages/mobile/e2e/src/utils/twilio.js`
2. Run the verification spec from `packages/mobile` with `yarn test:e2e:ios -d -f Verify.spec.js `

### Backwards compatibility

This only impacts e2e tests for verification.

PR #3056, opened 2022-11-01
test(e2e): automated FiatConnect cash outs with KYC

anthonydaoud:
### Description

Currently, there is an e2e test for cashing out crypto via FiatConnect without KYC. This PR adds another e2e test for cash out with KYC. The test onboards a new wallet through Persona's Inquiry process before setting up a fiat account and transferring out crypto. The test only runs on iOS and the return user cash out flow is also tested. 

For this test to work, the `.env.enc` file in the `e2e` directory is updated with two new values: `MOCK_PROVIDER_BASE_URL` and `MOCK_PROVIDER_API_KEY`.

### Other changes

None

### Tested

N/A

### How others should test

N/A

### Related issues

- Fixes ACT-495

### Backwards compatibility

N/A


src/account/Persona.tsx
```diff
@@ -19,6 +20,7 @@ import { walletAddressSelector } from 'src/web3/selectors'
 import pjson from '../../package.json'
 
 const TAG = 'PERSONA'
+const E2E_TEMPLATE_ID = 'itmpl_5FYHGGFhdAYvfd7FvSpNADcC'
```
satish-ravi, line 23: Why do we need this? Isn't this the same template as in firebase?

e2e/src/usecases/FiatConnectTransferOut.js
```diff
@@ -74,62 +77,210 @@ async function submitTransfer() {
...
+      .toBeVisible()
+      .withTimeout(10000)
+    await element(by.type('PersonaPhoneNumberKit2.PhoneNumberTextField')).typeText('0123456789')
+    await element(by.text('Phone Number')).tap() // Tap away to unfocus from input
+    await element(by.text('Continue')).tap()
+
+    await element(by.text('Done')).tap() // End of Persona flow
+
+    // Check that Mock Provider info is defined
+    if (MOCK_PROVIDER_BASE_URL && MOCK_PROVIDER_API_KEY) {
```
satish-ravi, line 266: should the else part be an error?

e2e/src/usecases/FiatConnectTransferOut.js
```diff
@@ -74,62 +77,210 @@ async function submitTransfer() {
...
+    await element(by.label('shutter button')).tap()
+    await element(by.label('shutter button')).tap()
+    await element(by.label('shutter button')).tap()
+
+    // Name, number, address form
+    await waitFor(element(by.text('Phone Number')))
+      .toBeVisible()
+      .withTimeout(10000)
+    await element(by.type('PersonaPhoneNumberKit2.PhoneNumberTextField')).typeText('0123456789')
+    await element(by.text('Phone Number')).tap() // Tap away to unfocus from input
```
satish-ravi, line 260: do we need to focus out and then click continue? Any issue with clicking "Continue" directly?

satish-ravi (commented):
PR looks good, just a few minor things. Also, can you update the description to include what's updated in the `e2e/.env.enc` file?

src/account/Persona.tsx
anthonydaoud: Yes it is, but firebase is [not initialized](https://github.com/valora-inc/wallet/blob/main/src/firebase/saga.ts#L51) in e2e tests so I manually set the template id.

e2e/src/usecases/FiatConnectTransferOut.js
anthonydaoud: That part is for developers who are external to Valora who also want to run the e2e tests. I don't think they should fail because they don't have access to the encrypted env variables, which is akin to what is [currently done](https://github.com/valora-inc/wallet/blob/main/e2e/src/usecases/NewAccountPhoneVerification.js#L54) for the `NewAccountPhoneVerification` test.

e2e/src/usecases/FiatConnectTransferOut.js
anthonydaoud: The focus out tap is necessary because the keyboard is still active on the screen, which blocks the `Continue` button from being visible.

e2e/src/usecases/FiatConnectTransferOut.js
satish-ravi: could the entire test be skipped instead of partially running the test if this is the case?

.github/workflows/e2e-ios.yml
```diff
@@ -44,6 +44,12 @@ jobs:
         run: git diff --exit-code
       - name: Check E2E wallet balance
         run: NODE_OPTIONS='--unhandled-rejections=strict' yarn ts-node ./e2e/scripts/check-e2e-wallet-balance.ts
+      - name: Create iOS E2E .env File
+        env:
+          MOCK_PROVIDER_BASE_URL: ${{ secrets.MOCK_PROVIDER_BASE_URL }}
+          MOCK_PROVIDER_API_KEY: ${{ secrets.MOCK_PROVIDER_API_KEY }}
+        working-directory: e2e
+        run: echo MOCK_PROVIDER_BASE_URL=$MOCK_PROVIDER_BASE_URL >> .env && echo MOCK_PROVIDER_API_KEY=$MOCK_PROVIDER_API_KEY >> .env
```
satish-ravi, line 52: what's the purpose of this if the MOCK_PROVIDER config is also set in `e2e/.env.enc`?

src/account/Persona.tsx
satish-ravi: hmm. personally not a fan of having e2e specific stuff in components, especially given that this has to be in sync with what we set in firebase. Guess we can leave this as is for now and revisit if this becomes more dynamic (e.g., different templates for different kyc requirements)

cajubelt: perhaps a template ID getter could be put in the firebase or config module instead?

.github/workflows/e2e-ios.yml
anthonydaoud: The github workflow doesn't decrypt the .env.enc file, so the tests will fail in github if this manual .env setup is not included. This is based on an [earlier approach](https://github.com/valora-inc/wallet/pull/534/files#diff-3e103440521ada06efd263ae09b259e5507e4b8f7408308dc227621ad9efa31eR46) to adding config values.

e2e/src/usecases/FiatConnectTransferOut.js
anthonydaoud: Yes, just updated the test.

satish-ravi (approved):

.github/workflows/e2e-ios.yml
anthonydaoud: Yes that is my understanding.

PR #3064, opened 2022-11-03
fix: failed CPV during onboarding due to missing signed message

kathaypacific:
### Description

Context: https://valora-app.slack.com/archives/C044B8ULVC0/p1667413326982599

This PR introduces a loading state to the phone verification start button while the signed message is not available. Previously on android this can fail: 
- during onboarding the "initialise account" action is dispatched on this screen load
- the signed message is only generated after the account has been generated, but this is an asynchronous action
- if the user navigates to the next CPV step before the previous step is complete, there is no signed message for authentication with CPV, so the verification will always fail here.

### Other changes
N/A

### Tested

Manually + unit tests. 

### How others should test

On android device:
- create a new account + verify phone number. it should succeed

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/src/verify/VerificationStartScreen.tsx
+++ b/src/verify/VerificationStartScreen.tsx
@@ -256,6 +256,7 @@ const styles = StyleSheet.create({
   container: {
     flex: 1,
     backgroundColor: colors.onboardingBackground,
+    justifyContent: 'center',
   },
   scrollContainer: {
     flex: 1,
--- a/src/verify/VerificationStartScreen.tsx
+++ b/src/verify/VerificationStartScreen.tsx
@@ -26,6 +26,7 @@ import { navigate, navigateHome } from 'src/navigator/NavigationService'
 import { Screens } from 'src/navigator/Screens'
 import { TopBarTextButton } from 'src/navigator/TopBarButton'
 import { StackParamList } from 'src/navigator/types'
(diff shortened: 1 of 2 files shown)


M-shin (approved):
Verified on my pixel 5 (android 13) with both manual and auto-filled phone number.

MuckT (approved):
Was able to verify using a Galaxy 5 on Android 6.0.1.

src/verify/VerificationStartScreen.tsx
```diff
@@ -140,6 +142,17 @@ function VerificationStartScreen({
     }
   }, [route.params?.selectedCountryCodeAlpha2])
 
+  useEffect(() => {
+    const interval = setInterval(async () => {
+      if (!signedMessageCreated) {
```
jeanregisser, line 147: I'm surprised this works. `signedMessageCreated` is always `false` and this hook doesn't get called when it changes.
Doesn't it continue to call `await retrieveSignedMessage()` indefinitely (until the component is unmounted)?

kathaypacific: yeah good callout, i missed this. i remember thinking about the outdated reference to `signedMessageCreated` for a moment but clearly never bothered to go back and test. i updated the hook here https://github.com/valora-inc/wallet/pull/3071

PR #3067, opened 2022-11-04
feat: add loader to swap input when initially loading exchange rate

kathaypacific:
### Description

Sometimes it can take a few seconds to fetch the exchange rate, and it can be confusing to have no visual feedback to indicate that there is a background fetch happening.

Context https://valora-app.slack.com/archives/C018VLA3YAK/p1667520356022889?thread_ts=1667471639.586799&cid=C018VLA3YAK

This PR:
- add a loader _only_ when there is no `inputValue` on the text field. if the value is just being updated, we already have a light grey loading state.

https://user-images.githubusercontent.com/20150449/199971754-c62214f0-f54f-46c5-994a-36ec095678a6.mp4

### Other changes

N/A

### Tested

Manually + updated a unit test

### How others should test

On the swap screen, enter a "to" or "from" value. The other field should display a loader before the exchange rate is fetched.

### Related issues

Fixes RET-447

### Backwards compatibility

Yes


src/swap/SwapAmountInput.tsx
```diff
@@ -57,40 +58,53 @@ const SwapAmountInput = ({
...
-          }
-        />
+        {showInputLoader ? (
+          <View style={styles.loadingContainer}>
+            <ActivityIndicator
+              size="small"
+              color={Colors.greenBrand}
+              testID="SwapAmountInput/Loader"
+            />
+          </View>
```
jeanregisser, line 74: nit: no risk of things moving slightly while loading because the height is different?

jeanregisser (approved):
Awesome! 👩‍🎨

src/swap/SwapAmountInput.tsx
jeanregisser: ah yes, one way to avoid the jump is to keep rendering the normal `TextInput`, but with `opacity: 0` while loading, and render the loading container absolutely on top.

kathaypacific: ah, that is so clever! i didn't think of this. i like it, i updated the component to do this

PR #3072, opened 2022-11-07
feat: add loader while fetching verification status of selected contact for send flow

kathaypacific:
### Description

This PR:
- Add loading spinner on the right hand side of the contact list (send flow)
- On select contact, display loading spinner while the verification status is being fetched
- If contact address can be found OR if escrow is enabled, navigate to send screen as usual
- If no contact address can be found and escrow is disabled, display invite modal

Change to the flow on the send screen:
- on select contact, do not navigate. instead, set the selected contact
- when selected contact is set and there is no address, fetch the address
- update the verification status on finish fetch
- navigate or show modal depending on the verification status

Note that this removes the need for checking the verification status again on the send amount screen. If escrow is enabled, this change introduces a delay for all users (the verification status fetch would normally be done in the background while inputting the send amount) but we are expecting that escrow is disabled when this code is live. I only put the escrow check in there in case we need to delay CPV.

Recording: https://valora-app.slack.com/archives/C029Z1QMD7B/p1667831479772129

### Other changes

N/A

### Tested

Manually, + more unit tests

### How others should test

- Send to a contact that is on Valora. You should be able to navigate the send flow as normal
- Send to a contact that is not on Valora. You should see the invite modal.

### Related issues

- Fixes RET-439

### Backwards compatibility

Yes


jacob-israel-turner (approved):
LGTM!

src/send/ValidateRecipientAccount.tsx
```diff
@@ -272,7 +273,7 @@ export class ValidateRecipientAccount extends React.Component<Props, State> {
         : singleDigitInputValueArr.filter((entry) => /[a-f0-9]/gi.test(entry)).length === 4
 
     return (
-      <View style={styles.container}>
+      <SafeAreaView style={styles.container}>
```
kathaypacific, line 276: this change was unrelated (noticed it while testing the flow) but small so i kept it

PR #3074, opened 2022-11-08
fix(swap): allow input with decimal separator other than `.`

jeanregisser:
### Description

It was impossible to swap decimal amounts on devices using a decimal separator different from "."
This fixes it.

### Other changes

Also display the values on the Swap screen using the device's region decimal separator. Instead of "." for all regions.

### Tested

**before/after**

### How others should test

- Set your device region to a country using "," as decimal separator (France, Germany, etc)
- Go to the Swap screen
- Enter a decimal value using the country decimal separator
- Swap should work

### Related issues

- Fixes RET-485

### Backwards compatibility

Yes


kathaypacific (approved):
✨ mooi!!! 🚀

PR #3077, opened 2022-11-08
fix(fiatconnect): add a spinner when posting kyc after persona flow

satish-ravi:
### Description

Adds a new `personaInProgress` state variable, which is used for showing a spinner on `KycLanding` page. This is set to `true` when persona flow starts and `false` when persona finishes (for success, it waits until post kyc is complete before setting it to `false`). 

This also fixes a race condition which was caused by re-using `selectFiatConnectQuote` after persona completes. There could be delays in persona webhooks and the GET kyc status may return persona is not complete, which will result in the user landing on persona flow again. To fix this, we have a new saga `postKyc` that is invoked on persona success, which assumes persona is completed and attempts to post kyc to the provider directly.

First time flow:
https://user-images.githubusercontent.com/5062591/200665405-3b8c1815-4779-4ee7-bb04-69d300666f52.mp4

Expired flow with bank account saved:
https://user-images.githubusercontent.com/5062591/200665458-739698c3-2ab2-48d7-a98a-d96b7b861ba2.mp4

Persona flow cancel:
https://user-images.githubusercontent.com/5062591/200665592-5d5fb3a1-60d1-49e5-a068-d4620664f74d.mp4

### Other changes

N/A

### Tested

Unit tests, manual

### How others should test

with an account with no kyc / fiat account with mock-provider: 
1. Start withdraw flow with an amount with last digit odd number (e.g., 0.1) to trigger kyc with test provider
2. Select mock-provider, which should land on the "Setup Bank Account" page with step 1 enabled
3. Go through persona flow
4. On completing persona flow, ensure a spinner is shown before step 2 is enabled
5. Go through step 2 and enter bank details and ensure it lands on KycPending screen

Return user flow with kyc expired
1. Set kyc status for test-provider to expired for the same account (using curl or directly on firebase)
2. Start a withdraw flow as above and ensure "KycExpired" screen is shown
3. Select "Verify Again", which should land on the "Setup Bank Account" page with step 1 enabled
4. Go through persona flow
5. On completing persona flow, ensure a spinner is shown before navigating to KycPending screen

### Related issues

- Fixes ACT-519

### Backwards compatibility

yes


src/fiatconnect/saga.ts
```diff
@@ -711,6 +715,60 @@ export function* handleSelectFiatConnectQuote({
...
+        kycSchema,
+        cachedQuoteParams: {
+          cryptoAmount: quote.getCryptoAmount(),
+          fiatAmount: quote.getFiatAmount(),
+          flow: quote.flow,
+          cryptoType: quote.getCryptoType(),
+          fiatType: quote.getFiatType(),
+        },
+      })
+    )
```
satish-ravi, line 745: These two lines are duplicated in `handleFiatConnectQuote` and does exactly the same for the case when persona is approved, but kyc is missing on provider. I didn't think it was worth the effort to share the code, since both needs to fire a different "completion" action (personaFinished vs selectFiatConnectQuoteCompleted) at the end

src/fiatconnect/KycLanding.tsx
```diff
@@ -43,6 +45,16 @@ export interface Props {
 
 export default function KycLanding(props: StackScreenProps<StackParamList, Screens.KycLanding>) {
   const { quote, flow, step, personaKycStatus } = props.route.params
+  const personaInProgress = useSelector(personaInProgressSelector)
```
satish-ravi, line 48: If this state changes, I believe it cause the kycAgreement checkmark to be unset always. Should we pull the kycAgreement state to this component?

src/fiatconnect/saga.ts
```diff
@@ -711,6 +715,60 @@ export function* handleSelectFiatConnectQuote({
...
+    // clear persona status
+    yield put(personaFinished())
+  } catch (error) {
+    // Error while attempting to post to kyc or selecting fiat account
+    Logger.debug(
+      TAG,
+      `handlePostKyc* Error while attempting to post kyc for provider ${quote.getProviderId()}`,
+      error
+    )
+    yield put(personaFinished())
```
jh2oman, line 758: I'm pretty sure this would cause the KycLanding screen to flash in the non-spinning state unless you move it to be after the navigate and a 500 delay like the other places.

src/fiatconnect/KycLanding.test.tsx
```diff
@@ -65,31 +64,60 @@ describe('KycLanding', () => {
...
+    })
+    it('dispatches post kyc when persona succeeds', () => {
+      const { getByTestId } = render(
+        <Provider store={store}>
+          <KycLanding {...props} />
+        </Provider>
       )
-      jest.restoreAllMocks()
+      fireEvent.press(getByTestId('checkbox'))
+      fireEvent.press(getByTestId('PersonaSuccessButton'))
```
jh2oman, line 86: Could you add a comment to each of these explaining? Its kind of hacky and took me a bit to understand what was going on
```suggestion
      // 'PersonaSuccessButton' is a mock that calls onSuccess when pressed
      fireEvent.press(getByTestId('PersonaSuccessButton'))
```

src/fiatconnect/saga.ts
jophish: +1, also with the call above.

src/fiatconnect/saga.ts
```diff
@@ -86,6 +88,8 @@ const TAG = 'FiatConnectSaga'
 
 const KYC_WAIT_TIME_MILLIS = 3000
 
+const PERSONA_COMPLETION_STATUSES = new Set([PersonaKycStatus.Approved, PersonaKycStatus.Completed])
```
jophish, line 91: Should this be used somewhere in this PR?

src/fiatconnect/KycLanding.tsx
jophish: Does it matter? Maybe ask Nitya to see if it's a consideration? Otherwise, I would vote to just keep it as it.

src/fiatconnect/KycLanding.test.tsx
```diff
@@ -62,34 +62,64 @@ describe('KycLanding', () => {
...
+      expect(store.dispatch).toHaveBeenCalledWith(personaFinished())
+    })
+    it('dispatches persona finished when persona flow is canceled', () => {
+      const { getByTestId } = render(
+        <Provider store={store}>
+          <KycLanding {...props} />
+        </Provider>
+      )
+      fireEvent.press(getByTestId('checkbox/unchecked'))
+      fireEvent.press(getByTestId('PersonaCancelButton'))
```
satish-ravi, line 108:
```suggestion
      // 'PersonaCancelButton' is a mock that calls onCanceled when pressed
      fireEvent.press(getByTestId('PersonaCancelButton'))
```

src/fiatconnect/KycLanding.tsx
satish-ravi: chatted with Nitya, and updated to set the checkbox to true if step 1 is disabled

jh2oman (approved):

jh2oman (approved):

PR #3086, opened 2022-11-10
feat(fiatconnect): Add Mobile Money to Link Account Screen

anthonydaoud:
### Description

Currently, the link account screen is hardcoded to bank account values. This PR updates the link account screen so that it shows the correct text when a Mobile Money account type is selected. The KYC landing screen is also updated to have the correct account type in the screen header. 

MobileMoneyLinkAccountScreen                        MobileMoneyKycLanding

### Other changes

None

### Tested

Unit tests added to verify the correct text is shown on the link account screen.

### How others should test

N/A, Mobile Money is not yet supported in the wallet.

### Related issues

- Fixes ACT-517

### Backwards compatibility

N/A


src/fiatconnect/LinkAccountScreen.tsx
```diff
@@ -120,8 +129,11 @@ FiatConnectLinkAccountScreen.navigationOptions = ({
       }}
     />
   ),
-  // NOTE: title should be dynamic when we support multiple fiat account types
-  headerTitle: i18n.t('fiatConnectLinkAccountScreen.bankAccount.header'),
+  headerTitle: i18n.t(
+    `fiatConnectLinkAccountScreen.${convertAccountTypeToTranslation(
+      route.params.quote.getFiatAccountType()
+    )}.header`
```
jh2oman, line 135: As a norm we should try to avoid making strings parameterized like this. you can have a similar function, but the strings should be written out in full like `fiatConnectLinkAccountScreen.mobileMoney.header` and `fiatConnectLinkAccountScreen.bankAccount.header`. This makes them easier to find when searching for the code and potentially do in an automated way in the future.

src/fiatconnect/KycLanding.tsx
```diff
@@ -68,8 +69,11 @@ KycLanding.navigationOptions = ({
       }}
     />
   ),
-  // NOTE: title should be dynamic when we support multiple fiat account types
-  headerTitle: i18n.t('fiatConnectLinkAccountScreen.bankAccount.header'),
+  headerTitle: i18n.t(
+    `fiatConnectLinkAccountScreen.${convertAccountTypeToTranslation(
+      route.params.quote.getFiatAccountType()
```
jh2oman, line 74: same string comment

src/fiatconnect/LinkAccountScreen.tsx
```diff
@@ -72,10 +73,18 @@ export function LinkAccountSection(props: {
 
   return (
     <SafeAreaView style={styles.content}>
-      <Text style={styles.title}>{t('fiatConnectLinkAccountScreen.bankAccount.bodyTitle')}</Text>
+      <Text style={styles.title}>
+        {t(
+          `fiatConnectLinkAccountScreen.${convertAccountTypeToTranslation(
+            quote.getFiatAccountType()
```
jh2oman, line 79: same string comment

src/fiatconnect/LinkAccountScreen.tsx
```diff
@@ -72,10 +73,18 @@ export function LinkAccountSection(props: {
...
+          )}.bodyTitle`
+        )}
+      </Text>
       <Text testID="descriptionText" style={styles.description}>
         <Trans
-          i18nKey={'fiatConnectLinkAccountScreen.bankAccount.description'}
+          i18nKey={`fiatConnectLinkAccountScreen.${convertAccountTypeToTranslation(
+            quote.getFiatAccountType()
```
jh2oman, line 86: same string comment

src/fiatconnect/LinkAccountScreen.tsx
anthonydaoud: Good point, I've updated the code so it keeps the string intact.

src/fiatconnect/index.ts
```diff
@@ -47,15 +47,35 @@ function convertToFiatConnectCryptoCurrency(cicoCurrency: CiCoCurrency): CryptoT
...
+  description: string
+  header: string
+}
+
 /**
  * Small helper function to map the account type to its name used in the base translation
  * file.
  */
-export function convertAccountTypeToTranslation(accountType: FiatAccountType): string {
+export function accountTypeToLinkAccountTranslations(
```
jh2oman, line 60: nit: I think this function would probably be better placed in the LinkAccountScreen.tsx since its use is so closely tied to that component. 

Nit: also if you do move it to that file you can probably make the function name something less verbose like `getTranslationStrings`

jh2oman (approved):

PR #3091, opened 2022-11-11
fix(fiatconnect): Get quotes using the value the user inputs

jh2oman:
### Description

This change calls fiatconnect getQuotes using fiatAmount or cryptoAmount depending on which one the user originally inputed when trying to Add/Withdraw.

### Tested

Added unit tests. Saw logs where correct amount was being sent to valora-rest-api

### How others should test

N/A, only changes cash-in and its not ready yet.

### Related issues

https://linear.app/valora/issue/ACT-403/update-selectproviders-screen-for-transfers-in

### DIFF
--- a/src/fiatconnect/index.ts
+++ b/src/fiatconnect/index.ts
@@ -6,6 +6,7 @@ import {
   QuoteErrorResponse,
   QuoteResponse,
 } from '@fiatconnect/fiatconnect-types'
+import { isUserInputCrypto } from 'src/fiatExchanges/FiatExchangeAmount'
 import { CICOFlow } from 'src/fiatExchanges/utils'
 import { LocalCurrencyCode } from 'src/localCurrency/consts'
 import { getPassword } from 'src/pincode/authentication'
@@ -104,6 +105,7 @@ export type QuotesInput = {
   localCurrency: LocalCurrencyCode
   digitalAsset: CiCoCurrency
   cryptoAmount: number
+  fiatAmount: number
   country: string
   address: string
 }
@@ -132,6 +134,7 @@ export async function getFiatConnectQuotes(
     localCurrency,
     digitalAsset,
     cryptoAmount,
+    fiatAmount,
     country,
     flow,
     address,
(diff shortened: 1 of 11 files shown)


src/fiatExchanges/utils.tsx
```diff
@@ -291,3 +291,6 @@ export const filterProvidersByPaymentMethod = (
 ) => {
   return externalProviders?.find((quote) => quote.paymentMethods.includes(paymentMethod))
 }
+
+export const isUserInputCrypto = (flow: CICOFlow, currency: Currency | CiCoCurrency): boolean =>
```
cajubelt, line 295: I know this isn't new, but I don't actually understand this logic. Why must a user designate an amount in crypto when purchasing Celo? Would be nice if there was a docstring explaining this

src/fiatconnect/index.ts
```diff
@@ -142,7 +144,8 @@ export async function getFiatConnectQuotes(
   const quoteParams: CreateQuoteParams = {
     fiatType,
     cryptoType,
-    cryptoAmount: cryptoAmount.toString(),
+    ...(isUserInputCrypto(flow, digitalAsset) && { cryptoAmount: cryptoAmount.toString() }),
+    ...(!isUserInputCrypto(flow, digitalAsset) && { fiatAmount: fiatAmount.toString() }),
```
cajubelt, line 148:
```suggestion
    ...(isUserInputCrypto(flow, digitalAsset) ? { cryptoAmount: cryptoAmount.toString() } :  { fiatAmount: fiatAmount.toString() }),
```
code golf (more DRY)

cajubelt: probly not formatted right but you get the idea

cajubelt (approved):

src/fiatExchanges/utils.tsx
jh2oman: I don't either :) 

I brought it up to Nitya & Laura. They plan on discussing this week https://valora-app.slack.com/archives/C03D3K379FE/p1668122181856179

PR #3099, opened 2022-11-15
fix: dynamic links not persisted on Android after app download sometimes

kathaypacific:
### Description

A user who does not have Valora should be directed to the play store after consuming a deep link on Android. It has been observed that the dynamic link is not passed to the app unless the user navigates directly from the play store to the app. This seems to be because we are not allowing Firebase to process the initial url on app start on Android.

This PR changes the deep link handling:
- firebase always handles the initial link for dynamic links
- clevertap should ignore firebase dynamic links so that links are not double handled

### Other changes

I saw that deep links on Android are being handled twice - once by `Linking` and once by Clevertap. It doesn't seem to be causing problems but could introduce weird bugs? I added a class property to prevent duplicate handling.

### Tested

Given a dynamic link, on iOS and Android:
1. consume the link, and proceed to the app store
2. download / side load the app manually
3. exit the app store, do something else on your device
4. on opening the newly installed app, the `opened_via_invite_url` event should be dispatched with the expected inviter

Given a normal deep link, on iOS and Android where the app already exists:
1. consume the link
2. the link should redirect you to the expected location

Given a normal deep link, on iOS and Android where the app does not exist:
1. consume the link, and proceed to the app store
2. download / side load the app manually
4. exit the app store, do something else on your device
5. on opening the newly installed app, the link should redirect you as expected

### How others should test

As above

### Related issues

- Fixes RET-494

### Backwards compatibility

Y


jacob-israel-turner (approved):

jeanregisser (approved):
LGTM 👍 

If this continues to be an issue we should probably add some unit tests.

PR #3101, opened 2022-11-15
fix: prevent loader from appearing on send screen for recent addresses

kathaypacific:
### Description

Context: https://valora-app.slack.com/archives/C018VLA3YAK/p1668522773243539

This PR fixes the issue where the right hand side loader displays on the send screen for all contacts that have addresses only.

### Other changes

Vertically center the receipient details even when there is one line.

Before: Simulator Screen Shot - iPhone 11 - 2022-11-15 at 18 55 14

After: Simulator Screen Shot - iPhone 11 - 2022-11-15 at 19 05 15

### Tested

Manually

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

Yes


src/recipients/RecipientPicker.tsx
```diff
@@ -92,6 +92,7 @@ function RecipientPicker(props: RecipientProps) {
 
   const isFetchingVerificationStatus = (recipient: Recipient) => {
     return (
+      props.selectedRecipient &&
       recipient.e164PhoneNumber === props.selectedRecipient?.e164PhoneNumber &&
```
kathaypacific, line 96: this line caused the loader to appear - for contacts with address only, the `e164PhoneNumber` is undefined. same as the `selectedRecipient.e164PhoneNumber` since `selectedRecipient` is undefined when the screen first loads and the user hasn't selected a contact to send to

src/recipients/RecipientPicker.tsx
```diff
@@ -92,7 +92,8 @@ function RecipientPicker(props: RecipientProps) {
 
   const isFetchingVerificationStatus = (recipient: Recipient) => {
     return (
-      recipient.e164PhoneNumber === props.selectedRecipient?.e164PhoneNumber &&
```
kathaypacific, line 95: this line caused the loader to appear - for contacts with address only, the e164PhoneNumber is undefined. same as the selectedRecipient.e164PhoneNumber since selectedRecipient is undefined when the screen first loads and the user hasn't selected a contact to send to

MuckT (approved):
Some snapshot tests are still failing, but otherwise LGTM. Tested using the recent address of `0xe5F5363e31351C38ac82DBAdeaD91Fd5a7B08846`.

jeanregisser (approved):
Thanks 🚀 

Too bad we don't have unit tests on RecipientPicker 😅

PR #3110, opened 2022-11-16
chore: skip ASv1 CPV migration for users who have no signed message

kathaypacific:
### Description

For users who created their accounts before app version 1.32 and have never entered their pincode, they will not have a signed message to use for authentication. This PR skips the CPV migration for these users, since it will always fail.

### Other changes

N/A

### Tested

Unit tests

### How others should test

N/A

### Related issues

Fixes RET-490

### Backwards compatibility

Yes


jeanregisser (approved):
🚀

PR #3114, opened 2022-11-16
feat(fiatconnect): handle transfer in requests

jh2oman:
### Description

Refactored `handleCreateFiatConnectTransfer` to create two helper functions 
* `_initiateTransfer` for starting a fiatconnect transfer regardless of in/out
* `_initiateTransaction` for initiating the transaction on blockchain which only happens for transfers out

Then added some small functionality to only call `_initiateTransaction` for transfers out.

#### Video

Ignore the copy on the Review screen. It is not ready yet and in  a separate PR

https://user-images.githubusercontent.com/8432644/202314313-2a7c2171-6298-4eff-b640-219aefc28bcb.mp4

### Tested

Added quite a few more tests and improved the existing ones. Also tested out the flow on device

### How others should test

N/a

### Related issues

https://linear.app/valora/issue/ACT-406/initiate-transfer-in


src/fiatconnect/saga.ts
```diff
@@ -893,97 +893,120 @@ export function* handleFetchFiatConnectProviders() {
...
+  Logger.info(TAG, `${transferType} succeeded`, JSON.stringify(transferResult))
+  return transferResult
+}
 
-      const transferResult = result.unwrap()
+/**
+ * Initiates a transaction on-chain to send funds to a specific address. This is intended to be used for
+ * cashing out after a transfer has been initiated with a provider
+ **/
+export function* _initiateTransaction({
```
jophish, line 939: nit: make the naming a bit more fiatconnect-specific? been digging into our methods around sending TXs recently (of which there are many!) and they all have very similar names, so good to disambiguate.

src/fiatconnect/saga.ts
```diff
@@ -893,97 +893,120 @@ export function* handleFetchFiatConnectProviders() {
   }
 }
 
-export function* handleCreateFiatConnectTransfer({
+/**
+ * Initiates a fiatconnect transfer using the fiatconnect SDK to call a transfer/out or transfer/in endpoint
+ * for a provider.
+ **/
+export function* _initiateTransfer({
```
jophish, line 900: nit: same comment with naming; `_initateTransfer` and `_initiateTransaction` are pretty smiliar names

src/fiatconnect/saga.ts
```diff
@@ -893,97 +893,120 @@ export function* handleFetchFiatConnectProviders() {
...
-      )
-    } catch (err) {
-      Logger.error(TAG, 'Transfer out failed..', err)
-      yield put(createFiatConnectTransferFailed({ flow, quoteId }))
-    }
-  } else {
-    throw new Error('not implemented')
+    )
+  } catch (err) {
+    Logger.error(TAG, `Transfer for ${flow} failed..`, err)
```
jophish, line 1008: Should we log an analytics event here on error? There are code paths in the callees of this method that could throw but don't send an analytics event.

jh2oman: mmm, good catch. I'll move the Analytics error log here.

src/analytics/Properties.tsx
```diff
@@ -1116,8 +1116,13 @@ interface FiatExchangeEventsProperties {
...
     flow: CICOFlow
   }
+  [FiatExchangeEvents.cico_fc_transfer_error]: {
+    error?: string
+    provider: string
+    flow: CICOFlow
+  }
   [FiatExchangeEvents.cico_fc_transfer_success]: {
-    txHash?: string
+    txHash: string | null
```
satish-ravi, line 1125: Not directly related to this PR, but we have other events from the status screen that also optionally take txHash (`cico_fc_transfer_success_complete` and `cico_fc_transfer_success_view_tx`). Should we change the type for those as well?

src/analytics/Properties.tsx
```diff
@@ -1116,8 +1116,13 @@ interface FiatExchangeEventsProperties {
     provider: string
     flow: CICOFlow
   }
+  [FiatExchangeEvents.cico_fc_transfer_error]: {
+    error?: string
```
satish-ravi, line 1120: Should error be required here?

src/fiatconnect/saga.ts
```diff
@@ -893,97 +893,128 @@ export function* handleFetchFiatConnectProviders() {
   }
 }
 
-export function* handleCreateFiatConnectTransfer({
+/**
+ * Initiates a fiatconnect transfer using the fiatconnect SDK to call a transfer/out or transfer/in endpoint
+ * for a provider.
+ **/
+export function* _initiateFiatConnectTransferWithProvider({
```
satish-ravi, line 900: nit: Would `_initiateFiatConnectTransferWithProvider` suffice? I saw Joe's earlier comment, but I think "FiatConnect" is redundant here since this is in the fiatconnect/saga file and the `withProvider` adds context that this is something to do with FC provider.

src/fiatconnect/saga.ts
```diff
@@ -893,97 +893,128 @@ export function* handleFetchFiatConnectProviders() {
...
+  Logger.info(TAG, `${transferType} succeeded`, JSON.stringify(transferResult))
+  return transferResult
+}
 
-      const transferResult = result.unwrap()
+/**
+ * Initiates a transaction on-chain to send funds to a specific address. This is intended to be used for
+ * cashing out after a transfer has been initiated with a provider
+ **/
+export function* _initiateFiatConnectTx({
```
satish-ravi, line 939: nit: related to above. Maybe `_initiateSendTxToProvider`?

src/fiatconnect/saga.ts
```diff
@@ -893,97 +893,128 @@ export function* handleFetchFiatConnectProviders() {
...
+/**
+ * Initiates a fiatconnect transfer using the fiatconnect SDK to call a transfer/out or transfer/in endpoint
+ * for a provider.
+ **/
+export function* _initiateFiatConnectTransferWithProvider({
   payload: params,
 }: ReturnType<typeof createFiatConnectTransfer>) {
   const { flow, fiatConnectQuote, fiatAccountId } = params
   const quoteId = fiatConnectQuote.getQuoteId()
+  const transferType = flow === CICOFlow.CashIn ? 'transferIn' : 'transferOut'
```
satish-ravi, line 905: nit: maybe call this `transferFn` or `transferFnName`? Took me a couple of reads of the call method below to understand this was the function name

src/fiatconnect/saga.ts
satish-ravi: what are the other paths that could throw? We'd end up logging multiple events in some cases right? If for example `transfer_api_error` or `tx_error` is logged.

jh2oman: A runtime error in `buildAndSendPayment` or maybe an error setting up `fiatConnectClient` ?

I thought it definitely isn't ideal that we log two events in some (most) cases. But since this is such an important step in the flow I thought it was acceptable so that we cover all possible cases to be able to help us debug.

src/analytics/Properties.tsx
jh2oman: fixed

src/fiatconnect/TransferStatusScreen.tsx
```diff
@@ -118,7 +118,7 @@ function SuccessSection({
     ValoraAnalytics.track(FiatExchangeEvents.cico_fc_transfer_success_complete, {
       flow,
       provider,
-      txHash: fiatConnectTransfer.txHash ?? undefined,
+      txHash: fiatConnectTransfer.txHash ?? null,
```
satish-ravi, line 121:
```suggestion
      txHash: fiatConnectTransfer.txHash,
```
I think this should work since the state variable is also `string | null`

src/fiatconnect/saga.ts
```diff
@@ -893,97 +893,120 @@ export function* handleFetchFiatConnectProviders() {
...
+  if (error) {
+    ValoraAnalytics.track(FiatExchangeEvents.cico_fc_transfer_tx_error, {
+      flow: CICOFlow.CashOut,
+      error: error.message,
+      transferAddress: transferAddress,
+      provider: fiatConnectQuote.getProviderId(),
+    })
+    throw error
+  }
+  Logger.info(TAG, 'Completed transfer out transaction..')
```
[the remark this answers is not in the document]
satish-ravi: bump on this

satish-ravi (approved):

PR #3115, opened 2022-11-16
Require pin blocklist at the module scope

silasbw:
This is to facilitate some other work on module resolution. I'm running into issues with this specific change.


silasbw:
Well, it looks like there's some state in my working directory that's impacting this success of `yarn test` locally.  With a fresh clone, `yarn test` passes.

jeanregisser (approved):

PR #3116, opened 2022-11-17
fix(CPV): empty verification id when sms arrives very quickly

kathaypacific:
### Description

Context: https://valora-app.slack.com/archives/C044B8ULVC0/p1668675675593179

On the verify sms screen, basically there is a race condition between receiving the response from `verifyPhoneNumber` (specifically, capturing the `verificationId`) and receiving the SMS code which kicks off the `verifySMSCode` request. If the SMS code is received and processed first, the lack of `verificationId` causes the request to fail.

This PR waits for the `verificationId` before sending off the SMS verification request.

### Other changes

I noticed a lint warning for `await act` saying that `await` has no effect on `act`. which i found weird, since you can't remove the `await` and still get the same test result. i still wanted to fix the warning through and ended up using `await waitFor` instead of `await act(() => { ... flushMicroservicesQueue() }`, it seems more right.

### Tested

Manually added a delay for `verifyPhoneNumber` and verified that the `verifySMSCode` is not called until after the `verifyPhoneNumber` response is processed.

+ added unit test

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

Yes


src/verify/VerificationCodeInputScreen.test.tsx
```diff
@@ -131,6 +126,38 @@ describe('VerificationCodeInputScreen', () => {
     expect(navigate).toHaveBeenCalledWith(Screens.OnboardingSuccessScreen)
   })
 
+  it('waits for the verificationId to be captured before verifying sms', async () => {
+    mockFetch.mockImplementation(async (url?: string | Request) => {
+      if (url === networkConfig.verifyPhoneNumberUrl) {
+        await sleep(1000) // some arbitrary network delay
```
jeanregisser, line 132: nit: does this increase this unit test run by 1 sec, or are advancing timers to avoid this?

jeanregisser (approved):
Our users with fast SMS delivery thank you 🚀

src/verify/VerificationCodeInputScreen.test.tsx
kathaypacific: we're advancing timers :D this test doesn't use real timers

PR #3117, opened 2022-11-17
feat(fiatconnect): add mobile money to fiat details screen

anthonydaoud:
### Description

Currently, the Fiat Details screen does not have any support for a Mobile Money account. This PR allows Mobile Money accounts to be added on the Fiat Details screen. There isn't much phone number validation yet, just a check that the number is at most 15 digits as per the [international standards](https://en.wikipedia.org/wiki/E.164).

https://user-images.githubusercontent.com/20508929/202808974-e5ebbd04-e7c0-4849-b9e5-5cb8417a0a01.mp4

### Other changes

N/A

### Tested

Added unit tests.

### How others should test

N/A, Mobile Money is not yet supported in the wallet.

### Related issues

- Fixes ACT-516

### Backwards compatibility

N/A


src/fiatconnect/fiatAccountSchemas/mobileMoney.ts
```diff
@@ -0,0 +1,73 @@
...
+    country: { name: 'country', value: implicitParams.country },
+    fiatAccountType: { name: 'fiatAccountType', value: FiatAccountType.MobileMoney },
+    accountName: {
+      name: 'accountName',
+      computeValue: ({ operator, mobile = '' }) =>
+        `${operator} (${getObfuscatedAccountNumber(mobile)})`,
+    },
+    institutionName: {
+      name: 'institutionName',
+      computeValue: ({ operator }) => `${operator}`,
```
anthonydaoud, line 70: I discussed the institution name field with @cajubelt and we decided to hide it from the user as it seems it would not be different from the mobile operator.

src/fiatconnect/fiatAccountSchemas/mobileMoney.ts
```diff
@@ -0,0 +1,73 @@
...
+      isValid,
+      errorMessage: isValid ? undefined : errorMessageText,
+    }
+  }
+
+  const mobileFormatter = (input: string) => {
+    if (input.length && input[0] != '+') {
+      return `+${input}`
+    } else if (input === '+') {
+      return ''
```
satish-ravi, line 34: This would mean that if the user starts typing with a plus, it would immediately be replaced with `''` right? Is that intentional? Also since we're automatically including the `+`, should there be some text indicating they should include their area code? Otherwise they could just type their local number right?

src/fiatconnect/fiatAccountSchemas/mobileMoney.ts
```diff
@@ -0,0 +1,73 @@
...
+      infoDialog: {
+        title: i18n.t('fiatAccountSchema.mobileMoney.mobileDialog.title'),
+        actionText: i18n.t('fiatAccountSchema.mobileMoney.mobileDialog.dismiss'),
+        body: i18n.t('fiatAccountSchema.mobileMoney.mobileDialog.body'),
+        testID: 'mobileMoneyMobileDialog',
+      },
+      format: mobileFormatter,
+      validate: mobileValidator,
+      placeholderText: i18n.t('fiatAccountSchema.mobileMoney.mobile.placeholderText'),
+      keyboardType: 'number-pad',
```
satish-ravi, line 59: There's also a `phone-pad` keyboardType, although I guess if we don't allow user to type +, this may not be required

src/fiatconnect/fiatAccountSchemas/mobileMoney.ts
```diff
@@ -0,0 +1,73 @@
...
+export const getMobileMoneySchema = (
+  implicitParams: {
+    country: string
+    fiatAccountType: FiatAccountType
+  },
+  countryOverrides?: FiatAccountSchemaCountryOverrides
+): FiatAccountFormSchema<FiatAccountSchema.MobileMoney> => {
+  const overrides = countryOverrides?.[implicitParams.country]?.[FiatAccountSchema.MobileMoney]
+  const mobileValidator = (input: string) => {
+    const regex = overrides?.mobile?.regex ? new RegExp(overrides.mobile.regex) : /^[0-9+]+$/
```
satish-ravi, line 16: this needs to start with a `+` right? You can also do the length with regex. E.g.,
`/^+\d{3-15}$`.
Also, do we want to support overrides for this field? Seems like there are UI specific rules like adding a `+` automatically, which could cause issues if we push a regex without a `+` for example.

src/fiatconnect/fiatAccountSchemas/types.ts
```diff
@@ -4,6 +4,13 @@ import { KeyboardType } from 'react-native'
 export interface FormFieldParam {
   name: string
   label: string
+  infoDialog?: {
+    title: string
+    actionText: string
+    body: string
+    testID?: string
```
satish-ravi, line 11: Could we just automatically add testID using name instead of making this a form field param. E.g., we can generate testID like `${name}/dialog`

locales/base/translation.json
```diff
@@ -1438,6 +1439,22 @@
...
       "errorMessageDigit": "Account number must contain only digits"
     },
+    "mobileMoney": {
+      "operator": {
+        "label": "Network",
+        "placeholderText": "Select a network (ex. T-Mobile)"
+      },
+      "mobile": {
+        "label": "Phone Number",
+        "placeholderText": "Enter number (ex. +00 123 1234)",
```
satish-ravi, line 1449: Spaces aren't allowed in this field right? Maybe update the example? Also would be good include a real country code instead of `+00`

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -270,7 +290,32 @@ function FormField({
...
+            hitSlop={variables.iconHitslop}
+            testID={`infoIcon-${field.infoDialog.testID}`}
+          >
+            <InfoIcon size={18} color={colors.gray3} />
+          </TouchableOpacity>
+        )}
+      </View>
+      {field.infoDialog && (
+        <Dialog
+          testID={field.infoDialog.testID}
```
satish-ravi, line 308: Should the testID for the dialog and icon be different?

src/fiatExchanges/quotes/FiatConnectQuote.ts
```diff
@@ -212,8 +213,13 @@ export default class FiatConnectQuote extends NormalizedQuote {
...
   }
 
-  getFiatAccountSchemaAllowedValues(): { [key: string]: string[] } {
-    return this.quoteResponseFiatAccountSchema.allowedValues
+  getFiatAccountSchemaAllowedValues(key: string): string[] {
+    if (key === 'operator') {
+      return (
+        this.quoteResponseFiatAccountSchema.allowedValues[key] ?? Object.keys(SupportedOperatorEnum)
+      )
+    }
```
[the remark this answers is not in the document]
anthonydaoud: Good call, moved  "default allowed values" to its own file

locales/base/translation.json
```diff
@@ -1438,6 +1439,22 @@
       "errorMessageDigitLength": "Account number must be {{length}} digits",
       "errorMessageDigit": "Account number must contain only digits"
     },
+    "mobileMoney": {
+      "operator": {
+        "label": "Network",
+        "placeholderText": "Select a network (ex. T-Mobile)"
```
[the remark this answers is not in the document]
anthonydaoud: I just removed the `placeholderText` because you're right that it should always be a dropdown.

src/fiatconnect/fiatAccountSchemas/types.ts
anthonydaoud: Added with the format `dialog-${name}` to match other test ids in the file.

locales/base/translation.json
anthonydaoud: Updated so now it uses the Nigeria's country code.

src/fiatconnect/fiatAccountSchemas/mobileMoney.ts
```diff
@@ -0,0 +1,60 @@
...
+import { getObfuscatedAccountNumber } from 'src/fiatconnect'
+import { FiatAccountFormSchema } from 'src/fiatconnect/fiatAccountSchemas/types'
+import i18n from 'src/i18n'
+
+export const getMobileMoneySchema = (implicitParams: {
+  country: string
+  fiatAccountType: FiatAccountType
+}): FiatAccountFormSchema<FiatAccountSchema.MobileMoney> => {
+  const mobileValidator = (input: string) => {
+    const regex = /^\+\d{3,15}$/
```
cajubelt, line 11: Can you put a code comment next to this explaining what it means? It looks like any string starting with + followed by 3 to 15 digits, is that right? If so, seems like 3 digits is too short

cajubelt: also we should use [0-9] and not \d which includes Eastern Arabic numerals

src/fiatconnect/fiatAccountSchemas/mobileMoney.ts
```diff
@@ -0,0 +1,60 @@
...
+    const isValid = regex.test(input)
+    const errorMessageText = i18n.t('fiatAccountSchema.mobileMoney.mobile.errorMessage')
+    return {
+      isValid,
+      errorMessage: isValid ? undefined : errorMessageText,
+    }
+  }
+
+  const mobileFormatter = (input: string) => {
+    if (input.length && input[0] != '+') {
```
cajubelt, line 21: prefer strict equality

cajubelt: !==

src/fiatconnect/FiatDetailsScreen.test.tsx
```diff
@@ -326,4 +326,34 @@ describe('FiatDetailsScreen', () => {
...
+    })
+
+    const { getByTestId, getByText } = render(
+      <Provider store={store}>
+        <FiatDetailsScreen {...mmScreenProps} />
+      </Provider>
+    )
+
+    expect(getByTestId('dialog-mobile')).not.toBeVisible()
+    fireEvent.press(getByTestId(`infoIcon-mobile`))
```
satish-ravi, line 353:
```suggestion
    fireEvent.press(getByTestId('infoIcon-mobile'))
```

src/fiatconnect/FiatDetailsScreen.test.tsx
```diff
@@ -137,7 +137,7 @@ describe('FiatDetailsScreen', () => {
 
     const { getByTestId, getByText } = render(<Provider store={store}>{headerTitle}</Provider>)
 
-    expect(getByText('fiatDetailsScreen.header')).toBeTruthy()
+    expect(getByText('fiatDetailsScreen.headerBankAccount')).toBeTruthy()
```
satish-ravi, line 140: add a test checking for mobile money header. Could just use it.each on this test with two different quotes (one for BankAccount and one for MobileMoney).

src/fiatconnect/fiatAccountSchemas/mobileMoney.test.ts
```diff
@@ -0,0 +1,31 @@
...
+      // failure returns default error message
+      expect(validate(invalidPhoneNumber).errorMessage).toEqual(
+        'fiatAccountSchema.mobileMoney.mobile.errorMessage'
+      )
+    })
+    it('succeeds on valid input', () => {
+      expect(validate(validPhoneNumber).isValid).toBeTruthy()
+      expect(validate(validPhoneNumber).errorMessage).toBeUndefined()
+    })
+    it('formatter adds "+" symbol', () => {
```
satish-ravi, line 25: nit: either move this to a different describe since it doesn't test `validate` or just drop the `mobileMoney.validate` describe and rename above tests to include "validate". e.g., `validate fails/succeeds on`

src/fiatconnect/fiatAccountSchemas/mobileMoney.ts
anthonydaoud: Yes thats right, but there doesn't seem to be an explicit minimum digit length from the official international numbering [doc](https://www.itu.int/rec/T-REC-E.164-201011-I/en). I'll increase it a bit though.

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -1,3 +1,5 @@
+import { FiatAccountType } from '@fiatconnect/fiatconnect-types'
+import { TouchableOpacity } from '@gorhom/bottom-sheet'
```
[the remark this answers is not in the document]
anthonydaoud: Good catch! I didn't pay enough attention to what VS code auto-imported I guess

src/fiatconnect/FiatDetailsScreen.test.tsx
```diff
@@ -326,4 +326,34 @@ describe('FiatDetailsScreen', () => {
     )
     expect(getByTestId('checkmark')).toBeTruthy()
   })
+  it('Mobile info dialog works on mobile money screen', () => {
+    const mmQuote = new FiatConnectQuote({
+      quote: mockFcQuote,
+      fiatAccountType: FiatAccountType.BankAccount,
+      flow: CICOFlow.CashIn,
+    })
+    mmQuote.fiatAccountType = FiatAccountType.MobileMoney
```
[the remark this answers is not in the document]
anthonydaoud: I can't swap them out and comment out the test because the linter doesn't allow commented out tests. Added comments to remove/replace the code as part of ACT-518.

satish-ravi (approved):

src/fiatconnect/fiatAccountSchemas/mobileMoney.ts
```diff
@@ -0,0 +1,61 @@
...
+import { FiatAccountFormSchema } from 'src/fiatconnect/fiatAccountSchemas/types'
+import i18n from 'src/i18n'
+
+export const getMobileMoneySchema = (implicitParams: {
+  country: string
+  fiatAccountType: FiatAccountType
+}): FiatAccountFormSchema<FiatAccountSchema.MobileMoney> => {
+  const mobileValidator = (input: string) => {
+    //Phone numbers should start with a plus sign and be 6-15 digits long
+    const regex = /^\+\d{6,15}$/
```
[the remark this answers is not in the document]
cajubelt: did you miss this from my last review?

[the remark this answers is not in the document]
anthonydaoud: Oops, must have overlooked that.

PR #3122, opened 2022-11-18
fix: ensure that clearing and re-entering swap amount generates a new quote

kathaypacific:
### Description

This one has been bugging me for a while. There is some [workaround](https://github.com/valora-inc/wallet/blob/26288d3099ae0479001b423f2369f4a7282a160e/src/swap/useSwapQuote.ts#L21) to prevent duplicate quotes from being fetched, which also prevented genuine quotes from being refetched.

Before:
https://user-images.githubusercontent.com/20150449/202700637-66ea9459-7202-47fc-b6dc-465a3989870e.mp4

After:
https://user-images.githubusercontent.com/20150449/202700612-bbebae0e-a015-404b-82fd-13e42c03af22.mp4

### Other changes

N/A

### Tested

Manually

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

N/A


jeanregisser (approved):
Thanks so much,  this has been bothering me too!! Glad you got to it.

But no test to ensure it's fixed and won't come back? 😅

kathaypacific:
> Thanks so much, this has been bothering me too!! Glad you got to it.
> 
> But no test to ensure it's fixed and won't come back? 😅

fiiiiine i added a test 🙈 i also got sidetracked into fixing the last test that was flakey (i've noticed it breaking in a few unrelated PRs, with a jest error like `Ran 100000 timers, and there are still more!`) - i suspect it is because of `runAllTimers` so i changed everything that i could to `runOnlyPendingTimers`. so there's a few more changes than expected in the test file, maybe you want to take another look @jeanregisser

MuckT (approved):
The tests look great! Double checked by passing values that would fail and they failed correctly.

PR #3123, opened 2022-11-18
Use babel-plugin-module-resolver to resolve src and test paths

silasbw:
This moves away from the package.json which _seemed_ non-standard, but was making it hard to use `depcheck` because depcheck assumed those subdirectories were separate modules.


jeanregisser (approved):
🙌

kathaypacific (approved):
woah nice find 🧙🏻✨

PR #3135, opened 2022-11-19
refactor(fiatconnect): use enum for transfer status screen states

satish-ravi:
### Description

In preparation for adding a "Transaction is still processing" screen, this replaces the two booleans (isSending, failure) with a status field in fiatConnect.transfer state.

### Other changes

N/A

### Tested

Unit tests, manual

### How others should test

N/A

### Related issues

- Part of ACT-533

### Backwards compatibility

Yes


src/redux/migrations.ts
```diff
@@ -892,4 +892,11 @@ export const migrations = {
     ...state,
     app: _.omit(state.app, 'celoWithdrawalEnabledInExchange'),
   }),
+  93: (state: any) => ({
+    ...state,
+    fiatConnect: {
+      ...state.fiatConnect,
+      transfer: null,
```
satish-ravi, line 899: I am not sure actually if we need this migration, since we reset transfer to `null` on `REHYDRATE` and all the schema changes are under transfer. I can remove this if anyone can confirm this is not required.

src/fiatconnect/TransferStatusScreen.tsx
```diff
@@ -166,49 +166,57 @@ export default function FiatConnectTransferStatusScreen({ route, navigation }: P
...
+          </TextButton>
+        ),
+      })
+      return (
+        <SafeAreaView style={styles.content}>
+          <FailureSection flow={flow} normalizedQuote={normalizedQuote} fiatAccount={fiatAccount} />
+        </SafeAreaView>
+      )
+    case SendingTransferStatus.Pending:
+      // TODO
```
satish-ravi, line 215: will do this in a follow up PR.

src/redux/migrations.ts
kathaypacific: i don't think this migration is needed for the rehydrate reason you mentioned. but at the same time it doesn't hurt to exist if you want to leave it. the main purpose for migrations is to prevent breaking changes on app update - migrations which deal with removing properties are not strictly required but do become important if we were ever going to remove a property then bring it back with a different meaning. e.g. if in version 0 we had property A and in version 2 removed property A but brought it back in version 10 with a different meaning, a user who is upgrading from version 0 to 10 would have a property A which could potentially cause problems. seems unlikely that you are going to run into this edge case with this redux property though

cajubelt (approved):

PR #3139, opened 2022-11-21
fix: use CPV with e2e tests after komenci sunset

kathaypacific:
### Description

Komenci has been shut down and the onboarding pincode screen checks for it's uptime before navigating to the phone verification screen. This is not necessary with CPV, this PR adds this logic.

### Other changes

Update test id's - rather than update the tests i updated the test id of the component :/ didn't think it mattered to much

### Tested

Run CI, test manually

### How others should test

Create a new account, you should see phone verification in the onboarding flow

### Related issues

n/a

### Backwards compatibility

Y


jeanregisser (approved):
🙌

PR #3143, opened 2022-11-22
test(e2e): dismiss bottom sheet on home screen for fc tests

satish-ravi:
### Description

Dismiss bottom sheet before asserting home screen element after return user flow, which should fix occasional failures on the fc transfer out tests

### Other changes

N/A

### Tested

CI

### How others should test

N/A

### Related issues

N/A

### Backwards compatibility

N/A


kathaypacific (approved):
🚀

PR #3144, opened 2022-11-22
fix: dapplistUrl not using default remote config value in e2e tests

kathaypacific:
### Description

In the E2E tests, even though the remote config default value for `dappListApiUrl` is set it comes through as `null` in the test apps. This is because the redux store is initialising the variable using `null` instead of the remote config default.

Note that the default type is set to empty string instead of `null` because the default values are passed to `remoteConfig().setDefaults` which expects only primitive types and does not accept `null`. (e.g. https://github.com/valora-inc/wallet/actions/runs/3523615618/jobs/5907988901)  In practice this should not make a difference.

Context: https://linear.app/valora/issue/RET-416#comment-97fb3ab7

### Other changes

N/A

### Tested

Locally on my e2e tests

### How others should test

N/A

### Related issues

- Relates to RET-416

### Backwards compatibility

N/A


src/firebase/remoteConfigValuesDefaults.ts
```diff
@@ -56,6 +56,7 @@ export const REMOTE_CONFIG_VALUES_DEFAULTS: Omit<
   showPriceChangeIndicatorInBalances: false,
   paymentDeepLinkHandler: PaymentDeepLinkHandler.Disabled,
   dappsWebViewEnabled: false,
+  dappListApiUrl: '',
```
dievazqu, line 59: Just curious, why not setting `REMOTE_CONFIG_VALUES_DEFAULTS.dappListApiUrl` here?

dievazqu (approved):
LGTM

src/firebase/remoteConfigValuesDefaults.ts
kathaypacific: hum, i hope you don't mind - i didn't end up making this change. i think it's safer to leave this as an empty string so that it has the same behaviour has the default of `null` before. i think in reality it makes very little difference to the user what the initial value is anyway. it seems like we use a different value for alfajores and mainnet in remote config too, i'm not sure if it is intentional 🙈 i'm going to merge like this, lmk if you have strong opinions about this default value and i can always change it in a new PR :)

PR #3148, opened 2022-11-22
fix(style): prevent spinner from jumping when text appears

cajubelt:
updates text color instead of text content to keep position of other elements the same

https://user-images.githubusercontent.com/7979215/203415600-acb34666-1c84-42ad-a45d-abaa9cc5fdf0.mp4


src/fiatconnect/TransferStatusScreen.tsx
```diff
@@ -200,11 +201,13 @@ export default function FiatConnectTransferStatusScreen({ route, navigation }: P
     navigateHome()
   }
 
-  // add loading description if sending is taking a while
-  const [loadingDescription, setLoadingDescription] = useState('')
+  // make loading description visible if sending is taking a while
+  const [loadingDescriptionColor, setLoadingDescriptionColor] = useState(
```
cajubelt, line 205: I wanted to avoid hardcoding the "invisible" color to white, since we may support dark mode in the future

satish-ravi (approved):

PR #2324, opened 2022-04-14
chore: remove WalletConnect v2 code

jeanregisser:
### Description

This removes the last workspace packages that will allow us to move to a traditional RN app repo structure.
Part of the work to fully move off of the legacy monorepo structure.

Note: in the future once v2 is stable and we need to support it, we can either publish the `walletconnect-v1` and `walletconnect-v2` wrapper packages to NPM. Or WalletConnect will hopefully have a simpler way to support both versions 🤞 

### Tested

Updated tests.

### How others should test

Connecting Dapps with WalletConnect works as before.

### Related issues

N/A

### Backwards compatibility

Technically backward incompatible, but v2 has been disabled for a while.

PR #3161, opened 2022-11-24
chore: add walletconnect v2 to redux

kathaypacific:
### Description

As the title - I kept all the actions and redux state that were removed in #2324 but had to update the types. I tried to minimise the size of each PR so this PR only adds redux support (reducer, actions, selectors) but the actions are not yet used. I've done my best first pass at the types but they could change after I start hooking up the functionality.

### Other changes

n/a

### Tested

n/a

### How others should test

n/a

### Related issues

- Fixes RET-479

### Backwards compatibility

n/a


jeanregisser (approved):
🚀

PR #3162, opened 2022-11-24
feat: add initialise client functionality for walletconnect v2

kathaypacific:
### Description

This PR:
- add plumbing to enable initialising a walletconnect v2 connection
- there are sagas to handle initialising the v2 client, and initialise the pairing
- nothing further works yet in this PR

### Other changes

n/a

### Tested

Ensure that after scanning QR, the client initialised + pairing initialised actions are dispatched

### How others should test

n/a

### Related issues

- Fixes RET-479

### Backwards compatibility

Y


src/walletConnect/v2/saga.ts
```diff
@@ -0,0 +1,124 @@
...
+  if (!client) {
+    Logger.debug(TAG + '@createWalletConnectChannel', `init start`)
+
+    client = yield call([SignClient, 'init'], {
+      logger: 'debug',
+      // TODO get proper project ID - having issues setting up a valora
+      // walletconnect account
+      // is this a secret or env variable?
+      // https://docs.walletconnect.com/2.0/advanced/relay-server
+      projectId: '906f08083218680fedb1502f372b0b35',
```
jeanregisser, line 48: We already have an account. I'll share access with you.
And yes we need to store this like our other API keys.

src/walletConnect/v2/saga.ts
```diff
@@ -0,0 +1,124 @@
...
+      relayUrl: networkConfig.walletConnectEndpoint,
+    })
```
jeanregisser, line 50: Should we also add the [`metadata`](https://docs.walletconnect.com/2.0/javascript/guides/react-native#create-session)?

yarn.lock
```diff
@@ -21699,6 +21810,11 @@ ws@^7, ws@^7.0.0, ws@^7.4.6, ws@^7.5.0:
   resolved "https://registry.yarnpkg.com/ws/-/ws-7.5.8.tgz#ac2729881ab9e7cbaf8787fe3469a48c5c7f636a"
   integrity sha512-ri1Id1WinAX5Jqn9HejiGb8crfRio0Qgu8+MtL36rlTA6RLsMdWt1Az/19A2Qij6uSHUMphEFaTKa4WG+UNHNw==
 
+ws@^7.5.1:
+  version "7.5.9"
+  resolved "https://registry.yarnpkg.com/ws/-/ws-7.5.9.tgz#54fa7db29f4c7cec68b1ddd3a89de099942bb591"
+  integrity sha512-F+P9Jil7UiSKSkppIiD94dN07AwvFixvLIj1Og1Rl9GGMuNipJnV9JzjD6XuqmAeiswGvUmNLjr5cFuXwNS77Q==
+
```
jeanregisser, line 21817: Would be nice to deduplicated with the above since they are semver compatible, by manually removing both entries and running `yarn install` again.

jeanregisser (commented):
Great! 👍

src/walletConnect/v2/saga.ts
kathaypacific: thanks! i've updated this :D

src/walletConnect/v2/saga.ts
kathaypacific: good shout! added

yarn.lock
kathaypacific: 🫣 updated

jeanregisser (approved):

PR #3164, opened 2022-11-24
feat: support WalletConnect v2 accept and deny session

kathaypacific:
### Description

This PR allows connecting a session to WC v2.

Key changes:
- add sagas to respond to pairing request + accept and deny actions
- update the session request UI to display correct components for V2 connections

After initiating a pairing using the QR scanner / deep link:
- the dapp bottom sheet should be displayed with the correct session proposal info
- on tap "deny", the session should be denied and the pending session removed from redux
- on tap "accept", the dapp should be successfully connected +  the pending session removed from redux + the session added to redux

https://user-images.githubusercontent.com/20150449/204546355-bfe42e31-1fd7-48c2-b082-9a9c95495ccb.mov

Not included in this PR:
- analytics
- handling action requests
- handling pending sessions and actions

### Other changes

N/A

### Tested

Manually

### How others should test

N/A

### Related issues

- Fixes RET-479

### Backwards compatibility

Y


kathaypacific (commented):
the codecov patch percentage is low for this PR, i had a pass at making some unit tests but i really don't think they add much value because most of the logic is happening on the `SignClient` itself (e.g. after sending a message to the `client`, we expect the `client` internal state to have changed). i would prefer to not waste time creating heavily mocked unit tests that add no value, and test this in the e2e suite instead once the WC V2 functionality is working end to end. any objections?

jeanregisser (approved):
LGTM 👍

PR #3171, opened 2022-11-28
feat(fiatconnect): add new label to select providers screen

satish-ravi:
### Description

### Other changes

### Tested

### How others should test

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

### DIFF
--- a/src/fiatExchanges/PaymentMethodSection.tsx
+++ b/src/fiatExchanges/PaymentMethodSection.tsx
@@ -4,11 +4,13 @@ import { Image, LayoutAnimation, StyleSheet, Text, View } from 'react-native'
 import { useDispatch, useSelector } from 'react-redux'
 import { FiatExchangeEvents } from 'src/analytics/Events'
 import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
+import Dialog from 'src/components/Dialog'
 import Expandable from 'src/components/Expandable'
 import TokenDisplay from 'src/components/TokenDisplay'
 import Touchable from 'src/components/Touchable'
 import NormalizedQuote from 'src/fiatExchanges/quotes/NormalizedQuote'
 import { CICOFlow, PaymentMethod } from 'src/fiatExchanges/utils'
+import InfoIcon from 'src/icons/InfoIcon'
 import { localCurrencyExchangeRatesSelector } from 'src/localCurrency/selectors'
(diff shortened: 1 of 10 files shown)


src/fiatExchanges/PaymentMethodSection.tsx
```diff
@@ -117,14 +151,7 @@ export function PaymentMethodSection({
...
 
-      <View style={styles.imageContainer}>
-        <Image
-          testID={`image-${sectionQuotes[0].getProviderName()}`}
-          source={{ uri: sectionQuotes[0].getProviderLogo() }}
-          style={styles.providerImage}
-          resizeMode="center"
-        />
-      </View>
+      {renderProviderInfo(quote)}
```
jophish, line 154: Looks like you use `quote` here rather than `sectionQuotes[0]` which was previously being used... `quote` seems correct to use anyways since it's the argument to this function (though this function is only ever called with `sectionQuotes[0]` as an argument), but it makes me wonder why we're still using `sectionQuotes[0]` on line 147. Unrelated to this PR I guess, but should we actually be using `quote` there?

src/fiatExchanges/PaymentMethodSection.tsx
```diff
@@ -105,6 +108,37 @@ export function PaymentMethodSection({
...
+          style={styles.providerImage}
+          resizeMode="center"
+        />
+      </View>
+      {quote.isProviderNew() && (
+        <Touchable
+          testID={`newLabel-${quote.getProviderName()}`}
+          style={styles.newLabelContainer}
+          onPress={() => {
+            ValoraAnalytics.track(FiatExchangeEvents.cico_providers_new_info_opened, {
```
jophish, line 126: Should we track the fiat account schema here too, since a provider could have multiple options available?

src/fiatExchanges/PaymentMethodSection.tsx
satish-ravi: quote should always be `sectionQuote[0]` here.. removed the arg and added a comment.

src/fiatExchanges/PaymentMethodSection.tsx
satish-ravi: added paymentMethod since a quote here could be FC or non FC

jophish (approved):

PR #3180, opened 2022-11-29
feat(fiatconnect): Handle transfer out transaction errors

jophish:
### Description

For ACT-534. Modifies our transfer out flow to direct the user to the appropriate screen depending on the nature of the error we encounter while attempting a transfer on the blockchain. If there is a possibility that the transaction is in a state where it may possibly be pending on the blockchain, we direct the user to a new "pending" screen, otherwise we send them to the retry screen as normal. See flowchart [here](https://www.figma.com/file/KRK3jMiv3fjrEW9XHCznKR/FC-Transfer-Out-Failure-States?node-id=0%3A1).

### Other changes

N/A

### Tested

Unit tests, and tested by manually triggering errors to fire within the lowest-level internal send method. I have not been able to actually produce these errors yet in a _real_ transaction on the blockchain, however.

### How others should test

### Related issues

- Fixes ACT-534.

### Backwards compatibility

Yes.


src/transactions/send.ts
```diff
@@ -323,7 +323,35 @@ export function* wrapSendTransactionWithRetry(
   }
 }
 
-function shouldTxFailureRetry(err: any) {
+/**
+ * Given an error thrown while attempting a transaction,
+ * checks to see if that error could indicate that the transaction intent already
+ * exists on the blockchain.
+ **/
+export function isTxPossiblyPending(err: any): boolean {
```
jophish, line 331: These checks were pulled out, since the other cases within `shouldTxFailureRetry` do not indicate that the TX may potentially be in progress on the blockchain.

src/fiatconnect/types.ts
```diff
@@ -11,3 +11,15 @@ export interface FiatAccountSchemaCountryOverrides {
     }
   }
 }
+
+export class FiatConnectTransferOutError extends Error {
```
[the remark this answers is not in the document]
jophish: Ah, just blockchain TX's, good point.

src/fiatconnect/slice.ts
```diff
@@ -271,6 +276,17 @@ export const slice = createSlice({
...
       }
     },
+    createFiatConnectTransferTxProcessing: (
+      state,
+      action: PayloadAction<CreateFiatConnectTransferTxProcessingAction>
+    ) => {
+      state.transfer = {
+        quoteId: action.payload.quoteId,
+        flow: action.payload.flow,
+        txHash: null,
```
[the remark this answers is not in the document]
jophish: There may be a hash associated with the TX, but even if there is, I am not sure if it's actually accessible through the error object we get. I included it in the action payload because it's supported by the pending screen if we set one, but I can remove it for the time being.

jh2oman (approved):

src/fiatconnect/saga.ts
```diff
@@ -969,7 +972,16 @@ export function* _initiateSendTxToProvider({
       transferAddress: transferAddress,
       provider: fiatConnectQuote.getProviderId(),
     })
-    throw error
+    // If we've timed out, or the error is deemed unsafe to retry,
+    // it's possible that the transaction is already processing. Note that
+    // this check is not perfect; there may be false positives/negatives.
+    const txPossiblyPending =
+      error?.message === ErrorMessages.TRANSACTION_TIMEOUT || isTxPossiblyPending(error)
```
cajubelt, line 979: seems like this message check should also be done in the isTxPossiblyPending helper

src/fiatconnect/types.ts
```diff
@@ -11,3 +11,15 @@ export interface FiatAccountSchemaCountryOverrides {
...
 }
+
+export class FiatConnectTxError extends Error {
+  txPossiblyPending: boolean
+  error: any
+
+  constructor(message: string, txPossiblyPending: boolean, error: any) {
+    super(message)
+    Object.setPrototypeOf(this, FiatConnectTxError.prototype)
+    this.txPossiblyPending = txPossiblyPending
```
cajubelt, line 22: seems like you could just compute this value here, instead of taking it as a parameter. but maybe it was useful for testing to do it this way, i haven't looked at the tests closely

jophish: It's simpler for testing to do it outside of the error constructor. After moving the timeout check into the helper function though, we can just do this inline when passing args to the constructor.

src/transactions/send.ts
```diff
@@ -323,7 +323,41 @@ export function* wrapSendTransactionWithRetry(
...
+  if (!err || !err.message || typeof err.message !== 'string') {
+    return false
+  }
+
+  // Transaction has timed out; it may be on the blockchain already
+  if (err.message === ErrorMessages.TRANSACTION_TIMEOUT) {
```
satish-ravi, line 337: maybe log an error here similar to other cases where we return true?

satish-ravi (approved):

PR #3184, opened 2022-11-30
fix(fiatconnect): consistent settlement time strings

satish-ravi:
### Description

Used an enum to denote available settlement times which is then translated to the appropriate message on the SelectProvider and TransferStatus screens.

SelectProvider:

Transfer completion bank account

Transfer completion mobile money

### Other changes

N/A

### Tested

Unit tests, manual

### How others should test

Go through the withdraw flow and ensure settlement time strings are visible in the Select withdraw method screen. Complete withdraws with bank account and mobile money and ensure description contains corresponding settlement time

### Related issues

- Fixes ACT-538

### Backwards compatibility

Yes


locales/base/translation.json
```diff
@@ -1395,6 +1395,8 @@
       "header": "Success",
       "title": "Your funds are on their way!",
       "description": "Your transaction request has been sent & your funds will arrive in {{duration}}.",
+      "descriptionNumDays": "Your transaction request has been sent & your funds will arrive in 1-3 days.",
+      "description24Hours": "Your transaction request has been sent & your funds will arrive within 24 hours.",
```
satish-ravi, line 1399: Wasn't sure if we can dynamically fill in duration for within vs in (especially with other language translations), so explicitly added these strings similar to what we have for selectProvidersScreen strings. Eventually when we do dynamic strings based on quote settlement times, maybe we can make this dynamic, but didn't want to do it without the details on that, but open to other ideas.

src/fiatExchanges/quotes/FiatConnectQuote.ts
```diff
@@ -96,11 +96,6 @@ export default class FiatConnectQuote extends NormalizedQuote {
     this.quoteResponseFiatAccountSchema = quoteResponseFiatAccountSchema
   }
 
-  // TODO: Dynamically generate time estimation strings
-  private getSettlementEstimation(lower?: string, upper?: string) {
-    return strings.numDays
-  }
```
satish-ravi, line 102: Dropped this abstraction since we can just do this in `getTimeEstimation`

locales/base/translation.json
satish-ravi: I implemented another approach which uses placeholders for upper and lower bounds. So the two approaches are:
1. Using enum of available settlement time strings with hardcoded ranges
2. A more dynamic structure that supports any upper / lower bounds

1 is sufficient for now and also if we just use a predefined set of time ranges, but 2 sets us up better for when we do times based on what the quote says. Although a couple of things with 2:
- requires separate strings to handle plurals (1 hr vs 24 hrs) and different units, unless we switch to use something like [ICU message format](https://unicode-org.github.io/icu/userguide/format_parse/messages/); 
- makes an assumption that ranges always use the same units

jophish: I think one question that needs to be answered is what level of granularity is actually useful. I think I would lean towards option 1, with a sufficiently broad range of enum values that covers most of our bases. For example, we could have a couple of enum values for hour-based ranges, a couple for day-based ranges, etc. etc. We would then pick which one to use dynamically based off the quote's reported settlement time.. I think the fully dynamic approach with template strings, while ideally it conveys the maximum amount of information, would be a fair amount of work + lots of headache to implement correctly with translations, and be of questionable marginal value over option 1...

jh2oman: I also lean towards option 1 for the degree of granularity. More specific would seem odd to me as a user

jh2oman: "Your transaction request will be completed between 34 minutes and 5 days" 😖 🙃

jh2oman (approved):

jophish (approved):
:tada:

PR #3187, opened 2022-11-30
feat: handle pending sessions and actions for WalletConnect V2

kathaypacific:
### Description

This PR handles pending sessions and actions, so if a user quits the app before accepting a session / action request, the request will be displayed (if not expired) the next time the app opens.

https://user-images.githubusercontent.com/20150449/205060137-aac32b91-afd7-429d-a3f6-2c77407d0916.mov

### Other changes

Use only v3 of the `events` library, having both v1 and v2 creates issues where the `SignClient` cannot be initialised again due to `this.events.getMaxListeners is not a function. (In 'this.events.getMaxListeners()', 'this.events.getMaxListeners' is undefined)`

### Tested

Manually close the app before accepting a session or action. The app should display these again on the next session.

### How others should test

N/A

### Related issues

- Fixes RET-479

### Backwards compatibility

Y


jeanregisser (approved):
Looks great! 🚀

src/walletConnect/v2/reducer.ts
```diff
@@ -80,7 +80,20 @@ export const reducer = (
...
         ),
       }
-
+    case Actions.REMOVE_EXPIRED_SESSIONS_V2: {
+      const unexpiredSessions = state.sessions.filter(
+        (session) => session.expiry > action.dateInSeconds
+      )
+      return {
+        ...state,
+        sessions: unexpiredSessions,
```
jeanregisser, line 89: Shouldn't we also check `pendingSessions`?

kathaypacific: good catch! i didn't notice the expiry param before, i added now!

PR #3188, opened 2022-11-30
feat: handle close session with WalletConnect v2

kathaypacific:
### Description

### Other changes

### Tested

### How others should test

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

### DIFF
--- a/src/walletConnect/screens/Sessions.tsx
+++ b/src/walletConnect/screens/Sessions.tsx
@@ -1,3 +1,4 @@
+import { SessionTypes, SignClientTypes } from '@walletconnect/types'
 import React, { useState } from 'react'
 import { useTranslation } from 'react-i18next'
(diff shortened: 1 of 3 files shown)


jeanregisser (approved):
Looks great 👍 

Small question, in the video there's no title for the dapp in the settings, is this going to be fixed in a different PR?

kathaypacific:
> Looks great 👍
> 
> Small question, in the video there's no title for the dapp in the settings, is this going to be fixed in a different PR?

haha nice spot, actually the dapp name is supposed to come from the dapp but this wallet connect test dapp has `metadata: {
 description: ""
 icons: []
 name: ""
 url: https://react-app.walletconnect.com/
}` - perhaps we should set some kind of default dapp name if it's missing from dapps? i thought it was bad practice to use empty strings for required params though and was hoping that not many dapps would do this 🙈

PR #3192, opened 2022-12-01
test(e2e): add homefeed tests for iOS

MuckT:
### Description

Tests were only added on iOS as the current Detox implementation of `.getAttributes()` does not work when multiple items are matched - https://github.com/wix/Detox/issues/3196

Adds homefeed e2e tests:
- should load more items on scroll
- should show correct information on tap of feed item'

### Other changes

- Snapshots updated and testID's added

### Tested

Tested locally on iOS.

### How others should test

N/A

### Related issues

RET-417

# Backwards compatibility

Yes


src/transactions/feed/TransactionFeed.tsx
```diff
@@ -103,11 +103,23 @@ function TransactionFeed() {
...
+          <View testID="TransactionListItem">
+            <TransferFeedItem key={tx.transactionHash} transfer={tx} />
+          </View>
+        )
       case 'NftTransferV2':
-        return <NftFeedItem key={tx.transactionHash} transaction={tx} />
+        return (
+          <View testID="TransactionListItem">
+            <NftFeedItem key={tx.transactionHash} transaction={tx} />
+          </View>
```
jeanregisser, line 121: I'd prefer we pass a `testID` and modify these components instead of wrapping them with an unnecessary view.

It's a micro optimization, but if it's not strictly needed, let's avoid the extra view.

e2e/src/HomeFeed.spec.js
```diff
@@ -0,0 +1,48 @@
...
+import { quickOnboarding, sleep, waitForElementId } from './utils/utils'
+
+const jestExpect = require('expect')
+
+beforeAll(async () => {
+  await quickOnboarding()
+})
+
+beforeEach(async () => {
+  await reloadReactNative(false)
```
jeanregisser, line 11: Sorry, I know I've already asked, but can we do without reloading React Native?

e2e/src/HomeFeed.spec.js
```diff
@@ -0,0 +1,48 @@
...
+})
+
+// iOS only as getAttributes on multiple elements is not supported on Android
+describe('Home Feed :ios:', () => {
```
jeanregisser, line 15: How is this skipped on Android?

jeanregisser (commented):
Awesome! 👍

e2e/src/HomeFeed.spec.js
MuckT: If `:ios:` or `:android:` is in the describe block name or test name it will only attempt the suite / test on that OS.

e2e/src/HomeFeed.spec.js
MuckT: I reloaded in this case so that the first test that scrolls the transaction feed won't impact the state in the second test that is checking an item in the feed. I'll swap the test order and try without reloading the feed.

src/transactions/feed/TransactionFeed.tsx
MuckT: Removed in https://github.com/valora-inc/wallet/pull/3192/commits/d893332e71e3946ec1209ad84fbec663686d6373!

e2e/src/HomeFeed.spec.js
MuckT: Removed in https://github.com/valora-inc/wallet/pull/3192/commits/d893332e71e3946ec1209ad84fbec663686d6373!

e2e/src/HomeFeed.spec.js
```diff
@@ -0,0 +1,40 @@
...
+
+const jestExpect = require('expect')
+
+beforeAll(async () => {
+  await quickOnboarding()
+})
+
+// iOS only as getAttributes on multiple elements is not supported on Android
+describe('Home Feed :ios:', () => {
+  it.only('should show correct information on tap of feed item', async () => {
```
jeanregisser, line 12:
```suggestion
  it('should show correct information on tap of feed item', async () => {
```

jeanregisser (approved):
Approved! But we need to remove the `.only`. 😄

PR #3201, opened 2022-12-01
chore: switch to react-native-sms-retriever from our fork

silasbw:
There are two sets of changes: parallel SMS and using androidx packages

  https://github.com/Bruno-Furtado/react-native-sms-retriever/compare/master...valora-inc:react-native-sms-retriever:master

The parallel SMS isn't relevant with CPV, and upstream includes the androidx ones.


jeanregisser (approved):
Great! I've fixed the tests and successfully ran it locally, the SMS was still auto populated 💯 👍

PR #3207, opened 2022-12-02
chore: remove react-native-udp from dependencies

silasbw:
It appears unused is even though it was flagged by depcheck


jeanregisser (commented):
Ah yes this one is a [`peerDependency` of `react-native-ntp-client`](https://github.com/artem-russkikh/react-native-ntp-client/blob/65dd6ccb8acbd2434ef78f31ee20eefafe5f8f0e/package.json#L51), removing it will break it.
I've also thought it wasn't needed in the past.

But we could probably remove `react-native-ntp-client` now since I don't think there's any module left needing clock sync. Komenci and CIP8 features needed an up-to-date client clock for their auth mechanism.

silasbw:
ah, thanks!

silasbw:
@jeanregisser is working on this "the right way"!

jeanregisser:
@silasbw we still need this one, I haven't remove react-native-udp from #3229 :D

silasbw:
:facepalm:

jeanregisser (approved):

PR #3208, opened 2022-12-02
fix: misc walletconnect v2 issues

kathaypacific:
### Description

This PR fixes 3 issues:
1. no event is emitted on the SignClient when a new session is created, so there was a race condition between acknowledging the session proposal and storing the new session (the new session data is on the client itself) - this resulted in sometimes the session being established correctly but the app did not have the new session data to store in redux. fixed this by adding a retry to fetch the new session from the client with a timeout.
2. pass the raw tx to the `eth_signTransaction` method 
3. increase the WC connection timeout (maybe it is the test app that is slow, but i saw frequently that the session proposal event can take more than 10 seconds to be emitted)

### Other changes

N/A

### Tested

Manually

### How others should test

N/A

### Related issues

- Fixes RET-479

### Backwards compatibility

Y


src/walletConnect/v2/saga.ts
```diff
@@ -51,6 +51,8 @@ let client: SignClient | null = null
 
 const TAG = 'WalletConnect/saga'
 
+const CONNECTION_TIMEOUT = 10_000
```
jeanregisser, line 54: Can we use a different name so there's no confusion with the the other `CONNECTION_TIMEOUT`?

Maybe?
```suggestion
const GET_SESSION_TIMEOUT = 10_000
```

jeanregisser (approved):
👍

src/walletConnect/v2/saga.ts
kathaypacific: yikes i thought i pushed a change to update this but it didn't go through, i'll address this in the next PR

PR #3209, opened 2022-12-02
chore: add analytics for walletconnect v2 interactions

kathaypacific:
### Description

This PR adds all the WC events that is available for WC v1, and was taken pretty much directly from https://github.com/valora-inc/wallet/pull/2324

### Other changes

N/A

### Tested

Manually

### How others should test

N/A

### Related issues

- Fixes RET-479

### Backwards compatibility

Y


src/walletConnect/v2/saga.ts
```diff
@@ -52,6 +58,62 @@ let client: SignClient | null = null
...
+  const permissionsJsonrpcMethods: string[] = []
+  const permissionsNotificationsTypes: string[] = []
+
+  const requiredNamespaces = isSessionProposalType(session)
+    ? session.params.requiredNamespaces
+    : session.requiredNamespaces
+  Object.keys(requiredNamespaces).forEach((key) => {
+    permissionsBlockchains.push(...requiredNamespaces[key].chains)
+    permissionsJsonrpcMethods.push(...requiredNamespaces[key].methods)
+    permissionsNotificationsTypes.push(...requiredNamespaces[key].events)
```
kathaypacific, line 84: @jeanregisser some of the wc variable names have changed so i wasn't sure if the `permissionsXXX` variables are mapped correctly here. a session proposal looks something like this, and you can see the `requiredNamespaces` there:
```
{
  "type": "WALLETCONNECT/SESSION_PROPOSAL_V2",
  "payload": {
    "session": {
      "id": 1669989187506938,
      "params": {
        "expiry": 1669989496,
        "proposer": {
          "publicKey": "f4284dc764da82e9b62d625f4dfea4088142f477c0d7420cdec2a0f49959c233",
          "metadata": {
            "url": "https://react-app.walletconnect.com",
            "icons": [],
            "description": "",
            "name": ""
          }
        },
        "requiredNamespaces": {
          "eip155": {
            "events": [
              "chainChanged",
              "accountsChanged"
            ],
            "chains": [
              "eip155:44787"
            ],
            "methods": [
              "eth_sendTransaction",
              "eth_signTransaction",
              "eth_sign",
              "personal_sign",
              "eth_signTypedData"
            ]
          }
        },
        "id": 1669989187506938,
        "relays": [
          {
            "protocol": "irn"
          }
        ],
        "pairingTopic": "ab7c79764b6838abd24669ab735f6ce40bb26ca4d54cf948daca8e80a2eb6db1"
      }
    }
  }
}
```

jeanregisser (approved):
👍

src/walletConnect/v2/saga.ts
kathaypacific: are you worried about the type of requiredNamespaces causing issues? i don't see any existing event properties as an object but i can always do a json stringify of the value, or something clever like https://github.com/valora-inc/wallet/blob/2d425973dee8bbf5c87854c840999f1343ba7985/src/analytics/selectors.ts#L82-L89? having a single key (or derived keys) is more accurate, so i'd prefer that if we're not fussed about keeping the events the same as before :)

kathaypacific:
@jeanregisser ready for re-review after our discussion - with the dynamic properties we lose static typing but i added a test so make up for it

src/analytics/Properties.tsx
```diff
@@ -1229,9 +1229,6 @@ export interface WalletConnect2Properties {
   dappUrl: string
   dappDescription: string
   dappIcon: string
-  permissionsBlockchains: string[]
-  permissionsJsonrpcMethods: string[]
-  permissionsNotificationsTypes: string[]
```
jeanregisser, line 1234: Could we mention as a comment that we should have `eip155Events`,  `eip155Chains` and `eip155Methods` added dynamically?

jeanregisser (approved):
🚀

src/analytics/Properties.tsx
kathaypacific: yup, added a comment!

PR #3212, opened 2022-12-05
feat(fiatconnect): add review screen fee disclaimer

anthonydaoud:
### Description

Currently, users can receive a FiatConnect quote in a currency that is different from the one used by the user's bank. When this occurs, the user may be charged a fee by their bank, but we do not yet inform the user of this possibility. This PR adds a disclaimer on the Review Screen about these possible additional fees. 

FeeDisclaimer

### Other changes

N/A

### Tested

Unit tests.

### How others should test

When Cashing In/Out via FiatConnect, a disclaimer should be shown if the in-app currency does the match the currency used in the user's geographical location.

### Related issues

- ACT-530

### Backwards compatibility

N/A


src/fiatconnect/ReviewScreen.tsx
```diff
@@ -36,15 +42,35 @@ import { Currency } from 'src/utils/currencies'
...
+  )
+}
+
 export default function FiatConnectReviewScreen({ route, navigation }: Props) {
   const { t } = useTranslation()
   const dispatch = useDispatch()
   const { flow, normalizedQuote, fiatAccount, shouldRefetchQuote } = route.params
   const fiatConnectQuotesLoading = useSelector(fiatConnectQuotesLoadingSelector)
   const fiatConnectQuotesError = useSelector(fiatConnectQuotesErrorSelector)
+  const localCurrency = useSelector(getDefaultLocalCurrencyCode)
```
jh2oman, line 66: nit: please call this something like `defaultLocalCurrency` b/c localCurrency is the variable name we use throughout the wallet for the currency that the user selects

src/fiatconnect/constants.ts
```diff
@@ -0,0 +1,25 @@
+import { FiatAccountSchema } from '@fiatconnect/fiatconnect-types'
+import { LocalCurrencyCode } from 'src/localCurrency/consts'
+
+export const CURRENCIES_WITH_FEE_DISCLAIMER: Partial<
```
jh2oman, line 4: the AC for the ticket says this list should be the list of *countries* not currencies. Any reason why you did currency instead?

src/fiatconnect/constants.ts
```diff
@@ -0,0 +1,25 @@
...
+  Record<FiatAccountSchema, Set<LocalCurrencyCode>>
+> = {
+  [FiatAccountSchema.AccountNumber]: new Set([
```
jh2oman, line 7: Should this list be in firebase? Otherwise we will have to wait a whole release cycle for the disclaimer to be in a new country thats supported.

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -36,15 +42,35 @@ import { Currency } from 'src/utils/currencies'
 
 type Props = NativeStackScreenProps<StackParamList, Screens.FiatConnectReview>
 
+const disclaimerIsNeeded = (
+  flow: CICOFlow,
+  quote: FiatConnectQuote,
+  localCurrency: LocalCurrencyCode
+) => {
+  const fiatType = quote.getFiatType()
+  const currencyEligibleForDisclaimer =
```
cajubelt, line 51: I don't get this `currencyEligibleForDisclaimer` thing. Is there some reason we wouldn't want to show this disclaimer when the local currency differs from the fiat currency on the quote?

I suspect that this was an effort to comply with the last AC of the issue in Linear https://linear.app/valora/issue/ACT-530/disclaimer-for-users-with-different-app-currency-than-cash-out , but I wonder if that AC was taken the wrong way. Instead of geo-fencing the feature, maybe that AC was supposed to give an MVP set of locales this should work in, which we can work to expand later on if more eng effort is needed. 

Also, why only for cash out? If you try to send USD from an NGN bank account (for instance), and it deducts more NGN than you expect, it still seems warning-worthy.

cajubelt: I realize that cash out is in the AC also but idk if it makes sense either

cajubelt (commented):
mostly confused by the AC in the Linear issue

locales/base/translation.json
```diff
@@ -1348,6 +1348,7 @@
     "transactionDetails": "Transaction Details",
     "paymentMethodVia": "Connected via {{providerName}}",
     "receiveAmount": "You will receive",
+    "feeDisclaimer": "Your bank may charge additional fees for your transfer.",
```
jh2oman, line 1351: This string is bank specific. Shouldn't it support mobile money as well?

src/fiatconnect/ReviewScreen.tsx
anthonydaoud: Removed that variable in a subsequent commit

src/fiatconnect/constants.ts
anthonydaoud: I initially interpreted the 2nd AC to mean we just use the currency codes of those countries, but I've updated the code to rely on countries.

locales/base/translation.json
anthonydaoud: Good catch!

src/fiatconnect/ReviewScreen.tsx
anthonydaoud: Good points! After discussion on slack, I've made it so the PR verifies that the disclaimer can show up in the MVP locales but will also display whenever the local currency differs from the quote fiat currency.

src/fiatconnect/ReviewScreen.test.tsx
```diff
@@ -14,6 +14,32 @@ import { Currency } from 'src/utils/currencies'
 import { createMockStore, getMockStackScreenProps, sleep } from 'test/utils'
 import { mockFiatConnectQuotes } from 'test/values'
 
+const MVP_SUPPORTED_COUNTRIES: Partial<Record<FiatAccountSchema, Array<string>>> = {
```
jophish, line 17: Where does this list come from? Is this just "countries that have some amount of support for cash out, but do not have USD as one of their currency types"? I feel like this might get out of date/unmaintained as we add support for more countries.

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -24,27 +29,47 @@ import { createFiatConnectTransfer, refetchQuote } from 'src/fiatconnect/slice'
...
   const fiatConnectQuotesLoading = useSelector(fiatConnectQuotesLoadingSelector)
   const fiatConnectQuotesError = useSelector(fiatConnectQuotesErrorSelector)
+  const { countryCodeAlpha2 } = useSelector(userLocationDataSelector)
   const [showingExpiredQuoteDialog, setShowingExpiredQuoteDialog] = useState(
     normalizedQuote.getGuaranteedUntil() < new Date()
   )
+  const showFeeDisclaimer = useMemo(
+    () => disclaimerIsNeeded(normalizedQuote.getFiatType(), countryCodeAlpha2),
+    [normalizedQuote, countryCodeAlpha2]
+  )
```
jh2oman, line 72: I think this logic can be pretty drastically simplified by just using the `getDefaultLocalCurrencyCode` selector which is what the app uses to select the users default currency when the create a new account.

I believe this is what charlie meant when he said
> In pseudocode, this is what I'd expect:
if (selectedCurrency !== localCurrency) {
  show disclaimer
}

Something like this
```suggestion
  const defaultLocaleCurrencyCode =  useSelector(getDefaultLocalCurrencyCode)
  const showFeeDisclaimer = useMemo(
    () => normalizedQuote.getFiatType() !== convertToFiatConnectFiatCurrency(defaultLocaleCurrencyCode),
    [normalizedQuote, defaultLocaleCurrencyCode]
  )
```

src/fiatconnect/ReviewScreen.test.tsx
anthonydaoud: This list comes directly from the 3rd AC of ACT-530. After a [discussion](https://valora-app.slack.com/archives/C02KBT0DAHJ/p1670441199274289) on slack, I interpret that AC to mean we should ensure the MVP countries receive the disclaimer if the currency is unsupported locally (which USD is). I do agree that it may get out of date with time, but it appears we are trying to verify that at least these listed countries are supported initially.

src/fiatconnect/ReviewScreen.tsx
anthonydaoud: Yeah I originally had implemented it using `getDefaultLocalCurrencyCode` but some countries support multiple currencies, which means that the disclaimer may be shown unnecessarily in those cases.

anthonydaoud: Ahh gotcha, I've updated the code again.

src/localCurrency/consts.ts
```diff
@@ -10,6 +10,7 @@ export enum LocalCurrencyCode {
   EUR = 'EUR',
   GBP = 'GBP',
   GHS = 'GHS',
+  GMD = 'GMD',
```
jh2oman, line 13: These new currencies are not part of the fiatconnect spec, so we probably should not add them in https://github.com/fiatconnect/specification/blob/main/fiatconnect-api.md#921-fiattypeenum

src/fiatconnect/ReviewScreen.test.tsx
```diff
@@ -272,5 +339,38 @@ describe('ReviewScreen', () => {
...
+        FiatAccountType.BankAccount,
+        FiatAccountSchema.AccountNumber,
+        'fiatConnectReviewScreen.bankFeeDisclaimer',
+      ],
+      [
+        FiatAccountType.MobileMoney,
+        FiatAccountSchema.MobileMoney,
+        'fiatConnectReviewScreen.mobileMoneyFeeDisclaimer',
+      ],
+    ])('Fee Disclaimer for %s', (accountType, schema, disclaimer) => {
```
jh2oman, line 353: This test doesn't seem like it adds much value. Its essentially testing that for every country in the list the default currency is not USD. Thats testing the `country-data` package more than our actual code which is probably unnecessary. Especially since it adds ~20 more tests

src/fiatconnect/ReviewScreen.test.tsx
satish-ravi: nit: Is there a reason to test a specific list of currencies? Looking at the code, seems like we just need to test 2 cases for each schema, one with fiat type === default currency and one with !==

src/fiatconnect/ReviewScreen.test.tsx
jh2oman: I think its safe to remove along with `MVP_SUPPORTED_COUNTRIES`

src/fiatconnect/ReviewScreen.test.tsx
```diff
@@ -8,12 +9,59 @@ import FiatConnectReviewScreen from 'src/fiatconnect/ReviewScreen'
...
+  const originalModule = jest.requireActual('src/localCurrency/selectors')
+
+  return {
+    ...originalModule,
+    getDefaultLocalCurrencyCode: jest.fn(),
+  }
+})
+const mockGetDefaultLocation = getDefaultLocalCurrencyCode as jest.MockedFunction<
+  typeof getDefaultLocalCurrencyCode
+>
```
satish-ravi, line 62: nit: you can use the `mocked` util instead of casting type with `as`

src/fiatconnect/ReviewScreen.test.tsx
```diff
@@ -56,6 +104,9 @@ describe('ReviewScreen', () => {
   beforeEach(() => {
     store.dispatch = jest.fn()
   })
+  afterAll(() => {
+    jest.resetAllMocks()
```
satish-ravi, line 108: what's the purpose of this reset in afterAll?

src/fiatconnect/ReviewScreen.test.tsx
```diff
@@ -78,6 +130,20 @@ describe('ReviewScreen', () => {
         'fiatConnectReviewScreen.paymentMethodVia, {"providerName":"Provider Two"}',
       ])
     })
+    it('shows fee disclaimer when quote currency does not match user locale currency', () => {
+      const mockProps = getProps(CICOFlow.CashIn)
+      //Re-renders, need to mock twice
```
satish-ravi, line 135: nit: might be cleaner if you just set a default mock in before each, and drop the once when mocking here so the dependency on how many times the component renders is removed.

src/fiatconnect/ReviewScreen.test.tsx
anthonydaoud: Ok, I'll remove the list and just do those 2 tests per schema.

src/fiatconnect/ReviewScreen.test.tsx
anthonydaoud: Oops, removed

src/fiatconnect/ReviewScreen.test.tsx
anthonydaoud: Good idea!

PR #3214, opened 2022-12-06
test: add e2e tests for wallet connect v2

kathaypacific:
### Description

### Other changes

### Tested

### How others should test

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

### DIFF
--- a/e2e/src/usecases/WalletConnectV2.js
+++ b/e2e/src/usecases/WalletConnectV2.js
@@ -1,50 +1,147 @@
-import WalletConnectClient, { CLIENT_EVENTS } from '@walletconnect/client'
+import { newKit } from '@celo/contractkit'
+import Client from '@walletconnect/sign-client'
+import fetch from 'node-fetch'
 import { formatUri } from '../utils/encoding'
-import { scrollIntoView, sleep, waitForElementId } from '../utils/utils'
+import { enterPinUiIfNecessary, scrollIntoView, waitForElementId } from '../utils/utils'
 
-const WalletConnectAPIKey = process.env.WALLET_CONNECT_API_KEY || ''
-let uri, walletConnector
+const jestExpect = require('expect')
+
+const dappName = 'WalletConnectV2 E2E'
+const DEFAULT_PROJECT_ID =
+  process.env.WALLET_CONNECT_PROJECT_ID || 'b8620cf33f99d407708c298d68224d8a'
+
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -260,6 +260,7 @@
     "jest-html-reporter": "^3.5.0",
     "jest-junit": "^10.0.0",
     "jest-snapshot": "^26.1.0",
+    "lokijs": "^1.5.12",
```
kathaypacific, line 263: this is needed to run a wallet connect client on the server side. i put it in dev dependencies since we're just using it for the tests

e2e/src/usecases/WalletConnectV2.js
```diff
@@ -1,68 +1,164 @@
...
+const kitUrl = process.env.FORNO_URL || 'https://alfajores-forno.celo-testnet.org'
+const kit = newKit(kitUrl)
+
+const walletAddress = (
+  process.env.E2E_WALLET_ADDRESS || '0x6131a6d616a4be3737b38988847270a64bc10caa'
+).toLowerCase()
+
+async function formatTestTransaction(address) {
+  try {
+    const response = await fetch(kitUrl, {
```
kathaypacific, line 22: i'm not sure if it's really necessary to get the real nonce here, but the example dapp that walletconnect provides does this. i think it will make the test less flakey if the nonce is correct (we're already using forno to check the tx receipt so it doesn't add an extra source of flake here)

jeanregisser (approved):
w00t w00t!! 🎉 
So happy to see this! Awesome work 🥇

.github/workflows/e2e-android.yml
```diff
@@ -24,6 +24,11 @@ jobs:
         run: git diff --exit-code
       - name: Check E2E wallet balance
         run: NODE_OPTIONS='--unhandled-rejections=strict' yarn ts-node ./e2e/scripts/check-e2e-wallet-balance.ts
+      - name: Create Android E2E .env File
+        env:
+          WALLET_CONNECT_PROJECT_ID_E2E: ${{ secrets.WALLET_CONNECT_PROJECT_ID_E2E }}
+        working-directory: e2e
+        run: echo WALLET_CONNECT_PROJECT_ID_E2E=$WALLET_CONNECT_PROJECT_ID_E2E >> .env
```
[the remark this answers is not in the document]
kathaypacific: 👍 updated this in the iOS file as well

PR #3217, opened 2022-12-06
refactor: avoid "exposing" the cloud functions base URL

silasbw:
We expose it specifically for Sentry, but another piece of code was using it to construct a REST API endpoint (instead of following the convention of defining API endpoints explicitly in networkConfig.ts). This change is pretty mundane, but the intent it to make it harder to avoid the convention.

### DIFF
--- a/src/consumerIncentives/saga.test.ts
+++ b/src/consumerIncentives/saga.test.ts
@@ -89,7 +89,7 @@ describe('fetchAvailableRewardsSaga', () => {
   }
   const error = new Error('Unexpected error')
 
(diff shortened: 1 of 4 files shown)


jh2oman (approved):

kathaypacific (approved):
🚀 makes sense!

jeanregisser (commented):
Ah thanks, but I actually wanted to do the opposite and stop writing this boilerplate for each new function we use 😅

silasbw:
That makes sense @jeanregisser, but at least this converges on a single convention.  I don't think I feel strongly if we move to another convention (just as long as we follow something).

PR #3227, opened 2022-12-08
chore: bump react-native-config (and use npm package)

silasbw:


silasbw:
:facepalm: ts build fails even though this is a minor version bump

silasbw:
Easy to fix!

src/config.ts
```diff
@@ -52,7 +52,7 @@ export const TOKEN_MIN_AMOUNT = 0.00000001
 export const DOLLAR_MIN_AMOUNT_ACCOUNT_FUNDED = 0.01
 // The number of seconds before the sender can reclaim the payment.
 export const ESCROW_PAYMENT_EXPIRY_SECONDS = 1 // The contract doesn't allow 0 seconds.
-export const DEFAULT_TESTNET = Config.DEFAULT_TESTNET
+export const DEFAULT_TESTNET = Config.DEFAULT_TESTNET!
```
satish-ravi, line 55: Is it safe to use `!` here and below? Wonder if we should check for existence and throw if not found or at least have some defaults?

jeanregisser (approved):
Thanks! 🚀 

`Podfile.lock` needs to be updated and committed, running `cd ios; bundle exec pod install`

src/config.ts
jeanregisser: Ah yes it could be a bit better 👍

PR #3229, opened 2022-12-09
chore: remove react-native-ntp-client

jeanregisser:
### Description

Related to the work to remove `react-native-udp`: https://github.com/valora-inc/wallet/pull/3207

The app used `react-native-ntp-client` in a few places to get the time from a NTP server. It also enforced the device clock was not too far from it, and showed a blocking screen at app launch if that was not the case. This was bad UX and I’ve always wanted to remove the need for this.

Things that needed a clock in sync: Komenci, Stokado (CIP8) because they were signing a message with a timestamp for authenticating users. But we don’t use these anymore.

Bye bye clock sync screen:

 

### Tested

Tests pass.

### How others should test

The app should continue to work as usual.

### Related issues

N/A

### Backwards compatibility

Yes


silasbw (approved):
:partying_face:

PR #3230, opened 2022-12-09
fix(fiatconnect): CELO fiatconnect withdrawals

cajubelt:
### Description

We were converting CELO to cGLD, and then comparing to a list of strings that used CELO, not cGLD

### Other changes

slightly updated test mocks

### Tested

- automated regression test for _initiateSendTxToProvider
- manually tested cash out with CELO with mock-provider on Alfajores

### How others should test

na

### Related issues

fixes https://linear.app/valora/issue/BIT-7/selecting-celo-results-in-failed-cash-out

### Backwards compatibility

na


src/fiatExchanges/quotes/FiatConnectQuote.ts
```diff
@@ -203,6 +203,11 @@ export default class FiatConnectQuote extends NormalizedQuote {
     return this.quote.quote.cryptoAmount
   }
 
+  getCryptoTypeString(): string {
+    // todo eventually rename this to getCryptoType and rename getCryptoType to getCryptoCurrency
```
cajubelt, line 207: I didn't do the rename in this PR because getCryptoType is actually used in many places, so it added a lot of clutter to the review

src/fiatconnect/saga.test.ts
```diff
@@ -1796,6 +1835,44 @@ describe('Fiatconnect saga', () => {
         .returns('0x12345')
         .run()
     })
+    it.each(Object.keys(CryptoType) as CryptoType[])(
```
cajubelt, line 1838: I verified this test fails for CELO with the old implementation. It also happens to fail for cREAL

test/values.ts
```diff
@@ -458,7 +460,9 @@ export const makeExchangeRates = (
   },
 })
 
-export const mockTokenBalances = {
+export const mockTokenBalances: Record<string, StoredTokenBalance> = {
+  // NOTE: important to keep 'symbol' fields in this object matching their counterparts from here: https://github.com/valora-inc/address-metadata/blob/main/src/data/mainnet/tokens-info.json ,
+  //  particularly for CICO currencies
   '0x00400FcbF0816bebB94654259de7273f4A05c762': {
```
satish-ravi, line 466: Should all addresses in this list be lowercase?

satish-ravi (approved):

src/fiatconnect/saga.test.ts
satish-ravi: nit: combine this test with the test above? I am guessing the above tests the same thing with cUSD

test/values.ts
```diff
@@ -458,7 +460,9 @@ export const makeExchangeRates = (
   },
 })
 
-export const mockTokenBalances = {
+export const mockTokenBalances: Record<string, StoredTokenBalance> = {
```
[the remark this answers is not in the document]
cajubelt: thanks for the catch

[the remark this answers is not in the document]
cajubelt: I think it was because I was adding a nonzero balance for one of the tokens

test/values.ts
cajubelt: fixed

test/values.ts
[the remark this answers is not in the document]
cajubelt: oof turns out tests in tokens/saga.test.ts were very brittle and relied specifically on some values in mockTokenBalances to stay exactly as they were. I updated those tests to be be more isolated but could use a review to make sure my changes were valid

src/fiatconnect/saga.test.ts
cajubelt: yeah I largely copied the above test.

cajubelt: I'll delete the other one and update the name of this one to make it more clear what it's testing

src/tokens/saga.test.ts
```diff
@@ -12,99 +12,94 @@ import {
...
-import { fetchTokenBalancesFailure, setTokenBalances, StoredTokenBalances } from 'src/tokens/slice'
+import {
+  fetchTokenBalancesFailure,
+  setTokenBalances,
+  StoredTokenBalance,
+  StoredTokenBalances,
+} from 'src/tokens/slice'
 import { walletAddressSelector } from 'src/web3/selectors'
 import { createMockStore } from 'test/utils'
-import { mockAccount, mockCeurAddress, mockCusdAddress, mockTokenBalances } from 'test/values'
```
cajubelt, line 18: when I added fixture data to `mockTokenBalances` it broke these tests-- even when I added balances of 0 for some tokens. There was an assumption that fixture data in this file would stay in sync with `mockTokenBalances`, otherwise a test would fail. I updated the code here to remove that assumption so we can add shared fixture data without breaking dependencies.

PR #3237, opened 2022-12-13
feat(cico): setup add funds crypto exchange -> qr code experiment

satish-ravi:
### Description

Experiment: https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/experiments/add_funds_crypto_exchange_qr_code/setup

### Other changes

Renamed existing statsig ExperimentParams to LayerParams

### Tested

Manually by uncommenting the line to get experiments and checking logs for exp params

### How others should test

N/A

### Related issues

- Fixes ACT-545

### Backwards compatibility

Yes


src/fiatExchanges/SelectProvider.tsx
```diff
@@ -324,6 +359,18 @@ function ExchangesSection({
 }) {
   const { t } = useTranslation()
 
+  // TODO(any): uncomment below and use params
+  // @ts-ignore
+  const { exchangesText, exchangesFlow } = useMemo(() => {
+    // if (flow === CICOFlow.CashIn) {
+    //   return getAddFundsCryptoExchangeExperimentParams()
+    // }
```
satish-ravi, line 367: Leaving this commented because if a version is cut before this is actually used, this will impact the experiment (unless we gate it by versions)

src/fiatExchanges/SelectProvider.tsx
```diff
@@ -324,6 +358,18 @@ function ExchangesSection({
 }) {
   const { t } = useTranslation()
 
+  // TODO(any): uncomment below and use params
+  // @ts-ignore
+  const { exchangesText, exchangesFlow } = useMemo(() => {
```
jh2oman, line 363:
```suggestion
  const { exchangesText, exchangesLink } = useMemo(() => {
```

src/fiatExchanges/SelectProvider.tsx
```diff
@@ -57,6 +61,36 @@ const TAG = 'SelectProviderScreen'
 
 type Props = NativeStackScreenProps<StackParamList, Screens.SelectProvider>
 
+export function getAddFundsCryptoExchangeExperimentParams() {
+  const { addFundsExchangesLink, addFundsExchangesText } =
```
jh2oman, line 65: nit: IMO the destructuring here makes the code more confusing because its not clear what `addFundsExchangesLink` and `addFundsExchangesLink` are. It might be clearer, if a little more verbose, to just call this variable `params` and reference that variable through the rest of the function 
```js
const exchangesText = experiment.get(
      params.addFundsExchangesText.paramName,
      params.addFundsExchangesText.defaultValue
    )
```

don't feel too strongly.

satish-ravi: updated, I was also not too convinced with the destructuring

PR #3242, opened 2022-12-14
feat: allow screenshots of the backup phrase on Android

silasbw:
This was already allowed on iOS and we're unsure if this feature on Android encourages good or bad security practices. We're removing to parity with iOS.


jeanregisser (approved):
Awesome! 💯 

Already approved but we should also delete https://github.com/valora-inc/wallet/blob/f7bd8ddf479540589b46e5366837ea991128d05c/__mocks__/react-native-flag-secure-android.ts

PR #3245, opened 2022-12-14
feat(qr-code): Allow QR code to display the user's address

jh2oman:
### Description

Allow's the user's address to be scanned via QR code. Other wallets such as Coinbase & Metamask can scan the QR code to send funds to the user's address.

Will be used later in the [CICO Exchange QR experiment](https://www.figma.com/file/QMG12y2ToSm6MQ1Svdq6BI/CICO-via-Exchange?node-id=1%3A3&t=Id3pOXwDldiblGYx-0)

### Tested

Tested with:
* Coinbase 
* Metamask

### How others should test

Run locally, change the existing QR code to use Address as the data type, attempt to scan using the Coinbase/Metamask app

### Related issues

https://linear.app/valora/issue/ACT-543/update-qr-code-within-qr-code-screen


src/qrcode/QRCode.tsx
```diff
@@ -15,19 +15,23 @@ import { currentAccountSelector } from 'src/web3/selectors'
...
-export default function QRCodeDisplay({ qrSvgRef }: Props) {
+export default function QRCodeDisplay({ qrSvgRef, dataType }: Props) {
   const data = useSelector(mapStateToProps, shallowEqual)
+
+  // ValoraDeepLink generates a QR code that deeplinks into the walletconnect send flow of the valora app
+  // Address generates a QR code that has the walletAddress as plaintext that is readable by wallets such as Coinbase and Metamask
   const qrContent = useMemo(
-    () => urlFromUriData(data),
-    [data.address, data.displayName, data.e164PhoneNumber]
+    () => (dataType === QRCodeDataType.ValoraDeepLink ? urlFromUriData(data) : data.address),
```
cajubelt, line 33: I used a qr code scanner to see what data is in the MetaMask QR code and found this: `ethereum:0x589865Bdd130C77C28B9906DC630bC4349820AEa@42220` which looks like [EIP-681](https://eips.ethereum.org/EIPS/eip-681)

cajubelt: Coinbase Wallet uses just the address. However, I think Coinbase can scan EIP-681 contents

cajubelt: looks like we could use this https://www.npmjs.com/package/eth-url-parser

cajubelt: I think EIP681 sets us up a bit better for cross chain, though supposedly some [bugs have resulted from other wallets not scanning it correctly](https://www.coindesk.com/tech/2021/03/25/coinbase-and-zengo-spar-over-qr-code-standards-that-could-strand-erc-20-tokens/). I don't really care, though

cajubelt: well I guess I would care if some centralized exchange parsed a QR code incorrectly and ended up sending funds to a contract address instead of the user's address

not as worried about other wallets, though

jh2oman: Forgot to link these discussions together https://valora-app.slack.com/archives/C02KBT0DAHJ/p1671051933339429

cajubelt: Jacob replied to this [here](https://valora-app.slack.com/archives/C02898GN22V/p1671057639473929?thread_ts=1671051997.691509&cid=C02898GN22V)

cajubelt, line 33: could be nice if this was a switch that threw an error if the dataType was not recognized, so it wouldn't just fail silently if the dataType was some other QRCodeDataType that we forgot to account for here.

could even get a compile-time error if there was some object mapping QRCodeDataType keys to functions that get qr contents from `data`

cajubelt: you could also get a compile-time error if there was an object mapping QRCodeDataType keys to expected QR code contents in QRCode.test.tsx, which could be nicer

cajubelt (approved):
added suggestion for avoiding silent failures if qr code types are added but the useMemo callback isn't updated

PR #3246, opened 2022-12-14
chore: bump web3.js to latest stable

silasbw:


package.json
```diff
@@ -189,7 +189,7 @@
     "uuid": "^8.3.0",
     "victory-native": "^36.6.8",
     "vm-browserify": "^1.1.2",
-    "web3": "1.3.5"
+    "web3": "1.8.1"
```
satish-ravi, line 192: do we still need to pin or can we use range?

silasbw:
@jh2oman wondering why you only did the patch bump in this relatively recent commit 2afe840466

satish-ravi (approved):

package.json
silasbw: No idea about the context, but looks like there was a habit of adding pinned packages at one point and we're just carrying it forward.  I'll add a `^`.

jh2oman:
@silasbw One of the `@celo/` dependencies (I forget which one) depends on that exact version of web3 and doesn't play nice in Typescript with some of our code that also uses the web3 package unless they are the exact same versions. You can see some of the failures in the unit tests

package.json
jh2oman: also there's a resolution for web3 🙃  https://github.com/valora-inc/wallet/pull/3246/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R337

PR #3247, opened 2022-12-15
feat(exchanges): Add new Exchange QR code screen

jophish:
### Description

For ACT-542.

Adds in a new `ExchangeQR` screen to eventually be displayed in the CICO Cash In flow in place of the current Exchanges screen. Uses a new QR code generator, which, while unable to match the designs for this ticket perfectly, does have limited capabilities for displaying a logo/customization, which we can explore at some later date.

See [this Slack thread](https://valora-app.slack.com/archives/C02KBT0DAHJ/p1671045146288859) for more details on the QR code generation situation.

### Other changes

Adds in a new QR code generator; while many capabilities of this generator are currently unused, this should set us up a bit better for later when we want to explore our options.

### Tested

Unit and manual testing. See video below. Note that accessing the new `ExchangeQR` screen through the button on the homescreen was only for testing purposes; this behavior is not included in the PR.

https://user-images.githubusercontent.com/569401/207771345-930a9d01-8d44-43f3-abbf-d98478648f9c.mp4

### How others should test

N/A, as this screen is not currently used anywhere in-app.

### Related issues

- Fixes ACT-542.

### Backwards compatibility

Yes.


locales/base/translation.json
```diff
@@ -1410,19 +1410,29 @@
...
     "spend": {
       "fiatExchangeTitle": "Spend Funds",
       "fiatExchangeSubtitle": "Spend on gift cards and other purchases.",
       "selectCurrencyTitle": "Spend Funds"
+    },
+    "exchange": {
+      "copyAddress": "Copy Address",
+      "informationPartOne": "Only use exchanges or assets ",
+      "informationPartTwo": "compatible with the Celo network ",
+      "informationPartThree": "to prevent loss of funds.",
```
satish-ravi, line 1433: Would this break up work for other languages?

src/qrcode/StyledQRGen.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+function StyledQRCode({
+  value,
+  size = 100,
+  svgRef,
+}: {
+  value: string
+  size: number
+  color?: string
+  backgroundColor?: string
```
satish-ravi, line 13: Would these params be used eventually? if so, should there be a todo and/or is this tracked in a separate ticket?

src/qrcode/QRCode.tsx
```diff
@@ -39,6 +40,19 @@ export default function QRCodeDisplay({ qrSvgRef }: Props) {
   )
 }
 
+export function StyledQRCodeDisplay({ qrSvgRef }: Props) {
```
satish-ravi, line 43: nit: Would this eventually replace the existing `QRCodeDisplay` or do we want to keep both? If we want to keep both, maybe both should be regular exports or be in separate files (like `StyledQRCodeGen`) instead of `QRCodeDisplay` being a default export?

src/fiatExchanges/ExchangeQR.tsx
```diff
@@ -0,0 +1,191 @@
+import { RouteProp } from '@react-navigation/native'
```
satish-ravi, line 1: nit: Should this file be named `ExchangeQRScreen.tsx` to stay consistent with other files in this folder? Although I guess we also `ExternalExchanges.tsx` so don't feel too strongly

src/fiatExchanges/ExchangeQR.tsx
```diff
@@ -0,0 +1,191 @@
...
+import { ExternalExchangeProvider } from 'src/fiatExchanges/ExternalExchanges'
+
+type Props = NativeStackScreenProps<StackParamList, Screens.ExchangeQR>
+
+export default function ExchangeQR({ route, navigation }: Props) {
+  const qrSvgRef = useRef<SVG>()
+  const { flow, exchanges } = route.params
+  const { t } = useTranslation()
+  const dispatch = useDispatch()
+  const address = useSelector(currentAccountSelector)
```
satish-ravi, line 39: https://github.com/valora-inc/wallet/blob/main/src/web3/selectors.ts#L4-L5
Should we use `walletAddressSelector`?

src/fiatExchanges/ExchangeQR.tsx
```diff
@@ -0,0 +1,191 @@
...
+      flow,
+    })
+    dispatch(shareQRCode(qrSvgRef.current))
+  }
+
+  const onPressCopy = () => {
+    ValoraAnalytics.track(FiatExchangeEvents.external_exchange_qr_copy_address, {
+      flow,
+    })
+    Clipboard.setString(address || '')
```
satish-ravi, line 62: Is the `|| ''` for type safety?

src/components/ExchangesBottomSheet.tsx
```diff
@@ -0,0 +1,83 @@
...
+  return (
+    <BottomSheet isVisible={isVisible} onBackgroundPress={onClose}>
+      <>
+        <Text style={styles.title}>{t('fiatExchangeFlow.exchange.bottomSheetTitle')}</Text>
+        <Text style={styles.info}>{t('fiatExchangeFlow.exchange.bottomSheetInfo')}</Text>
+        {exchanges.map((exchange, index) => {
+          return (
+            <React.Fragment key={`exchange-${exchange.name}`}>
+              {index > 0 && <View style={styles.separator} />}
+              <ExchangeOption exchange={exchange} onPress={onExchangePressed(exchange)} />
```
satish-ravi, line 51: nit: maybe rename `onExchangePressed` to something like `getOnExchangePress` or just inline the fn definition? It took me a couple of reads to understand this was returning a function.

src/components/ExchangesBottomSheet.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+  beforeAll(() => {
+    // @ts-ignore This avoids an error, see: https://github.com/software-mansion/react-native-reanimated/issues/1380
+    global.__reanimatedWorkletInit = jest.fn()
+  })
+
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  function renderPicker(visible: boolean) {
```
satish-ravi, line 45: nit: Picker -> renderExchangesBottomSheet?

src/fiatExchanges/ExchangeQR.tsx
```diff
@@ -0,0 +1,191 @@
...
+  headerLeft: () => (
+    <BackButton
+      eventName={FiatExchangeEvents.external_exchange_qr_back}
+      eventProperties={{ flow: route.params.flow }}
+    />
+  ),
+  headerTitle:
+    route.params.flow === CICOFlow.CashIn
+      ? i18n.t(`fiatExchangeFlow.cashIn.depositExchangeTitle`)
+      : i18n.t(`fiatExchangeFlow.cashOut.withdrawExchangeTitle`),
```
satish-ravi, line 153: Does this just exist so we handle this case as well and won't need a new version? For the exp, we'd just have cash in navigate to this screen right?

locales/base/translation.json
jophish: I was also wondering this; I don't know, but without breaking the text up like this in the translations file, I'm not sure how we'd be able to style only a subset like the designs require.

src/qrcode/StyledQRGen.tsx
jophish: I've removed these params for the time being, they were left over from my testing. I'll spin up a ticket to figure out proper styling for the new QR code.

src/qrcode/QRCode.tsx
jophish: AFAIK, for the time being I think we'll want to keep both -- I'll move the styled QR code wrapper over to its own file.

src/fiatExchanges/ExchangeQR.tsx
jophish: Yeah, we can also do `address!` here, but I think I prefer `|| ''` and it's a pattern we use elsewhere as well.

src/fiatExchanges/ExchangeQR.tsx
jophish: Good callout, thanks!

src/fiatExchanges/ExchangeQR.tsx
jophish: I'm not sure what convention we ought to be using for screen naming; I recall hearing conflicting opinions in the past about whether or not to suffix with `Screen`. I like having a `Screen` suffix for screens, but not sure what other folks think.

locales/base/translation.json
satish-ravi: could you do something like this https://github.com/valora-inc/wallet/blob/main/src/fiatconnect/LinkAccountScreen.tsx#L111-L123 where we have different stylings for different parts of this text: https://github.com/valora-inc/wallet/blob/main/locales/base/translation.json#L1302

jophish: done! worked out of the box.

src/fiatExchanges/ExchangeQR.tsx
```diff
@@ -0,0 +1,189 @@
...
+      />
+
+      <Text style={[styles.infoWrapper, fontStyles.regular, styles.exchangeText]}>
+        <Trans i18nKey="fiatExchangeFlow.exchange.informationText">
+          <Text
+            testID="bottomSheetLink"
+            style={[fontStyles.regular600, styles.link]}
+            onPress={onPressInfo}
+          >
+            {t('fiatExchangeFlow.exchange.informationPartTwo')}
```
satish-ravi, line 122: there's no partTwo anymore right?

jophish: :facepalm:

src/fiatExchanges/ExchangeQR.tsx
```diff
@@ -0,0 +1,187 @@
...
+        testID="copyButton"
+      />
+
+      <Text style={[styles.infoWrapper, fontStyles.regular, styles.exchangeText]}>
+        <Trans i18nKey="fiatExchangeFlow.exchange.informationText">
+          <Text
+            testID="bottomSheetLink"
+            style={[fontStyles.regular600, styles.link]}
+            onPress={onPressInfo}
+          ></Text>
```
satish-ravi, line 121: optional nit
```suggestion
          />
```

satish-ravi (approved):

PR #3248, opened 2022-12-15
chore: refactor recent and favourite dapps to use ids

kathaypacific:
### Description

Use the dappsList as the source of truth for recent and favourited dapps. There was already a code comment left ages ago to make this change and Jacob mentioned it again here https://github.com/valora-inc/wallet/pull/3243#pullrequestreview-1218173556, so i thought i'd just bite the bullet.

### Other changes

n/a

### Tested

n/a

### How others should test

n/a

### Related issues

- Fixes RET-522

### Backwards compatibility

Y


src/dapps/selectors.ts
```diff
@@ -60,3 +60,17 @@ export const dappCategoriesByIdSelector = createSelector(
...
+  dappsListSelector,
+  recentDappsIdsSelector,
+  (dapps, recentDappIds) => {
+    return recentDappIds.reduce((acc, recentDappId) => {
+      const recentDapp = dapps.find((dapp) => dapp.id === recentDappId)
+      if (recentDapp) {
+        return [...acc, recentDapp]
+      }
+      return acc
+    }, [] as Dapp[])
```
jeanregisser, line 74: nit: I find `reduce` always hard to read compared to a plain old loop :D

src/dapps/slice.ts
```diff
@@ -8,16 +8,15 @@ export interface State {
   dappsWebViewEnabled: boolean
   activeDapp: ActiveDapp | null
   maxNumRecentDapps: number
-  recentDapps: Dapp[]
+  recentDapps: string[]
```
jeanregisser, line 11:
```suggestion
  recentDappIds: string[]
```

src/dapps/slice.ts
```diff
@@ -8,16 +8,15 @@ export interface State {
...
   dappListApiUrl: string | null
-  // TODO: update type of recentDapps and activeDapp to be string
   dappsList: Dapp[]
   dappsListLoading: boolean
   dappsListError: string | null
   dappsCategories: DappCategory[]
   dappConnectInfo: DappConnectInfo
   dappFavouritesEnabled: boolean
-  favouriteDapps: Dapp[]
+  favouriteDapps: string[]
```
jeanregisser, line 19:
```suggestion
  favouriteDappIds: string[]
```

src/dapps/selectors.ts
kathaypacific: lol should we actively try to avoid reduce? we haven't been consistent about this, i like reduce but i know you don't 😅

jeanregisser: lol yeah I think our future selves will thank us for not abusing `reduce`

PR #3251, opened 2022-12-15
feat(analytics): integrate statsig experimentation with segment

anthonydaoud:
### Description

Currently, Statsig is integrated into the wallet separately from Segment. This PR changes the way that Statsig is initialized so that the ids used to bucket users are compatible with Segment. When events are logged to Segment, they can now be  sent to Statsig.

### Other changes

N/A

### Tested

No testing need

### How others should test

N/A

### Related issues

- ACT-467

### Backwards compatibility

N/A


src/analytics/ValoraAnalytics.ts
```diff
@@ -108,7 +108,8 @@ class ValoraAnalytics {
         : null
 
       await Statsig.initialize(STATSIG_API_KEY, stasigUser, {
-        overrideStableID: uniqueID, //Received an error if Stable ID not manually specified
+        // StableID should match Segment anonymousId
+        overrideStableID: await Analytics.getAnonymousId(),
```
cajubelt, line 112: good find.

worth noting that this means there will be a backwards-incompatible change to the data flowing through Statsig, meaning the data that goes from Statsig to BQ is going to have a breaking change. 

Can you let Gowtham know? And can you deal with the fallout from the breaking change? We probly need to copy the data we have into something that's clearly named to indicate it's old/incompatible, clear out the statsig tables in BQ and start fresh. Lmk if you want to discuss.

cajubelt (approved):

src/analytics/ValoraAnalytics.ts
anthonydaoud: I let him know and setup the tables in BQ.

src/config.ts
```diff
@@ -94,6 +94,7 @@ export const FIATCONNECT_NETWORK =
 export const STATSIG_ENV = {
   tier: DEFAULT_TESTNET === 'mainnet' ? 'production' : 'development',
 }
+export const DEFAULT_STATSIG_ID = 'default_statsig_id'
```
cajubelt, line 97: kind of a nit but I think this is a bit misleading. The only case we want this to be a user's statsig ID is if it's an e2e test. Seems safer to explicitly include e2e_test in the variable name.

src/analytics/ValoraAnalytics.ts
```diff
@@ -107,8 +108,10 @@ class ValoraAnalytics {
           }
         : null
 
+      const overrideStableID = isE2EEnv ? DEFAULT_STATSIG_ID : await Analytics.getAnonymousId()
```
cajubelt, line 111: why do we need this? can you add an explanation in a code comment?

cajubelt (approved):

src/analytics/ValoraAnalytics.ts
anthonydaoud: It's unclear why, but `getAnonymousId` caused many e2e tests to fail. I tried to figure out what was going on under the hood but am still not certain. Added a brief comment

PR #3252, opened 2022-12-15
feat(celo-news): add feed components

jeanregisser:
### Description

This PR adds the Celo news feed components. 
It's using fake data for now, another PR will connect it to the API.

### Tested

https://user-images.githubusercontent.com/57791/208106513-70f410b8-bf8d-403c-8e9f-45baa2c2c7b5.mp4

### How others should test

N/A

### Related issues

- RET-519

### Backwards compatibility

Yes


src/exchange/CeloNewsFeedItem.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+describe('CeloNewsFeedItem', () => {
+  it('renders and behaves correctly', async () => {
+    const tree = render(<CeloNewsFeedItem article={TEST_ARTICLE} />)
+
+    // Check key elements are rendered
+    expect(tree.queryByText(TEST_ARTICLE.title)).toBeTruthy()
+    expect(tree.queryByText(TEST_ARTICLE.author)).toBeTruthy()
+    expect(tree.queryByText('1 month')).toBeTruthy()
+    expect(tree.UNSAFE_getByType(Image).props.source).toEqual({ uri: TEST_ARTICLE.articleImage })
```
kathaypacific, line 30: do you prefer this UNSAFE method rather than a test id on the image?

kathaypacific (approved):
Looks great!!

src/exchange/CeloNewsFeedItem.test.tsx
jeanregisser: Not really, I guess I should use `testID` instead. Will work better if we add another `Image` (author image?).

src/exchange/CeloNewsFeed.tsx
```diff
@@ -0,0 +1,140 @@
...
+  return <View style={styles.separator} />
+}
+
+export default function () {
+  const { t } = useTranslation()
+
+  function onPressReadMore() {
+    // TODO: use a remote config for this URL
+    navigate(Screens.WebViewScreen, { uri: 'https://blog.celo.org' })
+  }
```
jacob-israel-turner, line 89:
```suggestion
  useCallback(function onPressReadMore() {
    // TODO: use a remote config for this URL
    navigate(Screens.WebViewScreen, { uri: 'https://blog.celo.org' })
  }, [])
```

src/exchange/CeloNewsFeed.tsx
```diff
@@ -0,0 +1,140 @@
...
+      keyExtractor={keyExtractor}
+      ItemSeparatorComponent={ItemSeparator}
+      ListHeaderComponent={
+        <View style={styles.header}>
+          <Text style={styles.headerTitle}>{t('celoNews.headerTitle')}</Text>
+          <Text style={styles.headerDescription}>{t('celoNews.headerDescription')}</Text>
+        </View>
+      }
+      ListFooterComponent={
+        <>
```
jacob-israel-turner, line 104: This should be componentized or memoized as well

src/exchange/CeloNewsFeedItem.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+interface Props {
+  article: CeloNewsArticle
+  testID?: string
+}
+
+export default function CeloNewsFeedItem({ article, testID }: Props) {
+  function onPress() {
+    navigate(Screens.WebViewScreen, { uri: article.link })
+  }
```
jacob-israel-turner, line 22:
```suggestion
  useCallback(function onPress() {
    navigate(Screens.WebViewScreen, { uri: article.link })
  }, [article.link])
```

jacob-israel-turner: There are some handy linting rules that should help us catch un-memoized props, I'll look at adding them to the wallet.

jacob-israel-turner (approved):
Some nits, but looks good to me!

src/exchange/CeloNewsFeedItem.tsx
kathaypacific: this is my opinion but i don't believe that pre or over optimising brings the best outcomes...there's an upfront cost to memoising something, are we really going to gain much from memoising simple functions like these here?

jeanregisser: +1 to what @kathaypacific said. My memory is a bit fuzzy but I think we discussed this in the team a long time ago and decided to not only memoize callbacks when we identify clear bottlenecks.
I'll try to find the discussion.

jacob-israel-turner: To be clear - this is to prevent re-renders of children components due to referential inequality, not because the function itself is expensive.

I think there are generally two schools of thought regarding memoization.

1) **Memoize when it's needed** - When performance issues arise (due to either intense computations while rendering, or massive re-renders because of unstable references as props), you add memoization. This requires that you monitor for performance issues. It also means breaking out some tooling and combing through component trees to identify the source of performance issues. Fortunately React is generally really fast. Unfortunately, if you _do_ find yourself in this position, fixing the issue is generally non-trivial - but that's just my experience. There are wizards in this realm that can identify and fix perf issues very quickly.

2) **Memoize aggressively to pre-empt performance issues** - Memoize all non-primitive values passed as props to child components. Yes, this is a pre-optimization. The cost is more memory usage (although I believe it's really quite a small increase, as React stores props between renders regardless if the value is memoized) and cognitive load. Fortunately the cognitive load can be solved with linting. The benefit here is that you're much less likely to need to invest significant time tracking down performance issues.

I wouldn't say I have a _strong_ opinion in this regard, but I've generally been a member of the second camp. I'd rather invest a little up-front cost in adding linting rules to enforce memoization, than potentially have to invest a bunch of time tracking down and fixing performance issues in the future.

jacob-israel-turner: I'm ok with that approach if it's something the team has intentionally decided upon. I'd love to find the discussion if you can find it.

MuckT: The code legibility hit if we use memoization for everything is also a consideration. It would be interesting to see the results of the profiler for this simple case to see if there is enough of an improvement to justify using memoize more aggressively. Right now I don't think we a clear picture on how much of a performance boost we would get with memoization, but think that would be helpful for the discussion.

kathaypacific: maybe it is worth discussing again but i am firmly in the first group you describe @jacob-israel-turner. performance optimisations often comes at a cost (legibility and maintenance, as Tom mentions above) and in my opinion we should be intentional and strategic about what we are doing. not only in the performance topic but also more generally. this component is already a leaf component so i'm not sure that i understand your point about preventing re-renders of children components here, it seems very inexpensive to let react native handle re-renders of views and texts and images efficiently.

PR #3253, opened 2022-12-15
chore: bump svgs to latest version

silasbw:


satish-ravi (approved):

jeanregisser:
Ah this reminds me we could even remove this dependency and change all imports to use `react-native-svg` instead.

silasbw:
thanks for pointing that out @jeanregisser .  I'll try moving us over to `react-native-svg` completely and removing `svgs`.

PR #3256, opened 2022-12-16
feat: add dapp favouriting section

kathaypacific:
### Description

This PR:
- Add dapp favourites to the dapps screen.
- Break up the dapps screen file as it was getting too big

Simulator Screen Shot - iPhone 11 - 2022-12-16 at 11 29 21

Not included in this PR: 
- adding the stars on dapp card
- analytics

### Other changes

N/A

### Tested

Added unit tests

### How others should test

n/a

### Related issues

- Fixes RET-522

### Backwards compatibility

y


src/dapps/types.ts
```diff
@@ -2,6 +2,8 @@ export enum DappSection {
   RecentlyUsed = 'recently used',
   Featured = 'featured',
   All = 'all',
+  FavoritesDappScreen = 'favorites dapp screen',
+  FavoritesHomeScreen = 'favorites home screen',
```
jeanregisser, line 6: What's the difference between these 2?

jeanregisser (approved):
Great! 👍 

Though it's a bit hard for reviewers to combine refactor which moves things + a new feature.
Would have been best to split in 2 PRs.

src/dapps/types.ts
jeanregisser: Oh ok, I guess the screen is already part of the other props attached, but doesn't hurt to add this explicitely.

PR #3257, opened 2022-12-16
feat: add UI for dapp favouriting on dapp cards

kathaypacific:
### Description

This PR adds the UI for the favouriting / unfavouriting action + analytics events.

https://user-images.githubusercontent.com/20150449/208096028-e4425748-92fc-44db-83af-b06d3f45c2f7.mp4

### Other changes

n/a

### Tested

Manually

### How others should test

n/a

### Related issues

- Fixes RET-522

### Backwards compatibility

Y


src/analytics/Properties.tsx
```diff
@@ -1323,10 +1323,13 @@ interface CICOEventsProperties {
...
 
-interface DappEventProperties {
+interface DappProperties {
   categoryId: string
   dappId: string
   dappName: string
+}
+
+interface DappEventProperties extends DappProperties {
   section: DappSection
```
jeanregisser, line 1333: Why not keeping this even for favorite / unfavorite?

jeanregisser (approved):
👍

src/analytics/Properties.tsx
kathaypacific: good question, i didn't include these because the favourite / unfavourite actions can only come from one place (we're removing the dapp from the "all" list on favourite for MVP after all...) and the horizontalPosition was for recently used dapps only

jeanregisser: Ok, but aren't we gonna use a horizontal scrollview for favorites on the home screen?

kathaypacific: yes absolutely, but i haven't created events for that yet. those events can use the `DappEventProperties` type i think

PR #3259, opened 2022-12-16
chore: update PR template

jeanregisser:
### Description

I'm proposing to update our PR template.
- I removed the "Other changes" section to discourage doing unrelated changes that should be better done in separate PRs.
- Combined the "Tested" and "How others should test" into a new "Test plan" section, thanks @kathaypacific for the suggestion
- Turning the descriptive placeholders into hidden text to avoid seeing them unless we're editing.

What do you think?

### How others should test

N/A

### Backwards compatibility

Yes


kathaypacific (approved):
nice! what do you think about combining the "tested" and "how others should test" items? (could just remove "how others should test" because the steps are usually the same as in "tested"?)

jacob-israel-turner (approved):
🎉

jeanregisser:
> nice! what do you think about combining the "tested" and "how others should test" items? (could just remove "how others should test" because the steps are usually the same as in "tested"?)

Great idea, do you have a good placeholder update for this? Otherwise I'll take a stab at it 😄

kathaypacific:
> > nice! what do you think about combining the "tested" and "how others should test" items? (could just remove "how others should test" because the steps are usually the same as in "tested"?)
> 
> Great idea, do you have a good placeholder update for this? Otherwise I'll take a stab at it 😄

hmmmm i have been using the test sections to describe any manual testing steps (since we can already read the unit tests from the PR) - maybe this is the way you have been using them too? i was thinking a description like "If it is not obvious, add any manual testing steps or scenarios that were used to test this PR", what do you think?

jeanregisser:
Thanks @kathaypacific, I took your feedback and adapted it.
What do you think?

kathaypacific:
> Thanks @kathaypacific, I took your feedback and adapted it. What do you think?

Looks great!! 🚀

PR #3261, opened 2022-12-16
feat(celo-news): connect it to our API

jeanregisser:
### Description

Connect the Celo news feed to our API.

### Tested

**loading + success**

https://user-images.githubusercontent.com/57791/208414473-88673824-0bb2-42ed-abc8-062c0f1f4325.mp4

**loading + error with retry**

https://user-images.githubusercontent.com/57791/208414532-7256d792-392c-4247-832a-128e657398a2.mp4

### Related issues

- Fixes RET-519

### Backwards compatibility

Yes

### DIFF
--- a/src/components/EmptyView.tsx
+++ b/src/components/EmptyView.tsx
@@ -1,0 +1,34 @@
+import React from 'react'
+import { StyleProp, StyleSheet, Text, View, ViewStyle } from 'react-native'
+import fontStyles from 'src/styles/fonts'
+import { Spacing } from 'src/styles/styles'
+
+interface Props {
+  style?: StyleProp<ViewStyle>
+  text: string
+  children?: React.ReactNode
+}
+
+/**
+ * A component to render an empty state with a message and optional children.
+ */
+export default function EmptyView(props: Props) {
+  return (
+    <View style={[styles.container, props.style]}>
+      <Text style={styles.text}>{props.text}</Text>
+      {props.children}
+    </View>
+  )
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    padding: Spacing.Thick24,
+    paddingTop: 75,
+  },
+  text: {
+    ...fontStyles.regular,
+    marginBottom: Spacing.Regular16,
+  },
+})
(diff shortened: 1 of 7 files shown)


src/components/EmptyView.tsx
```diff
@@ -0,0 +1,34 @@
...
+interface Props {
+  style?: StyleProp<ViewStyle>
+  text: string
+  children?: React.ReactNode
+}
+
+/**
+ * A component to render an empty state with a message and optional children.
+ */
+export default function EmptyView(props: Props) {
```
kathaypacific, line 15: i don't have a big problem with this component but do you think it's worth abstracting out into the general components folder? my gut reaction to this component is that we'll forget that it exists or want to use it but the padding it comes with is doesn't match the design...

kathaypacific (approved):
looks great!!

one thought i had was maybe we want to keep the header title and description around even when there is an error? i think it'd minimise the content moving around + add more context for the error, but just my opinion!

src/components/EmptyView.tsx
jeanregisser: Yes my goal is to try to force us to think about empty / error states and display them consistently.
I hope we can get to that :D

src/exchange/CeloNewsFeed.test.tsx
```diff
@@ -20,9 +91,20 @@ describe('CeloNewsFeed', () => {
 
     // Check we can see the Celo news header
     expect(tree.queryByText('celoNews.headerTitle')).toBeTruthy()
+    // Check we can see the loading spinner
+    expect(tree.queryByTestId('CeloNewsFeed/loading')).toBeTruthy()
```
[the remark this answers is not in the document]
jeanregisser: Yes but it can't be used like this: `expect(tree.getByTestId('CeloNewsFeed/loading')).toBeFalsy()` which I forgot to add later in the test.

jeanregisser:
> looks great!!
> 
> one thought i had was maybe we want to keep the header title and description around even when there is an error? i think it'd minimise the content moving around + add more context for the error, but just my opinion!

Yes I'm discussing this with Nitya / Reed, I didn't like how it looked with my initial attempt, but we'll come up with something nice I'm sure.

jeanregisser:
Merging now, I'll address changes to the error UI in a follow up PR.

PR #3264, opened 2022-12-19
feat: use redux to store dapp favorited state

kathaypacific:
### Description

This PR:
- hooks up the favoriting functionality to redux
- updates selectors to remove dapps from the normal dapps list if favorited
- removes empty categories from the dapps list (in case all dapps in that category are favorited)
- renames test data to be generic in unit tests

https://user-images.githubusercontent.com/20150449/208614723-3a5d27e5-e7a6-4bae-aaa9-faea68f123ac.mp4

Not included:
- the toast that appears after a dapp is favorited. it seems like our current toast lib is way too limited for what we need (does not accept a onPress callback) so i'll see if i can put together a custom toast quickly.

### Other changes

n/a

### Tested

n/a

### How others should test

n/a

### Related issues

- Fixes RET-523

### Backwards compatibility

n/a


src/dapps/selectors.ts
```diff
@@ -51,8 +54,11 @@ export const dappCategoriesByIdSelector = createSelector(
       }
     })
     dapps.forEach((dapp) => {
-      mappedCategories[dapp.categoryId]?.dapps.push(dapp)
+      if (!favoriteDappIds.includes(dapp.id)) {
+        mappedCategories[dapp.categoryId]?.dapps.push(dapp)
+      }
```
jeanregisser, line 59: A comment here would be nice, to explain why we're doing this.

src/dapps/slice.ts
```diff
@@ -83,6 +87,14 @@ export const slice = createSlice({
       state.dappsListLoading = false
       state.dappsListError = action.payload.error
     },
+    favoriteDapp: (state, action: PayloadAction<FavoriteDappAction>) => {
+      state.favoriteDappIds = [...new Set([action.payload.dappId, ...state.favoriteDappIds])]
```
jeanregisser, line 91: Intuitively I would have added the new favorite at the end. If not already done, maybe it's worth checking with Reed?

jeanregisser (approved):
Awesome! 👍 
Can't wait to see the animation ✨

src/dapps/selectors.ts
kathaypacific: hmmm i added a comment as you suggested but not sure why it's not showing up in this diff, it is showing up on the file if you browse the branch though https://github.com/valora-inc/wallet/blob/9332286834a5e1c023b22e1eb6c564a0a110069e/src/dapps/selectors.ts#L57 🙈 take my word for it

src/dapps/slice.ts
kathaypacific: thanks for flagging this! i put the new dapps at the top because i was thinking about the recently used dapps / carousel on the home screen but the reverse also makes sense. i checked with reed and he said to follow whatever rule we have on the dapps screen now, so adding new dapps to the bottom of a section makes sense. i've updated the order

PR #3266, opened 2022-12-19
feat: add custom toast for dapp favorited action

kathaypacific:
### Description

This PR:
- add a custom Toast component that accepts a callback on press
- add hook that manages the Toast appearing / on press behaviour on the dapps screen
- add some state management to ensure that favoriting in quick succession produces a toast that lasts for the full timeout

https://user-images.githubusercontent.com/20150449/208656801-bcd989c5-ed53-43ac-87f7-a4ff46876034.mp4

### Other changes

N/A

### Tested

As in the video

### How others should test

N/A

### Related issues

- Fixes RET-523

### Backwards compatibility

Y


src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -88,6 +88,63 @@ export function DAppsExplorerScreen() {
     setHelpDialogVisible(false)
   }
 
+  // create sections for all screen content to facilitate scroll to top using a SectionList
+  const sections = []
```
kathaypacific, line 92: i had to refactor how we generate sections to enable the scroll to top functionality on tap of the toast. for a section list, you can only scroll to sections / section items but the top of this screen is at the top of the section list header. the only way is to specify an offset _in pixels_ to account for the header, which doesn't work for us if the favorites list is in the header since its length is dynamic.

https://reactnative.dev/docs/sectionlist#scrolltolocation

src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -125,28 +182,8 @@ export function DAppsExplorerScreen() {
...
-                    <FavoriteDappsSection onPressDapp={onSelectDapp} />
-                  </>
-                )}
-
-                {(featuredDapp || dappFavoritesEnabled) && (
-                  <Text style={styles.sectionTitle}>{t('dappsScreen.allDapps')}</Text>
-                )}
-              </>
-            }
+            // @ts-ignore TODO: resolve type error
```
kathaypacific, line 185: spent way too long trying to get this error to go away and i couldn't do it, i think it might be that this reanimated lib type definition doesn't play nice with the RN types. it didn't inspire confidence to see this `ref` type named as `LegacyRef` in the reanimated type doc :/ open for ideas on how to resolve this / i think it might not be worth the effort to resolve this

src/dappsExplorer/useDappFavoritedToast.tsx
```diff
@@ -0,0 +1,57 @@
+import React, { useEffect, useMemo, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import { SectionList } from 'react-native'
+import ToastWithCTA from 'src/components/ToastWithCTA'
+import { Dapp } from 'src/dapps/types'
+
+const TOAST_DISMISS_TIMEOUT = 5000
+
+const useDappFavoritedToast = (sectionListRef: React.RefObject<SectionList>) => {
```
kathaypacific, line 9: made this hook to look like the `useOpenDapp` hook, to decouple the favoriting functionality as much as possible from the screen itself

src/dappsExplorer/useDappFavoritedToast.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const { t } = useTranslation()
+
+  // do not use favoritedDapp to show / hide the toast, as the content of the toast depends on the dapp name during transitions
+  const [favoritedDapp, setFavoritedDapp] = useState<Dapp | null>(null)
+  const [showToast, setShowToast] = useState(false)
+
+  const timeoutIdRef = useRef<NodeJS.Timeout | null>(null)
```
kathaypacific, line 16: this bit is a bit convoluted, happy for any simplification suggestions. what i really want is to reset the timeout after every dapp favorite action, so that the toast stays on the screen for the full 5 seconds. before i started tracking the timeout id, favoriting a dapp then favoriting another dapp 4 seconds later = the first timeout got in the way and dismissed the toast for the second dapp 1 second later.

src/dappsExplorer/DAppsExplorerScreen.tsx
jeanregisser: I don't fully get this, could we scroll to the top of the whole list instead? The favorite should always be visible from the very top, right?
Sorry may be missing some details on the behavior.

src/dappsExplorer/useDappFavoritedToast.tsx
jeanregisser: I'm proposing a simplification, let me know if I missed a case: https://github.com/valora-inc/wallet/pull/3273

jeanregisser (commented):
Looks great! 😍

src/dappsExplorer/useDappFavoritedToast.tsx
kathaypacific: thank you for this! 🧙🏻‍♂️

src/dappsExplorer/DAppsExplorerScreen.tsx
jeanregisser: Ok thanks for the explanation.
`SectionList` is using a ScrollView under the hood,
what about using `sectionListRef.current?.getScrollResponder()?.scrollTo({ y: 0, animated: true })` instead?

kathaypacific: that works! i didn't see this in the docs but i see it in the types def now, thanks for the suggestion! i've updated :)

src/dappsExplorer/DappCard.tsx
```diff
@@ -48,6 +49,9 @@ function DappCard({ dapp, section, onPressDapp }: Props) {
     } else {
       ValoraAnalytics.track(DappExplorerEvents.dapp_favorite, eventProperties)
       dispatch(favoriteDapp({ dappId: dapp.id }))
+      if (onFavoriteDapp) {
+        onFavoriteDapp(dapp)
+      }
```
jeanregisser, line 54: ultra nit:
```suggestion
      onFavoriteDapp?.(dapp)
```

jeanregisser (approved):
🚀

PR #3267, opened 2022-12-19
fix(qrcode): Switch the Exhange QR code screen to encode address

jh2oman:
### Description

Old QR screens were already updated to be able to handle address, new screen needs it too. 

Exchange QR screen should always encode address.

### Test plan

Tested by updating the code locally to display the Exchange QR Code screen and successfully scanned the QR code with metamask and coinbase. 

### Related issues

https://linear.app/valora/issue/ACT-543/update-qr-code-within-qr-code-screen

### DIFF
--- a/src/fiatExchanges/ExchangeQR.tsx
+++ b/src/fiatExchanges/ExchangeQR.tsx
@@ -1,33 +1,33 @@
+import Clipboard from '@react-native-clipboard/clipboard'
 import { RouteProp } from '@react-navigation/native'
-import { StackParamList } from 'src/navigator/types'
 import { NativeStackScreenProps } from '@react-navigation/native-stack'
-import React, { useRef, useState, useLayoutEffect } from 'react'
+import React, { useLayoutEffect, useRef, useState } from 'react'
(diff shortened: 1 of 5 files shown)


src/fiatExchanges/ExchangeQR.tsx
```diff
@@ -88,7 +88,7 @@ export default function ExchangeQR({ route, navigation }: Props) {
   return (
     <View style={styles.container}>
       <View style={styles.qrContainer}>
-        <StyledQRCode qrSvgRef={qrSvgRef} />
+        <StyledQRCode dataType={QRCodeDataType.Address} qrSvgRef={qrSvgRef} />
```
satish-ravi, line 91: any easy way to unit test this?

jh2oman: I couldn't find an easy way to test this. We could add snapshots, but IDK if its worth that. I did add new unit tests to the hook so we have at least some additional security.

satish-ravi (approved):

PR #3269, opened 2022-12-20
chore(dev-deps): remove @babel/runtime

satish-ravi:
### Description

Flagged by depcheck. Per [docs](https://babeljs.io/docs/en/babel-runtime), looks like this needs to be used as a runtime dependency (not a dev dependency) along with [@babel/plugin-transform-runtime](https://babeljs.io/docs/en/babel-plugin-transform-runtime), but we don't use this plugin.

### Test plan

CI

### Related issues

N/A

### Backwards compatibility

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -199,7 +199,6 @@
     "@babel/plugin-proposal-decorators": "^7.12.1",
     "@babel/plugin-proposal-numeric-separator": "^7.14.5",
     "@babel/plugin-transform-named-capturing-groups-regex": "^7.17.12",
-    "@babel/runtime": "^7.12.5",
     "@faker-js/faker": "^5.5.3",
     "@graphql-codegen/add": "^1.10.0",
     "@graphql-codegen/cli": "^1.10.0",


jeanregisser (approved):
I have a vague memory this may still be used indirectly.
But maybe not relevant anymore, which seems to be confirmed by the tests working right.
Unless it was needed on older OS versions?

satish-ravi:
@jeanregisser Yea I wasn't 100% sure about this. This is still part of the RNDiffApp https://github.com/react-native-community/rn-diff-purge/blob/version/0.70.6/RnDiffApp/package.json#L18, although not sure how it's used. Also, this package still gets installed as a sub dependency (yarn.lock never changed on this PR). But CI passes and local builds also work so I am leaning towards removing it. But we can also leave it in and add to .depcheckrc ignores. What do you think?

jeanregisser:
Ok, I think the reason that it's specified explicitly is to make sure a compatible version is present.
We do have other deps pulling it, so we're fine.
But maybe I've a slight preference to add it to the ignore deps, so next time we upgrade React Native, we just follow the template and don't think about it.
What do you think?

satish-ravi:
yeah sounds good. Closing in favor of adding to ignore list.

PR #3271, opened 2022-12-20
feat(celo-news): remote config for "Read more" link

jeanregisser:
### Description

Use the remote config for the link opened by tapping on "Read more".
Also moves Celo News related config into a `celoNews` Object parameter, since they are closely related.
Later, we may add additional params, like page size.

### Test plan

Tapping on the "Read more" button at the bottom of the Celo News feed should open the link configured in the remote config under the `celoNews` parameter.

### Related issues

- Fixes RET-519

### Backwards compatibility

Yes


src/exchange/CeloNewsFeed.tsx
```diff
@@ -61,10 +63,14 @@ export default function CeloNewsFeed() {
   const { t } = useTranslation()
 
   const asyncArticles = useFetchArticles()
+  const readMoreUrl = useSelector(celoNewsConfigSelector).readMoreUrl
```
kathaypacific, line 66: `const { readMoreUrl } = useSelector(celoNewsConfigSelector)`?

src/redux/store.ts
```diff
@@ -23,7 +23,7 @@ const persistConfig: PersistConfig<RootState> = {
   key: 'root',
   // default is -1, increment as we make migrations
   // See https://github.com/valora-inc/wallet/tree/main/WALLET.md#redux-state-migration
-  version: 99,
+  version: 100,
```
kathaypacific, line 26: wow, 3 digits now! 🧙🏻‍♂️

kathaypacific (approved):
LGTM!

i don't mind this approach, i'm not sure that i totally see the benefits over having 2 separate remote config variables though 😅 it seems like the enabled / the read more url variables are not very tightly coupled in terms of use cases (we would change `enabled` for feature flagging and experiments, but idk when we'd change the url) so updating one exposes the other one to updates since the whole variable is refreshed. idk, i don't have strong objections to this approach either so i've approved

PR #3272, opened 2022-12-20
fix: only display invite rewards active if user is verified via CPV

kathaypacific:
### Description

This is something that I overlooked when implementing the invite rewards UI some weeks ago - for v4 where we require CPV, we should only display the UI if the user has verified their phone number.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y


M-shin (approved):
lgtm

PR #3273, opened 2022-12-20
Proposal to simplify timeout management

jeanregisser:
Tried it locally, seems is covering what we need.
The toast always stays for the full duration after every favoriting action.


kathaypacific (approved):
i think this was the obvious solution that i was looking for but missed 🤦🏻‍♀️ thank you!

PR #3274, opened 2022-12-20
feat(celo-news): skeleton placeholder while loading

jeanregisser:
### Description

This replaces the loading spinner by a skeleton placeholder with a shimmering animation in the Celo news feed.

Why?
> - Skeleton screens are perceived as being shorter in duration when compared against a blank screen and a spinner.
> - Use of loader / spinner creates a period of uncertainty for the user since the load time is unknown.
> - It drives user’s attention to progress instead of waiting time.
> - For creating an illusion of short page load times in apps and on the web.

From https://blog.prototypr.io/skeleton-loader-an-overview-purpose-usage-and-design-173b5340d0e1

Next: use this in other screens where it makes sense (transaction list, dapp list, etc).

I had a look at several libraries to implement this and settled on https://github.com/chramos/react-native-skeleton-placeholder as it had the "best" implementation using a mask and linear gradient so the effect works across several isolated views, with an animation driven by the native driver.
In the future, I could see us fork the lib or include it in our components to further customize it to our needs.

### Test plan

https://user-images.githubusercontent.com/57791/208921876-cdc18140-76f8-4a53-8ca9-6e505de8c31c.mp4

Note: I added an artificial delay of 5 secs in the above video so we can see it better

### Related issues

- Fixes RET-519

### Backwards compatibility

Yes


src/exchange/CeloNewsFeedItem.tsx
```diff
@@ -51,6 +53,28 @@ export default function CeloNewsFeedItem({ article, testID }: Props) {
...
+    <View style={styles.contentContainer}>
+      <View style={{ ...styles.author, width: 60 }} />
+      <View style={styles.row}>
+        <View style={{ flex: 1 }}>
+          <View style={{ ...styles.title, flex: undefined, marginBottom: 4 }} />
+          <View style={{ ...styles.title, flex: undefined, width: '60%' }} />
+        </View>
+        <View style={styles.image} />
+      </View>
+    </View>
```
jeanregisser, line 74: I broke our rule of not using inlines styles here because I thought it made more sense to have it directly there.
But let me know if you think I should really not do this.

kathaypacific (approved):
LGTM!! any plans to submit the patch to the skeleton project? :D

jeanregisser:
Yes the patch was submitted already: https://github.com/chramos/react-native-skeleton-placeholder/pull/71
I left a comment about this in it.

PR #3276, opened 2022-12-20
chore(e2e): update e2e env variable

anthonydaoud:
### Description

Currently, the e2e `env.enc` file has a value called `MOCK_PROVIDER_API_KEY` that is set to an old API key value. This causes the Fiat Connect KYC transfer out test to fail when running the e2e tests locally. 

This PR updates the `MOCK_PROVIDER_API_KEY` with the correct value, which can be found in google secret manager.

### Test plan

Verified that this fixes the issue by running e2e tests locally with the updated `env.enc` file.

### Related issues

N/A

### Backwards compatibility

N/A


satish-ravi (approved):
Approved assuming this was tested by running e2e tests locally. Can you include that in the test plan section of the PR description?

anthonydaoud:
Just updated it!

PR #3287, opened 2022-12-22
test(e2e): dapp list recently used

MuckT:
### Description

- Adds the `getElementText` utility method.
- Adds the `getElementTextList` utility method.
- Creates a `DappList.spec.js` which contains the `DappListRecent` test suite.

```
  Dapp List
    Recent Dapps
      ✓ should show most recently used dapp leftmost on home screen :ios: (15801 ms)
      ✓ should show prompt to open most recently used dapp (3013 ms)
      ✓ should correctly open most recently used dapp (16275 ms)
      ✓ should correctly navigate via all dapps card (14152 ms)
      ○ skipped should show most recently used dapp leftmost on home screen :android:
```

### Test plan

Tested locally on Android and iOS.

### Related issues

- RET-416

### Backwards compatibility

Yes


MuckT:
@kathaypacific it appears the dapp list is failing on the initial loads in CI as seen in [this run](https://github.com/valora-inc/wallet/actions/runs/3833335073/jobs/6524744208).

e2e/src/usecases/DappListRecent.js
```diff
@@ -0,0 +1,180 @@
...
+async function navigateToDappList() {
+  await device.disableSynchronization()
+  await waitForElementId('Hamburger')
+  await element(by.id('Hamburger')).tap()
+  await waitForElementId('dapps-explorer-icon')
+  await element(by.id('dapps-explorer-icon')).tap()
+  await waitFor(element(by.id('DAppExplorerScreen/loading')))
+    .not.toBeVisible()
+    .withTimeout(30 * 1000)
+  await sleep(5 * 1000)
```
kathaypacific, line 20: could we replace this waitFor loader and sleep with something like "wait for there to be elements with test id `DappCard`"? since the loader appears asynchronously (triggered by a saga) there might be a race condition where sometimes line 17 evaluates to true already before the loader has had a chance to appear...it'd also be nice not to wait 5 seconds if we can avoid it :)

e2e/src/usecases/DappListRecent.js
```diff
@@ -0,0 +1,180 @@
...
+      const dappPressedName = dappPressed.label.split('Go to ')[1]
+
+      // Confirm dapp open and navigate back to home
+      await element(by.id('ConfirmDappButton')).tap()
+      await element(by.text('Close')).tap()
+      await navigateToHome()
+
+      // Check that recently used dapp is now first in list
+      const endRecentlyUsedDapps = await getElementTextList('RecentDapp-name')
+      jestExpect(dappPressedName).toEqual(endRecentlyUsedDapps[0])
```
kathaypacific, line 77: maybe we can also add a `jestExpect(endRecentlyUsedDapps.length).toEqual(startRecentDappCount + 1)`?

kathaypacific (commented):
Looks great! What do you think about not having iOS and android specific tests? it looks like the 2nd and 3rd test for each platform is the same, so could we just combine the tests and have some conditions for iOS / android in the first test?

if i understand correctly, the main difference in the first test is that android doesn't scroll well so we use the first dapp in the list as the recently used. i think it's reasonable to do this for iOS too to avoid a different implementation, but maybe i am missing something :)

MuckT:
@kathaypacific reused the 2 - 4th tests on Android, but kept the first test separate to avoid branching logic for iOS and Android within the test. I was able to remove the sleep from in `navigateToDappList`; however, checking for the existence of the `DappCard` wasn't working due to multiple matching elements - instead of using the sometimes flaky `AtIndex(0)` I kept the waitFor `DAppExplorerScreen/loading` not to be visible.

src/home/DappsCarousel.tsx
```diff
@@ -149,7 +149,12 @@ function DappsCarousel({ onSelectDapp }: Props) {
                   resizeMode="cover"
                   testID={`${testID}/Icon`}
                 />
-                <Text style={styles.dappName} numberOfLines={1} ellipsizeMode="tail">
+                <Text
+                  style={styles.dappName}
+                  numberOfLines={1}
+                  ellipsizeMode="tail"
+                  testID="RecentDapp-name"
```
kathaypacific, line 156: nit: testID={`${testID}/name`}

e2e/src/DappList.spec.js
```diff
@@ -0,0 +1,14 @@
+import DappListRecent from './usecases/DappListRecent'
+import { quickOnboarding } from './utils/utils'
+import { launchApp } from './utils/retries'
+
+describe('Dapp List', () => {
+  beforeAll(async () => {
+    await quickOnboarding()
+    // Relaunch app to ensure dapp list loads
+    // TODO: Remove this once we have a better way to ensure dapp list loads
+    await launchApp({ newInstance: true })
```
kathaypacific, line 10: the action to fetch the dapps list is called for the first time when the app sagas are kicked off, so it seems like relaunching the app shouldn't add any value in ensuring the dapp list loads? could we either remove this or add some details to the comment about why this helps?

e2e/src/usecases/DappListRecent.js
```diff
@@ -0,0 +1,158 @@
...
+ */
+async function navigateToDappList() {
+  await device.disableSynchronization()
+  await waitForElementId('Hamburger')
+  await element(by.id('Hamburger')).tap()
+  await waitForElementId('dapps-explorer-icon')
+  await element(by.id('dapps-explorer-icon')).tap()
+  await waitFor(element(by.id('DAppExplorerScreen/loading')))
+    .not.toBeVisible()
+    .withTimeout(30 * 1000)
```
kathaypacific, line 19: i think this is racy because the loader is not guaranteed to be there when the screen first loads, i see in the last test there is an assertion like `await waitFor(element(by.id('DAppExplorerScreen/DappsList')))
      .toExist()
      .withTimeout(10 * 1000)` to see that the dapps list exists, would that be a better assertion here too?

e2e/src/usecases/DappListRecent.js
```diff
@@ -0,0 +1,158 @@
...
+}
+
+export default DappListRecent = () => {
+  it('should show most recently used dapp leftmost on home screen :ios:', async () => {
+    // Get recently used dapps at start of test
+    const startRecentlyUsedDapps = await getElementTextList('RecentDapp-name')
+    const startRecentDappCount = startRecentlyUsedDapps.length
+
+    // If no recently used dapps check that RecentDapp element does not exist
+    // Else check that All Dapps button exists
```
kathaypacific, line 55: i was confused for a sec about why the test would need conditions (like, why isn't it deterministic?) but then i remembered you said that you wanted this test to be able to rerun from any state left by previous tests. could you also mention this in the comment so that we don't forget?

kathaypacific (commented):
This is looking fab! sorry i had a few more small comments

e2e/src/usecases/DappListRecent.js
MuckT: Aadjusted and removed the sleep.
```
  await waitFor(element(by.id('DAppExplorerScreen/DappsList')))
    .toExist()
    .withTimeout(10 * 1000)
```

e2e/src/usecases/DappListRecent.js
MuckT: Added!

src/home/DappsCarousel.tsx
MuckT: Changed in 2578442707407aa15bbeba6e1e55b3b213dae616!

e2e/src/DappList.spec.js
MuckT: Added a comment in 1bcfb9348276381412d0cbc345d7cf59116b3c44, but this is still needed for the tests to pass on the first run.

e2e/src/usecases/DappListRecent.js
MuckT: Commented in https://github.com/valora-inc/wallet/pull/3287#discussion_r1062786920!

e2e/src/usecases/DappListRecent.js
MuckT: Added a comment in d30b567a9d4522ff412edde8e618d5b1f284f183!

PR #3291, opened 2022-12-22
fix: exchange rate display on swap review screen when to token is set last

MuckT:
### Description

Fixes an issue where the exchange rate on the swap review page would show incorrectly if the To token was set last.

### Test plan

Tested locally on iOS and unit tests added.

### Related issues

N/A

### Backwards compatibility

Yes


src/swap/SwapReviewScreen.test.tsx
```diff
@@ -152,25 +154,80 @@ describe('SwapReviewScreen', () => {
...
+    }
+
+    const newMockStore = createMockStore(newStore)
+
+    mockFetch.mockResponse(
+      JSON.stringify({
+        unvalidatedSwapTransaction: {
+          sellToken: mockCusdAddress,
+          buyToken: mockCeloAddress,
+          buyAmount: '1000000000000000000',
```
jeanregisser, line 190: I think we should use a different amount here to make sure the price is showing as a unit.
And is not inadvertently reusing the buyAmount.
Same for the other examples.

src/swap/SwapReviewScreen.test.tsx
```diff
@@ -152,25 +154,80 @@ describe('SwapReviewScreen', () => {
...
+    )
+
+    await waitFor(() => {
+      // Swap From
+      expect(getByTestId('FromSwapAmountToken')).toHaveTextContent('3.10 cUSD')
+      expect(getByTestId('FromSwapAmountTokenLocal')).toHaveTextContent('$3.10')
+      // Swap To
+      expect(getByTestId('ToSwapAmountToken')).toHaveTextContent('1.00 CELO')
+      // Exchange Rate
+      expect(getByTestId('ExchangeRate')).toHaveTextContent('3.10 cUSD ≈ 1 CELO')
```
jeanregisser, line 212: I wonder what's best to display for our users? 
`3.10 cUSD ≈ 1 CELO` or `1 cUSD ≈ 0.32 CELO`

jeanregisser (approved):
Great! 👍

src/swap/SwapReviewScreen.test.tsx
MuckT: Adjusted the values for the `should display correct exchange rate when buyAmount is used` test in https://github.com/valora-inc/wallet/pull/3291/commits/10fa405b099e6b47aa87169f751558a1f03b0b05!

src/swap/SwapReviewScreen.test.tsx
MuckT: `3.10 cUSD ≈ 1 CELO` was the easiest with our current implementation so I went with that.

PR #3292, opened 2022-12-23
fix(analytics): device id in super props

satish-ravi:
### Description

`getUniqueId` returns a promise in react-native-device-info@v10, which was upgraded on https://github.com/valora-inc/wallet/pull/2909. This was resulting in the data including a serialized promise, hence the `s_device_id_40`, `s_device_id_55`, etc columns. 
This PR consolidates getting all device info from a single place and adds type checking.

### Test plan

Unit tests, locally by logging analytics props

### Related issues

- Fixes ACT-574

### Backwards compatibility

Yes


silasbw (approved):

src/analytics/ValoraAnalytics.ts
```diff
@@ -25,7 +25,36 @@ import { Statsig } from 'statsig-react-native'
 
 const TAG = 'ValoraAnalytics'
 
-async function getDeviceInfo() {
+interface DeviceInfoType {
```
kathaypacific, line 28: ah, how annoying that there's no type definition from the lib itself 🙈

kathaypacific (approved):
✨

src/analytics/selectors.ts
```diff
@@ -90,10 +89,6 @@ export const getCurrentUserTraits = createSelector(
       localCurrencyCode,
       hasVerifiedNumber,
       hasCompletedBackup,
-      deviceId: DeviceInfo.getUniqueId(),
-      appVersion: DeviceInfo.getVersion(),
-      appBuildNumber: DeviceInfo.getBuildNumber(),
-      appBundleId: DeviceInfo.getBundleId(),
```
jeanregisser, line 96: I think we should have kept these.

kathaypacific: how come? is it because of the saga that calls the `identify` method on the analytics when these things change? i thought that these variables were unlikely to change during a session so seemed like a safe change?

jeanregisser: I commented here: https://github.com/valora-inc/wallet/pull/3294

src/analytics/ValoraAnalytics.ts
cajubelt: there is actually

src/analytics/ValoraAnalytics.ts
```diff
@@ -71,7 +100,7 @@ const SEGMENT_OPTIONS: analytics.Configuration = {
 
 class ValoraAnalytics {
   sessionId: string = ''
-  deviceInfo: object = {}
```
cajubelt, line 74: @kathaypacific this was our type before. it allowed a promise as a value with no compile time error.

cajubelt: so the fact that the library was typed did not save us because we were just dumping a mix of promises and strings here

src/analytics/selectors.ts
cajubelt: yeah it's annoying that they're in every track event in BQ twice, but that doesn't mean it's safe to remove them from one of the places yet imo

cajubelt: also if we're going to remove them i'm not sure i'd do it from here

PR #3294, opened 2022-12-23
fix(analytics): restore existing logic for user traits

jeanregisser:
### Description

Follow up for #3292, which was changing the behavior for the  `identify` call.
It's hard to tell if it would have had a real impact, but for instance the removed properties would have been missing from the CleverTap user profiles. 
Which we could argue is probably fine for `appVersion` as the CleverTap lib already adds this internally.
But just to be on the safe side with this, I restored the previous behavior.

### Test plan

- Tests updated 
- Added types to `getCurrentUserTraits` to ensure changes to `react-native-device-info` won't bring back the initial bug fixed by #3292.

### Related issues

N/A

### Backwards compatibility

Yes


kathaypacific (approved):
ah, thanks! i missed that the updateUserTraits saga is the _only_ place that calls the `identify` method. i wrongly thought that identify was already called on app start with all the super user traits, and that this saga was only updating on changes. 

on a side note, i'm very confused about why react-native-device-info changed the existing `getUniqueId` method to return a promise, when they could still return this directly in a string with `getUniqueIdSync` :/ seems like an unnecessary breaking change. i tried to see if there was a difference between the implementations but couldn't understand 🙈

jeanregisser:
Yes the reason is this: https://github.com/react-native-device-info/react-native-device-info/commit/449e1a8579d057e27aa331963204e595a5179a17

But there are more things to change if we don't want to be accessing it before consent.

PR #3295, opened 2022-12-23
feat: add favorited dapps to the home screen

kathaypacific:
### Description

As the title. When there are favorited dapps, display those instead of the recently used dapps.

All analytics events are shared, the only difference is the `section` property on the events.

Not included in this PR: UI changes to the carousel items. Discussing with Reed some details there.

### Test plan

https://user-images.githubusercontent.com/20150449/209326854-e97ce136-7305-4064-b183-9155bf3eee1b.mp4

### Related issues

- Fixes RET-521

### Backwards compatibility

Yes


jeanregisser (approved):
🚀

PR #3296, opened 2022-12-23
feat(celo-news): add analytics event when the feed is mounted

jeanregisser:
### Description

Since the Celo News feed is currently displayed in the existing `ExchangeHomeScreen` screen, it would have been more difficult to differentiate users having been exposed to it vs the existing CELO transaction feed.
Hence the dedicated event.

### Test plan

- Updated tests

### Related issues

- Fixes RET-519

### Backwards compatibility

Yes


kathaypacific (approved):
🚀

PR #3298, opened 2022-12-23
feat(celo-news): skip Celo education carousel when celo news is enabled

jeanregisser:
### Description

From the [experiment design](https://www.notion.so/valora-inc/Celo-news-feed-43e09037425949b38f2fdc79812a62e3) we learned users tapping the CELO menu item drop at the Celo education carousel.
So when Celo News is enabled, we now skip it too.

Note: I slightly refactored the menu item thing so we don't have to repeat ourselves.

### Test plan

- When Celo News is enabled, tapping the CELO menu item should directly show the news feed, without the Celo education carousel.

### Related issues

Discussed on [Slack](https://valora-app.slack.com/archives/C0494B52JTH/p1671805415760139).

### Backwards compatibility

Yes


kathaypacific (approved):
🚀

PR #3299, opened 2022-12-25
fix(statsig): use wallet address instead of account address

satish-ravi:
### Description

Use wallet address instead of account address for statsig user

### Test plan

Unit tests

### Related issues

N/A

### Backwards compatibility

Yes


kathaypacific (approved):
The changes look fine to me, on the PR description we say it's backwards compatible but once this change is rolled out we wont be able to carry over the userID from previous sessions / app versions for some users? idk if this is an issue, just wanted to check

PR #3300, opened 2022-12-28
feat: enable restoring wallet with 12 word seed phrase (ETH derivation path)

kathaypacific:
### Description

In the restore flow:
- if a user enters a 24 word recovery phrase, restore the wallet using the CELO derivation path
- for all other recovery phrase lengths, restore the wallet using the ETH derivation path
- update the allowed mnemonic lengths to be 12 and 24

### Test plan

For the following groups of users, they should be able to still transact and use their wallets as usual. I tested these manually.
1. Existing user who receives this PR as an app update
2. Restoring user using 12 word seed phrase (e.g. Metamask wallet)
3. Restoring user using 24 word seed phrase (Valora user)
4. New user going through the onboarding flow

### Related issues

- Fixes RET-539

### Backwards compatibility

Yes


src/backup/utils.ts
```diff
@@ -95,16 +96,17 @@ export function formatBackupPhraseOnEdit(phrase: string) {
   return phrase.replace(/\s+/gm, ' ')
 }
 
-export function formatBackupPhraseOnSubmit(phrase: string) {
```
kathaypacific, line 98: i removed this abstraction, didn't seem to be adding much value

src/backup/utils.ts
```diff
@@ -95,16 +96,17 @@ export function formatBackupPhraseOnEdit(phrase: string) {
...
 }
 
-function isValidMnemonic(phrase: string, length: number) {
-  return !!phrase && countMnemonicWords(formatBackupPhraseOnEdit(phrase)) === length
-}
+export function getDerivationPath(phrase: string) {
+  const wordCount = countMnemonicWords(phrase)
+  const derivationPath = wordCount === 24 ? CELO_DERIVATION_PATH_BASE : ETHEREUM_DERIVATION_PATH
 
-export function isValidBackupPhrase(phrase: string) {
```
kathaypacific, line 106: also combined isValidMnemonic and isValidBackupPhrase, seemed to do the same thing

src/backup/utils.ts
```diff
@@ -14,6 +14,7 @@ import { currentAccountSelector } from 'src/web3/selectors'
 const TAG = 'Backup/utils'
 
 const MNEMONIC_STORAGE_KEY = 'mnemonic'
+const ETHEREUM_DERIVATION_PATH = "m/44'/60'/0'"
```
kathaypacific, line 17: happy to move this to a config file but it's only used here, i see it declared as a variable like this in the celo monorepo too so didn't think too much further on where to place this

jeanregisser (approved):
Yay 🚀

PR #3302, opened 2022-12-28
test: add e2e tests for restoring with 12 word recovery phrase

kathaypacific:
### Description

As the title

### Test plan

n/a
### Related issues

- Fixes RET-539

### Backwards compatibility

Yes


e2e/src/usecases/RestoreAccountOnboarding.js
```diff
@@ -1,90 +1,90 @@
...
-  //   await element(by.id('ChooseLanguage/en-US')).tap()
-  // })
-  it('Onboarding Education', async () => {
-    // Onboarding education has 3 steps
-    for (let i = 0; i < 3; i++) {
-      await element(by.id('Education/progressButton')).tap()
-    }
   })
 
-  it('Welcome', async () => {
```
kathaypacific, line 26: i removed these very short test blocks, i'm not sure that they added much value. this let me do the `it.each` more easily on the top level for 12 / 24 word tests, rather than having multiple levels of `describe`s

e2e/src/usecases/RestoreAccountOnboarding.js
```diff
@@ -1,90 +1,90 @@
...
-    // confirmation popup skip
-    await element(by.id('PhoneVerificationSkipDialog/PrimaryAction')).tap()
-  })
+      // verify that we land on the home screen
+      await expect(element(by.id('SendOrRequestBar'))).toBeVisible()
 
-  it('Wallet Home', async () => {
-    await expect(element(by.id('SendOrRequestBar'))).toBeVisible()
-  })
+      // verify that the correct account was restored
```
kathaypacific, line 81: i updated the assertion to check for the account address rather than checking for wallet home, seems more accurate (although we don't test using this wallet, seeing the correct wallet address restored is a good start)

e2e/src/usecases/RestoreAccountOnboarding.js
jeanregisser: Looks fine, we can also use `describe.each`, or good old loops, to mostly keep the existing structure.
I'll let @MuckT chime in too.

jeanregisser (approved):
👍

e2e/src/usecases/RestoreAccountOnboarding.js
kathaypacific: i clicked into a few other test files and didn't see any other tests that were broken up so finely (even in the NewAccountOnboarding one) so i thought this file might just be old / outdated from the new patterns. i'd be curious if there is benefit in breaking up tests very finely, i guess the error messages will look different but i think it's still clear where the test failed in either implementation

kathaypacific: i caught up with @MuckT yesterday and seems like he is fine with removing the existing structure, this is an older pattern that we want to move away from and he mentioned that removing the fine grain tests is more "end to end" here :)

PR #3303, opened 2022-12-28
chore: remove escrow and invite experiment code

kathaypacific:
### Description

Remove:
- `inviteMethod` experiment variable and redundant code paths (with REFERRAL_URL as the chosen variant) + update redux migrations
- remove code paths related to transferring to escrow
- remove unused invite modals
- rename variables that had "orInvite" in them (many were called things like `sendPaymentOrInviteXXX`)

Did not touch:
- logic for displaying escrow claims on the home cards and tx feed, for backwards compatibility

### Test plan

n/a

### Related issues

- Fixes RET-514

### Backwards compatibility

Yes


src/send/SendConfirmationLegacy.tsx
```diff
@@ -475,17 +446,20 @@ function SendConfirmationLegacy(props: Props) {
...
-        recipientAddress,
-        amount: amount.valueOf(),
-        currency,
-        balance: balance.toString(),
-        includeDekFee: !isDekRegistered,
-      }
-    : { feeType: FeeType.INVITE, account, amount, currency, balance: balance.toString() }
+
+  if (!recipientAddress) {
+    throw Error('Recipient address is required')
```
kathaypacific, line 451: this is one part that i was iffy about :/ there are 3 entry points to the SendConfirmationLegacy screen: 
1) scanning QR code if the tx is legacy - really unsure which real user flow this is?
2) IncomingPaymentRequestListItem - this seems to be the related to paying out a payment request, i think it's not possible for the recipient not to have an address if they've sent a payment request so i'm fairly sure this is safe
3) ValidateRecipientAccount - this seems to be part of the secure send flow, again i think it's not possible to reach this stage in the user journey and have no address on the recipient object.
would love any feedback on whether removing the InviteAndSendModal is a bad idea

src/send/SendAmount/index.tsx
```diff
@@ -117,21 +106,15 @@ function SendAmount(props: Props) {
...
   const showInputInLocalAmount = usingLocalAmount && tokenHasUsdPrice
 
-  const localCurrencySymbol = useSelector(getLocalCurrencySymbol)
   const recipientVerificationStatus = useRecipientVerificationStatus(recipient)
-  const feeType =
-    recipientVerificationStatus === RecipientVerificationStatus.VERIFIED
-      ? FeeType.SEND
-      : FeeType.INVITE
-  const shouldFetchNewFee =
-    recipientVerificationStatus !== RecipientVerificationStatus.UNKNOWN && !isOutgoingPaymentRequest
```
jeanregisser, line 134: it seems this wasn't only for the escrow case and we should keep the existing logic, right?

src/send/saga.ts
```diff
@@ -388,15 +387,6 @@ export function* sendPaymentOrInviteSagaLegacy({
...
       yield call(sendPaymentLegacy, recipientAddress, amount, comment, currency, feeInfo)
-    } else if (recipientHasNumber(recipient)) {
-      yield call(
-        sendInvite,
-        recipient.e164PhoneNumber,
-        amount,
-        tokenInfo.usdPrice ? amount.multipliedBy(tokenInfo.usdPrice) : null,
-        tokenInfo.address,
-        feeInfo
-      )
```
jeanregisser, line 399: What do you think of throwing here if we don't have an address?

src/send/saga.ts
```diff
@@ -441,8 +431,6 @@ export function* sendPaymentOrInviteSaga({
           amount: amount.toString(),
         })
       }
-    } else if (recipientHasNumber(recipient)) {
-      yield call(sendInvite, recipient.e164PhoneNumber, amount, usdAmount, tokenAddress, feeInfo)
     } else {
       throw new Error('')
```
jeanregisser, line 435: Interesting we didn't provide a message here 🤔 
We should probably add one.

jeanregisser (commented):
Great to see this removed! 🚀

src/send/SendConfirmationLegacy.tsx
jeanregisser: 1. this is for secure send too, confirming the recipient by scanning a QR code
2. yes what's a bit confusing is payment requests can use the secure send flow. I don't recall the details about this. From reading the code it seems this is checked when the phone number is present in the payment request, to ensure it goes to the expected address, even though the requester address is there too.
3. yep this is what I understand too.

In brief, it seems safe to remove it 👍

src/send/SendAmount/useTransactionCallbacks.ts
```diff
@@ -173,9 +169,6 @@ function useTransactionCallbacks({
         isOutgoingPaymentRequest: true,
         origin,
       })
-    } else if (recipientVerificationStatus !== RecipientVerificationStatus.VERIFIED) {
-      ValoraAnalytics.track(RequestEvents.request_unavailable, continueAnalyticsParams)
-      navigate(Screens.PaymentRequestUnavailable, { transactionData })
```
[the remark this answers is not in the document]
jeanregisser: Makes sense, thanks for double checking.

src/send/SendAmount/index.tsx
kathaypacific: i think this is the same as above, a user should only land here if they have the verified status

src/send/saga.ts
kathaypacific: added!

jeanregisser (approved):

PR #3304, opened 2022-12-28
chore(mtwAddress): remove accountAddress selector to prevent bugs

cajubelt:
### Description

Remove accountAddressSelector to prevent bugs like the one Jean fixed [here](https://github.com/valora-inc/wallet/pull/3294/files#diff-8d33cbb0da323fba8e3986c827ba5c2d362c018756e78445f3b6c0606f413780R137). 

The name of the accountAddressSelector introduces high risk of mistaking it for a wallet address selector, but in reality, it selects MTW address and uses wallet address as a fallback. We have largely discontinued use of the MTW address, so this selector should not be needed anymore. 

### Test plan

added unit tests

### Related issues

na

### Backwards compatibility

should be backwards compatible. assumed in 1 place that wallet address is always non-null if MTW address is non-null-- see code comment in `app/selectors.test.ts` for details

### DIFF
--- a/src/analytics/selectors.ts
+++ b/src/analytics/selectors.ts
@@ -12,11 +12,11 @@ import { getLocalCurrencyCode } from 'src/localCurrency/selectors'
 import { userLocationDataSelector } from 'src/networkInfo/selectors'
 import { coreTokensSelector, tokensWithTokenBalanceSelector } from 'src/tokens/selectors'
 import { sortByUsdBalance } from 'src/tokens/utils'
-import { mtwAddressSelector, walletAddressSelector } from 'src/web3/selectors'
+import { mtwAddressSelector, rawWalletAddressSelector } from 'src/web3/selectors'
 
 export const getCurrentUserTraits = createSelector(
   [
-    walletAddressSelector,
+    rawWalletAddressSelector,
     mtwAddressSelector,
     defaultCountryCodeSelector,
     userLocationDataSelector,
@@ -30,7 +30,7 @@ export const getCurrentUserTraits = createSelector(
     superchargeInfoSelector,
   ],
   (
-    walletAddress,
+    rawWalletAddress,
     mtwAddress,
     phoneCountryCallingCode,
     { countryCodeAlpha2 },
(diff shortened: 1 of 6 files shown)


src/app/selectors.ts
```diff
@@ -78,7 +78,7 @@ export const superchargeTokenConfigByTokenSelector = (state: RootState) =>
   state.app.superchargeTokenConfigByToken
 
 export const rewardsEnabledSelector = createSelector(
-  [accountAddressSelector, superchargeTokenConfigByTokenSelector],
+  [walletAddressSelector, superchargeTokenConfigByTokenSelector],
```
cajubelt, line 81: since `accountAddressSelector` was just `mtwAddress ?? walletAddress`, this is only different if `mtwAddress` is non-null but `walletAddress` is null, which should never happen

satish-ravi: if `mtwAddress` is not null, this value would be `mtwAddress` in the existing code regardless of what `walletAddress` is right? Isn't this changing behavior for user's with `mtwAddress`?

cajubelt: if I understand correctly, rewardsEnabledSelector uses the outputs of accountAddressSelector and superchargeTokenConfigByTokenSelector to compute a boolean. My argument is that in all cases we care about, the output of rewardsEnabledSelector will be the same by just using walletAddressSelector in its place here

satish-ravi: ah, I didn't read the actual implementation of the selector 😁. This sounds good to me.

satish-ravi (approved):

src/analytics/selectors.ts
```diff
@@ -59,8 +59,8 @@ export const getCurrentUserTraits = createSelector(
     // Don't rename these unless you have a really good reason!
     // They are used in users analytics profiles + super properties
     return {
-      accountAddress,
-      walletAddress,
+      accountAddress: mtwAddress ?? rawWalletAddress,
+      walletAddress: rawWalletAddress?.toLowerCase(),
```
cajubelt, line 63: This is necessary because accountAddressSelector did not lowercase the wallet address when taking it from redux state, but walletAddressSelector does. We've been lower-casing wallet address before saving it to redux state since 11/2019, but for older users they might not be lowercase. (There's no redux migration in place for those users!) So I did this to be safe...

satish-ravi (approved):

jeanregisser (commented):
💯 🙏

PR #3307, opened 2022-12-29
chore: remove CPV experiment code and decentralised verification screens

kathaypacific:
### Description

This PR includes:
- remove CPV remote config variable, update redux migrations
- remove directly related code paths where CPV was false

This PR is already massive, so I draw the line at "directly related code paths" for the CPV flag. This PR removes display components but does not remove all komenci related code, there will be follow up PRs for that.

### Test plan

n/a

### Related issues

- Fixes RET-504

### Backwards compatibility

Yes


src/app/selectors.ts
```diff
@@ -36,31 +27,6 @@ export const sessionIdSelector = (state: RootState) => {
...
-  }
-
-  const e164Number = e164NumberSelector(state)
-  const saltCache = e164NumberToSaltSelector(state)
-  const shouldUseKomenci = shouldUseKomenciSelector(state)
-  const { komenci } = verificationStatusSelector(state)
-  const hideVerification = hideVerificationSelector(state)
-  if (hideVerification) {
-    return false
-  }
```
jeanregisser, line 52: Shouldn't we keep this?

src/account/Settings.tsx
```diff
@@ -416,7 +413,7 @@ export class Account extends React.Component<Props, State> {
               title={t('editProfile')}
               onPress={this.goToProfile}
             />
-            {!numberVerified && verificationPossible && (
```
jeanregisser, line 419: Looks like we may want to keep this for when `hideVerification` is true.

src/home/NotificationBox.tsx
```diff
@@ -244,7 +239,7 @@ function useSimpleActions() {
     }
   }
 
-  if (!dismissedGetVerified && !numberVerified && verificationPossible) {
```
jeanregisser, line 247: ditto

src/onboarding/registration/NameAndPicture.tsx
```diff
@@ -247,7 +243,6 @@ function NameAndPicture({ navigation, route }: Props) {
           type={BtnTypes.ONBOARDING}
           disabled={!nameInput?.trim()}
           testID={'NameAndPictureContinueButton'}
-          showLoading={asyncKomenciReadiness.loading}
```
jeanregisser, line 250: Yay, less unnecessary wait for our users 👍

src/send/Send.tsx
```diff
@@ -165,7 +163,7 @@ function Send({ route }: Props) {
   }
 
   const renderListHeader = () => {
-    if (!numberVerified && verificationPossible && !skipContactsImport) {
```
jeanregisser, line 168: Same here, we may want to keep this.

jeanregisser (commented):
w00t w00t 💥 🚀 

Feels so good to see all this code removed.

src/app/selectors.ts
kathaypacific: there is an early return above where the rest of this logic isn't executed if CPV is enabled, so i think it is safe to remove?

jeanregisser: I guess the question is, do we want to still be able to use `hideVerification`? while leaving CPV enabled.

kathaypacific: hmm, i lean towards no. what do you think? 

the name is a little misleading, we're using this to disable the "start" button on the decentralised verification screen rather than hiding the verification step during onboarding. i struggle to think of a valid scenario where we'd want to remotely disable phone verification on the front end - we can stop accepting new mappings from the backend, so i'm thinking the only reason why we'd want separate control of the front end is if our backend service is severly broken / we want to hide errors on the front end... do you think we need to prepare for this kind of fallback?

jeanregisser: True, the name is bit misleading, at least for the onboarding part.
I think it's fine to remove it. We should also remove the flag then. But can be done in a separate PR of course :D

kathaypacific: cool, i've removed it!

src/account/Settings.tsx
kathaypacific: i think that we resolved this in the conversation above :)

src/home/NotificationBox.tsx
kathaypacific: and this one

src/send/Send.tsx
kathaypacific: same :D

jeanregisser (approved):
🎉

PR #3308, opened 2022-12-29
feat: update styling for dapps carousel on home screen

kathaypacific:
### Description

As the description, also checked in with Reed about the new shade of green added to the colors.

Simulator Screen Shot - iPhone 11 - 2022-12-29 at 17 39 56

### Test plan

n/a

### Related issues

- Fixes RET-521

### Backwards compatibility

Y


src/home/DappsCarousel.tsx
```diff
@@ -145,10 +139,9 @@ function DappsCarousel({ onSelectDapp }: Props) {
           <Touchable
             key={dapp.id}
             onPress={() => onSelectDapp({ ...dapp, openedFrom: section })}
-            style={styles.dappContainer}
             testID={`${testID}/Dapp`}
           >
-            <>
+            <Card style={styles.card} rounded={true} shadow={Shadow.SoftLight}>
```
jeanregisser, line 144: I think we should put the `Touchable` inside the card like we did for the dapp list: https://github.com/valora-inc/wallet/blob/5ce3b5c541ccdb688ea3165bbaec8b588c6e2338/src/dappsExplorer/DappCard.tsx#L57-L58

This way the ripple effect on Android is contained within it.

src/home/DappsCarousel.tsx
```diff
@@ -158,9 +151,20 @@ function DappsCarousel({ onSelectDapp }: Props) {
...
+        <Touchable onPress={onPressAllDapps}>
+          <Card style={styles.viewAllCard} rounded={true} shadow={Shadow.SoftLight}>
+            <View style={[styles.icon, styles.viewAllIcon]}>
+              <ArrowRight />
+            </View>
+            <Text style={styles.dappName} numberOfLines={1} ellipsizeMode="tail">
+              {t('allDapps')}
+            </Text>
+          </Card>
+        </Touchable>
```
jeanregisser, line 167: nit: maybe all items should be the same component with different data?

jeanregisser (commented):
Nice!

src/home/DappsCarousel.tsx
kathaypacific: to get this guy to print within the dapps loop, i'd have to make up some data to get this into the `Dapp` interface (many redundant properties) / make the `onPress` responsive to the dapp "type" / add a conditional styling on the image since this one has a green background - i could also export the image with the green background but it makes the icon less reusable. with the implementation here, i also could easily use a different test id here (i.e. none, lol) so the tests feel more correct. i did think about this suggestion before but i don't think it is as readable as this, what do you think?

src/home/DappsCarousel.tsx
kathaypacific: ah, thanks! i didn't know this, i've updated

src/home/DappsCarousel.tsx
jeanregisser: Yes that's fine, we just need to remember to update both places.
Then we need the Touchable inside the Card here too.

kathaypacific: oops 🙈

jeanregisser (approved):
👍 

Just double checking, do we need a test covering this?

kathaypacific:
> 👍
> 
> Just double checking, do we need a test covering this?

i don't think so - the functionality of rendering recently used / favourited dapps + navigating through the all dapps link is already covered by unit tests :)

PR #3311, opened 2022-12-30
fix: add fallback dapp name for WC bottom sheets

kathaypacific:
### Description

Context: https://valora-app.slack.com/archives/C02UAE4MGUF/p1672339362860479

This PR adds a fallback dapp name in case the wallet connect metadata is not complete from the dapp. In priority order, the displayed dapp name should be from: the WC payload -> the active dapp metadata (i.e. from the dapps list) -> the WC payload url (for externally connected dapps) -> empty string (status quo).

### Test plan

Tested manually

### Related issues

N/A

### Backwards compatibility

Y


src/walletConnect/screens/RequestContent.tsx
```diff
@@ -28,6 +30,30 @@ interface Props {
 
 const DAPP_IMAGE_SIZE = 60
 
+export const useDappMetadata = (metadata?: IClientMeta | CoreTypes.Metadata | null) => {
+  const activeDapp = useSelector(activeDappSelector)
```
jeanregisser, line 34: The `activeDapp` can be misleading in some cases:
- user reaching a different dapp from the initial dapp opened
- an unrelated WC requests coming while a user has opened a dapp

Even though these are arguably edge cases, I think we should aim for correctness, as it could have security implications.

What do you think of dropping this fallback on the activeDapp?

kathaypacific: yeah you're right, that's a valid edge case. i think that the `activeDapp` use case might be common though, and the url fallback is a little ugly since for the fallback image we use the first letter of the dapp name (in this case, probably "H" for "https://www.dapp.com" which seems confusing) 

Since the logic is abstracted away in a hook, it can be ugly - what do you think about the `activeDapp` condition being "if there is an active dapp AND the active dapp url domain matches the payload domain"?

jeanregisser: Yes we could check the domain matches.
I guess since we control the listed dapps we should require they provide a valid name and avoid that extra logic, up to you.

For the url fallback, we should probably use the domain instead of the the full url.

kathaypacific: yay thanks! i added the check for the domain and fallback to the domain.

we do implicitly check that dapps that we list has correct logic for WC since we test before merging the dappslist PR but it looks like things can become broken :/ i think it's better to have a proper fallback and when we do spot it then raise it to the dapps

src/walletConnect/screens/RequestContent.tsx
```diff
@@ -28,6 +30,36 @@ interface Props {
...
+    // should never happen
+    return {
+      url: '',
+      dappName: '',
+      dappImageUrl: '',
+    }
+  }
+
+  const { url, name, icons } = metadata
+  const dappHostname = new URL(url).hostname
```
jeanregisser, line 46: If the url can't be parsed, this will crash.
Maybe we should cover this with a unit test? :D

kathaypacific: thank you! i always forget this :( i added unit tests like you suggested

src/walletConnect/screens/RequestContent.tsx
```diff
@@ -28,6 +30,36 @@ interface Props {
...
+    // should never happen
+    return {
+      url: '',
+      dappName: '',
+      dappImageUrl: '',
+    }
+  }
+
+  const { url, name, icons } = metadata
+  const dappHostname = url ? new URL(url).hostname : ''
```
jeanregisser, line 46: Sorry this can still crash with `new URL('/en-US/docs')` for instance.
We have to use a try/catch block.

kathaypacific: ah, third time's the charm...i read the docs and updated an existing hook that uses this logic

jeanregisser (approved):
🚀

PR #3323, opened 2023-01-05
fix: incorrectly navigating back after entering PIN on iOS

jeanregisser:
### Description

Fix for RET-559, causing a temporary navigate back action after entering the PIN on iOS.

I can't find anymore why I needed to add this buggy workaround in the first place, when we switched to native-stack in https://github.com/valora-inc/wallet/pull/3004 🤷  

### Test plan

- Tested the reproduction steps
- Tested related flows where PIN is asked to make sure they behave correctly

### Related issues

- Fixes RET-559

### Backwards compatibility

Yes


jeanregisser:
Maybe @MuckT you remember the steps that caused the PIN screen to remain visible on iOS with the modal when we switched to native-stack?

kathaypacific (approved):
i _think_ this is fine, i had a quick test and LGTM. it'd be good for @MuckT to have a look in case he remembers what this workaround was for though. i vaguely recall the reset account flow causing issues with this native stack and thought maybe this workaround is related to that, but it seems like the reset account flow works the same with or without this workaround.

MuckT:
@jeanregisser and @kathaypacific this was added due to the reset flow; however, I am unable to reproduce the original error which is detailed below:
1. Restore account
2. Navigate to Settings
3. Tap Reset
4. Confirm on modal
5. Tap back on pin entry - blank screen is served.

MuckT (approved):
I am unable to reproduce the original error that this fixed. 🚀

PR #3329, opened 2023-01-05
fix(fiatconnect): update mobile money to match fc spec

anthonydaoud:
### Description

When adding a Mobile Money account, the wallet currently provides a dropdown showing a set of supported network operators by default. The fiatconnect specification was [updated](https://github.com/fiatconnect/specification/pull/88) such that we should no longer show this dropdown by default, and instead we should allow users to enter the name of their operator using their keyboard. This PR implements this functionality. This PR also updates the both the `fiatconnect-sdk` and `fiatconnect-types` packages to the latest version which no longer features the `SupportedOperatorEnum`

UpdatedMobileMoneyOperator

### Test plan

Manually verified adding a mobile money account behaves as expected.

### Related issues

- ACT-581

### Backwards compatibility

Yes.


src/fiatExchanges/quotes/constants.ts
```diff
@@ -1,20 +1,12 @@
...
     [Property in keyof Partial<FiatAccountSchemas[Schema]>]: string[]
   }
 }>
 
-export const DEFAULT_ALLOWED_VALUES: DefaultAllowedValues = {
-  [FiatAccountSchema.MobileMoney]: {
-    operator: Object.keys(SupportedOperatorEnum),
-  },
-}
+export const DEFAULT_ALLOWED_VALUES: DefaultAllowedValues = {}
```
cajubelt, line 9: seems like this isn't used anymore. please get rid of it.

anthonydaoud: Removed it.

cajubelt (approved):

PR #3330, opened 2023-01-06
feat!: new User-Agent header on iOS

jeanregisser:
### Description

PR for iOS to unify the `User-Agent` header for all network requests.

### Test plan

Tested the new `User-Agent` is used for all requests (fetch, Images, etc) using [Proxyman](https://proxyman.io/).

Examples:
Screenshot 2023-01-06 at 11 09 32
Screenshot 2023-01-06 at 11 13 57

### Related issues

- Fixes RET-558

### Backwards compatibility

No, services depending on the existing `User-Agent` need to be updated.

### DIFF
--- a/ios/celo/AppDelegate.m
+++ b/ios/celo/AppDelegate.m
@@ -13,6 +13,7 @@
 #import <React/RCTBundleURLProvider.h>
 #import <React/RCTRootView.h>
 #import <React/RCTLinkingManager.h>
+#import <React/RCTHTTPRequestHandler.h>
 
 @import Firebase;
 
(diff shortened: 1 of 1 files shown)


kathaypacific (approved):
LGTM! do we want this value to ever be "Celo", like how the app name is Celo if you don't decrypt the branding? i'm not sure why that would be useful, it was just a thought 😅

jeanregisser:
> LGTM! do we want this value to ever be "Celo", like how the app name is Celo if you don't decrypt the branding? i'm not sure why that would be useful, it was just a thought 😅

Yes I was wondering about that and decided to leave it always as Valora to avoid issues.
Since some services use this header to tailor some responses (dapplist, etc), I thought it was better to have it fixed.

PR #3331, opened 2023-01-06
feat!: new User-Agent header on Android

jeanregisser:
### Description

PR for Android to unify the `User-Agent` header for all network requests.

### Test plan

Tested the new `User-Agent` is used for all requests (fetch, Images, etc) using [Proxyman](https://proxyman.io/).

Examples:
- forno: Screenshot 2023-01-06 at 12 34 07
- an image: Screenshot 2023-01-06 at 12 43 31

Browsing in our WebView keeps the default User-Agent, as expected, here with Bidali:

Screenshot 2023-01-06 at 12 38 13

### Related issues

- Fixes RET-558

### Backwards compatibility

No, services depending on the existing `User-Agent` need to be updated.


kathaypacific (approved):
🚀

PR #3332, opened 2023-01-06
fix: android restore wallet flow navigation stall

kathaypacific:
### Description

There are 2 problems this PR solves:

#1: on android, sometimes the recovery phrase text box does not appear (i find this reproducible if i land on this screen with a recovery phrase already on my clipboard). Video of the issue:

https://user-images.githubusercontent.com/20150449/211043417-0ded2018-6cec-45b8-a77a-7d87f5598af3.mp4

This one is annoying, I am fairly sure it has something to do with the Layout animation of the "paste" button. The value of the clipboard content is always an empty string to start with (since fetching the clipboard content is async) so the `ClipboardAwarePasteButton` component always needs to do the layout animation twice as `shouldShow` changes from `false` initially to `true` when the clipboard content is picked up. The issue does not exist if this `shouldShow` prop doesn't change value between renders, i.e. if there is no recovery phrase on the clipboard). I couldn't find a way to fix this, so i just disabled this animation since it doesn't seem like a critical feature....

#2: the navigation seems to be broken after the recovery phrase is entered. on force close then restart, the app does function normally. video below:

https://user-images.githubusercontent.com/20150449/211043522-4aee3914-f136-46ff-ae6e-4ed78acfc6ac.mp4

Since we've disabled the swipe to navigate back on iOS and we have control over the hardward back button on android, there's no more need to adjust the navigation stack. The change in this PR removes the navigation stack logic and makes the back button on android (and the cancel button on the header) go to the welcome screen. Video after the changes in this PR:

https://user-images.githubusercontent.com/20150449/211043464-107125df-ce3d-42bb-9d6e-b16d2c32b495.mp4

### Test plan

Test on iOS and Android:
- restore flow
- navigating back on restore wallet screen
- create flow

### Related issues

- Fixes RET-562

### Backwards compatibility

Y


src/components/CodeInput.tsx
```diff
@@ -61,6 +61,12 @@ export default function CodeInput({
 }: Props) {
   const [forceShowingPasteIcon, clipboardContent, getFreshClipboardContent] = useClipboard()
 
+  const clipboardRef = useRef(clipboardContent)
```
kathaypacific, line 64: i need to rework this part, the ref resolves the main issue but does lag one render cycle behind so the paste button doesn't appear right away :/ any ideas appreciated

src/navigator/NavigationService.ts
```diff
@@ -95,6 +96,12 @@ export function navigate<RouteName extends keyof StackParamList>(
     : [RouteName, StackParamList[RouteName]]
 ) {
   const [routeName, params] = args
+
+  if (routeName === Screens.ImportWallet) {
```
kathaypacific, line 100: i don't love this approach because it is a little unexpected / easy to miss. BUT, the previous implementation had this logic in the screen itself which for me was just as unexpected....with this approach at least the navigation functionality lives together and it's not possible to forget to clear the stack when navigating to this screen

src/navigator/NavigationService.ts
```diff
@@ -128,6 +135,43 @@ export const navigateClearingStack: SafeNavigate = (...args) => {
...
+// the Welcome screen, even if there were intermediary screens in between. It is
+// meant to be used during onboarding,
+export const navigateClearingOnboardingStack: SafeNavigate = (...args) => {
+  const [routeName, params] = args
+
+  ensureNavigator()
+    .then(() => {
+      Logger.debug(`${TAG}@navigateClearingOnboardingStack`, `Dispatch ${routeName}`)
+
+      const state = navigationRef.current?.getRootState()
```
kathaypacific, line 148: most of this logic is untouched from UseBackToWelcomeScreen

src/navigator/NavigationService.ts
jeanregisser: Yes though what this doesn't fully cover is the case where the app is launched and the initial screen is set to `ImportWallet`.

https://github.com/valora-inc/wallet/blob/2af220ad9b557963e35dda38f2a00fd3fb01351e/src/navigator/Navigator.tsx#L604-L607

That's why I had created the `UseBackToWelcomeScreen` and put the logic close to the screen.

That being said, without this edge case handling, I think the user is still able to proceed, just the animation is not right when tapping "cancel", after launching the app and landing directly on the import screen.

src/components/ClipboardAwarePasteButton.tsx
```diff
@@ -17,7 +17,9 @@ export default function ClipboardAwarePasteButton({
   const { t } = useTranslation()
 
   useLayoutEffect(() => {
-    LayoutAnimation.easeInEaseOut()
+    if (Platform.OS !== 'android') {
```
kathaypacific, line 20: i think this is our only option...reading the [LayoutAnimation docs](https://reactnative.dev/docs/0.66/layoutanimation) the first paragraph already calls out that `Note that in order to get this to work on Android you need to set the following flags via UIManager` but we've explicitly disabled this in the past https://github.com/valora-inc/wallet/blob/83630ce2878cde47431d966380e0c3c3f179a40a/src/app/App.tsx#L62-L67

src/import/ImportWallet.tsx
```diff
@@ -80,8 +78,11 @@ function ImportWallet({ navigation, route }: Props) {
       headerLeft: () => (
         <TopBarTextButtonOnboarding
           title={t('cancel')}
-          // Note: redux state reset is handled by UseBackToWelcomeScreen
-          onPress={() => navigate(Screens.Welcome)}
+          onPress={() => {
+            dispatch(cancelCreateOrRestoreAccount())
+            ValoraAnalytics.track(OnboardingEvents.restore_account_cancel)
+            navigate(Screens.Welcome)
```
jeanregisser, line 84: We need to make sure these actions are triggered when using the Android back button.
(swipe to go back is already prevented on iOS on this screen).

We can use this:

https://github.com/valora-inc/wallet/blob/2b28b691640c52bdeeb1b49a7751c5bdf3b764fe/src/utils/useBackHandler.ts#L4

jeanregisser (commented):
Thanks so much for taking a look at this. 
I bumped into this recently a couple of times using a real device and wanted to fix it too. This is definitely related to our switch to native-stack. @MuckT and I were discussing it while working on the switch, but somehow we missed it before merging.

src/components/ClipboardAwarePasteButton.tsx
jeanregisser: Yes we RN's builtin LayoutAnimation is quite buggy on Android and I've been burned a couple of times trying to re-enable it.

src/navigator/NavigationService.ts
kathaypacific: actually this case i did test, and i think it works okay still (i haven't tested on release mode though, but i hope it doesn't make a difference). to me the animation here still looks okay, what do you think? 

https://user-images.githubusercontent.com/20150449/211284317-d2172170-ba34-45d0-9170-f4e9b4a39939.mp4

src/components/ClipboardAwarePasteButton.tsx
```diff
@@ -17,7 +17,9 @@ export default function ClipboardAwarePasteButton({
...
+      LayoutAnimation.easeInEaseOut()
+    }
```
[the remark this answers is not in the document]
kathaypacific: yeah i'm not really sure what's going on here, but maybe `setLayoutAnimationEnabledExperimental` means "enable it to work kind of well" instead of "enable it to work at all"? idk but for me turning off the animation makes all the difference

src/navigator/NavigationService.ts
kathaypacific: clicking on the back button takes you out of the app though, maybe this is the issue you were referring to? i can play around with the `useBackHandler` hook for this i think

src/import/ImportWallet.tsx
kathaypacific: added! 

actually as i was adding this i was thinking, do we need any special navigation at all in this case? the only way that a user can navigate back is by the "cancel" button or with the android hardware back button. in both those places, we have control to clear the navigation stack and go straight back to the Welcome screen. so do we need to worry about clearing the navigation stack before navigating to (or, before this change, when we land on) this screen....?

kathaypacific: i might be missing something, but the only benefit i see for having something like `navigateClearingOnboardingStack` is that you're able to navigate back to the onboarding carousels from the Welcome screen. i honestly don't think this is valuable functionality though.

src/navigator/NavigationService.ts
jeanregisser: Yes I think it's fine. The incorrect animation is more on iOS. But of course a very small detail.

src/import/ImportWallet.tsx
jeanregisser: Yes I think you're right, we can remove clearing the stack now that iOS "swipe to go back" is disabled on this screen.
This would make this whole thing simpler.

The initial intention was to use `UseBackToWelcomeScreen` on most onboarding steps, without having to remember to disabled iOS's  back gesture or the Android back handler, but with the different iterations it ended up being used only on the `ImportWallet` screen.

MuckT:
@kathaypacific It looks like there is an e2e test failing on Both iOS and Android - `FiatConnectTransferOut.spec.js` Looking at the artifacts it appears to fail when attempting to restore an empty account: this appears related to the modal / navigation.

Alfajores Seed for Testing

```
void yard vanish antique usage party tennis orange twelve fluid room jump myth rich hobby brisk siege couple cup hunt legend claw canyon letter
```

https://user-images.githubusercontent.com/26950305/211416552-f8196229-16c6-49e0-8e37-993c89379a44.mov

kathaypacific:
> 

Hey @MuckT thanks for flagging! honestly not really sure what happened there but i did remove the navigation service workaround mentioned here https://github.com/valora-inc/wallet/pull/3332#discussion_r1064492982, i can no longer repro the issue manually and the CI is happy again.

jeanregisser (approved):
🎉

PR #3333, opened 2023-01-06
test(e2e): fix getElementText utility method

MuckT:
### Description

Fixes reruns of the `should correctly navigate via all dapps card` test on iOS.
Fixes the `getElementText` utility method to correctly return text when multiple elements match.

### Test plan

Tests pass in CI and locally.

### Related issues

N/A

### Backwards compatibility

Yes


jeanregisser (approved):
LGTM 👍

PR #3338, opened 2023-01-09
refactor(tokens): Deprecate goldToken & stableTokens in favor of tokens infrastructure

jh2oman:
### Description

Some historical context: 

> Long long ago, in a company far far away there was an App called Valora with only CELO and cUSD for play. 

> Soon cEUR was added too! But this wouldn't do. It needed more, MORE, and thus multitoken support was born. 🌟 

When multitoken support got added there were still large chunks of the app using the old data sources for CELO, CUSD, and cEUR. With this PR, those all get deprecated in favor of using the new multitoken data from `src/tokens`.

This PR:
* Deprecates most of `src/goldToken`
  * moves `educationComplete` into `state.account` where its better suited
  * keeps a legacy send function around and names it 'legacy*' so we are extra sure
* Deprecates most of `src/stableToken`
  * keeps a legacy send function around and names it 'legacy*' so we are extra sure

Most of the deprecation is very straightforward and has a near 1:1 support.

### Test plan

Manually going through some of the send & cico flows to ensure they still work with the new data sources. Trusting our good e2e tests for much of it

### Related issues

Paves the way for https://linear.app/valora/issue/ACT-390/ramp-creal

### Backwards compatibility

Should be fully backwards compatible


src/account/reducer.ts
```diff
@@ -31,6 +31,7 @@ export interface State {
   accountToRecoverFromStoreWipe: string | undefined
   dismissedKeepSupercharging: boolean
   dismissedStartSupercharging: boolean
+  goldEducationCompleted: boolean
```
jh2oman, line 34: Moved `state.goldToken.educationCompleted` to `state.account.goldEducationCompleted` since it doesn't have anything to do with the goldToken data and it allows us to delete the `goldToken` state from redux for good

src/send/saga.ts
```diff
@@ -174,7 +174,7 @@ function* sendPaymentLegacy(
     switch (currency) {
       case Currency.Celo: {
         yield put(
-          transferGoldToken({
+          transferGoldTokenLegacy({
```
jh2oman, line 177: making it clear that these are legacy functions waiting on other cleanup to be removed

src/account/actions.ts
```diff
@@ -30,6 +30,7 @@ export enum Actions {
   DISMISS_KEEP_SUPERCHARGING = 'ACCOUNT/DISMISS_KEEP_SUPERCHARGING',
   DISMISS_START_SUPERCHARGING = 'ACCOUNT/DISMISS_START_SUPERCHARGING',
   SAVE_SIGNED_MESSAGE = 'ACCOUNT/SAVE_SIGNED_MESSAGE',
+  SET_GOLD_EDUCATION_COMPLETED = 'ACCOUNT/SET_GOLD_EDUCATION_COMPLETED',
```
jeanregisser, line 33: nit: Gold term is really old :D
```suggestion
  SET_CELO_EDUCATION_COMPLETED = 'ACCOUNT/SET_CELO_EDUCATION_COMPLETED',
```

src/account/actions.ts
```diff
@@ -153,6 +154,11 @@ export interface SaveSignedMessage {
   type: Actions.SAVE_SIGNED_MESSAGE
 }
 
+export interface SetGoldEducationCompletedAction {
+  type: Actions.SET_GOLD_EDUCATION_COMPLETED
+  goldEducationCompleted: boolean
+}
```
jeanregisser, line 160:
```suggestion
export interface SetCeloEducationCompletedAction {
  type: Actions.SET_CELO_EDUCATION_COMPLETED
  celoEducationCompleted: boolean
}
```

src/exchange/ExchangeHomeScreen.test.tsx
```diff
@@ -17,8 +16,22 @@ const exchangeRates: ExchangeRates = makeExchangeRates('0.11', '10')
...
+            balance: '10',
+          },
+          [mockCeloAddress]: {
+            ...mockTokenBalances[mockCeloAddress],
+            balance: '2',
+          },
+          [mockCeloAddress]: {
+            ...mockTokenBalances[mockCeloAddress],
+            balance: '2',
+          },
```
jeanregisser, line 32: It's set twice for Celo:
```suggestion
          [mockCeloAddress]: {
            ...mockTokenBalances[mockCeloAddress],
            balance: '2',
          },
```

src/identity/PhoneNumberLookupQuotaScreen.tsx
```diff
@@ -15,14 +14,15 @@ import { LOOKUP_GAS_FEE_ESTIMATE } from 'src/identity/privateHashing'
...
 import fontStyles from 'src/styles/fonts'
+import { useTokenInfoByCurrency } from 'src/tokens/hooks'
+import { Currency } from 'src/utils/currencies'
 
 type Props = NativeStackScreenProps<StackParamList, Screens.PhoneNumberLookupQuota>
 
 function PhoneNumberLookupQuotaScreen(props: Props) {
   const [isSending, setIsSending] = useState(false)
-  const userBalance = useSelector(cUsdBalanceSelector) ?? null
+  const userBalance = useTokenInfoByCurrency(Currency.Dollar)?.balance.toNumber() ?? null
```
jeanregisser, line 25: So we don't loose precision:
```suggestion
  const userBalance = useTokenInfoByCurrency(Currency.Dollar)?.balance.toString() ?? null
```

src/identity/privateHashing.ts
```diff
@@ -183,15 +183,16 @@ function* navigateToQuotaPurchaseScreen() {
     })
 
     const ownAddress: string = yield select(currentAccountSelector)
-    const userBalance = yield select(cUsdBalanceSelector)
+    const tokens: CurrencyTokens = yield select(tokensByCurrencySelector)
+    const userBalance = tokens[Currency.Dollar]?.balance.toNumber() ?? null
```
jeanregisser, line 187: ditto

src/web3/dataEncryptionKey.ts
```diff
@@ -188,13 +187,10 @@ export function* registerAccountDek() {
...
-    const stableBalance = yield select(cUsdBalanceSelector)
-    const celoBalance = yield select(celoTokenBalanceSelector)
-    if (
-      (stableBalance === null || stableBalance === '0') &&
-      (celoBalance === null || celoBalance === '0')
-    ) {
+    const tokens: CurrencyTokens = yield select(tokensByCurrencySelector)
+    const cusdBalance = tokens[Currency.Dollar]?.balance
+    const celoBalance = tokens[Currency.Celo]?.balance
+    if (!cusdBalance && !celoBalance) {
```
jeanregisser, line 193: Looks like we may need to keep the 0 check?

jeanregisser (commented):
w00t 🚀 
Thanks so much for refactoring this 💯!

jeanregisser (approved):
Amazing 💯 🎉

PR #3348, opened 2023-01-10
feat: use 12 word mnemonic and base eth derivation path

MuckT:
### Description

Adds a remote config `useTwelveWordMnemonic` boolean which when enabled will cause new wallets to be created using the Eth derivation path and a 12 word seed phrase.

The restore flow already supports restoring from a 12 word seed phrase using an Eth derivation path so any user who creates a wallet using either the 12 word Eth derivation path seed phrase or 24 word Celo derivation path will be able to effectively restore.

### Test plan

Tested locally on iOS.

### Related issues

- Fixes RET-540

### Backwards compatibility

Yes


MuckT:
@kathaypacific I kept the logic for the Remote config key `useTwelveWordSeedPhrase` in app state, but I am open to moving it to into the web3 section if that's more appropriate: something similar was done send state.

src/app/saga.ts
```diff
@@ -204,6 +204,7 @@ export interface RemoteConfigValues {
   networkTimeoutSeconds: number
   dappFavoritesEnabled: boolean
   celoNews: CeloNewsConfig
+  useTwelveWordSeedPhrase: boolean
```
jeanregisser, line 207: nit: to follow similar other properties we have
```suggestion
  twelveWordSeedPhraseEnabled: boolean
```

src/app/selectors.ts
```diff
@@ -206,3 +206,6 @@ export const inviterAddressSelector = (state: RootState) => state.app.inviterAdd
 export const networkTimeoutSecondsSelector = (state: RootState) => state.app.networkTimeoutSeconds
 
 export const celoNewsConfigSelector = (state: RootState) => state.app.celoNews
+
+export const useTwelveWordSeedPhraseSelector = (state: RootState) =>
+  state.app.useTwelveWordSeedPhrase
```
jeanregisser, line 211: so it doesn't look like a hook:
```suggestion
export const twelveWordSeedPhraseEnabledSelector = (state: RootState) =>
  state.app.twelveWordSeedPhraseEnabled
```

src/web3/saga.ts
```diff
@@ -84,7 +92,19 @@ export function* getOrCreateAccount() {
...
+
+    const keys = yield call(
+      generateKeys,
+      mnemonic,
+      undefined,
+      undefined,
+      undefined,
+      bip39,
+      DERIVATION_PATH
+    )
```
jeanregisser, line 107: Can we add tests for `getOrCreateAccount` for the 12 words case?

src/app/reducers.ts
```diff
@@ -55,6 +55,7 @@ export interface State {
   inviterAddress: string | null
   networkTimeoutSeconds: number
   celoNews: CeloNewsConfig
+  useTwelveWordSeedPhrase: boolean
```
jeanregisser, line 58: nit: maybe this could live in the web3 module we have?

jeanregisser (commented):
🚀 🚀 🚀

src/web3/saga.ts
```diff
@@ -84,7 +92,19 @@ export function* getOrCreateAccount() {
       throw new Error('Failed to generate mnemonic')
     }
 
-    const keys = yield call(generateKeys, mnemonic, undefined, undefined, undefined, bip39)
+    const DERIVATION_PATH = useTwelveWordSeedPhrase
+      ? ETHEREUM_DERIVATION_PATH
+      : CELO_DERIVATION_PATH_BASE
+
+    const keys = yield call(
```
kathaypacific, line 99: could we use `generateKeysFromMnemonic` instead? that would remove the need to specify the derivation path here

src/firebase/remoteConfigValuesDefaults.e2e.ts
```diff
@@ -64,4 +64,5 @@ export const REMOTE_CONFIG_VALUES_DEFAULTS: Omit<
   networkTimeoutSeconds: 30,
   dappFavoritesEnabled: false,
   celoNews: JSON.stringify({ enabled: false } as RemoteConfigValues['celoNews']),
+  useTwelveWordSeedPhrase: false,
```
kathaypacific, line 67: ah, i'm conflicted about what this value should be 😅 it's a shame we can only test one flow, but do you think we should set this to true and prioritise the 12 word seed phrase?

kathaypacific (commented):
🥳 wow nice!!

src/firebase/remoteConfigValuesDefaults.e2e.ts
MuckT: As long as we are sure that the restore flow works for both either create flow seems fine for me.

src/web3/saga.ts
MuckT: Done in 67c68573612d31507eca1d95de789002ea28ea2f!

src/app/saga.ts
MuckT: Done in b1a5a1d4d9a4e90b601b5797f8e6c32cd78b30c9!

src/app/selectors.ts
MuckT: Done in b1a5a1d4d9a4e90b601b5797f8e6c32cd78b30c9!

src/web3/saga.ts
```diff
@@ -61,6 +60,10 @@ export function* getOrCreateAccount() {
   try {
     Logger.debug(TAG + '@getOrCreateAccount', 'Creating a new account')
 
+    const twelveWordSeedPhraseEnabled = yield select(twelveWordSeedPhraseEnabledSelector)
+    const MNEMONIC_BIT_LENGTH = twelveWordSeedPhraseEnabled
```
kathaypacific, line 64: super nit: maybe just camel case for this?

kathaypacific (approved):
Looks great!! 🚀

i kind of lean towards using the 12 word seed phrase in the CI, that seems like the flow we should prioritise not breaking going forward? we could also wait to update this after this change is fully rolled out :)

src/web3/saga.test.ts
```diff
@@ -51,31 +51,33 @@ describe(getOrCreateAccount, () => {
...
-      ])
-      .put(setAccount(EXPECTED_ADDRESS))
-      .put(setDataEncryptionKey(EXPECTED_DEK))
-      .returns(EXPECTED_ADDRESS)
-      .run()
-  })
+  it.each`
+    expectedAddress                                 | expectedDEK                                                               | mnemonic
+    ${'0xE025583d25Eff2C254999b5904C97bAe9B3F8D83'} | ${'0xb6812219f7003c27cc1ef17c2033c033a38cfc52d83f176a0667086787d59d39'}   | ${'avellana novio zona pinza ducha íntimo amante diluir toldo peón ocio encía gen balcón carro lingote millón amasar mármol bondad toser soledad croqueta agosto'}
+    ${'0x49a4728b4f94e5ccc4ddfe2fe2269ccd431dc052'} | ${'0x0275da8df87a83403d0a0e9aec60e26ac11ebd9764ebe3bbfc5fecb3631e3e0a99'} | ${'Sufrir maniquí índice viral boina reino choza denso gafas aviso nadar glaciar'}
```
kathaypacific, line 57: i spent some time chasing why this test doesn't work, but couldn't exactly figure it out. the tests fail because the seed phrase generates a different wallet address than the one that is expected. however, running on a live environment, the same seed phrase does generate the expected address. from these symptoms you'd assume that we're mocking something wrongly in the tests but i couldn't find which mocks would cause this behaviour.

one thing to note, this test does pass if you use an english seed phrase (e.g. the one we use for the e2e test). 
i'll note what i found and we can decide where to go from there....

the `getOrCreateAccount` saga generates a mnemonic and then gets the private key (+ account address) from the `generateKeysFromMnemonic` function. in the tests we mock generating the mnemonic, but we do not mock any part of generating the keys from the mnemonic. all `generateKeysFromMnemonic` does is decide the derivation path based on number of words in the seed phrase, then call `generateKeys` from the celo lib. the inputs passed in are the mnemonic, bip39, and the derivation path.

i logged these 3 inputs to the `generateKeys` function from the `generateKeysFromMnemonic` function, and compared the results when running the test vs restoring an account in the simulator. the results were the same, but the keys that `generateKeys` returned were different. i then went and put some log statements in the @celo/cryptographic-utils node module, those logs were printed so we are hitting the real function and not mocking it. we are also not mocking bip39 anywhere. `generateKeys` uses bip32 to generate the keys under the hood, we're not mocking that anywhere either. i just cannot understand why the same inputs to this `generateKeys` function is producing different results in a live env vs in the test env. scrolling through our mocks in the mocks folder, i couldn't see any mocks that would obviously interfere with this function.

i've exhausted the timebox i want to spend on this, maybe something in the above sparks some idea for you @MuckT ? we also discussed the value of this test yesterday, if we cover with e2e tests then perhaps it's not worth spending too much time on this unit test?

MuckT: Two things are causing this issue. The first was the the mnemonic in Spanish had accents that were different than what was seen in: https://github.com/bitcoin/bips/blob/master/bip-0039/bip-0039-wordlists.md.
Second there appears to be a mismatch of the DEK's generated by the wallet and those registered on chain. I'll create a follow-up spike to investigate the DEK issue.

jeanregisser: I had a look the DEK mismatch and it's all good.
We were comparing the public key to the private key, hence the mismatch 🤦 

For `0x49A4728b4F94E5ccC4dDFe2fE2269ccD431dc052` (mnemonic: `avellana novio zona pinza ducha íntimo amante diluir toldo peón ocio encía gen balcón carro lingote millón amasar mármol bondad toser soledad croqueta agosto`, here's the DEK:
- Public key: `0x0334b8d45fc99b433d3f294786d5d32aff0e0ce914f92ff518a42b8854b85e30fb`
- Private key: `0xb6812219f7003c27cc1ef17c2033c033a38cfc52d83f176a0667086787d59d39` 

For `0x49A4728b4F94E5ccC4dDFe2fE2269ccD431dc052` (mnemonic: `sufrir maniquí índice viral boina reino choza denso gafas aviso nadar glaciar`, here's the DEK:
- Public key: `0x0275da8df87a83403d0a0e9aec60e26ac11ebd9764ebe3bbfc5fecb3631e3e0a99`
- Private key: `0x596371d7f1dc3a1dfb9e00b51cb5ca46b73993b962c0fb0e04302004581bf45d`

src/app/saga.ts
```diff
@@ -204,6 +204,7 @@ export interface RemoteConfigValues {
   networkTimeoutSeconds: number
   dappFavoritesEnabled: boolean
   celoNews: CeloNewsConfig
+  twelveWordSeedPhraseEnabled: boolean
```
jeanregisser, line 207: nit: sorry I realize we're using `mnemonic` throughout the code base, I'd suggest we use the same terminology everywhere in this PR too.
```suggestion
  twelveWordMnemonicEnabled: boolean
```

jeanregisser (approved):
🚀

src/app/saga.ts
MuckT: Changed in 3536da4d134b87d614431868fec162e397310245!

PR #3349, opened 2023-01-11
chore: remove onboarding carousel order experiment

MuckT:
### Description

Removes the shuffle and onboarding carousel order experimentation code. Selected the most popular ordering based on [the completion data](https://docs.google.com/spreadsheets/d/1QQZQrKiAD9t0inMD3JH-DtGSDILqd3iOjwMNG8w2wmM/edit#gid=854901054). Also removed [fast-shuffle](https://www.npmjs.com/package/fast-shuffle) as it is no longer needed.

### Test plan

Tested locally in iOS. Carousel should show the consistent ordering and not switch when changing languages. 

### Related issues

- Fixes RET-543

### Backwards compatibility

Yes - Translations were kept that were used in the experiment for users on older versions of the App.


src/onboarding/education/OnboardingEducationScreen.tsx
```diff
@@ -21,36 +20,27 @@ function useStep() {
...
-        image: onboardingEducation3,
+        image: onboardingEducation1,
         topic: EducationTopic.onboarding,
-        valueProposition: 'spend',
-        variant,
       },
-    ]) as any
-    const order = steps.map((item: { valueProposition: string }) => item.valueProposition).join('-')
-    return { steps, variant, order }
+    ]
```
jeanregisser, line 43: nit: wondering if it may be confusing for localization, since stepX is no longer the real step X 😅

jeanregisser (approved):

src/onboarding/education/OnboardingEducationScreen.tsx
MuckT: I could see how that would be confusing, but as this page doesn't change often it should be ok.

PR #3357, opened 2023-01-16
fix: broken deep links on the homecards

kathaypacific:
### Description

Context: https://valora-app.slack.com/archives/C02N3AR2P2S/p1673628550702569?thread_ts=1673565367.051299&cid=C02N3AR2P2S

URLSearchParams from the `url` lib doesn't work as expected, in other places we're using this from the global scope.

### Test plan

Tested manually, added a e2e test

### Related issues

N/A

### Backwards compatibility

Y


e2e/src/usecases/HandleDeepLinkSend.js
```diff
@@ -99,6 +100,13 @@ export default HandleDeepLinkSend = () => {
       await openDeepLink(PAY_URL)
       await expect(element(by.id('SendAmount'))).not.toBeVisible()
     })
+
+    it('should navigate to another screen correctly', async () => {
+      await openDeepLink(deepLinks.navigateToSend)
+      await waitFor(element(by.id('RecipientPicker')))
+        .toBeVisible()
+        .withTimeout(10 * 1000)
+    })
```
jeanregisser: Great adding a test! 
But now I'm surprised this works, as we don't to allow deep links to arbitrary screens, for security reasons. Except when it's from ourselves (push notifications).

https://github.com/valora-inc/wallet/blob/f3b008bb6297ec679419ad8604449e474029632a/src/app/saga.test.ts#L119-L126

I wonder what code path it takes and if CleverTap may be handling it, causing `secureOrigin` to be true: https://github.com/valora-inc/wallet/blob/f3b008bb6297ec679419ad8604449e474029632a/src/app/App.tsx#L90

src/app/App.tsx
```diff
@@ -100,7 +105,9 @@ export class App extends React.Component<Props> {
       }
     })
 
-    Linking.addEventListener('url', this.handleOpenURL)
+    Linking.addEventListener('url', async (event) => {
+      return this.handleOpenURL(event, OVERRIDE_DEEP_LINK_SECURE_ORIGIN)
+    })
```
jeanregisser, line 110: We shouldn't override this for this test. It's a security feature and we should use a different deep link instead for testing.
Would be nice to actually test this security feature via the e2e test too.
i.e. using calling `openScreen` deeplink will not open said screen.

e2e/src/usecases/HandleDeepLinkSend.js
```diff
@@ -99,6 +100,13 @@ export default HandleDeepLinkSend = () => {
       await openDeepLink(PAY_URL)
       await expect(element(by.id('SendAmount'))).not.toBeVisible()
     })
+
+    it('should navigate to another screen correctly', async () => {
+      await openDeepLink(quote(deepLinks.navigateToSend))
+      await waitFor(element(by.id('RecipientPicker')))
+        .toBeVisible()
+        .withTimeout(10 * 1000)
+    })
```
jeanregisser, line 109: Great adding a test!
We should use a different test link though as we don't support opening arbitrary screens from deep links for security reasons.

https://github.com/valora-inc/wallet/blob/f3b008bb6297ec679419ad8604449e474029632a/src/app/saga.test.ts#L119-L126

jeanregisser (commented):
Thanks!

e2e/src/usecases/HandleDeepLinkSend.js
```diff
@@ -16,13 +18,34 @@ const launchDeepLink = async (url, newInstance = true) => {
...
+}
+
 const openDeepLink = async (payUrl) => {
   await reloadReactNative()
   await device.openURL({ url: payUrl })
 }
 
 export default HandleDeepLinkSend = () => {
   describe('When Launching Deeplink - App Closed', () => {
+    it('The should handle deeplink to another screen', async () => {
```
jeanregisser, line 42: Looks like we could move this to [handleNotifications](https://github.com/valora-inc/wallet/blob/f3b008bb6297ec679419ad8604449e474029632a/e2e/src/usecases/HandleNotification.js) instead: 
```suggestion
    it('Launch app from push notification should allow deeplink to another screen', async () => {
```

jeanregisser (approved):
Thanks so much 💯 🙏

PR #3360, opened 2023-01-17
chore: update cocoapods to 1.11.3

jeanregisser:
### Description

Initially started this to fix RET-568, but it will only be fixed with CocoaPods 1.12 which hasn't been released yet. So we'll have to wait. In the meantime we'll still benefit from updating, 1.10 is pretty old.

### Test plan

- CI tests pass

### Related issues

- Fixes RET-566

### Backwards compatibility

Yes


kathaypacific (approved):
LGTM, still works on my machine ✨ 🚀

PR #3361, opened 2023-01-17
fix: build with Xcode 14

jeanregisser:
### Description

This makes it possible to build and run the app with Xcode 14. This is needed to run the app with Xcode on a real device running iOS >= 16.

### Test plan

- Tested building with Xcode 14.2 works locally

### Related issues

- Fixes RET-568

### Backwards compatibility

Yes


kathaypacific (approved):
🥳

PR #3371, opened 2023-01-18
fix: ensure max number of dapps in dapps carousel

kathaypacific:
### Description

During the refactor to create the dapps carousel, the maximum number of dapps got lost :( This PR fixes that, and adds tests. The remote config variable is a bit confusingly named now since it is not specific to recently used dapps, i can change this including the actual remote config variable in a follow up PR. 

### Test plan

Added unit tests, tested manually

### Related issues

N/A

### Backwards compatibility

Y


jeanregisser (approved):
LGTM 👍

PR #3375, opened 2023-01-19
feat(cico): Enabling cREAL for cash in

jh2oman:
### Description

A fairly large refactor to end up with a small feature. The refactor is basically removing all uses of the `Currency` enum which only supports cUSD, cEUR, and CELO. This enum is assoicated with a lot of legacy code that only supports those tokens so its easier to just use our own new enum `CiCoCurrency` to add cREAL support.

https://user-images.githubusercontent.com/8432644/213536235-e50c6f2a-69a0-4bca-a4f0-368e6cc097ef.mp4

### Test plan

Should probably have a bugapalooza to go through the cash in / cash out flows and verify that everything is still working correctly. Between this and my other `goldToken` / `stableToken` refactor a lot of code has changed for the CICO flow.

### Related issues

https://linear.app/valora/issue/ACT-390/ramp-creal


src/fiatconnect/ReviewScreen.tsx
```diff
@@ -465,23 +462,34 @@ function TransactionDetails({
   feeEstimate?: FeeEstimateState
 }) {
   const exchangeRates = useSelector(localCurrencyExchangeRatesSelector)!
+  const tokenInfo = useTokenInfoBySymbol(normalizedQuote.getCryptoType())
+  const usdTokenInfo = useTokenInfoBySymbol(CiCoCurrency.cUSD)!
+  const networkFee = useLocalToTokenAmount(
+    feeEstimate?.usdFee ? new BigNumber(feeEstimate?.usdFee) : new BigNumber(0),
+    usdTokenInfo.address
+  )!
```
satish-ravi, line 470: nit: is it possible to avoid the duplication of `networkFee` computation in both `ReceiveAmount` and `TransactionDetails`?

src/fiatconnect/ReviewScreen.test.tsx
```diff
@@ -100,9 +133,9 @@ describe('ReviewScreen', () => {
       expect(queryByText('fiatConnectReviewScreen.transactionDetails')).toBeTruthy()
       expect(queryByText('fiatConnectReviewScreen.cashIn.transactionDetailsAmount')).toBeTruthy()
       expect(queryByTestId('txDetails-total/value')?.children).toEqual(['', '$', '100.00'])
-      expect(queryByTestId('txDetails-converted/value')?.children).toEqual(['', '$', '98.94'])
-      expect(queryByTestId('txDetails-fee/value')?.children).toEqual(['', '$', '1.06'])
-      expect(queryByTestId('txDetails-exchangeRate/value')?.children).toEqual(['', '$', '0.9894'])
+      expect(queryByTestId('txDetails-converted/value')?.children).toEqual(['', '$', '99.15'])
+      expect(queryByTestId('txDetails-fee/value')?.children).toEqual(['', '$', '0.84'])
+      expect(queryByTestId('txDetails-exchangeRate/value')?.children).toEqual(['', '$', '0.9915'])
```
satish-ravi, line 138: are these values changing because of different mock values for fee estimation?

src/fiatExchanges/utils.tsx
```diff
@@ -34,7 +34,7 @@ interface ProviderRequestData {
   userLocation: UserLocationData
   walletAddress: string
   fiatCurrency: LocalCurrencyCode
-  digitalAsset: CiCoCurrency
+  digitalAsset: string
```
satish-ravi, line 37: why is this changing to string?

src/fiatExchanges/quotes/FiatConnectQuote.ts
```diff
@@ -208,8 +225,8 @@ export default class FiatConnectQuote extends NormalizedQuote {
     return this.quote.quote.cryptoType
   }
 
-  getCryptoType(): Currency {
-    return resolveCurrency(this.quote.quote.cryptoType)!
+  getCryptoType(): CiCoCurrency {
+    return this.quote.quote.cryptoType as unknown as CiCoCurrency
```
satish-ravi, line 229: is this a safe typecast? would it be safer to just use a map to convert?

src/fiatExchanges/quotes/FiatConnectQuote.ts
```diff
@@ -110,19 +107,39 @@ export default class FiatConnectQuote extends NormalizedQuote {
...
+      return fee
+    }
+    const tokenUsdPrice = tokenInfo?.usdPrice
+    const usdExchangeRate = exchangeRates[Currency.Dollar]
+    if (!tokenUsdPrice || !usdExchangeRate || !fee) {
+      return null
     }
-    return convertLocalAmountToCurrency(this._getFee(), exchangeRates[this.getCryptoType()])
+
+    return fee.dividedBy(usdExchangeRate).dividedBy(tokenUsdPrice)
```
satish-ravi, line 124: Seems like this conversion logic is duplicated in ExternalQuote and here. Would be good to just have a single implementation.

src/utils/currencies.ts
```diff
@@ -4,11 +4,20 @@ export enum Currency {
...
   CELO = 'CELO',
-  CUSD = 'CUSD',
-  CEUR = 'CEUR',
-  CREAL = 'CREAL',
+  cUSD = 'cUSD',
+  cEUR = 'cEUR',
+  cREAL = 'cREAL',
+}
+
+export const cicoCurrencyTranslationKeys = {
```
satish-ravi, line 16: nit: does this need to be here? Or can this just be closer to the front end component?

src/analytics/Properties.tsx
```diff
@@ -768,21 +768,21 @@ interface FiatExchangeEventsProperties {
   [FiatExchangeEvents.cico_landing_how_to_fund]: undefined
   [FiatExchangeEvents.cico_currency_chosen]: {
     flow: FiatExchangeFlow
-    currency: Currency
+    currency: CiCoCurrency
```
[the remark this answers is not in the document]
jh2oman: yea, good point. I'll update it to keep this backwards compatible

src/fiatExchanges/utils.tsx
jh2oman: The old values were CiCoCurrency matched up with the values that cloud-function's fetchProviders expects, but that is no longer the case

src/fiatExchanges/quotes/FiatConnectQuote.ts
jh2oman: yes, good idea

src/utils/currencies.ts
jh2oman: ty for catching that. At one point i used it in multiple files, but that closed away

src/fiatExchanges/quotes/FiatConnectQuote.ts
jh2oman: yep , i joined it with some other logic and added unit tests

src/fiatconnect/ReviewScreen.test.tsx
jh2oman: because of different values for the exchange rate

src/fiatconnect/ReviewScreen.tsx
jh2oman: done

src/redux/migrations.test.ts
```diff
@@ -767,6 +768,77 @@ describe('Redux persist migrations', () => {
...
+  })
+  it('works from v104 to v105', () => {
+    const oldSchema = {
+      ...v104Schema,
+      fiatConnect: {
+        ...v104Schema.fiatConnect,
+        cachedFiatAccountUses: [
+          {
+            providerId: 'provider-two',
+            cryptoType: CiCoCurrency.cUSD,
```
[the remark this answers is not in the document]
jh2oman: thats actually just a typo on my part

src/fiatExchanges/FiatExchangeCurrency.tsx
```diff
@@ -142,15 +144,27 @@ function FiatExchangeCurrency({ route, navigation }: Props) {
...
             title="CELO"
-            selected={selectedCurrency === Currency.Celo}
-            onSelect={() => setSelectedCurrency(Currency.Celo)}
+            selected={selectedCurrency === CiCoCurrency.CELO}
+            onSelect={() => setSelectedCurrency(CiCoCurrency.CELO)}
             enabled={flow !== FiatExchangeFlow.Spend}
+            containerStyle={{
+              borderTopWidth: 0.5,
+              borderBottomWidth: 0.5,
+            }}
```
cajubelt, line 153: nit: now that there are multiple middle currency radio items, the style could be re-used between them

src/fiatExchanges/FiatExchangeCurrency.tsx
```diff
@@ -142,15 +144,27 @@ function FiatExchangeCurrency({ route, navigation }: Props) {
...
+            selected={selectedCurrency === CiCoCurrency.cREAL}
+            onSelect={() => setSelectedCurrency(CiCoCurrency.cREAL)}
+            enabled={flow === FiatExchangeFlow.CashIn}
             containerStyle={{
               borderTopWidth: 0.5,
               borderBottomLeftRadius: 8,
               borderBottomRightRadius: 8,
             }}
-            testID="radio/CELO"
+            testID="radio/CREAL"
```
cajubelt, line 167:
```suggestion
            testID="radio/cREAL"
```
nit. matches the other test id's better.

cajubelt (commented):
reviewed up thru ExternalQuote.ts, skipping tests for now

src/fiatExchanges/quotes/NormalizedQuote.ts
```diff
@@ -4,13 +4,20 @@ import { FiatExchangeEvents } from 'src/analytics/Events'
...
+import { CiCoCurrency, Currency } from 'src/utils/currencies'
 
 export default abstract class NormalizedQuote {
   abstract getPaymentMethod(): PaymentMethod
-  abstract getFeeInFiat(exchangeRates: { [token in Currency]: string | null }): BigNumber | null
-  abstract getFeeInCrypto(exchangeRates: { [token in Currency]: string | null }): BigNumber | null
-  abstract getCryptoType(): Currency
+  abstract getFeeInFiat(
+    exchangeRates: { [token in Currency]: string | null },
+    tokenInfo: TokenBalance | undefined
```
cajubelt, line 14: the tokenInfo param being optional here and for getFeeInCrypto confuses me. It seems like it is required to get a meaningful result almost all the time, except for getFeeInFiat on the ExternalQuote class.  I'd prefer if it was just made required and the consumers did null handling themselves, to make it less opaque what happens when you don't have a truthy tokenInfo value.

src/fiatExchanges/quotes/normalizeQuotes.ts
```diff
@@ -34,8 +35,14 @@ export const quotesByFeeComparator = (quote1: NormalizedQuote, quote2: Normalize
     [Currency.Dollar]: '1',
     [Currency.Euro]: '1',
   }
-  const providerFee1 = quote1.getFeeInFiat(exchangeRates) ?? new BigNumber(Infinity)
-  const providerFee2 = quote2.getFeeInFiat(exchangeRates) ?? new BigNumber(Infinity)
+  // also dummy token info. all we need is the usdPrice
+  const dummyTokenInfo = {
```
cajubelt, line 39: I commented on this already in NormalizedQuote, but this is a good example of where the type is misleading. If the second parameter was actually optional, you shouldn't have to do something like this.

src/fiatExchanges/utils.tsx
cajubelt: shouldn't there be a type for what fetchProviders expects? So we can have compile-time checks for when there might be an error in a request to that cloud function?

src/fiatExchanges/utils.tsx
```diff
@@ -255,20 +255,19 @@ export const filterLegacyMobileMoneyProviders = (
   )
 
   return activeProviders.filter((provider) =>
-    provider[selectedCurrency === CiCoCurrency.CUSD ? 'cusd' : 'celo'].countries.includes(
+    provider[selectedCurrency === CiCoCurrency.cUSD ? 'cusd' : 'celo'].countries.includes(
```
cajubelt, line 258: I realize you aren't introducing this in this PR, but it seems weird that the function assumes `selectedCurrency` is either cUSD or CELO, but the type for that parameter doesn't reflect that assumption. Seems like the type should be `CiCoCurrency.cUSD | CiCoCurrency.CELO`

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -70,6 +67,11 @@ export default function FiatConnectReviewScreen({ route, navigation }: Props) {
   const tokenAddress = tokenList.find((token) => token.symbol === cryptoType)?.address
   const feeEstimates = useSelector(feeEstimatesSelector)
   const feeEstimate = tokenAddress ? feeEstimates[tokenAddress]?.[feeType] : undefined
+  const usdTokenInfo = useTokenInfoBySymbol(CiCoCurrency.cUSD)!
+  const networkFee = useLocalToTokenAmount(
+    feeEstimate?.usdFee ? new BigNumber(feeEstimate?.usdFee) : new BigNumber(0),
+    usdTokenInfo.address
+  )!
```
cajubelt, line 74:
```suggestion
  const networkFee = useLocalToTokenAmount(
    feeEstimate?.usdFee ? new BigNumber(feeEstimate?.usdFee) : new BigNumber(0),
    usdTokenInfo.address
  ) ?? new BigNumber(0)
```

cajubelt: I'd like to try avoiding non-null assertion operators wherever possible

src/fiatExchanges/FiatExchangeAmount.tsx
```diff
@@ -295,6 +292,37 @@ function FiatExchangeAmount({ route }: Props) {
...
+        <TokenDisplay
+          amount={tokenInfo.balance}
+          tokenAddress={tokenInfo.address}
+          showLocalAmount={!displayCrypto}
+          hideSign={false}
+          style={styles.headerSubTitle}
+        />
+      </Trans>
+    ) : (
+      // TODO: a null balance doesn't necessarily mean it's loading
```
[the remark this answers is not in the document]
jh2oman: No, this is copy/pasted from when this component was in navigator/Headers.tsx. Its a 2 year old comment and I think its meant to explain the ternary operator above. 

It honestly does not make sense to me, but maybe it will be helpful if we have a bug here someday so I suggest we keep it in

src/tokens/utils.ts
```diff
@@ -80,3 +80,39 @@ function usdBalance(token: TokenBalance): BigNumber {
...
+  tokenInfo: TokenBalance | undefined
+  exchangeRates: { [token in Currency]: string | null }
+}) {
+  const tokenUsdPrice = tokenInfo?.usdPrice
+  const usdExchangeRate = exchangeRates[Currency.Dollar]
+  if (!tokenUsdPrice || !usdExchangeRate || !localAmount) {
+    return null
+  }
+
+  return localAmount.dividedBy(usdExchangeRate).dividedBy(tokenUsdPrice)
```
cajubelt, line 99:
```suggestion
  const localFiatPerDollar = exchangeRates[Currency.Dollar]
  if (!tokenUsdPrice || !usdExchangeRate || !localAmount) {
    return null
  }

  return localAmount.dividedBy(localFiatPerDollar).dividedBy(tokenUsdPrice)
```
nit: I never remember what the units are of our "exchange rates". How about this rename

src/tokens/utils.ts
```diff
@@ -80,3 +80,39 @@ function usdBalance(token: TokenBalance): BigNumber {
...
+  tokenInfo: TokenBalance | undefined
+  exchangeRates: { [token in Currency]: string | null }
+}) {
+  const tokenUsdPrice = tokenInfo?.usdPrice
+  const usdExchangeRate = exchangeRates[Currency.Dollar]
+  if (!tokenUsdPrice || !usdExchangeRate || !tokenAmount) {
+    return null
+  }
+
+  return tokenAmount.multipliedBy(tokenUsdPrice).multipliedBy(usdExchangeRate)
```
cajubelt, line 117:
```suggestion
  const localFiatPerDollar = exchangeRates[Currency.Dollar]
  if (!tokenUsdPrice || !usdExchangeRate || !tokenAmount) {
    return null
  }

  return tokenAmount.multipliedBy(tokenUsdPrice).multipliedBy(localFiatPerDollar)
```

cajubelt: same idea

src/redux/migrations.ts
```diff
@@ -978,4 +998,15 @@ export const migrations = {
       celoEducationCompleted: state.goldToken.educationCompleted,
     },
   }),
+  105: (state: any) => ({
+    ...state,
+    fiatConnect: {
+      ...state.fiatConnect,
+      cachedFiatAccountUses: state.fiatConnect.cachedFiatAccountUses.map((use: any) => ({
```
cajubelt, line 1005:
```suggestion
      cachedFiatAccountUses: state.fiatConnect.cachedFiatAccountUses.map((use: {cryptoType: Currency}) => ({
```

src/redux/migrations.ts
```diff
@@ -8,7 +8,27 @@ import { SendingFiatAccountStatus } from 'src/fiatconnect/slice'
...
 import { TokenTransaction } from 'src/transactions/types'
-import { Currency } from 'src/utils/currencies'
+import { CiCoCurrency, Currency } from 'src/utils/currencies'
+
+export function updateCachedQuoteParams(cachedQuoteParams: {
+  [providerId: string]: {
+    [kycSchema: string]: any
+  }
+}) {
+  const newCachedQuoteParams: any = {}
```
cajubelt, line 18: why not use a more descriptive return type here? seems like it would only make the code safer

cajubelt (commented):
reviewed the rest, minus tests

src/fiatExchanges/FiatExchangeAmount.tsx
```diff
@@ -295,6 +292,37 @@ function FiatExchangeAmount({ route }: Props) {
...
+      i18n.t('loading')
+    )
+
+  return <HeaderTitleWithSubtitle title={title} subTitle={subTitle} />
```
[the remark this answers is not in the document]
jh2oman: I did something similar to what you suggested but instead combined this header function with the existing `HeaderTitleWithTokenBalance` component

src/fiatExchanges/FiatExchangeAmount.tsx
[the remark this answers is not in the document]
jh2oman: sidestepped now by not using this implementation

src/fiatExchanges/utils.tsx
```diff
@@ -294,3 +300,15 @@ export const filterProvidersByPaymentMethod = (
...
+export function resolveCloudFunctionDigitalAsset(
+  currency: CiCoCurrency
+): CloudFunctionDigitalAsset {
+  const mapping: Record<CiCoCurrency, CloudFunctionDigitalAsset> = {
+    [CiCoCurrency.CELO]: CloudFunctionDigitalAsset.CELO,
+    [CiCoCurrency.cUSD]: CloudFunctionDigitalAsset.CUSD,
+    [CiCoCurrency.cEUR]: CloudFunctionDigitalAsset.CEUR,
+    [CiCoCurrency.cREAL]: CloudFunctionDigitalAsset.CREAL,
+  }
+  return mapping[currency] || CiCoCurrency.CELO
```
cajubelt, line 313:
```suggestion
  return mapping[currency]
```
this fallback is not necessary and a little misleading imo. If a currency is missing from the mapping, there will be a compile-time error.

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -416,7 +418,8 @@ function getDisplayAmounts({
   } else {
     const receive = normalizedQuote.getCryptoAmount()
     const total = normalizedQuote.getFiatAmount()
-    const fee = normalizedQuote.getFeeInFiat(exchangeRates, tokenInfo)
+    const fee =
+      (!!tokenInfo && normalizedQuote.getFeeInFiat(exchangeRates, tokenInfo)) || new BigNumber(0)
```
cajubelt, line 422:
```suggestion
      tokenInfo ? normalizedQuote.getFeeInFiat(exchangeRates, tokenInfo) : new BigNumber(0)
```
nit: this is a bit clearer imo

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -375,7 +376,8 @@ function getDisplayAmounts({
   const cryptoType = normalizedQuote.getCryptoType()
   const fiatType = normalizedQuote.getFiatType()
   if (flow === CICOFlow.CashOut) {
-    const providerFee = normalizedQuote.getFeeInCrypto(exchangeRates, tokenInfo) ?? new BigNumber(0)
+    const providerFee =
+      (!!tokenInfo && normalizedQuote.getFeeInCrypto(exchangeRates, tokenInfo)) || new BigNumber(0)
```
cajubelt, line 380:
```suggestion
      tokenInfo ? normalizedQuote.getFeeInCrypto(exchangeRates, tokenInfo) : new BigNumber(0)
```

cajubelt: nvm, this messes up the types since it's not actually the same

src/fiatconnect/ReviewScreen.tsx
cajubelt: nvm, this messes up the types since it's not actually the same

cajubelt (approved):

satish-ravi (approved):

PR #3376, opened 2023-01-20
chore: restore merchant payments feature

MuckT:
### Description

Restores the Merchant Payment feature removed in #3190.

### Test plan

Tested on Android using https://sandbox-celo-demo-store.demo.firstdag.com. All steps except the final step were successful - this is likely an issue with the sandbox site.

1. Select an item.
2. Scan with QR code scanner in Valora.
3. Send Payment.
4. Observe payment sent and balance updated.
5. Observe payment on website is successful.

#### Video

https://user-images.githubusercontent.com/26950305/214186864-f4432e2c-769f-49bd-ac56-c95187c2f022.mov

### Related issues

- Fixes RET-564

### Backwards compatibility

Yes


MuckT:
Associated [PR](https://github.com/celo-org/payments/pull/33) to update dependencies in [@celo-org/payments](https://github.com/celo-org/payments).

MuckT:
@Kathy @Jean there appears to be an issue with the initial start of the app and getting the DEK which is needed for the merchant payment features. On a fresh install after a restore when attempting to scan the QR code from https://sandbox-celo-demo-store.demo.firstdag.com. The DEK is undefined when hitting `if (!wallet.isAccountUnlocked(account) || !wallet.isAccountUnlocked(dekAddress))` in `src/merchantPayment/hooks.ts`; however, if I close and reopen the app the DEK is present and merchant payments is working as expected.

MuckT:
@jeanregisser and @kathaypacific just want to test this a bit more, but I think this is ready for review.

locales/base/translation.json
```diff
@@ -1238,6 +1238,13 @@
   "others": "Others",
   "nomSpaceRecipient": "Nomspace: {{name}}",
   "lastDay": "Last 24 hrs",
+  "merchant": "Merchant Payment",
+  "merchantMoneyEscrow": "You're about to send a direct payment to a merchant.",
+  "merchantPaymentNumberVerificationMessage": "You need to connect your number",
+  "merchantPaymentSubmitFail": "Your payment failed to submit",
+  "merchantPaymentSetup": "There was an error loading the merchant information, please try scanning the QRCode again",
+  "merchantPaymentSubmitSuccess": "Your payment successfully submitted",
```
kathaypacific, line 1246:
```suggestion
  "merchantPaymentSubmitSuccess": "Your payment was successfully submitted",
```

src/merchantPayment/types.ts
```diff
@@ -0,0 +1,4 @@
+export enum PaymentDeepLinkHandler {
```
kathaypacific, line 1: this type is a little odd, should we just have a boolean `merchantPaymentEnabled` remote config variable instead? we can add more variants if we use an enum but seems unlikely that we want to build on this?

src/app/saga.ts
```diff
@@ -280,6 +283,10 @@ export function* handleDeepLink(action: OpenDeepLink) {
     const pathParts = rawParams.path.split('/')
     if (rawParams.path.startsWith('/pay')) {
       yield call(handlePaymentDeeplink, deepLink)
+    } else if (rawParams.path.startsWith('/payment')) {
```
kathaypacific, line 286: maybe this condition should go ahead of the previous one that checks for `/pay`? seems like a path starting with `/payment` also starts with `/pay` so does this block get triggered?

src/merchantPayment/FeeContainer.test.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { createMockStore } from 'test/utils'
+
+describe('FeeContainer', () => {
+  it('renders correctly', () => {
+    const tree = render(
+      <Provider store={createMockStore({})}>
+        <FeeContainer amount={new BigNumber(1337)} />
+      </Provider>
+    )
+    expect(tree).toMatchSnapshot()
```
kathaypacific, line 16: sorry this was probably there in the previous implementation but can we avoid adding new snapshot tests and test for the expected text etc instead? i think we discussed this some time ago, snapshots are quick to write but difficult to read / review the changes later on and serve as worse documentation about what the component is supposed to do

src/merchantPayment/FeeContainer.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { useTokenInfoBySymbol } from 'src/tokens/hooks'
+import { fetchTokenBalances } from 'src/tokens/slice'
+import { Currency } from 'src/utils/currencies'
+import Logger from 'src/utils/Logger'
+import { isDekRegisteredSelector } from 'src/web3/selectors'
+
+export default function FeeContainer({ amount }: { amount: BigNumber }) {
+  const LOG_TAG = BASE_TAG + 'FeeContainer'
+
+  const dispatch = useDispatch() as (...args: unknown[]) => void
```
kathaypacific, line 21: i don't think we need this `as` type casting?

src/merchantPayment/FeeContainer.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const feeEstimate = feeEstimates[tokenAddress]?.[FeeType.SEND]
+  const storedDekFee = feeEstimates[tokenAddress]?.[FeeType.REGISTER_DEK]
+  const securityFee = feeEstimate?.usdFee ? new BigNumber(feeEstimate.usdFee) : undefined
+  const dekFee = storedDekFee?.usdFee ? new BigNumber(storedDekFee.usdFee) : undefined
+  const totalFee = securityFee?.plus(dekFee ?? 0)
+
+  return (
+    <View style={styles.feeContainer}>
+      <FeeDrawer
+        testID={'feeDrawer/SendConfirmation'}
```
kathaypacific, line 63: this doesn't matter much but should this be a different test id?

src/merchantPayment/MerchantPaymentScreen.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+      },
+      exchange: { exchangeRates },
+    })
+
+    const tree = render(
+      <Provider store={store}>
+        <MerchantPaymentScreen {...mockScreenProps} />
+      </Provider>
+    )
+    expect(tree).toMatchSnapshot()
```
kathaypacific, line 45: ah i see the snapshot test here too. this isn't the hill i want to die on, maybe it's not worth the effort to update the tests for this merchant payments feature

kathaypacific (commented):
Looks good, i had some nit picky comments that you can also ignore since most of this code is just being restored and passed reviews in the past.

one thing that i'm not sure about is the new vulnerability https://github.com/advisories/GHSA-hj79-42mx-m4gr that we would be pulling in. it sounds a little scary. it exists in a dependency of an version 1.2.0 of @celo/utils, which the @celo/payments-types and @celo/payments-sdk packages depend on. it would be great if the payments sdks could use a newer version of @celo/utils (version 1.5.2 seems to be fine, it doesn't even have to be the newest version) - idk if you've looked into this and decided that it isn't worth the effort? i'm not sure what the effort is there, but i vaguely remember discussing this at a refinement

src/app/saga.ts
MuckT: Order changed in 93d3cbb9bd74fa53dbf9a1d6db2de055ddd939bf!

src/merchantPayment/FeeContainer.tsx
MuckT: Remove type casting in 651eed8bb42eee9d8c13004ff2a3f6a262999709!

locales/base/translation.json
MuckT: Fixed in 47521465ab4da4fe5b17f591fb05d3db2b1cdc75!

src/merchantPayment/hooks.ts
```diff
@@ -0,0 +1,95 @@
...
+  }, [])
+
+  const submit = useCallback(async () => {
+    if (!charge || paymentStatus === PaymentStatus.Pending) return
+
+    setStatus(PaymentStatus.Pending)
+    try {
+      // Temporarily disabled for celo connect
+      // await charge.submit({ phoneNumber: e164PhoneNumber })
+      await charge.submit({ address: { country: 'ES' } })
```
MuckT, line 68: @kathaypacific not sure what we should replace this with, but hardcoding a country doesn't seem correct.

kathaypacific: sorry i don't have much context on this feature, what is this part for? if it wasn't needed for celo connect, is it needed now? naively looking at this i'd expect it to use the same country as the user's location but i'm not sure. maybe @jeanregisser you have an idea?

MuckT: I'll use the users country from the user location selector for now.

jeanregisser: Yes because phone number verification was quite unreliable at the time, they switched to just using the address.
I guess hardcoded to ES for simplicity, but it seems logical to use the user country 👍

jeanregisser (approved):
LGTM 👍

PR #3384, opened 2023-01-20
chore!: disable split ABI, it's not needed with AAB

jeanregisser:
### Description

Using split ABI (one APK per arch), causes the Android `versionCode` to be different per arch ([see documentation](https://developer.android.com/studio/build/configure-apk-splits#configure-APK-versions)).
This can cause confusion where the value is displayed (Play Store, Sentry, Slack release automation message).
And makes it different from the build version shown on iOS.

Since we are using the AAB format to upload to the Play Store, we don't actually need to do that anymore.
And we can go back to the `versionCode` being the commit timestamp.

See the related change for the build automation: https://github.com/valora-inc/release-automation/pull/54

### Related issues

N/A

### Backwards compatibility

The Android versionCode will now be the same across architectures.


kathaypacific (commented):
LGTM, but the android CI is failing...on the package.json we explicitly reference the x86_64 apk for the tests, maybe this needs to be updated?

jeanregisser:
> LGTM, but the android CI is failing...on the package.json we explicitly reference the x86_64 apk for the tests, maybe this needs to be updated?

Thanks, updated 👍

kathaypacific (approved):
✨

PR #3385, opened 2023-01-20
feat(ramp): Enable Ramp Cash-Outs

jophish:
### Description

For ACT-555. Enables Ramp Cash-Outs in the Select Provider screen. Currently pending Jacob's related PR in the cloud-function repo to merged and deployed. Our data model for legacy provider quotes is getting stretched to its limit; we now consider a new class of provider responses which _are marked_ as available, but contain no detailed quote information.

### Test plan

Unit and manual tested, see video:

https://user-images.githubusercontent.com/569401/213820519-c07df08f-54c9-4c79-b6bf-b7c144436887.mp4

### Related issues

- Fixes ACT-555.

### Backwards compatibility

Yes, backwards compatible.


src/fiatExchanges/quotes/normalizeQuotes.ts
```diff
@@ -78,20 +79,42 @@ export function normalizeFiatConnectQuotes(
...
+        if (provider.quote.length) {
+          quotes.push(...provider.quote)
+        } else if (!provider.unavailable && !provider.restricted) {
+          // If no quotes are provided, but the provider is not marked as unavailable,
+          // this means the provider supports transfers, but does not give specific quote details.
+          // We construct a partial quote object for each payment method. N.B., this will never occur
+          // for Simplex quotes, which are a special case.
+          provider.paymentMethods.forEach((paymentMethod) =>
+            quotes.push({
+              digitalAsset,
```
jophish, line 103: @jh2oman Now that `RawProviderQuote` includes _optional_ fields, rather than post-hoc constructing partial quote objects in the wallet, we can actually construct them in the cloud function instead. This would make the wallet code a bit cleaner (and prevent us from passing `digitalAsset` through here), with little(?) added complexity in the cloud function, I think.

src/fiatExchanges/quotes/ExternalQuote.ts
```diff
@@ -73,11 +73,15 @@ export default class ExternalQuote extends NormalizedQuote {
...
-      ? new BigNumber(this.quote.fiat_money.total_amount).minus(
-          new BigNumber(this.quote.fiat_money.base_amount)
-        )
-      : new BigNumber(this.quote.fiatFee)
+    if (isSimplexQuote(this.quote)) {
+      return new BigNumber(this.quote.fiat_money.total_amount).minus(
+        new BigNumber(this.quote.fiat_money.base_amount)
+      )
+    } else if (typeof this.quote.fiatFee === 'number') {
+      // Can't just check for truthiness since '0' fails this and introduces a regression
```
anthonydaoud, line 81: I'm a bit confused by this comment in part because I can't figure out the scenario in which a value of '0' is received. Aren't the quotes supposed to return a number from the [cloud function](https://github.com/valora-inc/cloud-functions/blob/main/src/cico/fetchProviders.ts#L40) for fiatFee (assuming they're not a simplex quote)?

anthonydaoud:
Looks pretty good to me overall!

src/fiatExchanges/quotes/ExternalQuote.ts
jophish: I'm not sure it's impossible for `fiatFee` to be `0`; I assumed there was a possibility that could happen; if that's the case, and it's possible for our cloud function to return the number `0`, then just checking truthiness would introduce a regression here. (This was actually caught by one of our existing unit tests, which uses mock data where the `fiatFee` was already set to `0`.)

test/values.ts
```diff
@@ -674,6 +674,19 @@ export const mockProviders: FetchProvidersOutput[] = [
     cashIn: true,
     cashOut: false,
   },
+  {
+    name: 'Ramp',
+    restricted: false,
+    unavailable: false,
+    paymentMethods: [PaymentMethod.Card, PaymentMethod.Bank],
```
jh2oman, line 681: nit: will only be Bank for cash outs

jh2oman (approved):

test/values.ts
jophish: yeah, but just wanted to verify that we do in fact generate one quote per payment method in this case.

src/fiatExchanges/quotes/ExternalQuote.ts
satish-ravi: nit: can you use backtick instead of single quote for 0, I originally read this as a string containing 0

satish-ravi (approved):

PR #3390, opened 2023-01-24
fix: android modal background not covering the whole screen

kathaypacific:
### Description

On android, some modal backgrounds do not go all the way down to the bottom of the screen. The fix suggested in the react-native-modal [doc](https://github.com/react-native-modal/react-native-modal#the-backdrop-is-not-completely-filledcovered-on-some-android-devices-galaxy-for-one) is outdated and points to an abandoned project, but luckily  https://github.com/th3rdwave/react-native-safe-area-context#usesafeareaframe seems to work!

Before:
Screenshot_20230124-143954_Valora

After:
Screenshot_20230124-143146_Alfajores (dev)

### Test plan

Tested manually.

### Related issues

n/a

### Backwards compatibility

Y


kathaypacific:
it is distressing that https://github.com/Sunhat/react-native-extra-dimensions-android is no longer maintained....i'll see if there's another way to hack at this

jeanregisser:
> it is distressing that https://github.com/Sunhat/react-native-extra-dimensions-android is no longer maintained....i'll see if there's another way to hack at this

Could we use https://github.com/th3rdwave/react-native-safe-area-context#usesafeareaframe instead?

src/components/Modal.tsx
```diff
@@ -14,12 +14,15 @@ interface Props {
...
 export default function Modal({ children, isVisible, style, testID, onBackgroundPress }: Props) {
+  const { height } = useSafeAreaFrame()
+
   return (
     <ReactNativeModal
       testID={testID}
       isVisible={isVisible}
       backdropOpacity={0.1}
       onBackdropPress={onBackgroundPress}
+      deviceHeight={height}
```
jeanregisser, line 25: nit:
```suggestion
      // The default uses `Dimensions.get('window').height` but sometimes reports an incorrect height
      // `useSafeAreaFrame()` seems to work better
      deviceHeight={height}
```

jeanregisser (approved):
👩‍🎨 👍

PR #3391, opened 2023-01-24
fix: remove phone number from app after revoke action

kathaypacific:
### Description

This PR removes the phone number and country code from the redux store after a user revokes their number. This prevents the phone number from being displayed after a revoke action.

### Test plan

Tested manually - this is not a standard user action.

### Related issues

n/a

### Backwards compatibility

Y


jeanregisser (approved):
LGTM 👍

PR #3400, opened 2023-01-25
fix(fiatconnect): avoid allowing user to retry if tx might still be pending

cajubelt:
### Description

We've seen a case in testing where a user got this error in their logs
```
info [2023-01-23T17:15:17.387Z] FiatConnectSaga :: Transaction failed: c2ba9fc0-4f75-4526-b383-b59e947fc981 :: Failed to check for transaction receipt:
{} in _fireError@/private/var/containers/Bundle/Application/DA4D4DB7-095F-4CFC-A18B-4162D7E9283D/celo.app/ :: network connected true
```
and was presented with the option to retry the FiatConnect transfer, but when they retried, they ended up with 2 transfers. 

If we get that tx error, we should avoid allowing the user to retry their transfer, since it may still be confirmed.

### Test plan

unit tested. not sure we can reproduce this one manually.

### Related issues

na

### Backwards compatibility

yes

### DIFF
--- a/src/transactions/send.test.ts
+++ b/src/transactions/send.test.ts
@@ -73,6 +73,10 @@ describe('isTxPossiblyPending', () => {
     const result = isTxPossiblyPending(new Error('nonce too low error!!!'))
     expect(result).toBe(true)
   })
+  it('returns true when failed to check for tx receipt', () => {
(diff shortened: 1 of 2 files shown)


jophish (approved):
Thanks for adding this in, glad the fix is so simple.

PR #3401, opened 2023-01-25
feat(fiatconnect): cache transfers & quotes by tx hash

anthonydaoud:
### Description

Currently, there is no mechanism to store a successful FiatConnect transaction. This is needed because in ACT-578, we plan to populate FiatConnect transfers in the transaction feed with greater details about the transaction. This PR adds a new `cachedTransfers` array to the redux state, and updates the FiatConnect saga to append successful transactions to it.  

### Test plan

Added new unit tests

### Related issues

- ACT-579

### Backwards compatibility

Backwards compatible


satish-ravi (commented):
Changes look good, just a comment on the state variable type

src/fiatconnect/saga.ts
```diff
@@ -1024,6 +1028,17 @@ export function* handleCreateFiatConnectTransfer(
         txHash: transactionHash,
       })
     )
+    if (transactionHash) {
```
cajubelt, line 1031: we won't have the transactionHash for transfers in, so it seems like this could be merged with the section above.
```suggestion
```

even if we changed the spec to have the provider give us transactionHash for transfers in, they'd need to debit the user's fiat account first before giving this, meaning we'd get the tx hash async after a potentially substantial delay. so I think it's very safe to assume that we'll only have transaction hash for transfers out.

src/fiatconnect/selectors.ts
```diff
@@ -20,3 +20,6 @@ export const cachedQuoteParamsSelector = (state: RootState) => state.fiatConnect
 export const schemaCountryOverridesSelector = (state: RootState) =>
   state.fiatConnect.schemaCountryOverrides
 export const personaInProgressSelector = (state: RootState) => state.fiatConnect.personaInProgress
+export const getCachedFiatConnectTransfer = (txHashToFind: string) => (state: RootState) => {
+  return state.fiatConnect.cachedTransfers.find(({ txHash }) => txHash === txHashToFind)
```
cajubelt, line 24: are you sure both of these values will always have the same case?

src/fiatconnect/slice.ts
```diff
@@ -47,11 +47,20 @@ export interface CachedQuoteParams {
   fiatType: FiatType
 }
 
+export interface TransferDetails {
```
cajubelt, line 50: nit: maybe CachedTransferDetails?

src/fiatconnect/slice.ts
```diff
@@ -147,6 +157,14 @@ export interface CreateFiatConnectTransferCompletedAction {
   txHash: string | null
 }
 
+export interface CacheFiatConnectTransferAction {
```
cajubelt, line 160: this looks the same as the type you defined above. Seems like making it an alias for the other type would be more DRY

src/fiatconnect/slice.ts
```diff
@@ -147,6 +157,14 @@ export interface CreateFiatConnectTransferCompletedAction {
...
+  txHash: string
+  transferId: string
+  providerId: string
+  fiatAccountId: string
+  quote: FiatConnectQuoteSuccess
+}
```
cajubelt, line 166:
```suggestion
export type CacheFiatConnectTransferAction = TransferDetails
```
like this

src/fiatconnect/slice.ts
```diff
@@ -47,11 +47,20 @@ export interface CachedQuoteParams {
   fiatType: FiatType
 }
 
+export interface TransferDetails {
+  txHash: string
+  transferId: string
+  providerId: string
+  fiatAccountId: string
+  quote: FiatConnectQuoteSuccess
```
cajubelt, line 55: seems like we don't actually need a lot of the stuff from FiatConnectQuoteSuccess, like what fiat account schemas were allowed, or what kyc was required. What about just saving the stuff from the "quote" field of the [quote response](https://github.com/fiatconnect/specification/blob/main/fiatconnect-api.md#34112-responses)?

cajubelt (commented):
looks close

src/fiatconnect/saga.ts
```diff
@@ -1002,13 +1003,25 @@ export function* handleCreateFiatConnectTransfer(
...
+    )
 
     if (flow === CICOFlow.CashOut) {
       transactionHash = yield call(_initiateSendTxToProvider, {
         transferAddress,
         fiatConnectQuote,
       })
+      yield put(
+        cacheFiatConnectTransfer({
+          txHash: transactionHash!,
```
anthonydaoud, line 1018: I'm aware we are trying not to use the non-null assertion operator. However, the return type from the generator is currently `any`, so we're already casting to make the transactionHash into a string.

src/fiatconnect/selectors.ts
```diff
@@ -20,3 +20,6 @@ export const cachedQuoteParamsSelector = (state: RootState) => state.fiatConnect
 export const schemaCountryOverridesSelector = (state: RootState) =>
   state.fiatConnect.schemaCountryOverrides
 export const personaInProgressSelector = (state: RootState) => state.fiatConnect.personaInProgress
+export const getCachedFiatConnectTransfer = (txHashToFind: string) => (state: RootState) => {
```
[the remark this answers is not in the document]
anthonydaoud: updated!

src/fiatconnect/saga.ts
anthonydaoud: Updated it

src/fiatconnect/slice.ts
anthonydaoud: Good call, switched to use that

src/fiatconnect/selectors.ts
anthonydaoud: Not sure, will make lowercase just to be safe.

src/fiatconnect/selectors.ts
```diff
@@ -20,3 +21,8 @@ export const cachedQuoteParamsSelector = (state: RootState) => state.fiatConnect
 export const schemaCountryOverridesSelector = (state: RootState) =>
   state.fiatConnect.schemaCountryOverrides
 export const personaInProgressSelector = (state: RootState) => state.fiatConnect.personaInProgress
+export const getCachedFiatConnectTransfer =
```
satish-ravi, line 24: nit: Looks like all methods end with the name Selector.. maybe `getCachedFiatConnectTransferSelector` or `getCachedTransferSelector`?

src/fiatconnect/saga.ts
```diff
@@ -1002,13 +1003,25 @@ export function* handleCreateFiatConnectTransfer(
...
+    )
 
     if (flow === CICOFlow.CashOut) {
       transactionHash = yield call(_initiateSendTxToProvider, {
         transferAddress,
         fiatConnectQuote,
       })
+      yield put(
+        cacheFiatConnectTransfer({
+          txHash: transactionHash!.toLowerCase(),
```
satish-ravi, line 1018: I guess `_initiateSendTxToProvider` guarantees returning a string, but maybe safer to just do an additional null check here and log a warning/error and skip caching transfer if its null?

Or you can also use a local variable to avoid the `!` operator. E.g.,
```js
const cashOutTxHash: string = yield call(...)
yield put(cacheFiatConnectTransfer({txHash: cashOutTxHash.toLowerCase(), ...}))
transactionHash = cashOutTxHash
```

src/fiatconnect/types.ts
```diff
@@ -22,3 +22,5 @@ export class FiatConnectTxError extends Error {
     this.error = error
   }
 }
+
+export type QuoteResponseQuoteField = QuoteResponse['quote']
```
satish-ravi, line 26: nit: you could just inline this type in `CachedTransferDetails`. Don't think we get much out of this exported type

src/fiatconnect/selectors.ts
anthonydaoud: Ok, updated to `getCachedFiatConnectTransferSelector`

src/fiatconnect/saga.ts
```diff
@@ -997,13 +998,26 @@ export function* handleCreateFiatConnectTransfer(
...
   try {
-    const { transferAddress }: TransferResponse = yield call(_initiateTransferWithProvider, action)
+    const { transferAddress, transferId }: TransferResponse = yield call(
+      _initiateTransferWithProvider,
+      action
+    )
 
     if (flow === CICOFlow.CashOut) {
-      transactionHash = yield call(_initiateSendTxToProvider, {
+      const cashOutTxHash: string = yield call(_initiateSendTxToProvider, {
```
jophish, line 1007: Why not just keep the assignment to `transactionHash` here? We won't have one for cash-ins anyways, so don't think we need to differentiate here.

src/fiatconnect/slice.ts
```diff
@@ -47,11 +48,19 @@ export interface CachedQuoteParams {
   fiatType: FiatType
 }
 
+export interface CachedTransferDetails {
+  transferId: string
+  providerId: string
+  fiatAccountId: string
+  quote: QuoteResponse['quote']
```
jophish, line 55: This `quote` is typed to pull the `quote` object out of the [quote response schema](https://github.com/fiatconnect/fiatconnect-types/blob/main/src/quote.ts#L117), right? Why don't we store the entire quote (i.e., the `kyc` and `fiatAccount` objects as well)?

src/fiatconnect/slice.ts
```diff
@@ -47,11 +48,19 @@ export interface CachedQuoteParams {
   fiatType: FiatType
 }
 
+export interface CachedTransferDetails {
```
jophish, line 51: Is it potentially worth it to include the `transactionHash` itself in the `CachedTransferDetails` interface? I know it's used as a key to the `cachedTransfers` map, but it may be useful so that callers don't have to keep track of `transactionHash` separately.

anthonydaoud: Added it!

satish-ravi (approved):

jophish (approved):

jophish:
Noticed that tests are failing; A pending a passing build!

PR #3402, opened 2023-01-26
feat: move dapps disclaimer from bottom sheet to bottom of dapps screen

kathaypacific:
### Description

This PR:
- adds a remote config variable for the dapps disclaimer experiment
- for the variant group, display disclaimer at the bottom of the dapps screen and do not display disclaimer bottom sheet on select dapp

https://user-images.githubusercontent.com/20150449/214815475-185d080f-1135-46ea-bfed-2d865556506b.mp4

### Test plan

Updated unit tests

### Related issues

- Fixes RET-550

### Backwards compatibility

Y


src/app/saga.ts
```diff
@@ -205,6 +205,7 @@ export interface RemoteConfigValues {
   dappFavoritesEnabled: boolean
   celoNews: CeloNewsConfig
   twelveWordMnemonicEnabled: boolean
+  dappsDisclaimerMinimal: boolean
```
jeanregisser, line 208: ultra nit: naming is hard, but I wonder if we should use this for clarity
```suggestion
  dappsMinimalDisclaimerEnabled: boolean
```

jeanregisser (approved):
LGTM 👍

src/app/saga.ts
kathaypacific: ah, thank you. that is better! i reeeally struggled to name this guy, i'll update

PR #3403, opened 2023-01-26
feat: add dapps info bottom sheet

kathaypacific:
### Description

This PR makes the help dialog a bottom sheet, with updated copy and a button that links to the dapps support page. Also adds analytics events on the buttons.

https://user-images.githubusercontent.com/20150449/214863147-617b8941-c2f7-40e6-9564-160acdb85146.mp4

### Test plan

Added a unit test, updated E2E test.

### Related issues

- Fixes RET-550

### Backwards compatibility

Y


src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -70,6 +69,7 @@ export function DAppsExplorerScreen() {
 
   const { onSelectDapp, ConfirmOpenDappBottomSheet } = useOpenDapp()
   const { onFavoriteDapp, DappFavoritedToast } = useDappFavoritedToast(sectionListRef)
+  const { onOpenDappInfo, DappInfoBottomSheet } = useDappInfoBottomSheet()
```
jeanregisser, line 72: Is there a downside to making it an explicit screen instead?
Is it because the sheet remains visible while animating to show the in-app browser? and we want the component mounted here.

Why am I asking? I'm wondering whether it's a good or bad pattern to use a hook to return a component and some actions with it 😅 

One thing is it makes it less natural to pass custom props to the component. We don't need custom props here, but thinking more generally about this pattern.

I guess I would have a slight preference for keeping the same pattern we use for `<Dialog />` or `<Modal />`. Even if it means having a `useState` to control the visibility (or a `ref`).
As it is just simple React.

Curious what you think here.

src/dappsExplorer/useDappInfoBottomSheet.tsx
```diff
@@ -0,0 +1,97 @@
...
+            testID="DAppsExplorerScreen/InfoBottomSheet/PrimaryAction"
+          />
+        </View>
+      </BottomSheet>
+    ),
+    [animatedSnapPoints, animatedHandleHeight, animatedContentHeight, handleContentLayout]
+  )
+
+  return {
+    onOpenDappInfo,
```
jeanregisser, line 74:
```suggestion
    openSheet,
```

jeanregisser (commented):
Awesome! 👍

src/dappsExplorer/DAppsExplorerScreen.tsx
kathaypacific: yeah good point, i briefly considered using a screen since there's a pattern i could copy with the WC stuff but i thought it didn't make a lot of sense semantically. this sheet isn't a standalone screen and is more of a component of the dapps screen - i guess this is not an objective reason though and could be disputed 😅

the other reason i didn't go with a screen is i couldn't think of how this pattern would scale. i can imagine us wanting to use bottom sheets in many places going forward - does that mean we'd have a new screen per sheet? or a generic sheet screen where we pass in the contents as navigation params? neither option seemed appealing to me, but maybe there's an obvious appealing alternative option that i missed?

as for hooks returning JSX...yeah that feels funny but the main advantage i see is that it keeps related logic together and allows the dapps screen component to be easy to read. with the previous `<Dialog />` approach, the dapps screen component itself needed to orchestrate the visibility and actions within the dialog and keep extra state. that's fine but makes the component way more difficult to read / update. especially with this bottom sheet where we need to do some magic to make the height adjust dynamically to the content height, i just preferred to keep this together with the bottom sheet itself and not pollute the whole screen component. i think this pattern works well from a readability perspective but yeah i agree, i wouldn't use this pattern for hooks that could be reused widely or where the returned values need to be highly customisable.

kathaypacific: i was trying to remember why we started this pattern - and i see that it was a suggestion from Gonza https://github.com/valora-inc/wallet/pull/2039#discussion_r813143587 i see you mentioned using a separate screen back then too. i don't have strong opinions on this, i'm also happy to go with a screens approach if you think it's worth the tradeoff.

i did keep this pattern also for the `useDappFavoritedToast` hook, i think the dapps screen component without this hook and without `useDappFavoritedToast` would be a lot harder and scarier to maintain though

PR #3405, opened 2023-01-27
fix: crash when dappListApiUrl is empty

jeanregisser:
### Description

Title says it all.

### Test plan

- Added new tests
- Manually by setting an empty `dappListApiUrl` and seeing the crash before the fix

### Related issues

- Fixes RET-585

### Backwards compatibility

Yes

### DIFF
--- a/src/navigator/DrawerNavigator.tsx
+++ b/src/navigator/DrawerNavigator.tsx
@@ -261,7 +261,7 @@ export default function DrawerNavigator() {
         celoMenuItem
       )}
 
-      {dappsListUrl && (
+      {!!dappsListUrl && (
         <Drawer.Screen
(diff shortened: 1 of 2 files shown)


src/navigator/DrawerNavigator.tsx
```diff
@@ -261,7 +261,7 @@ export default function DrawerNavigator() {
         celoMenuItem
       )}
 
-      {dappsListUrl && (
+      {!!dappsListUrl && (
```
jeanregisser, line 264: This bug is easy to overlook. When the string is empty, it returns it. But in RN text is only valid within `<Text>` components.

It's likely a linting rule could help us catch this too.

MuckT (approved):
Great fix! Thanks @jeanregisser.

PR #3406, opened 2023-01-30
feat: improved dapps bottom sheet UI

kathaypacific:
### Description

This PR implements most of the new styles in this design https://www.figma.com/file/vfd4W7wT2W6jGwfIscJT9n/Dapp-Disclaimer?node-id=1317%3A2946&t=yGhlJHXU86MuiEaR-0. Specifically:
- the dapp logos are now smaller / left aligned / with shadows
- the copy for the bottom sheet titles have been simplified
- the description for connect / transact bottom sheets have been updated. the action request details are now part of the description.
- the transaction details are now copy-able, with a toast
- no more "cancel" button - swipe to dismiss the sheet + deny the request

This PR does not address:
- automatically adjusting the height of the bottom sheet according to content size
- adding the dapp disclaimer

### Test plan

Video of WC interaction:

https://user-images.githubusercontent.com/20150449/215516403-08f3d87d-b5a1-4710-bde5-04a334e34ce0.mp4

Video of dappkit interaction:

https://user-images.githubusercontent.com/20150449/215516420-2b830405-de71-449d-8b17-6f985edcf504.mp4

### Related issues

- Fixes RET-551

### Backwards compatibility

Yes - have used new transaction strings to avoid breaking existing apps


e2e/src/usecases/WalletConnectV1.js
```diff
@@ -58,24 +52,10 @@ export default WalletConnect = () => {
   })
 
   beforeEach(async () => {
-    // Sleep a few seconds for runs in ci
+    // wait for any banners to disappear
     await sleep(3 * 1000)
   })
 
-  // Used to prevent a failing test spec form failing the entire test suite
-  afterEach(async () => {
```
kathaypacific, line 66: i think the components don't exist in the WC components anymore, they were from the full screen implementation

e2e/src/usecases/WalletConnectV1.js
```diff
@@ -310,7 +285,7 @@ export default WalletConnect = () => {
     }
 
     let result = walletConnector.sendCustomRequest(customRequest)
-    await waitFor(element(by.text('Sign a Celo transaction')))
+    await waitFor(element(by.text('WalletConnectV1 E2E would like to sign a Celo transaction.')))
```
kathaypacific, line 288: in a future PR i might change these to reference test id's instead :/ it's annoying to remember to update these with copy changes

jeanregisser: Yes alternatively I wonder if we should load the locale by id and avoid adding testID everywhere. 🤔

src/analytics/Properties.tsx
```diff
@@ -1070,6 +1070,11 @@ interface WalletConnectProperties {
   [WalletConnectEvents.wc_request_deny_error]: WalletConnectRequestDenyProperties & {
     error: string
   }
+
+  [WalletConnectEvents.wc_copy_request_payload]: {
```
jeanregisser, line 1074: I suggest we reuse `WalletConnectDefaultProperties` to follow the other events.

src/dappkit/DappKitAccountScreen.tsx
```diff
@@ -78,21 +79,20 @@ const DappKitAccountScreen = ({ route }: Props) => {
...
       />
-    </View>
+    </SafeAreaView>
   )
 }
 
 const styles = StyleSheet.create({
   container: {
     padding: Spacing.Thick24,
+    paddingBottom: 0, // SafeAreaView already adds enough space here
```
jeanregisser, line 94: What about on devices with no home bar?

src/dappkit/DappKitSignTxScreen.tsx
```diff
@@ -61,55 +68,57 @@ const DappKitSignTxScreen = ({ route }: Props) => {
...
       </RequestContent>
-    </View>
+    </SafeAreaView>
   )
 }
 
 const styles = StyleSheet.create({
   container: {
     padding: Spacing.Thick24,
+    paddingBottom: 0, // SafeAreaView already adds enough space here
```
jeanregisser, line 104: same question here.

src/walletConnect/screens/WalletConnectRequest.tsx
```diff
@@ -41,15 +46,15 @@ function WalletConnectRequest({ route: { params } }: Props) {
...
       {params.type === WalletConnectRequestType.TimeOut && <ConnectionTimedOut />}
-    </View>
+    </SafeAreaView>
   )
 }
 
 const styles = StyleSheet.create({
   container: {
     padding: Spacing.Thick24,
+    paddingBottom: 0, // SafeAreaView already adds enough space here
```
jeanregisser, line 56: ditto

jeanregisser (approved):
Awesome! 🚀 

Already approved with a couple of very small comments.

src/analytics/Properties.tsx
kathaypacific: haha 🙈 i avoided doing this in the first place because it is hard but i added it now - we were previously using a saga to get all the default props so i abstracted what i could from there to share with this component. i also updated the name of the corresponding dappkit event. since there were non trivial changes maybe you want to take another look @jeanregisser ?

src/dappkit/DappKitSignTxScreen.tsx
kathaypacific: yeah good point - this PR has some quick hacks (this is one of them) because i had the dynamic heights on the bottom sheets but took that out to get reviewed separately. the 0 padding bottom is removed in the follow up https://github.com/valora-inc/wallet/pull/3412 so if you're okay with it i'll not spend time changing this here

locales/de/translation.json
```diff
@@ -1229,6 +1229,11 @@
     "message": "Du bist dabei, eine Anwendung zu öffnen, die nicht von Valora betrieben wird.\n\nValora gibt keine Zusicherungen zu dieser Anwendung ab. Die Nutzung erfolgt auf eigene Gefahr.",
     "button": "Gehe zu {{dappName}}"
   },
+  "dappsScreenInfoSheet": {
```
jeanregisser, line 1232: Is this PR up to date? we shouldn't touch non base locales.

src/analytics/Properties.tsx
```diff
@@ -1070,6 +1070,10 @@ interface WalletConnectProperties {
   [WalletConnectEvents.wc_request_deny_error]: WalletConnectRequestDenyProperties & {
     error: string
   }
+
+  [WalletConnectEvents.wc_copy_request_payload]: WalletConnectDefaultProperties & {
+    method: string
```
jeanregisser, line 1075: Isn't this the same as the `requestMethod` from `WalletConnectDefaultProperties`?

src/analytics/Properties.tsx
jeanregisser: great thanks, just did it.

src/dappkit/DappKitSignTxScreen.tsx
jeanregisser: Sounds good!

src/analytics/Properties.tsx
kathaypacific: oh `requestMethod` exists on `WalletConnectRequestDefaultProperties` but the type is `WalletConnectDefaultProperties` here....i can change it to WalletConnectRequestDefaultProperties but the extra params didn't seem that useful on first glance 😅

src/walletConnect/analytics.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export function getDefaultRequestTrackedPropertiesV2(
+  request: SignClientTypes.EventArguments['session_request']
+) {
+  const { id, params } = request
+
+  return {
+    requestChainId: params.chainId,
+    requestId: id,
+    requestJsonrpc: '2.0',
```
kathaypacific, line 61: this was previously params.request.params which is wrong, i don't think this field is so relevant in v2 anymore

src/analytics/Properties.tsx
kathaypacific: updated!

PR #3407, opened 2023-01-30
feat: display FiatConnect transfers differently in transaction feed

anthonydaoud:
### Description

Currently, the wallet display FiatConnect transfers in the same style as any other transfer in the transaction feed. This PR adds new functionality to check if a transfer is a FiatConnect transfer, and if so populates the transfer feed item so that it matches the style shown in this [figma (the Withdraw cUSD row)](https://www.figma.com/file/bwIY7peeykI8K4wTrXs4iN/Fiat-Connect?node-id=4264%3A49583&t=hvdNzAVgoJ4zkEkv-0).

fcTransactionFeedItem

### Test plan

Added unit test.

### Related issues

- ACT-578

### Backwards compatibility

Backwards Compatible.


src/transactions/transferFeedUtils.ts
```diff
@@ -301,12 +322,18 @@ export function useTransferFeedDetails(transfer: FeedTokenTransfer) {
...
+  let title, subtitle, customLocalAmount
 
   switch (type) {
     case TokenTransactionTypeV2.Sent: {
-      title = t('feedItemSentTitle', { displayName })
-      subtitle = t('feedItemSentInfo', { context: !comment ? 'noComment' : null, comment })
+      if (fcTransferDisplayInfo) {
+        title = fcTransferDisplayInfo.title
+        subtitle = fcTransferDisplayInfo.subtitle
+        customLocalAmount = fcTransferDisplayInfo.localAmount
```
cajubelt, line 332:
```suggestion
        ;({ title, subtitle, localAmount: customLocalAmount } = fcTransferDisplayInfo)
```
code golf

src/transactions/transferFeedUtils.ts
```diff
@@ -400,3 +427,52 @@ export function isTransferTransaction(
...
+  const cachedFiatAccountUses = useSelector(cachedFiatAccountUsesSelector)
+  const localCurrency = useSelector(getLocalCurrencyCode)
+  const account = useMemo(
+    () =>
+      cachedFiatAccountUses.find(
+        (account) => account.fiatAccountId === fcTransferDetails?.fiatAccountId
+      ),
+    [cachedFiatAccountUses, fcTransferDetails]
+  )
+  if (!account || !fcTransferDetails) {
```
cajubelt, line 445: seems like it would be more efficient to short circuit earlier, before going thru all the cached fiat account uses, if `fcTransferDetails` or `fcTransferDetails.fiatAccountId` was null

src/transactions/transferFeedUtils.ts
```diff
@@ -400,3 +427,52 @@ export function isTransferTransaction(
...
+function useFiatConnectTransferDisplayInfo({ amount, transactionHash }: TokenTransfer) {
+  const { t } = useTranslation()
+  const tokenInfo = useTokenInfo(amount.tokenAddress)
+  const fcTransferDetails = useSelector(getCachedFiatConnectTransferSelector(transactionHash))
+  const cachedFiatAccountUses = useSelector(cachedFiatAccountUsesSelector)
+  const localCurrency = useSelector(getLocalCurrencyCode)
+  const account = useMemo(
+    () =>
+      cachedFiatAccountUses.find(
+        (account) => account.fiatAccountId === fcTransferDetails?.fiatAccountId
```
cajubelt, line 441: "account" shadows a variable name from outer scope, not ideal

cajubelt, line 441:
```suggestion
        ({ fiatAccountId }) => fiatAccountId === fcTransferDetails?.fiatAccountId
```

src/transactions/transferFeedUtils.ts
```diff
@@ -400,3 +427,52 @@ export function isTransferTransaction(
...
+  if (!account || !fcTransferDetails) {
+    return
+  }
+
+  const receive = Number(fcTransferDetails.quote.fiatAmount)
+  const total = Number(amount.value)
+  const exchangeRate = String(receive / total)
+  const sign = fcTransferDetails.quote.transferType === TransferType.TransferOut ? -1 : 1
+  const localAmount: LocalAmount = {
+    value: new BigNumber(fcTransferDetails.quote.fiatAmount).multipliedBy(sign),
```
cajubelt, line 454:
```suggestion
    value: new BigNumber(receive).multipliedBy(sign),
```

src/transactions/transferFeedUtils.ts
```diff
@@ -400,3 +427,52 @@ export function isTransferTransaction(
...
+    [cachedFiatAccountUses, fcTransferDetails]
+  )
+  if (!account || !fcTransferDetails) {
+    return
+  }
+
+  const receive = Number(fcTransferDetails.quote.fiatAmount)
+  const total = Number(amount.value)
+  const exchangeRate = String(receive / total)
+  const sign = fcTransferDetails.quote.transferType === TransferType.TransferOut ? -1 : 1
```
cajubelt, line 452: if you're not sure if the user is receiving or sending, it seems weird to call the variable assigned to fiat amount "receive"

src/transactions/transferFeedUtils.ts
cajubelt: also why are you using a bignumber here but not for exchangeRate?

src/transactions/transferFeedUtils.ts
```diff
@@ -400,3 +427,52 @@ export function isTransferTransaction(
...
+  let subtitle: string
+  switch (account.fiatAccountType) {
+    case FiatAccountType.BankAccount:
+      subtitle = t('feedItemFcTransferBankAccount')
+      break
+    case FiatAccountType.MobileMoney:
+      subtitle = t('feedItemFcTransferMobileMoney')
+      break
+    default:
+      // should never happen
```
cajubelt, line 468: this actually seems relatively likely to happen. In any case, is it really worth erroring out? Seems like a debug log and returning null would be sufficient

src/transactions/transferFeedUtils.ts
cajubelt: I'd probly just call it fiatAmount

src/fiatconnect/index.ts
```diff
@@ -41,6 +41,10 @@ export function convertToFiatConnectFiatCurrency(
   return FiatType[localCurrency as unknown as FiatType]
 }
 
+export function convertToLocalCurrency(fiatType: FiatType): LocalCurrencyCode | undefined {
+  return LocalCurrencyCode[fiatType as unknown as LocalCurrencyCode]
```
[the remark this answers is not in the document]
anthonydaoud: Added the currency codes/symbols in

src/transactions/transferFeedUtils.ts
anthonydaoud: That's nicer!

src/transactions/transferFeedUtils.ts
anthonydaoud: Updated

src/transactions/transferFeedUtils.ts
anthonydaoud: Added the debug log

src/transactions/transferFeedUtils.ts
anthonydaoud: That is how the `LocalAmount` type is [defined](https://github.com/valora-inc/wallet/blob/266b2b56658e3d2f5d740d9f6588be19d564f87e/src/transactions/types.ts#L113), which is used [here](https://github.com/valora-inc/wallet/blob/266b2b56658e3d2f5d740d9f6588be19d564f87e/src/transactions/feed/TransferFeedItem.tsx#L61) to render the local currency in each `TransferFeedItem`. I'm not sure why `LocalAmount` was defined with exchangeRate as a string

src/fiatconnect/index.ts
```diff
@@ -41,6 +41,40 @@ export function convertToFiatConnectFiatCurrency(
   return FiatType[localCurrency as unknown as FiatType]
 }
 
+export function convertToLocalCurrency(fiatType: FiatType): LocalCurrencyCode {
+  return {
```
jophish, line 45: nit: pull this out into a const?

src/transactions/transferFeedUtils.ts
```diff
@@ -400,3 +426,53 @@ export function isTransferTransaction(
...
+
+// Note: This hook is tested from src/transactions/feed/TransferFeedItem.test.ts
+function useFiatConnectTransferDisplayInfo({ amount, transactionHash }: TokenTransfer) {
+  const { t } = useTranslation()
+  const tokenInfo = useTokenInfo(amount.tokenAddress)
+  const fcTransferDetails = useSelector(getCachedFiatConnectTransferSelector(transactionHash))
+  const cachedFiatAccountUses = useSelector(cachedFiatAccountUsesSelector)
+  const account = useMemo(
+    () =>
+      fcTransferDetails
```
jophish, line 438: I think this should work?

```suggestion
      cachedFiatAccountUses.find(
        ({ fiatAccountId }) => fiatAccountId === fcTransferDetails?.fiatAccountId
      ),
```

src/transactions/transferFeedUtils.ts
```diff
@@ -400,3 +426,53 @@ export function isTransferTransaction(
...
+          )
+        : undefined,
+    [cachedFiatAccountUses, fcTransferDetails]
+  )
+  if (!account || !fcTransferDetails) {
+    return
+  }
+
+  const fiatAmount = Number(fcTransferDetails.quote.fiatAmount)
+  const total = Number(amount.value)
```
jophish, line 450: `total` is a bit vague; maybe `cryptoAmount`?

src/transactions/transferFeedUtils.ts
```diff
@@ -400,3 +426,53 @@ export function isTransferTransaction(
...
+    case FiatAccountType.MobileMoney:
+      subtitle = t('feedItemFcTransferMobileMoney')
+      break
+    default:
+      Logger.debug(TAG, 'useFiatConnectTransferDisplayInfo received an unsupported FiatAccountType')
+      return
+  }
+
+  return {
+    title: t('feedItemFcTransferWithdraw', { crypto: fcTransferDetails.quote.cryptoType }),
```
jophish, line 473: This currently hardcodes the title for withdraws, but we do in fact check the transfer type above to determine the sign of the amount displayed. Since this doesn't work for deposits (cash-ins) atm, should we log an error and return `null` if `fcTransferDetails.quote.transferType === TransferType.TransferIn`?

src/transactions/transferFeedUtils.ts
anthonydaoud: Yeah I had it like that previously, but it [was suggested](https://github.com/valora-inc/wallet/pull/3407#discussion_r1092494511) that I terminate early so that a `find` is not done unnecessarily.

src/transactions/transferFeedUtils.ts
anthonydaoud: I added the `transferType` as another check that will cause an early return with `null`, so now the code only supports transfers out.

src/fiatconnect/consts.ts
```diff
@@ -0,0 +1,42 @@
...
+  [FiatType.UGX]: LocalCurrencyCode.UGX,
+  [FiatType.USD]: LocalCurrencyCode.USD,
+  [FiatType.UYU]: LocalCurrencyCode.UYU,
+  [FiatType.VES]: LocalCurrencyCode.VES,
+  [FiatType.XAF]: LocalCurrencyCode.XAF,
+  [FiatType.XOF]: LocalCurrencyCode.XOF,
+  [FiatType.ZAR]: LocalCurrencyCode.ZAR,
+}
+
+export const WALLET_CRYPTO_TO_FIATCONNECT_CRYPTO = {
```
cajubelt, line 37:
```suggestion
export const WALLET_CRYPTO_TO_FIATCONNECT_CRYPTO: Record<CiCoCurrency, CryptoType> = {
```
an explicit type is important to add here because it gives us compile-time guarantees that if a CiCoCurrency is added, the set of keys in this object must be exhaustive, else an error will be thrown

src/fiatconnect/consts.ts
```diff
@@ -0,0 +1,42 @@
+import { CryptoType, FiatType } from '@fiatconnect/fiatconnect-types'
+import { LocalCurrencyCode } from 'src/localCurrency/consts'
+import { CiCoCurrency } from 'src/utils/currencies'
+
+export const FIATCONNECT_CURRENCY_TO_WALLET_CURRENCY = {
```
cajubelt, line 5:
```suggestion
export const FIATCONNECT_CURRENCY_TO_WALLET_CURRENCY: Record<FiatType, LocalCurrencyCode> = {
```
see next comment

src/fiatconnect/index.ts
```diff
@@ -41,13 +45,12 @@ export function convertToFiatConnectFiatCurrency(
   return FiatType[localCurrency as unknown as FiatType]
 }
 
+export function convertToLocalCurrency(fiatType: FiatType): LocalCurrencyCode {
```
cajubelt, line 48: this function doesn't seem necessary anymore

src/fiatconnect/index.ts
```diff
@@ -41,13 +45,12 @@ export function convertToFiatConnectFiatCurrency(
...
+}
+
 function convertToFiatConnectCryptoCurrency(cicoCurrency: CiCoCurrency): CryptoType {
-  return {
-    [CiCoCurrency.CELO]: CryptoType.CELO,
-    [CiCoCurrency.cEUR]: CryptoType.cEUR,
-    [CiCoCurrency.cUSD]: CryptoType.cUSD,
-    [CiCoCurrency.cREAL]: CryptoType.cREAL,
-  }[cicoCurrency]
+  return WALLET_CRYPTO_TO_FIATCONNECT_CRYPTO[cicoCurrency]
```
cajubelt, line 53: same here

src/transactions/transferFeedUtils.ts
```diff
@@ -400,3 +426,56 @@ export function isTransferTransaction(
...
   return (tx as TransferItemFragment).address !== undefined
 }
+
+// Note: This hook is tested from src/transactions/feed/TransferFeedItem.test.ts
+function useFiatConnectTransferDisplayInfo({ amount, transactionHash }: TokenTransfer) {
+  const { t } = useTranslation()
+  const tokenInfo = useTokenInfo(amount.tokenAddress)
+  const fcTransferDetails = useSelector(getCachedFiatConnectTransferSelector(transactionHash))
+  const cachedFiatAccountUses = useSelector(cachedFiatAccountUsesSelector)
+  const account = useMemo(
```
cajubelt, line 436: it appears that the "account" variable is serving two purposes now:
- storing information about the fiat account type that was used in a fiatconnect transfer
- acting as a flag for whether the transfer type is supported yet in the transaction feed, by being undefined if it was a transfer in, even if the user does actually have a fiat account id matching one in the fiatconnect transfer details

It makes the code more complex and hard to maintain to have one thing serve multiple purposes. I recommend finding another way to deal with transfers in.

src/transactions/transferFeedUtils.ts
```diff
@@ -400,3 +426,56 @@ export function isTransferTransaction(
...
+    () =>
+      fcTransferDetails?.quote.transferType === TransferType.TransferOut
```
cajubelt, line 438:
```suggestion
      fcTransferDetails?.fiatAccountId
```

src/transactions/transferFeedUtils.ts
```diff
@@ -400,3 +426,56 @@ export function isTransferTransaction(
...
+        ? cachedFiatAccountUses.find(
+            ({ fiatAccountId }) => fiatAccountId === fcTransferDetails.fiatAccountId
+          )
+        : undefined,
+    [cachedFiatAccountUses, fcTransferDetails]
+  )
+  if (!account || !fcTransferDetails) {
```
cajubelt, line 445:
```suggestion
  if (!account || !fcTransferDetails || (fcTransferDetails?.quote.transferType !== TransferType.TransferOut)) {
```
this is one thing you could do to deal with transfers in for now. I'm less concerned about efficiency than I am about simplicity and maintainability

src/transactions/transferFeedUtils.ts
```diff
@@ -400,3 +426,56 @@ export function isTransferTransaction(
...
+
+  const fiatAmount = Number(fcTransferDetails.quote.fiatAmount)
+  const cryptoAmount = Number(amount.value)
+  const exchangeRate = String(fiatAmount / cryptoAmount)
+  const localAmount: LocalAmount = {
+    //Negative sign because currently only withdraws are supported
+    value: -new BigNumber(fiatAmount),
+    currencyCode: convertToLocalCurrency(fcTransferDetails.quote.fiatType),
+    exchangeRate,
+  }
```
cajubelt, line 460:
```suggestion
  const fiatAmount = new BigNumber(fcTransferDetails.quote.fiatAmount)
  const localAmount: LocalAmount = {
    //Negative sign because currently only withdraws are supported
    value: fiatAmount.multipliedBy(-1),
    currencyCode: convertToLocalCurrency(fcTransferDetails.quote.fiatType),
    exchangeRate: fiatAmount.div(amount.value).toFixed(2),
  }
```
my previous comment was less about why BigNumber is used for "value", and more about why it *isn't* used for the arithmetic used in calculating exchangeRate. We should all be scared of [floating point errors in javascript](https://ellenaua.medium.com/floating-point-errors-in-javascript-node-js-21aadd897bf8)

src/fiatconnect/consts.ts
anthonydaoud: Makes sense, good catch

src/transactions/transferFeedUtils.ts
anthonydaoud: Ok cool, yeah I see how the changes you suggest make the code a bit cleaner/easier to reason about. Updated it

src/transactions/transferFeedUtils.ts
anthonydaoud: I see, yeah that is a safer way to calculate the exchangeRate

src/transactions/transferFeedUtils.ts
```diff
@@ -435,28 +435,30 @@ function useFiatConnectTransferDisplayInfo({ amount, transactionHash }: TokenTra
   const cachedFiatAccountUses = useSelector(cachedFiatAccountUsesSelector)
   const account = useMemo(
     () =>
-      fcTransferDetails?.quote.transferType === TransferType.TransferOut
+      fcTransferDetails?.quote
```
cajubelt, line 438:
```suggestion
      fcTransferDetails?.fiatAccountId
```

cajubelt: maybe I'm missing something but it seems more valuable to check if there's a fiatAccountId before iterating thru the cached fiat account uses, instead of checking if there's a quote, which seems unrelated to the task of this hook, which is to find the correct fiat account

cajubelt (commented):
looks close!

src/transactions/transferFeedUtils.ts
anthonydaoud: Good point, updated!

cajubelt (approved):

src/transactions/transferFeedUtils.ts
```diff
@@ -400,3 +426,58 @@ export function isTransferTransaction(
...
+            ({ fiatAccountId }) => fiatAccountId === fcTransferDetails.fiatAccountId
+          )
+        : undefined,
+    [cachedFiatAccountUses, fcTransferDetails]
+  )
+  if (
+    !account ||
+    !fcTransferDetails ||
+    fcTransferDetails.quote.transferType !== TransferType.TransferOut
+  ) {
```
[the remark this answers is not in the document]
anthonydaoud: Sure thing, separated them

PR #3412, opened 2023-01-31
feat: dapps bottom sheet dynamic height based on content size

kathaypacific:
### Description

We should have the bottom sheet height respond to the content size according to the new design. There's no out of the box way to do it yet using the bottom sheet navigation lib (but there is an[ open request](https://github.com/th3rdwave/react-navigation-bottom-sheet/issues/7) to support it, with an example of how someone else did it). The issue has been open for a while but there's a chance that it could be supported soon, so opted for a patch for now so that we can still use the `useBottomSheetDynamicSnapPoints` hook.

Not addressed in this PR:
- handling displays for large text (probably need to add scrolling)

### Test plan

With WalletConnect:

https://user-images.githubusercontent.com/20150449/215788229-e2eaf2fd-ec68-4b6f-8736-7ec72522628b.mp4

With DappKit:

https://user-images.githubusercontent.com/20150449/215788377-dca31a5d-8d54-42f2-91af-4c6204ba89f8.mp4

### Related issues

- Fixes RET-551

### Backwards compatibility

Y


patches/@th3rdwave+react-navigation-bottom-sheet+0.2.2.patch
```diff
@@ -0,0 +1,52 @@
...
+diff --git a/node_modules/@th3rdwave/react-navigation-bottom-sheet/lib/typescript/types.d.ts b/node_modules/@th3rdwave/react-navigation-bottom-sheet/lib/typescript/types.d.ts
+index e03cf4e..c040f23 100644
+--- a/node_modules/@th3rdwave/react-navigation-bottom-sheet/lib/typescript/types.d.ts
++++ b/node_modules/@th3rdwave/react-navigation-bottom-sheet/lib/typescript/types.d.ts
+@@ -30,7 +30,7 @@ export declare type BottomSheetNavigationOptions = Omit<BottomSheetProps, 'conta
+      * snapPoints={['%100']}
+      * @type Array<string | number>
+      */
+-    snapPoints?: Array<string | number>;
++    snapPoints?: Array<string | number> | { value: Array<string | number> };
```
kathaypacific, line 23: the actual type here should be `SharedValue` from the react-native-reanimated lib but this lib doesn't have this as a dependency 😅 so i took this shortcut

src/walletConnect/screens/WalletConnectRequest.tsx
```diff
@@ -53,13 +55,15 @@ function WalletConnectRequest({ route: { params } }: Props) {
 const styles = StyleSheet.create({
   container: {
     padding: Spacing.Thick24,
-    paddingBottom: 0, // SafeAreaView already adds enough space here
-    flex: 1,
+  },
+  loadingTimeoutContainer: {
+    alignItems: 'center',
+    minHeight: 370,
```
kathaypacific, line 61: because the WC connection flow is always connecting -> session request, and while connecting we don't know the height of the session request content, there is a bit of jumping in the height of the bottom sheet. i've attempted to reduce the jump with a min height here but there is still some movement when the session request is displayed as the sheet adjusts to the new content height....

before setting a min height:

https://user-images.githubusercontent.com/20150449/216942255-b0908e98-d71b-4087-8c79-6a0d92269702.mp4



after setting a min height:

https://user-images.githubusercontent.com/20150449/216942286-59cd20d0-9a7e-4888-aeae-07f3ac643a14.mp4

src/walletConnect/screens/RequestContent.tsx
```diff
@@ -111,54 +110,52 @@ function RequestContent({
...
-        )}
-
-        {children}
-      </View>
+        </View>
+      )}
+      <Text style={styles.header} testID={`${testId}Header`}>
+        {title}
+      </Text>
+      {description && <Text style={styles.description}>{description}</Text>}
```
jeanregisser, line 136:
```suggestion
      {!!description && <Text style={styles.description}>{description}</Text>}
```

src/walletConnect/screens/WalletConnectRequest.tsx
jeanregisser: Great! 
I wonder if the min height should be smaller, this way it's more likely it goes from small -> actual height which is bigger so it looks slightly nicer?

Alternatively, should we use a skeleton placeholder? out of scope of this PR of course, we could discuss with Reed.

patches/@th3rdwave+react-navigation-bottom-sheet+0.2.2.patch
```diff
@@ -0,0 +1,52 @@
+diff --git a/node_modules/@th3rdwave/react-navigation-bottom-sheet/lib/commonjs/BottomSheetView.js b/node_modules/@th3rdwave/react-navigation-bottom-sheet/lib/commonjs/BottomSheetView.js
+index 04a1ee9..8008098 100644
+--- a/node_modules/@th3rdwave/react-navigation-bottom-sheet/lib/commonjs/BottomSheetView.js
++++ b/node_modules/@th3rdwave/react-navigation-bottom-sheet/lib/commonjs/BottomSheetView.js
+@@ -105,7 +105,7 @@ function BottomSheetView(_ref2) {
+       key: route.key // Make sure index is in range, it could be out if snapToIndex is persisted
+       // and snapPoints is changed.
+       ,
+-      index: Math.min((_ref3 = (_route$snapToIndex = route.snapToIndex) !== null && _route$snapToIndex !== void 0 ? _route$snapToIndex : index) !== null && _ref3 !== void 0 ? _ref3 : 0, snapPoints.length - 1),
++      index: Math.min((_ref3 = (_route$snapToIndex = route.snapToIndex) !== null && _route$snapToIndex !== void 0 ? _route$snapToIndex : index) !== null && _ref3 !== void 0 ? _ref3 : 0, Array.isArray(snapPoints) ? snapPoints.length - 1 : snapPoints.value.length - 1),
```
jeanregisser, line 10: Is Metro also loading this one?

src/walletConnect/screens/WalletConnectRequest.tsx
kathaypacific: hmm, i'm actually not sure that this is a straightforward problem to solve :( 

i haven't found the exact parts of the bottom sheet lib that cause the problem but from observation i think that the animated content height / snap points managed by the bottom sheet lib are long living (they are using `useSharedValue`). after a bottom sheet is mounted / measured, that value persists and is the first value that is used for the next bottom sheet. 

in the video above, there are 3 heights - the first one is from the previous connection bottom sheet, the second one is the minHeight (370) that is set on the loading state, the third one is the height of the sheet with the session request content. i need to spend more time to confirm this theory though

kathaypacific: i tried adding some logs to the `useBottomSheetDynamicSnapPoints` hook and i think my theory is right - after a bottom sheet is dismissed, the `animatedContentHeight` doesn't get reset to 0 so the next sheets use the last known value. i think this is just because we've got `useBottomSheetDynamicSnapPoints` at the very top level. idk if this is a problem worth solving yet 😅 i think the experience is pretty fine

jeanregisser: oh yeah right, I feared this would happen with our approach. 
Not the end of the world like you said, would be nice to have a sheet independent solution in the lib itself.

PR #3413, opened 2023-01-31
feat(qr): Parameterize QR tab screen by styling and data encoding

jophish:
### Description

For ACT-550.

Parameterizes the existing QR code screen by both QR code styling and data encoding. Defaults to (current) deeplink encoding with current designs.

N.B., while this PR _does_ enable us to parameterize the QR code screen via route params (which is useful), I believe it would be easier/cleaner to determine the value of the parameters directly in the `QRNavigator` component when we move to set up the experiment for the following reasons:
* There are many places besides the Home Screen that allow the user to navigate to this QR page (the `QRNavigator`, in particular); if we passed experiment params via route params, this would get very messy (assuming we want consistent experiences across any way a user might arrive at the `QRNavigator` screen).
* The `QRNavigator` is a _tab_-based screen, and you can navigate to _either_ the "Scan" tab, or the "Code" tab; when navigating to the "Scan" tab (as is done in several places), it is not immediately obvious to me how to _also_ pass potential parameters to the "Code" tab, such that they will be honored by the "Code" tab if the user switches to it. AFAICT, you can only pass parameters to the tab you're directly navigating to.

Instead, we can just determine the desired parameters in the `QRNavigator` screen and pass them to the new `QRCodePicker` component in place of the props; the content of this PR puts us in a place to do this with relative ease.

### Test plan

Unit and manual tested, tested for regression on the `ExchangeQR` screen as well. Videos below. To differentiate the encoding styles, compare the actual content/size of the QR code data.

#### `ExchangeQR` Screen regression test

(The QR button on the home screen will never navigate the user to the `ExchangeQR` screen, I just hooked it up for testing purposes.)

https://user-images.githubusercontent.com/569401/215821516-bb92ea9a-266a-4847-a4ee-858cd4bf522e.mp4

#### New Styling, Address Encoding

https://user-images.githubusercontent.com/569401/215821585-e98ed0d2-c139-47ff-ac03-eb80375622ed.mp4

#### New Styling, Deeplink Encoding

https://user-images.githubusercontent.com/569401/215821682-646e1d87-ffd0-4e41-928a-0dbf98ee49d2.mp4

#### Legacy Styling, Address Encoding

https://user-images.githubusercontent.com/569401/215821745-b7139be9-8973-430e-890c-59a9d916fcd0.mp4

#### Legacy Styling, Deeplink Encoding

https://user-images.githubusercontent.com/569401/215821760-c8879d1a-f373-4063-b645-ff719caa2d31.mp4

### Related issues

- Fixes ACT-550

### Backwards compatibility

Yes!


src/qrcode/NewQRCode.tsx
```diff
@@ -0,0 +1,140 @@
...
+  qrSvgRef: React.MutableRefObject<SVG>
+  exchanges: ExternalExchangeProvider[]
+  dataType: QRCodeDataType
+  onCloseBottomSheet?: () => void
+  onPressCopy?: () => void
+  onPressInfo?: () => void
+  onPressExchange?: (exchange: ExternalExchangeProvider) => void
+}
+
+export default function NewQRCodeDisplay(props: Props) {
```
jophish, line 30: This file largely ported over from `ExchangeQR.tsx` and generalized a bit.

src/qrcode/QRTabBar.tsx
```diff
@@ -46,6 +46,10 @@ export default function QRTabBar({ state, descriptors, navigation, position, qrS
     dispatch(shareQRCode(qrSvgRef.current))
   }
 
+  const getParams = (route: string) => {
```
jophish, line 49: Previously, params were discarded when a user switched tab and switched back again.

src/qrcode/schema.ts
```diff
@@ -42,6 +42,11 @@ export enum QRCodeDataType {
   Address = 'Address',
 }
 
+export enum QRCodeStyle {
+  New = 'New',
```
cajubelt, line 46: feel like it's a slippery slope from this to NewModuleRedesigned. How about a more descriptive version name, like `Jan2023QRCodeStyle`? or `Jan2023Version`?

cajubelt (approved):

PR #3414, opened 2023-02-01
feat: display dapps disclaimer on dapps bottom sheet

kathaypacific:
### Description

This PR adds the dapps disclaimer to the dapps bottom sheet for the variant group of the dapps disclaimer experiment.

### Test plan

Added a unit test for the happy flow, that seemed like the most important one. Screenshot:

Simulator Screen Shot - iPhone 14 Pro Max - 2023-02-01 at 13 57 29

### Related issues

- Fixes RET-551

### Backwards compatibility

Y

### DIFF
--- a/locales/base/translation.json
+++ b/locales/base/translation.json
@@ -1248,6 +1248,8 @@
     "buttonLabel": "Tell me more"
   },
   "dappsDisclaimerAllDapps": "{{appName}} is not affiliated with these dapps and makes no representations or claims in using them. Please acknowledge the risk before using a dapp by reviewing the user agreement or privacy policy.",
+  "dappsDisclaimerSingleDapp": "{{appName}} is not affiliated with this dapp and makes no representations in using it. Always review a dapps user agreement or privacy policy.",
+  "dappsDisclaimerUnlistedDapp": "This dapp is not listed in {{appName}}. {{appName}} is not affiliated with this dapp and makes no representations in using it. Always review a dapps user agreement or privacy policy.",
   "recentlyUsedDapps": "Recently Used Dapps",
   "favoritedDapps": "Favorited Dapps",
   "allDapps": "All Dapps",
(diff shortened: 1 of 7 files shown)


src/walletConnect/screens/DappsDisclaimer.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  if (dappsMinimalDisclaimerEnabled) {
+    return (
+      <Text style={styles.dappNotListedDisclaimer}>
+        {isDappListed ? t('dappsDisclaimerSingleDapp') : t('dappsDisclaimerUnlistedDapp')}
+      </Text>
+    )
+  }
+
+  if (!isDappListed && dappConnectInfo === DappConnectInfo.Basic) {
```
kathaypacific, line 28: i know this is not a 1:1 change as before, we used to also show this disclaimer on the dapp connect bottom sheet but i think the direction we want to move to is to display disclaimers on "risky" actions only, and we had discussed how connecting to a dapp is very low risk

locales/base/translation.json
```diff
@@ -1248,6 +1248,8 @@
     "buttonLabel": "Tell me more"
   },
   "dappsDisclaimerAllDapps": "{{appName}} is not affiliated with these dapps and makes no representations or claims in using them. Please acknowledge the risk before using a dapp by reviewing the user agreement or privacy policy.",
+  "dappsDisclaimerSingleDapp": "{{appName}} is not affiliated with this dapp and makes no representations in using it. Always review a dapps user agreement or privacy policy.",
+  "dappsDisclaimerUnlistedDapp": "This dapp is not listed in {{appName}}. {{appName}} is not affiliated with this dapp and makes no representations in using it. Always review a dapps user agreement or privacy policy.",
```
M-shin, line 1252: Is the "dapps" not supposed to have an apostrophe?

M-shin (approved):

locales/base/translation.json
kathaypacific: nice catch!! i think it should have one indeed..i updated :D

PR #3416, opened 2023-02-01
fix: walletconnect should deny requests on error

kathaypacific:
### Description

After a user accepts a WC action request, we trigger a saga to execute the approval. If the saga fails, we should trigger the flow to reject the action so that the pending action does not persist in the redux store (to be handled in the next app sessions). the saga fails if the user is prompted for their pin and cancel the pincode screen. 

I _think_ that is the right thing to do, in the v1 saga i see we attempt to reject the request on the `connector` directly but do not handle the redux state. in the v2 saga we just do nothing.

### Test plan

Manually tested

### Related issues

- Fixes RET-551

### Backwards compatibility

Y

### DIFF
--- a/src/walletConnect/v1/saga.ts
+++ b/src/walletConnect/v1/saga.ts
@@ -207,7 +207,7 @@ function* acceptRequest(r: AcceptRequest) {
     SentryTransactionHub.finishTransaction(SentryTransaction.wallet_connect_transaction)
   } catch (e) {
     Logger.debug(TAG + '@acceptRequest', e.message)
-    connector?.rejectRequest({ id, jsonrpc, error: e.message })
+    yield put(denyRequestAction(peerId, request, e.message))
     ValoraAnalytics.track(WalletConnectEvents.wc_request_accept_error, {
       ...defaultTrackedProperties,
       error: e.message,
--- a/src/walletConnect/v2/saga.ts
+++ b/src/walletConnect/v2/saga.ts
@@ -419,6 +419,7 @@ function* handleAcceptRequest({ request }: AcceptRequest) {
       ...defaultTrackedProperties,
       error: e.message,
     })
+    yield put(denyRequest(request, getSdkError('USER_REJECTED')))
   }
 
   yield call(handlePendingStateOrNavigateBack)
--- a/src/walletConnect/walletConnect.ts
+++ b/src/walletConnect/walletConnect.ts
@@ -73,6 +73,7 @@ export function* handleLoadingWithTimeout(origin: WalletConnectPairingOrigin) {
     sessionRequestReceivedV1: take(ActionsV1.SESSION_V1),
     sessionRequestReceivedV2: take(ActionsV2.SESSION_PROPOSAL_V2),
     actionRequestReceived: take(ActionsV1.PAYLOAD_V1),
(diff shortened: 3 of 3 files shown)


src/walletConnect/v2/saga.ts
```diff
@@ -419,6 +419,7 @@ function* handleAcceptRequest({ request }: AcceptRequest) {
       ...defaultTrackedProperties,
       error: e.message,
     })
+    yield put(denyRequest(request, getSdkError('USER_REJECTED')))
```
kathaypacific, line 422: there could be many errors, i am not sure which one we should send back to the WC server but USER_REJECTED seems accurate enough?

src/walletConnect/v1/saga.ts
```diff
@@ -207,7 +207,7 @@ function* acceptRequest(r: AcceptRequest) {
     SentryTransactionHub.finishTransaction(SentryTransaction.wallet_connect_transaction)
   } catch (e) {
     Logger.debug(TAG + '@acceptRequest', e.message)
-    connector?.rejectRequest({ id, jsonrpc, error: e.message })
+    yield put(denyRequestAction(peerId, request, e.message))
```
jeanregisser, line 210: At this point the action is not anymore in Redux. It's cleared when `ACCEPT_REQUEST_V1` is dispatched. And after that this saga is triggered.
So dispatching `denyRequest` does nothing and I think the previous behavior was good.

https://github.com/valora-inc/wallet/blob/0d6bff725be9f9ea634f6b7328fe318068bf55bc/src/walletConnect/v1/reducer.ts#L66-L71

src/walletConnect/v2/saga.ts
jeanregisser: Similar here, but I think we should also directly reject too.

src/walletConnect/walletConnect.ts
```diff
@@ -73,6 +73,7 @@ export function* handleLoadingWithTimeout(origin: WalletConnectPairingOrigin) {
     sessionRequestReceivedV1: take(ActionsV1.SESSION_V1),
     sessionRequestReceivedV2: take(ActionsV2.SESSION_PROPOSAL_V2),
     actionRequestReceived: take(ActionsV1.PAYLOAD_V1),
+    actionRequestReceivedV2: take(ActionsV2.SESSION_PAYLOAD_V2),
```
jeanregisser, line 76: Should this be done separately in a different PR?

jeanregisser (commented):
Thanks, I'm a little confused by the changes. Maybe I missed something?

src/walletConnect/v1/saga.ts
jeanregisser: > So dispatching `denyRequest` does nothing and I think the previous behavior was good.

I meant in the Redux state. But it will be handled again by the saga. Which I think we shouldn't do as it will have other side effects, like capturing another analytics event for instance. Which is gonna be a bit misleading here. We shouldn't use the action for this case.

kathaypacific: yeah i think you're right, i came across a buggy looking scenario when i was testing the bottom sheets and thought this change fixed it but i cannot remember the exact details of the scenario anymore :( i'll close this PR for now and try to remember what i was trying to fix

PR #3421, opened 2023-02-03
fix: low allowance swap issues

MuckT:
### Description

We are calling 0x api twice, one to get the details of the swaps (the amounts and prices) and another to confirm everything is ready to execute the swap. This increases how long the swaps take to complete, two fetches instead of one and produces errors with the guaranteed price when the buy amount is set. To simplify we are now using the transaction object returned from the `/approveSwap` endpoint instead of subsequently calling the `/executeSwap` endpoint and using the swap transaction present in that response. This is more inline with the [0x API docs](https://docs.0x.org/0x-api-swap/guides/swap-tokens-with-0x-api) and will make future swap router implementations easier.

### Test plan

Tested locally on Android and saga tests updated.

### Related issues

- Fixes RET-572

### Backwards compatibility

Yes

### DIFF
--- a/src/swap/saga.test.ts
+++ b/src/swap/saga.test.ts
@@ -7,10 +7,8 @@ import { store } from 'src/redux/store'
 import { swapSubmitSaga } from 'src/swap/saga'
 import { swapApprove, swapError, swapExecute, swapPriceChange } from 'src/swap/slice'
 import { sendTransaction } from 'src/transactions/send'
-import { fetchWithTimeout } from 'src/utils/fetchWithTimeout'
 import Logger from 'src/utils/Logger'
 import { getContractKit } from 'src/web3/contracts'
-import networkConfig from 'src/web3/networkConfig'
 import { getConnectedUnlockedAccount } from 'src/web3/saga'
 import { walletAddressSelector } from 'src/web3/selectors'
 import { mockAccount } from 'test/values'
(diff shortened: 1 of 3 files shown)


src/swap/saga.test.ts
```diff
@@ -61,26 +59,13 @@ describe(swapSubmitSaga, () => {
...
-    ok: true,
-    json: () => {
-      return { validatedSwapTransaction: mockSwapTransaction }
-    },
-  }
-
-  const buyToken = '0xd8763cba276a3738e6de85b4b3bf5fded6d6ca73'
-  const sellToken = '0xe8537a3d056da446677b9e9d6c5db704eaab4787'
-  const buyAmount = '10000000000000000'
-  const executeSwapUri = `${networkConfig.executeSwapUrl}?buyToken=${buyToken}&sellToken=${sellToken}&buyAmount=${buyAmount}&userAddress=${mockAccount}`
```
kathaypacific, line 74: i think `executeSwapUrl` can also be removed from networkConfig

src/swap/saga.test.ts
```diff
@@ -104,29 +89,14 @@ describe(swapSubmitSaga, () => {
...
-
-  it('should set swap state correctly when response ok is false', async () => {
-    await expectSaga(swapSubmitSaga, mockSwap)
-      .withState(store.getState())
-      .provide([
-        [select(walletAddressSelector), mockAccount],
-        [call(getContractKit), contractKit],
-        [call(getConnectedUnlockedAccount), mockAccount],
-        [call(fetchWithTimeout, executeSwapUri), { ok: false }],
+        [call(getConnectedUnlockedAccount), throwError(new Error('fake error'))],
```
kathaypacific, line 92: nit: we're testing for a unlock account error, i wonder if it's more realistic to test for an error on `sendTransaction`?

src/swap/saga.test.ts
MuckT: Fixed in 8f39cc75470d2b6d2ccf4f8d83e66d187f639b4b & 2831d9770994f63a05b1a2bf66339f9818c09f7d!

src/swap/saga.test.ts
MuckT: Fixed in 5abe29b41bd423565ee1fd3a31106bd8842e1964!

PR #3423, opened 2023-02-03
feat: add alert that can be dismissed for the session

kathaypacific:
### Description

This PR allows the offline banner to be achknowledged and hidden for the rest of the session.

Approach:
- alerts redux now has a list of dismissed alerts, which will not be displayed again
- create a new type of hide action that dismisses the alerts for the whole session

This is a quick-ish fix, it doesn't address the issue of how we can let users know that their balances are up to date again but the current implementation also doesn't cater for this.

### Related issues

- Fixes RET-597


kathaypacific:
Closing in favour of #3424

PR #3424, opened 2023-02-04
chore: turn offline banner alert into non blocking toast

kathaypacific:
### Description

Aside from this blocking banner, we don't currently have another way to show users that their balance is not up to date. Until we have a way to let users know that they are back online, i think we have no choice but to leave some kind of messaging on the screen. I see instagram uses a toast for connectivity issues, which i quite like so thought i'd try it here.

I like it because it does not forfeit any existing functionality. Users are still able to trigger a refresh via the toast, and they can still navigate the app. The toast persists until the problem goes away, so that users can know that their balance is up to date again. I also like that it's a relatively unintrusive change + we can reuse an existing component.

There is some extra state management to preserve the toast dismiss animation. Since there is only one alert active at a time, there could be a scenario where there is an existing toast -> a new error banner is triggered. Without some extra state management, the toast would display the new error banner message while animating off the screen. This is a very small issue of course but it was also simple to fix with an extra state variable. Example with slow animation:

https://user-images.githubusercontent.com/20150449/216957689-5ee1c979-5422-479d-a5ea-6e330d572fb3.mp4

### Test plan

https://user-images.githubusercontent.com/20150449/216770235-3b2859d2-57a3-4a7a-8fbf-2046b33710b9.mp4

### Related issues

- Fixes RET-597

### Backwards compatibility

Y


src/tokens/saga.ts
```diff
@@ -253,25 +251,27 @@ interface UserBalancesResponse {
...
-    variables: {
-      address,
-    },
-    fetchPolicy: 'network-only',
-    errorPolicy: 'all',
-  })
-
-  return response.data.userBalances.balances
+  await sleep(100)
+  throw new Error('test')
```
kathaypacific, line 255: as with the other PR, this is for testing! in case you want to check out the branch and play with the error. i'm faking a refetch + fail after 1 s

jeanregisser (commented):
Awesome! I think I prefer this approach for now (compared to https://github.com/valora-inc/wallet/pull/3423) cause it displays so much nicer and doesn't need more state 😄

kathaypacific:
> Awesome! I think I prefer this approach for now (compared to #3423) cause it looks so much nicer and doesn't need more state 😄

yay me too! glad you like it too, i tidied up the PR so that it's ready for review again. i will also touch base with Reed later on today so that he is aware :)

jeanregisser (approved):
🚀

PR #3425, opened 2023-02-07
feat: use guaranteed price in swaps based on remote config

MuckT:
### Description

Sets up the use of guaranteed price from 0x based on the remote config to allow an a/b test to see how using the value impacts the percent of swaps completed.

### Test plan

Tested locally on an iOS mainnet build.

### Related issues

- Fixes RET-573
### Backwards compatibility

Yes


jeanregisser (approved):
Great! 👍

PR #3426, opened 2023-02-07
chore: ensure that dapps bottom sheet handles font scaling

kathaypacific:
### Description

This PR ensures that the dapps content is visible if users have small screens / large fonts. If the bottom sheet content exceeds the screen height, allow the content to scroll.

### Test plan

https://user-images.githubusercontent.com/20150449/217264210-ed63dd23-6448-4547-a8b9-e589f8d053a6.mp4

### Related issues

- Fixes RET-511

### Backwards compatibility

Y


src/navigator/constants.ts
```diff
@@ -0,0 +1 @@
+export const BOTTOM_SHEET_DEFAULT_HANDLE_HEIGHT = 24
```
kathaypacific, line 1: idk where to put this guy! the bottom sheet lib doesn't export this :(

src/dappkit/DappKitAccountScreen.tsx
```diff
@@ -28,6 +30,9 @@ type Props = BottomSheetScreenProps<StackParamList, Screens.DappKitAccountScreen
 const DappKitAccountScreen = ({ route, handleContentLayout }: Props) => {
   const { dappKitRequest } = route.params
 
+  const [scrollEnabled, setScrollEnabled] = useState(false)
+  const { height } = useSafeAreaFrame()
```
jeanregisser, line 34: Shouldn't we include the top area height with the notch?

Otherwise it goes too high from the video taken:
Screenshot 2023-02-07 at 19 12 29

src/dappkit/DappKitAccountScreen.tsx
```diff
@@ -59,32 +64,44 @@ const DappKitAccountScreen = ({ route, handleContentLayout }: Props) => {
...
-            value: account,
-          },
-        ]}
-        testId="DappKitSessionRequest"
-      />
-    </SafeAreaView>
+    <BottomSheetScrollView
+      style={{ maxHeight: height - BOTTOM_SHEET_DEFAULT_HANDLE_HEIGHT }}
+      scrollEnabled={scrollEnabled}
+    >
```
jeanregisser, line 78: nit: could we factorize this into our own BottomSheetScrollView to avoid repeating the same logic in different screens?

jeanregisser (approved):
🚀

src/dappkit/DappKitAccountScreen.tsx
kathaypacific: ah yes, i was lazy and thought about the dappkit components as temporary 🙈 i refactored into a separate component now

src/dappkit/DappKitAccountScreen.tsx
kathaypacific: yes you're right! i accounted for the notch now when scrolling, but ideally i wanted the bottom sheet handle not to overlap with the status bar / notch. i couldn't figure out a way to do this with dynamic snap points, it seems like this is the default behaviour when you set the snap point to 100% :( 

updated scrolling:

https://user-images.githubusercontent.com/20150449/217485829-693b1780-9442-4d57-9944-f575f14a882f.mp4


https://user-images.githubusercontent.com/20150449/217485842-2a988487-9748-4fbb-afda-35b95c6091ee.mp4

jeanregisser: Great! Nothing to do now, but if we wanted the top of the sheet to stop just before the top safe area, we could calculate the max percentage to use for the snap point. Knowing 100% is fullscreen and the height of the window + top inset.

kathaypacific: can you elaborate on how set the max percentage for the snap point? this is the part i am stuck on, since we just define a dynamic snap point and the `useBottomSheetDynamicSnapPoints` sets the actual value. i'd love to cap the maximum snap point, so keen if you have any ideas.

for now i'll merge this to not block the release but i'll follow up if we can limit the top of the sheet!

PR #3431, opened 2023-02-08
fix(fiatconnect): handle undefined feeInfo

jh2oman:
### Description

https://linear.app/valora/issue/ACT-625/timeboxed-bug-investigation-failed-fiatconnect-transfers-from-null

When we implemented transfers out, we did not handle feeEstimate very well. Specifically we did not handle cases where its undefined and didn't take advantage of other useful info like the `loading` or `error` fields. This PR uses the same model as the `Send` flow to be more defensive about feeEstimates in the fiatconnect transfer out flow.

| Review screen while Fee Loading | Review Screen while Fee Error |
:----------------------------------:|:---------------------------------:
| Simulator Screen Shot - iPhone 13 - 2023-02-08 at 11 01 48 | Simulator Screen Shot - iPhone 13 - 2023-02-08 at 11 01 36 |

### Test plan

Unit tests added

### Related issues

Addresses https://linear.app/valora/issue/ACT-625/timeboxed-bug-investigation-failed-fiatconnect-transfers-from-null


src/fiatconnect/ReviewScreen.test.tsx
```diff
@@ -244,7 +320,60 @@ describe('ReviewScreen', () => {
         'fiatConnectReviewScreen.paymentMethodVia, {"providerName":"Provider Two"}',
       ])
     })
-
+    it('disables the submit button if feeInfo is not available', () => {
```
cajubelt, line 323: i guess if this is what we do for the Send flow it's parity and worth merging for now, but this seems like kind of a confusing UX. To just have the submit button greyed out with no reason why.

cajubelt: maybe we should log when this happens so we can figure out how frequent it is and prioritize possible improvements

cajubelt (approved):

src/fiatconnect/ReviewScreen.test.tsx
jh2oman: We have some existing events that should cover that well. I'll put together a report

PR #3440, opened 2023-02-10
chore: upgrade react-native-gesture-handler to v2

kathaypacific:
### Description

We mostly don't need to use this dependency directly. I think that we were using some components from there because of auto import suggestions, in this PR i updated the touchables to be used from either the Touchable component or from react native directly. Also simplified some styles along the way.

This is a good one to upgrade because it removes the dependency on RNGestureHandlerEnabledRootView which was causing conflicts while upgrading react native.

### Test plan

Tested all the changes manually. Relying on CI to check the rest.

### Related issues

- Fixes RET-587

### Backwards compatibility

Y


e2e/src/usecases/OffRamps.js
```diff
@@ -48,13 +48,6 @@ export default offRamps = () => {
       await waitForElementId('RNWebView')
       await expect(element(by.text('Bidali'))).toBeVisible()
     })
-
-    // TODO(tomm) debug why this is failing on Android
-    it(':ios: Then should not be able to spend CELO', async () => {
```
kathaypacific, line 53: I removed this test because it was failing and i don't think it works as intended - i was surprised to find that it's not straighforward to test if a button is disabled using detox because detox only has access to native elements (not JS ones). https://github.com/wix/Detox/issues/3203

this was previously passing because `celoRadioButton` returns a native element which doesn't have any property called `enabled`. i think that e2e tests are better for testing flows rather than display state, so i didn't feel too bad about moving this test into the FiatExchangeCurrency unit test. wdyt @MuckT ?

jeanregisser (approved):
Awesome! 👍 
I'm curious what @MuckT thinks about the e2e test case.

PR #3441, opened 2023-02-10
chore: upgrade lottie

kathaypacific:
### Description

The latest version of lottie-react-native no longer requires us to directly depend on lottie-ios :D

### Test plan

Tested the loading spinner animation on ios and android manually to see that lottie still works on both platforms.

### Related issues

- Fixes RET-587

### Backwards compatibility

Y


jeanregisser (approved):
👍

PR #3454, opened 2023-02-20
chore: add remote config variables for supercharge v2

kathaypacific:
### Description

As the title, this PR just adds the two new remote config variables we need. I briefly thought about using statsig, but these variables will not be used for experimentation so it didn't seem like there was much benefit.

### Test plan

n/a

### Related issues

- Fixes RET-513

### Backwards compatibility

Y


jeanregisser (approved):
LGTM 👍

PR #3455, opened 2023-02-20
feat: add integration with supercharge v2 backend

kathaypacific:
### Description

This PR makes the wallet work with the v2 supercharge backend. Of note:
- add a new type for the v2 response
- add a new code path for the v2 stuff when fetching the rewards / displaying the rewards on the supercharge screen / claiming the rewards
- update unit tests

Not included in this PR: handling the error when the user is not verified - will make a follow up PR because this one is getting massive.

### Test plan

We should be able to:
- see and claim rewards as usual with the current system
- see and claim rewards in the new system
- see and claim rewards in the new system, when some of the rewards come from the old system

### Related issues

- Fixes RET-513

### Backwards compatibility

Y


src/consumerIncentives/saga.ts
```diff
@@ -126,18 +150,78 @@ function* claimReward(reward: SuperchargePendingReward, index: number, baseNonce
   }
 }
 
+function* claimRewardV2(reward: SuperchargePendingRewardV2, index: number, baseNonce: number) {
+  const { transaction, details } = reward
+
+  const superchargeRewardContractAddress = yield select(superchargeRewardContractAddressSelector)
+  if (superchargeRewardContractAddress !== transaction.to) {
```
dievazqu, line 157: Remember that in the V2 endpoint we will also return the pending rewards from the previous program. So the `transation.to` could be different in those cases.

src/consumerIncentives/saga.ts
```diff
@@ -126,18 +150,78 @@ function* claimReward(reward: SuperchargePendingReward, index: number, baseNonce
...
+    )
+  }
+
+  const kit: ContractKit = yield call(getContractKit)
+  const tokens: TokenBalances = yield select(tokensByAddressSelector)
+  const walletAddress: string = yield call(getConnectedUnlockedAccount)
+
+  Logger.debug(TAG, `Start claiming reward at index ${index}: ${JSON.stringify(reward)}`)
+
+  applyChainIdWorkaround(transaction, yield call([kit.connection, 'chainId']))
```
dievazqu, line 169: nit: I think this is not needed... I tested locally and I was able to send the transaction as it is returned by the API.

dievazqu (commented):
Looks good overall!
I left a comment that I think it should be addressed before merging.

src/consumerIncentives/saga.ts
kathaypacific: ahhhh such a good catch, thanks! damn, i think i'll need to add another remote config for this 🥲

src/consumerIncentives/saga.ts
kathaypacific: hmmmm maybe...i don't fully understand the scenarios where this workaround makes a differentce but there is a [long comment](https://github.com/valora-inc/wallet/blob/9ccce4df6a9ad4e836118e5961028b83942dc75d/src/web3/utils.ts#L64-L69) mentioning a `bug in TxParamsNormalizer` so i'm reluctant to remove it. it's used in other places in the same way, and it doesn't seem to cause issues (and it does do something, the `chainId` without this workaround is a number but after this workaround is a `0x` base 16 string)

src/consumerIncentives/saga.ts
kathaypacific: aaaaah damn. the `to` address is from the merkle contract 😵‍💫 i need to think about this some more

kathaypacific: added this! https://github.com/valora-inc/wallet/pull/3475

src/consumerIncentives/ConsumerIncentivesHomeScreen.tsx
```diff
@@ -157,24 +166,33 @@ function SuperchargingInfo() {
...
-  const rewardsByToken: { [tokenAddress: string]: BigNumber | undefined } = rewards.reduce(
-    (acc, reward) => {
+  const rewardsByToken: { [tokenAddress: string]: BigNumber } = {}
+
+  if (isSuperchargePendingRewardsV2(rewards)) {
+    rewards.forEach((reward) => {
+      const tokenAddress = reward.details.tokenAddress.toLowerCase()
+
+      rewardsByToken[tokenAddress] = (rewardsByToken[tokenAddress] || new BigNumber(0)).plus(
+        new BigNumber(reward.details.amount).div(WEI_PER_TOKEN)
```
jeanregisser, line 184: nit: This could be incorrect if we decided to reward with tokens not using 18 decimals.
I know this just copies what used to be done. Just noting it.

src/consumerIncentives/saga.ts
```diff
@@ -48,15 +59,27 @@ export function* claimRewardsSaga({ payload: rewards }: ReturnType<typeof claimR
...
       amount: string
       tokenAddress: string
       txHash: string
-    }[] = yield all(rewards.map((reward, index) => call(claimReward, reward, index, baseNonce)))
+    }[] = []
+
+    if (rewards.length > 0) {
+      const claimRewardFn = isSuperchargePendingRewardsV2(rewards) ? claimRewardV2 : claimReward
+      receivedRewards = yield all(
+        // @ts-ignore remove this ignore when upgrading to TS 4.2+ https://github.com/microsoft/TypeScript/issues/36390
```
jeanregisser, line 78: This way the compiler will tell us when we can remove it:
```suggestion
        // @ts-expect-error remove this when upgrading to TS 4.2+ https://github.com/microsoft/TypeScript/issues/36390
```

src/consumerIncentives/ConsumerIncentivesHomeScreen.tsx
```diff
@@ -223,7 +241,7 @@ export default function ConsumerIncentivesHomeScreen() {
 
   const onPressCTA = async () => {
     if (canClaimRewards) {
-      dispatch(claimRewards(superchargeRewards))
+      dispatch(claimRewards())
```
jeanregisser, line 244: Why not continuing to pass the rewards?
I find it better to avoid global state when possible.

src/consumerIncentives/saga.ts
jeanregisser: `applyChainIdWorkaround` is really needed to sign WC transactions where the `chainId` is stripped by WC.
But it should be harmless to keep it.

jeanregisser (approved):
Looking great! 👍

src/consumerIncentives/saga.ts
kathaypacific: thanks! i didn't know this, i updated

src/consumerIncentives/saga.ts
kathaypacific: ah thanks for the info! i didn't know, i removed it since it seems it's not needed

src/consumerIncentives/ConsumerIncentivesHomeScreen.tsx
kathaypacific: what prompted this is in the slice, the definition of `claimRewards` has an unused variable :/ i then saw that we are passing the rewards as an action payload only to use it in the saga. can you clarify why it's better to avoid global state? the saga already reads from the store, are we saying that it's better to reduce the number of selectors used?

it also seemed more error prone to respond to an action payload, when that payload comes from the store anyway so there might be mismatches depending on when the action was dispatched, e.g. we fetch the supercharge rewards on the wallet home screen, when users come into the supercharge screen another fetch is kicked off. if the responses are different for whatever reason, and the user presses claim with the payload of the first response before the second response comes back, then they'd claim rewards that are different to what we expect? idk how real this scenario is, but it felt like

jeanregisser: This is more in general. I know the supercharge code is handled quite globally for now in Redux.

The difference is that we're acting exactly on what the user did in the component: "claim this specific reward".
Which is a good thing. As it can't be impacted by another requests changing the value under its feet.

> supercharge rewards on the wallet home screen, when users come into the supercharge screen another fetch is kicked off. if the responses are different for whatever reason, and the user presses claim with the payload of the first response before the second response comes back, then they'd claim rewards that are different to what we expect? idk how real this scenario is, but it felt like

Yes and I'd expect the initial reward to be claimed.
But like you said, it's not a real scenario.

However if we take the example of WC requests or sending transactions. It helps to see possible problems from using values from the global state that can change anytime.

Scenario:
- App receives WC request to approve, stores it in single global variable
- User sees the screen to approve
- User clicks to approve
- Before the approve can be processed, app receives another requests and changes global variable
- Result is the wrong requests gets approved, which could have really bad consequences.

kathaypacific: thanks for the explanation! makes sense. like you said, i'm not sure that this scenario is likely to impact supercharge because of the global implementation but it sounds like this change isn't necessarily an improvement so i reverted it. it wasn't necessary anyway

PR #3456, opened 2023-02-21
fix: wrongful navigation to language screen on dev reload

kathaypacific:
### Description

Not really sure what caused this problem but it has been annoying. During dev, after saving a file, the app will often navigate to the language selection screen and then we'd need to force close the app to avoid going through the onboarding flow again. This was annoying.

When kicking off the root saga, we do have a blocking saga that waits for the redux store to be rehydrated before allowing all other sagas to be spawn. This saga only runs if a global variable `didHydrate` is false - between reloads, the previous value of this variable is preserved but the store is actually rehydrated. This causes all sagas to kick off using the un-hydrated (i.e. default) state. It is the `checkAccountExistenceSaga` and takes us into this onboarding flow.

The rehydrate happens pretty quickly and does not introduce a meaningful delay, we should always wait for this.

### Test plan

Tested manually

### Related issues

N/A

### Backwards compatibility

Y


src/redux/persist-helper.ts
```diff
@@ -18,10 +18,8 @@ export function getRehydratePayload(action: RehydrateAction, key: string) {
 
 // Note: Shouldn't be necessary except for in one place in the root saga
 export function* waitForRehydrate() {
```
jeanregisser, line 20: w00t w00t, thanks so much for looking into this!! ❤️ 

Since this helper is used in a single place and not supposed to be used anywhere else, 
what do you think of moving it to the root saga where it is used and removing the export?
This way we don't even need the global variable and this workaround.

jeanregisser (approved):

src/redux/persist-helper.ts
kathaypacific: thanks for the suggestion! i like it, it does simplify things a lot. i guess it was originally like this to keep all the persist logic together, but i think either way a 1 liner in the root saga to wait for the rehydrate is a little easy to miss so i like this simplified version much better!

jeanregisser (approved):
🎉

PR #3458, opened 2023-02-21
chore(onboarding): update styling on biometry screen

cajubelt:
### Description

Update styling on biometry screen to match designs [here](https://www.figma.com/file/NAp0jPTvldgl8AVajej3ns/Onboarding-Updates?node-id=1%3A3&t=gk6prOnR6PS6MH3t-0)

### Test plan

rendered on ios simulator:

Screenshot 2023-02-21 at 1 49 40 PM
Screenshot 2023-02-21 at 1 49 52 PM

android:

Screenshot 2023-02-21 at 5 01 23 PM
Screenshot 2023-02-21 at 5 01 55 PM
Screenshot 2023-02-21 at 5 03 21 PM

### Related issues

fixes [ACT-605](https://linear.app/valora/issue/ACT-605/update-biometry-screen)

### Backwards compatibility

must be included in release with other onboarding styling updates


src/onboarding/registration/EnableBiometry.tsx
```diff
@@ -166,32 +155,27 @@ EnableBiometry.navigationOptions = nuxNavigationOptions
...
   },
-  imageContainer: {
-    marginBottom: Spacing.Thick24,
-  },
-  description: {
-    ...fontStyles.regular,
-    textAlign: 'center',
-    marginBottom: Spacing.Thick24,
+  biometryButton: {
+    width: '100%',
```
jh2oman, line 167: I suspect that this is unnecessary b/c of the `size={BtnSizes.FULL}` on line 139

src/onboarding/registration/EnableBiometry.tsx
```diff
@@ -67,11 +65,7 @@ export default function EnableBiometry({ navigation }: Props) {
     navigation.setOptions({
```
jh2oman, line 65: The back arrow and skip button need to have their colors changed

src/onboarding/registration/EnableBiometry.tsx
cajubelt: looks like this without it
Screenshot 2023-02-21 at 1 10 14 PM

cajubelt: i futzed with this for a while. I think it's because of [this line](https://github.com/valora-inc/wallet/blob/caefb3dbb968e7da5c5466a7e69ad9d5012adfea/src/components/Button.tsx#L199) in the Button component. When I added a View wrapping the button and made its flexDirection "row", it sort of worked, but the width was incorrect. Which led me to using 100% width, at which point it was possible to remove the wrapper with no effect (somehow...)

cajubelt: the weird thing is that in other onboarding steps, BtnSizes.FULL is sufficient to get the full screen appearance that we want here. so I don't really get it.

src/onboarding/registration/EnableBiometry.tsx
cajubelt: whoops. thanks

jh2oman (approved):

src/onboarding/registration/EnableBiometry.tsx
jh2oman: weird, but if it works 👍

PR #3460, opened 2023-02-22
chore: add remote config variables for dapps explorer search and filter

MuckT:
### Description

Adds the remote config variables `dappsFilterEnabled` and `dappsSearchEnabled`.

### Test plan
 
Tested locally with unit tests and in CI.

### Related issues

- RET-581

### Backwards compatibility

Yes


kathaypacific (approved):
🚀

PR #3461, opened 2023-02-22
chore: rename existing dapps explorer as legacy and update test ids

MuckT:
### Description

Renames the existing dapps explorer components that will need changes to support dapp filtering as legacy.

### Test plan

Tested locally and in CI.

### Related issues

- RET-581

### Backwards compatibility

Yes


e2e/src/utils/dappList.js
```diff
@@ -36,7 +36,7 @@ export async function scrollToDapp(dappIndex = 0) {
   try {
     await waitFor(element(by.id('DappCard')).atIndex(dappIndex))
       .toBeVisible(100)
-      .whileElement(by.id('DAppExplorerScreen/DappsList'))
+      .whileElement(by.id('DAppsExplorerScreen/DappsList'))
```
M-shin, line 39: should this be `DAppsExplorerScreenLegacy`?

M-shin (approved):
+1 with comment

PR #3463, opened 2023-02-22
test(e2e): improve error message and add timeout param

MuckT:
### Description

A small improvement for `waitForElementId` when the utils functions fails to find a matching element. It can be difficult to determine what element was not found in CI without downloading and reviewing e2e artifacts; this change should make it clear what test id failed to be found.

```
    Element with testID 'DAppsExplorerScreenTest/HelpIcon' not found

      130 |       .withTimeout(timeout)
      131 |   } catch {
    > 132 |     throw new Error(`Element with testID '${elementId}' not found`)
          |           ^
      133 |   }
      134 | }
```

### Test plan

Tested locally and in CI.

### Related issues

N/A

### Backwards compatibility

Yes


e2e/src/utils/utils.js
```diff
@@ -118,11 +118,19 @@ export async function isElementVisible(elementId, index) {
...
+ * @param {number} timeout timeout in milliseconds
+ */
+export async function waitForElementId(elementId, timeout = 10 * 1000) {
+  try {
+    await waitFor(element(by.id(elementId)))
+      .toBeVisible()
+      .withTimeout(timeout)
+  } catch {
+    throw new Error(`Element with testID '${elementId}' not found`)
+  }
```
jeanregisser, line 133: We can attach the underlying error. Could be useful in some cases.
https://twitter.com/housecor/status/1608795552644476928

```suggestion
  } catch (err) {
    throw new Error(`Element with testID '${elementId}' not found`, { cause: err })
  }
```

jeanregisser (approved):
Awesome!! 👍

PR #3471, opened 2023-02-27
chore(supercharge): bypass fetching supercharge v2 rewards for non CPV users

kathaypacific:
### Description

The current supercharge service returns no rewards for an unverified user, but the new supercharge service returns a 400 error. This PR:
- prevent fetching supercharge rewards if user is not verified (without this, all users who are not verified receive an error banner)
- only surface an error if the wallet thinks the user is verified but the backend does not
- update the error logs to show the response, to help with debugging when users contact support

### Test plan

n/a

### Related issues

- Fixes RET-513

### Backwards compatibility

Y


jeanregisser (approved):
🚀

PR #3474, opened 2023-02-27
feat: dapp filter screens

MuckT:
### Description

- Adds the `DAppsExplorerScreenFilters.tsx`,  `DappFilterChip.tsx`, `FavoriteDappsSection.tsx` and `NoResults.tsx`.
- Associated tests for `DAppsExplorerScreenFilters.tsx`, `FavoriteDappsSection.tsx`.
- Modifies the `dappCategoriesByIdSelector` to `dappCategoriesSelector` to work with both DappV1 and DappV2 lists.
- Adds `dappsCategoriesAlphabeticalSelector` to show dapp categories in alphabetical order in multiple languages.
- Adds a DappFilter type.
- Adds translations for new strings.
- Show `DAppsExplorerScreenLegacy.tsx` or new `DAppsExplorerScreenFilters.tsx` screen based on remote config in `DrawerNavigator.tsx`

### Video

https://user-images.githubusercontent.com/26950305/222035244-60140bc1-cdc7-4577-81f8-9d4d20e024bc.mov

### Test plan

### Related issues

- Fixes: RET-581

### Backwards compatibility

Yes


src/analytics/Properties.tsx
```diff
@@ -1141,6 +1141,10 @@ interface DappExplorerEventsProperties {
   [DappExplorerEvents.dapp_unfavorite]: DappProperties
   [DappExplorerEvents.dapp_open_info]: undefined
   [DappExplorerEvents.dapp_open_more_info]: undefined
+  [DappExplorerEvents.dapp_filter]: {
+    id: string
+  }
+  [DappExplorerEvents.dapp_filter_remove]: undefined
```
kathaypacific, line 1147: i think it'd be nice to combine these two events, then you could also track more specifically which dapp filter is selected and removed? the event property could take the dapp id + a `isApplied` boolean?

src/navigator/DrawerNavigator.tsx
```diff
@@ -264,7 +272,9 @@ export default function DrawerNavigator() {
       {!!dappsListUrl && (
         <Drawer.Screen
           name={Screens.DAppsExplorerScreen}
-          component={DAppsExplorerScreenLegacy}
+          component={
```
kathaypacific, line 275: i noticed with this approach, if the app is already loaded and the remote config values change, this change isn't applied until the app is completely reloaded. e.g. if you load the app without filters, then the remote config value changes (in the slice.ts, change the value of dappsFilterEnabled and save), the screen component is already registered in the drawer so it doesn't change?

i think instead of doing this here, the component can always be `DAppsExplorerScreen` and within `DAppsExplorerScreen`, we can do the split. we did something similar with CPV https://github.com/valora-inc/wallet/blob/06e188c9d69325d25eb2ff84bd413f3a83364bf8/src/verify/VerificationEducationScreen.tsx#L68-L76 - the benefit is the user can use the filters on the first session that the remote config value updates (instead of the second) and when we remove the legacy code, we don't have to touch the navigator

src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,352 @@
...
+        // Filter chips color based on selected filter
+        chipFilter.id === selectedFilter.id
+          ? { backgroundColor: colors.onboardingBlue }
+          : { backgroundColor: colors.onboardingLightBlue },
+        // First Chip has slightly different margins
+        chipFilter.id === 'all'
+          ? { marginLeft: Spacing.Regular16 }
+          : { marginLeft: Spacing.Smallest8 },
+        // Last Chip has slightly different right margin
+        lastChip && { marginRight: Spacing.Regular16 },
```
kathaypacific, line 282: hmm i think we can simplify the styles a lot here, but it's hard to explain. i'll submit a PR and you can see if you prefer my suggestion

src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,352 @@
...
+function DescriptionView({ message, title }: { message: string; title: string }) {
+  return (
+    <View>
+      <Text style={styles.pageHeaderText}>{title}</Text>
+      <Text style={styles.pageHeaderSubText}>{message}</Text>
+    </View>
+  )
+}
+
+function DappFilterChip({ chipFilter, selectedFilter, setFilter, lastChip }: DappFilterChip) {
```
kathaypacific, line 263: this file is pretty long, can DappFilterChip be it's own file?

src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,352 @@
...
+  const error = useSelector(dappsListErrorSelector)
+  const categories = useSelector(dappsCategoriesAlphabeticalSelector)
+  const dappFavoritesEnabled = useSelector(dappFavoritesEnabledSelector)
+  const dappsMinimalDisclaimerEnabled = useSelector(dappsMinimalDisclaimerEnabledSelector)
+  const dappList = useSelector(dappsListSelector)
+  const favoriteDappsById = useSelector(favoriteDappIdsSelector)
+  const [selectedFilter, setSelectedFilter] = useState({
+    id: 'all',
+    name: t('dappsScreen.allDapps'),
+  })
```
kathaypacific, line 78: we discussed this but can we use just a string value for this, instead of keeping the derived "name" value in state?

src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,352 @@
...
+                      lastChip={false}
+                      key={'all'}
+                    />
+                    {/* Category Filter Chips */}
+                    {categories.map((category, idx) => {
+                      return (
+                        <DappFilterChip
+                          chipFilter={{ id: category.id, name: category.name }}
+                          selectedFilter={selectedFilter}
+                          setFilter={setSelectedFilter}
```
kathaypacific, line 158: nit: can this be `onPress` instead, so that we immediately know how it's used?

src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,352 @@
...
+                {dappFavoritesEnabled && (
+                  <Text style={styles.sectionTitle}>{t('dappsScreen.allDappsUpper')}</Text>
+                )}
+              </>
+            }
+            style={styles.sectionList}
+            contentContainerStyle={{
+              padding: Spacing.Thick24,
+              paddingBottom: Math.max(insets.bottom, Spacing.Regular16),
+              flexGrow: 1,
```
kathaypacific, line 189: we have a stylesheet already, can we move any styles that don't require logic to the stylesheet? it's not just here, there's quite a few places in this component where we should remove inline styles

src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,352 @@
...
+import useOpenDapp from 'src/dappsExplorer/useOpenDapp'
+import Help from 'src/icons/Help'
+import DrawerTopBar from 'src/navigator/DrawerTopBar'
+import { TopBarIconButton } from 'src/navigator/TopBarButton'
+import colors from 'src/styles/colors'
+import fontStyles from 'src/styles/fonts'
+import { Spacing } from 'src/styles/styles'
+
+const AnimatedSectionList =
+  Animated.createAnimatedComponent<SectionListProps<DappV1 | DappV2, SectionData>>(SectionList)
```
kathaypacific, line 44: shouldn't this component only reference `DappV2`? I think you can remove all references to `DappV1`, and the correct type for `category` in `SectionData` should be string instead of `CategoryWithDapps`?

src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,352 @@
...
+            dapp.categories &&
+            dapp.categories.includes(filter.id)
+        )
+  // Return empty array if no results
+  if (data.length === 0) return []
+  // Else return dapps in all section
+  return [
+    {
+      data,
+      category: 'all' as unknown as CategoryWithDapps,
```
kathaypacific, line 249: if you follow my comment above and update the type of `SectionData`, you can remove this `as unknown as CategoryWithDapps`

src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,352 @@
...
+            // @ts-ignore TODO: resolve type error
+            ref={sectionListRef}
+            ListFooterComponent={
+              <>
+                {dappsMinimalDisclaimerEnabled ? (
+                  <Text style={styles.disclaimer}>{t('dappsDisclaimerAllDapps')}</Text>
+                ) : null}
+              </>
+            }
+            ListHeaderComponent={
```
kathaypacific, line 128: this SectionList is a little hard to read when the props are 53 lines long. what do you think about taking the whole horizontally scrolling carousel and turning it into a component / new file?

src/dappsExplorer/DAppsExplorerScreen.tsx
```diff
@@ -0,0 +1,352 @@
...
+                      )
+                    })}
+                  </ScrollView>
+                </View>
+                {dappFavoritesEnabled && (
+                  <>
+                    <Text style={styles.sectionTitle}>{t('dappsScreen.favoriteDappsUpper')}</Text>
+                    <FavoriteDappsSection
+                      filter={selectedFilter}
+                      removeFilter={() => {
```
kathaypacific, line 171: the remove filter functionality is repeated, can we abstract this into a function?

src/dappsExplorer/NoResults.tsx
```diff
@@ -0,0 +1,82 @@
...
+interface Props {
+  filter: DappFilter | null
+  removeFilter: () => void
+  testID: string
+}
+
+export function NoResults({ filter, removeFilter, testID }: Props) {
+  const { t } = useTranslation()
+
+  const removePress = () => {
```
kathaypacific, line 21: can we stick to the vanilla naming conventions so that it's really easy to guess what's happening 🙈 i think generally we use `onPress` or `onPressXXXX` or `handlePressXXXX` for click handers, and the rest for executing logic

src/dappsExplorer/NoResults.tsx
```diff
@@ -0,0 +1,82 @@
...
+        testID={`${testID}/NoResults/RemoveFilter`}
+      >
+        <Text style={styles.removeFilterText}>{t('dappsScreen.emptyResults.removeFilter')}</Text>
+      </Touchable>
+    </View>
+  )
+}
+
+const styles = StyleSheet.create({
+  favorites: {
```
kathaypacific, line 48: `container`?

src/dappsExplorer/NoResults.tsx
```diff
@@ -0,0 +1,82 @@
...
+    paddingVertical: 0,
+    paddingHorizontal: 8,
+    backgroundColor: Colors.gray2,
+    borderRadius: 16,
+  },
+  removeFilterText: {
+    ...fontStyles.regular500,
+    fontWeight: '500',
+    fontSize: 12,
+    lineHeight: 20,
```
kathaypacific, line 79: i think i commented on the previous PR but what do you think about just using the closest font that we have defined (e.g. small500). it means that when we update the design system, places like this will not be missed (since it has a bunch of custom styles that overwrite the design system). fwiw i think that not customising the design system styles keeps the integrity of the system + it's less code to maintain :) even if this means that we're off by a few pixels from the designs

src/dappsExplorer/FavoriteDappsSection.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  // If dapp filter enabled display favorites matching filter or no results section
+  // Favorites dapps matching category filter
+
+  const favoritedDappsFiltered = favoriteDapps.filter(
+    (dapp) => isDappV2(dapp) && (dapp.categories.includes(filter?.id ?? '') || filter?.id === 'all')
+  )
+
+  // Matching favorite dapp(s) found
+  if (favoriteDapps.length > 0 && favoritedDappsFiltered.length > 0) {
```
kathaypacific, line 45: i think the `favoriteDapps.length > 0` check is redundant since we already return out of this function in line 26 if this is false

src/icons/Help.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react'
+import Svg, { Circle, Path } from 'react-native-svg'
+import colors from 'src/styles/colors'
+
+interface Props {
+  color?: colors
+}
+
+const Help = ({ color = colors.gray3 }: Props) => (
```
kathaypacific, line 9: argh, i strongly feel like we don't need to duplicate this Help icon just because of the props, i know you said you got build issues so i'll try from my side

kathaypacific (commented):
Looks good so far, i didn't get round to reviewing the test yet though

src/dappsExplorer/DAppsExplorerScreen.tsx
kathaypacific: what do you think about https://github.com/valora-inc/wallet/pull/3492?

src/icons/Help.tsx
kathaypacific: does this work for you? https://github.com/valora-inc/wallet/pull/3493

locales/base/translation.json
```diff
@@ -1219,18 +1219,24 @@
...
   "dappsScreen": {
     "title": "Dapps",
-    "message": "Explore ways to use your crypto",
+    "message": "Explore ways to use your assets",
     "errorMessage": "There was an error loading dapps",
     "featuredDapp": "Featured",
     "allDapps": "All",
+    "allDappsUpper": "ALL",
     "favoriteDapps": "My Favorites",
+    "favoriteDappsUpper": "MY FAVORITES",
```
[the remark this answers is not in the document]
MuckT: Fixed in f88218b514955d71753d29a0e6187ea9fb7f8af5!

src/navigator/DrawerNavigator.tsx
MuckT: Made this change in ac6f9bd7f74e9a0e9f7bfbe92f5c4dbfb80dba19! This will make implementing the search a lot easier. There might be a bit of code duplication, but I agree it's a better experience for users.

MuckT: After this change when developing locally I am seeing a possible bug. 
Reproduction Steps.

1. Set `dappsFilterEnabled = true` in `remoteConfigValuesDefaults.ts`.
2. Launch the app and navigate to the dapps explorer.
3. Observe `DappsExplorerScreenFilter` is served and WAI.
4. Set the `dappsFilterEnabled = false` in `remoteConfigValuesDefaults.ts`.
5. Navigate to the dapps explorer.
6. Observe `DappsExplorerScreenFilter` is served incorrectly and the dapps are not displayed. This probably because we are fetching the V1 dapp list but attempting to display the results in the DappExplorerScreenFilter.tsx which expects v2.

MuckT: Removed this change in 48909fcb59866c42bfc08c3c71b11a34ae3c42a4.

src/dappsExplorer/DAppsExplorerScreen.tsx
MuckT: Fixed in f17dfe22ae5c764d169170c818bfa012e6257f05!

src/dappsExplorer/DAppsExplorerScreen.tsx
MuckT: #3492 looks good to me! I left the key value, but otherwise made the changes in f17dfe22ae5c764d169170c818bfa012e6257f05.

src/dappsExplorer/DAppsExplorerScreen.tsx
MuckT: Fixed in 5e4e9be0f975baa845bf43846989f61cb2a4ef99!

src/dappsExplorer/DAppsExplorerScreen.tsx
MuckT: Fixed in 5e4e9be0f975baa845bf43846989f61cb2a4ef99!

src/dappsExplorer/FavoriteDappsSection.tsx
MuckT: fixed in d2b815fb337c8617bcf131e8dfec60e69a7f81d4!

src/dappsExplorer/DAppsExplorerScreen.tsx
MuckT: Fixed in 55b20eb68d73ebf49213e9c26a4ae5f7a8d34e89!

src/dappsExplorer/NoResults.tsx
MuckT: Fixed in e5341bcc2662f046e048379299861b43e3205765!

src/dappsExplorer/NoResults.tsx
MuckT: Fixed in e5341bcc2662f046e048379299861b43e3205765!

src/dappsExplorer/NoResults.tsx
MuckT: Fixed in e5341bcc2662f046e048379299861b43e3205765!

src/dappsExplorer/DAppsExplorerScreen.tsx
MuckT: Fixed in 9c80d3cf4f50d1e3ff64f90628b46785f3fc02aa!

src/dappsExplorer/DAppsExplorerScreenFilter.tsx
```diff
@@ -0,0 +1,319 @@
...
+                    </Text>
+                    <FavoriteDappsSection
+                      filter={selectedFilter}
+                      removeFilter={removeFilter}
+                      onPressDapp={onSelectDapp}
+                    />
+                  </>
+                )}
+
+                {dappFavoritesEnabled && (
```
kathaypacific, line 183: nit: this can be combined with the previous block

src/dappsExplorer/DAppsExplorerScreenFilter.tsx
```diff
@@ -0,0 +1,319 @@
...
+  const { onFavoriteDapp, DappFavoritedToast } = useDappFavoritedToast(sectionListRef)
+  const { openSheet, DappInfoBottomSheet } = useDappInfoBottomSheet()
+
+  useEffect(() => {
+    dispatch(fetchDappsList())
+    ValoraAnalytics.track(DappExplorerEvents.dapp_screen_open)
+  }, [])
+
+  const removeFilter = () => {
+    ValoraAnalytics.track(DappExplorerEvents.dapp_filter, { id: selectedFilter.id, remove: true })
```
kathaypacific, line 84: could we make the set / remove filter implementation consistent? i don't mind which solution to take, but it would be nice to either declare the set / remove filter functions in this file and pass it down OR pass the `setSelectedFilter` down directly. i.e. manage the same analytics event in the same file.

right now, the remove filter is declared here and passed to the favourites and no results components. meanwhile the filter chip component takes the `setSelectedFilter` directly and handles the analytics. i know it probably made sense to keep the `removeFilter` here because of the scroll ref, but i think it'd be nice to have the chip filter component also just receive a callback that it can execute without worrying about any extra logic. what do you think?

src/dappsExplorer/DappFilterChip.tsx
```diff
@@ -0,0 +1,72 @@
...
+interface DappFilterChip {
+  chipFilter: DappFilter
+  isSelected: boolean
+  onPress: (filter: DappFilter) => void
+  style?: StyleProp<ViewStyle>
+}
+
+function DappFilterChip({ chipFilter, isSelected, onPress, style }: DappFilterChip) {
+  const filterPress = () => {
+    ValoraAnalytics.track(DappExplorerEvents.dapp_filter, { id: chipFilter.id, remove: false })
```
kathaypacific, line 20: i think this should be `remove:  isSelected` no?

src/dappsExplorer/NoResults.tsx
```diff
@@ -0,0 +1,80 @@
...
+        <Trans
+          i18nKey={'dappsScreen.emptyResults.message'}
+          tOptions={{ filter: filter?.name.toLocaleUpperCase(language ?? 'en-US') }}
+        >
+          <Text style={styles.filterText} />
+        </Trans>
+      </Text>
+      <View style={styles.removeFilterTouchableContainer}>
+        <Touchable
+          hitSlop={{ top: 10, bottom: 10, left: 10, right: 10 }}
```
kathaypacific, line 38: nit: there is an `iconHitslop` in the variables file for this

src/dappsExplorer/NoResults.tsx
```diff
@@ -0,0 +1,80 @@
...
+    color: Colors.gray5,
+    paddingHorizontal: 4,
+  },
+  removeFilterTouchableContainer: {
+    borderRadius: 16,
+    overflow: 'hidden',
+  },
+  removeFilterTouchable: {
+    paddingVertical: 0,
+    paddingHorizontal: 8,
```
kathaypacific, line 72: use the Spaces variable here?

src/dappsExplorer/DAppsExplorerScreenFilter.test.tsx
```diff
@@ -0,0 +1,426 @@
...
+  dapps: { dappListApiUrl: 'http://url.com', dappsList, dappsCategories },
+})
+
+describe(DAppsExplorerScreenFilter, () => {
+  beforeEach(() => {
+    defaultStore.clearActions()
+    jest.clearAllMocks()
+  })
+
+  it('renders correctly when no featured dapp is available', () => {
```
kathaypacific, line 41:
```suggestion
  it('renders correctly and fires the correct actions on press dapp', () => {
```

src/dappsExplorer/DAppsExplorerScreenFilter.test.tsx
```diff
@@ -0,0 +1,426 @@
...
+    fireEvent.press(getByTestId('Dapp/dapp1'))
+    fireEvent.press(getByTestId('ConfirmDappButton'))
+
+    expect(defaultStore.getActions()).toEqual([
+      fetchDappsList(),
+      dappSelected({ dapp: { ...dappsList[0], openedFrom: DappSection.All } }),
+    ])
+  })
+
+  it('opens the screen directly when using a deeplink', () => {
```
kathaypacific, line 62:
```suggestion
  it('renders correctly and fires the correct actions on press deep linked dapp', () => {
```

src/dappsExplorer/DAppsExplorerScreenFilter.test.tsx
```diff
@@ -0,0 +1,426 @@
...
+        id: '2',
+        remove: false,
+      })
+      expect(ValoraAnalytics.track).toHaveBeenCalledWith(DappExplorerEvents.dapp_filter, {
+        id: '2',
+        remove: true,
+      })
+    })
+  })
+})
```
kathaypacific, line 426: nice filters tests :) fwiw i think we should prefer `getByText` over `getByTestId` where it's possible. i started thinking this after reading this [article](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-querying-by-text) but also i think the way that we gain the most confidence is seeing our users use the feature. the closer we can get to testing what a real user sees, the more confidence we'd gain

src/dappsExplorer/NoResults.tsx
```diff
@@ -0,0 +1,80 @@
...
+    <View testID={`${testID}/NoResults`} style={styles.favoritesContainer}>
+      <Text style={styles.filterAppliedText}>
+        <Trans
+          i18nKey={'dappsScreen.emptyResults.message'}
+          tOptions={{ filter: filter?.name.toLocaleUpperCase(language ?? 'en-US') }}
+        >
+          <Text style={styles.filterText} />
+        </Trans>
+      </Text>
+      <View style={styles.removeFilterTouchableContainer}>
```
kathaypacific, line 36: should we add a margin vertical here? the filter is a little tight once the content spills to two lines

Simulator Screen Shot - iPhone 14 Pro - 2023-03-08 at 12 50 23

kathaypacific (approved):
hey nice work!!! this looks great :) i left a few small comments that should be addressed before merging.

we spoke a couple times about whether keeping a filter object or a filter id in state is better. i wasn't sure if you had played around with this, but i took a really quick pass just to see what it looks like. up to you if you want to incorporate these changes, i do like having a simple type for the state + being able to remove the extra filter chip type, but this implementation does introduce a useMemo so that's the tradeoff. i'm not too fussed about which solution we end up going for, but wanted to add this idea https://github.com/valora-inc/wallet/pull/3512

src/dappsExplorer/NoResults.tsx
MuckT: Fixed in 3347408038796b8fd990bacb10192130b069ddd2!

src/dappsExplorer/NoResults.tsx
MuckT: Fixed in 3347408038796b8fd990bacb10192130b069ddd2!

src/dappsExplorer/NoResults.tsx
MuckT: Applying the margin vertical here is not working well when text is in one line. Hopefully after the text style refactor this will be less of an issue.

src/dappsExplorer/DAppsExplorerScreenFilter.test.tsx
MuckT: Fixed in cbac8b8506318a72cec64093ec4bcb2164fdc7a4!

PR #3475, opened 2023-02-28
chore(supercharge): add remote config variable for v1 addresses

kathaypacific:
### Description

Related to https://github.com/valora-inc/wallet/pull/3455#discussion_r1118950066.

During the first 4 weeks of the new system being live, we'll still have rewards that are claimed against the old system. I can't think of how else to "whitelist" those addresses, so I added a remote config variable...open to better ideas about how to cater for this.

### Test plan

n/a

### Related issues

- Fixes RET-513

### Backwards compatibility

Y


src/app/saga.ts
```diff
@@ -208,6 +208,7 @@ export interface RemoteConfigValues {
   guaranteedSwapPriceEnabled: boolean
   superchargeV2Enabled: boolean
   superchargeRewardContractAddress: string
+  superchargeV1Addresses: string[]
```
jeanregisser, line 211: Maybe we should add a comment about what this? and when we can remove it?

jeanregisser (approved):
Looks good to me 👍

PR #3485, opened 2023-03-01
chore(deps): update cross-fetch resolution

anthonydaoud:
### Description

As per [this dependabot alert](https://github.com/valora-inc/wallet/security/dependabot/59), `cross-fetch` needs to be at least at version 3.1.5 to prevent a security issue.

### Test plan

CI

### Related issues

N/A

### Backwards compatibility

Yes


MuckT (approved):

yarn.lock
```diff
@@ -8079,7 +8079,15 @@ create-require@^1.1.0:
   resolved "https://registry.yarnpkg.com/create-require/-/create-require-1.1.1.tgz#c1d7e8f1e5f6cfc9ff65f9cd352d37348756c333"
   integrity sha512-dcKFX3jn0MpIaXjisoRvexIJVEKzaq7z2rZKxf+MSr9TkdmHmsU4m2lcLojrj/FHl8mk5VxMmYA+ftRkP/3oKQ==
 
-cross-fetch@3.0.4, cross-fetch@^3.0.2, cross-fetch@^3.0.4, cross-fetch@^3.0.6, cross-fetch@^3.1.5:
+cross-fetch@3.0.4:
```
cajubelt, line 8082: not sure this added version of cross-fetch is intentional/desirable

anthonydaoud: That version comes from a dependency in the `@celo/identity` package. There don't seem to be any security, testing, or build size issues with using it, so it seems better to use the dependency specified by the package rather than override it.

cajubelt: look at what the lockfile was doing before-- for `cross-fetch@3.0.4`, it was using version 3.1.5. Why change that? It seems like you just reran yarn install and accepted whatever changes were auto-generated, but I don't think that's correct. We end up with this new version we didn't need before, that's older than the one we were using. 

My guess is that the only change that's actually needed to the lockfile is the removal of `cross-fetch@^3.0.2` from the things that use 3.1.5

PR #3488, opened 2023-03-01
fix(fees): don't show error banner if fee currency balance is zero

satish-ravi:
### Description

Estimating fees for a currency with zero balance, which results in the "Could not calculate fees" error banner. This case accounts for ~92% of all `estimate_fee_failed` events in the last 3 months: https://mixpanel.com/project/2786968/view/3321744/app/insights#QXXXnHBAkKzr
The saga throws "Error: execution reverted: transfer value exceeded balance of sender" when it tries to create a small tx and estimate gas.
Also, this banner will definitely show up for any user attempting to cash in first time since the wallet will have a 0 balance or if the swaps screen is loaded with 0 CELO (the default source for swaps). We don't need to show this banner for these cases and also for any other 0 balance, since we have other checks and error messages if the user attempts to send/swap/cashout with a currency with 0 balance.

This PR updates the feeEstimate saga to early bail if the balance is 0 and not show the banner. 

### Test plan

- Unit tests
- Manually by going through the below and ensuring the banner doesn't show up
  - CICO flow with a currency with 0 balance
  - Send flow with a currency with 0 balance
  - Swap screen with 0 CELO (the default source for swaps)

### Related issues

- Fixes part of ACT-653

### Backwards compatibility

Yes


cajubelt (approved):

src/fees/saga.test.ts
```diff
@@ -245,4 +247,57 @@ describe(estimateFeeSaga, () => {
...
+  })
+
+  it('marks as error without fee banner if token balance is zero', async () => {
+    const { effects } = await expectSaga(estimateFeeSaga, {
+      payload: { feeType: FeeType.SEND, tokenAddress: mockCeurAddress },
+    })
+      .withState(store.getState())
+      .provide([
+        [matchers.call.fn(buildSendTx), jest.fn(() => ({ txo: mockTxo }))],
+        [matchers.call.fn(estimateGas), throwError(new Error('cannot estimate fee'))],
```
jophish, line 285: We don't need to provide these for this test, since `estimateFeeSaga` will bail almost instantly, right?

jophish (approved):

src/fees/saga.test.ts
satish-ravi: yeah it's not used on this test, but I started writing this test as a failing test for when we don't early bail on 0 balance and left it as is.

PR #3490, opened 2023-03-02
feat(send): Require token selection on send/request flow

jophish:
### Description

For ACT-658.

Currently, we do not explicitly require the user to select a currency when sending/requesting. This can lead to confusion/lack of clarity about what token is being used for a given transfer. This PR requires the user to input a token during the send/request flow, via a bottom sheet that appears after a recipient has been selected.

For the send flow, the bottom sheet contains all transferable tokens for which the user has a non-zero balance. For the request flow, the bottom sheet contains the stable tokens. This behavior mirrors the currently selectable tokens in the `SendAmount` screen which follows.

This also rejiggers some of the parameters used in the `Send` and `SendAmount` screens. Previously, `forceTokenAddress` was responsible for two things:
1. Overriding the default token (by providing a token address as a string)
2. Preventing the user from changing tokens (by simply being non-null)

Since, for the regular send/request flows, we just want the behavior of 1., I've modified `forceTokenAddress` to be a `boolean` indicating whether or not the user should be able to switch from the default, and added a `defaultTokenOverride`, which just overrides the default token. In places where we used to use `forceTokenAddress`, I've updated it to use the _new_ `forceTokenAddress`, as well as `defaultTokenOverride`.

### Test plan

Unit and manual tested. See videos below:

#### Send Flow

https://user-images.githubusercontent.com/569401/222310926-dbe3f2b4-1be4-46ce-a736-f25744d9deb4.mp4

#### Request Flow

https://user-images.githubusercontent.com/569401/222310949-4a7a3349-0cdb-4ceb-8b4e-069e71603f2b.mp4

### Related issues

- Fixes ACT-658.

### Backwards compatibility

Yes.


jophish:
updating e2e tests, but should be otherwise good to review

src/send/Send.tsx
```diff
@@ -135,10 +137,36 @@ function Send({ route }: Props) {
...
+      } else {
+        setShowCurrencyPicker(true)
+      }
     },
     [isOutgoingPaymentRequest, searchQuery]
   )
 
+  const onTokenSelected = (token: string) => {
+    setShowCurrencyPicker(false)
+    if (!recipient) {
```
jh2oman, line 158: What case is this for? Shouldn't this only be callable after a recipient has been selected?

jh2oman (approved):
Question, but not blocking. Looks good!

src/send/Send.tsx
jophish: This is 1: to be defensive, since _technically_ there aren't any compile-time guarantees that this will always be defined, and 2. to appease the Typescript lords, since we can't pass `string | undefined` as a value for `recipient` to the `SendAmount` screen.

PR #3491, opened 2023-03-02
fix(swap): inconsistent fee currency when swapping from CELO

kathaypacific:
### Description

On the swap screen, we estimate the fee and use this to determine the "max" "from" amount. Then when we perform the swap, we calculate the gas price / currency again.

For a wallet with more stablecoins than CELO, when pressing "max" to swap all CELO to something else, the "useMaxAmount" hook returns the whole CELO balance because the stablecoin with the higher balance will pay for gas. However when we perform the swap, CELO is chosen as the gas currency and therefore the tx fails of course.

This PR:
- set the preferred gas currency before sending the swap, consistent with the start of the swap flow
- increase the fee estimate buffer. i was often seeing failures when swapping from max cEUR :/ this increased buffer seems to work better

### Related issues

- Fixes RET-609

### Backwards compatibility

Y


jeanregisser (approved):
Great! 👍 

This fix works and is good enough probably, 
but it seems the we should return the `feeCurrency` from `useMaxSendAmount` and pass it all the way to where we send the transaction to avoid this type of mismatch in the first place.

kathaypacific:
> Great! 👍
> 
> This fix works and is good enough probably, but it seems the we should return the `feeCurrency` from `useMaxSendAmount` and pass it all the way to where we send the transaction to avoid this type of mismatch in the first place.
> 
> This is another example where we should rethink how we present TX before sending. Right now, users can accept sending TX without knowing which token will be used to pay for fees and the estimated amount (which is recalculated) after approving. This isn't right.

this might be a good one for the eng backlog? the fee estimation stuff is one of things that i don't feel super confident with in the wallet, so i don't have a vision right now for what it should look like. not all places that need a fee currency calculation use the `useMaxSendAmount` hook, but i don't have a good overall idea of all the use cases. maybe this is something we can discuss and document, i don't feel confident that it's the right time to do this refactor to solve this bug?

i agree that surfacing the tx data to the user would solve some of these problems (at least, this bug would have been spotted sooner), i wonder if we should also allow changing the fee currency too :/

jeanregisser:
Yes totally, let's merge this and discuss the possible bigger changes separately 👍

PR #3493, opened 2023-03-02
chore(dapp filters): repurpose info icon

kathaypacific:
### Related issues

- Fixes #[issue number here]


MuckT:
@kathaypacific I am not a huge fan of this as it's a bit strange that this icon is different than all the other `src/navigation/icons` and requires changing the navigator itself. Is there a reason we shouldn't add the new help icon consistent with non-navigator icons instead of changing this one to work in both cases?

kathaypacific:
> @kathaypacific I am not a huge fan of this as it's a bit strange that this icon is different than all the other `src/navigation/icons` and requires changing the navigator itself. Is there a reason we shouldn't add the new help icon consistent with non-navigator icons instead of changing this one to work in both cases?

i guess i just found it strange that the design for this essentially didn't change but we need to add a new icon :/ i suppose it makes sense to have separate navigator icons because we can change them independently of the normal icons. but then the existing dapps screen should not use the navigator icon.

MuckT:
Fixed in #3504!

PR #3494, opened 2023-03-03
chore: add requireCPV remote config

kathaypacific:
### Description

This PR adds a remote config that we can switch to true when we want to enable manual verification for users who have not migrated to CPV.

### Test plan

n/a

### Related issues

- Fixes RET-638

### Backwards compatibility

Y

### DIFF
--- a/src/fiatExchanges/SimplexScreen.tsx
+++ b/src/fiatExchanges/SimplexScreen.tsx
@@ -7,7 +7,7 @@ import { useDispatch, useSelector } from 'react-redux'
 import { e164NumberSelector } from 'src/account/selectors'
 import { showError } from 'src/alert/actions'
 import { ErrorMessages } from 'src/app/ErrorMessages'
-import { numberVerifiedSelector } from 'src/app/selectors'
+import { phoneNumberVerifiedSelector } from 'src/app/selectors'
 import BackButton from 'src/components/BackButton'
 import Button, { BtnSizes } from 'src/components/Button'
 import WebView from 'src/components/WebView'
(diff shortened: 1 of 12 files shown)


src/fiatExchanges/SimplexScreen.tsx
```diff
@@ -37,7 +37,7 @@ function SimplexScreen({ route, navigation }: Props) {
 
   const account = useSelector(currentAccountSelector)
   const e164PhoneNumber = useSelector(e164NumberSelector)
-  const phoneNumberConfirmed = useSelector(numberVerifiedSelector)
+  const phoneNumberConfirmed = useSelector(phoneNumberVerifiedSelector)
```
kathaypacific, line 40: this is kind of unrelated but i don't think Simplex should just be considering the decentralised status

src/redux/migrations.ts
```diff
@@ -1063,4 +1063,18 @@ export const migrations = {
...
     ),
   }),
+  114: (state: any) => ({
+    ...state,
+    account: {
+      ...state.account,
+      dismissedGetVerified:
+        state.app.numberVerified && !state.app.phoneNumberVerified
+          ? false
+          : state.account.dismissedGetVerified,
```
kathaypacific, line 1073: this is a little sneaky but wanted to squeak this homecard back in

src/send/Send.tsx
```diff
@@ -221,9 +221,7 @@ function Send({ route }: Props) {
       <SendHeader isOutgoingPaymentRequest={isOutgoingPaymentRequest} />
       <DisconnectBanner />
       <SendSearchInput input={searchQuery} onChangeText={throttledSearch} />
-      {inviteRewardsEnabled && numberVerified && hasGivenContactPermission && (
-        <InviteRewardsBanner />
-      )}
+      {inviteRewardsEnabled && hasGivenContactPermission && <InviteRewardsBanner />}
```
kathaypacific, line 224: the `inviteRewardsEnabled` selector already takes into account phone verification, we shouldn't show this banner to non CPV users anyway

jeanregisser (approved):
LGTM 👍

PR #3496, opened 2023-03-06
chore: upgrade payments sdk

kathaypacific:
### Description

As the title. 

### Test plan

The [test website](https://sandbox-celo-demo-store.demo.firstdag.com/) for merchant payments is no longer available, so it's not really possible to test this upgrade fully. The provider of this website is (afaik) the only user of this feature. We're not sure is this feature is actively being used, I've checked with Ellen and we are fine to release this as is. I scanned the QR in the video in https://github.com/valora-inc/wallet/pull/3376 and seems to work well enough.

### Related issues

- Fixes RET-637

### Backwards compatibility

Probably.


jeanregisser (approved):
:rocket:

PR #3498, opened 2023-03-06
feat(swap): add warning when swapping max CELO

kathaypacific:
### Description

As the title:
- add a warning component
- display the component when the user has pressed the max button while swapping from CELO 
- hide the component when the user edits the swap amount
- add analytics to the max button

Unrelated:
- removed a bunch of unused icons - i was going through the folder to see if the icon that i need exists already, and found a bunch of redundant things.

### Test plan

Added small unit test.

https://user-images.githubusercontent.com/20150449/223147068-5510e697-ba9e-44f5-a03c-82480511114d.mp4

### Related issues

Fixes RET-609

### Backwards compatibility

Y


src/swap/SwapScreen.tsx
```diff
@@ -132,6 +134,7 @@ export function SwapScreen() {
 
     if (parsedSwapAmount[Field.FROM].gt(maxFromAmount)) {
       setFromSwapAmountError(true)
+      showMaxCeloSwapWarning()
```
kathaypacific, line 137: keen for thoughts on this behaviour - i thought if the goal is to warn users about future implications of not having gas fees, then if they manually enter a swap amount that is too close to their absolute token balance then this warning would be helpful too?

MuckT:
@kathaypacific seeing an issue with larger fonts.
<p align="center">

</p>

jeanregisser (approved):
Awesome 🚀

PR #3500, opened 2023-03-06
feat: format 12 word seed phrase into a table design

anthonydaoud:
### Description

Currently, the wallet displays both 12 and 24 word seed phrases as a list of words. This PR adds modifies the display of 12 word phrases so that they are now shown in a table.

**Before**

backupBefore

**After**

backupAfter

### Test plan

Updated e2e tests and jest snapshot.

### Related issues

- ACT-649

### Backwards compatibility

Yes.


src/backup/BackupPhraseContainer.tsx
```diff
@@ -12,6 +12,25 @@ import Logger from 'src/utils/Logger'
...
 const PhraseInput = withTextInputPasteAware(TextInput, { top: undefined, right: 12, bottom: 12 })
 
+type TwelveWordTableProps = {
+  words: string
+}
+
+function TwelveWordTable({ words }: TwelveWordTableProps) {
+  return (
+    <>
+      {words.split(' ').map((word, index) => (
```
cajubelt, line 22: nit: you already split the words to figure out if there are 12 words and this component should be used in the first place. would be more efficient to take a list of words as props, and avoid splitting the same string twice.

src/backup/BackupPhraseContainer.tsx
```diff
@@ -68,12 +88,14 @@ export class BackupPhraseContainer extends React.Component<Props> {
...
-              </Text>
-            )}
+          <View
+            style={styles.phraseContainer}
+            testID="AccountKeyWordsContainer"
+            accessibilityLabel={words ?? ''}
+          >
+            {isTwelveWords
+              ? !!words && <TwelveWordTable words={words} />
+              : !!words && <Text style={styles.phraseText}>{words}</Text>}
```
cajubelt, line 98:
```suggestion
            {words && (isTwelveWords
              ? <TwelveWordTable words={words}/>
              : <Text style={styles.phraseText}>{words}</Text>)}
```
code golf

cajubelt (approved):
did you test on different screen sizes?

kathaypacific:
*a wild kathy appeared* sorry i stumbled across this PR and this issue is existing, but while you're there what do you think about updating the grey container so that it is left / right aligned with the rest of the content? it made me sad to see that it is a different width to the other content 😅

anthonydaoud:
@cajubelt Yes I verified this looks good on different sized phones.

@kathaypacific Sure thing, just updated it so that the grey container is aligned!

PR #3505, opened 2023-03-06
refactor(onboarding): consolidate onboarding flow logic.

jh2oman:
### Description

This solves a few problems with the current onboarding flow logic:
1. The logic for which screen to go to next in onboarding is currently spread out throughout the different onboarding screens. This makes it difficult to see in-code how the onboarding flow actually works with its different branching options. With this refactor all that logic would be in one place
2. The logic for the 'Step X of Y' is currently totally separate from the logic of which screens come after one another in onboarding. This duplication of logic is confusing and messy. With this refactor we have a traversable graph that can be used to calculate the steps without separate logic that must change every time a screen is added.

### Test plan

1. More unit test coverage than there was before for the onboarding flow
2. Unit tests of the Onboarding directed graph ensuring that counters always increment and the graphs always end up at Home.

### Related issues

https://linear.app/valora/issue/ACT-597/refactor-onboarding-screen-logic

### Backwards compatibility

yes


src/onboarding/map.ts
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ * Traverses through the directed graph of onboarding navigate, navigateClearingStack, and navigateHome calls
+ * and count the number of screens until the given screen and the total number
+ */
+export function getStepValues(screen: Screens) {
+  let stepCounter = 0
+  let totalCounter = 0
+  let reachedScreeen = false
+  let currentScreen: Screens = Screens.NameAndPicture
```
cajubelt, line 34: would be nice if `Screens.NameAndPicture` was a constant somewhere named FIRST_ONBOARDING_SCREEN or something

src/onboarding/map.ts
```diff
@@ -0,0 +1,163 @@
...
+    if (!reachedScreeen) {
+      stepCounter++
+    }
+    if (currentScreen === screen) {
+      reachedScreeen = true
+    }
+  }
+
+  const navigateHome = () => {
+    currentScreen = Screens.WalletHome
```
cajubelt, line 48: seems like there could be a constant for `Screens.WalletHome` like AFTER_ONBOARDING_SCREEN . This could actually change in the future, for example with the "choose your adventure" experience

src/onboarding/map.ts
```diff
@@ -0,0 +1,163 @@
...
+    navigate: NavigationService.navigate,
+    popToScreen: NavigationService.popToScreen,
+    navigateHome: NavigationService.navigateHome,
+    navigateClearingStack: NavigationService.navigateClearingStack,
+    dispatch: store.dispatch,
+  })
+  onBoardingMap[screen]?.next()
+}
+
+function _getOnboardingMap({
```
cajubelt, line 83: this is really interesting! I love how it uses dependency injection to make it possible to just traverse the graph without creating a bunch of side effects.

I almost want to take it a step further where it takes all this state as a parameter-- or maybe uses some TS magic to only require the state that's needed from the start screen you're at??-- and returns a screen you need to go to, plus some way of telling what side effects need to take place if you're going to that screen in a "real" onboarding navigation flow. it could be just a function like "doSideEffects". the `navigateClearingStack` use case is annoying, but maybe there's some way of taking what that does and separating out the effectful part into the `doSideEffects` piece.

It's already in pretty good shape so it's hard to say what's an improvement about my idea exactly. Taking state as a parameter makes it a pure function, so that's good. I guess separating the side effects out into a function on the return value makes it feel more fair to use it for `getStepValues`. 

I think my main feeling is that this isn't really a "map"; it's more like a state-dependent set of edges between screens that have a start screen, an end screen, and a side effect for the transition between the screens (which you may or may not want to use!).

cajubelt (commented):
really cool!

src/onboarding/success/OnboardingSuccessScreen.tsx
```diff
@@ -4,14 +4,20 @@ import { Image, StyleSheet, Text, View } from 'react-native'
...
 import colors from 'src/styles/colors'
 import fontStyles from 'src/styles/fonts'
 import { Spacing } from 'src/styles/styles'
 
 function OnboardingSuccessScreen() {
+  const onboardingProps = useTypedSelector(onboardingPropsSelector)
   useEffect(() => {
-    const timeout = setTimeout(() => navigateHome(), 3000)
+    const timeout = setTimeout(
+      () => goToNextOnboardingScreen(Screens.VerificationStartScreen, onboardingProps),
```
cajubelt, line 18: i find this a bit confusing. the intuitive screen parameter seems like it would be Screens.OnboardingSuccessScreen. can you explain why it's not?

src/verify/VerificationStartScreen.tsx
```diff
@@ -81,7 +82,11 @@ function VerificationStartScreen({
   const cachedNumber = useSelector(e164NumberSelector)
   const cachedCountryCallingCode = useSelector(defaultCountryCodeSelector)
   const walletAddress = useSelector(walletAddressSelector)
-  const { step, totalSteps } = useSelector(registrationStepsSelector)
+  const onboardingProps = useTypedSelector(onboardingPropsSelector)
+  const { step, totalSteps } = getOnboardingStepValues(
+    Screens.VerificationStartScreen,
+    onboardingProps
+  )
```
cajubelt, line 89: looks like there is still some bespoke logic for determining the next onboarding screen to show from this one-- not ideal

cajubelt: it seems like the problem is that there are 2 screens you might want to show next-- OnboardingSuccessScreen if they successfully verify, or home if they skip. Two ideas for handling this branching:
- treat "home" as the next screen from this one, and pass a callback to OnboardingSuccessScreen as a prop that takes it to the "next screen in onboarding" (always home)
- put the branching logic inside `goToNextOnboardingScreen`, and choose the next screen by checking whether a user has verified their PN

cajubelt (commented):
still need to review onboarding/steps.ts and steps.test.ts-- skipped for now since they look similar to my last review

src/onboarding/steps.ts
```diff
@@ -0,0 +1,208 @@
...
+  supportedBiometryTypeSelector,
+} from 'src/app/selectors'
+import { setHasSeenVerificationNux } from 'src/identity/actions'
+import * as NavigationService from 'src/navigator/NavigationService'
+import { Screens } from 'src/navigator/Screens'
+import { RootState } from 'src/redux/reducers'
+import { store } from 'src/redux/store'
+
+export const FIRST_ONBOARDING_SCREEN = Screens.NameAndPicture
+export const END_OF_ONBOARDING_SCREEN = Screens.WalletHome
```
cajubelt, line 19: this might be a premature optimization, but in a future where we want to choose the first/last onboarding screen by the user's A/B test cohort, it would be more convenient to have these wrapped by getter functions, like `getFirstOnboardingScreen()`

cajubelt: they can still just return a constant value for now. just thinking if they're exported and used elsewhere, it might be better to wrap them in getter functions

src/onboarding/steps.ts
```diff
@@ -0,0 +1,208 @@
...
+
+interface NavigatorFunctions {
+  navigate: typeof NavigationService.navigate | ((screen: Screens) => void)
```
cajubelt, line 22: i don't really get the OR types here. why is that needed?

src/onboarding/steps.ts
```diff
@@ -0,0 +1,208 @@
...
+  let stepCounter = 1
+  let totalCounter = 1
+  let reachedScreeen = false
+  let currentScreen: Screens = FIRST_ONBOARDING_SCREEN
+
+  const navigate = (s: Screens) => {
+    totalCounter++
+    if (currentScreen === screen) {
+      reachedScreeen = true
+    }
```
cajubelt, line 75:
```suggestion
    reachedScreen = reachedScreen || (currentScreen === screen)
```
code golf. very optional.

src/onboarding/steps.ts
```diff
@@ -0,0 +1,208 @@
...
+ * Traverses through the directed graph of onboarding navigate, navigateClearingStack, and navigateHome calls
+ * and count the number of screens until the given screen and the total number
+ */
+export function getOnboardingStepValues(screen: Screens, onboardingProps: OnboardingProps) {
+  let stepCounter = 1
+  let totalCounter = 1
+  let reachedScreeen = false
+  let currentScreen: Screens = FIRST_ONBOARDING_SCREEN
+
+  const navigate = (s: Screens) => {
```
cajubelt, line 71:
```suggestion
  const navigate = (s: Screens) => {
      // dummy navigation function to help determine what onboarding step the user is on, without triggering side effects like actually cycling them back through the first few onboarding screens
```

cajubelt: also, I generally object to single-character variable names. `toScreen` could be more clear

src/onboarding/steps.ts
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * Traverses through the directed graph of onboarding navigate, navigateClearingStack, and navigateHome calls
+ * and count the number of screens until the given screen and the total number
+ */
+export function getOnboardingStepValues(screen: Screens, onboardingProps: OnboardingProps) {
+  let stepCounter = 1
+  let totalCounter = 1
+  let reachedScreeen = false
+  let currentScreen: Screens = FIRST_ONBOARDING_SCREEN
```
cajubelt, line 69:
```suggestion
  let currentScreen: Screens = FIRST_ONBOARDING_SCREEN  // pointer that we will update when simulating navigation through the onboarding screens to calculate "step" and "totalSteps"
```

src/onboarding/steps.ts
```diff
@@ -0,0 +1,208 @@
...
+}
+
+/**
+ * Traverses through the directed graph of onboarding navigate, navigateClearingStack, and navigateHome calls
+ * and count the number of screens until the given screen and the total number
+ */
+export function getOnboardingStepValues(screen: Screens, onboardingProps: OnboardingProps) {
+  let stepCounter = 1
+  let totalCounter = 1
+  let reachedScreeen = false
```
cajubelt, line 68:
```suggestion
  let reachedScreeen = false  // tracks whether we have reached the screen the user is on in onboarding, and we can stop incrementing stepCounter
```

src/onboarding/steps.ts
```diff
@@ -0,0 +1,208 @@
...
+    numberAlreadyVerifiedCentrally,
+  }
+}
+
+/**
+ * Traverses through the directed graph of onboarding navigate, navigateClearingStack, and navigateHome calls
+ * and count the number of screens until the given screen and the total number
+ */
+export function getOnboardingStepValues(screen: Screens, onboardingProps: OnboardingProps) {
+  let stepCounter = 1
```
cajubelt, line 66:
```suggestion
  let stepCounter = 1  // will increment this up to the onboarding step the user is on
```

src/onboarding/steps.ts
```diff
@@ -0,0 +1,208 @@
...
+      reachedScreeen = true
+    }
+    if (!reachedScreeen) {
+      stepCounter++
+    }
+    currentScreen = s
+  }
+
+  const navigateHome = () => {
+    currentScreen = END_OF_ONBOARDING_SCREEN
```
cajubelt, line 83:
```suggestion
    currentScreen = Screens.WalletHome
```

cajubelt: if we ever updated END_OF_ONBOARDING_SCREEN, we wouldn't want `navigateHome` to not take you home anymore

src/onboarding/steps.ts
```diff
@@ -0,0 +1,208 @@
...
+  const { navigate, popToScreen, navigateHome, navigateClearingStack } = navigator
+  const {
+    recoveringFromStoreWipe,
+    choseToRestoreAccount,
+    supportedBiometryType,
+    skipVerification,
+    numberAlreadyVerifiedCentrally,
+  } = props
+
+  const onboardingScreens: Partial<{
```
cajubelt, line 140: seems like this could be a switch

cajubelt (commented):
ran out of time before i could actually check the logic for each of the screen transitions, which is probly pretty important. would be great if another reviewer could look at that, or if you could self-review.

cajubelt:
also one thing that concerns me a little is that the user's state could potentially change between when they start onboarding and when they finish, which could introduce a bug like a step getting added to the number of total steps, or even a cycle. not sure what the best way of dealing with that is, though.. maybe just making sure it's covered in integration tests?

src/onboarding/steps.test.ts
```diff
@@ -0,0 +1,237 @@
...
+        supportedBiometryType: [null, BIOMETRY_TYPE.FACE_ID],
+        skipVerification: [true, false],
+        numberAlreadyVerifiedCentrally: [true, false],
+      }
+
+      options.recoveringFromStoreWipe.forEach((recoveringFromStoreWipe) => {
+        options.choseToRestoreAccount.forEach((choseToRestoreAccount) => {
+          options.supportedBiometryType.forEach((supportedBiometryType) => {
+            options.skipVerification.forEach((skipVerification) => {
+              options.numberAlreadyVerifiedCentrally.forEach((numberAlreadyVerifiedCentrally) => {
```
cajubelt, line 41: i think this could be a little cleaner if it were refactored as follows:
- take "options" out of the test body
- prep a list of cases to check by iterating over the options like you do here, and just pushing to an array of cases to check
- pass that list of cases to check into an it.each

you could also use [this SO post](https://stackoverflow.com/questions/12303989/cartesian-product-of-multiple-arrays-in-javascript) to prep the cases to check with a 1-line helper, but that's probably overkill

if nothing else, it would be nice to see this test written without all the nested indentation

src/onboarding/steps.test.ts
```diff
@@ -0,0 +1,237 @@
...
+                // Initiate Test Variables
+                const onboardingProps = {
+                  recoveringFromStoreWipe,
+                  choseToRestoreAccount,
+                  supportedBiometryType,
+                  skipVerification,
+                  numberAlreadyVerifiedCentrally,
+                }
+                let currentScreen: Screens = FIRST_ONBOARDING_SCREEN
```
cajubelt, line 50: this test looks like it largely duplicates the implementation of getOnboardingStepValues, and then checks that the answer is the same at the end. This seems too circular to me. Also, it's complex enough that there could easily be bugs in the test itself.

cajubelt: one way you could write this test is select a subset of cases for your test and hardcoding the expected values (after checking them manually). But computing them on the fly like this with essentially the same code is going to make the test less reliable, harder to maintain, and less useful overall

cajubelt (changes requested):
concerned about the complexity of one of the tests

src/onboarding/steps.test.ts
```diff
@@ -0,0 +1,237 @@
...
+
+                const navigate = (s: Screens) => {
+                  currentScreen = s
+                  stepCounter++
+                }
+                const navigateHome = () => {
+                  currentScreen = END_OF_ONBOARDING_SCREEN
+                }
+
+                const { totalSteps } = getOnboardingStepValues(
```
cajubelt, line 61: it looks like all your tests work with the same screen parameter-- would be good to get better coverage on that. the highest impact cases to cover are probably the first screen, the last screen, and some screen in the middle.

jh2oman: All of those screen parameters are tested below in the `expect(steps.totalSteps).toEqual(totalSteps)`, testing that throughout the flow the totalCounts are the same.

src/onboarding/steps.ts
jh2oman: No, the first type is needed for actually executing the navigation. The second type is needed for calculating the steps

src/verify/VerificationStartScreen.tsx
jh2oman: The way I think about it is VerificationStartScreen, VerificationInputScreen, and OnboardingSuccessScreen are all part of the Verification Step, and the logic for navigation between those screens should be handled internally since the logic is tightly coupled. 

The functions like `goToNextOnboardingScreen` should only be used to navigate between loosely coupled steps in the onboarding flow. Maybe I should change the name to `goToNextOnboardingStep` and create a new `OnboardingSteps` enum to capture the difference between screens and steps in this case.

src/onboarding/success/OnboardingSuccessScreen.tsx
jh2oman: Its really the first screen in the current 'step'. I've adjusted the variable names to make this very clear.

src/onboarding/steps.test.ts
```diff
@@ -0,0 +1,254 @@
...
+    ({ onboardingProps, screens, name }) => {
+      const expectedTotalSteps = screens.length
+      screens.forEach((screen, index) => {
+        const { totalSteps, step } = getOnboardingStepValues(screen, onboardingProps)
+        // Checking that the step number is correct
+        try {
+          expect(step).toEqual(index + 1)
+          expect(totalSteps).toEqual(expectedTotalSteps)
+        } catch (error) {
+          throw new Error(
```
cajubelt, line 81: nit: jest already emits a similar error message

cajubelt: maybe you were worried about figuring out which test failed. but you should be able to inject parameters from the test cases into the test description. (that being said, I tried adding "$name" to the end of the test description and it didn't work-- not sure why. maybe the jest version?)

https://jestjs.io/docs/26.x/api#testeachtablename-fn-timeout

src/onboarding/steps.test.ts
```diff
@@ -0,0 +1,254 @@
...
+          // eslint-disable-next-line jest/no-conditional-expect
+          expect(navigateHome).toHaveBeenCalled()
+        } else {
+          try {
+            // eslint-disable-next-line jest/no-conditional-expect
+            expect(navigate).toHaveBeenCalledWith(screens[index + 1])
+          } catch {
+            // eslint-disable-next-line jest/no-conditional-expect
+            expect(navigateClearingStack).toHaveBeenCalledWith(screens[index + 1])
+          }
```
cajubelt, line 102:
```suggestion
                expect(index === screens.length - 1 ? navigateHome : ((navigate as jest.MockedFunction<any>).mock.calls.length ? navigate : navigateClearingStack)).toHaveBeenCalled()
```
this is pretty ugly but I think it works

cajubelt: and it lets us not disable linting

cajubelt: oh just realized it means we can't use toHaveBeenCalledWith. Ugh. your original is probly better.

src/onboarding/steps.test.ts
```diff
@@ -0,0 +1,254 @@
...
+      Screens.NameAndPicture,
+      Screens.PincodeSet,
+      Screens.EnableBiometry,
+      Screens.ImportWallet,
+      Screens.VerificationStartScreen,
+    ],
+    name: 'importWalletFlowEverythingEnabled',
+  }
+
+  it.each([
```
cajubelt, line 66: the refactor looks very good!

src/onboarding/steps.ts
```diff
@@ -63,12 +63,13 @@ export function onboardingPropsSelector(state: RootState): OnboardingProps {
  * and count the number of screens until the given screen and the total number
  */
 export function getOnboardingStepValues(screen: Screens, onboardingProps: OnboardingProps) {
-  let stepCounter = 1
+  let stepCounter = 1 // will increment this up to the onboarding step the user is on
   let totalCounter = 1
-  let reachedScreeen = false
-  let currentScreen: Screens = FIRST_ONBOARDING_SCREEN
+  let reachedScreeen = false // tracks whether we have reached the screen the user is on in onboarding, and we can stop incrementing stepCounter
```
cajubelt, line 68:
```suggestion
  let reachedScreeen = false // tracks whether we have reached the step the user is on in onboarding, and we can stop incrementing stepCounter
```
since you clarified the screens vs steps thing, I realized this should be "step", not "screen"

cajubelt (approved):

src/onboarding/steps.ts
```diff
@@ -0,0 +1,214 @@
...
+      navigateHome: NavigationService.navigateHome,
+      navigateClearingStack: NavigationService.navigateClearingStack,
+    },
+    dispatch: store.dispatch,
+    props: onboardingProps,
+  })
+  stepInfo?.next()
+}
+
+export function _getStepInfo({ firstScreenInStep, navigator, dispatch, props }: GetStepInfoProps) {
```
jophish, line 137: I think a docstring here could be helpful, since this is now the source of truth for onboarding navigation logic. Probably would be useful to mention this.

src/onboarding/steps.ts
```diff
@@ -0,0 +1,214 @@
...
+  const { navigate, popToScreen, navigateHome, navigateClearingStack } = navigator
+  const {
+    recoveringFromStoreWipe,
+    choseToRestoreAccount,
+    supportedBiometryType,
+    skipVerification,
+    numberAlreadyVerifiedCentrally,
+  } = props
+
+  switch (firstScreenInStep) {
```
jophish, line 147: Should this have a default case which logs an error or something similar?

src/onboarding/steps.ts
```diff
@@ -0,0 +1,214 @@
...
+            dispatch(setHasSeenVerificationNux(true))
+            navigateHome()
+          } else {
+            navigateClearingStack(Screens.VerificationStartScreen)
+          }
+        },
+      }
+    case Screens.EnableBiometry:
+      return {
+        next: () => {
```
jophish, line 180: Why is the return signature of this function an object with a single property? Can we just return the `next` function directly?

src/onboarding/steps.ts
```diff
@@ -0,0 +1,214 @@
...
+ * Traverses through the directed graph of onboarding navigate, navigateClearingStack, and navigateHome calls
+ * and count the number of screens until the given screen and the total number
+ */
+export function getOnboardingStepValues(screen: Screens, onboardingProps: OnboardingProps) {
+  let stepCounter = 1 // will increment this up to the onboarding step the user is on
+  let totalCounter = 1
+  let reachedScreeen = false // tracks whether we have reached the screen the user is on in onboarding, and we can stop incrementing stepCounter
+  let currentScreen: Screens = FIRST_ONBOARDING_SCREEN // pointer that we will update when simulating navigation through the onboarding screens to calculate "step" and "totalSteps"
+
+  const navigate = (s: Screens) => {
```
jophish, line 71: nit: perhaps adjust naming on `navigate` and `navigateHome` to clarify that they're dummy functions? Without looking at the definitions, I would think that `navigate` and `navigateHome` actually do the thing they claim to.

src/onboarding/steps.ts
```diff
@@ -0,0 +1,214 @@
...
+
+export interface OnboardingProps {
+  recoveringFromStoreWipe: boolean
+  choseToRestoreAccount: boolean | undefined
+  supportedBiometryType: BIOMETRY_TYPE | null
+  skipVerification: boolean
+  numberAlreadyVerifiedCentrally: boolean
+}
+
+export function onboardingPropsSelector(state: RootState): OnboardingProps {
```
jophish, line 45: Maybe add a docstring explaining that any data that parameterizes onboarding screen logic must be included in this function's result?

jophish (commented):
such a cool approach!!

PR #3507, opened 2023-03-07
fix: phone number format not recognised on verification screen

kathaypacific:
### Description

Context: https://valora-app.slack.com/archives/C029MMTD529/p1677719177909599, https://valora-app.slack.com/archives/C025V1D6F3J/p1678194369634539?thread_ts=1677719863.631589&cid=C025V1D6F3J

We use the value in the phone number text input to do the phone number validation, but as soon as the validation succeeds, we add the country code to the text input. Subsequent validations are performed with a phone number that includes the country code, which always fails. This PR fixes this issue.

### Test plan

This phone number can now be validated:
https://user-images.githubusercontent.com/20150449/223467814-ae6beaeb-6ccd-4dfb-8ca7-77c9fcb376e9.mp4

We can also backspace on a validated number:
https://user-images.githubusercontent.com/20150449/223469228-5e1232c3-de01-49d8-9cee-4020ad71a903.mp4

### Related issues

- Fixes RET-553, ENG-29, RET-646

### Backwards compatibility

Y

### DIFF
--- a/src/components/ValidatedTextInput.tsx
+++ b/src/components/ValidatedTextInput.tsx
@@ -22,6 +22,10 @@ interface OwnProps {
     | null
 }
 
+export interface PhoneValidatorProps {
+  validator: ValidatorKind.Phone
+  countryCallingCode: string
+}
 export interface IntegerValidatorProps {
   validator: ValidatorKind.Integer
 }
@@ -43,19 +47,17 @@ export type ValidatorProps =
   | DecimalValidatorProps
   | CustomValidatorProps
 
-export interface PhoneValidatorProps {
-  validator: ValidatorKind.Phone
-  countryCallingCode: string
-}
-
-export type ValidatedTextInputProps = OwnProps & PhoneValidatorProps & TextInputProps
+export type ValidatedTextInputProps = OwnProps & ValidatorProps & TextInputProps
 
 export default class ValidatedTextInput extends React.Component<ValidatedTextInputProps> {
   onChangeText = (input: string): void => {
     // some countries support multiple phone number lengths. since we add the
     // country code to a validated phone number in the UI, we need to omit the
     // country code from subsequent validations if the user types a longer number
-    const rawInput = input.split(this.props.countryCallingCode)[1] ?? input
+    const rawInput =
+      this.props.validator === ValidatorKind.Phone
+        ? input.split(this.props.countryCallingCode)[1] ?? input
+        : input
     const validated = validateInput(rawInput, this.props)
 
     // Don't propagate change if new change is invalid
@@ -68,7 +70,10 @@ export default class ValidatedTextInput extends React.Component<ValidatedTextInp
   }
 
   getMaxLength = () => {
-    const { numberOfDecimals, value, decimalSeparator } = this.props
+    const { numberOfDecimals, validator, value, decimalSeparator } = this.props
+    if (validator !== ValidatorKind.Decimal || !numberOfDecimals) {
+      return undefined
+    }
     const decimalPos = value.indexOf(decimalSeparator ?? '.')
     if (decimalPos === -1) {
       return undefined
--- a/src/components/ValidatedTextInput.tsx
+++ b/src/components/ValidatedTextInput.tsx
@@ -3,10 +3,10 @@
  */
 
 import { validateInput } from '@celo/phone-utils'
+import { ValidatorKind } from '@celo/utils/lib/inputValidation'
 import * as React from 'react'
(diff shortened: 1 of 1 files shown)


src/components/ValidatedTextInput.tsx
```diff
@@ -27,32 +27,16 @@ export interface PhoneValidatorProps {
...
-export type ValidatedTextInputProps = OwnProps & ValidatorProps & TextInputProps
+export type ValidatedTextInputProps = OwnProps & PhoneValidatorProps & TextInputProps
 
 export default class ValidatedTextInput extends React.Component<ValidatedTextInputProps> {
   onChangeText = (input: string): void => {
-    const validated = validateInput(input, this.props)
+    // some countries support multiple phone number lengths. since we add the
+    // country code to a validated phone number in the UI, we need to omit the
+    // country code from subsequent validations if the user types a longer number
+    const rawInput = input.split(this.props.countryCallingCode)[1] ?? input
```
kathaypacific, line 37: the value of `this.props.countryCallingCode` is something like `+82`

silasbw: some qs:
- why do we add the country code to the txt field?
- can the country code appear in the number `+82 10-1282-1234` -> `['+', ' 10-12', '-1234']`?

silasbw (commented):
If numbers can have different lengths, won't the "ghost text" we show be incorrect for some users?

src/components/ValidatedTextInput.tsx
kathaypacific: we add the country code to the text field after the user's input is validated. so entering a dutch phone number, i will see something like "061912345" in the text field until i enter the last digit, then the input will change to "+31 619-123-456". this is not a technical constraint, i think it's a product requirement - basically this update to show the full phone number is the cause of all our problems. fwiw i do think it's nice to show the full phone number with the country code, as it's not clear from the UI what the country code is otherwise.

i don't see a reason to make the data appear as `['+', ' 10-12', '-1234']` - i think you are suggesting that we use a separate value for the data rather than the display value, but we already do this further up the chain of components. we keep a nice object that describes the display number / country code / e164number, but this component is the lowest level display component, that responds to the user input. so to separate the displayed value from the input to this onChange handler would require an extra layer of state, i'm not sure it's worth the extra complexity?

kathaypacific:
> If numbers for a given country can have different lengths, won't the "ghost text" we show be incorrect for some users?

i'm not sure that i see the problem - we just want to give an example of what the phone number should look like, if the example number is valid for the user's country (i.e. doesn't look unfamiliar to them) but is a different length to the user's actual number, that still seems fine?

src/components/ValidatedTextInput.tsx
```diff
@@ -27,32 +27,16 @@ export interface PhoneValidatorProps {
...
-  numberOfDecimals: number
-}
-
-// Required props for a custom input validator
-export interface CustomValidatorProps {
-  validator: ValidatorKind.Custom
-  customValidator: (input: string) => string
-}
-
-export type ValidatorProps =
```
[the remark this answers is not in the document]
jeanregisser: This wasn't for the decentralized flow but for validating input when we used to have the real keyboard for entering amounts.

I'd have left that change out of this PR and focused on the phone number bug.
As is, the component name is now a bit misleading.

src/components/ValidatedTextInput.tsx
jeanregisser: Initially we were showing +XX next to the flag,
And only showing the local format in the text input.
I forgot why we switched though. Maybe some users were still entering +XX (because pasting?) or auto-complete from the OS.

jeanregisser (commented):
Thanks for tackling this 🚀

src/components/ValidatedTextInput.tsx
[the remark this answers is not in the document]
kathaypacific: that's true, i put back the unrelated changes :)

silasbw (approved):
Amazing :artist:

kathaypacific:
> This may warrant a unit test. Sorry I know it's extra annoying when the component doesn't have tests in the first place. We can follow up separately.

i agree, usually i'd add the tests before i merge but i haven't thought of a good way to test this yet. the bug only occurs while the user is typing, which is not possible (i think?) to reproduce with react native testing library since we only have `fireEvent.changeText` which dispatches a native event with the whole text. so a unit test isn't really helpful here :/ we could do it with an e2e test, but that also feels weird....although i lean towards having an e2e test

PR #3511, opened 2023-03-08
feat: better packager logs

jeanregisser:
### Description

Some improvements:
- Use time instead of timestamp: simpler for us, humans :)
- Aligned times: easier to parse visually
- Add device model with darker color: avoids confusion when multiple devices are connected to the packager. The darker color helps with visual parsing too.

### Test plan

**Packager logs (before)**

Screenshot 2023-03-07 at 19 26 15

**Packager logs (after)**

Screenshot 2023-03-08 at 11 43 07

### Related issues

- Fixes RET-648

### Backwards compatibility

Yes

### DIFF
--- a/src/utils/stylize.ts
+++ b/src/utils/stylize.ts
@@ -1,0 +1,30 @@
+const colors = {
+  bold: [1, 22],
+  italic: [3, 23],
+  underline: [4, 24],
+  inverse: [7, 27],
+  white: [37, 39],
+  grey: [90, 39],
+  black: [30, 39],
+  blue: [34, 39],
+  cyan: [36, 39],
+  green: [32, 39],
+  magenta: [35, 39],
+  red: [31, 39],
+  yellow: [33, 39],
+}
+
+/**
+ * Add colors to the specified string, for colorized output in terminals
+ */
+export function stylize(str: string, color: keyof typeof colors) {
+  if (!str) {
+    return ''
+  }
+
+  const codes = colors[color]
+  if (codes) {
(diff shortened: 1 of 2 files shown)


src/utils/stylize.ts
```diff
@@ -0,0 +1,30 @@
+const colors = {
```
silasbw, line 1: Have you heard my rant about "utils" yet :stuck_out_tongue: ?  "friends don't let friends dump things in utils/ or utils.ts"

my not really data-informed opinion: utils/ or utils.ts encourage poorly thought-out implementations and organization. they grow without bound.  so does their usage.  pretty soon everything is being imported from utils.  there's this sprawling mess of unit tests nobody wants to touch. people are extending the poorly thought-out things dumped into utils in poorly thought-out ways.

my not really data-informed solution: if I avoid putting stuff in "utils" it forces me to think a little bit more about the implementation and avoids the anarchy I describe above.

So anyways, nothing is really wrong with the code in stylize.ts, except that it encourages dumping into utils by example.  Consider moving it to Logger.ts (which I realize is also in utils, but maybe it shouldn't be either).

silasbw (approved):

src/utils/stylize.ts
jeanregisser: Ahah yes, I had it in Logger initially, but it didn't feel right either.
For me, utils serves as a "staging" zone. Code there may upgrade to a dedicated package in the future if it becomes generally useful to more than this project.
Usually it needs to be somewhere so it can be used from more than 1 place. I don't really have a better name than `utils` for this now.

PR #3514, opened 2023-03-08
docs: add documentation for upcoming EIP-747 feature

silasbw:


docs/watching-assets.mdx
```diff
@@ -0,0 +1,48 @@
+# Registering Tokens with Users
+
+> This feature is in-development and not fully supported yet!
+
+Valora includes a default list of popular tokens. Each token in the
+default list must meet the asset listing <a href={props.assetGuidlines}>requirements</a>.
```
silasbw, line 6: I didn't want to hardcode a relative URL (that's relative to the usage of this component in the docs repo -- that'd be confusing and brittle); nor did I want to hardcode an absolute URL (because it wouldn't work across environments, like staging and production)...so I made this an mdx and the user needs to pass in the `assetGuidelines` URL.

jeanregisser (approved):
🙌

PR #3516, opened 2023-03-08
fix(send): Dismiss keyboard on Send screen bottom sheet

jophish:
### Description

See [this Slack thread](https://valora-app.slack.com/archives/C04B61SJ6DS/p1678289764167129) for more context.
Currently, on the send flow, it's possible for the keyboard to be present _on top_ of the bottom sheet under certain circumstances. This PR explicitly dismisses the keyboard when the bottom sheet is made visible.

### Test plan

Manual tested and sort-of-e2e tested. Video below:

https://user-images.githubusercontent.com/569401/223771083-e8f97839-1b80-4a6d-93d2-d470d3c8e6ae.mp4

### Backwards compatibility

Yes.

### DIFF
--- a/src/send/Send.tsx
+++ b/src/send/Send.tsx
@@ -117,12 +117,15 @@ function Send({ route }: Props) {
       return
     }
 
+    // Dismiss the keyboard as soon as we know the verification status, so it does not
+    // interfere with the invite modal or bottom sheet.
+    Keyboard.dismiss()
+
     if (recipientVerificationStatus === RecipientVerificationStatus.UNVERIFIED) {
       setShowInviteModal(true)
       return
     }
 
-    Keyboard.dismiss()
     // Only show currency picker once we know that the recipient is verified,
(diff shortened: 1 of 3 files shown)


src/send/Send.tsx
```diff
@@ -122,6 +122,7 @@ function Send({ route }: Props) {
       return
     }
 
+    Keyboard.dismiss()
```
kathaypacific, line 125: what do you think about moving this to line 119, so that the keyboard is dismissed as soon as we know the verification status? the benefit would be, the keyboard is also dismissed when the invite modal appears. sorry, this is an existing issue

kathaypacific (approved):
one small suggestion, but otherwise great!! thank you!!

src/send/Send.tsx
jophish: ohh, interesting -- yeah, that sounds good to me, two birds one stone etc. etc. updated!

PR #3518, opened 2023-03-09
feat: swaps v2 execute screen with lottie

MuckT:
### Description

Updates the `src/swap/SwapExecuteScreen.tsx` to use Lottie animations and a bottom bar for navigation control in swap success and error states.

#### iOS Swap Error / Android Success
https://user-images.githubusercontent.com/26950305/224263448-178537e4-ebe9-4d2d-9e61-345721d9243b.mov

#### Android Success
https://user-images.githubusercontent.com/26950305/224260911-00eb62cb-dfcc-4955-a5eb-f84a002bcd4b.mov

### Test plan

- `SwapReviewScreen.test.tsx` updated.

### Related issues

- Fixes RET-610

### Backwards compatibility

Yes


src/icons/GreenLoadingSpinner.tsx
```diff
@@ -0,0 +1,16 @@
+import LottieView from 'lottie-react-native'
```
kathaypacific, line 1: i guess we'll be adding more lottie animations in the near future, and this icons folder is massive already (i personally find it very hard to find the icon that i need / verify that something i need already exists in this folder) - what do you think about creating a new folder for the lottie animation files? at least then the json files won't also live in this folder, i think then you will also be able to reuse the same name for the lottie file as the react file :)

src/swap/SwapExecuteScreen.tsx
```diff
@@ -31,13 +32,51 @@ export function SwapExecuteScreen() {
   }
 
   const navigateToSupport = () => {
-    replace(Screens.SupportContact)
+    navigate(Screens.SupportContact)
   }
 
-  const NavigateHomeIcon = () => {
+  const DoneAndSwapAgain = () => {
```
kathaypacific, line 38: `SwapCompleteNavigationButtons`?

src/swap/SwapExecuteScreen.tsx
```diff
@@ -118,33 +152,23 @@ export function SwapExecuteScreen() {
...
+  }, [swapState])
+
+  const NavigationButtons = useMemo(() => {
+    switch (swapState) {
+      case SwapState.ERROR:
+        return <TryAgainAndDone />
+      case SwapState.COMPLETE:
+        return <DoneAndSwapAgain />
+      default:
+        return <View style={[styles.actionBar, { minHeight: 130 }]} />
```
kathaypacific, line 171: is this 130 magic number to ensure that the animated loader stays in the same place when there are no buttons? this seems a little brittle (e.g. if we ever updated the size of the buttons, this seems like an easy place to break) - but i also don't think it works for the error scenario since you've got `contentContainer` to position content in the middle, the error content is longer than the pending content so the icon does move around.

i think to keep the icon in the same place, you need to have `contentContainer` position content with `flex-start` so that you can always control where the top of the icon is. this might mean that the content is not vertically positioned on the screen, but i think that is a worthwhile tradeoff. what do you think?

kathaypacific (commented):
looking great!!

src/swap/SwapExecuteScreen.tsx
```diff
@@ -118,33 +152,23 @@ export function SwapExecuteScreen() {
...
-            style={[styles.iconContainer, { backgroundColor: colors.greenUI }]}
-          >
-            <Checkmark color={colors.light} />
-          </View>
-        )
+        return <GreenLoadingSpinnerToCheck />
+    }
+  }, [swapState])
+
+  const NavigationButtons = useMemo(() => {
```
jeanregisser, line 164: This isn't a component, just a variable.
```suggestion
  const navigationButtons = useMemo(() => {
```

jeanregisser (commented):
Awesome! 👍 

My main concern is that the positions of the animations jump a little when changing state (from loading to error mostly).
It would be  better to ensure existing elements remain exactly at the same place when changing state.

Also, isn't the loading animation a bit too close to the text below?

src/swap/SwapExecuteScreen.tsx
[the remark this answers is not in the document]
MuckT: @jeanregisser  Does 501ffc814b1b854154e549ad93d8d4e5d01b2721 resolve this issue?

MuckT:
@jeanregisser @kathaypacific incorporated feedback on icon placement in bae4ef6763bff1749645b16cc4dd45c37275e063 🙏

kathaypacific (commented):
looking good, although with small screen (iPhone SE) + slightly large text, the error screen doesn't look quite right. idk what to do really, but it seems like we should enable scrolling if we want the buttons to always be visible. or think about decreasing the margin top since there's a lot of space :/ i'm not sure, maybe something you want to discuss with Reed?

Simulator Screen Shot - iPhone SE (3rd generation) - 2023-03-15 at 11 02 32

MuckT:
@jeanregisser @kathaypacific added 6a7f5e70581f26bd90ca41ba0d294c77eb1b68d0 which moves the icon up and prevents the contact support link from being obscured by the navigation buttons as seen [here](https://github.com/valora-inc/wallet/pull/3518#pullrequestreview-1341037186). 

### iOS - iPhone 13 Failure Flow
https://user-images.githubusercontent.com/26950305/225732312-adf86689-7002-42ab-aa2c-333b97e4b3ca.mov

locales/base/translation.json
```diff
@@ -1552,6 +1552,21 @@
...
     },
+    "swapCompleteSection": {
+      "title": "Swap successful",
+      "subtitle": "Your assets are now in your wallet",
+      "done": "Done"
+    },
+    "swapErrorSection": {
+      "title": "We couldn’t complete your swap",
+      "subtitle": "Your assets are safe in your wallet.\nPlease try again.",
+      "contactSupport": "If you need additional help, you can contact\n<0>customer support<0>"
```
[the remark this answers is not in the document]
kathaypacific: yeah up to what you and reed decide - i thought that having a line break between the link is an okay experience, if anything it gives the link a bigger touch target

src/swap/SwapExecuteScreen.tsx
```diff
@@ -31,18 +32,59 @@ export function SwapExecuteScreen() {
...
+      case SwapState.COMPLETE:
+        return (
+          <View style={styles.actionBar}>
+            <Button
+              text={t('SwapExecuteScreen.swapActionBar.done')}
+              onPress={navigateHome}
+              type={BtnTypes.PRIMARY}
+              size={BtnSizes.FULL}
+              testID="SwapExecuteScreen/Done"
+              style={[styles.button, { paddingBottom: Spacing.Regular16 }]}
```
kathaypacific, line 70: the paddingBottom style already exists in the button style so it can be removed here :)

kathaypacific (approved):
looks good!! 

you've probably done this already but worth checking with Reed about the placement of the content now that it's further up the screen (on larger devices it looks a little unbalanced but that's just my subjective opinion) - the scrollview on the content fixes the layout issues on smaller devices so i don't think the content strictly needed to move up, it just depends on which device size we want to optimise the layout for :) i'm fine with how this looks now too, just wanted to make sure that we're optimising intentionally.

PR #3520, opened 2023-03-09
fix(wc): don't crash WC sagas when the icons property is incorrect

jeanregisser:
### Description

This is an immediate fix to prevent the WC sagas from crashing (and not handling subsequent actions) when  the `icons` property of the session is incorrect.

There will be a follow up PR to make sure the WC sagas (and other sagas throughout the app) aren't cancelled when an unexpected error is thrown.

This is one of the cases where the compile time TS types aren't enough.
I decided to apply the fix directly where the incorrect data is, in the received session, so the rest of the code doesn't need to be aware of this (UI and other handlers).

### Test plan

- Added unit test
- Manually ensuring the WC e2e tests now work when [`icons: undefined`](https://github.com/valora-inc/wallet/blob/0d40793f5705f7dcb0c1816f91e48ed22f97ebf1/e2e/src/usecases/WalletConnectV2.js#L116) is set

### Related issues

- Fixes RET-654

### Backwards compatibility

Yes


src/walletConnect/v2/saga.test.ts
```diff
@@ -95,3 +108,39 @@ describe('getDefaultSessionTrackedProperties', () => {
       .run()
   })
 })
+
+describe('applyIconFixIfNeeded', () => {
```
jeanregisser, line 112: Only added a test for WCv2, since WCv1 is deprecated.

src/walletConnect/v1/saga.ts
```diff
@@ -299,6 +301,17 @@ function* listenForWalletConnectMessages(walletConnectChannel: EventChannel<Wall
   }
 }
 
+// Though the WC types say `icons` is string[], we've seen buggy clients with no `icons` property
+// so to avoid crashing the code depending on this, we fix it here
+// Note: this method mutates the session
+function applyIconFixIfNeeded(session: WalletConnectSessionRequest | WalletConnectSession) {
+  const { peerMeta } = 'peerMeta' in session ? session : session.params[0]
```
kathaypacific, line 308: is there any chance that `session.params[0]` is undefined? 🙈 it's probably not likely but if we default to an empty object then this function shouldn't ever fail? idk, maybe overkill

kathaypacific (approved):
i love it! thank you for digging into this! 👨🏼‍🎨

src/walletConnect/v1/saga.ts
jeanregisser: Good question, I think it should fail if this happens.
We won't be able to do anything with that request and it should propagate the error.

PR #3521, opened 2023-03-09
chore: use the project's TypeScript version instead of the one bundled with VS Code

jeanregisser:
This avoids seeing incorrect TS issues in VS Code.

Taken from https://github.com/valora-inc/typescript-app-starter/pull/33


kathaypacific (approved):
❣️

PR #3499, opened 2023-03-06
feat: make gray text accessible on white

jeanregisser:
### Description

This PR updates the color palette so gray text is now accessible on white.
`gray3`, `gray4`, `gray5` which were used for text were previously not accessible on white.

Plus, also tweaked `gray2` so it's slightly darker.

### Test plan

I provided a test APK for Reed to check and approve the changes throughout the app.

**before/after**

### Related issues

Reed and Nitya decided this updated palette.

### Backwards compatibility

Yes

PR #3522, opened 2023-03-09
fix(dapps): tweak dapps title and description to match the designs

jeanregisser:
### Description

Following https://github.com/valora-inc/wallet/pull/3499, the dapps screen didn't look good anymore.

Plus, added an `xsmall` font style for places where we need 12px text.

Following the 

### Test plan

**before/after**

### Related issues

- Discussed with Reed on Slack

### Backwards compatibility

Yes


MuckT (approved):
Thanks for adding the xsmall font!

PR #3524, opened 2023-03-09
chore(onboarding): Set up mock function to return experiment parameters

jophish:
### Description

For ACT-638. Just adds in a dummy function to return the set of feature flags we care about for the onboarding/recovery phrase experiment. Wasn't sure where to put it, so added a new `index.ts` file in `src/onboarding`, happy to move it if folks think there's a better home for it.

### Test plan

N/A

### Related issues

- Fixes ACT-638.

### Backwards compatibility

N/A


src/onboarding/index.ts
```diff
@@ -0,0 +1,20 @@
+export interface OnboardingExperimentParams {
+  disableForcedBackup: boolean // Whether to disable forced 24h backup.
```
cajubelt, line 2: nit: having this be a "disable" flag makes it a little harder to reason about imo. It's like a double negative-- to enable something, I need to not not enable it. I mention this because I think it makes bugs more likely. 

the only reason I'd want to keep it as a "disable" is if there was an existing feature flag that it was overriding. I don't think that's the case, though-- I think forced backup is hardcoded

src/onboarding/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  showRecoveryPhraseInOnboarding: boolean // Whether to show recovery phrase and backup options in onboarding.
+  showCloudBackupFakeDoor: boolean // Whether to show cloud backup as a fakedoor recovery option. No effect if showRecoveryPhraseInOnboarding is false.
+  useNewBackupFlowCopy: boolean // Whether to use new copy in the backup flow.
```
cajubelt, line 5:
```suggestion
  useNewBackupFlowCopy: boolean
```

src/onboarding/index.ts
```diff
@@ -0,0 +1,20 @@
+export interface OnboardingExperimentParams {
+  disableForcedBackup: boolean // Whether to disable forced 24h backup.
+  showRecoveryPhraseInOnboarding: boolean // Whether to show recovery phrase and backup options in onboarding.
+  showCloudBackupFakeDoor: boolean // Whether to show cloud backup as a fakedoor recovery option. No effect if showRecoveryPhraseInOnboarding is false.
```
cajubelt, line 4:
```suggestion
  showCloudBackupFakeDoor: boolean // No effect if showRecoveryPhraseInOnboarding is false.
```

src/onboarding/index.ts
```diff
@@ -0,0 +1,20 @@
+export interface OnboardingExperimentParams {
+  disableForcedBackup: boolean // Whether to disable forced 24h backup.
+  showRecoveryPhraseInOnboarding: boolean // Whether to show recovery phrase and backup options in onboarding.
```
cajubelt, line 3:
```suggestion
  showRecoveryPhraseInOnboarding: boolean
```

src/onboarding/index.ts
cajubelt, line 2:
```suggestion
  disableForcedBackup: boolean
```

src/onboarding/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  showCloudBackupFakeDoor: boolean // Whether to show cloud backup as a fakedoor recovery option. No effect if showRecoveryPhraseInOnboarding is false.
+  useNewBackupFlowCopy: boolean // Whether to use new copy in the backup flow.
+  showBackupAlert: boolean // Whether to show an alert in the sidenav if the user hasn't backed up.
+  useNewBackupHomecard: boolean // Whether to use new designs for backup homecard.
```
cajubelt, line 7:
```suggestion
  useNewBackupHomecard: boolean
```

cajubelt (approved):
deleted some comments that I found redundant with the feature flag names

PR #3525, opened 2023-03-09
feat(fiatconnect): add PIX account schema

anthonydaoud:
### Description

Currently, the PIX account schema is not supported in the wallet. This PR adds the PIX account schema to the set of FiatConnect schemas, which allows the account details screen to show the relevant inputs. This PR does not yet add PIX account to the `SUPPORTED_FIAT_ACCOUNT_SCHEMAS` set.

 
https://user-images.githubusercontent.com/20508929/224419726-8d79ffd8-3c9a-437a-85d5-9ff2265cb7a5.mp4

### Test plan

Added unit tests.

### Related issues

- ACT-656

### Backwards compatibility

Yes.


src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -217,17 +220,17 @@ const FiatDetailsScreen = ({ route, navigation }: Props) => {
...
-                  errorMessage={errors.get(index)}
-                  onChange={(value) => {
-                    setInputValue(value, index)
+                  value={fieldNamesToValues.current[field.name]}
+                  errorMessage={errors.get(field.name)}
+                  onChange={(fieldName, value) => {
+                    setInputValue(fieldName, value)
                   }}
                   allowedValues={quote.getFiatAccountSchemaAllowedValues(field.name)}
+                  fieldNamesToValues={fieldNamesToValues.current}
```
jh2oman, line 233: Would it be more straightforward to just have a `isVisible` prop on FormField?

cajubelt: ➕

src/fiatExchanges/quotes/FiatConnectQuote.ts
```diff
@@ -236,7 +236,9 @@ export default class FiatConnectQuote extends NormalizedQuote {
   }
 
   getFiatAccountSchemaAllowedValues(key: string): string[] | undefined {
-    return this.quoteResponseFiatAccountSchema.allowedValues[key]
+    const schemaDefaultValues = DEFAULT_ALLOWED_VALUES[this.getFiatAccountSchema()]
+    const defaultValue = schemaDefaultValues?.[key as keyof typeof schemaDefaultValues]
+    return this.quoteResponseFiatAccountSchema.allowedValues[key] ?? defaultValue
```
jh2oman, line 241: can you add a unit test for the case where the default value is used

src/fiatconnect/FiatDetailsScreen.tsx
```diff
@@ -177,24 +175,29 @@ const FiatDetailsScreen = ({ route, navigation }: Props) => {
...
       }
     })
 
     setErrors(newErrorMap)
     setValidInputs(newErrorMap.size === 0)
   }
 
-  const setInputValue = (value: string, index: number) => {
-    fieldValues.current[index] = value
+  const setInputValue = (fieldName: string, value: string) => {
```
cajubelt, line 199:
```suggestion
  const setInputValue = ({fieldName, value}: {fieldName: string, value: string}) => {
```
this will make it harder to introduce bugs where you swap the order of the arguments by accident

cajubelt: i prefer to use destructuring whenever there are 2 consecutive positional arguments with the same type, since static type checks will not prevent this kind of bug otherwise

src/fiatconnect/fiatAccountSchemas/pixAccount.ts
```diff
@@ -0,0 +1,69 @@
...
+        break
+      case PIXKeyTypeEnum.PHONE:
+        regex = PIX_PHONE_KEY_REGEX
+        break
+      case PIXKeyTypeEnum.RANDOM:
+        regex = UUID_REGEX
+        break
+      default:
+        // Occurs when keyType isn't selected
+        regex = /.*/g
```
cajubelt, line 35: if keyType isn't selected, key could still have invalid values-- anything that isn't a uuid, phone, cpf, or email. seems like a union of all the other regexes might be better for this

src/fiatconnect/fiatAccountSchemas/pixAccount.ts
```diff
@@ -0,0 +1,69 @@
...
+} from '@fiatconnect/fiatconnect-types'
+import { getObfuscatedAccountNumber } from 'src/fiatconnect'
+import { INSTITUTION_NAME_FIELD } from 'src/fiatconnect/fiatAccountSchemas'
+import { FiatAccountFormSchema } from 'src/fiatconnect/fiatAccountSchemas/types'
+import i18n from 'src/i18n'
+
+const UUID_REGEX = /^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$/g
+
+export const getPixAccountSchema = (): FiatAccountFormSchema<FiatAccountSchema.PIXAccount> => {
+  const keyValidation = (input: string, fieldNamesToValues: Record<string, string>) => {
```
cajubelt, line 17: probly worth splitting this out so it can be unit tested

src/fiatconnect/fiatAccountSchemas/pixAccount.ts
```diff
@@ -0,0 +1,69 @@
...
+      placeholderText: i18n.t('fiatAccountSchema.pix.key.placeholderText'),
+      validate: keyValidation,
+      keyboardType: 'default',
+      isVisible: (fieldNamesToValues: Record<string, string>) => !!fieldNamesToValues['keyType'],
+    },
+    fiatAccountType: { name: 'fiatAccountType', value: FiatAccountType.BankAccount },
+    accountName: {
+      name: 'accountName',
+      computeValue: ({ institutionName, key = '', keyType }) =>
+        `${institutionName} (${getObfuscatedAccountNumber(key)})`,
```
cajubelt, line 66: for email, getObfuscatedAccountNumber isn't very helpful since it just shows the last 4 digits (for instance, `.com`)

cajubelt: probly need a new helper for emails that focuses on obfuscating the stuff before the `@` character

anthonydaoud: Added an email obfuscation function

src/fiatExchanges/quotes/FiatConnectQuote.test.ts
```diff
@@ -22,6 +22,20 @@ jest.mock('src/web3/contracts', () => ({
...
   })),
 }))
+jest.mock('src/fiatExchanges/quotes/constants', () => {
+  const originalModule = jest.requireActual('src/fiatExchanges/quotes/constants')
+
+  return {
+    ...originalModule,
+    DEFAULT_ALLOWED_VALUES: {
+      // Using AccountNumber because jest hoisting prevents us from using the
+      // FiatAccountSchema enum.
```
cajubelt, line 32: fwiw I'm usually fine with hardcoding things in test code since it keeps it more independent from the rest of the codebase, and less circular. this can actually help prevent bugs-- for instance, if you updated some constant to a buggy value, and then re-used that constant in test code, you'd be in trouble. But if the test code has some hardcoded version of it that's supposed to match, it's less likely that you made the same typo twice, and you might notice the new discrepancy from a failing test

src/fiatconnect/index.test.ts
```diff
@@ -271,4 +272,20 @@ describe('FiatConnect helpers', () => {
       expect(getObfuscatedAccountNumber('1')).toEqual('')
     })
   })
+
+  describe('getObfuscatedEmail', () => {
```
cajubelt, line 276: nice coverage

cajubelt (approved):
🥳

PR #3526, opened 2023-03-10
chore: add `safely` helper for sagas + documentation

jeanregisser:
### Description

This is a first PR to address the problem of unhandled errors in (some) sagas bubbling up with unexpected consequences.

Please see the new documentation added for the details.

Also see #3527 where it's used.

### Test plan

- Added unit tests

### Related issues

- Fixes RET-654

### Backwards compatibility

Yes


WALLET.md
```diff
@@ -592,6 +593,57 @@ If you're deleting or updating existing properties, please implement the appropr
 6. Optional: if the migration is not trivial, add a test for it in [src/redux/migrations.test.ts](src/redux/migrations.test.ts)
 7. Commit the changes
 
+### Redux-Saga pitfalls
+
+### Error bubbling
+
+It's important to understand how errors propagates with Redux-Saga.
```
kathaypacific, line 600:
```suggestion
It's important to understand how errors propagate with Redux-Saga.
```

WALLET.md
```diff
@@ -592,6 +593,57 @@ If you're deleting or updating existing properties, please implement the appropr
...
+
+function* someOtherSaga() {
+  // [...]
+}
+```
+
+If an exception is thrown from `sendPayment` or `notifyUser`, the whole `mySaga` will be cancelled.
+And won't handle `SEND_PAYMENT` AND `NOTIFY_USER` actions until the app is restarted.
+
+Since `mySaga` was spawned from the root saga, `someOtherSaga` won't be affected though.
```
kathaypacific, line 623: just to clarify, `mySaga` is not affected because the root saga isn't the direct parent saga of the error? or is there something special about the root saga, where it is not affected by errors in the same way other sagas would be? e.g. if the root saga instead looked like 
```
function* rootSaga() {
  yield takeEvery('SEND_PAYMENT', sendPayment)
  yield spawn(someOtherSaga)
}
```
and `sendPayment` throws, the error would get bubbled to the `rootSaga` and in this case, `someOtherSaga` would be affected (not spawned) right?

kathaypacific (approved):
LGTM!! ✨🚀

WALLET.md
jeanregisser: Good question, it's because it uses [`spawn`](https://redux-saga.js.org/docs/api/#spawnfn-args) which isolates it from the other tasks. 
I'll tweak the wording 👍 

In your example, if `sendPayment` throws, `someOtherSaga` would NOT be affected because it uses `spawn`.
`takeEvery` and `spawn` are not blocking, both of these would be applied. By the time a `SEND_PAYMENT` action is dispatched and fails in `sendPayment`, `someOtherSaga` would already have been spawned.

PR #3527, opened 2023-03-10
fix(wc): prevent unhandled errors from cancelling all WC action handlers

jeanregisser:
### Description

This PR fixes the issue observed where any unhandled error in a WC action handler would cancel all other listeners for WC actions. The only way to recover was to restart the app.

It uses the `safely` helper added in #3526.

Note: there are other places in the codebase suffering from the same issue. We can open additional PRs for them.

### Test plan

- Added unit test

### Related issues

- Fixes RET-654

### Backwards compatibility

Yes

### DIFF
--- a/src/walletConnect/v1/saga.ts
+++ b/src/walletConnect/v1/saga.ts
@@ -4,7 +4,7 @@ import '@react-native-firebase/messaging'
 import WalletConnectClient from '@walletconnect/client'
 import { IWalletConnectOptions } from '@walletconnect/legacy-types'
 import { EventChannel, eventChannel } from 'redux-saga'
-import { call, fork, put, select, take, takeEvery } from 'redux-saga/effects'
+import { call, fork, put, select, spawn, take, takeEvery } from 'redux-saga/effects'
 import { WalletConnectEvents } from 'src/analytics/Events'
 import { WalletConnect1Properties } from 'src/analytics/Properties'
 import { WalletConnectPairingOrigin } from 'src/analytics/types'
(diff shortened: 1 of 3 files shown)


src/walletConnect/v1/saga.ts
```diff
@@ -497,18 +498,18 @@ function* checkPersistedState(): any {
...
+  yield takeEvery(Actions.ACCEPT_REQUEST_V1, safely(acceptRequest))
+  yield takeEvery(Actions.DENY_REQUEST_V1, safely(denyRequest))
 
-  yield takeEvery(Actions.SESSION_V1, handleSessionRequest)
-  yield takeEvery(Actions.PAYLOAD_V1, handlePayloadRequest)
+  yield takeEvery(Actions.SESSION_V1, safely(handleSessionRequest))
+  yield takeEvery(Actions.PAYLOAD_V1, safely(handlePayloadRequest))
 
-  yield call(checkPersistedState)
+  yield spawn(checkPersistedState)
```
jeanregisser, line 512: Changing to spawn so an unhandled error won't cancel the whole saga.

src/walletConnect/v2/saga.ts
```diff
@@ -551,19 +562,19 @@ function* checkPersistedState() {
...
 
-  yield takeEvery(Actions.SESSION_PAYLOAD_V2, handleIncomingActionRequest)
-  yield takeEvery(Actions.ACCEPT_REQUEST_V2, handleAcceptRequest)
-  yield takeEvery(Actions.DENY_REQUEST_V2, handleDenyRequest)
+  yield takeEvery(Actions.SESSION_PAYLOAD_V2, safely(handleIncomingActionRequest))
+  yield takeEvery(Actions.ACCEPT_REQUEST_V2, safely(handleAcceptRequest))
+  yield takeEvery(Actions.DENY_REQUEST_V2, safely(handleDenyRequest))
 
-  yield call(checkPersistedState)
+  yield spawn(checkPersistedState)
```
jeanregisser, line 577: Same here.

kathaypacific (approved):
👨🏼‍🎨

PR #3534, opened 2023-03-14
refactor(statsig): helper for getting experiment parameters

cajubelt:
### Description

going for more type safety and less boiler plate.

### Test plan

- automated test on the new helpers
- navigated to select providers screen and confirmed that no warning is logged and the QR screen is shown (retrieved from Statsig successfully)

### Related issues

cc https://linear.app/valora/issue/ACT-659/statsig-abstraction (not done since this doesn't cover Layer params yet)

### Backwards compatibility

na


src/onboarding/index.ts
```diff
@@ -1,20 +1,7 @@
-export interface OnboardingExperimentParams {
```
cajubelt, line 1: the return type is the same, but I don't think it needs to be duplicated here. it's inferred from the defaultValues field on the experiment config

src/statsig/constants.ts
```diff
@@ -15,15 +15,24 @@ export const LayerParams = {
   },
 }
 
-export const ExperimentParams = {
+export const ExperimentConfigs = {
```
satish-ravi, line 18: is it possible to define a type for this that ensures the key and value of experimentName field are the same?

src/statsig/index.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+  describe('getExperimentParams', () => {
+    it('returns default values if getting statsig experiment throws error', () => {
+      ;(Statsig.getExperiment as jest.Mock).mockImplementation(() => {
+        throw new Error('mock error')
+      })
+      const defaultValues = { param1: 'defaultValue1', param2: 'defaultValue2' }
+      const experimentName = StatsigExperiments.RECOVERY_PHRASE_IN_ONBOARDING // chosen arbitrarily
```
satish-ravi, line 19: nit: maybe mock constants/types and use a test specific experiment if its easy to do? We will likely remove experiment specific code in the future right?

satish-ravi (approved):
Looks good. Just a couple of non blocking comments

src/statsig/index.ts
jh2oman, line 1: Elegante 💃

PR #3509, opened 2023-03-07
feat: logger improvements

jeanregisser:
### Description

This fixes a couple of issues with our logger:
- log file not containing the full info, only the first argument was written to the file (while showing the full line in the packager logs).
- warn and error levels treated as info. Making it easy to miss them in the packager logs. Though they were correctly already being written to the log file as `error` or `warn`
- aligns the data in the log file so it's easier to parse visually

### Test plan

**Log file (before)**:

```
info [2023-03-07T18:16:57.964Z] sentry/Sentry/Sentry not enabled
debug [2023-03-07T18:16:58.743Z] ValoraAnalytics/Queued call to identify, 0xa6d1e0bdb6960c3f1bda8ef8f1e91480cfc40dbb, [object Object]
error [2023-03-07T18:16:58.744Z] The above error occurred in task appVersionSaga

debug [2023-03-07T18:16:58.744Z] app/saga@monitorAppState/Starting monitor app state saga
debug [2023-03-07T18:16:58.744Z] ValoraAnalytics/Queued call to track, init_contractkit_start
info [2023-03-07T18:16:58.745Z] web3/contracts@initContractKit/Initializing contractkit
info [2023-03-07T18:16:58.745Z] web3/contracts@initContractKit/Initializing wallet, {"address":"0xa6d1e0bdb6960c3f1bda8ef8f1e91480cfc40dbb","createdAt":"2023-03-06T12:01:38.576Z"}
debug [2023-03-07T18:16:58.745Z] ValoraAnalytics/Queued call to track, init_contractkit_get_wallet_start
debug [2023-03-07T18:16:58.745Z] ValoraAnalytics/Queued call to track, init_contractkit_get_wallet_finish
debug [2023-03-07T18:16:58.747Z] web3/dataEncryptionKey@registerAccountDek/Skipping DEK registration because its already registered
```

**Log file (after)**:

```
INFO  [2023-03-07T18:23:25.607Z] sentry/Sentry/Sentry not enabled
DEBUG [2023-03-07T18:23:26.572Z] ValoraAnalytics/Queued call to identify, 0xa6d1e0bdb6960c3f1bda8ef8f1e91480cfc40dbb, [object Object]
ERROR [2023-03-07T18:23:26.572Z] [Error: You attempted to use 'firebase.database' but this module could not be found.

Ensure you have installed and imported the '@react-native-firebase/database' package.]
ERROR [2023-03-07T18:23:26.572Z] The above error occurred in task appVersionSaga

[...]

PR #3535, opened 2023-03-14
fix: error in appVersionSaga

jeanregisser:
### Description

This fixes the error in `appVersionSaga` which became more apparent from the changes done in #3509.

```
[Error: You attempted to use 'firebase.database' but this module could not be found.

Ensure you have installed and imported the '@react-native-firebase/database' package.]
```

### Test plan

**before/after**

 

Still a bunch of other warnings to address 🙂

### Related issues

- Fixes RET-648

### Backwards compatibility

Yes


src/firebase/firebase.ts
```diff
@@ -1,5 +1,6 @@
 import firebase, { ReactNativeFirebase } from '@react-native-firebase/app'
 import '@react-native-firebase/auth'
+import '@react-native-firebase/database'
```
jeanregisser, line 3: This is the fix.

src/transactions/saga.ts
```diff
@@ -1,8 +1,6 @@
 import { CeloTransactionObject, CeloTxReceipt, EventLog } from '@celo/connect'
 import { ContractKit } from '@celo/contractkit'
 import { EscrowWrapper } from '@celo/contractkit/lib/wrappers/Escrow'
-import '@react-native-firebase/database'
-import '@react-native-firebase/messaging'
```
jeanregisser, line 5: I removed these additional imports which shouldn't be needed.

kathaypacific (approved):
🙏🏻

PR #3538, opened 2023-03-14
feat(swaps): Updating default tokens in the swap screen

dievazqu:
### Description

Change the logic for the default tokens in the swap screen. Now the `fromToken` will be the token with the highest usd balance, while the `toToken` will be empty.

Special cases:
* Choosing the `fromToken` when selecting the `toToken`, will still result in a swap of tokens (from <-> to). If there is no token selected in the `toToken`, then after the swap there won't be a token selected in the `fromToken`.
* If the user doesn't have balance in any of the supported token, it will default to use celo as `fromToken`

### Test plan

I tested manually in the emulator and with the unit tests.

### Related issues

- Fixes RET-577

### Backwards compatibility

It's backward compatible

### Changed view

https://user-images.githubusercontent.com/12504748/225036871-68ebce98-0084-4137-ac6d-2d645f7fab37.mov


MuckT:
@dievazqu I am getting some zero balances back in the token bottom sheet, but this appears to be an existing issue.

src/swap/SwapScreen.tsx
```diff
@@ -46,15 +46,23 @@ export function SwapScreen() {
   const { t } = useTranslation()
   const dispatch = useDispatch()
 
-  const coreTokens = useSelector(coreTokensSelector)
+  const supportedTokens = useSelector(coreTokensSelector)
```
MuckT, line 49: This will change when we have the cloud function correct?

src/swap/SwapScreen.tsx
```diff
@@ -46,15 +46,23 @@ export function SwapScreen() {
...
-  )
-  const [fromToken, setFromToken] = useState(
-    coreTokens.find((token) => token.symbol === DEFAULT_FROM_TOKEN)
+  const CELO = supportedTokens.find(
+    (token) => token.symbol === DEFAULT_FROM_TOKEN && token.isCoreToken
   )
+
+  const defaultFromToken =
+    tokensSortedByUsdBalance.find(
+      (token) => token.balance.gt(0) && supportedTokensAddresses.includes(token.address)
```
MuckT, line 60: We might higher default for swappable assets something like percent of asset price, but that might be outside of the scope of this PR.

MuckT (approved):
This looks great! I have a few questions around handling low balances in swaps that are outside of the scope of this PR in the comments. Tested locally on Android and iOS.

src/fees/hooks.ts
```diff
@@ -79,6 +80,9 @@ export function useMaxSendAmount(
   const feeEstimate =
     useUsdToTokenAmount(new BigNumber(usdFeeEstimate ?? 0), tokenAddress) ?? new BigNumber(0)
 
+  if (!balance) {
+    return new BigNumber(0)
```
kathaypacific, line 84: why do we need changes to the fee estimation hook for this feature?

src/swap/SwapAmountInput.tsx
```diff
@@ -43,6 +44,7 @@ const SwapAmountInput = ({
   autoFocus,
   inputError,
   style,
+  noTokenDisplay,
```
kathaypacific, line 47: `buttonPlaceholder`? or `placeholder`?

src/swap/SwapScreen.tsx
```diff
@@ -46,15 +46,23 @@ export function SwapScreen() {
   const { t } = useTranslation()
   const dispatch = useDispatch()
 
-  const coreTokens = useSelector(coreTokensSelector)
+  const supportedTokens = useSelector(coreTokensSelector)
+  const supportedTokensAddresses = supportedTokens.map((token) => token.address)
```
kathaypacific, line 50: this is only used in one place, so maybe we don't need to declare it here?

variables are generally recreated when the component rerenders so depending on how big the list of supported tokens is, it might be something we want to wrap in `useMemo` to avoid doing the loop all the time. it doesn't matter much right now with 4 tokens of course, it just got me thinking that maybe we don't need this variable, and instead can use `supportedTokens` in `defaultFromToken` and wrap `defaultFromToken` in a `useMemo`. what do you think?

kathaypacific (approved):
Looks great!!! ✨

src/swap/SwapScreen.tsx
dievazqu: I'm not sure what cloud function you are referring to, but I think we will need to update the selector to return only the assets that we support for swapping.
https://linear.app/valora/issue/RET-611/[wallet]-more-tokens-support

src/swap/SwapScreen.tsx
dievazqu: The `tokensSortedByUsdBalance` already returns a sorted list of token by usd balance. So this will pick up the supported token with the highest usd balance.
So the `token.balance.gt(0)` is just to avoid the case where the user doesn't have any balance, so we should show CELO as the default token.
Now thinking about it, I think this could introduce a bug, only when the user has supported tokens but we are not able to fetch the price for them. So the sorting won't work correctly, and in that case we will show any of supported tokens.
I will add a `token.balance.gt(0) && token.usdPrice && supportedTokensAddresses.includes(token.address)` to address that case.

src/swap/SwapScreen.tsx
dievazqu: Yep, that makes a lot of sense, I'll do that :)

src/fees/hooks.ts
dievazqu: Yeah, to be honest, I'm a bit concern about this change!
Since this feature, it's possible that we call the `useMaxSendAmount` method with an empty string which throws an error with the old hook. Then I realized the hook was casting a possible `undefined` value, and I thought that wasn't good.
So I think it's ok if this hook returns 0 when the token is invalid instead of crashing the app.

I would really love to avoid calling this hook when we know there is no token, but I can't put the hook under a condition block, right? So I guess there's no other solution than changing the hook? 
Maybe I'm missing an easier solution 😅

kathaypacific: ah i see. this hook is getting kind of crazy, i think because the use cases are becoming too broad. there is already a place inside the `useEffect` that handles falsy balance, so i was confused about why we need another place. i think i understand the problem, could you provide a default for the balance? like `  const balance = useTokenInfo(tokenAddress)?.balance ?? new BigNumber(0)`

after talking about this, should we also hide the max button if there is no from token selected? up to you, on one hand i don't think the button is helpful but on the other hand idk if it's worth the extra logic because the button won't break 🤷🏻‍♀️

PR #3540, opened 2023-03-14
feat(onboarding): Add new onboarding recovery phrase screens

jophish:
### Description

For ACT-636/ACT-637.

First, apologies for the big PR -- I was working on ACT-637 and realized too late that it _only_ included the first screen, and not both... If this feels cumbersome to review, I can split it out into two separate PRs.

Adds in two new screens into the onboarding flow; the `ProtectWallet` and `OnboardingRecoveryPhrase` screens, per the Figma designs [here](https://www.figma.com/file/gGp6SVFWMbnSltkGzNzrpC/Wallet-Recovery?node-id=1%3A3&t=SbxOuTMooFp7Z67f-0).

These screens require account initialization to occur earlier in the onboarding flow than normal. As a result, a spinner is briefly shown on the first `ProtectWallet` screen, as the user's account may take a moment to initialize. 

### Test plan

Unit and manual tested:

#### Recovery Phrase & Cloud Fakedoor

https://user-images.githubusercontent.com/569401/225067340-b8a230b8-a930-41f8-8f27-027868fa6889.mp4

#### Recovery Phrase Only

https://user-images.githubusercontent.com/569401/225067268-53c20894-8678-4efa-8810-2edcae55bffc.mp4

### Related issues

- Fixes ACT-636, Fixes ACT-637.

### Backwards compatibility

Yes.


src/onboarding/registration/OnboardingRecoveryPhrase.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+    fireEvent.press(getByTestId('protectWalletBottomSheetContinue'))
+    expect(ValoraAnalytics.track).toHaveBeenCalledTimes(1)
+    expect(ValoraAnalytics.track).toHaveBeenCalledWith(OnboardingEvents.protect_wallet_complete)
+    expect(goToNextOnboardingScreen).toBeCalledWith({
+      firstScreenInCurrentStep: Screens.ProtectWallet,
+      onboardingProps: mockOnboardingProps,
+    })
+  })
+  // eslint-disable-next-line jest/no-disabled-tests
+  it.skip('opens bottom sheet when help pressed', async () => {
```
jophish, line 54: I've never been able to figure out how to test the behavior of logic that occurs within `useLayoutEffect` hooks; I wasn't able to find other examples of this being done elsewhere in the app, unfortunately..

src/onboarding/registration/ProtectWallet.tsx
```diff
@@ -0,0 +1,241 @@
...
+import { UNLOCK_DURATION } from 'src/web3/consts'
+import variables from 'src/styles/variables'
+
+type Props = NativeStackScreenProps<StackParamList, Screens.ProtectWallet>
+
+function getCloudBackupIndex(seed: string | null) {
+  if (!seed) {
+    return 1
+  } else {
+    return seedrandom(seed)() > 0.5 ? 0 : 1
```
jophish, line 36: `Math.random` doesn't allow you to provide a seed to get repeatable randomness, and I wasn't able to find anything easily accessible in libraries we currently used, so brought this in as a simple solution.

src/onboarding/registration/ProtectWallet.tsx
```diff
@@ -0,0 +1,241 @@
...
+  },
+  topSection: {
+    alignItems: 'center',
+  },
+  buttonStyle: {
+    paddingTop: 30,
+  },
+  bottomSheetTitle: {
+    ...fontStyles.h2,
+    marginTop: 22,
```
jh2oman, line 211: margin looks too high

jh2oman (commented):
A few minor style nits to start. looking at the rest now

src/onboarding/registration/ProtectWallet.tsx
```diff
@@ -0,0 +1,241 @@
...
+  const onboardingProps = useTypedSelector(onboardingPropsSelector)
+  const { step, totalSteps } = getOnboardingStepValues(Screens.ProtectWallet, onboardingProps)
+  const address = useSelector(walletAddressSelector)
+  const [showBottomSheet, setShowBottomSheet] = useState(false)
+  const [shouldNavigate, setShouldNavigate] = useState(false)
+  const { t } = useTranslation()
+
+  const cloudBackupIndex = getCloudBackupIndex(address)
+
+  // We need to ensure that the wallet is unlocked before navigating to the next screen,
```
jophish, line 52: An alternative here would be to pass the mnemonic as a prop to the next screen; during testing, this was a bit of an issue if the wallet was locked when navigating to this page, as calling `useAccountKey` prompts the user for their PIN if the wallet is locked. We can't call `useAccountKey` "on demand" when the user presses the button to go to the next screen, since React will yell at you if a different set of hooks is called on subsequent renders.

In practice, however, it's probably impossible for the user's account to be locked when arriving at this screen, so we could in theory just _assume_ that the user's account is unlocked here, call `useAccountKey` immediately, and pass the result as a prop to the next page. _If_ it so happens that the user's account is locked for whatever reason however, this would cause problems.

src/onboarding/steps.ts
```diff
@@ -40,6 +41,8 @@ export interface OnboardingProps {
   supportedBiometryType: BIOMETRY_TYPE | null
   skipVerification: boolean
   numberAlreadyVerifiedCentrally: boolean
+  showRecoveryPhrase: boolean
+  showCloudBackupFakeDoor: boolean
```
jh2oman, line 45: It doesn't appear that showCloudBackupFakeDoor is used in steps. Is it necessary here?

jophish: It's not used in `_getStepInfo`, but it _is_ used as a parameter throughout onboarding. If we remove it from here, we'd need an extra call to `getOnboardingExperimentParams` in `ProtectWallet`. On the one hand, it's an extra (potentially unnecessary call) but on the other, it means we wouldn't be "polluting" the `OnboardingProps` interface.

src/onboarding/steps.ts
```diff
@@ -14,6 +14,7 @@ import * as NavigationService from 'src/navigator/NavigationService'
 import { Screens } from 'src/navigator/Screens'
 import { RootState } from 'src/redux/reducers'
 import { store } from 'src/redux/store'
+import { getOnboardingExperimentParams } from 'src//onboarding'
```
jh2oman, line 17: is this `src//` a typo? 🤔

src/onboarding/steps.ts
jh2oman: I think OnboardingProps should just be for props that determine the steps, not props that dictate internal workings of individual steps.

src/onboarding/steps.ts
jophish: :scream: yes! ty

jophish: surprised that's not caught by the linter

src/onboarding/steps.ts
jophish: Sure, I think I agree -- happy to move it over.

src/onboarding/registration/ProtectWallet.tsx
jh2oman: Would the users account be locked if they completed entering their pin, arrived at this screen, then set down their phone for an hour?

jh2oman: Also, what about using the same approach that the settings drawer uses to gate whether or not a user can see their recovery phrase with `ensurePincode`? https://github.com/valora-inc/wallet/blob/main/src/navigator/DrawerNavigator.tsx#L119-L128

src/onboarding/steps.ts
jh2oman: me too, very weird

src/onboarding/registration/ProtectWallet.tsx
jophish: wow, i wish i had known about this sooner!! `ensurePincode` is exactly what we want here, and way cleaner. will replace with that. and to answer your question -- yeah, that's definitely a possibility.

src/onboarding/registration/OnboardingRecoveryPhrase.test.tsx
cajubelt: could you mock the useLayoutEffectHook?

jh2oman (approved):

src/navigator/Screens.tsx
```diff
@@ -49,13 +49,15 @@ export enum Screens {
...
   NameAndPicture = 'NameAndPicture',
+  OnboardingRecoveryPhrase = 'OnboardingRecoveryPhrase',
   OnboardingSuccessScreen = 'OnboardingSuccessScreen',
   OutgoingPaymentRequestListScreen = 'OutgoingPaymentRequestListScreen',
   PaymentRequestConfirmation = 'PaymentRequestConfirmation',
   PaymentRequestConfirmationLegacy = 'PaymentRequestConfirmationLegacy',
   PhoneNumberLookupQuota = 'PhoneNumberLookupQuota',
   PhotosEducation = 'PhotosEducation',
   PhotosNUX = 'PhotosNUX',
+  ProtectWallet = 'ProtectWallet',
```
cajubelt, line 60: nit: most of these are in alphabetical order-- we should probly try to keep it that way

src/onboarding/registration/OnboardingRecoveryPhrase.tsx
```diff
@@ -0,0 +1,202 @@
...
+    ValoraAnalytics.track(OnboardingEvents.protect_wallet_help)
+    setShowBottomSheet(true)
+  }
+  const onPressDismissBottomSheet = () => {
+    ValoraAnalytics.track(OnboardingEvents.protect_wallet_help_dismiss)
+    setShowBottomSheet(false)
+  }
+  const onPressCopy = () => {
+    ValoraAnalytics.track(OnboardingEvents.protect_wallet_copy_phrase)
+    Clipboard.setString(accountKey || '')
```
cajubelt, line 77:
```suggestion
    Clipboard.setString(accountKey ?? '')
```

src/onboarding/registration/OnboardingRecoveryPhrase.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+      <BottomSheet
+        testID="OnboardingRecoveryPhraseBottomSheet"
+        isVisible={showBottomSheet}
+        onBackgroundPress={onPressDismissBottomSheet}
+      >
+        <View>
+          <Text style={styles.bottomSheetTitle}>{t('recoveryPhrase.bottomSheet.title')}</Text>
+          <Text style={styles.bottomSheetBody}>{t('recoveryPhrase.bottomSheet.body1')}</Text>
+          <Text style={styles.bottomSheetBody}>{t('recoveryPhrase.bottomSheet.body2')}</Text>
```
cajubelt, line 128: these translation keys don't tell me what is being translated. I think this makes it harder for our translators too, since the base translation and the key are all they have to go off to think of how the string is used

cajubelt (commented):
looks pretty close, just left a few things

src/onboarding/registration/ProtectWallet.tsx
jophish: How would that work? We'd have to split the fakedoor variant up into two separate ones, right? That would cause any effect to be more difficult to discern in Statsig, right? Unless I'm missing something here. Agreed that it would be preferable to not have this sort of logic in-code, though.

src/onboarding/registration/OnboardingRecoveryPhrase.tsx
jophish: :facepalm:

src/onboarding/registration/OnboardingRecoveryPhrase.tsx
jophish: good callout; updated these to be a bit more descriptive.

src/onboarding/registration/ProtectWallet.tsx
cajubelt: yeah we'd end up with 2 fakedoor variants. I guess that would mean we'd get wider error bars on the metric for # of users clicking the fakedoor button. In total we'd need less manual analysis to break down by fakedoor button position, but more manual analysis to combine the fakedoor click statistic between the two fakedoor variants and compute an error bar. So maybe those two things cancel each other out. 

Reducing in-code logic could still be nice. Up to you.

src/onboarding/registration/ProtectWallet.tsx
```diff
@@ -41,28 +41,25 @@ function ProtectWallet({ navigation }: Props) {
...
-        navigate(Screens.OnboardingRecoveryPhrase)
-      }
-    } finally {
-      setShouldNavigate(false)
-    }
-  }, [shouldNavigate, address])
+  const navigateToRecoveryPhrase = () => {
+    ensurePincode()
+      .then((pinIsCorrect) => {
+        if (pinIsCorrect) {
```
cajubelt, line 55: if it's not correct do you get another chance to enter the pin?

cajubelt (approved):
i'm leaving it as optional to move the fake door position to a statsig parameter or not

src/onboarding/registration/ProtectWallet.tsx
satish-ravi: Here's the logic I wrote for bucketing an address for rolling out FC providers https://github.com/valora-inc/valora-rest-api/blob/2a64fefbc657804323466c4a571b9ccb38e4f2d1/src/fiatConnect/getProviderAvailability.ts#L34-L39

PR #3545, opened 2023-03-15
fix(logger): `[object Object]` logged instead of the serialized data

jeanregisser:
### Description

Long time issue finally fixed 😅 
Letting `console` do its job is better.

See also: #3546

### Test plan

**before**

```
 DEBUG  [08:35:39.590Z] iPhone 13 TEST/This is Logger.debug, [object Object]
 INFO   [08:35:39.591Z] iPhone 13 TEST/This is Logger.info, [object Object]
 WARN   [08:35:39.591Z] iPhone 13 TEST/This is Logger.warn, [object Object]
 ERROR  [08:35:39.595Z] iPhone 13 TEST :: This is Logger.error :: Test error in rootSaga$@http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&ru :: network connected true
```

**after**

```
 DEBUG  [08:37:33.787Z] iPhone 13 TEST This is Logger.debug {"someVal": "test"}
 INFO   [08:37:33.792Z] iPhone 13 TEST This is Logger.info {"someVal": "test"}
 WARN   [08:37:33.793Z] iPhone 13 TEST This is Logger.warn {"someVal": "test"}
 ERROR  [08:37:33.797Z] iPhone 13 TEST :: This is Logger.error :: Test error in rootSaga$@http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&ru :: network connected true
```

### Related issues

- Fixes RET-648

### Backwards compatibility

Not entirely as the data in the log file will change a bit, this shouldn't affect anything though.

PR #3546, opened 2023-03-15
chore(logs): remove unnecessary stringification

jeanregisser:
### Description

Following #3545, we no longer need to stringify objects passed to our logger.

### Test plan

- Manually checked logs have all the info needed

### Related issues

- Fixes RET-648

### Backwards compatibility

Not entirely as the data in log file will change a bit, this shouldn't affect anything though.


kathaypacific (approved):
🎈

PR #3547, opened 2023-03-15
fix(sagas): use `safely` in all our action listeners

jeanregisser:
### Description

Related to #3526 and #3527, apply `safely` everywhere we use `takeEvery`/`takeLatest`/`takeLeading`.
This will prevent unhandled errors from cancelling long lived action listeners.

### Test plan

- Tests pass

### Related issues

N/A

### Backwards compatibility

Yes


kathaypacific (approved):
🚀

PR #3548, opened 2023-03-15
chore: ensure supercharge v2 first session data refresh

kathaypacific:
### Description

Supercharge v2 is controlled by a remote config variable. When the app is launched, we fetch supercharge rewards and remote config at around the same time resulting in a race condition. It is very probable that the user will fire a request to fetch supercharge v1 rewards before the updated remote config value is received. This means that the user could navigate to the supercharge screen and try to claim their v1 rewards, which will always fail because the superchargeV2 remote config has updated by this time.

### Test plan

There are two ways the user can visit the supercharge screen:
1. open the app -> navigate to supercharge screen
2. open the app through a push notification -> land directly on the supercharge screen

In both cases, it is most probable that the request to supercharge v1 will be fired before receiving the updated remote config value. These are the ways I think the race condition could play out:
1. the updated remote config variable comes back before the supercharge v1 response: we will prevent the v1 response from being saved if supercharge v2 is enabled.
2. if the supercharge v1 response comes back first and is saved, when the updated remote config value is received then we will clear the saved rewards and fetch supercharge v2 rewards.

### Related issues

- Fixes RET-656

### Backwards compatibility

Y


src/consumerIncentives/saga.ts
```diff
@@ -220,6 +223,7 @@ export function* fetchAvailableRewardsSaga() {
     const response: Response = yield call(
       fetchWithTimeout,
       `${superchargeRewardsUrl}?address=${address}`,
+      null,
```
kathaypacific, line 226: this fixes a bug, i missed that this function takes a `body` input param

dievazqu (approved):

src/consumerIncentives/saga.ts
```diff
@@ -230,8 +234,18 @@ export function* fetchAvailableRewardsSaga() {
...
+    // supercharge rewards and fetching remote config values, prevent mismatched
+    // data from being stored in redux here
+    const mostRecentSuperchargeV2Enabled = yield select(superchargeV2EnabledSelector)
+    if (
+      data.availableRewards.length === 0 ||
+      (mostRecentSuperchargeV2Enabled && isSuperchargePendingRewardsV2(data.availableRewards)) ||
+      (!mostRecentSuperchargeV2Enabled && !isSuperchargePendingRewardsV2(data.availableRewards))
+    ) {
+      yield put(setAvailableRewards(data.availableRewards))
+      yield put(fetchAvailableRewardsSuccess())
```
jeanregisser, line 247: Oh no, a double dispatch 😢  No need to do anything though, most of the code here do that.

jeanregisser (commented):
Just curious, alternatively could we cancel and retry the existing fetch saga when the remote config value changes instead?

Probably doesn't matter since it's "temporary", but was feeling it was a bit hacky.

kathaypacific:
> Just curious, alternatively could we cancel and retry the existing fetch saga when the remote config value changes instead?
> 
> Probably doesn't matter since it's "temporary", but was feeling it was a bit hacky, since there are no tests helping confirm the fix.

i agree, this was hacky. i was thinking about some kind of cancel but got fixated on trying to cancel the network request yesterday which was too hard. thinking about it with fresh eyes, i think we can just use `takeLeading` for the fetch supercharge rewards saga which will automatically cancel inflight sagas. i think the changes are more isolated now and feeling less hacky, thanks for the suggestion!

jeanregisser (approved):

PR #3552, opened 2023-03-15
chore(statsig): log "started onboarding" timestamp to statsig

cajubelt:
### Description

Helps determine whether a user should be included in an experiment. The motivating example is our Recovery Phrase in Onboarding experiment, where we only want to show users the treatment if they see the onboarding screen first. (For instance, we do not want to expose users who have already completed onboarding before the experiment took place.)

The downside of this is we'll have to remember to update a targeting gate when we start the experiment to some future time to get this to work (even if it's just a few minutes). 

Alternatives considered:
- before querying Statsig for experiment parameters (which emits an exposure event), check some in-code logic first (like "is the user currently going thru onboarding, or has already gone thru onboarding while this experiment was on"). This would actually be pretty challenging since the Statsig SDK doesn't give us a way of checking if an experiment is running or not, and it would add what I'd describe as "experiment business logic" to our app. I'd prefer a solution that defers as much experiment biz logic as possible to Statsig.
- include existing users in the treatment group even if they won't see the new onboarding screens, and use a funnel metric  to exclude them in the decision metric analysis. I'm actually not sure how we'd analyze the control group in this case, but in any case, including users in a cohort that we don't actually want to analyze would have downstream costs (need more 1-off metrics, need to remember to use a funnel for all our analysis, ...).

### Test plan
- added unit tests
- Tested manually to see if this would let us check a feature gate that depends on it in getOnboardingStepValues. It worked. See exposure stream on [this feature gate](https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/gates/dummy_start_timestamp_late_enough)

### Related issues

cc https://linear.app/valora/issue/ACT-643

### Backwards compatibility

na


src/account/saga.ts
```diff
@@ -220,6 +221,25 @@ export function* handleUpdateAccountRegistration() {
   }
 }
 
+export function* handleChooseCreateAccount() {
+  try {
+    yield call([Statsig, 'updateUser'], {
+      custom: { startOnboardingTimestamp: new Date().toISOString() },
+    })
```
satish-ravi, line 228: is this guaranteed to fire immediately? Wondering if we'd run into a scenario where we call this, and immediately request a variant and the targeting gate fails and then later passes.
Also, just to confirm, the latest value is sufficient right in case the user goes back and forth between create / restore flow and this ends up firing multiple times

cajubelt: we talked offline and decided to move updateUser to the button click handler to make sure it gets invoked in time

cajubelt: > Also, just to confirm, the latest value is sufficient right in case the user goes back and forth between create / restore flow and this ends up firing multiple times

I thought of this too, and opted to keep implementation simpler by just updating the start timestamp each time a user chooses "create account". However, I ran a query just now to see how often a user onboards twice on the same device, and I was surprised to find that the number is actually pretty high, around 8-10% (depending on how you do the filtering; 8% if you exclude users who onboard twice in 30 seconds). This could pose a problem in our experiments, since a user onboarding a second time might behave differently than one onboarding the first time, and we'd likely want to filter them out if our goal is to get "real new users". 

So I'll probly try adding start timestamp to redux to avoid this, and avoid updating start timestamp twice for the same device. 

Unfortunately, I don't think we'll be able to solve this for users onboarding twice on two different devices-- they could get included in experiments, cross bucketed, or the like. But that's out of scope for this PR IMO.

satish-ravi (approved):

src/account/reducer.test.ts
```diff
@@ -0,0 +1,22 @@
...
+      expect(reducer(initialState, action)).toEqual(expectedState)
+    })
+    it('does not update onboarding time if already set', () => {
+      const action: ChooseCreateAccountAction = { type: Actions.CHOOSE_CREATE_ACCOUNT, now: 456 }
+      const expectedState = { ...initialState, startOnboardingTime: 123 }
+      expect(reducer(expectedState, action)).toEqual(expectedState)
+    })
+    it('sets choseToRestoreAccount to false', () => {
+      const action: ChooseCreateAccountAction = { type: Actions.CHOOSE_CREATE_ACCOUNT, now: 123 }
+      const expectedState = { ...initialState, choseToRestoreAccount: false }
```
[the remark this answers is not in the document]
cajubelt: yep, thanks

src/account/reducer.ts
```diff
@@ -95,10 +97,8 @@ export const initialState: State = {
   celoEducationCompleted: false,
 }
 
-export const reducer = (
-  state: State | undefined = initialState,
-  action: ActionTypes | RehydrateAction | Web3ActionTypes | AppActionTypes
-): State => {
+export type AccountAction = ActionTypes | RehydrateAction | Web3ActionTypes | AppActionTypes
```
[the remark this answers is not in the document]
cajubelt: thanks, i used that briefly in tests and then did something else. i'll revert

PR #3553, opened 2023-03-16
chore: update supercharge v2 screen copy

kathaypacific:
### Description

This PR:
- creates new translation keys for the "is supercharging" state of the supercharging screen. the copy is unchanged but this change enables us to update the translations over the air for older app versions only. 
- adds copy for reconnecting for DPV users. users who were supercharging but have not migrated to CPV will see this messaging, to explain why they need to verify again.

### Test plan

Simulator Screen Shot - iPhone 14 Pro - 2023-03-16 at 11 25 46

### Related issues

- Fixes RET-656

### Backwards compatibility

Y


src/consumerIncentives/ConsumerIncentivesHomeScreen.test.tsx
```diff
@@ -18,6 +19,20 @@ import { RootState } from 'src/redux/reducers'
 import { StoredTokenBalance } from 'src/tokens/slice'
 import { createMockStore } from 'test/utils'
 import { mockCeurAddress, mockCusdAddress } from 'test/values'
+
+jest.mock('react-i18next', () => ({
+  ...(jest.requireActual('../../__mocks__/react-i18next') as any),
+  Trans: ({ i18nKey, tOptions, children }: TransProps) => {
```
kathaypacific, line 25: i want to make this change in the __mocks__ file but will do that in a separate PR. the way we've mocked `Trans` is just to render all the child nodes, which is not helpful if you want to validate the key itself (which, i think is the more useful thing to test for)

src/consumerIncentives/ConsumerIncentivesHomeScreen.test.tsx
```diff
@@ -133,6 +148,30 @@ describe('ConsumerIncentivesHomeScreen', () => {
     )
 
     expectVisibleMainComponents(queryByTestId, 'SuperchargeInstructions')
+    expect(getByTestId('SuperchargeInstructions/ConnectNumber')).toHaveTextContent(
```
kathaypacific, line 151: unfortunately `getByText` looks for an exact match, whereas the mock i've defined outputs the key + the react nodes

jeanregisser (approved):
LGTM 👍

PR #3555, opened 2023-03-16
feat: Support multiple invite rewards versions in the wallet

M-shin:
### Description

Currently, we only support one invite rewards version in the wallet. Invite rewards is either "enabled" or "disabled". This PR brings support for different invite rewards versions. The `inviteRewardsEnabled` field in app state is replaced with `inviteRewardsVersion` ("none", "v4", "v5", ...). And relevant changes are made to the invite modal to support multiple versions.

**Note**: Some copy is TBD (including placeholder link to support article). I will update this PR with screenshots once design is finalized. I thought it would still be valuable to get review on the underlying plumbing to support multiple invite rewards versions.

### Test plan

Manual testing on small devices (with both large and small font sizes on the device). Ensuring that the UI matches design https://www.figma.com/file/DL6Huw867n1EZOVPJ2kNmp/Invite-Rewards?node-id=2705-34749&t=LkieO8VYs2111rGo-0. 
Two main paths:
1. Invite flow: Side menu -> "Invite" -> "Share Invite"
2. Send flow: Send -> Choose user who does not have Valora -> "Share Invite"

### Related issues

- Fixes RET-635

### Backwards compatibility

All translation changes were left backwards compatible. Unfortunately, this resulted in less clean code (eg. would have been nice to use `inviteWithUrl.rewardsActive.${version}.title`).

image

image


scripts/sync_branding.sh
```diff
@@ -21,7 +21,7 @@ echo $mobile_root
 cd "$mobile_root"
 
 # Please update the sha when valora branding updates are needed
-valora_branding_sha=d4f5c6cea5c6dcba2a18ba2c20487cce5e13441f
+valora_branding_sha=027dc2843a50e7f0b65d861dd3fd7172de28db3f
```
kathaypacific, line 24: i think this PR will be merged after https://github.com/valora-inc/wallet/pull/3554, which contains a more recent branding hash so this change shouldn't be needed

test/schemas.ts
```diff
@@ -379,7 +379,7 @@ export const v7Schema = {
   },
   send: {
     ...v6Schema.send,
-    inviteRewardsEnabled: false,
+    inviteRewardsVersion: 'disabled',
```
kathaypacific, line 382: we shouldn't need to change previous migrations?

src/invite/Invite.tsx
```diff
@@ -5,28 +5,44 @@ import { Share } from 'react-native'
...
+        title = t('inviteWithUrl.rewardsActive.title')
+        descriptionKey = 'inviteWithUrl.rewardsActive.body'
+        message = t('inviteWithRewards', { link: shareUrl })
+        image = inviteWithRewards
+        break
+      case 'v5':
+        title = t('inviteWithUrl.rewardsActiveV5.title')
+        descriptionKey = 'inviteWithUrl.rewardsActiveV5.body'
+        message = t('inviteWithRewardsV5', { link: shareUrl })
+        helpKey = 'inviteWithUrl.help'
```
kathaypacific, line 41: any reason why we want this help text to only be available for v5? it seems generic enough to the claiming process, at least as far as the copy shows right now. in which case instead of the helpKey and helpLink props, we could just have a `showCollectRewardHelp` or some boolean prop

src/invite/InviteModal.tsx
```diff
@@ -42,7 +60,14 @@ const InviteModal = ({
       <View style={styles.contentContainer}>
         <Image style={styles.imageContainer} source={imageSource} resizeMode="contain" />
         <Text style={[fontStyles.h2, styles.text]}>{title}</Text>
-        <Text style={[fontStyles.regular, styles.text]}>{description}</Text>
+        {description ? <Text style={[fontStyles.regular, styles.text]}>{description}</Text> : null}
+        {descriptionKey ? (
+          <Text style={[fontStyles.regular, styles.text]}>
+            <Trans i18nKey={descriptionKey}>
```
kathaypacific, line 66: @jeanregisser are we still keeping the door open for automatically deriving translation keys, when we're not sure how it would work with Crowdin and OTA? if so, we should avoid separating the translation key from the translation function like this (but because we need the `Trans` component, we'd need to pass the whole component in as a prop :/ )

src/invite/InviteModal.tsx
```diff
@@ -77,10 +111,28 @@ const styles = StyleSheet.create({
     width: 120,
     height: 120,
   },
+  helpContainer: {
+    width: 312,
+    height: 100,
+    left: 24,
```
kathaypacific, line 117: i don't think these styles are quite working. i'd recommend not setting the width and height and position, and letting the container size itself based on the content (otherwise you risk clipped text and weird alignment depending on the screen size and device font scaling). all that's really needed is a marginBottom i think, and for spaces we should use the `Spacing` variable.

Simulator Screen Shot - iPhone SE (3rd generation) - 2023-03-17 at 11 36 03

src/invite/InviteModal.tsx
```diff
@@ -77,10 +111,28 @@ const styles = StyleSheet.create({
...
+  },
+  helpText: {
+    color: colors.gray5,
+    fontSize: 12,
```
kathaypacific, line 121: i think you can use the `...fontStyles.xsmall` or something here instead

src/invite/InviteModal.tsx
```diff
@@ -77,10 +111,28 @@ const styles = StyleSheet.create({
...
+    color: colors.onboardingBlue,
+    flexWrap: 'wrap',
+    textDecorationLine: 'underline',
+  },
   text: {
     textAlign: 'center',
     marginBottom: Spacing.Regular16,
   },
+  textBold: {
+    fontFamily: 'Inter-SemiBold',
```
kathaypacific, line 134: would use `...fontStyles.xsmall, fontFamily: 'Inter-SemiBold'` to control the font size etc

kathaypacific (commented):
looking good!

one thought i had when reviewing this is that it's not very obvious what "v4" or "v5" means for a reward version :/ it's also confusing that "disabled" is a reward version. what do you think about using a word string instead to decide the reward type? like none / nft / cUSD? 

also, do we care about the scenario where someone is using an old version of the app in a region where v5 is enabled, and is expecting an NFT based on messaging in the app but gets cUSD instead? idk if anyone would be sad about that / if we've already handled that somehow in the backend?

src/invite/InviteModal.tsx
M-shin: oh okay sounds good. I copied from figma, but thought it didn't look right on different devices, and didn't know what to do, will add the margin and remove the absolute positioning

src/invite/Invite.tsx
M-shin: I don't think there is any reason to limit this, I'll add the link for v4. I do think we should still have `helpLink` because different reward programs can have different links.

test/schemas.ts
M-shin: ah, I just ran a grep for anywhere that used `inviteRewardsEnabled`, wasn't sure if this needed to be changed or not. I will change it back.

M-shin:
I think using reward type instead of version would be great for clarity, but this will make things more difficult if we ever run different programs of the same type (eg. 1 usd in Nigeria, 5 usd in EU). Since we don't know that will ever be a thing, I think going with reward type for now makes sense.

test/schemas.ts
M-shin: I get this error when I change this back to `inviteRewardsEnabled` (and remove `inviteRewardsVersion`):
```
 FAIL  test/schema.test.ts
  ● getLatestSchema › validates against the RootState schema

    expect(received).toBeNull()

    Received: [{"instancePath": "/send", "keyword": "required", "message": "must have required property 'inviteRewardsVersion'", "params": {"missingProperty": "inviteRewardsVersion"}, "schemaPath": "#/required"}, {"instancePath": "/send", "keyword": "additionalProperties", "message": "must NOT have additional properties", "params": {"additionalProperty": "inviteRewardsEnabled"}, "schemaPath": "#/additionalProperties"}]
```

kathaypacific: ah, on closer inspection, this is because you haven't done a redux migration. https://github.com/valora-inc/wallet/pull/3454 this is an example of the code we need to add to add a new redux state variable. we're renaming a redux variable in this PR so it is an addition + deletion. docs on why / how to do this are here https://github.com/valora-inc/wallet/blob/main/WALLET.md#redux-state-migration lmk if you have any questions

src/send/selectors.ts
```diff
@@ -16,11 +16,21 @@ export const isSendingSelector = (state: RootState) => {
...
 
-export const inviteRewardsEnabledSelector = (state: RootState) => state.send.inviteRewardsEnabled
+export const inviteRewardsTypeSelector = (state: RootState) => {
+  switch (state.send.inviteRewardsVersion) {
+    case 'v4':
+      return 'nft'
+    case 'v5':
+      return 'cUSD'
+    default:
+      return 'none'
```
kathaypacific, line 26: oh interesting selector, usually we've had the same values on remote config as in the code but i think this makes sense too since the version will make more sense to the product team. can we use an enum instead though, so that we can easily see all the available values? then when we use the type we can reference the enum instead of the actual value

kathaypacific: aside: i'd also recommend adding a description of this value in the remote config console, to explain the acceptable values (e.g. when Ellen goes to update this value, she needs to know what the available options are)

src/invite/InviteModal.tsx
```diff
@@ -1,34 +1,50 @@
...
 import colors from 'src/styles/colors'
 import fontStyles from 'src/styles/fonts'
 import { Spacing } from 'src/styles/styles'
 import variables from 'src/styles/variables'
 
 interface Props {
   title: string
-  description: string
+  description?: string
+  descriptionProps?: Record<string, any>
```
kathaypacific, line 21: i think you can use `TransProps` here instead

locales/base/translation.json
```diff
@@ -920,6 +925,7 @@
   "inviteWithoutPayment": "Hi! I’ve been using {{appName}} to send crypto worldwide. It’s easy to try and I want to invite you to check it out with this link: {{link}}",
   "inviteWithEscrowedPayment": "Hi! I’ve been using {{appName}} to send crypto worldwide. I just sent you {{amount}} {{token}}. Please download the {{appName}} App and redeem your {{token}} using this link: {{link}}",
   "inviteWithRewards": "Hey! I just sent you crypto with my {{appName}} wallet. Once you download the app and connect your phone number, we both get a special edition NFT. Use this link: {{link}}",
```
kathaypacific, line 927: i know this is out of the immediate scope but can we also update this key with the first part of the new copy? `Hi! 👋 I'm using {{appName}} as my crypto wallet. It's super quick to download the app and connect your phone number - plus, we both get ...` - i think the current copy is wrong, it still says "i've sent you crypto"

kathaypacific (approved):
Looks great! I think it's fine to merge after a few small suggestions, mainly around the tests / copy.

i also think that the help text is way too small, i had to pull out my magnifying glass to see it....there's so much space on the screen, can we make this a regular font size instead of xs?

kathaypacific:
> Looks great! I think it's fine to merge after a few small suggestions, mainly around the tests / copy.
> 
> i also think that the help text is way too small, i had to pull out my magnifying glass to see it....there's so much space on the screen, can we make this a regular font size instead of xs?
> 
> 

okay nevermind about the font size, i see we also use this tiny size in the dapps screen :/ so probably fine to leave as is and address holistically

src/invite/Invite.test.tsx
```diff
@@ -9,6 +9,10 @@ import * as InviteUtils from 'src/invite/utils'
 import { createMockStore } from 'test/utils'
 import Invite from './Invite'
 
+jest.mock('react-i18next', () => ({
```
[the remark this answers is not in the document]
M-shin: Yeah, I was thinking that this was needed to support the <Trans> mock, but I still wasn't able to get by text for some reason.. So I changed those to `getByTestId`, and I will remove this.

PR #3556, opened 2023-03-16
feat: Load .celo name instead of profile name when the user has one

H34D:
### Description

Celo Domain Names are soulbound domain names for your regenerative finance journey. Prosperity Passport Domains track your reputation and contribution across the ecosystem. They also unlock utility such as micro-loans, universal basic income, token rewards, and more.

This PR adds the Masa SDK and Soul name capabilities to valora.

### Related issues

- Fixes #[issue number here]


silasbw:
Hey @H34D -- thanks for sending this :raised_hands: It's impressive how simple the integration might be.  We're considering ways to allow more profile customization that isn't gated through this wallet repo (think something vaguely like a plugin).  I know a Masa <> Valora chat is getting scheduled, so we'd love to get your feedback on that topic then.

silasbw:
Thanks again!  Let's loop back around to this once Valora has a better plan for profile customization.

PR #3536, opened 2023-03-14
fix: ensure pincode buttons fit on all screen sizes

kathaypacific:
### Description

We need to ensure that the pincode screen content fits on the screen, or else users could be blocked from entering their pincode for transactions.

This PR is a quick solution for now - 
- make the pincode keyboard always available

### Test plan

Tested with iPhone SE (I could repo the original problem with this screen size) with slightly larger device font.

https://user-images.githubusercontent.com/20150449/225001593-4f12f240-e128-44cb-ba86-ad55d492c895.mp4

https://user-images.githubusercontent.com/20150449/225001599-d796f0c7-0cee-43c5-ba87-a7b2dae2bb9b.mp4

https://user-images.githubusercontent.com/20150449/225003939-980fa63f-f633-425d-8da6-9704cbd33370.mp4

### Related issues

- Fixes SUP-62

### Backwards compatibility

Y

PR #3558, opened 2023-03-16
test(e2e): use iPhone SE for iOS detox test device

MuckT:
### Description

Switches our iOS E2E tests to use a smaller device following up the incident that resulted in #3536. 

### Test plan

Tested in CI.

### Related issues

Fixes RET-665

### Backwards compatibility

Yes


MuckT:
@kathaypacific it looks like this might have uncovered an issue with the dapps bottom sheet.
Screen Shot 2023-03-16 at 7 45 00 PM

jeanregisser (approved):
Great idea!! 🚀

jeanregisser:
> @kathaypacific it looks like we'll need to scroll on some of the wallet connect tests. Screen Shot 2023-03-16 at 7 45 00 PM

Ah weird, this bottom sheet is supposed to measure itself and expand till it reaches the top so all content is always visible.

kathaypacific:
> > @kathaypacific it looks like we'll need to scroll on some of the wallet connect tests.
> 
> Ah weird, this bottom sheet is supposed to measure itself and expand till it reaches the top so all content is always visible.

yeah it's supposed to! i just tested on an SE and it seems okay for me. i'll rerun the CI, maybe it was a glitch / a screenshot taken during a transition
Simulator Screen Shot - iPhone SE (3rd generation) - 2023-03-17 at 09 24 54

MuckT:
@kathaypacific looks like the WC tests are passing just a small adjustment for fiat connect tests is needed.

PR #3559, opened 2023-03-17
test(e2e): update default pin to include a zero and most edge buttons

MuckT:
### Description

Switches our default e2e test pin to a pin that uses a 0 so that issues that required #3536 as a fix fail in e2e tests.

### Test plan

Tested in CI.

### Related issues

Fixes RET-665

### Backwards compatibility

Yes


e2e/src/utils/consts.js
```diff
@@ -7,7 +7,7 @@ export const SAMPLE_WALLET_ADDRESS = '0x6131a6d616a4be3737b38988847270a64bc10caa
 export const SAMPLE_BACKUP_KEY_12_WORDS =
   'embody siege middle glory soda solar nasty report swap now never any'
 export const SAMPLE_WALLET_ADDRESS_12_WORDS = '0x86b8f44386cb2d457db79c3dab8cf42f9d8a3fc0' // corresponds to the backup key above
-export const DEFAULT_PIN = '323122'
+export const DEFAULT_PIN = '130685'
```
kathaypacific, line 10: how about 130795?

M-shin (approved):

e2e/src/utils/consts.js
MuckT: Ended up using 193705 for the default and 507391 for the alternative pin as 130795 & 130685 were included in pin blocklist 🤦

MuckT: Ended up using 193705 & 507391 in d9a8f87e2018ab7124a550d1b1a617a060287843!

PR #3560, opened 2023-03-17
fix: unable to swap on Uniswap

jeanregisser:
### Description

This fixes swaps always failing on Uniswap.

Uniswap started using [Permit2](https://uniswap.org/blog/permit2-and-universal-router) for their swaps and it requires a signed EIP-712 message with some types which were unsupported by our implementation (specifically `uint160` and `uint48`).

I've submitted a PR to support the needed types in `sign-typed-data-utils` from `@celo/utils`: https://github.com/celo-org/celo-monorepo/pull/10224

This PR includes a patch, so we don't have to wait for a release of the Celo SDK.

### Test plan

- Updated e2e tests

https://user-images.githubusercontent.com/57791/225853763-5a35d4ca-3e55-4005-9051-ba48b3d05be7.mp4

This shows someone who's never swapped on Uniswap:
- First "Allow" is a one time max approval of cUSD for [Permit2](https://uniswap.org/blog/permit2-and-universal-router)
- Second "Allow" is the Permit2 signature
- Third "Allow" is the actual transaction signing

Subsequent swaps (of cUSD) won't need the one time approval.
And the Permit2 signature will only be needed if the existing one expired.
Right now Uniswap Permit2 signed messages are valid for 1 month with max approval.

The benefit is that if the user interacts with another app using Permit2, they won't need the initial transaction for the approval. They can just sign a Permit2 message which doesn't cost anything.

### Related issues

- Fixes RET-664

### Backwards compatibility

Yes


kathaypacific (approved):
wow, nice find!!! ✨

PR #3562, opened 2023-03-17
chore(test): update Trans mock to include translation key and params

kathaypacific:
### Description

The existing mocks for the `Trans` component only renders the children. In unit tests, it is therefore not possible to make an assertion like "expect this translation key to have been rendered". I think that it's important to be able to verify the translation key / params as well as the child components. This PR changes the mocks to include the translation key and param, as we do in the mocked `t` function.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y


__mocks__/react-i18next.js
```diff
@@ -2,36 +2,13 @@
...
-    }
-
-    return child
-  })
+const renderTrans = ({ i18nKey, tOptions, children }) => {
+  return (
+    <React.Fragment>
+      {[i18nKey, JSON.stringify(tOptions)].filter((value) => !!value).join(', ')}
+      {children}
+    </React.Fragment>
```
kathaypacific, line 10: this is the key change in this PR, the rest are updates to existing snapshots

__mocks__/react-i18next.js
```diff
@@ -2,36 +2,13 @@
...
-    if (typeof child === 'object' && !isElement) {
-      return Object.keys(child).reduce((str, childKey) => `${str}${child[childKey]}`, '')
-    }
-
-    return child
-  })
+const renderTrans = ({ i18nKey, tOptions, children }) => {
+  return (
+    <React.Fragment>
+      {[i18nKey, JSON.stringify(tOptions)].filter((value) => !!value).join(', ')}
```
jeanregisser, line 8: Maybe so we re-use the existing logic? and add the filter logic there?
```suggestion
      {translationFunction(i18nKey, tOptions)}
```

src/components/TokenTotalLineItem.test.tsx
```diff
@@ -74,7 +74,9 @@ describe('TokenTotalLineItem', () => {
     it('shows the right amounts', () => {
       const { getByTestId } = renderComponent({})
       expect(getElementText(getByTestId('TotalLineItem/Total'))).toEqual('R$15.00')
-      expect(getElementText(getByTestId('TotalLineItem/ExchangeRate'))).toEqual('R$1.50')
+      expect(getElementText(getByTestId('TotalLineItem/ExchangeRate'))).toEqual(
+        'tokenExchanteRate, {"symbol":"cUSD"}R$1.50'
```
jeanregisser, line 78: These look a bit weird, but that's fine.

jeanregisser (approved):
Awesome 🚀

src/components/TokenTotalLineItem.test.tsx
kathaypacific: i agree, this one is weird. but the `tokenExchanteRate` translation is `"{{symbol}} @ <0></0>"` which is the source of the weirdness i think....i didn't want to investigate why for this 🙈

M-shin (approved):

kathaypacific:
> @kathaypacific any ideas on why this increased the build size? I only see updates to the snapshots and some changes to the mocks, but emerge is reporting some new icons.

hey @MuckT great question, i was also surprised about this but then i saw that the app is Celo(test) so i didn't dig further. is the increased bundle size for the CI a problem?

PR #3563, opened 2023-03-17
feat: sprinkle haptic feedback in a handful of places

kathaypacific:
### Description

This PR adds the plumbing to do haptic feedback, and adds it to some key actions. I added a new lib because the `Vibration` that comes with react native is very limited. On iOS, the vibration is fixed at 400ms which was way to aggressive.

This PR:
- add haptic feedback lib
- add a haptic feedback helper file - so that we can give the vibrations semantic meanings and change them easily in one place
- add haptic feedback to in app notifications, primary buttons, copy action, dapp favourite, settings toggles

### Test plan

Manual

### Related issues

- Fixes RET-651

### Backwards compatibility

Y


android/settings.gradle
```diff
@@ -1,4 +1,6 @@
 rootProject.name = 'celo'
+include ':react-native-haptic-feedback'
+project(':react-native-haptic-feedback').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-haptic-feedback/android')
```
jeanregisser, line 3: Is this necessary? Doesn't support auto-linking?

android/app/src/main/java/org/celo/mobile/MainApplication.java
```diff
@@ -15,6 +15,7 @@
 import com.facebook.react.modules.network.OkHttpClientProvider;
 import com.facebook.react.shell.MainReactPackage;
 import com.facebook.soloader.SoLoader;
+import com.mkuczera.RNReactNativeHapticFeedbackPackage;
```
jeanregisser, line 18: ditto

ios/Podfile
```diff
@@ -32,6 +32,8 @@ target "celo" do
   pod 'Permission-Camera', :path => "#{permissions_path}/Camera"
   pod 'Permission-AppTrackingTransparency', :path => "#{permissions_path}/AppTrackingTransparency"
 
+  pod 'RNReactNativeHapticFeedback', :path => '../node_modules/react-native-haptic-feedback'
+
```
jeanregisser, line 36: No auto linking too? 😢

jeanregisser (approved):
OMG!!! Love this 🚀 🚀 🚀 

Can't wait to try this on real device, but the code looks great 👍

kathaypacific:
> @jeanregisser @kathaypacific should we give users an option to disable haptic feedback in settings?

ooooh interesting question. i'll check with Reed, i think having a way to opt out would be nice indeed

kathaypacific:
> > @jeanregisser @kathaypacific should we give users an option to disable haptic feedback in settings?
> 
> ooooh interesting question. i'll check with Reed, i think having a way to opt out would be nice indeed

i checked with Reed and we'll add something to the settings screen. i'll add in a separate PR

PR #3569, opened 2023-03-21
feat(onboarding): hook up choose your adventure experiment

jh2oman:
### Description

Created a new [choose_your_adventure](https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/experiments/choose_your_adventure/setup) statsig experiment with three parameters
* _onboardingNameScreenEnabled_ now directs the app to show or not show the name screen. Additional logic was implemented with the step counters to account for different entrypoints into onboarding
* _chooseAdventureEnabled_ shows / hides the new choose adventure screen after onboarding
* _cashInBottomSheetEnabled_ shows / hides the bottom sheet on the home screen. This replaces a very old remote config value that did the same thing.

https://user-images.githubusercontent.com/8432644/226719156-d4c0ca60-993b-4d0f-8285-9a05cd0945a1.mp4

### Test plan

Unit tests added.

### Related issues

https://linear.app/valora/issue/ACT-661/setup-statsig-experiment-and-use-parameters-in-wallet

### Backwards compatibility

Yes


src/onboarding/steps.ts
```diff
@@ -41,8 +41,30 @@ export interface OnboardingProps {
...
+  recoveringFromStoreWipe,
+}: {
+  onboardingNameScreenEnabled: boolean
+  recoveringFromStoreWipe: boolean
+}) {
+  if (onboardingNameScreenEnabled) {
+    return Screens.NameAndPicture
+  } else {
+    if (recoveringFromStoreWipe) {
+      return Screens.ImportWallet
```
jophish, line 63: nit: add unit tests for this?

src/home/WalletHome.tsx
```diff
@@ -120,7 +122,11 @@ function WalletHome() {
         ?.balance.isGreaterThan(CELO_TRANSACTION_MIN_AMOUNT) ?? false
     const isAccountBalanceZero = hasStable === false && hasCelo === false
 
-    return cashInButtonExpEnabled && isAccountBalanceZero
+    const { cashInBottomSheetEnabled } = getExperimentParams(
```
jophish, line 125: see below comment on accessing experiment params

src/onboarding/index.ts
```diff
@@ -3,8 +3,12 @@ import { ExperimentConfigs } from 'src/statsig/constants'
 import { StatsigExperiments } from 'src/statsig/types'
 
 export function getOnboardingExperimentParams() {
+  const { onboardingNameScreenEnabled, chooseAdventureEnabled } = getExperimentParams(
```
jophish, line 6: The way that we're accessing experiment parameters here (and elsewhere) is kind of confusing to me.

`getOnboardingExperimentParams` now returns parameters for multiple experiments. For `RECOVERY_PHRASE_IN_ONBOARDING`,  it returns _all_ parameters; anywhere we want to access a parameter for this experiment (whether or not it's used in onboarding, per se) we do so through this function. For `CHOOSE_YOUR_ADVENTURE`, we only return a subset of the available experiment params, seemingly, just the ones that are actually used in onboarding. If we want one of the other parameters, we have to call `getExperimentParams` directly. 

I'm wondering if we should consider one of these alternatives instead:
* expose _all_ parameters for both experiments in `getOnboardingExperimentParams`, and the caller just picks what it needs.
* Have two different functions, e.g., `getOnboardingRecoveryExperimentParams` and `getChooseYourAdventureExperimentParams`, and have both of them return all their respective parameters.

It mainly seems weird to me that we can get _most_ of the experiment params through this function, but _some_ have to gotten directly from a call to `getExperimentParams`. I think I like the first option above more, but curious to hear other thoughts/if I'm missing something obvious.

src/onboarding/welcome/Welcome.tsx
```diff
@@ -23,12 +26,23 @@ export default function Welcome() {
...
+      })
+    )
+  }
 
   const navigateNext = () => {
     if (!acceptedTerms) {
       navigate(Screens.RegulatoryTerms)
     } else {
-      navigate(Screens.NameAndPicture)
+      startOnboarding()
```
jophish, line 45: add a test for this path?

jophish (commented):
Couple of small comments, and a question about how we access experiment parameters

src/onboarding/index.ts
jh2oman: I see what you mean. I think what you suggested as the first option is probably cleaner.
> expose all parameters for both experiments in getOnboardingExperimentParams, and the caller just picks what it needs.

I'll update for that

src/onboarding/steps.ts
```diff
@@ -77,10 +103,15 @@ export function onboardingPropsSelector(state: RootState): OnboardingProps {
...
 export function getOnboardingStepValues(screen: Screens, onboardingProps: OnboardingProps) {
+  const firstScreen = firstOnboardingScreen({
+    onboardingNameScreenEnabled: onboardingProps.onboardingNameScreenEnabled,
+    recoveringFromStoreWipe: onboardingProps.recoveringFromStoreWipe,
+  })
+
   let stepCounter = 1 // will increment this up to the onboarding step the user is on
   let totalCounter = 1
   let reachedStep = false // tracks whether we have reached the step the user is on in onboarding, and we can stop incrementing stepCounter
-  let currentScreen: Screens = FIRST_ONBOARDING_SCREEN // pointer that we will update when simulating navigation through the onboarding screens to calculate "step" and "totalSteps"
```
anthonydaoud, line 83: I think the `FIRST_ONBOARDING_SCREEN` is unused now, not sure if you want to keep it.

src/onboarding/index.ts
satish-ravi: +1 to Charlie's comment. Why not just directly call getExperiments like you did in WalletHome and avoid this extra layer?

src/onboarding/welcome/Welcome.test.tsx
```diff
@@ -39,4 +43,22 @@ describe('Welcome', () => {
...
+  it('goes to the onboarding screen', () => {
+    const store = createMockStore({
+      account: {
+        acceptedTerms: true,
+      },
+    })
+    mocked(firstOnboardingScreen).mockReturnValue(Screens.NameAndPicture)
+    const { getByTestId } = render(
+      <Provider store={store}>
+        <Welcome /*{...getMockStackScreenProps(Screens.)}*/ />
```
satish-ravi, line 55: remove comment?

src/onboarding/index.ts
jophish: Didn't consider the exposure event issue -- this solution makes sense to me!

jophish (approved):
LGTM! one optional code golf comment

PR #3570, opened 2023-03-21
feat(analytics): Add analytics events for "choose your own adventure" flow

jophish:
### Description

For ACT-663.

Implements all the analytics events, instrumented with the proper data, that we'll need in order to satisfy ACT-663. Metrics still need to be set up in Statsig, however.

### Test plan

Unit tested.

### Related issues

- Fixes ACT-663.

### Backwards compatibility

Yes.


src/analytics/Properties.tsx
```diff
@@ -60,6 +60,8 @@ import { RecipientType } from 'src/recipients/recipient'
 import { Field } from 'src/swap/types'
 import { Currency, CurrencyOrCREAL, StableCurrency } from 'src/utils/currencies'
 import { Awaited } from 'src/utils/typescript'
+import { AdventureCardMap } from 'src/onboarding/ChooseYourAdventure'
```
satish-ravi, line 63: I guess we have other screen imports here that uses Analytics, but this introduces a cyclic dependency right? 
```
ChooseYourAdventure -> ValoraAnalytics -> Properties -> ChooseYourAdventure
```
Should we avoid introducing another one?

src/analytics/Properties.tsx
```diff
@@ -336,6 +338,12 @@ interface OnboardingEventsProperties {
   [OnboardingEvents.protect_wallet_help_dismiss]: undefined
   [OnboardingEvents.protect_wallet_copy_phrase]: undefined
   [OnboardingEvents.protect_wallet_complete]: undefined
+  [OnboardingEvents.cya_button_press]: {
+    name: string
+    index: number
+    indexMap: AdventureCardMap
```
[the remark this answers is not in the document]
jophish: Hmm, no real benefit. I'll switch it to an array and make the index 1-based. I can rename it to something like `position` also.

src/analytics/Properties.tsx
jophish: Yeah, I can refactor things in order to avoid this cycle.

jh2oman (approved):
Looks good! The data seem sufficient to calculate all the metrics

PR #3571, opened 2023-03-22
chore: allow users to opt out of haptic feedback

kathaypacific:
### Description

This PR:
- adds a haptic feedback variable to redux
- allow users to switch off haptic feedback in settings
- add an analytics event to track how many users are doing this

### Test plan

Tested on a physical device. 

Simulator Screen Shot - iPhone SE (3rd generation) - 2023-03-22 at 11 14 53

### Related issues

- Fixes RET-651

### Backwards compatibility

Y


src/styles/hapticFeedback.ts
```diff
@@ -1,18 +1,27 @@
...
+import { hapticFeedbackEnabledSelector } from 'src/app/selectors'
+import { store } from 'src/redux/store'
 
 const options = {
   enableVibrateFallback: false,
   ignoreAndroidSystemSettings: false,
 }
 
+const triggerHapticFeedback = (type: ReactNativeHapticFeedback.HapticFeedbackTypes) => {
+  const hapticFeedbackEnabled = hapticFeedbackEnabledSelector(store.getState())
```
kathaypacific, line 11: idk if it's good to read from the store so much, but i also don't know enough to gauge how bad it actually is :/ i only used redux here to persist the setting across app sessions (and also to keep consistency with how we're handling the other app settings items)

i could also have a global variable in this file that changes only when the redux value changes (with a watcher saga, or just having the `hapticFeedbackSet` action also change this global variable)? wdyt @jeanregisser ?

jeanregisser: Good question, I think it's fine for specific use cases like this.
It creates a tight coupling between redux and this module, but we could easily decouple like you said.

jeanregisser (approved):
Great! 👍

PR #3573, opened 2023-03-22
chore: add super property for cpv status

kathaypacific:
### Description

This information is missing from the logs, which makes debugging more difficult for some scenarios.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y

### DIFF
--- a/src/analytics/selectors.ts
+++ b/src/analytics/selectors.ts
@@ -4,7 +4,7 @@ import DeviceInfo from 'react-native-device-info'
 import * as RNLocalize from 'react-native-localize'
 import { createSelector } from 'reselect'
 import { defaultCountryCodeSelector, pincodeTypeSelector } from 'src/account/selectors'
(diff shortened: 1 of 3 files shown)


src/analytics/selectors.ts
```diff
@@ -24,7 +24,7 @@ export const getCurrentUserTraits = createSelector(
     tokensWithTokenBalanceSelector,
     coreTokensSelector,
     getLocalCurrencyCode,
-    numberVerifiedSelector,
+    phoneVerificationStatusSelector,
```
kathaypacific, line 27: open to suggestions for how else to do this, it seems like `createSelector` expects to have a maximum of 12 selectors as input. adding a 13th one makes typescript complain :/ i found an [old issue](https://github.com/reduxjs/reselect/issues/378) that references the problem, and the suggestion was to use a `createSelector` as an input...

PR #3575, opened 2023-03-22
chore(l10n): update translations from Crowdin

valora-bot-crowdin:


locales/pl-PL/translation.json
MuckT, line 1: These translations seem a bit suspect as there is the missing {{mnemonicLength}} and profile translated to English.

kathaypacific: nice catch @MuckT !! i'll flag it to Paula

PR #3576, opened 2023-03-23
feat: dapp explorer screen search

MuckT:
### Description

Adds a search only variant for dapp search and filters.

Adds primarily four new files:
- `src/dappsExplorerDAppsExplorerScreenSearch.tsx`
- `src/dappsExplorerFavoriteDappsSectionSearch.tsx` 
- `src/dappsExplorerNoResultsSearch.tsx`
- `src/dappsExplorer/utils.ts`
 
There's some lifting up of state to `DAppsExplorerScreenSearch.tsx` to handle the case where both section lists are empty so we don't serve double no results sections. The search uses a relevance score based on a combination of name, description, category to hopefully show users the most desired results.

### Video

https://user-images.githubusercontent.com/26950305/227128295-0ca65ea6-0d0f-4e9a-a0d2-f82fcf4c45e5.mov

### Test plan

WIP

### Related issues

- Fixes RET-580

### Backwards compatibility

Yes


src/dappsExplorer/utils.ts
```diff
@@ -0,0 +1,17 @@
+import { DappV1, DappV2, isDappV2 } from 'src/dapps/types'
+
+export const calculateSearchScore = (dapp: DappV2 | DappV1, searchQuery: string) => {
```
kathaypacific, line 3: if i understand this correctly, there are some rules here to order the search results depending on how much they "match" the search results. is that right? i don't remember talking about anything like this at refinement, is this kind of logic necessary for this experiment? can we just do a black and white search, where we display all matched dapps in the existing order form the backend?

tbh i feel like this kind of ranking logic shouldn't be on the client :/ i'm not sure exactly where this should go, but it's feasible to imagine a world where we decide next month that actually we want to change this logic completely (e.g. we no longer care at all about matching the category) and would want this change to affect every app version. if this logic is baked into apps, then we don't have control anymore and we're bound to the release process for changes. 

unless this logic is very important to the experiment, i'd suggest leaving it out. if it is important, we should maybe have a quick brainstorm and weigh up pros / cons of various ways to implement this.

src/dappsExplorer/utils.ts
```diff
@@ -0,0 +1,17 @@
+import { DappV1, DappV2, isDappV2 } from 'src/dapps/types'
```
kathaypacific, line 1: i think i'm also in favour of not creating new "utils" files when possible, after hearing silas go on about this many times 😅 they tend to evolve into collections of miscellaneous and unfocused functions

src/dappsExplorer/FavoriteDappsSectionSearch.tsx
```diff
@@ -0,0 +1,96 @@
...
+import fontStyles from 'src/styles/fonts'
+import { Spacing } from 'src/styles/styles'
+
+interface Props {
+  onPressDapp: (dapp: ActiveDapp) => void
+  searchQuery: string
+  setFavoriteResultsEmpty: (empty: boolean) => void
+}
+
+export function FavoriteDappsSectionSearch({
```
kathaypacific, line 21: i'm starting to wonder about the value of this favourite dapps abstraction :/ i think it makes sense with the existing dapps screen because the "all" dapps section is an actual section list with categories as sections, whereas the favourites is a list of cards.

but with the filters and search, it's no longer the same section list. the favourites and all sections are now pretty much the same component, with different labels and placeholders and cards....i wonder if we should combine them. this is not a very well thought out suggestion yet, i just thought of it because i notice that we're doing the search logic both here and in the main dapps screen :/ and now on closer inspection, much of the rendering code is similar too.

MuckT:
@kathaypacific I considered pulling 48ff9a8f9dd3fece0d70e1cb5fb9cd9575b71b85 out in it's own PR but this branch is the best one to my knowledge to test the fix on.

src/dappsExplorer/utils.ts
MuckT: This will ultimately be shared by the search and filterSearch variants so I made a shared utils file. I am not sure of a better place for it if we want to share it between the two screens.

src/dappsExplorer/DAppsExplorerScreenSearch.tsx
```diff
@@ -0,0 +1,326 @@
...
+  const dappList = useSelector(dappsListSelector)
+  const language = useSelector(currentLanguageSelector)
+  const favoriteDappsById = useSelector(favoriteDappIdsSelector)
+
+  // Some state lifted up from all and favorite sections
+  const [searchQuery, setSearchQuery] = React.useState('')
+  // @ts-expect-error 'previousSearchQuery' is declared but its value is never read
+  const [previousSearchQuery, setPrevSearchQuery] = React.useState('')
+  const [value] = useDebounce(searchQuery, 1000)
+  const [previousValue] = useDebounce(value, 1000)
```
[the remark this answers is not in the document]
MuckT: I think I need at least 3 of the 4 too keep the analytics correct and the fourth to avoid build errors 🤔 

#### Example: User enters 'Eth'
----- E -----
`value=''`
`previousValue=''`
`searchQuery='E'`
`previousSearchQuery=''`

----- t -----
`searchQuery='Et'`
`previousSearchQuery='E'`

----- h -----
`searchQuery='Eth'`
`previousSearchQuery='Et'`

----- 1000 MS -----
`value='Eth'`
Analytics for search submitted:
`ValoraAnalytics.track(DappExplorerEvents.dapp_search, { query: 'Eth', clearing: false })`

----- 1000 MS -----
`previousValue='Eth'`

----- User Clears Input -----
`previousSearchQuery='Eth'`
`searchQuery=''`

----- 1000 MS -----
`value=''`
Analytics for search submitted:
`ValoraAnalytics.track(DappExplorerEvents.dapp_search, { query: 'Eth', clearing: true })`

PR #3580, opened 2023-03-23
feat: dapp filter improvements

MuckT:
### Description

Based on the feedback from Bugapalooza this PR does two things:
1. Tapping a selected filter will deselect that filter and make all the selected filter.
2. The header for the second section list under  'MY FAVORITES'  updates to the selected dapp filter name.

#### Video

https://user-images.githubusercontent.com/26950305/227607810-3358f623-b3cf-4f79-8ef8-329577d41fce.mov

### Test plan

- Tested locally on iOS
- Test case added
- Tested in CI

### Related issues

N/A

### Backwards compatibility

Yes


src/dappsExplorer/DappFilterChip.tsx
```diff
@@ -17,8 +17,8 @@ interface DappFilterChip {
 
 function DappFilterChip({ filterId, filterName, isSelected, onPress, style }: DappFilterChip) {
   const filterPress = () => {
+    isSelected ? onPress('all') : onPress(filterId)
```
kathaypacific, line 20: this is unnecessary right? since the `filterPress` callback that's passed in already does this check

src/dappsExplorer/DAppsExplorerScreenFilter.tsx
```diff
@@ -88,6 +88,10 @@ export function DAppsExplorerScreenFilter() {
     horizontalScrollView.current?.scrollTo({ x: 0, animated: true })
   }
 
+  const filterPress = (filterId: string) => {
+    selectedFilter === filterId ? setSelectedFilter('all') : setSelectedFilter(filterId)
```
kathaypacific, line 92: do we need to ensure that `all` cannot be toggled? e.g. `selectedFilter === filterId || selectedFilter === 'all' ? setSelectedFilter('all') : setSelectedFilter(filterId)`

kathaypacific (commented):
👯‍♂️

src/dappsExplorer/DAppsExplorerScreenFilter.tsx
MuckT: I don't think we need to ensure 'alll cannot be toggled. When we tap all and it's already selected we shouldn't do anything special since the `selectedFilter === 'all' === filterId` and we call `setSelectedFilter('all')` and the button remains selected. 

With the example code I was unable to toggle other filters and playing around with it left me with the same results as what's already implemented.

kathaypacific: oh right, of course! makes sense

kathaypacific (approved):
✨

PR #3582, opened 2023-03-24
fix: ensure always refetch supercharge rewards when remote config refreshed

kathaypacific:
### Description

This was caught in bugapalooza https://valora-app.slack.com/archives/C04B61SJ6DS/p1679593002858459?thread_ts=1679591552.442839&cid=C04B61SJ6DS

I had tunnel vision when i implemented this saga before - I was initially trying to fix existing incompatible data in redux, but the use case that was not covered is when:
1. the user receives response from v1 that there is no supercharge rewards
2. then the remote config for v2 enabled returns true
3. we need to dispatch a v2 fetch request here, but it was missed :(

User impact:
- users in this situation will not see the "claim your rewards" homecard
- users can still claim their rewards because they need to go to the supercharge screen which kicks off a new request, but they may not know to go there if they have push notifications turned off.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

n/a


jeanregisser (approved):
In general it would be nice to cover bug fixes with a test 🙈 

I mentioned there was no test [initially](https://github.com/valora-inc/wallet/pull/3548#pullrequestreview-1341930821), but we may want to add one now.
It helps think about how the code is used.

kathaypacific:
> In general it would be nice to cover bug fixes with a test 🙈
> 
> I mentioned there was no test [initially](https://github.com/valora-inc/wallet/pull/3548#pullrequestreview-1341930821), but we may want to add one now. It helps think about how the code is used.

thanks for the callout! i added a unit test

PR #3583, opened 2023-03-24
chore: remove supercharge from redux persist blacklist

kathaypacific:
### Description

To ensure that remote config values in the store are persisted (and we don't always hit the v1 supercharge endpoint for no reason once v2 is enabled.)

### Test plan

Manually tested

### Related issues

n/a

### Backwards compatibility

Y

### DIFF
--- a/src/consumerIncentives/slice.ts
+++ b/src/consumerIncentives/slice.ts
@@ -78,24 +78,25 @@ const slice = createSlice({
     }),
   },
   extraReducers: (builder) => {
-    builder.addCase(REHYDRATE, (state, action: RehydrateAction) => ({
-      ...state,
-      ...getRehydratePayload(action, 'supercharge'),
-      loading: false,
-      error: false,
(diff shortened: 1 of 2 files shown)


src/consumerIncentives/slice.ts
```diff
@@ -76,15 +78,25 @@ const slice = createSlice({
...
-    )
+    builder
+      .addCase(REHYDRATE, (state, action: RehydrateAction) => ({
+        ...state,
+        ...getRehydratePayload(action, 'supercharge'),
+        loading: false,
+        error: false,
+        fetchAvailableRewardsLoading: false,
+        fetchAvailableRewardsError: false,
+        availableRewards: [],
```
jeanregisser, line 89: Site note: why don't we keep the fetched available rewards? At least for **some** time? Seems wasteful to query on every app start, right?

jeanregisser (approved):
🙌

PR #3584, opened 2023-03-24
fix: invite modal height sometimes incorrect on android

kathaypacific:
### Description

As the title

### Test plan

Before:

After:

### Related issues

- Fixes RET-602

### Backwards compatibility

Y


src/invite/InviteModal.tsx
```diff
@@ -46,7 +48,7 @@ const InviteModal = ({
   }
 
   return (
-    <SafeAreaView style={styles.container}>
+    <SafeAreaView style={[styles.container, { height, width }]}>
```
jeanregisser, line 51: Curious, do we even need the explicit width and height? Wouldn't it try to fill the available space?

jeanregisser (approved):
🧑‍🎨

src/invite/InviteModal.tsx
kathaypacific: it is position absolute and doesn't seem to play nice without the width and height....

PR #3585, opened 2023-03-24
chore: add workflow to check that only base translations are modified

kathaypacific:
### Description

Add a GH workflow that fails a PR if translation files were edited, to prevent Crowdin translations from being overwritten.

### Test plan

Test when there are translation files modified: https://github.com/valora-inc/wallet/actions/runs/4544258918/jobs/8010040802

Test when only base translations are modified: https://github.com/valora-inc/wallet/actions/runs/4544276572/jobs/8010080504?pr=3585

### Related issues

- Fixes RET-604

### Backwards compatibility

Y


.github/scripts/checkTranslationFilesOnPr.ts
```diff
@@ -0,0 +1,24 @@
...
+
+if (modifiedTranslationFiles.length > 0) {
+  echo(
+    `❌ Only base translation files should be modified! The following files should not be modified:`
+  )
+  echo(modifiedTranslationFiles.join('\n'))
+  process.exit(1)
+}
+
+echo('✅ All good to go 🚀')
```
jeanregisser, line 24: I think we could maybe even remove this script by relying on the `files` feature from `tj-actions/changed-files`

```
    - name: Get changed files
      id: changed-files
      uses: tj-actions/changed-files@v35
      with:
        files: |
          locales/*/translation.json
          !locales/base/translation.json
          ios/celo/*.lproj/InfoPlist.strings 
          !ios/celo/Base.lproj/InfoPlist.strings
```

Makes it a bit easier to read the paths too.

And checking the `any_changed` output from the script.

.github/scripts/checkTranslationFilesOnPr.ts
```diff
@@ -0,0 +1,24 @@
...
+
+echo('Checking modified files')
+const updatedFilenames = process.argv[2]
+const modifiedTranslationFiles = updatedFilenames.split(',').filter((filename) => {
+  return filename.match(TRANSLATION_FILES)
+})
+
+if (modifiedTranslationFiles.length > 0) {
+  echo(
+    `❌ Only base translation files should be modified! The following files should not be modified:`
```
jeanregisser, line 18: I think we still need to allow this for Crowdin PRs, right?

jeanregisser (commented):
Great idea! 👍

jeanregisser (commented):
Great, but what about:
> I think we still need to allow this for Crowdin PRs, right?

kathaypacific:
> Great, but what about:
> 
> > I think we still need to allow this for Crowdin PRs, right?

i added a branch-ignore for the crowdin branch :) https://github.com/valora-inc/wallet/pull/3585/files#diff-50523cfd1d1be276394f879b02868b81fd0baa57b1abef223e931198e98ce301R7

kathaypacific:
bump on this @jeanregisser

jeanregisser (approved):

jeanregisser:
> i added a branch-ignore for the crowdin branch :) https://github.com/valora-inc/wallet/pull/3585/files#diff-50523cfd1d1be276394f879b02868b81fd0baa57b1abef223e931198e98ce301R7

Oh didn't see that. But then we can't make this check required to merge branches, right?
probably fine for now.

jeanregisser:
Or we could put the branch check within one step?

kathaypacific:
> Or we could put the branch check within one step?

ooooohhhhh yes good shout. i assumed (wrongly) that we could still include workflows that didn't apply to the branch as a requirement for merging (like, the absence of it !== a failure) but indeed this is probably not the case.

i can put the branch check in a step. i'll see if adding an "if" condition for the whole job works too

jeanregisser:
interesting, there's an official solution for this problem: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks

It's a bit weird though, no?

kathaypacific:
> interesting, there's an official solution for this problem: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks
> 
> It's a bit weird though, no?

ah, that is too weird...and kind of brittle. 

what do you think of checking the branch in the last step? ideally i'd have liked the first step to check the branch and if it's a crowdin one, then exit the job. but it seems like there's no way to exit a job early 😅 so if i have a step like that, i'd have to add an "if" to every subsequent step to skip the step depending on the branch.

checking on the last step of the branch feels a little less intrusive, it's a waste to diff the files but it only takes 18 seconds so seems okay? wdyt?

jeanregisser:
Looks all good 🚀

PR #3597, opened 2023-03-29
fix(statsig): handle Statsig in e2e test environments

anthonydaoud:
### Description

Currently, there is no in-wallet mechanism to control the value returned from Statsig in the e2e tests. When running e2e tests in a local wallet, the experiment parameter is based on the anonymousId/userId, which is randomly generated for some of the tests. This could lead to unexpected e2e test behaviors. 

This PR resolves this by setting Statsig to [local mode](https://docs.statsig.com/client/jsClientSDK#how-can-i-mock-the-sdkoverride-gates-in-tests) when running e2e tests. This means that the default values for the experiment parameters will be used in the tests.

### Test plan

CI, updated unit tests

### Related issues

- ACT-712

### Backwards compatibility

Yes


cajubelt:
have you been able to verify that this results in defaults being used for e2e tests?

anthonydaoud:
Yes I verified that by running the e2e tests locally.

PR #3598, opened 2023-03-29
feat: calculate search score for dapp list search

MuckT:
### Description

Broken out from #3576. This contains a calculateSearchScore function in `src/dappsExplorer/utils.tsx`

### Test plan

- Tested locally on iOS and Android within #3576.
- Tests added in `src/dappsExplorer/utils.test.tsx`

### Related issues

RET-580

### Backwards compatibility

Yes


src/dappsExplorer/utils.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const nameScore = dapp.name.toLowerCase().includes(searchTerm.toLowerCase()) ? 1 : 0
+  const descriptionScore = dapp.description.toLowerCase().includes(searchTerm.toLowerCase()) ? 1 : 0
+  const categoryScore = isDappV2(dapp)
+    ? dapp.categories.some((category) => category.toLowerCase().includes(searchTerm.toLowerCase()))
+      ? 1
+      : 0
+    : dapp.categoryId.toLowerCase().includes(searchTerm.toLowerCase())
+    ? 1
+    : 0
+  return nameMatchingCaseScore + nameScore + descriptionScore + categoryScore
```
kathaypacific, line 25: do we want to include the category in the scoring? i wonder if this would be confusing - for example, the user has no filters selected and they search for "swap". in this case, bitssa would be displayed but it's not clear why since there's no "swap" in the name or description :/ i wonder if this would look like a broken search, since in this scenario we do not display the dapp category anywhere.

i figure that you probably included this so that people can search for categories and get a list of matching dapps, but this is what the filters are for right? if we want the search to do something similar, it kind of feels like we should surface the category of the dapp on the dapp card itself to make it less confusing why the search result includes the dapp? 

idk, just my 2 cents. maybe Reed has an opinion on this too

MuckT: Yeah I agree searching on the category id is is not optimal. I'd expect as a user if I type swap to see bittsa since they provide swaps. What would be better is category name because then I would be localized.

MuckT: @kathaypacific added the localized category name to search in e2fe4326379d19296c03a40a63c76f33bd2df86a.

We could remove considering the category name entirely, but this might lead to some strange interactions between search and filter & search.

#### Without Fix or without Categories considered Search Example:
- User Searches 'Usa cripto'
- Dapps Listed: None

#### With Fix or without Categories considered Search Example:
- User Searches 'Usa cripto'
- Dapps Listed: Bidali, Flywallet, ChiSpend

#### Without Fix of without Categories Filter & Search Example:
- User Searches 'Usa cripto'
- User selects 'Usa cripto' Filter
- Dapps Listed: None
-

src/dappsExplorer/utils.tsx
```diff
@@ -0,0 +1,31 @@
...
+ * - +2 if the name matches the search term in a case sensitive way
+ * - +1 if the name matches the search term in a case insensitive way
+ * - +1 if the description matches the search term in a case insensitive way
+ * - +1 if the category matches the search term in a case insensitive way
+ * @param dapp expects a dapp with a categoryName or categoryNames
+ * @param searchTerm
+ * @returns {number} the score of the dapp 0 - 5
+ */
+export const calculateSearchScore = (
+  dapp: DappV1WithCategoryName | DappV2WithCategoryNames,
```
jeanregisser, line 14: Do we need the search for v1 dapps?

jeanregisser (commented):
Great! 👍

src/dappsExplorer/utils.tsx
MuckT: We do not this also simplifies some of the typing needed in #3599.

src/dapps/selectors.ts
```diff
@@ -117,6 +117,46 @@ export const favoriteDappsSelector = createSelector(
   }
 )
 
+export const favoriteDappsWithCategoryNamesSelector = createSelector(
```
dievazqu, line 120: nit: These 2 functions (`favoriteDappsWithCategoryNamesSelector` and  `dappListWithCategoryNamesSelector`) are pretty similar, can you put the code in an auxiliar function to avoid the duplicated code?

src/dappsExplorer/utils.tsx
dievazqu: I agree with Kathy, I think the categories should be filtered only using the filter chips. Or, we have to add some indication in the UI so it's clear when there is a match because of the category.
But I don't know if you have already talked about this

src/dappsExplorer/utils.tsx
```diff
@@ -0,0 +1,24 @@
...
+ * Calculates a score based on the search term and dapp
+ * - +2 if the name matches the search term in a case sensitive way
+ * - +1 if the name matches the search term in a case insensitive way
+ * - +1 if the description matches the search term in a case insensitive way
+ * - +1 if the category matches the search term in a case insensitive way
+ * @param dapp: dapp with categoryNames
+ * @param searchTerm: the user input search string
+ * @returns {number} the score of the dapp 0 - 5
+ */
+export const calculateSearchScore = (dapp: DappV2WithCategoryNames, searchTerm: string) => {
```
dievazqu, line 13: I think we are not considering the case where [the user inputs more than one word](https://www.figma.com/file/zcOyuTp5316yiguvL15as1?node-id=130:6385#351150969). I guess we have to increase the score for each matched word. Also consider if the exact searched term matched, in that case we should prioritize that match.

dievazqu (commented):
Looks good, but I think we need to address to multi word search.

jeanregisser (approved):
Looks like there's some remaining suggestions from Kathy and Diego as to how search should work.
But it looks good to me, search can easily turn into a rabbit hole 😄

MuckT:
Closing in favor of #3629.

PR #3599, opened 2023-03-30
feat: dapp list search

MuckT:
### Description

Broken out from https://github.com/valora-inc/wallet/pull/3576.  

- Adds new translation keys in `locales/base/translation.json`
- Adds `src/dappsExplorer/search/DAppsExplorerScreenSearch.tsx` and `src/dappsExplorer/search/FavoriteDappsSection.tsx` which both use `searchDappList` from `src/dappsExplorer/searchDappList.tsx`.
- Adds `src/dappsExplorer/search/NoResults.tsx` which uses a trans element to contain the no results text and the searchTerm input by the user.
- Updates `src/navigator/DrawerNavigator.tsx` to route to `DAppsExplorerScreenSearch.tsx` if search is enabled in firebase.

### Video iOS Keyboard Enabled
https://user-images.githubusercontent.com/26950305/230747341-49a15669-7899-412a-b6dd-70287e5e4212.mov

### Test plan

- New tests added in `src/dappsExplorer/search/DAppsExplorerScreenSearch.test.tsx`

### Related issues

- Fixes #RET-580

### Backwards compatibility

Yes


src/dappsExplorer/search/NoResults.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { Spacing } from 'src/styles/styles'
+
+interface Props {
+  searchTerm: string
+  testID?: string
+}
+
+function NoResults({ testID, searchTerm }: Props) {
+  return (
+    <View testID={`${testID}/NoResults`} style={styles.viewContainer}>
```
jeanregisser, line 16: I'd expect this if passed a specific testID:
```suggestion
    <View testID={testID} style={styles.viewContainer}>
```

src/dappsExplorer/search/FavoriteDappsSection.tsx
```diff
@@ -0,0 +1,93 @@
...
+    (a, b) => calculateSearchScore(b, searchTerm) - calculateSearchScore(a, searchTerm)
+  )
+
+  useEffect(() => {
+    if (favoritedDappsSearched.length > 0 && searchTerm !== '') {
+      setFavoriteResultsEmpty(false)
+    } else {
+      setFavoriteResultsEmpty(true)
+    }
+  }, [favoritedDappsSearched])
```
jeanregisser, line 41:
```suggestion
  }, [favoritedDappsSearched, searchTerm])
```

src/dappsExplorer/search/FavoriteDappsSection.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  if (favoritedDappsSearched.length === 0 && searchTerm !== '') {
+    return <NoResults testID="FavoriteDappsSection" searchTerm={searchTerm} />
```
jeanregisser, line 44:
```suggestion
    return <NoResults testID="FavoriteDappsSection/NoResults" searchTerm={searchTerm} />
```

src/dappsExplorer/search/FavoriteDappsSection.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { calculateSearchScore } from 'src/dappsExplorer/utils'
+import StarIllustration from 'src/icons/StarIllustration'
+import Colors from 'src/styles/colors'
+import fontStyles from 'src/styles/fonts'
+import { Spacing } from 'src/styles/styles'
+
+interface Props {
+  onPressDapp: (dapp: ActiveDapp) => void
+  searchTerm: string
+  setFavoriteResultsEmpty: (empty: boolean) => void
```
jeanregisser, line 18:
```suggestion
  onShowSearchResult: (empty: boolean) => void
```

src/dappsExplorer/search/DAppsExplorerScreenSearch.tsx
```diff
@@ -0,0 +1,327 @@
...
+            refreshControl={
+              <RefreshControl
+                tintColor={colors.greenBrand}
+                colors={[colors.greenBrand]}
+                style={styles.refreshControl}
+                refreshing={loading}
+                onRefresh={() => dispatch(fetchDappsList())}
+              />
+            }
+            // @ts-ignore TODO: resolve type error
```
jeanregisser, line 137: Too bad we didn't address this, now it's being copied in multiple places.

```suggestion
            // TODO: resolve type error
            // @ts-expect-error 
```

src/dappsExplorer/search/DAppsExplorerScreenSearch.tsx
```diff
@@ -0,0 +1,327 @@
...
+    if (allResultsParsed.length === 0) {
+      setAllResultEmpty(true)
+    } else {
+      setAllResultEmpty(false)
+    }
+    return allResultsParsed
+  }, [dappList, searchTerm, favoriteDappsById])
+
+  const emptyListComponent = React.useMemo(() => {
+    if (allResultEmpty && favoriteResultsEmpty) return null
```
jeanregisser, line 99: Shouldn't we display something?

src/dappsExplorer/search/DAppsExplorerScreenSearch.tsx
```diff
@@ -0,0 +1,327 @@
...
+                  }}
+                  value={searchTerm}
+                  multiline={false}
+                  placeholderTextColor={colors.gray4}
+                  underlineColorAndroid="transparent"
+                  placeholder={t('dappsScreen.searchPlaceHolder')}
+                  showClearButton={true}
+                  allowFontScaling={false}
+                />
+                {dappFavoritesEnabled && (
```
jeanregisser, line 164: Is it possible to be in that screen and `dappFavoritesEnabled` to not be `true`?

src/dappsExplorer/search/DAppsExplorerScreenSearch.tsx
```diff
@@ -0,0 +1,327 @@
...
+}
+
+function parseResultsIntoAll(
+  dappList: (DappV1WithCategoryName | DappV2WithCategoryNames)[],
+  searchTerm: string,
+  favoriteDappsById: string[]
+) {
+  // If there is no search query, return the non favorite dapps in the all section
+  const nonFavoriteDapps = dappList.filter(
+    (dapp: DappV1 | DappV2) => !favoriteDappsById.includes(dapp.id)
```
jeanregisser, line 244: Can we do it without the explicit type?
```suggestion
    (dapp) => !favoriteDappsById.includes(dapp.id)
```

src/dappsExplorer/search/DAppsExplorerScreenSearch.tsx
```diff
@@ -0,0 +1,327 @@
...
+      {
+        data: nonFavoriteDapps,
+        category: 'all',
+      },
+    ]
+  }
+
+  // Filter out dapps that don't match the search query
+  const nonFavoriteDappsMatchingSearch = nonFavoriteDapps.filter(
+    (dapp: DappV1WithCategoryName | DappV2WithCategoryNames) =>
```
jeanregisser, line 257: Same here?
```suggestion
    (dapp) =>
```

jeanregisser (commented):
Awesome work! 👍 

Does the video from https://github.com/valora-inc/wallet/pull/3576 still reflect the changes applied here?
If so we should add it to this PR description.

src/dappsExplorer/search/DAppsExplorerScreenSearch.tsx
MuckT: 🤔 I don't think so, but this does bring up a good point. I doubt we will be removing dapp favorites so it might be worth a clean-up pr to remove all the logic around if favorites is enabled and display it by default.

src/dappsExplorer/search/DAppsExplorerScreenSearch.tsx
MuckT: This is the weird case where favorites section is empty and the all section is empty. The favorites will display a no results message and we want to display nothing here.

src/dappsExplorer/search/DAppsExplorerScreenSearch.tsx
MuckT: Fixed in 6b9882a368accea810f9c533e9082c7fba72f920!

src/dappsExplorer/search/DAppsExplorerScreenSearch.tsx
dievazqu, line 1: This file seems have a lot of duplicated code in `DAppsExplorerScreenFilter`...
I wonder if you tried to merge both files. I think that adding the search logic to the previous component should be fine, but maybe I'm missing something.
I see that you did some refactor, so maybe that's a problem... But I just don't want us to create a new component unless it's really necessary, so we avoid maintaining legacy code.

dievazqu (commented):
Looks good, I just wonder if we could avoid duplicating all the code from the filter component.

jeanregisser (approved):
🚀

src/dappsExplorer/search/DAppsExplorerScreenSearch.tsx
MuckT: Added this in #3629 in the `searchDappList` function. This removes a lot of the duplicate logic across the screens. Having them be separate is nice as if we ultimately decide to remove one or two after the experiment we can remove those components without worry. In my initial filter implementation I had both all the logic in the legacy screen for filters and it got quite confusing reading through the component. I imagine it being much worse to debug if a filter, search, search and filter variant are added to one component.

src/dappsExplorer/search/NoResults.tsx
MuckT: Fixed in 6b9882a368accea810f9c533e9082c7fba72f920!

src/dappsExplorer/search/FavoriteDappsSection.tsx
MuckT: Fixed in d3a9def639513851398c015134bd7fa503cdd10f!

src/dappsExplorer/search/FavoriteDappsSection.tsx
MuckT: Fixed in 6b9882a368accea810f9c533e9082c7fba72f920!

PR #3602, opened 2023-03-30
feat: dapp list search analytics

MuckT:
### Description

Broken out from #3576. Adds one analytic event for the dapp list search. Uses `lodash`'s debounce to assist with getting the desired user analytics. This event is triggered 1000ms after the user stops typing in the search input.

```
[DappExplorerEvents.dapp_search]: {
  searchTerm: string
}
```

#### Example Analytics Flow: User enters 'Eth'
----- E -----
`searchTerm='E'`

----- t -----
`searchTerm='Et'`

----- h -----
`searchTerm='Eth'`

----- 1000 MS -----
`searchTerm='Eth'`
Analytics for search submitted:
`ValoraAnalytics.track(DappExplorerEvents.dapp_search, { searchTerm })`

### Test plan

### Related issues
Fixes #RET-580

### Backwards compatibility

Yes


kathaypacific (commented):
is it possible to simplify the requirements here? what is the `clearing` property and why is it important to track this (i.e. what do we learn from the interaction here)?

my assumption (without having put a lot of thought in, so i might be missing something) is that we want to understand what users are searching for. so, if they begin typing and pause, that can be interpreted as a search that they have "committed to". so it makes sense to track this. if it's something that a user types and backspaces quickly, then i'm not sure what the value of tracking that is.

i had mentioned a simplification in this [comment](https://github.com/valora-inc/wallet/pull/3576#discussion_r1150452546), i'm not sure if it was clear so i made a [branch](https://github.com/valora-inc/wallet/compare/tomm/dapps-list/search-analytics...kathy/dapp-search-analytics-alternative?expand=1) to quickly show what i meant. unless the backspace / "clearing" property very important from an analysis perspective, i'd favour a simple approach like this since it avoids some TS errors + 3 less state variables + one less new dependency? wdyt?

MuckT:
@kathaypacific I tried the above approach from [linked branch](https://github.com/valora-inc/wallet/compare/tomm/dapps-list/search-analytics...kathy/dapp-search-analytics-alternative?expand=1) and it does not appear to be working as intended.

#### Example analytics from the linked branch:

```
Analytics Fired with:  'T'
Analytics Fired with:  'Th'
Analytics Fired with:  'Thi'
Analytics Fired with:  'This'
Analytics Fired with:  'This '
Analytics Fired with:  'This i'
Analytics Fired with:  'This is'
...
```
#### Example Analytics from this PR
```
Analytics Fired with:  'This is a test'
```

M-shin (approved):
lgtm. We might miss some `clearing` events if the user begins typing something new within 1000ms, will that be fine?

MuckT:
@M-shin I checked with @ellenwoconnor and it sounds like we can instead look at whether the user opened a dapp after searching to determine if a search was successful. Knowing this I've removed the clear analytics in 3a2da0f62e1056f0927993f823cb8d2f80769c19 which simplifies things.

package.json
```diff
@@ -184,6 +184,7 @@
     "statsig-react-native": "^4.9.2",
     "svgs": "^4.2.0",
     "tslib": "^2.4.1",
+    "use-debounce": "^9.0.3",
```
jeanregisser, line 187: Since we already use `debounce` from `lodash`, could we avoid this new dependency?

MuckT: I tried using Lodash and couldn't quite get to it to work. `use-debounce` is a pretty small library, plays with React well, and we also need the functionality in #3613. It would be nice to keep if the build size isn't increased much.

jeanregisser: yes it's not so much the build size, but trying to keep dependencies under control, since we're already struggling a bit with this.
Happy to help here.

MuckT: Got it sorted in 9f1d1e1367e7eafb7f2d7714c5993f0e67676d29 with help from @dievazqu!

src/dappsExplorer/search/DAppsExplorerScreenSearch.test.tsx
```diff
@@ -345,5 +346,39 @@ describe(DAppsExplorerScreenSearch, () => {
       // Names display correctly in the all dapps section
       expect(within(allDappsSection).getByText(dappsList[0].name)).toBeTruthy()
     })
+
+    it('triggers events when searching', async () => {
+      jest.useRealTimers()
```
jeanregisser, line 351: I'm proposing a patch to avoid real timers: https://github.com/valora-inc/wallet/pull/3628

jeanregisser (approved):
Approving already but see the suggested change before merging.

PR #3603, opened 2023-03-30
chore: add remote config variable for decentralised phone verification paths

kathaypacific:
### Description

When odis v1 is EOL'ed, we will no longer be able to get the phone hash using ODIS. This means we should:
- do not try to fetch on-chain phone numbers during the phone number lookup in the send flow
- do not try to background migrate to CPV
- remove option to revoke on-chain mapping in the debug settings menu

### Test plan

n/a

### Related issues

- Fixes RET-666

### Backwards compatibility

Y


src/account/Settings.tsx
```diff
@@ -100,6 +101,7 @@ interface StateProps {
   supportedBiometryType: BIOMETRY_TYPE | null
   shouldShowRecoveryPhraseInSettings: boolean
   hapticFeedbackEnabled: boolean
+  odisV1EOL: boolean
```
jeanregisser, line 104: nit: seems simpler, no?
```suggestion
  odisV1Enabled: boolean
```

src/app/saga.ts
```diff
@@ -384,8 +386,9 @@ export function* handleSetAppState(action: SetAppState) {
 }
 
 export function* runCentralPhoneVerificationMigration() {
+  const odisV1EOL = yield select(odisV1EOLSelector)
   const shouldRunVerificationMigration = yield select(shouldRunVerificationMigrationSelector)
```
jeanregisser, line 390: nit: We should update `shouldRunVerificationMigrationSelector` instead and add the new condition there?
Otherwise this selector name becomes a bit misleading.

jeanregisser (approved):
Thanks! 👍

src/account/Settings.tsx
kathaypacific: lol, yes this is better. i updated

src/app/saga.ts
kathaypacific: ah oops, yes you're right. updated!

src/account/Settings.tsx
kathaypacific: lol, i'm also fine with that. updated

PR #3606, opened 2023-03-30
fix: Add invite reward banner for cUSD invite rewards

M-shin:
### Description

Fixes a missing change that should have been in #3555. The invite rewards banner in the send flow should properly indicate the expected reward type.

### Test plan

Added unit test and manual tested in emulator.

### Related issues

- Fixes RET-635

### Backwards compatibility

Backwards compatible. No previous translation keys changed, only added new.

### DIFF
--- a/locales/en-US/translation.json
+++ b/locales/en-US/translation.json
@@ -596,6 +596,10 @@
     "title": "Invite friends and collect NFTs",
     "body": "Once they set up their {{appName}} wallet and connect their number, you’ll each get an exclusive NFT."
   },
+  "inviteRewardsBannerCUSD": {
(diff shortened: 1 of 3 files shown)


locales/en-US/translation.json
MuckT, line 1: We want to avoid modifying this file otherwise the Crowdin PR's can have issues.

MuckT (changes requested):

kathaypacific (approved):
thanks! like Tom said, we should revert any changes to the en-US translation file before merging :)

MuckT (approved):
👍

PR #3609, opened 2023-03-31
chore: reduce unnecessary fetching of supercharge rewards

kathaypacific:
### Description

Since the supercharge v2 backend computation is relatively intensive, it'd be nice to reduce the redundant calculations. This will save us some rpc calls quota. This PR aims to implement some kind of cache.

There are 2 places that fetch supercharge rewards:
1. home screen - specifically, the homecards. the rewards data is used to display the supercharge homecards.
2. supercharge screen

I think it is unnecessary to keep fetching supercharge rewards on the home screen. The worst case scenario that comes from this change, is that the "claim your rewards" homecard is displayed later.

This PR:
- add a new redux variable / migration to track last reward fetch time
- in the notifications, only fetch rewards if the previous fetch was more than 24 hours ago

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y

### DIFF
--- a/src/home/NotificationBox.tsx
+++ b/src/home/NotificationBox.tsx
@@ -25,6 +25,8 @@ import SimpleMessagingCard, {
 } from 'src/components/SimpleMessagingCard'
 import { RewardsScreenOrigin } from 'src/consumerIncentives/analyticsEventsTracker'
 import {
+  availableRewardsSelector,
+  rewardsRefreshTimestampSelector,
   superchargeInfoSelector,
   userIsVerifiedForSuperchargeSelector,
 } from 'src/consumerIncentives/selectors'
@@ -96,6 +98,8 @@ interface Notification {
   id: string
 }
 
+const REFRESH_REWARDS_INTERVAL = 1000 * 60 * 60 * 24 // 24 hours
+
 function useSimpleActions() {
   const {
     backupCompleted,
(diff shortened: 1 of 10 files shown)


src/home/NotificationBox.tsx
```diff
@@ -96,6 +98,8 @@ interface Notification {
   id: string
 }
 
+const REFRESH_REWARDS_INTERVAL = 1000 * 60 * 60 * 24 // 24 hours
```
kathaypacific, line 101: we could also move this to remote config? 🤔

jeanregisser (commented):
Great idea, wondering if we should do this client side change or set the right cache headers instead in the Cloud Function response? this way we wouldn't even need to add this in the client.

jeanregisser:
I mean the normal HTTP cache headers that all well behaved HTTP clients respect.

Specifically, we could return something like `Cache-Control: max-age=86400` in the supercharge response headers so the wallet would continue serving the cached response for 1 day.

kathaypacific:
> I mean the normal HTTP cache headers that all well behaved HTTP clients respect.
> 
> Specifically, we could return something like `Cache-Control: max-age=86400` in the supercharge response headers so the wallet would continue serving the cached response for 1 day.

ooooohhhhhh makes sense, well then if it doesn't need a client side change then i'll do it when i'm back :D

PR #3610, opened 2023-03-31
feat(fiatExchange): Don't show add funds when pressing withdraw home action

jophish:
### Description

Fixes ACT-705. Adds a new version of the add/withdraw screen, without the "add" option, and navigates to it from the home carousel.

There were two "issues" with this.
1. Components used for Drawer navigator screens cannot accept props; Typescript will yell at you.
2. We will want to be able to _not_ have the Add/Withdraw section appear in the drawer nav, but make it navigable from the home actions section. In order to do this, we must _not_ register the Add/Withdraw screen with the Drawer navigator (so it doesn't appear in the side nav); removing the screen from the Drawer navigator and _not_ adding it to our regular navigator makes it un-navigable, so we must somehow register it with our normal navigator.

In order to do this, the `FiatExchange` screen was component-ized and parameterized. Two wrappers were created, neither accepting props; one to be used by the drawer navigator (with behavior unchanged) and a new which removes the add funds section and drawer nav header.

### Test plan

Unit and manual tested.

https://user-images.githubusercontent.com/569401/229833998-b8346a89-445f-4b4c-a6f8-dc70cb4b74a1.mp4

### Related issues

- Fixes ACT-705.

### Backwards compatibility

Yes.


src/navigator/DrawerNavigator.tsx
```diff
@@ -315,6 +315,7 @@ export default function DrawerNavigator({ route }: Props) {
       )}
       <Drawer.Screen
         name={Screens.FiatExchange}
+        // @ts-expect-error component type in native-stack v6
```
jophish, line 318: this is copied from the `BackupIntroduction` component above

src/account/FiatExchange.tsx
```diff
@@ -23,9 +23,20 @@ import { navigateToURI } from 'src/utils/linking'
 import Logger from 'src/utils/Logger'
 
 function FiatExchange() {
+  return <FiatExchangeSection />
+}
+
+export function FiatExchangeSection({
+  hideAddFunds,
+  hideDrawerTopNav,
```
satish-ravi, line 31: nit: should we use `show` instead of `hide`, Since we've been using `show` for our parameters?

cajubelt: ➕

cajubelt (approved):
looks good, though mind the extra padding Satish noted

cajubelt:
in the future, please add screenshots or videos from manual tests so reviewers don't need to run simulators themselves to catch issues like the one Satish found

src/account/FiatExchange.tsx
```diff
@@ -72,21 +83,23 @@ function FiatExchange() {
 
   return (
     <SafeAreaView style={styles.container}>
-      <DrawerTopBar />
+      {!hideDrawerTopNav && <DrawerTopBar testID="FiatExchange/DrawerBar" />}
```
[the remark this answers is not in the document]
jophish: I think we'd have to move the top-most `SafeAreaView` out into `FiatExchange` and `WithdrawSpend` in that case; I went with this since it felt a bit simpler.

PR #3612, opened 2023-03-31
docs(bottom-sheet): mark BottomSheet.tsx as deprecated with jsdoc

cajubelt:
### Description

per Kathy's advice [here](https://valora-app.slack.com/archives/C029MMTD529/p1680163435208949)

This makes it harder to accidentally add new usages of the component we don't want to use anymore, because an IDE will remind you that it's deprecated
Screenshot 2023-03-31 at 2 33 55 PM

### Test plan

na

### Related issues

na

### Backwards compatibility

na


satish-ravi (approved):
💯

jophish (approved):

PR #3613, opened 2023-04-01
feat(swaps): Adding support for new tokens and search box

dievazqu:
### Description

Adding search box and new tokens to the swaps bottom sheet.
Adding sticky header feature to the Bottom Sheet component.

### Test plan

Turn on the config in the `swapping_non_native_tokens` statsig experiment.

### Related issues

- Fixes #RET-611

### Backwards compatibility

The changes are backward compatible.

### Demo
https://user-images.githubusercontent.com/12504748/229369125-821c8221-a82a-4a6a-b687-262dbb9833f5.mp4


src/components/TokenBottomSheet.test.tsx
```diff
@@ -89,25 +94,26 @@ describe('TokenBottomSheet', () => {
     jest.clearAllMocks()
   })
 
-  function renderPicker(visible: boolean, isInvite: boolean = false) {
```
dievazqu, line 92: Nobody was calling renderPicker with `isInvite`.

src/tokens/selectors.ts
```diff
@@ -124,6 +124,10 @@ export const celoAddressSelector = createSelector(coreTokensSelector, (tokens) =
   return tokens.find((tokenInfo) => tokenInfo.symbol === 'CELO')?.address
 })
 
+export const swappableTokensSelector = createSelector(tokensByUsdBalanceSelector, (tokens) => {
+  return tokens.filter((tokenInfo) => tokenInfo.isSwappable === true)
```
silasbw, line 128: do you need to `=== true` (or can you test for truthy)?

src/send/Send.tsx
```diff
@@ -221,6 +222,10 @@ function Send({ route }: Props) {
     return null
   }
 
+  const sortedTokens = (isOutgoingPaymentRequest ? stableTokens : tokensWithBalance).sort(
+    sortFirstStableThenCeloThenOthersByUsdBalance
```
silasbw, line 226: Is this change needed for `Adding support for new tokens and search box`?  It's not apparent it's related.

`sortFirstStableThenCeloThenOthersByUsdBalance` connotes a function behavior that doesn't sound super scalable either (and I'm wondering if we should avoid using that function).

src/tokens/utils.ts
```diff
@@ -25,6 +25,19 @@ export function sortByUsdBalance(token1: TokenBalance, token2: TokenBalance) {
   return token2UsdBalance.comparedTo(token1UsdBalance)
 }
 
+export function sortByUsdBalanceThenByAlphabetical(token1: TokenBalance, token2: TokenBalance) {
```
silasbw, line 28: If this is used only in the `SwapScreen` I'd personally rather have this function live with that code and renamed to be, like, `tokenCompare` to match existing naming conventions (e.g., `.comparedTo` or `localeCompare`, etc).

silasbw (commented):
awesome

src/tokens/selectors.ts
dievazqu: lol, no

src/send/Send.tsx
dievazqu: I extract this logic from the TokenBottom sheet component.

src/tokens/utils.ts
dievazqu: Sure, I can move it there

src/send/Send.tsx
dievazqu: I think we will need to revisit how to sort when we add support for multichain

src/components/TokenBottomSheet.tsx
```diff
@@ -56,33 +61,115 @@ function TokenOption({ tokenInfo, onPress }: { tokenInfo: TokenBalance; onPress:
...
+    setTokenList(tokens)
+  }
+
+  const onCloseBottomSheet = () => {
+    resetSearchState()
+    onClose()
+  }
+
   const onTokenPressed = (tokenAddress: string) => () => {
+    resetSearchState()
```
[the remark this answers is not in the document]
dievazqu: It was to clean up the state. Without this, when you open the bottom sheet again the searched text was still there.
I will put the reset after calling the `onTokenSelected`, so the bottom sheet should be dismissed before reseting the search text. Or should I reset the state somewhere else?

src/components/TokenBottomSheet.tsx
```diff
@@ -56,33 +61,115 @@ function TokenOption({ tokenInfo, onPress }: { tokenInfo: TokenBalance; onPress:
...
+function TokenBottomSheet({
+  isVisible,
+  origin,
+  onTokenSelected,
+  onClose,
+  tokens,
+  searchEnabled,
+  titleText,
+}: Props) {
+  const [tokenList, setTokenList] = useState(tokens)
```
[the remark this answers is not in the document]
dievazqu: I see, it makes sense! Thanks for the explanation!

src/components/TokenBottomSheet.tsx
```diff
@@ -56,33 +61,115 @@ function TokenOption({ tokenInfo, onPress }: { tokenInfo: TokenBalance; onPress:
...
   const onTokenPressed = (tokenAddress: string) => () => {
+    resetSearchState()
     ValoraAnalytics.track(SendEvents.token_selected, {
       origin,
       tokenAddress,
     })
     onTokenSelected(tokenAddress)
   }
 
+  const throttleSearch = throttle((searchInput: string) => {
```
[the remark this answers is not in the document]
jeanregisser: @MuckT needed to do something similar for the dapps search.
I think he opted to do the debounce only for the analytics event.
Which seems good given the data to filter was relatively small.
I guess we could decide to do the same here too.

locales/base/translation.json
```diff
@@ -1664,6 +1664,10 @@
     "generateName": "Generate a name",
     "namePlaceholder": "Select a name ex) {{exampleName}}"
   },
+  "tokenBottomSheet": {
+    "searchAssets": "Search Assets",
+    "noTokenInResult": "No matching token for \"{{searchQuery}}\""
```
MuckT, line 1669: I think this should be similar to the no text used in search so we can bold just the searchQuery / searchTerm to match the styling of Dapps Search: [Figma Comment](https://www.figma.com/file/jRtYI8GiSvPAaClBnRLLpu?node-id=2520:26583#363942229).
- [Translations file](https://github.com/valora-inc/wallet/pull/3599/files#diff-d3d67f4b3f8dbec345426f534268e40d9b087dfa5d83750165d63caf136cb24cR1288).
- [Styling in component](https://github.com/valora-inc/wallet/pull/3599/files#diff-0a07a8e746d1acf171a5b752ade521ea308d045d70283553323e2fc19af6d2cbR14-R27).

MuckT:
@dievazqu I think we can remove a lot of the debounce logic if we use [use-debounce](https://www.npmjs.com/package/use-debounce) for search analytics instead.

```JavaScript
const [searchTerm, setSearchTerm] = useState('')
const [analyticsSearchTerm] = useDebounce(searchTerm, 1000)

  // Fire Analytics event when search term changes and is not empty
useEffect(() => {
  if (analyticsSearchTerm) {
    ValoraAnalytics.track(TokenBottomSheetEvents.search_token, {
      origin,
      searchInput: analyticsSearchTerm,
    })
  }
}, [analyticsSearchTerm])
  ```

**Comment by MuckT** [comment]
@dievazqu also for some reason this is crashing for me on Android, I have only been able to test on iOS. Since we don't have swap e2e tests this is something to keep in mind.

File: src/components/BottomSheet.tsx [hunk 9]
```diff
@@ -19,6 +19,8 @@ interface Props {
   testID?: string
   opacity?: number
   backgroundColor?: string
+  stickyHeader?: JSX.Element
+  fullHeight?: boolean
```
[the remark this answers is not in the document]
jeanregisser: Thanks for explaining! Your solution looks good then 👍

src/components/TokenBottomSheet.test.tsx
```diff
@@ -80,34 +88,113 @@ const onTokenSelectedMock = jest.fn()
...
           tokens={tokens}
+          searchEnabled={searchEnabled}
         />
       </Provider>
     )
   }
 
+  // TODO before merge: Investigate why this tests are failing if put at the end of the file.
+  it('renders and behaves correctly when the search is enabled', async () => {
+    jest.useRealTimers()
```
jeanregisser, line 118: Can we avoid this by advancing timers?
Otherwise it slows down tests unnecessarily.

jeanregisser:
> @dievazqu I think we can remove a lot of the debounce logic if we use [use-debounce](https://www.npmjs.com/package/use-debounce) for search analytics instead.

I'd recommend not adding another dependency which we can be covered by `lodash`
https://github.com/valora-inc/wallet/pull/3602#discussion_r1157001660

src/components/TokenBottomSheet.test.tsx
dievazqu: I have tried to use fake timers, but I couldn't make it work. I'll give it a try today again, but I think I will need help. Can I schedule a short meeting to do pair programming session to solve this?

dievazqu:
@jeanregisser @silasbw @MuckT - Thanks for the reviews! I have done all the suggested changes

Also I created a [PR](https://github.com/valora-inc/wallet/pull/3626) only with the changes in the bottom sheet to add the sticky header, so it's easier to review :)

jeanregisser (commented):
Thanks for splitting the Bottom sheet changes.
Could you rebase this PR on top of it, to make this one easier to review then?

src/components/TokenBottomSheet.test.tsx
jeanregisser: Sorry didn't see that. Feel free to schedule meetings whenever you want.

dievazqu: No worries, I finally found the problem was solved by using `jest.useFakeTimers('modern')` 🙌

src/components/TokenBottomSheet.tsx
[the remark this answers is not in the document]
MuckT: There may be some value to debouncing before searching even if it's a low threshold such as 200ms.
> Purposefully adding a delay to a process can actually increase its perceived value and instill a sense of trust, even when the process itself actually takes much less time.

[Doherty Threshold](https://lawsofux.com/doherty-threshold/)

jeanregisser (approved):
🚀

PR #3614, opened 2023-04-03
feat: hide Swap and Add & Withdraw menu items for Home Screen Actions experiment

anthonydaoud:
### Description

As part of an upcoming [experiment](https://www.notion.so/actions-on-the-homescreen-50725477376047a892f736d40eaa089f), one of the experiment variants will not have the Swap and Add & Withdraw actions visible in the hamburger menu. This PR sets up the logic to hide these items based on the experiment parameters obtained from Statsig.

#### Menu without the actions

menuWithoutSwapAddWithdraw

### Test plan

Added unit tests.

### Related issues

- ACT-703

### Backwards compatibility

Yes.


src/navigator/DrawerNavigator.test.tsx
```diff
@@ -64,4 +72,33 @@ describe('DrawerNavigator', () => {
     )
     expect(queryByTestId('DrawerItem/dappsScreen.title')).toBeNull()
   })
+
+  it('hides add/withdraw menu item based on statsig experiment param', () => {
+    ;(getExperimentParams as jest.Mock).mockReturnValueOnce({
```
satish-ravi, line 77: nit: use `mocked` util from jest over `as jest.Mock`

src/navigator/DrawerNavigator.tsx
```diff
@@ -313,11 +316,13 @@ export default function DrawerNavigator({ route }: Props) {
           initialParams={{ showDrawerTopBar: true }}
         />
       )}
-      <Drawer.Screen
-        name={Screens.FiatExchange}
-        component={FiatExchange}
-        options={{ title: t('addAndWithdraw'), drawerIcon: AddWithdraw }}
-      />
+      {showAddWithdrawOnMenu && (
```
satish-ravi, line 319: one side effect of this change is that the "Withdraw" h no longer works for the variant where this is hidden. I believe @jophish is working on making two different screens, one for the drawer and one for the home action, which we should also do for the Swap screen. Doesn't need to be on this PR and could be part of ACT-704, but maybe we wait on merging this until those changes go in since it breaks a variant?

cajubelt (approved):

PR #3615, opened 2023-04-03
fix(statsig): initialize before home screen requests experiments

satish-ravi:
### Description

Currently statsig initializes asynchronously (part of ValoraAnalytics.init) and the home screen requests experiments before initialization completes, resulting in loading the default experience and then switching to the appropriate variant. This makes analytics init a blocking call before rendering the first screen (home / onboarding)

### Test plan

Unit test and manual. Also ensured the below `getExperimentParams` warn log doesn't show up:

```
getExperimentParams Error getting experiment params [Error: Call and wait for initialize() to finish first.] 
```

Before:

https://user-images.githubusercontent.com/5062591/229637645-73c24b62-61d9-475a-9459-a74d2e928c0d.mp4

After:

https://user-images.githubusercontent.com/5062591/229638156-244bcac9-9199-4593-bc05-ae194a40ab2d.mp4

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/src/app/AppInitGate.test.tsx
+++ b/src/app/AppInitGate.test.tsx
@@ -26,7 +26,7 @@ jest.mock('src/navigator/NavigationService', () => ({
 const mockedI18n = I18n as jest.Mocked<typeof I18n>
 const setLanguageSpy = jest.spyOn(I18nActions, 'setLanguage')
 
-const renderI18nGate = (language: string | null) =>
+const renderAppInitGate = (language: string | null) =>
   render(
     <Provider store={createMockStore({ i18n: { language } })}>
       <AppInitGate loading={<Text>Loading component</Text>}>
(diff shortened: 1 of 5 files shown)


src/app/AppInitGate.test.tsx
satish-ravi, line 1: Repurposed this to initialize both i18n and analytics (which includes statsig. Open to suggestions on other names for this component.

src/app/AppInitGate.test.tsx
```diff
@@ -96,4 +102,14 @@ describe('I18nGate', () => {
     expect(mockedI18n.initI18n).toHaveBeenCalledTimes(1)
     expect(navigateToError).toHaveBeenCalledWith('appInitFailed', 'some error')
   })
+
+  it('should show error screen in case of failed analytics init', async () => {
```
satish-ravi, line 106: this should never happen because we wrap both segment and statsig inits in try catches [here](https://github.com/valora-inc/wallet/blob/main/src/analytics/ValoraAnalytics.ts#L108-L152). We could also add another try / catch wrapper for ValoraAnalytics.init and don't throw error in case of failure and continue with app load. Thoughts?

src/app/AppInitGate.tsx
```diff
@@ -18,35 +19,41 @@ interface Props {
...
-      }
+      Logger.debug('AppInitGate', 'Starting init')
+      await Promise.all([i18nInitializer(), ValoraAnalytics.init()])
+      Logger.debug('AppInitGate', 'init completed')
     },
     [],
     {
       onError: (error) => {
-        Logger.error('i18n', 'Failed init i18n', error)
+        Logger.error('AppInitGate', 'Failed init', error)
```
jh2oman, line 49: super nit: Might as well make a TAG variable since we have three logs for 'AppInitGate'

jh2oman (approved):

jophish (approved):

PR #3616, opened 2023-04-03
fix(statsig): Refactor Statsig initialization and user object updates

jophish:
### Description

Refactors our Statsig initialization and user object updating to ensure the following:
1. `startOnboardingTime` is available to statsig on app launch.
2. _Meaningful_ fetches to Statsig will never occur while the SDK is uninitialized.

See [this Slack thread](https://valora-app.slack.com/archives/C02KBT0DAHJ/p1680715814180279) and the following discussion for more information and context on the approach used here.

What this does/some more context:

Previously, we were updating the Statsig user object whenever a user trait changes, in the background. This can happen at any time, asynchronously. When updating the Statsig user, the Statsig SDK is briefly re-initialized, during which experiment fetching fails due to the SDK being uninitialized. This meant that there was a race condition where experiments could return default values if fetched while a user trait was being updated.

Additionally, when initializing the Statsig SDK, and while updating the Statsig user object, we were not passing _all_ properties, just the ones we wanted to update. `Statsig.updateUser` actually functions as a `PUT` operation instead of a `PATCH`, so we need to specify all properties we want, _every time_ we call `Statsig.updateUser`.

To solve the first problem, we now _explcitly_ call `Statsig.updateUser` at key moments and _block on its completion_ instead of asynchronously. Namely, when:
1. `startOnboardingTime` is determined for the first (and only) time.
2. Onboarding is completed, and the user's `walletAddress` is guaranteed to exist.

`Statsig.updateUser` is _never called_ after onboarding is complete. It is called _exactly_ twice for a user, ever. For 1., we block on `Statsig.updateUser` _before_ navigating to the first onboarding page that requires experiment parameters, in order to ensure that the correct parameters are fetched. For 2., when onboarding is completed, we fire off a new action that triggers a saga responsible for calling `Statsig.updateUser`, waiting for it to finish, and navigating to a specified post-onboarding screen (which is assumed to depend on the Statsig user object being up-to-date). Because the navigate necessarily occurs after the `Statsig.updateUser` resolves, the post-onboarding screen will always fetch the correct parameters, and not result in uninitialized fetches.

This approach of using a saga _does present_ the possibility of a small number of non-meaningful fetches to Statsig being made in the uninitialized state. This is because the last screen in onbaording will _always_ depend on Statsig experiment params. WLOG, consider the Phone Verification screen. When a user presses the "Skip" button, we kick off the saga to update the Statig user object and navigate to the post-onboarding screen. Pressing the "Skip" button _also_ induces state change, which triggers a re-render. The Phone Verification screen relies `getOnboardingStepValues` method in order to determine the steps to display in the header, which in turn relies on a fetch to Statsig to get experiment params. If this re-render occurs _while_ `Statsig.updateUser` is being called via the saga, the SDK will be briefly uninitialized, and the calls made to Statsig while re-determining onboarding steps for the re-render will return default values due to the SDK being uninitialized. The end-user impact of this is that there is a _chance_ that step numbers may (very) briefly be incorrect between the "Skip" button press and the navigate to the post-onboarding screen. This is something that's tolerable for the time being, though not ideal, and can likely be avoided when we make the switch to using the RN-specific `StatsigProvider` pattern.

To solve the _second_ problem from above (not passing all properties to Statsig `initialize` and `updateUser`), we've done two things:
1. When initializing the Statsig SDK, fetch all relevant parameters needed for the user object from Redux.
2. Wrap all calls to `Statsig.updateUser` in a `PATCH`-like interface, which merges default values from Redux with optional overrides from the caller.

### Test plan

Unit and manual tested. I would appreciate if reviewers could test this locally by performing the following steps:

1. Pull this branch.
2. Build/run a dev build.
3. In the dev build, reset Valora.
4. Open Flipper (or whatever you use to monitor logs) and filter for "Statsig".
5. Begin onboarding, and stop on the Phone Verification screen. It does not matter what experiment variants you are bucketed into. You should not see any logs.
6. Press the "Skip" button. You should see an action dispatched, and potentially, a small number of logs warning about some uninitialized calls to Statsig (as mentioned above).
7. Mess around in the app briefly, close and re-open, etc. You should not see any more logs (ever), and should have the same experimental variants as before.

While testing, I would encourage you to follow along with the exposure events in the diagnostics view of the [recovery phrase experiment](https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/experiments/recovery_phrase_in_onboarding/diagnostics) or the [choose your adventure experiment](https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/experiments/choose_your_adventure/diagnostics) (make sure to enable non-production logs), and make sure that everything looks right to you.

In between steps 5. and 6., you can try closing the app and re-opening it; the behavior should be the same as mentioned above (we experienced some issues this during testing, but have fixed them; testing this is mostly just a sanity check).

### Related issues

N/A

### Backwards compatibility

Yes.


src/statsig/index.ts
```diff
@@ -52,3 +56,14 @@ export function getDynamicConfigParams<T extends Record<string, StatsigParameter
     return defaultValues
   }
 }
+
+export async function updateStatsigUser(statsigUser?: StatsigUser) {
+  const state = store.getState()
```
jophish, line 61: I wasn't sure of the correct pattern to use to get state from the store outside a component. We use this pattern in a few other places that I saw, so copied it here, but open to other suggestions!

src/onboarding/welcome/Welcome.test.tsx
```diff
@@ -5,12 +5,15 @@ import { navigate } from 'src/navigator/NavigationService'
...
 import Welcome from 'src/onboarding/welcome/Welcome'
-import { Statsig } from 'statsig-react-native'
+import { updateStatsigUser } from 'src/statsig'
 import { createMockStore } from 'test/utils'
 import { mocked } from 'ts-jest/utils'
 
 jest.mock('src/onboarding/steps')
-jest.mock('statsig-react-native')
+jest.mock('src/statsig', () => ({
+  ...(jest.requireActual('src/statsig') as any),
```
satish-ravi, line 14: any reason to requireActual everything else here?

src/statsig/index.ts
```diff
@@ -52,3 +56,14 @@ export function getDynamicConfigParams<T extends Record<string, StatsigParameter
...
+  const defaultUser = {
+    userID: walletAddressSelector(state),
+    custom: {
+      startOnboardingTime: startOnboardingTimeSelector(state),
+    },
+  }
+  await Statsig.updateUser(_.merge(defaultUser, statsigUser))
```
satish-ravi, line 72: does this do a deep merge? Can you add a test case for when custom includes properties but `startOnboardingTime` isn't one of them? We don't have a use case now, but in the future there could be a case where the default user includes multiple custom properties and `updateStatsigUser` is invoked with a subset of custom properties

src/statsig/index.ts
```diff
@@ -52,3 +56,14 @@ export function getDynamicConfigParams<T extends Record<string, StatsigParameter
     return defaultValues
   }
 }
+
+export async function updateStatsigUser(statsigUser?: StatsigUser) {
```
satish-ravi, line 64: This makes an assumption that whatever is calling this with an updated property also updates the defaults in redux right? Should we document it?

Also, wonder if this could lead to the following race condition
1. Welcome screen updates startOnboardingTime and then updates redux store
2. Some user trait changes, which invokes identify
3. Identify invokes updateStatsigUser, but startOnboardingTime is not yet on redux store
4. StatsigUser no longer includes startOnboardingTime

Don't think this specifically can happen since the wallet address is created later in the flow, but this could be problem more generally if we have multiple custom properties and the redux store isn't updated with latest values right?

satish-ravi (commented):
Have you also tested that app reloads include `startOnboardingTime` on statsig requests?

src/onboarding/welcome/Welcome.test.tsx
jophish: yeah, we want `getExperimentParams` to return the default value rather than being a jest mock; tests fail otherwise

src/statsig/index.ts
jophish: yeah it does do a deep merge -- added a test to explicitly test for this

jophish:
@satish-ravi After adding an `updateStatsigUser` call to `ValoraAnalytics.init`, I verified that subsequent requests to Statsig after app reload do indeed contain a `startOnboardingTime` field. See [https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/experiments/recovery_phrase_in_onboarding/diagnostics], and look for subsequent requests from stable ID `19d4d967-73fd-4da6-8ff0-1bf7aa9fdb60` in non-production logs.

src/analytics/ValoraAnalytics.ts
```diff
@@ -130,14 +131,8 @@ class ValoraAnalytics {
...
-      const { walletAddress } = getCurrentUserTraits(store.getState())
-      const statsigUser =
-        typeof walletAddress === 'string'
-          ? {
-              userID: walletAddress,
-            }
-          : null
-
+      const defaultStatsigUser = getDefaultStatsigUser()
+      const statsigUser = defaultStatsigUser.userID ? defaultStatsigUser : null
```
satish-ravi, line 135: Should we set this regardless of the userID existing since there are other properties?

src/statsig/index.ts
```diff
@@ -56,3 +60,26 @@ export function getDynamicConfigParams<T extends Record<string, StatsigParameter
     return defaultValues
   }
 }
+
+/* Updates the current Statsig user. If no argument is given, a default StatsigUser
```
satish-ravi, line 64:
```suggestion
/**
 * Updates the current Statsig user. If no argument is given, a default StatsigUser
```

to make this a jsdoc

src/statsig/index.ts
satish-ravi: If there's any chance at all of affecting the onboarding experiments, it's worth addressing right? If not, should be ok, but something to watch out for if/when we add more props

cajubelt: i feel like this kind of problem is hard to fully avoid if we cache statsig user properties in redux with the goal of doing patch updates to the statsig user. without synchronously updating redux before calling updateStatsigUser, there's always some chance that the cache could be out of date when we update the statsig user. and I'm not really sure what to do about that.

src/analytics/ValoraAnalytics.ts
```diff
@@ -203,23 +205,25 @@ class ValoraAnalytics {
...
       return
     }
 
+    // Ensure that Statsig user is updated even if analytics is disabled.
+    // This is primarily for dev builds, where analytics is disabled by default.
     try {
-      void Statsig.updateUser({ userID })
+      await updateStatsigUser({ userID })
     } catch (error) {
       Logger.warn(TAG, 'Error updating statsig user', error)
```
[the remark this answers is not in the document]
jophish: Error as in `Logger.error`? I agree, will update.

src/analytics/ValoraAnalytics.ts
jophish: I think it's safe to do so? I added this line in here for the sake of backwards compatibility, but I think we should be okay to set custom properties even when the user ID doesn't exist.

src/analytics/ValoraAnalytics.ts
```diff
@@ -203,23 +205,25 @@ class ValoraAnalytics {
...
     }
 
+    if (!this.isEnabled()) {
```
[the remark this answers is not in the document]
jophish: Agreed, I like `trackingEnabled` -- updated.

src/onboarding/steps.ts
```diff
@@ -127,8 +129,8 @@ export function getOnboardingStepValues(screen: Screens, onboardingProps: Onboar
     currentScreen = nextScreen
   }
 
-  const toHomeStep = () => {
-    currentScreen = Screens.WalletHome
+  const finishOnboarding = (nextScreen: Screens) => {
```
cajubelt, line 132: i wonder if this should log an error or something if nextScreen isn't in END_OF_ONBOARDING_SCREENS

cajubelt (approved):
looks good, didn't have time to test yet

src/account/saga.ts
```diff
@@ -220,6 +227,21 @@ export function* handleUpdateAccountRegistration() {
   }
 }
 
+export function* updateStatsigAndNavigate(action: UpdateStatsigAndNavigateAction) {
+  // Wait for wallet address to exist before updating statsig user
+  yield call(getWalletAddress)
+  yield call(patchUpdateStatsigUser)
+  if (action.screen === Screens.WalletHome) {
+    navigateHome()
+  } else if (action.screen) {
```
satish-ravi, line 236:
```suggestion
  } else {
```
we don't need this if since it's required now right?

src/statsig/index.ts
```diff
@@ -30,11 +37,18 @@ export function getExperimentParams<T extends Record<string, StatsigParameter>>(
...
+        experiment
+      )
+    }
     return getParams({ config: experiment, defaultValues })
   } catch (error) {
     Logger.warn(
-      'getExperimentParams',
-      `Error getting params for experiment: ${experimentName}`,
+      TAG,
+      `getExperimentParams: Error getting params for experiment: ${experimentName}`,
```
satish-ravi, line 51: nit: similar change for `getDynamicConfig`?

src/onboarding/steps.ts
satish-ravi: we could just make the arg of type of elements in the `END_OF_ONBOARDING_SCREENS` array so we can catch it compile time.

satish-ravi (approved):

src/account/saga.ts
jophish: just wasn't sure if this would fall through to `navigate` after calling `navigateHome` if we're going to the home screen

PR #3622, opened 2023-04-05
fix(notifications): show notifs perms dialogue on first homescreen visit

cajubelt:
### Description

Previously, the dialogue for setting permissions on sending push notifications was shown on wallet creation. This meant we'd get the dialogue showing up in awkward places, particularly for users in the Recovery Phrase in Onboarding experiment, where it would cover up information about backing up your recovery phrase. 

This change waits for the user to visit the home screen before running the cloud messaging part of the firebase init saga.

### Test plan

manually and with unit tests on the new saga.

https://user-images.githubusercontent.com/7979215/230508789-c6be6a0f-ebe3-415f-a503-98674ac45c1d.mp4

### Related issues

fixes https://linear.app/valora/issue/ACT-715/bug-where-notification-pop-up-shows-up-on-protect-wallet-screen

### Backwards compatibility

na


src/firebase/saga.ts
```diff
@@ -36,16 +37,25 @@ const EXCHANGE_RATES = 'exchangeRates'
 const VALUE_CHANGE_HOOK = 'value'
 const FIREBASE_CONNECT_RETRIES = 3
 
-let firebaseAlreadyAuthorized = false
-export function* waitForFirebaseAuth() {
-  if (firebaseAlreadyAuthorized) {
+const actionSeen: Record<string, boolean> = {}
+
+export function* waitForAction(action: Actions | HomeActions) {
```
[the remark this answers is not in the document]
cajubelt: I worry that `takeOnce` over promises, since the in-memory cache of actions seen will be reset whenever the app restarts. I went about renaming and documenting this, but people don't always read the fine print when something is named a certain way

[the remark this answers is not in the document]
cajubelt: maybe `takeWithInMemoryCache` would be transparent enough? bit of a mouthful but I'm fine with it

src/firebase/saga.test.ts
```diff
@@ -0,0 +1,39 @@
...
+import firebase from '@react-native-firebase/app'
+
+jest.mock('@react-native-firebase/app', () => ({
+  app: () => ({ options: { databaseURL: 'https://test-firebase.valoraapp.com' } }),
+}))
+jest.mock('src/firebase/firebase')
+jest.mock('@celo/utils/lib/async')
+
+describe('firebase saga', () => {
+  it('waitForAction', async () => {
```
cajubelt, line 15: notably, this test doesn't verify that the saga does *not* proceed if the action is not dispatched. I tested this manually, but was not able to get it in a unit test here. Open to suggestions of how that could be done

src/firebase/saga.ts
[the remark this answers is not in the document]
jh2oman: good point. `takeOnceInMemory`? 

Also fine with a small comment to ward people off from just using without understanding.

src/firebase/saga.ts
```diff
@@ -67,6 +77,7 @@ function* initializeFirebase() {
 
         yield call(initializeAuth, firebase, address)
         yield put(firebaseAuthorized())
+        yield waitForVisitHome()
         yield call(initializeCloudMessaging, firebase, address)
```
jophish, line 81: Are there any side effects/other consequences of waiting to call `initializeCloudMessaging` until the home screen is seen? Just scanning, that function seems to do a lot. Is the only real consequence that push notifications won't be enabled until the home screen is reached?

src/firebase/saga.test.ts
jophish: I think `expectSaga` [has a default timeout](https://github.com/jfairbank/redux-saga-test-plan/blob/master/docs/integration-testing/timeout.md#timeout) of 250ms by default. I wonder if you could _not_ dispatch the action, and assert that the timeout is indeed triggered? Dunno if that's doable.

jh2oman: Joe's suggestion is what I thought of. I pulled the code and tried to write a test around it but it wasn't working. The test would just hit jest's 5000 ms timeout limit

jophish (approved):

src/firebase/saga.test.ts
cajubelt: i had the same issue

jh2oman (approved):

src/firebase/saga.ts
cajubelt: I had the same thought, and my initial reaction was well, this still happens every time a user visits home, which is most of the time when they open the app.

But I'm a bit uneasy about this, particularly as we move things around in the app-- if there wasn't a singular homepage anymore, for example, and some users spent lots of time away from home, this could cause problems in a very unexpected and hard to detect way. There's some advice in the docstring for `app.messaging().onTokenRefresh` (see [line 207 in firebase.ts](https://github.com/valora-inc/wallet/blob/0b4e518cc46dc98bbc19aca9bbd81222ca904981/src/firebase/firebase.ts#L189)) that says to run it on app open. 

I'm looking into a way to be more precise on when to wait for the homescreen, which might be better to do exactly [here](https://github.com/valora-inc/wallet/blob/0b4e518cc46dc98bbc19aca9bbd81222ca904981/src/firebase/firebase.ts#L154), since it only happens if the user hasn't already enabled messaging. If they have enabled messaging previously, all the other stuff in that initializeCloudMessaging function can still take place.

src/firebase/saga.test.ts
```diff
@@ -0,0 +1,37 @@
...
+import { initializeAuth, initializeCloudMessaging } from 'src/firebase/firebase'
+import firebase from '@react-native-firebase/app'
+
+jest.mock('@react-native-firebase/app', () => ({
+  app: () => ({ options: { databaseURL: 'https://test-firebase.valoraapp.com' } }),
+}))
+jest.mock('src/firebase/firebase')
+jest.mock('@celo/utils/lib/async')
+
+describe('firebase saga', () => {
```
cajubelt, line 14: these tests don't actually test anything new anymore; I wrote them when I was doing this a different way. But we have them now, so I left them in.

src/firebase/firebase.ts
```diff
@@ -150,6 +151,7 @@ export function* initializeCloudMessaging(app: ReactNativeFirebase.Module, addre
     yield call([app.messaging(), 'hasPermission'])
   Logger.info(TAG, 'Current messaging authorization status', authStatus.toString())
   if (authStatus === firebase.messaging.AuthorizationStatus.NOT_DETERMINED) {
+    yield take(HomeActions.VISIT_HOME)
```
cajubelt, line 154: if you reviewed an earlier version of this PR you may recall that I was doing some in-memory caching on home visits in the prev version. This is no longer necessary, since this particular block will rarely get entered more than once. If it does get evaluated more than once, it means the user quit the app and came back before accepting or rejecting notifications-- and in this case, an in-memory cache wouldn't do us any good.

src/firebase/firebase.test.ts
```diff
@@ -50,6 +50,7 @@ describe(initializeCloudMessaging, () => {
     let catchedError
 
     await expectSaga(initializeCloudMessaging, app, mockAccount)
+      .dispatch({ type: 'HOME/VISIT_HOME' })
```
cajubelt, line 53: these tests fail with a timeout without the home screen visit action. I wasn't able to turn that into an automated test for reasons discussed previously, but between this and the manual test result, I'm comfortable moving forward anyway.

src/firebase/saga.ts
cajubelt: k just updated the PR

src/firebase/saga.ts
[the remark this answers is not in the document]
cajubelt: i removed the changes here so hitting resolve on this one

src/firebase/firebase.ts
jophish: This only takes a single action, right? If this function throws after this action has been taken, the [caller](https://github.com/valora-inc/wallet/blob/d61919c182998847e8fe4c02bb614a6a53cc45d6/src/firebase/saga.ts#L63) will try calling this function again in a loop, at which point we'll block forever on this `take`, since the action will not have been dispatched again, right?

jophish (approved):

PR #3623, opened 2023-04-05
feat(home-actions): add new swap screen used by home actions

anthonydaoud:
### Description

Part 1 of ACT-704. This PR adds a new swap screen that the "Swap" home action will redirect to. The new screen removes the hamburger menu and replaces it with a back button that redirects to the home screen.

For the same reasons as described in #3610, it is necessary to add a new Swap screen rather than use existing one which is used in the Drawer navigation.

https://user-images.githubusercontent.com/20508929/230185109-349a8f0c-8d05-41e2-9982-3e3763cabcac.mp4

### Test plan

Added unit tests

### Related issues

- Partially resolves ACT-704

### Backwards compatibility

Yes

### DIFF
--- a/src/navigator/Screens.tsx
+++ b/src/navigator/Screens.tsx
@@ -85,6 +85,7 @@ export enum Screens {
   StoreWipeRecoveryScreen = 'StoreWipeRecoveryScreen',
   Support = 'Support',
   SupportContact = 'SupportContact',
+  SwapActionScreen = 'SwapActionScreen',
   SwapScreen = 'SwapScreen',
   SwapExecuteScreen = 'SwapExecuteScreen',
   SwapReviewScreen = 'SwapReviewScreen',
--- a/src/swap/SwapActionScreen.tsx
+++ b/src/swap/SwapActionScreen.tsx
@@ -1,0 +1,8 @@
+import React from 'react'
+import { SwapScreenSection } from 'src/swap/SwapScreen'
+
+function SwapActionScreen() {
+  return <SwapScreenSection showDrawerTopNav={false} />
+}
+
+export default SwapActionScreen
--- a/src/swap/SwapScreen.test.tsx
+++ b/src/swap/SwapScreen.test.tsx
@@ -9,7 +9,7 @@ import { TRANSACTION_FEES_LEARN_MORE } from 'src/brandingConfig'
 import { navigate } from 'src/navigator/NavigationService'
 import { Screens } from 'src/navigator/Screens'
 import { setSwapUserInput } from 'src/swap/slice'
-import SwapScreen from 'src/swap/SwapScreen'
+import SwapScreen, { SwapScreenSection } from 'src/swap/SwapScreen'
 import { Field } from 'src/swap/types'
 import networkConfig from 'src/web3/networkConfig'
 import { createMockStore } from 'test/utils'
@@ -27,7 +27,7 @@ jest.mock('react-native-localize', () => ({
 
 const now = Date.now()
 
-const renderScreen = ({ celoBalance = '10', cUSDBalance = '20.456' }) => {
+const renderScreen = ({ celoBalance = '10', cUSDBalance = '20.456', showDrawerTopNav = true }) => {
   const store = createMockStore({
     tokens: {
       tokenBalances: {
@@ -91,7 +91,11 @@ const renderScreen = ({ celoBalance = '10', cUSDBalance = '20.456' }) => {
 
   const tree = render(
     <Provider store={store}>
-      <SwapScreen />
+      {showDrawerTopNav ? (
+        <SwapScreen />
+      ) : (
+        <SwapScreenSection showDrawerTopNav={showDrawerTopNav} />
+      )}
     </Provider>
   )
   const [swapFromContainer, swapToContainer] = tree.getAllByTestId('SwapAmountInput')
@@ -116,10 +120,11 @@ describe('SwapScreen', () => {
   })
 
   it('should display the correct elements on load', () => {
-    const { getByText, swapFromContainer, swapToContainer } = renderScreen({})
+    const { getByText, swapFromContainer, swapToContainer, queryByTestId } = renderScreen({})
 
     expect(getByText('swapScreen.title')).toBeTruthy()
     expect(getByText('swapScreen.review')).toBeDisabled()
+    expect(queryByTestId('SwapScreen/DrawerBar')).toBeTruthy()
 
     expect(within(swapFromContainer).getByText('swapScreen.swapFrom')).toBeTruthy()
     expect(within(swapFromContainer).getByTestId('SwapAmountInput/MaxButton')).toBeTruthy()
@@ -594,5 +599,23 @@ describe('SwapScreen', () => {
         }),
       ])
     )
+  })
+
+  it('should be able to hide top drawer nav when parameter is set', () => {
+    const { getByText, swapFromContainer, swapToContainer, queryByTestId } = renderScreen({
+      showDrawerTopNav: false,
+    })
+
+    expect(queryByTestId('SwapScreen/DrawerBar')).toBeFalsy()
+    expect(getByText('swapScreen.review')).toBeDisabled()
+
+    expect(within(swapFromContainer).getByText('swapScreen.swapFrom')).toBeTruthy()
+    expect(within(swapFromContainer).getByTestId('SwapAmountInput/MaxButton')).toBeTruthy()
+    expect(within(swapFromContainer).getByTestId('SwapAmountInput/TokenSelect')).toBeTruthy()
+    expect(within(swapFromContainer).getByText('CELO')).toBeTruthy()
+
+    expect(within(swapToContainer).getByText('swapScreen.swapTo')).toBeTruthy()
+    expect(within(swapToContainer).getByTestId('SwapAmountInput/TokenSelect')).toBeTruthy()
+    expect(within(swapToContainer).getByText('swapScreen.swapToTokenSelection')).toBeTruthy()
   })
 })
--- a/src/swap/SwapScreen.tsx
+++ b/src/swap/SwapScreen.tsx
@@ -4,7 +4,7 @@ import React, { useEffect, useMemo, useState } from 'react'
 import { Trans, useTranslation } from 'react-i18next'
 import { Keyboard, StyleSheet, Text, View } from 'react-native'
 import { getNumberFormatSettings } from 'react-native-localize'
-import { SafeAreaView } from 'react-native-safe-area-context'
+import { Edge, SafeAreaView } from 'react-native-safe-area-context'
 import { useDispatch, useSelector } from 'react-redux'
 import { showError } from 'src/alert/actions'
 import { SwapEvents } from 'src/analytics/Events'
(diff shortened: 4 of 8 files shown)


src/swap/SwapScreen.tsx
```diff
@@ -251,25 +255,30 @@ export function SwapScreen() {
...
+              {exchangeRate && fromToken && toToken && (
+                <Text
+                  style={[headerStyles.headerSubTitle, fetchingSwapQuote ? styles.mutedHeader : {}]}
+                >
+                  {`1 ${fromToken.symbol} ≈ ${new BigNumber(exchangeRate).toFormat(
+                    5,
+                    BigNumber.ROUND_DOWN
+                  )} ${toToken.symbol}`}
+                </Text>
+              )}
```
satish-ravi, line 277: Should we retain this on the header of the new screen?

src/swap/SwapScreen.tsx
```diff
@@ -42,7 +42,11 @@ const DEFAULT_SWAP_AMOUNT: SwapAmount = {
 
 const { decimalSeparator } = getNumberFormatSettings()
 
-export function SwapScreen() {
+function SwapScreen() {
+  return <SwapScreenSection showDrawerTopNav={true} />
+}
+
+export function SwapScreenSection({ showDrawerTopNav }: { showDrawerTopNav?: boolean }) {
```
satish-ravi, line 49: nit: make this required? we pass it in both uses.

src/swap/SwapScreen.test.tsx
```diff
@@ -91,7 +91,11 @@ const renderScreen = ({ celoBalance = '10', cUSDBalance = '20.456' }) => {
 
   const tree = render(
     <Provider store={store}>
-      <SwapScreen />
+      {showDrawerTopNav ? (
+        <SwapScreen />
+      ) : (
+        <SwapScreenSection showDrawerTopNav={showDrawerTopNav} />
+      )}
```
satish-ravi, line 98: not sure I understand this. Wouldn't this mean that for the else part, `showDrawerTopNav` is always false?

src/swap/SwapActionScreen.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react'
+import { SwapScreenSection } from 'src/swap/SwapScreen'
+
+function SwapActionScreen() {
```
jh2oman, line 4: I'm not a fan of the `SwapActionScreen` name because it is referring to how a user gets to this screen (through the home actions) which doesn't have anything specifically do do with the component itself. In the future there may be other ways that a user can get to the swap screen and then this name will be confusing. Maybe naming this screen simply `SwapScreen` and naming the existing component `SwapDrawerScreen` to differentiate?

src/swap/SwapScreen.tsx
anthonydaoud: Updated the code so it keeps that subtitle

src/swap/SwapActionScreen.tsx
anthonydaoud: Good call, renamed the components

satish-ravi: Renaming existing screens have implications on analytics right? With this change, we'd start logging "SwapScreen" for the screen reached from the actions whereas previously this would refer to the screen from the drawer

src/swap/SwapScreen.test.tsx
anthonydaoud: Updated to remove the conditional and just render the `SwapScreenSection`

src/swap/SwapActionScreen.tsx
anthonydaoud: Okay, I kept the drawer screen's name as `SwapScreen` and the non-drawer is now called `SwapStartScreen`.

src/swap/SwapScreen.tsx
```diff
@@ -92,6 +106,33 @@ export function SwapScreen() {
...
+          {`1 ${fromToken.symbol} ≈ ${new BigNumber(exchangeRate).toFormat(
+            5,
+            BigNumber.ROUND_DOWN
+          )} ${toToken.symbol}`}
+        </Text>
+      ) : undefined,
+    [exchangeRate, fromToken, toToken, fetchSwapQuoteError]
+  )
+
+  useLayoutEffect(() => {
```
satish-ravi, line 122: Would be nice to have this in the SwapScreenWithBack component, but tricky since this relies on local state variables and we'd have to move them to redux? If so, probably not worth it.

src/swap/SwapScreen.test.tsx
```diff
@@ -8,12 +9,19 @@ import { ErrorMessages } from 'src/app/ErrorMessages'
 import { TRANSACTION_FEES_LEARN_MORE } from 'src/brandingConfig'
 import { navigate } from 'src/navigator/NavigationService'
 import { Screens } from 'src/navigator/Screens'
+import { StackParamList } from 'src/navigator/types'
 import { setSwapUserInput } from 'src/swap/slice'
-import SwapScreen from 'src/swap/SwapScreen'
+import { SwapScreenSection } from 'src/swap/SwapScreen'
```
satish-ravi, line 14: Can you also include a single test for SwapScreen component and make sure it renders the top drawer?

src/swap/SwapScreenWithBack.tsx
```diff
@@ -0,0 +1,13 @@
+import { NativeStackScreenProps } from '@react-navigation/native-stack'
```
satish-ravi, line 1: similar comment as above, add a single test for this validating the drawer is not rendered and nav is..

src/swap/SwapScreen.tsx
```diff
@@ -42,7 +44,19 @@ const DEFAULT_SWAP_AMOUNT: SwapAmount = {
...
+function SwapScreen() {
+  return <SwapScreenSection showDrawerTopNav={true} />
+}
+
+export function SwapScreenSection({
+  showDrawerTopNav,
+  props,
+}: {
+  showDrawerTopNav: boolean
+  props?: Props
```
satish-ravi, line 58: could this explicitly be a "navigation" object? We don't need all the props of the SwapScreenWithBack component right?

src/swap/SwapScreen.test.tsx
```diff
@@ -26,8 +34,14 @@ jest.mock('react-native-localize', () => ({
...
 const now = Date.now()
-
-const renderScreen = ({ celoBalance = '10', cUSDBalance = '20.456' }) => {
+const emptyProps = {} as NativeStackScreenProps<StackParamList, Screens.SwapScreenWithBack>
+
+const renderScreen = ({
+  celoBalance = '10',
+  cUSDBalance = '20.456',
+  showDrawerTopNav = true,
+  props = emptyProps,
```
satish-ravi, line 43: nit: you can just default this to undefined since its not a required prop right?

src/swap/SwapScreen.tsx
anthonydaoud: Yeah I'd agree that it'd be preferable not to have this here, but I think moving some local state to redux is required.

src/swap/SwapScreen.tsx
anthonydaoud: Good call, changed it to that

src/swap/SwapScreen.test.tsx
anthonydaoud: I can't because the type of `props` is inferred from the default value. If `props` is default `undefined`, I can't later call the `renderScreen` function with any other value for `props`

src/swap/SwapScreen.tsx
kathaypacific: we actually have a task to remove the exchange rate from the header, this is the new design for the swap screen https://www.figma.com/file/jRtYI8GiSvPAaClBnRLLpu/In-App-Swap?node-id=2528-24667&t=HLZ14cPGl2sHRofV-0 - if we're creating extra complexity just to preserve the exchange rate in the header, i think it's not worth it :/ assuming that the exchange rate will not be in the header, this implementation becomes quite simple because we just need one prop to determine if the left icon is a back or a drawer icon right? it removes the need for these 2 screen wrappers? if that's the case i think i can just have a pass at the swap screen restyling tomorrow, it should be quick to knock out (famous last words 😅)

anthonydaoud: I think we'd still need the screen wrappers because (1) drawer navigator components cannot accept props and (2) one of the cohorts in this experiment will not render the swap screen in the drawer, so the home actions cannot navigate to the swap screen unless we add it to the regular navigator. Joe explains this issue a bit more in #3610, and this PR mimics that approach

satish-ravi: yeah we still need the 2 screens, but sounds like we can drop the subtitle stuff and all the messy passing around of navigation props.

src/swap/SwapScreen.test.tsx
satish-ravi: you can define types explicitly instead of inferred types, but don't feel too strongly on this

src/swap/SwapScreen.tsx
kathaypacific: sorry i don't really understand the problem fully (even after reading Joe's comment in #3610). screens registered in the normal navigator can accept props, it's only the drawer navigator screens that cannot? i didn't know this, but i don't see how this is a blocking problem here (could just make all props optional, and pass the necessary props through the normal navigator?) we currently have one screen `Screens.BackupIntroduction` which maybe has the same use case, it appears conditionally in the drawer menu for some users and for others it is a normal screen that we navigate to from the settings screen. i might be missing some details about why we need a different approach for these home nav cards, so ignore this suggestion if that's the case :) tagging @jophish in case he has some more to add

kathaypacific: either way, you can move forward in this PR with the assumption that we no longer need to display the exchange rate in the header. that should simplify at least the props and how the header is created :)

anthonydaoud: Yeah I think that the reason the `BackupIntroduction` screen doesn't have this issue is because there is a `ts-expect-error` comment [added here](https://github.com/valora-inc/wallet/blob/main/src/navigator/DrawerNavigator.tsx#L303) in the `DrawerNavigator`. If we add that, it would remove the need for the two screens but not sure that we want to ignore that error.

Also, thanks for speeding through that exchange rate in header PR :)

satish-ravi: Thanks @kathaypacific , with regards to the 2 screens, the primary reason is the 2nd one mentioned by @anthonydaoud . We can workaround the ts error, but for one of the variants, we don't want to show the swap screen on the drawer menu, so it won't be registered with the drawer navigator and the action button on home screen won't be able to navigate to the screen. This is unlike the BackupIntroduction screen, which is also behind flags, but we don't want to navigate to it from any other place too if those flags are false.

src/navigator/Screens.tsx
```diff
@@ -86,6 +86,7 @@ export enum Screens {
   Support = 'Support',
   SupportContact = 'SupportContact',
   SwapScreen = 'SwapScreen',
+  SwapScreenWithBack = 'SwapScreenWithBack',
```
satish-ravi, line 89: Another option for the name is just `Swap`, we don't include the `Screen` suffix for many screens.

satish-ravi (approved):

PR #3624, opened 2023-04-05
fix(forced-backup): disable forced backup if experiment param says to

cajubelt:
### Description

for the backup in onboarding experiment, we want to turn off forced backup if the experiment param says to do so

### Test plan

automated tests

### Related issues

should have been done in https://linear.app/valora/issue/ACT-643/set-up-statsig-experiment-determine-bucketing-in-wallet-disable-forced

### Backwards compatibility

na

### DIFF
--- a/src/backup/selectors.ts
+++ b/src/backup/selectors.ts
@@ -3,7 +3,7 @@ import { Screens } from 'src/navigator/Screens'
 import { RootState } from 'src/redux/reducers'
 import { ONE_DAY_IN_MILLIS } from 'src/utils/time'
 
-export const shouldForceBackupSelector = (state: RootState) => {
+export const pastForcedBackupDeadlineSelector = (state: RootState) => {
   if (state.account.backupCompleted) {
     return false
   }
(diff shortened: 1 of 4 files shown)


src/backup/selectors.ts
```diff
@@ -3,7 +3,7 @@ import { Screens } from 'src/navigator/Screens'
 import { RootState } from 'src/redux/reducers'
 import { ONE_DAY_IN_MILLIS } from 'src/utils/time'
 
-export const shouldForceBackupSelector = (state: RootState) => {
+export const pastForcedBackupDeadlineSelector = (state: RootState) => {
```
cajubelt, line 6: renamed to make it clearer that this is only one thing that should be considered to determine whether we should show the force backup screen

cajubelt: (the other thing being the experiment parameter)

satish-ravi (approved):

satish-ravi (commented):
I think this would be good to test manually by reducing the 24h deadline on a local build and test behavior for all variants.

cajubelt:
i tested both values of the experiment parameter per Satish's suggestion and verified that we get forced backup only in the control

PR #3625, opened 2023-04-06
fix: dismiss keyboard on bottom sheet open

MuckT:
### Description

Noticed when working on #3599. Currently if the keyboard is visible the when `openSheet` is called the bottom sheet will display behind the keyboard. If there is no keyboard visible on the screen, then the `Keyboard.dismiss()` method will not have any effect.

#### Added to
- `openSheet` from `src/dappsExplorer/useDappInfoBottomSheet.tsx`
- `src/components/BottomSheet.tsx`

#### Before
https://user-images.githubusercontent.com/26950305/230246866-57bd2246-cb55-4d79-9672-6712e071fb9b.mov

#### After
https://user-images.githubusercontent.com/26950305/230247068-85bc3ada-f6dc-4302-9537-e0e30c3c2642.mov

### Test plan

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/src/components/BottomSheet.tsx
+++ b/src/components/BottomSheet.tsx
@@ -1,6 +1,7 @@
-import React, { useState } from 'react'
+import React, { useEffect, useState } from 'react'
 import {
   Dimensions,
+  Keyboard,
   LayoutChangeEvent,
   StyleSheet,
   TouchableWithoutFeedback,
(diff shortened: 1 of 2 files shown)


jh2oman (approved):

src/components/BottomSheet.tsx
```diff
@@ -38,6 +39,10 @@ function BottomSheet({
   const [pickerHeight, setPickerHeight] = useState(0)
   const safeAreaInsets = useSafeAreaInsets()
 
+  useEffect(() => {
+    if (isVisible) Keyboard.dismiss()
```
dievazqu, line 43: nit: I'm a bit concern about this being triggered when we don't want to...
What would happen if there is a component with `autoFocus` in the BottomSheet? So we will expect to have the keyboard ready.
Maybe we should leave this logic to the BottomSheet caller?
Anyways, I think this is a fine.

dievazqu (approved):

src/components/BottomSheet.tsx
MuckT: We should not be putting components with text inputs within bottom sheets in general. I don't see in a place in the code where we are calling `autoFocus={true}` on a component in a bottom sheet currently so I think this is fine for now too.

PR #3626, opened 2023-04-06
feat(bottom sheet): Adding Sticky header to the bottom sheet

dievazqu:
### Description

Adding Sticky header to the bottom sheet. Extracted logic from #3613


src/send/Send.tsx
```diff
@@ -246,7 +251,8 @@ function Send({ route }: Props) {
         origin={TokenPickerOrigin.Send}
         onTokenSelected={onTokenSelected}
         onClose={closeCurrencyPicker}
-        tokens={isOutgoingPaymentRequest ? stableTokens : tokensWithBalance}
+        tokens={sortedTokens}
```
jeanregisser, line 254: Looks like this tokens change should be done separately, right?

jeanregisser (commented):
Thanks for splitting.

src/send/Send.tsx
dievazqu: What do you mean? This is not changing anything actually, we were sorting inside the TokenBottomSheet, now each component can decide how to sort that list...

jeanregisser: Oh sorry I see! All good, I initially thought it was an unrelated change.

jeanregisser (approved):

PR #3627, opened 2023-04-06
fix(onboarding): land on the correct screen if bailed on recovery phrase

satish-ravi:
### Description

Currently if the user bails on recovery phrase screen in onboarding, the next time the app is launched, the user lands on the PN verification screen. This updates it so that for the experiment variant, the user lands on the recovery phrase screen.

### Test plan

Manual and unit

Control:

https://user-images.githubusercontent.com/5062591/230488403-ea89bb01-9e4a-4e87-9dfa-0b6fae786730.mp4

V1 - backup on onboarding

https://user-images.githubusercontent.com/5062591/230488682-7d4721f9-81c9-46f1-8a01-2152d1cce103.mp4

V2 - fake door

https://user-images.githubusercontent.com/5062591/230488716-0f2c3317-6b03-4bbe-8292-9e5689df15c8.mp4

### Related issues

- Fixes ACT-700

### Backwards compatibility

Yes

### DIFF
--- a/src/navigator/Navigator.tsx
+++ b/src/navigator/Navigator.tsx
@@ -20,7 +20,6 @@ import AccounSetupFailureScreen from 'src/account/AccountSetupFailureScreen'
 import GoldEducation from 'src/account/GoldEducation'
 import Licenses from 'src/account/Licenses'
 import Profile from 'src/account/Profile'
-import { PincodeType } from 'src/account/reducer'
 import StoreWipeRecoveryScreen from 'src/account/StoreWipeRecoveryScreen'
 import SupportContact from 'src/account/SupportContact'
 import { CeloExchangeEvents } from 'src/analytics/Events'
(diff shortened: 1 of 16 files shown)


src/navigator/Navigator.tsx
```diff
@@ -634,6 +639,12 @@ export function MainStackScreen() {
       initialRoute = Screens.Welcome
     } else if (!account) {
       initialRoute = choseToRestoreAccount ? Screens.ImportWallet : Screens.Welcome
+    } else if (
```
jh2oman, line 642: At first glance it feels like this if/else chain is ripe for refactor, for example doing a `lastOnboardingScreenSeen` variable in redux, integrating it into steps.ts and deciding which screen to load after a bailout based on that. 

However, that would mean a few real product changes, like taking the user directly to name screen, pin screen, biometrics screen which we may or may not want to do. 

I'm sure you thought about this, but I'm okay with your current implementation b/c all the exceptions to the rules right now I think would defeat the point of a refactor.

satish-ravi: yeah, we should eventually do this, but outside the scope of the current ticket as we discussed in the sync. Will create a follow up ticket. One idea to retain current behavior is to have some onboarding milestones, that way we don't land on certain screens we currently don't.

cajubelt (approved):

PR #3628, opened 2023-04-07
test(dapp-search): don't use real timers in jest

jeanregisser:
Patch for https://github.com/valora-inc/wallet/pull/3602

It's best to avoid real timers in jest whenever possible, so tests don't run for longer than needed.


src/dappsExplorer/search/DAppsExplorerScreenSearch.test.tsx
```diff
@@ -347,8 +353,7 @@ describe(DAppsExplorerScreenSearch, () => {
       expect(within(allDappsSection).getByText(dappsList[0].name)).toBeTruthy()
     })
 
-    it('triggers events when searching', async () => {
-      jest.useRealTimers()
+    it('triggers events when searching', () => {
```
jeanregisser, line 356: Strangely, I couldn't make it work leaving `async`.
It's not needed here so I removed, but still could be confusing in the future if we add `async` again.

MuckT (approved):
🎉

PR #3629, opened 2023-04-08
feat: calculate dapp score and sort for search

MuckT:
### Description

An overhaul of #3598. This PR uses both exact match and fuzz match to determine the search score, for former having a having a heavier weight than that later. This implementation gives us several advantages to an exact match search only.

It uses Levenshtein distance for fuzzy match which is the the minimum number of single-character edits (insertions, deletions or substitutions) required to change one word into the other. The library added is [fast-levenshtein](https://www.npmjs.com/package/fast-levenshtein) which as of this writing has as 30 million weekly downloads: we probably won't have to fix this library ourselves if it breaks.

Some advantages of Levenshtien:
- Allows for a match if a search term includes the string e.g. Searching "Swaps" would fuzzy match with "Swap" using exact match this would not match, same with "cambiar" and "cambia".
- Allow for some misspellings in a language agonistic way. The allowed variance of a word from in proportional to the word length. "Swxp" would match "Swap", but "I" would not match "e".

Some advantages of using exact match:
- Works well when the searchTerm is a small subset of a larger string, where the levenshtien distance would be rather large in this case e.g. "Pool" & "PoolTogether"
- Early returns if there is an exact match to avoid running unneeded calculations

#### Levenshtein Distance Example
```
const levenshteinThreshold = .7
searchTerm = "swaps"
haystackTerm = "swap"
maxLength = Math.Max(searchTerm.length, haystackTerm.length) // 5
const weightedDistance = get(searchTerm, haystackTerm, { useCollator: true }) / maxLength // .20
return 1 - weightedDistance >= .7 // True in this case
```

This PR includes a `searchDappList` method for sorting the dapp list based on a search the main advantage of this is we can reduce duplicate code and calculate the search score for each dapp only once per search term. This is more performant than the previous sort method which would calculate the dapp score each comparison of the sort.

#### Some notes about other implementations tried

- BM25 worked decent but was slower and more complex. Had some issues with smaller strings.
- [flexsearch](https://www.npmjs.com/package/flexsearch#contextual) using - contextual indexing: worked well, but required additional stemmers, which are language specific, to fuzzy match well. Documentation was a bit unclear and it's unclear if this library is actively maintained.

### Test plan

Unit tests added for `scoreDapp` method and new selectors in:
- `src/dappsExplorer/searchDappList.test.tsx`
- `src/dapps/selectors.test.ts`

### Related issues

- Fixes: RET-580

### Backwards compatibility

Yes


src/dappsExplorer/searchDappList.tsx
```diff
@@ -0,0 +1,111 @@
...
+    } else {
+      for (const token of descriptionTokens) {
+        const similarity = getSimilarity(token, word)
+        if (similarity >= LEVENSHTEIN_THRESHOLD) {
+          scoreMap.set(word, (scoreMap.get(word) ?? 0) + ScoreValues.Low)
+        }
+      }
+    }
+
+    // Category scoring
```
dievazqu, line 68: I'm still not sure if we want to add this to the scoring.

src/dappsExplorer/searchDappList.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+  })
+
+  it.each`
+    searchTerm
+    ${'Ubeswap'}
+    ${'Cambia'}
+  `('returns an array of dapps that match the search term', ({ searchTerm }) => {
+    expect(searchDappList([dappV2], searchTerm)).toEqual([dappV2])
+  })
+})
```
dievazqu, line 58: nit: Can you add a test case with multiple dapps and check against multiple search terms?
So we can check is sorting and filtering correctly.

src/dappsExplorer/searchDappList.tsx
MuckT: I am worried without this we are hampering search and making search and filter a bit more difficult.

src/dappsExplorer/searchDappList.test.tsx
MuckT: Added in 9639b79fcc7656fe19f9581897316b2afebb7dff!

src/dappsExplorer/searchDappList.tsx
MuckT: Checked with @reedeereed and they seemed fine with this behavior.

PR #3630, opened 2023-04-10
feat: search filter sans analytics

MuckT:
## Description

Implements the search and filter variant of the dapps list. This looks like a bit PR, but most of the logic is simply copied over from the search and filter variants of the dapps list. Below I'll call specific behavior unique to this PR.

### New Components
- `src/dappsExplorer/search/NoResults.tsx` has been added which is a combination of the components of the same name in search and in filter. It has conditional logic to display the approriate no results section based on the presence of absence of selected filters and search term.

-  `src/dappsExplorer/search/FavoriteDappsSection.tsx` has been added which is also a combination of the components of the same name in search and filter. The conditional logic displays No Results, favorites or no favorites added sections.
Below is the pattern for how we're choosing what section to display.

#### Applying filters and sorts

`src/dappsExplorer/searchFilter/FavoriteDappsSection.tsx`
```
  const favoriteDappsWithCategoryNames = useSelector(favoriteDappsWithCategoryNamesSelector)
  const favoriteResultsFiltered = favoriteDappsWithCategoryNames.filter(
    (dapp) => dapp.categories.includes(filterId) || filterId === 'all'
  )
  const favoriteResults =
    searchTerm === ''
      ? favoriteResultsFiltered
      : (searchDappList(favoriteResultsFiltered, searchTerm) as DappV2[])
```

You might notice this is similar to the logic from the `parseResultsIntoAll` function in `src/dappsExplorer/searchFilter/DAppsExplorerScreenSearchFilter.tsx`

```
function parseResultsIntoAll(
  nonFavoriteDapps: DappV2WithCategoryNames[],
  searchTerm: string,
  filterId: string
) {
  // Dapps in the all section are all the non favorite dapps that match the filter
  const dappsMatchingFilter =
    filterId === 'all'
      ? nonFavoriteDapps
      : nonFavoriteDapps.filter(
          (dapp: DappV2) => dapp.categories && dapp.categories.includes(filterId)
        )
  // If there are no dapps matching the filter return an empty array
  if (dappsMatchingFilter.length === 0) return []
  // If there is no search term return the dapps matching the category filter
  if (searchTerm === '') {
    return [
      {
        data: dappsMatchingFilter,
        category: 'all',
      },
    ]
  } else {
    // Score and sort the dapps matching the category filter
    const results = searchDappList(dappsMatchingFilter, searchTerm) as DappV2WithCategoryNames[]
    // If there are no dapps matching search term return an empty array
    if (results.length === 0) return []
    return [
      {
        data: results,
        category: 'all',
      },
    ]
  }
}
```

#### Determining when to show the no results section
`src/dappsExplorer/searchFilter/FavoriteDappsSection.tsx`
```
useEffect(() => {
  setFavoriteResultsEmpty(favoriteResults.length <= 0)
}, [favoriteResults, searchTerm, filterId])
```

`src/dappsExplorer/searchFilter/DAppsExplorerScreenSearchFilter.tsx`
```
const allSectionResults: SectionData[] = useMemo(() => {
  const allResultsParsed = parseResultsIntoAll(
    nonFavoriteDappsWithCategoryNames,
    searchTerm,
    selectedFilter
  )
  setAllResultEmpty(allResultsParsed.length === 0)
  return allResultsParsed
}, [nonFavoriteDappsWithCategoryNames, searchTerm, selectedFilter])
```

#### Demo Video iOS
https://user-images.githubusercontent.com/26950305/231023224-30a52e95-1f57-4ebf-8988-a6b3a6fbb461.mov

### Test plan

Tested locally on iOS

### Related issues

- Fixes: RET-705

### Backwards compatibility

Yes


MuckT:
Looking into an issue on this branch the `<NoResults />` component isn't updating the search term if no matches are found. Likely a missing a dependency in a memo somewhere 👀


https://user-images.githubusercontent.com/26950305/231013933-d9bc5930-21bc-4848-9c40-8d051a08de3d.mov

src/dappsExplorer/searchFilter/FavoriteDappsSection.tsx
```diff
@@ -0,0 +1,113 @@
...
+    searchTerm === ''
+      ? favoriteResultsFiltered
+      : (searchDappList(favoriteResultsFiltered, searchTerm) as DappV2[])
+
+  useEffect(() => {
+    if (favoriteResults.length > 0) {
+      onShowSearchResult(false)
+    } else {
+      onShowSearchResult(true)
+    }
```
kathaypacific, line 47: nit: this could be `onShowSearchResult(favoriteResults.length <= 0)`

src/dappsExplorer/searchFilter/FavoriteDappsSection.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const favoriteResults =
+    searchTerm === ''
+      ? favoriteResultsFiltered
+      : (searchDappList(favoriteResultsFiltered, searchTerm) as DappV2[])
+
+  useEffect(() => {
+    if (favoriteResults.length > 0) {
+      onShowSearchResult(false)
+    } else {
+      onShowSearchResult(true)
```
kathaypacific, line 46: i was confused about what `onShowSearchResult true` means...maybe this should just be `setFavoriteResultsEmpty`?

src/dappsExplorer/searchFilter/DAppsExplorerScreenSearchFilter.test.tsx
```diff
@@ -0,0 +1,640 @@
+import { fireEvent, render, within } from '@testing-library/react-native'
```
kathaypacific, line 1: is this file all copy paste, or are there key changes that should be reviewed?

src/dappsExplorer/searchFilter/DAppsExplorerScreenSearchFilter.tsx
```diff
@@ -0,0 +1,387 @@
...
+  const onScroll = Animated.event([{ nativeEvent: { contentOffset: { y: scrollPosition } } }])
+  const dispatch = useDispatch()
+  const loading = useSelector(dappsListLoadingSelector)
+  const error = useSelector(dappsListErrorSelector)
+  const categories = useSelector(dappsCategoriesAlphabeticalSelector)
+  const dappsMinimalDisclaimerEnabled = useSelector(dappsMinimalDisclaimerEnabledSelector)
+  const dappList = useSelector(dappsListSelector)
+  const dappListWithCategoryNames = useSelector(dappListWithCategoryNamesSelector)
+  const language = useSelector(currentLanguageSelector)
+  const favoriteDappsById = useSelector(favoriteDappIdsSelector)
```
kathaypacific, line 70: can we simplify this? it seems like `favoriteDappsById` is only ever used with `dappListWithCategoryNames`, to filter out the favourite dapps from the dapps list. can we just have a selector that does this, so that we don't need both? the benefit would be that we don't make this computation as often.

src/dappsExplorer/searchFilter/DAppsExplorerScreenSearchFilter.test.tsx
MuckT: Only this section is new: https://github.com/valora-inc/wallet/blob/193a9bc20bd2096c4b64d54324e91db001d73ae5/src/dappsExplorer/searchFilter/DAppsExplorerScreenSearchFilter.test.tsx#L564-L640.

src/dappsExplorer/searchFilter/DAppsExplorerScreenSearchFilter.tsx
MuckT: I made the changes in ee9b0761a714c730d5421ef47e1f1caa702a5aab. Replaced 3 selector imports with one!

src/dappsExplorer/searchFilter/FavoriteDappsSection.tsx
MuckT: Fixed in dd6b52699420a45ff16c1d098a54499930f379ea!

src/dappsExplorer/searchFilter/FavoriteDappsSection.tsx
MuckT: Fixed in dd6b52699420a45ff16c1d098a54499930f379ea! This is also used in the search only variant, but I think a follow up pr aligning the behavior in those components would be better than adding more in this PR 😅

src/dappsExplorer/searchFilter/DAppsExplorerScreenSearchFilter.tsx
```diff
@@ -0,0 +1,387 @@
...
+                  {/* If no matching dapps in all section and favorite section display favoriteDappsAndAll*/}
+                  {allResultEmpty && favoriteResultsEmpty ? (
+                    <Text style={styles.sectionTitle}>
+                      {t('dappsScreen.favoriteDappsAndAll').toLocaleUpperCase(language ?? 'en-US')}
+                    </Text>
+                  ) : (
+                    <Text style={styles.sectionTitle}>
+                      {t('dappsScreen.favoriteDapps').toLocaleUpperCase(language ?? 'en-US')}
+                    </Text>
+                  )}
```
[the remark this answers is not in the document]
MuckT: Fixed in dd6b52699420a45ff16c1d098a54499930f379ea! I think the second option, calling the translations in ternary, is a bit easier to read personally.

src/dappsExplorer/searchFilter/DAppsExplorerScreenSearchFilter.tsx
```diff
@@ -0,0 +1,387 @@
...
+  const horizontalScrollView = useRef<ScrollView>(null)
+
+  const onScroll = Animated.event([{ nativeEvent: { contentOffset: { y: scrollPosition } } }])
+  const dispatch = useDispatch()
+  const loading = useSelector(dappsListLoadingSelector)
+  const error = useSelector(dappsListErrorSelector)
+  const categories = useSelector(dappsCategoriesAlphabeticalSelector)
+  const dappsMinimalDisclaimerEnabled = useSelector(dappsMinimalDisclaimerEnabledSelector)
+  const dappList = useSelector(dappsListSelector)
+  const dappListWithCategoryNames = useSelector(dappListWithCategoryNamesSelector)
```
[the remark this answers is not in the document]
MuckT: ee9b0761a714c730d5421ef47e1f1caa702a5aab addresses this too.

src/dappsExplorer/searchFilter/NoResults.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+  const onPress = () => {
+    removeFilter()
+  }
+
+  // No Match Search & Filter
+  // If filterId is not 'all' and searchTerm is not empty - No Match Filter & Search
+  // Else if filterId is not 'all' and searchTerm is empty - No Match Filter
+  // Else - No Match Search
+  if (filterId !== 'all' && searchTerm !== '') {
```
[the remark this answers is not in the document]
MuckT: Fixed in 8c163a81a48976bb1467c44d842441eb52c2ced7!

src/dappsExplorer/searchFilter/DAppsExplorerScreenSearchFilter.test.tsx
```diff
@@ -0,0 +1,640 @@
...
+      expect(getByTestId('FavoriteDappsSection/NoResults')).toBeTruthy()
+
+      // All Section should show only 'dapp 1'
+      const allDappsSection = getByTestId('DAppsExplorerScreenSearchFilter/DappsList')
+      expect(within(allDappsSection).getByText(dappsList[0].name)).toBeTruthy()
+      expect(within(allDappsSection).getByText(dappsList[0].description)).toBeTruthy()
+      expect(within(allDappsSection).queryByText(dappsList[1].name)).toBeFalsy()
+      expect(within(allDappsSection).queryByText(dappsList[1].description)).toBeFalsy()
+    })
+  })
```
kathaypacific, line 639: do you think there's value to add a test case for the `favoriteDappsAndAll` scenario where there is search and filter applied and no results in favourites or all?

MuckT:
@kathaypacific I'll add a test for the favorites being empty and remove the selector abstraction: I was on the fence about it too. For the styling of the remove filter button I'll make a follow up PR since I'll have to adjust the one in filters too.

PR #3631, opened 2023-04-10
chore(deps): upgrade segment (along with segment plugins adjust and clevertap)  and firebase

cajubelt:
### Description

upgrade `@react-native-firebase/*` deps from v11 to v17
upgrade `@segment/*` deps from v1 to v2

### Audit Advisory

This audit advisory should be able to be ignored by us because its part of the `@expo/config-plugins` package which we definitely don't make use of. (Expo is a platform to develop react native that we don't use).

### Test plan

I have manually tested the following: 
* identify, screen, track events all posted to segment with the data we expect
* statsig client still initialized with same anonymousId as Segment uses
* values are still read correctly from firebase realtime DB
* feature flags are loaded correctly from firebase remote config
* notifications still received from firebase messaging
* deep links still work from firebase dynamic links

I recommend we re-test notifications and deeplinks in bugapalooza. 

### Related issues

https://linear.app/valora/issue/ACT-727/upgrade-segment-firebase-deps-in-wallet-repo

### Backwards compatibility

Yes


patches/@segment+sovran-react-native+1.0.1.patch
```diff
@@ -0,0 +1,10 @@
+diff --git a/node_modules/@segment/sovran-react-native/lib/typescript/src/index.d.ts b/node_modules/@segment/sovran-react-native/lib/typescript/src/index.d.ts
+index 961c442..aba4c3a 100644
+--- a/node_modules/@segment/sovran-react-native/lib/typescript/src/index.d.ts
++++ b/node_modules/@segment/sovran-react-native/lib/typescript/src/index.d.ts
+@@ -1,3 +1,4 @@
+-export { createStore, type Store, type Notify, type Unsubscribe, } from './store';
```
jh2oman, line 6: TODO: This file can be removed when we upgrade to Typescript 4.4.X or higher. 
It was causing an error similar to this with our current typescript version https://github.com/segmentio/analytics-react-native/issues/607

firebase.json
```diff
@@ -1,5 +0,0 @@
-{
-  "react-native": {
-    "messaging_ios_auto_register_for_remote_messages": false
```
jh2oman, line 3: Leftover from a bug in 6.X.X . Registering is done automatically by the firebase client

src/account/saga.ts
```diff
@@ -206,7 +204,11 @@ export function* handleUpdateAccountRegistration() {
 
   let fcmToken
   try {
-    fcmToken = yield call([firebase.app().messaging(), 'getToken'])
+    const isEmulator = yield call([DeviceInfo, 'isEmulator'])
```
jophish, line 207: What used to happen here when this was called inside of an emulator? We just threw an error and it was caught below?

src/analytics/InjectTraits.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Plugin that injects the user traits to every event
+ *
+ * Added to migrate to Segment v2 without breaking existing dashboards that rely on context_traits_*. All this data is
+ *  available as "super" properties, so we can eventually get rid of it.
+ *
+ * TODO eventually: remove dependencies and delete this plugin, see https://linear.app/valora/issue/ENG-60/remove-dependencies-on-context-traits-
+ */
+export class InjectTraits extends PlatformPlugin {
```
jophish, line 11: How was this data previously being logged to segment? Was this happening automatically under the hood/is the removal of this behavior from segment v1 to v2 documented anywhere?

src/account/saga.ts
jh2oman: It failed silently before by just returning undefined. the latest 17v made it throw an error instead. https://github.com/invertase/react-native-firebase/blob/main/CHANGELOG.md#-breaking-changes

src/analytics/InjectTraits.ts
jh2oman: It was done automatically. Here is the documentation for it https://github.com/segmentio/analytics-react-native/blob/c3e41ead9de261bd3232305444eccd824294acc7/MIGRATION_GUIDE.md#context-traits

src/analytics/InjectTraits.ts
```diff
@@ -0,0 +1,28 @@
...
+  type = PluginType.before
+
+  execute(event: SegmentEvent) {
```
cajubelt, line 14: it might be worth putting the implementation of this in a try/catch and just logging an error if it doesn't work. probably better to get some data than no data and potentially block the user

src/analytics/ValoraAnalytics.ts
```diff
@@ -86,33 +87,32 @@ async function getDeviceInfo(): Promise<DeviceInfoType> {
...
+  private segmentClient: SegmentClient | undefined
 
   async init() {
     let uniqueID
     try {
       if (!SEGMENT_API_KEY) {
         throw Error('API Key not present, likely due to environment. Skipping enabling')
       }
-      await Analytics.setup(SEGMENT_API_KEY, SEGMENT_OPTIONS)
+      this.segmentClient = createClient({
```
cajubelt, line 104: @kathaypacific fyi this upgrade may impact your work to initialize Segment earlier..

src/firebase/firebase.ts
```diff
@@ -184,11 +185,12 @@ export function* initializeCloudMessaging(app: ReactNativeFirebase.Module, addre
       yield put(pushNotificationsPermissionChanged(pushNotificationsEnabled))
     }
   }
-
-  // `registerDeviceForRemoteMessages` must be called before calling `getToken`
-  // Note: `registerDeviceForRemoteMessages` is really only required for iOS and is a no-op on Android
-  yield call([app.messaging(), 'registerDeviceForRemoteMessages'])
```
cajubelt, line 190: this isn't needed anymore?

yarn.lock
```diff
@@ -5375,6 +5483,33 @@
...
+  dependencies:
+    "@typescript-eslint/types" "5.58.0"
+    "@typescript-eslint/visitor-keys" "5.58.0"
+    debug "^4.3.4"
+    globby "^11.1.0"
+    is-glob "^4.0.3"
+    semver "^7.3.7"
+    tsutils "^3.21.0"
+
+"@typescript-eslint/utils@5.58.0":
```
cajubelt, line 5499:
```suggestion
"@typescript-eslint/utils@5.58.0", "@typescript-eslint/utils@^5.10.0":
```
we *could* do this (and delete the one below), though it's not that important

yarn.lock
```diff
@@ -10414,6 +10551,18 @@ glob@7.1.3:
     once "^1.3.0"
     path-is-absolute "^1.0.0"
 
+glob@7.1.6, glob@^7.0.3, glob@^7.1.1, glob@^7.1.2, glob@^7.1.3, glob@^7.1.4, glob@^7.1.6:
```
cajubelt, line 10554: combine with the one below?

cajubelt (commented):
excited!

src/firebase/firebase.ts
jh2oman: Not anymore now that we've turned off [messaging_ios_auto_register_for_remote_messages](https://github.com/valora-inc/wallet/pull/3631/files#r1173119135)

docs here https://rnfirebase.io/app/json-config#messaging_ios_auto_register_for_remote_messages

yarn-audit-known-issues
```diff
@@ -1 +1,2 @@
...
...ble_versions":"<0.5.0","module_name":"xml2js","severity":"high","github_advisory_id":"GHSA-776f-qx25-q3cc","cves":["CVE-2023-0842"],"access":"public","patched_versions":">=0.5.0","cvss":{"score":7.3,"vectorString":"CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:L/I:L/A:L"},"updated":"2023-04-10T19:01:30.000Z","recommendation":"Upgrade to version 0.5.0 or later","cwe":["CWE-1321"],"found_by":null,"deleted":null,"id":1091653,"references":"- https://nvd.nist.gov/vuln/detail/CVE-2023-0842\n- https://fluidattacks.com/advisories/myers/\n- https://github.com/Leonidas-from-XIV/node-xml2js/issues/663\n- https://github.com/Leonidas-from-XIV/node-xml2js/pull/603/commits/581b19a62d88f8a3c068b5a45f4542c2d6a495a5\n- https://github.com/advisories/GHSA-776f-qx25-q3cc","created":"2023-04-05T21:30:24.000Z","reported_by":null,"title":"xml2js is vulnerable to prototype pollution ","npm_advisory_id":null,"overview":"xml2js versions before 0.5.0 allows an external attacker to edit or add new properties to an object. This is possible because the application does not properly validate incoming JSON keys, thus allowing the `__proto__` property to be edited.","url":"https://github.com/advisories/GHSA-776f-qx25-q3cc"}}}
```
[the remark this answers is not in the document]
jh2oman: This issue should be able to be ignored by us because its part of the `@expo/config-plugins` package which we definitely don't make use of. (Expo is a platform to develop react native that we don't use).

yarn.lock
cajubelt: never mind, we need at least 2 versions of glob because we have `glob@7.1.6` and `glob@7.2.0` which do not accept ranges

yarn.lock
cajubelt: actually never mind, since ^5.10.0 should be able to go above 5.58.0 if there are newer versions, whereas 5.58.0 always needs 5.58.0

yarn.lock
```diff
@@ -8302,7 +8460,7 @@ debug@^3.2.7:
   dependencies:
     ms "^2.1.1"
 
-debug@^4.0.1, debug@^4.1.0, debug@^4.1.1:
+debug@^4.0.1, debug@^4.1.0:
```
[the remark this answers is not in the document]
cajubelt: never mind, the ranged versions should be able to exceed 4.3.4, but we also have a fixed version `debug@4.3.4`

yarn-audit-known-issues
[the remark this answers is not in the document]
jh2oman: added that to the description too

src/analytics/InjectTraits.ts
jh2oman: done

src/analytics/InjectTraits.ts
```diff
@@ -0,0 +1,33 @@
...
+              ...event.context,
+              traits: {
+                ...event.context?.traits,
+                ...this.analytics.userInfo.get().traits,
+              },
+            },
+          }
+        : event
+    } catch (error) {
+      Logger.error('SegmentPlugin InjectTraits', 'Error injecting traits', error)
```
cajubelt, line 30:
```suggestion
      Logger.error('SegmentPlugin InjectTraits', 'Error injecting traits', error)
      return event
```

cajubelt (commented):
thanks for deduping the deps!

src/analytics/ValoraAnalytics.test.ts
```diff
@@ -152,6 +143,21 @@ const defaultProperties = {
...
+  const mockPromiseFactory = () =>
+    jest.fn().mockReturnValue({
+      then: jest.fn(),
+      catch: jest.fn(),
+    })
+  const mockSegmentClient = {
+    identify: mockPromiseFactory(),
+    track: mockPromiseFactory(),
+    screen: mockPromiseFactory(),
+    flush: mockPromiseFactory(),
```
satish-ravi, line 155: nit: You can just do `jest.fn().mockResolvedValue(undefined)`  instead of the promise factory right?

satish-ravi (approved):

src/firebase/firebase.ts
jeanregisser: Removing this + `messaging_ios_auto_register_for_remote_messages` will cause iOS to show the push permission popup directly on app launch, which is a bit too intrusive I think.

Are we sure we want this?

patches/@segment+sovran-react-native+1.0.1.patch
jeanregisser: Maybe we should add this as a comment in the patch so we'll know more easily?

jeanregisser (commented):
Awesome work! ❤️ 

Maybe we should update the PR title to also reflect that Adjust and CleverTap were upgraded?

package.json
```diff
@@ -97,10 +98,11 @@
...
     "@reduxjs/toolkit": "^1.9.2",
-    "@segment/analytics-react-native": "^1.5.3",
-    "@segment/analytics-react-native-adjust": "^1.5.3",
-    "@segment/analytics-react-native-clevertap": "^1.5.3",
-    "@segment/analytics-react-native-firebase": "^1.5.3",
+    "@segment/analytics-react-native": "^2.13.4",
+    "@segment/analytics-react-native-plugin-adjust": "^0.6.1",
+    "@segment/analytics-react-native-plugin-clevertap": "^1.0.0",
+    "@segment/analytics-react-native-plugin-firebase": "^0.3.7",
+    "@segment/sovran-react-native": "1.0.1",
```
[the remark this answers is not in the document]
jh2oman: yes because it has a patch/ file

PR #3635, opened 2023-04-11
fix: display reconnect info on supercharge screen when decentrally verified

kathaypacific:
### Description

As I was doing the last checks for the supercharge plan, I noticed that I used the wrong selector to get the decentrally verified state :( not a show stopper but thought i'd fix it anyway in case users run into this in the next release.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y


src/home/NotificationBox.tsx
```diff
@@ -106,7 +106,7 @@ function useSimpleActions() {
   } = useSelector((state) => state.account)
 
   const phoneNumberVerified = useSelector(phoneNumberVerifiedSelector)
-  const numberVerifiedDecentrally = useSelector(numberVerifiedSelector)
+  const numberVerifiedDecentrally = useSelector(numberVerifiedDecentrallySelector)
```
kathaypacific, line 109: this is the key change, aside from the renaming

jeanregisser: I don't fully follow.
The same redux state is read: `state.app.numberVerified`. What's different?

jeanregisser (approved):

PR #3640, opened 2023-04-12
fix: do not display all header when results are empty

MuckT:
### Description

As the title describes.

#### Before / After iOS

<p align="left">

</p>

### Test plan

- Tested locally on iOS
- Unit tests updated

### Related issues

N/A

### Backwards compatibility

Yes


kathaypacific (approved):

jeanregisser:
How did it look before the change?

MuckT:
@jeanregisser I can grab a picture if needed, but essentially the all section header would display without any content below it.

PR #3641, opened 2023-04-12
chore: remove non celo news code path

MuckT:
### Description

- Removes the non Celo news code path from the Celo page.
- Old Celo exchange transaction feed items are now display as swaps.
- Adjusts the logic on the `src/exchange/CeloGoldHistoryChart.tsx` and `src/exchange/ExchangeHomeScreen.tsx` for getting the current CELO price in c5684ca7cf70b6f1e7bf73292dc270ece80cc197.

#### iOS Transaction Feed with exchange item

https://user-images.githubusercontent.com/26950305/232859439-272c17e6-86c4-4d8d-a0a3-65734b2afd47.mov

### Test plan

- Tests replaced in #3648.

### Related issues

- Fixes: RET-624

### Backwards compatibility

Yes


src/exchange/ExchangeHomeScreen.tsx
```diff
@@ -85,8 +71,6 @@ function ExchangeHomeScreen() {
 
   const { t } = useTranslation()
 
-  const { RESTRICTED_CP_DOTO } = useCountryFeatures()
```
jeanregisser, line 88: Can we remove `RESTRICTED_CP_DOTO` there too?

src/exchange/ExchangeHomeScreen.tsx
```diff
@@ -171,23 +155,7 @@ function ExchangeHomeScreen() {
...
-                  <CeloExchangeButtons />
-                ))}
-              <ItemSeparator />
-              <CeloGoldOverview testID="ExchangeAccountOverview" />
-              <ItemSeparator />
-              <SectionHead text={t('activity')} />
-              <TransactionsList feedType={FeedType.EXCHANGE} />
-            </>
-          )}
+          {isCeloNewsEnabled && <CeloNewsFeed />}
```
jeanregisser, line 158: We can remove `isCeloNewsEnabled` too, right?

src/transactions/TransactionsList.tsx
```diff
@@ -125,9 +125,7 @@ function mapExchangeStandbyToFeedItem(
   // Find amount relative to the queried currency
   const accountAmount = [makerAmount, takerAmount].find((amount) => {
     const amountCurrency = amount.currencyCode as Currency
-    return feedType === FeedType.EXCHANGE
```
jeanregisser, line 128: Can we remove `FeedType` too?

jeanregisser (approved):
Awesome!! Feels good to see all this removed. 💥 

I think we can remove even more things, right?
The old transaction list with support for Celo comes to mind.

We could also rename `src/exchanges` to `src/celoNews`.

src/exchange/ExchangeHomeScreen.tsx
MuckT: I debated this, but I though it might be a good idea to have a way to disable the news feed if we ever need it.

MuckT:
@jeanregisser I found a few more files to remove 😬

MuckT:
@kathaypacific I updated the schema in a29c6d0c757b3311cc5940d114cae5dfbf7a760c, but I am unsure if we need migration as the old exchange items appear to be handled by the feed correctly as swaps.

kathaypacific:
> @kathaypacific I updated the schema in [a29c6d0](https://github.com/valora-inc/wallet/commit/a29c6d0c757b3311cc5940d114cae5dfbf7a760c), but I am unsure if we need migration as the old exchange items appear to be handled by the feed correctly as swaps.

i don't think that we need a migration since we haven't modified any persisted state here. if i understand the ExchangeHomeScreen correctly, we're always fetching new data via apollo to generate the feed and we don't store it anywhere. so we should be safe :)

src/exchange/WithdrawCeloReviewScreen.tsx
```diff
@@ -125,7 +125,7 @@ WithdrawCeloReviewScreen.navigationOptions = ({
         icon={<BackButton />}
         testID="EditButton"
         onPress={onEdit}
-        eventName={CeloExchangeEvents.celo_sell_edit}
+        eventName={CeloExchangeEvents.celo_withdraw_edit}
```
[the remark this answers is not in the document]
MuckT: We don't need to change it, but based on the name the incorrect analytic event was tracked.

src/exchange/ExchangeHomeScreen.tsx
MuckT: If for some reason we lost access to our account for social media, it would be great to have a switch for disabling the news feed.

src/fiatExchanges/saga.test.ts
```diff
@@ -205,9 +205,9 @@ describe(tagTxsWithProviderInfo, () => {
       address: mockProviderAccount,
     },
     {
-      __typename: 'TokenExchange',
-      type: TokenTransactionType.Exchange,
-      hash: exchangeHash,
+      __typename: 'TokenTransfer',
```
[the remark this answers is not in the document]
MuckT: Removing this was causing some tests to fail.

PR #3642, opened 2023-04-12
chore: add cache control for fetching supercharge rewards

kathaypacific:
### Description

This PR expects that a `cache-control: max-age=X` will be returned in the response header for the fetch supercharge service.

After claiming rewards, refresh the cache. 

### Test plan

Checking that cache-control works:
1. Load the app - verify that the server has been hit, the response is stored
2. Reload the app - verify that the server is not hit again

Checking that the refresh works (after modifying the response on the server):
3. After dispatching an action that triggers the force refresh: verify that the server has been hit, the new response is stored
4. Reload the app - verify that the server is not hit again, and the response in step 3 (instead of 1) is stored

### Related issues

- Fixes RET-707

### Backwards compatibility

Y

### DIFF
--- a/android/app/src/main/java/org/celo/mobile/MainApplication.java
+++ b/android/app/src/main/java/org/celo/mobile/MainApplication.java
@@ -65,7 +65,7 @@ public class MainApplication
     super.onCreate();
     SoLoader.init(this, /* native exopackage */false);
     initializeFlipper(this, getReactNativeHost().getReactInstanceManager());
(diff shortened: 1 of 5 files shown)


jeanregisser (approved):
🚀

jeanregisser:
All good on Android too?

jeanregisser:
Oh great find! Thanks for checking and fixing! 💪 🚀 🎸

PR #3643, opened 2023-04-12
feat(home-actions): Show back button on screens reached from home actions

anthonydaoud:
### Description

Part 2 of ACT-704. Currently, the screens reached from the home actions can display the hamburger menu or the 'X' icon in the top left of the screen. This PR replaces these navigation methods with a back button only when the screens are accessed from the home actions. When reaching the screens via any other methods, the existing navigation option will still be shown. Note that navigation for the swap screen is handled in #3623.

https://user-images.githubusercontent.com/20508929/232064472-422cc988-c514-406b-b0ee-06d94d17e4c5.mp4

### Test plan

Added unit tests and updated snapshots.

### Related issues

- ACT-704

### Backwards compatibility

Yes.


src/qrcode/QRTabBar.tsx
```diff
@@ -5,17 +5,28 @@ import Animated from 'react-native-reanimated'
...
 import Times from 'src/icons/Times'
+import { navigate } from 'src/navigator/NavigationService'
+import { Screens } from 'src/navigator/Screens'
 import { TopBarIconButton } from 'src/navigator/TopBarButton'
 import { shareQRCode, SVG } from 'src/send/actions'
 import colors from 'src/styles/colors'
 
 type Props = MaterialTopTabBarProps & {
   qrSvgRef: React.MutableRefObject<SVG>
+  showBackButton: boolean
```
satish-ravi, line 19: Instead of a boolean, can we use an enum? it's not obvious that setting this to false means `X` is used

src/navigator/types.tsx
```diff
@@ -35,8 +35,8 @@ import {
 // Typed nested navigator params
 type NestedNavigatorParams<ParamList> = {
   [K in keyof ParamList]: undefined extends ParamList[K]
-    ? { screen: K; params?: ParamList[K] }
-    : { screen: K; params: ParamList[K] }
+    ? { screen: K; params?: ParamList[K]; showBackButton?: boolean }
+    : { screen: K; params: ParamList[K]; showBackButton?: boolean }
```
satish-ravi, line 39: I guess this is only used in the QRNavigator, but this doesn't seem the right place to add this prop since NestedNavigatorParams sounds like it could be used with other screens. Can we make this QR screen specific?

src/icons/BackChevron.tsx
```diff
@@ -13,11 +16,11 @@ function BackChevron({ color, height }: Props) {
       height={height}
       width={height && height / 2}
       viewBox="0 0 8 16"
-      fill="none"
       xmlns="http://www.w3.org/2000/svg"
       testID="BackChevron"
     >
-      <Path
+      <AnimatedPath
+        // @ts-ignore
```
satish-ravi, line 23: What's the purpose of ts-ignore here?

anthonydaoud: Typescript yells at you if that isn't present and makes you remove all the properties. Adding that ts-ignore is based on what is already done in the 'Times' component [here](https://github.com/valora-inc/wallet/blob/main/src/icons/Times.tsx#L24)

src/qrcode/QRTabBar.tsx
anthonydaoud: Sure thing, changed it to `qrCloseButton`

satish-ravi: Could we name this more generic like `closeIcon` / `leftNavIcon`?  Also, can you also make a similar change to the Send / Request screen? Sorry, I should've been more explicit on my original comment.

src/icons/BackChevron.tsx
satish-ravi: Do we know why TS doesn't like these params? If it's a known limitation, can we use `ts-expect-error` and also add a comment on why we expect error. I don't think we should be adding generic `ts-ignore`s, those can hide real issues in the future.

src/home/ActionsCarousel.test.tsx
```diff
@@ -21,11 +21,21 @@ describe('ActionsCarousel', () => {
...
 
   it.each([
-    [HomeActionName.Send, 'send', Screens.Send, undefined],
-    [HomeActionName.Receive, 'receive', Screens.QRNavigator, undefined],
+    [HomeActionName.Send, 'send', Screens.Send, { showBackButton: true }],
+    [
+      HomeActionName.Receive,
+      'receive',
+      Screens.QRNavigator,
+      { screen: Screens.QRCode, qrCloseButton: 'BackChevron' },
```
satish-ravi, line 29: nit: use enum instead of hardcoded string

src/qrcode/QRTabBar.tsx
anthonydaoud: No worries, updated the code so it has a `closeIcon` param on those screens

src/icons/BackChevron.tsx
anthonydaoud: Yeah I think it's because we were using the package `svgs` instead of `react-native-svg`. I'll update the `Times` component to use that package instead in a separate PR

satish-ravi: Seems like `svgs` is just a wrapper around `react-native-svg` and was surprised why this all of a sudden succeeded. Dig a little more digging on this and turns out we override ts declarations for svgs here: https://github.com/valora-inc/wallet/blob/44b94f0fb450156749872d240ce5f1e875968338/src/index.d.ts#L12
Git blame says this line is from the initial commit, probably this was added when this package didn't have type definitions. We should remove this.
Also, I wonder if we should consider switching to just `react-native-svg` throughout. Don't see any value in using `svgs` since everything in this repo is react-native.

src/qrcode/QRTabBar.tsx
```diff
@@ -38,9 +50,19 @@ export default function QRTabBar({ state, descriptors, navigation, position, qrS
...
 
-  const onPressClose = () => {
-    navigation.getParent()?.goBack()
-  }
+  const { closeIconComponent, onPressClose } = useMemo(
+    () =>
+      closeIcon === CloseIcon.BackChevron
+        ? {
+            closeIconComponent: <BackChevron color={color} />,
+            onPressClose: () => navigate(Screens.WalletHome),
```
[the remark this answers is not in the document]
anthonydaoud: I think I was having issues with navigation when I first added it, not sure why though because the `getParent` navigation works. I'll change it back that

satish-ravi (approved):

PR #3647, opened 2023-04-13
fix(swaps): Fix errors when swapping all the gas token.

dievazqu:
### Description

Our current swap fee estimation is a bit short sometimes, and if the users try to swap all the balance of the token they will pay the gas with, they usually get a swap rejection (because the swap balance exceeded the balance - fee, and even in some cases the error is a bit cryptic (see: [example1](https://explorer.celo.org/mainnet/tx/0x6a76234525a828a9a269b4124ce3e4be51f2a5a46d4d6ced5b40b93e16e4d1ad) or [example2](https://explorer.celo.org/mainnet/tx/0x2c9f8b5135939792b3c04f49678f3ba264adfaf7b81704c387f199bc8f23971a)).
All these problems seems to solved if we leave more room in the user's balance.

This PR:
* Increases the swap fee estimation multipliers (these are only used to subtract the total balance when pressing the max button) to avoid swap errors.
* Fixes the swap review to show the correct fee token, and a more precise amount (we were using the celo fee amount always).

### Test

Tested locally on a iOS emulator.

#### Testing results

I have been testing swapping the max amount with different amounts (the maximum I swapped was around 150-200 usd worth values), with the current multipliers it leaves around 2 cents USD worth of token as dust after the swap (this might vary depending on the specific swap route). I tested several times with the current multipliers and it worked always.

### Related issues

- Fixes RET-678 RET-688

### Backwards compatibility

It's backward compatible


MuckT (approved):
🚀 
Tested locally on Android and had a similar experience swapping around 60 cUSD to and from CELO was successful with a dust of  ~1 - 3 cents USD.

src/fees/saga.ts
```diff
@@ -161,11 +163,16 @@ export function* estimateSwapFee(tokenAddress: string) {
...
   // To calculate the fee, you need the txo but in the in-app swaps flow the
   // swap amount is required to query the 0x API for the txo. For now,
-  // approximate the fee to be 4x that of a simple transfer, to take into
-  // account long swap routes. The maximum i saw on Ubeswap was 3, choosing 4 as
-  // a buffer. The tradeoff for this approximation is that users may have more
-  // dust (on the order of 3x the gas fee)
-  const feeInfo: FeeInfo = yield call(calculateFeeForTx, tx.txo, SWAP_FEE_ESTIMATE_MULTIPLIER)
+  // approximate the fee to be 8x that of a simple transfer, to take into
+  // account long swap routes.
+  // Increased multiplier for CELO swaps because the ratio swap_fee / simple_transaction_fee is higher
```
jeanregisser, line 168: I don't fully follow the explanation for CELO, why do we need a bigger multiplier for CELO?

src/swap/SwapReviewScreen.tsx
```diff
@@ -193,11 +225,8 @@ export function SwapReviewScreen() {
                 <View style={styles.tokenDisplayView}>
                   <TokenDisplay
                     style={[styles.amountText, { color: colors.greenUI }]}
-                    // Fix: I think we shouldn't subtract gas from the buyAmount here.
                     amount={divideByWei(
```
jeanregisser, line 228: This isn't right for tokens not using 18 decimals, is this the bug you talked about on Slack?

src/swap/SwapReviewScreen.tsx
```diff
@@ -59,8 +60,39 @@ export function SwapReviewScreen() {
...
+    const feeCurrencyUsdPrice = tokensByAddress[feeCurrency]?.usdPrice
+
+    if (!celoUsdPrice || !feeCurrencyUsdPrice) {
+      return new BigNumber(0)
+    }
+
+    const estimatedCeloFeeAmount = divideByWei(
+      new BigNumber(swapResponse.unvalidatedSwapTransaction.gas).multipliedBy(
+        new BigNumber(swapResponse.unvalidatedSwapTransaction.gasPrice)
+      )
```
jeanregisser, line 83: nit: I think we can do this:
```suggestion
      new BigNumber(swapResponse.unvalidatedSwapTransaction.gas).multipliedBy(
        swapResponse.unvalidatedSwapTransaction.gasPrice
      )
```

jeanregisser (commented):
Thanks!

src/fees/saga.ts
dievazqu: Yeah, sorry, I didn't know how to explain this.
But basically, in our calculations, all non-CELO transaction fees are much higher than the CELO transaction fee. So, using the same multiplier for CELO is not reaching the fee needed for the swaps. But it does work for non-CELO tokens.

Imaginary example:

Transaction fee in CELO: 0.0001
Transaction fee in cUSD: 0.001

Minimum swap fee needed in CELO: 0.01
Minimum swap fee needed in cUSD: 0.01

So using the same multiplier is won't work here. I mean, we can use the CELO multiplier for everything, but we will leave more dust for non-CELO tokens.

src/swap/SwapReviewScreen.tsx
dievazqu: I saw a similar issue in other places, I'll add this case to the [task](https://linear.app/valora/issue/RET-712/[wallet]-swapping-tokens-with-different-decimals-is-not-working).

PR #3651, opened 2023-04-14
fix: build issue with XCode 14 and M1 machines

kathaypacific:
### Description

I have been getting build errors for iOS locally, and found a related GH issue for the same problem. 

### Test plan

My local build is running fine now, if it doesn't create problems for the CI then it should be a safe change.

### Related issues

N/A

### Backwards compatibility

Y


MuckT:
@kathaypacific not sure why this is failing I have a similar patch locally and it's working. Maybe how the patch was generated? I have to do the following to get patches to generate correctly:
1. `rm -rf node_modules
2. yarn.
3. Edits needed for the patch
4. `yarn patch-package <target-dependency>`
5. Commit that patch before building.

MuckT:
Opened #3656 which has a the fix, but a slightly different patch file.

kathaypacific:
Closing for #3656

PR #3653, opened 2023-04-14
chore: remove redundant haptic on settings item

kathaypacific:
### Description

As the title, this component already has native haptics so we don't need to add another one.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y


jeanregisser (approved):
📳

PR #3655, opened 2023-04-14
fix(swap): ensure that outdated exchange rates are not shown on switch token

kathaypacific:
### Description

Changing the swap token is displayed instantly on the screen, but it takes more time for the exchange rate to be refreshed. We weren't handling this before, so upon switch the token there would be a short time when we actually display the previous exchange rate with the new token. This is obviously wrong, and became more obvious with the new design.

This PR:
- essentially add metadata to the exchange rate, so that we know the params it was for. prevent displaying the exchange rate if the tokens or swap amounts change, in case the rate is wrong.
- the exchange rate disappears sometimes and looks a bit abrupt, there is a task in the backlog for adding a skeleton loader which is not included here.

### Test plan

I didn't add a unit test for this since this code change affects only an intermediary step. the final outcomes are already covered in existing tests.

https://user-images.githubusercontent.com/20150449/232015769-5c8bb240-bb5b-4b04-8b74-45f3b5b5f6a9.mp4

Previous behaviour:

https://user-images.githubusercontent.com/20150449/232016280-414ceed8-6ceb-4103-ac53-799468963fd1.mp4

### Related issues

- Fixes RET-714

### Backwards compatibility

Y


MuckT (approved):
👍 Tested locally and the expected results are matching on Android!

PR #3656, opened 2023-04-15
fix: yoga build issue with XCode 14

MuckT:
### Description

Backports a fix for a yoga build error in [react-native](https://github.com/facebook/react-native) for XCode 14 - [Related GitHub Issue](https://github.com/facebook/yoga/issues/1224) - [React Native Fix](https://github.com/facebook/yoga/commit/f174de70afdde2492e8677bd0e716eb41bf64469)
Adds the patches folder to the cache path to avoid a cache hit when only a patch file has changed.

### Test plan

Tested locally on iOS.

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -32,6 +32,7 @@ jobs:
       - uses: actions/cache@v3
         with:
           path: |
+            patches
             node_modules
             */*/node_modules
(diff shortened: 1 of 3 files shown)


kathaypacific (approved):
🙏🏻

PR #3660, opened 2023-04-18
chore: resolve warning for bottom sheet navigator

MuckT:
### Description

Noticed this warning when running in dev mode a lot. Applied the recommended fix from [here](https://github.com/react-navigation/react-navigation/issues/8517#issuecomment-663838298).

```Looks like you're passing an inline function for 'component' prop for the screen 'DappKitSignTxScreen' (e.g. component={() => <SomeComponent />}). Passing an inline function will cause the component state to be lost on re-render and cause perf issues since it's re-created every render. You can pass the function as children to 'Screen' instead to achieve the desired behaviour.```

### Test plan

Tested locally on iOS by swapping cUSD to Ethix on Ubeswap.

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/src/navigator/Navigator.tsx
+++ b/src/navigator/Navigator.tsx
@@ -8,10 +8,7 @@ import {
   createNativeStackNavigator,
   NativeStackNavigationOptions,
 } from '@react-navigation/native-stack'
-import {
-  BottomSheetScreenProps,
-  createBottomSheetNavigator,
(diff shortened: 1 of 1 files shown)


jeanregisser:
Thanks, note that it only hides the warning and doesn't address the possible performance issue.

https://reactnavigation.org/docs/screen/#children

PR #3661, opened 2023-04-19
feat: celo price chart localization

MuckT:
### Description

While working on the #3641, it became clear this small change could allow users to view the Celo Price history chart in their local currency. The biggest downside is are converting the historical USD price to the local currency instead of storing the historical price in the local currency; however, building out historical pricing for each available currency is app would be a much larger task.

#### Video iOS

https://user-images.githubusercontent.com/26950305/232938101-31dec0a1-c2fc-4c7a-813c-2c80a151902e.mov

### Test plan

Tested locally on iOS

### Related issues

N/A

### Backwards compatibility

Yes


src/exchange/CeloGoldHistoryChart.tsx
```diff
@@ -175,7 +178,7 @@ function Loader() {
 function CeloGoldHistoryChart({ testID, i18n }: Props) {
   // We hardcode localCurrencyCode to null, hence the chart will always show cGLD to cUSD no matter what.
   // TODO: revert this back to `useLocalCurrencyCode()` when we have history data for cGDL to Local Currency.
```
kathaypacific, line 180: nit: these comments can be removed

kathaypacific (approved):
this is great!! i had a quick sanity check for the prices we display vs what is on coin market cap and the prices are close enough (within a few cents in most cases) ✨

PR #3662, opened 2023-04-19
feat: debounce requests to resolve-id endpoint

silasbw:
We're going to start rate-limiting requests soon.

### Test plan

Start the send flow and type a valid .nom with a mapped address.


src/recipients/RecipientPicker.tsx
```diff
@@ -36,6 +36,7 @@ import colors from 'src/styles/colors'
 import fontStyles from 'src/styles/fonts'
 import Logger from 'src/utils/Logger'
 import networkConfig from 'src/web3/networkConfig'
+import AwesomeDebouncePromise from 'awesome-debounce-promise'
```
silasbw, line 39: `lodash.debounce` has weird behavior with promises / requires annoying plumbing: https://github.com/lodash/lodash/issues/4400

Selected `awesome-debounce-promise` based on recommendation from `react-async-hook` documentation.

MuckT: Had you considered using [use-debounce](https://www.npmjs.com/package/use-debounce)? If we're adding a debounce library I'd strongly prefer that as it's more react specific and better maintained. I see the last update of [awesome-debounce-promise](https://www.npmjs.com/package/awesome-debounce-promise) was 4 years ago vs 14 days for [use-debounce](https://www.npmjs.com/package/use-debounce). There was also a fair amount of discussion around why we shouldn't add a debounce library in https://github.com/valora-inc/wallet/pull/3602#discussion_r1157001660.

silasbw: That library has the behavior I'm trying to avoid with `awesome-debounce-promise`.  I can shoehorn an implementation into lodash.debounce for now.

silasbw:
If we try to add it to Send.tsx, we'll end up debouncing all result filtering (not just API calls).  For example, we'll add the debounce delay to local contact list filtering too.

silasbw:
It's not super clear to me how to move forward.  I can:
1. rate limit API calls at some other point in the code (but I'd like to avoid rate limiting other things as a side-effect, like the contact list); 
2. see if we can contort lodash.debounce for our needs (but see https://github.com/lodash/lodash/issues/4400); or

kathaypacific (commented):
the react async hooks lib is not very active (the last publish was 2 years ago) so i think that it's reasonable to be suspicious about its recommendations and how current / active they are. i'd love to not introduce a maybe abandoned new dependency for this one use case.

i see that we are trying to debounce the promise itself, which lodash doesn't handle nicely. what if we tried to debounce the trigger for the `useAsync` hook instead? the `useAsync` hook is currently triggered any time the `props.searchQuery` value changes, but what if we replace the hook trigger with a `debouncedSearchQuery` variable instead - we can add a new state variable and update this state variable in a debounced way, and use that in the dependency array of `useAsync`. idk, something like 
```
  const [debouncedSearchQuery, setDebouncedSearchQuery] = useState(props.searchQuery)

  const debounceSearchQuery = React.useCallback(
    debounce((query: string) => {
      setDebouncedSearchQuery(query)
    }, TYPING_DEBOUNCE_MILLSECONDS),
    []
  )

  React.useEffect(() => {
    debounceSearchQuery(props.searchQuery)
  }, [props.searchQuery])
```
i think in this case it's safe to use the lodash debounce, unless i'm missing something?

MuckT (approved):
@silasbw tested locally and the debounce is working well! 🚀 

#### Entering a phone number in quickly before
```
 LOG    [19:26:57.582Z] SM-G991U1 resolveId
 LOG    [19:26:58.000Z] SM-G991U1 resolveId
 LOG    [19:26:58.463Z] SM-G991U1 resolveId
 LOG    [19:26:59.272Z] SM-G991U1 resolveId
 LOG    [19:26:59.838Z] SM-G991U1 resolveId
 LOG    [19:27:00.244Z] SM-G991U1 resolveId
 LOG    [19:27:01.345Z] SM-G991U1 resolveId
 LOG    [19:27:01.893Z] SM-G991U1 resolveId
 LOG    [19:27:02.661Z] SM-G991U1 resolveId
 LOG    [19:27:03.077Z] SM-G991U1 resolveId 
 ```
#### Entering a phone number in quickly after
```
 LOG    [19:28:35.647Z] SM-G991U1 resolveId
```

PR #3663, opened 2023-04-20
chore: ensure backwards compatibility when dapps v2 data is used

kathaypacific:
### Description

We persist the dapps data in redux, so existing users will have the v1 dapps data schema. When the experiment is switched on for the user, they will first land on the new dapps screens with v1 data while the request to fetch new data is kicked off. During this time, the screen tries to render and crashes due to data incompatibility. 

### Test plan

To repro:
1. Load the app, visit the dapps screen. Observe current dapps screen loads.
2. In statsig, assign the search or search + filter variant to your user
3. Reload the app, visit the dapps screen. Note that the app crashes, and note that the v2 data is not received in time (so it is still v1 data in the redux store). **This means that the user will continue to to get app crashes every time they visit the dapps screen.**

https://user-images.githubusercontent.com/20150449/233394570-3fe528a6-907a-4239-9b7c-2178cf2299ce.mp4

### Related issues

N/A

### Backwards compatibility

Y


src/dapps/selectors.ts
```diff
@@ -23,6 +21,8 @@ export const dappsCategoriesSelector = (state: RootState) => state.dapps.dappsCa
 
 export const dappsListSelector = (state: RootState) => state.dapps.dappsList
 
+export const dappsV2ListSelector = (state: RootState) => state.dapps.dappsList.filter(isDappV2)
```
satish-ravi, line 24: Is it possible for dappsList to include both v1 and v2 dapps? Or is it always one or the other?
Also, should the other selector return v1? Or it's usages can work with both v1 and v2 schema?

satish-ravi (commented):
Looks like some tests are failing

src/dapps/selectors.ts
kathaypacific: `dappListSelector` returns the dapps list regardless of if it's v1 or v2, that's okay for most cases where we just care about the dapp and not the category (e.g. featured, favourited, recently used etc)

it's not possible for the dappsList to include both v1 and v2 dapps, i actually took a shortcut here in the selector. what i really want is for the dappsList to be cleared when the experiment is switched on for the first time since the new screens is going to expect a different data format for rendering. i could either watch the experiment and clear the store or just not return the dappsList until the fetch is complete / the store is refreshed with new data. i opted for the latter because it is easier and also i don't know how to watch statsig experiment variables lol 😅

dievazqu (approved):

jeanregisser (approved):
Great catch! 🙌

MuckT (approved):
Tested locally with the [Statsig setup](https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/experiments/dapps_filters_and_search/setup) and all the variants were working switching between after app restart 🚀

PR #3666, opened 2023-04-21
chore: ensure statsig init before sagas init

kathaypacific:
### Description

Satish alerted me to this `'Statsig', 'getExperimentParams: Error getting params for experiment: dapps_filters_and_search', [Error: Call and wait for initialize() to finish first.]` error that occurs on app start every time. This error occurs because the root saga kicks off the dapps saga which kicks off the saga to fetch the dapps list. We query statsig for the experiment variables in this fetch dapps list saga. All of this happens when the redux store is initialised, which is actually before the AppInitGate.

I'm not sure of the context of why the statsig initialisation was done inside the analytics class. Does it need to be? If not, it would be nice to decouple the 2 (since the dependency between the 2 is not obvious). I put the statsig init inside the app init saga, which is run before all other sagas are spawned. This makes it safe to call statsig anywhere else in the app. Yes, this does delay the app since all sagas now wait for statsig, but from the AppInitGate i think this is the result we were looking to achieve anyway.

### Test plan

Actually this error only happens on app load, there are subsequent requests to fetch the dapps list which succeed. So no visible impact, experiment should work as usual. The error is no longer logged, and the app does not feel slower to start.

### Related issues

n/a

### Backwards compatibility

Y

### DIFF
--- a/__mocks__/src/statsig/index.ts
+++ b/__mocks__/src/statsig/index.ts
@@ -1,0 +1,7 @@
+module.exports = {
+  patchUpdateStatsigUser: jest.fn(),
+  getDefaultStatsigUser: jest.fn(),
+  getDynamicConfigParams: jest.fn(),
+  getExperimentParams: jest.fn(),
+  initializeStatsig: jest.fn(),
+}
--- a/src/analytics/ValoraAnalytics.ts
+++ b/src/analytics/ValoraAnalytics.ts
@@ -15,14 +15,10 @@ import {
   isE2EEnv,
   SEGMENT_API_KEY,
   STATSIG_ENV,
-  E2E_TEST_STATSIG_ID,
-  STATSIG_API_KEY,
 } from 'src/config'
 import { store } from 'src/redux/store'
 import Logger from 'src/utils/Logger'
 import { isPresent } from 'src/utils/typescript'
-import { getDefaultStatsigUser } from 'src/statsig'
-import { Statsig } from 'statsig-react-native'
 
 const TAG = 'ValoraAnalytics'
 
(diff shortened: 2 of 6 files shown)


__mocks__/src/statsig/index.ts
```diff
@@ -0,0 +1,7 @@
+module.exports = {
```
satish-ravi, line 1: Looks like this may have broken getExperimentParams mocks in various test files. Any reason to have these generic mocks?

src/analytics/ValoraAnalytics.ts
```diff
@@ -129,21 +125,6 @@ class ValoraAnalytics {
     } catch (error) {
       Logger.error(TAG, `Segment setup error: ${error.message}\n`, error)
     }
-
-    try {
-      const statsigUser = getDefaultStatsigUser()
-      // getAnonymousId causes the e2e tests to fail
-      const overrideStableID = isE2EEnv ? E2E_TEST_STATSIG_ID : await Analytics.getAnonymousId()
```
satish-ravi, line 136: I guess this was originally here because we assumed getting anonymousId should be after initializing segment, but this may not be true. If we don't need it and if segment consistently returns the ID before / after initialization, I guess we can also revert adding ValoraAnalytics.init() to AppInitGate and do it async as it was prior to this commit https://github.com/valora-inc/wallet/commit/5a982019f89ef552c8fc9e5f1b0fee7bdedd7c19

MuckT:
@kathaypacific played around with this locally on Android and it appears to be working if I am on the dapps page and swap my statsig bucketing.

kathaypacific:
Closing for #3672

PR #3667, opened 2023-04-21
chore: upgrade to react native 0.67

kathaypacific:
### Description

As the title.

This was a pretty small upgrade, but one thing did change: we use `<Modal>` components pretty extensively, these have children and a `visible` prop. if `visible` is false, the Modal doesn't render its children in real life but in the jest environment the children used to be there. This is wrong - in many tests were were asserting that elements inside `<Modal visible={false}>` components were present / tappable etc, but actually in real life the `<Modal>` component wouldn't have rendered any children. All the snapshot updates in this PR are for this change, and several tests needed to be updated to be correct.

### Test plan

Did a quick manual test on iOS and Android, rely on CI

### Related issues

- Fixes RET-683

### Backwards compatibility

Y


patches/react-native+0.67.5.patch
```diff
@@ -1,40 +1,8 @@
...
-+++ b/node_modules/react-native/Libraries/Components/ScrollView/ScrollView.js
-@@ -1178,11 +1178,6 @@ class ScrollView extends React.Component<Props, State> {
-         );
-       }
-     }
--    if (Platform.OS === 'android') {
--      if (this.props.keyboardDismissMode === 'on-drag' && this._isTouching) {
--        dismissKeyboard();
--      }
--    }
```
kathaypacific, line 13: this patch isn't needed anymore

yarn.lock
```diff
@@ -15544,18 +15544,18 @@ react-async-hook@^4.0.0:
   resolved "https://registry.yarnpkg.com/react-async-hook/-/react-async-hook-4.0.0.tgz#1f0467586654e1f33b7433bd98c300a0c5f9b3d0"
   integrity sha512-97lgjFkOcHCTYSrsKBpsXg3iVWM0LnzedB749iP76sb3/8Ouu4nHIkCLEOrQWHVYqrYxjF05NN6GHoXWFkB3Kw==
 
-react-devtools-core@4.26.0:
```
MuckT, line 15547: This looks like it downgraded. This seems suspect with the [yarn lock check fail.](https://github.com/valora-inc/wallet/actions/runs/4766429475/jobs/8473430942?pr=3667)

MuckT: Otherwise this is working for me locally on iOS and Android.

kathaypacific: it didn't downgrade, if you scroll down you can see that it upgraded from 4.13 to 4.19 (just the order was switched so the diff looks a bit weird)

kathaypacific: the yarn.lock job uses `actions/cache@v3` but i don't think this is working as expected. i tried without this step and the job succeeded :/

.github/workflows/check.yml
```diff
@@ -32,10 +32,9 @@ jobs:
       - uses: actions/cache@v3
         with:
           path: |
-            patches
             node_modules
             */*/node_modules
-          key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
+          key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock', '**/patches') }}
```
kathaypacific, line 37: i think what we actually want is to fetch fresh node modules if the yarn.lock OR any patches change. without taking into account the patches, PRs like [this](https://github.com/valora-inc/wallet/pull/3656) would fail - examples of failures are in the CI in https://github.com/valora-inc/wallet/pull/3651.

it seems like `path` was the wrong place to put the patches though. according to the [docs](https://github.com/actions/cache#inputs) this would mean that we are caching the patches, which is unnecessary. it also causes the wrong patch to be applied (e.g. https://github.com/valora-inc/wallet/actions/runs/4779094863/jobs/8495947030)

it seems [possible](https://docs.github.com/en/actions/learn-github-actions/expressions#example-with-multiple-patterns) to add multiple things to the `hashFiles` function. i tested this change by seeing that the [CI still passes](https://github.com/valora-inc/wallet/actions/runs/4779175382/attempts/1) -> [rerunning the same job and verifying that the cache is applied](https://github.com/valora-inc/wallet/actions/runs/4779175382/jobs/8496118656) -> creating a commit that just changes a patch file and verifying that [the cache is not hit](https://github.com/valora-inc/wallet/actions/runs/4779203159/jobs/8496132136?pr=3667), fresh node modules are fetched and the job still runs.

i left this change here because it is small and directly affecting this PR, if someone feels strongly then i can move it out to a separate PR. FYI @MuckT

MuckT (approved):
@kathaypacific LGTM thanks for fixing and explaining the patch issue!

.github/workflows/check.yml
jeanregisser: Also proposing to match only the patches at the root of the repo ([docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#patterns-to-match-file-paths)):
```suggestion
          key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock', 'patches/*.patch') }}
```

src/account/ShakeForSupport.test.tsx
```diff
@@ -22,24 +20,4 @@ describe('ShakeForSupport', () => {
...
-
-  it('navigates to support screen after button press', async () => {
-    const { getByTestId } = render(
-      <Provider store={createStore(AppState.Active)}>
-        <ShakeForSupport />
-      </Provider>
-    )
-    fireEvent.press(getByTestId('ContactSupportFromShake'))
-    expect(navigate).toHaveBeenCalledWith(Screens.SupportContact)
-  })
```
jeanregisser, line 44: Why are we removing these?

jeanregisser (approved):
Awesome 🚀

src/account/ShakeForSupport.test.tsx
kathaypacific: oh sorry, should have explained.

it relates to the key change where <Modal visible={false} /> components are no longer rendered in the test dom. notice how in this unit test, we never bother to actually _launch_ the bottom sheet - we make all these assertions and actions but the whole time, the Modal has `visible={false}`.

the launching of the sheet is done by shake action. i couldn't figure out how to do this with this testing lib. we also cover this exact test in the e2e tests so i thought it was okay to okay to remove these

jeanregisser: Oh ok, thanks, if it's cover by the e2e test it's all good then 👍

PR #3669, opened 2023-04-21
chore: update Supercharge screen copy for V.1.53 and earlier

mpgaarciaa:


locales/base/translation.json
```diff
@@ -209,8 +209,8 @@
     "body": "You’ll get rewarded weekly on your Celo stablecoin with the highest value (cEUR, cUSD, cREAL).",
     "dismiss": "Dismiss"
   },
-  "superchargingTitle": "Your {{token}} is being Supercharged",
```
MuckT, line 212: Wait will removing token cause a crash on v1.53?

MuckT (changes requested):

locales/base/translation.json
MuckT: Tested locally on 1.53 with the new strings and it appears to be fine to remove `{{token}}`.

MuckT (approved):

PR #3672, opened 2023-04-24
chore: ensure statsig / analytics / i18n initialised before rendering app

kathaypacific:
### Description

Satish alerted me to this `'Statsig', 'getExperimentParams: Error getting params for experiment: dapps_filters_and_search', [Error: Call and wait for initialize() to finish first.]` error that occurs on app start every time. This error occurs because the root saga kicks off the dapps saga which kicks off the saga to fetch the dapps list. We query statsig for the experiment variables in this fetch dapps list saga. All of this happens when the redux store is initialised, which is actually before the AppInitGate.

We essentially need to sync the "side effects" and "render" parts of the app. 

This PR: 
- await sentry / analytics / i18n in app init saga
- ensure AppInitGate waits for all sagas to finish initialising before starting the "render" part of the app
- ensure that analytics and i18n are not referenced before the AppInitGate

### Test plan

Actually this error only happens on app load, there are subsequent requests to fetch the dapps list which succeed. So no visible impact, experiment should work as usual. The error is no longer logged, and the app does not feel slower to start.

### Related issues

n/a

### Backwards compatibility

Y


src/app/saga.ts
```diff
@@ -93,14 +94,19 @@ const DO_NOT_LOCK_PERIOD = 30000 // 30 sec
...
   // This step is important if the user if offline and unable to fetch remote
   // config values, we can use the persisted value instead of an empty one
   const sentryNetworkErrors = yield select(sentryNetworkErrorsSelector)
   Logger.setNetworkErrors(sentryNetworkErrors)
 
   const supportedBiometryType = yield call(Keychain.getSupportedBiometryType)
   yield put(setSupportedBiometryType(supportedBiometryType))
+
+  const t1 = Date.now()
+  ValoraAnalytics.track(AppEvents.app_tooling_initialized, { durationMs: t1 - t0 })
```
cajubelt, line 109: this seems like a bit of a hack to get performance profiling in this one place. I'd prefer if product analytics and performance analytics were more separate and used domain-specific tooling for each. Could we try to use Sentry profiling for this instead?

src/analytics/Events.tsx
```diff
@@ -22,6 +22,8 @@ export enum AppEvents {
 
   account_funded = 'account_funded',
   account_liquidated = 'account_liquidated',
+
+  app_tooling_initialized = 'app_tooling_initialized', // when app tooling like sentry and analytics have finished initializing
```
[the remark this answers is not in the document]
kathaypacific: i intentionally left this vague, like this used to be sentry and now we included analytics...it didn't seem far away that we might want to add something else. i didn't want us to have to change the analytics name every time we wanted to update this logic :/

src/app/saga.ts
kathaypacific: it is kind of 😅 i guess i want to understand how much of a performance hit our users are incurring. but i don't trust in our understanding of sentry - it's easy enough to put sentry instrumentation in here but we've seen all kinds of weird reporting in sentry. idk, i can also use sentry for this but i'd like to keep the analytics event either way. it seems useful for measuring MAU's (since we don't want to use the "fetch balances" one)

MuckT: We actually already have a a custom instrumentation implemented in #2728 that we could use.

kathaypacific: @satish-ravi this will only be triggered for a cold start, if the app has been backgrounded for long enough and been terminated by the device then it will be run. but for a quick background - foreground switch then it will not run.

it seems kind of wild to use sentry to measure sentry's init time, it didn't log any errors so i put it in since everyone feels so strongly about it. but i would like to keep the analytics event in as a sanity check, in case the sentry data does not make sense

kathaypacific: fwiw i do believe in using sentry, but recently i've had a hard time understanding and trusting the data that we see in the dashboards. so i was reluctant to throw more data there that we might not understand, but maybe it's just me who can't make sense of the data. it'd be nice to have a deep dive brownbag if anyone has more knowledge (for example, the dapps screen has a p95 value of [over 1 minute](https://valora-inc.sentry.io/performance/summary/?project=1250733&query=&referrer=performance-transaction-summary&statsPeriod=14d&transaction=DAppsExplorerScreen&unselectedSeries=p100%28%29) but we cache the results so there should be virtually no load time for the user...)

src/app/AppInitGate.tsx
```diff
@@ -41,9 +41,9 @@ const AppInitGate = ({ loading, children }: Props) => {
 
   const initResult = useAsync(
     async () => {
-      Logger.debug(TAG, 'Starting init')
-      await Promise.all([i18nInitializer(), ValoraAnalytics.init()])
-      Logger.debug(TAG, 'init completed')
+      Logger.debug(TAG, 'Starting AppInitGate init')
+      await Promise.all([i18nInitializer(), waitUntilSagasFinishLoading()])
+      Logger.debug(TAG, 'AppInitGate init completed')
```
jeanregisser, line 46: Nothing to do in this PR, but this reminded me we should measure how long it takes from app start to [dimissing the loading splash screen](https://github.com/valora-inc/wallet/blob/6d076bc9b2a83fa340941a8a1a241af8e5180382/src/navigator/Navigator.tsx#L643).

src/app/saga.ts
```diff
@@ -93,14 +96,22 @@ const DO_NOT_LOCK_PERIOD = 30000 // 30 sec
 // Work that's done before other sagas are initalized
 // Be mindful to not put long blocking tasks here
 export function* appInit() {
-  yield call(initializeSentry)
+  SentryTransactionHub.startTransaction(SentryTransaction.app_tooling_initialised)
+  const t0 = Date.now()
```
jeanregisser, line 100: nit: it's better to use a [monotonic clock](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now#performance.now_vs._date.now), which don't suffer from possible time adjustments.
```suggestion
  const t0 = performance.now()
```

src/app/saga.ts
```diff
@@ -93,14 +96,22 @@ const DO_NOT_LOCK_PERIOD = 30000 // 30 sec
...
+
   // This step is important if the user if offline and unable to fetch remote
   // config values, we can use the persisted value instead of an empty one
   const sentryNetworkErrors = yield select(sentryNetworkErrorsSelector)
   Logger.setNetworkErrors(sentryNetworkErrors)
 
   const supportedBiometryType = yield call(Keychain.getSupportedBiometryType)
   yield put(setSupportedBiometryType(supportedBiometryType))
+
+  const t1 = Date.now()
```
jeanregisser, line 112:
```suggestion
  const t1 = performance.now()
```

jeanregisser (approved):
Great!

src/app/useDeepLinks.ts
```diff
@@ -0,0 +1,92 @@
+import dynamicLinks from '@react-native-firebase/dynamic-links'
```
kathaypacific, line 1: these changes are from App.tsx, only i split apart handling initial links / future links to make it easier to read

kathaypacific:
@jeanregisser @cajubelt @satish-ravi @MuckT @silasbw this PR took a turn today as i went deeper down this rabbit hole.

i was trying to ensure that the correct tools were initialised before they were referenced, but found the i18n + analytics were both being referenced in the App.tsx and sagas. this kind of defeats the purpose of having an AppInitGate.

i bounced these ideas around with Jean to check that i'm not totally crazy. the key changes:
- the app init saga should be the place where we initialise app wide tools, this is executed first before spawning the rest of the sagas. this includes sentry, analytics, i18n so far.
- the AppInitGate's purpose now is to block the app from rendering until the sagas are spawned. it also starts the analytics session and attaches deep link handlers - these tasks were done by App.tsx before, which was racy since analytics and firebase weren't guaranteed to have initialised yet
-  the app_launched event now effectively tracks the performance impact that i was interested in, so no need to add a new event

src/app/App.tsx
```diff
@@ -74,103 +68,22 @@ export class App extends React.Component<Props> {
...
-      appLoadDuration,
-      language: i18n.language || store.getState().i18n.language,
-    })
   }
 
   componentWillUnmount() {
-    CleverTap.removeListener('CleverTapPushNotificationClicked')
-    Linking.removeEventListener('url', this.handleOpenURL)
-    this.dynamicLinksRemoveListener?.()
     store.dispatch(appUnmounted())
```
jeanregisser, line 74: Maybe we should move it closer to `appMounted`?

src/app/useDeepLinks.ts
```diff
@@ -0,0 +1,92 @@
...
+    }
+  }
+
+  useEffect(() => {
+    return () => {
+      CleverTap.removeListener('CleverTapPushNotificationClicked')
+      linkingUnsubscribeRef.current?.remove()
+      dynamicLinksUnsubscribeRef.current?.()
+    }
+  })
```
jeanregisser, line 87:
```suggestion
  }, [])
```

src/app/saga.ts
```diff
@@ -93,14 +108,33 @@ const DO_NOT_LOCK_PERIOD = 30000 // 30 sec
...
+    call(ValoraAnalytics.init),
+    call(
+      initI18n,
+      language || bestLanguage || DEFAULT_APP_LANGUAGE,
+      allowOtaTranslations,
+      otaTranslationsAppVersion
+    ),
+  ])
+
   // This step is important if the user if offline and unable to fetch remote
```
jeanregisser, line 129:
```suggestion
  // This step is important if the user is offline and unable to fetch remote
```

src/sentry/SentryTransactions.tsx
```diff
@@ -59,4 +60,8 @@ export const SentryTransactions: Record<values, SentryTransactionInfo> = {
     name: 'Wallet Connect Transaction',
     op: 'wallet_connect_transaction',
   },
+  AppToolingInitialised: {
+    name: 'App Tooling Initialised',
+    op: 'app_tooling_initialised',
+  },
```
jeanregisser, line 66: Maybe?
```suggestion
  AppInitSaga: {
    name: 'App Init Saga',
    op: 'app_init_saga',
  },
```

src/app/AppInitGate.tsx
```diff
@@ -1,61 +1,72 @@
...
+      ValoraAnalytics.startSession(AppEvents.app_launched, {
+        deviceHeight: height,
+        deviceWidth: width,
+        reactLoadDuration,
+        appLoadDuration,
+        language: i18n.language || language,
+      })
+
+      Logger.debug(TAG, 'AppInitGate init completed')
+      dispatch(appMounted())
```
jeanregisser, line 56: Now that I'm seeing this, I wonder if this part should maybe remain in `App.tsx` and called by a `onLoaded` callback passed to `AppInitGate`. The deeplinks would remain in `App.tsx` too.

src/app/AppInitGate.tsx
```diff
@@ -1,61 +1,72 @@
...
     async () => {
-      Logger.debug(TAG, 'Starting init')
-      await Promise.all([i18nInitializer(), ValoraAnalytics.init()])
-      Logger.debug(TAG, 'init completed')
+      Logger.debug(TAG, 'Starting AppInitGate init')
+      await waitUntilSagasFinishLoading()
+
+      if (!language && bestLanguage) {
+        await changelanguage(bestLanguage)
+      }
```
jeanregisser, line 40: Nothing to do now, but I wonder why we have this here. Shouldn't `initI18n` somehow take care of this?

jeanregisser (commented):
Thanks!

src/app/App.tsx
kathaypacific: yes good idea, done!

src/app/useDeepLinks.ts
jeanregisser: More to this, I'm a bit concerned this could cause subtle bugs in the future, since adding the listeners is done outside of this.
For instance in [React Strict](https://react.dev/reference/react/StrictMode) mode, the effects will be applied twice in dev mode.
I know we don't have this yet, but can we do this differently?

src/app/AppInitGate.tsx
kathaypacific: curious, what is the benefit of keeping this code in App.tsx? it'd be less changes to this PR of course, but aside from the reactLoadTime and appStartedMillis variables none of the other logic is tied to this component. we'd have functions here that are executed in a child component, why not put the function where it will be run?

jeanregisser: Yes I don't know, it's more a gut feeling, this part is arguably not directly part of the init.
It's what happens after init is done. Seeing it within `AppInitGate` feels like the component does too many things.
"oh look it also handles deeplinks".

src/app/useDeepLinks.ts
kathaypacific: maybe this also addresses the "oh look, this component also does deep links" concern? although now it's like "hey look the nav wrapper also does deep links" 😅 let's decide on an approach first, every time i move things around the tests are a pain to update so i'll wait until the end to update the tests

jeanregisser: Yes the deep links changes look good 👍

jeanregisser (approved):

PR #3673, opened 2023-04-24
feat: add recipient type to send/request events

anthonydaoud:
### Description

Currently, the Send/Request Payment flow logs some events without including whether the recipient is a phone number, wallet address, or nom space address. This PR adds this additional info to these events, as well as a property indicating if the recipient info was obtained via a QR code scan. 

### Test plan

Updated unit tests.

### Related issues

- ACT-723

### Backwards compatibility

Yes.


src/analytics/Properties.tsx
```diff
@@ -491,8 +491,8 @@ interface EscrowEventsProperties {
 interface SendEventsProperties {
   [SendEvents.send_scan]: undefined
   [SendEvents.send_select_recipient]: {
-    // TODO: decide what recipient info to collect, now that RecipientKind doesn't exist
     usedSearchBar: boolean
+    recipientType?: RecipientType
```
cajubelt, line 495: do you know what the cases are where recipientType is undefined? it would be helpful to document in a code comment here since I think this is where people look to understand the analytics data

src/analytics/Properties.tsx
```diff
@@ -632,6 +632,8 @@ interface RequestEventsProperties {
   [RequestEvents.request_confirm_back]: undefined
   [RequestEvents.request_confirm_request]: {
     requesteeAddress: string
+    recipientType?: RecipientType
+    isScan: boolean
```
cajubelt, line 636: nit: would be great to align the field names with the types (instead of `isScan` and `isFromScan` having 2 different names), and maybe to make them a bit more self-documenting too. I'd probly go with `isFromQRScan`

src/navigator/types.tsx
```diff
@@ -236,10 +236,12 @@ export type StackParamList = {
   [Screens.OutgoingPaymentRequestListScreen]: undefined
   [Screens.PaymentRequestConfirmation]: {
     transactionData: TransactionDataInput
+    isFromScan?: boolean
```
cajubelt, line 239: seems like it would be safer to make this required. otherwise it seems like we could easily just leave out the optional field, and start getting bad data, because later it is assumed that when `isFromScan` is undefined, it means false.

cajubelt: setting it explicitly everywhere seems safer

src/paymentRequest/PaymentRequestConfirmation.tsx
```diff
@@ -92,6 +92,8 @@ function PaymentRequestConfirmation({ route }: Props) {
 
     ValoraAnalytics.track(RequestEvents.request_confirm_request, {
       requesteeAddress: paymentInfo.requesteeAddress,
+      recipientType: route.params.transactionData.recipient.recipientType,
+      isScan: !!route.params.isFromScan,
```
cajubelt, line 96: here is where it is assumed that when isFromScan is undefined, it means false

cajubelt (commented):
note that CI fails due to test coverage

src/analytics/Properties.tsx
cajubelt: this property is being added to an existing event, so there isn't a backwards compatibility issue to just changing it here. But keeping it consistent with the property names on other events sounds fine

src/navigator/types.tsx
anthonydaoud: Updated so it's non-optional

src/send/SendConfirmation.tsx
```diff
@@ -82,6 +82,7 @@ export function useRecipientToSendTo(paramRecipient: Recipient) {
         // Setting the phone number explicitly so Typescript doesn't complain
         e164PhoneNumber: paramRecipient.e164PhoneNumber,
         address: recipientAddress ?? undefined,
+        recipientType: recipientAddress ? RecipientType.Address : RecipientType.PhoneNumber,
```
cajubelt, line 85: this is inside a block entered when phone number is defined and address is not defined. shouldn't the recipientType be phone number?

src/send/utils.ts
```diff
@@ -135,7 +136,7 @@ export function* handleSendPaymentData(
 export function* handlePaymentDeeplink(deeplink: string) {
   try {
     const paymentData = uriDataFromUrl(deeplink)
-    yield call(handleSendPaymentData, paymentData)
+    yield call(handleSendPaymentData, paymentData, false)
```
cajubelt, line 139: aren't payment deep links usually from QR codes that users scan?

src/transactions/actions.ts
```diff
@@ -184,5 +184,9 @@ export const navigateToPaymentTransferReview = (
 }
 
 export const navigateToRequestedPaymentReview = (transactionData: TransactionDataInput) => {
-  navigate(Screens.SendConfirmationLegacy, { transactionData, origin: SendOrigin.AppRequestFlow })
+  navigate(Screens.SendConfirmationLegacy, {
+    transactionData,
+    origin: SendOrigin.AppRequestFlow,
+    isFromScan: false,
```
cajubelt, line 190: It looks like this is only used in a notifications saga right now, but it seems to me like navigating to a requested payment confirmation screen is something that could actually take place in response to a QR code scan.  In light of that, it might make more sense to make isFromScan a parameter of this function.

src/merchantPayment/MerchantPaymentScreen.tsx
```diff
@@ -125,7 +126,8 @@ function MerchantPaymentScreen({ route }: Props) {
                 recipient={{
                   name: businessInformation?.name,
                   thumbnailPath: businessInformation?.imageUrl,
-                  address: businessInformation?.address?.city || 'Missing address', // TODO: what do we do here
+                  address: '0x0', // Use as default address, only used for Avatar color
```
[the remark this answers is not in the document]
anthonydaoud: Yeah I think that's reasonable, I added the TODO comment.

src/send/utils.ts
anthonydaoud: Yeah good catch, changed to isFromScan to true

src/send/SendConfirmation.tsx
anthonydaoud: Yeah good call, changed to phone number

cajubelt (approved):
🥳

PR #3684, opened 2023-04-27
chore(deps): upgrade Segment to v2 (along with all its plugins) and Firebase to v17

jh2oman:
### Description

upgrade `@react-native-firebase/*` deps from v11 to v17
upgrade `@segment/*` deps from v1 to v2

### Audit Advisory

This audit advisory should be able to be ignored by us because its part of the `@expo/config-plugins` package which we definitely don't make use of. (Expo is a platform to develop react native that we don't use).

### Test plan

I have manually tested the following: 
* identify, screen, track events all posted to segment with the data we expect
* statsig client still initialized with same anonymousId as Segment uses
* values are still read correctly from firebase realtime DB
* feature flags are loaded correctly from firebase remote config
* notifications still received from firebase messaging
* deep links still work from firebase dynamic links

I recommend we re-test notifications and deeplinks in bugapalooza. 

### Related issues

https://linear.app/valora/issue/ACT-727/upgrade-segment-firebase-deps-in-wallet-repo

### Backwards compatibility

Yes


src/analytics/ValoraAnalytics.ts
```diff
@@ -109,6 +110,9 @@ class ValoraAnalytics {
         writeKey: SEGMENT_API_KEY,
         storePersistor: AsyncStoragePersistor,
       })
+      await LegacySegment.setup('n/a', {})
+      const anonymousId = await LegacySegment.getAnonymousId()
```
cajubelt, line 114: I think it might be overkill to import LegacySegment just to do this. Can we import NativeModules from react-native and call `NativeModules.RNAnalytics.getAnonymousId()` here instead?

jh2oman: `RNanalytics` is actually a custom native module made by segment inside of the legacy `@segment/react-native-analytics` package. So we would have to re-implement large parts of that to replace it

cajubelt: dang

src/analytics/AsyncStoragePersistor.ts
```diff
@@ -0,0 +1,34 @@
...
+import Logger from 'src/utils/Logger'
+
+const TAG = 'AsyncStoragePersistor'
+/**
+ * Persistor implementation using AsyncStorage
+ * From https://github.com/segmentio/analytics-react-native/blob/master/packages/sovran/src/persistor/async-storage-persistor.ts
+ *
+ * This is almost exactly the same as the default persistor that segment uses however our build was not liking the persistor linked above due to its
+ * use of require('@react-native-async-storage/async-storage'). The require was resolving to { default: AsynctStorage } instead of just AsyncStorage
+ * so calls to AsyncStorage.getItem and AsyncStorage.setItem were not happening because they were undefined. This is a workaround to that issue.
```
cajubelt, line 11: this looks fine, but would a patch work for this instead? Seems like it might be a reasonable use case for one

src/analytics/ValoraAnalytics.ts
```diff
@@ -109,6 +110,9 @@ class ValoraAnalytics {
         writeKey: SEGMENT_API_KEY,
         storePersistor: AsyncStoragePersistor,
       })
+      await LegacySegment.setup('n/a', {})
```
cajubelt, line 118: nit: could be clearer what the n/a replaces

cajubelt:
```suggestion
      await LegacySegment.setup('n/a', {})  // n/a used in place of write key, which isn't needed because LegacySegment is only used for getAnonymousId
```

cajubelt: btw did you test this without calling setup first? would be nice if we didn't need it

cajubelt (approved):
did you find a way to test that this keeps anonymous id consistent after upgrading from an older version?

src/analytics/ValoraAnalytics.ts
jh2oman: I did test this without setup. doesn't work unfortunately :(

jh2oman:
@cajubelt Yes I tested upgrading from older versions to this version. Still maintains the same anonymousId!

PR #3688, opened 2023-04-28
fix: unable to navigate to some screens after entering PIN on iOS

jeanregisser:
### Description

The PIN entry screen is displayed with the native `modal` presentation style on iOS.

The problem is that in the force backup flow, after the PIN is entered, it navigates to the `AccountKeyScreen` which is in the same native stack but since that screen isn't itself displayed with the native `modal` style, it is inserted below the modal.

Though maybe surprising as we'd expect the modal to be dismissed in that case, this is how native iOS navigation behaves with React Navigation.

To prevent similar issues from happening in the future, I've moved all screens that don't need the native `modal` presentation to `MainStack`. Navigating to a screen within that stack always closes the top level modal. 

Also all screens in the `ModalStack` now don't need to manually specify the `modal` presentation style in their navigation options. This way we don't inadvertently add a screen there without the `modal` style. Avoiding new instances of that bug.

One small thing to note is that `RegulatoryTerms` and `LanguageModal` screens now use the native `modal` presentation style on iOS (with the stacking effect, instead of being full screen), like the PIN screen. Since I haven't moved them to `MainStack` as I think they belong in the `ModalStack`.

### Test plan

Tested manually. We don't yet have an easy way to test the force backup flow from the e2e test. But maybe we should work on that?

**before**

https://user-images.githubusercontent.com/57791/235169348-5a145417-d354-443d-9e79-0ec8df70c802.mp4

**after**

https://user-images.githubusercontent.com/57791/235169389-54992693-5282-4db5-8f5d-95b5550ec41e.mp4

### Related issues

- Fixes ACT-752

### Backwards compatibility

Yes


cajubelt (approved):
nice find-- thank you! any testing needed on the other affected screens?

jophish (approved):
:tada:

jeanregisser:
I did more manual testing with the affected screens. Seems ok so far.
Merging it.

PR #3657, opened 2023-04-16
fix(deps): downgrade dependency react-native-fast-crypto to ^2.0.0

MuckT:
### Description

Reverts the changes #3366 due to failing restores on Android 6 & 7. 
More details on the issue with react-native-fast-crypto can be found [here](https://github.com/EdgeApp/react-native-fast-crypto/issues/52).

### Test plan

Tested locally on Android 6.

### Related issues

N/A

### Backwards compatibility

Yes

PR #3689, opened 2023-04-28
test(e2e): run on android sdk 24

MuckT:
### Description
[Android API Levels](https://developer.android.com/tools/releases/platforms)

- Runs Android E2E tests on Android 7.0 (API level 24) and Android 11.0 (API level 30) on merges to main.
- Runs Android E2E tests on Android 11.0 (API level 30) on pull requests.

### Test plan

Tested in CI.

### Related issues

- Fixes RET-727
- Related #3657
### Backwards compatibility

N/A


kathaypacific (commented):
we only have 4 android machines, now 3 of them will be occupied running each api level after a PR is merged right? which is pretty annoying for other open PRs. it's not a very scalable pattern either, once we add one more version in here then the whole pipeline shuts down after a PR is merged.

in a perfect world, we'd have lots of CI machines and run tests on all api levels. but since we don't, should we be more strategic about which api levels we run on? i was thinking that the chances of an issue affecting 29 but not 24 or 30 is pretty slim. should we continue running the tests on 29, given the resource constraints or just replace it with the lowest version?

OR another option is to run a separate nightly job against all api levels that we care about, once a day during a dead time. wdyt?

also slightly related,`21 is failing 9 specs` - this is the extra effort i was alluding to for fully supporting older api levels. it's not as straightforward as just enabling the tests...we need to fix the app 😅 we can create tasks for this

MuckT:
@kathaypacific we have 5 android machines currently after a PR is merged 3 of them will be occupied by the checks leaving two open. Only one set of these checks should be in the queue at any given time due to the concurrency we have.

kathaypacific:
> @kathaypacific we have 5 android machines currently after a PR is merged 3 of them will be occupied by the checks leaving two open. Only one set of these checks should be in the queue at any given time due to the concurrency we have. These checks on main already run on a daily [action](https://github.com/valora-inc/wallet/actions/runs/4846271020). So we should be fine to add this check to the existing ones, but if we want to add more we should rethink how / what tests are running.

yes but this isn't a very scalable approach. like yes, we added a 3rd api level that we want to run tests on and it's fine because we have 5 machines. but what happens when we add in a 4th and 5th and 6th? i am wondering if we should only run on the highest and lowest api levels after every merge to main, because that would catch most bugs and slow down the CI the least. and nightly, we could have a more comprehensive set of tests that run against all api versions that we care about? 

ideally we'd want to catch all bugs before they're merged. but since we're running these tests against main anyway and the feature is already merged, i think it makes little difference whether the bug is caught in the CI within the hour (in the workflow that is triggered by merge) or within the day (by the scheduled workflow). i prefer to have the CI running more quickly during working hours, i think the tradeoff here is worth it but that's my personal opinion. wdyt? @jeanregisser wbu?

jeanregisser:
Given the constraints, I like what you're proposing @kathaypacific:
- running the lowest and highest supported API versions on all PRs
- running more versions on nightly/daily workflow to catch intermediary incompatibilities

MuckT:
@jeanregisser @kathaypacific - I've adjusted the behavior so that we run PR's on 29 and main workflow on 24 & 30.

kathaypacific (approved):
sorry i'm being annoying with all these comments about such a small change 😅 - i think of the PR CI being more like the real gatekeeper, where we would want to run the tests we really care about before allowing it to be merged. the tests that run on main are almost less priority, since by that point things have been merged already and we should have fairly high confidence that nothing has broken. so along this train of thought (if you agree) i would expect that we run the PRs against the android OS that we care most about. this is usually the most popular one that most users are on, it could well be 29 but it might be 30?

anyway, as it is i think the coverage is there so i've approved but would have a slight preference for running against 30 in PRs and 24, 29 on main as a temporary step. eventually it'd be nice to have separate workflows, one "light weight" one that runs after PR merge against the highest and lowest API versions we support, and one comprehensive one that runs nightly against all API versions we support (or just have enough build machines where running a comprehensive set of tests doesn't create a bottleneck, but i think this approach generally doesn't scale well...)

jeanregisser:
I still strongly think there shouldn't be differences between main and PRs checks.
Before this change we had at least overlap. Now it's different APIs versions with no overlap.
This isn't predictible and will cause issues.

MuckT:
@jeanregisser @kathaypacific - we could run sdk 24 and sdk 30 on every PR, And Run 24, 29, 30 on merges to main with a concurrency of 2. The biggest downside I see with this is on merge days we'll likely be using all the runners. I would like to get coverage for sdk 24 somewhere in our pipeline.

MuckT:
Latest changes will run Android SDK 30 on pull requests to main and Android SDK 24 & 30 on merges to main.

jeanregisser (approved):
Thanks, looking great like this! 👍

kathaypacific (approved):

PR #3690, opened 2023-04-28
feat: send recipient as e.164 (if possible) to resolveId endpoint

silasbw:
This helps for cases where typing in a local number (and eliding the country code).

### DIFF
--- a/src/recipients/RecipientPicker.tsx
+++ b/src/recipients/RecipientPicker.tsx
@@ -1,6 +1,7 @@
 import { parsePhoneNumber } from '@celo/phone-utils'
 import { isValidAddress } from '@celo/utils/lib/address'
 import { NameResolution, ResolutionKind } from '@valora/resolve-kit'
+import { debounce } from 'lodash'
 import * as React from 'react'
 import { useState } from 'react'
 import { useAsync } from 'react-async-hook'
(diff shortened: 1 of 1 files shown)


src/recipients/RecipientPicker.tsx
```diff
@@ -1,6 +1,7 @@
 import { parsePhoneNumber } from '@celo/phone-utils'
 import { isValidAddress } from '@celo/utils/lib/address'
 import { NameResolution, ResolutionKind } from '@valora/resolve-kit'
+import { debounce } from 'lodash'
```
silasbw, line 4: I'm not sure what happened with these re-orderings.  I'm trying to transition from emacs to vscode and maybe this is some vscode thing I did by accident?

satish-ravi: This is because of this setting https://github.com/valora-inc/wallet/blob/696fe1df4d3233fe0d54d712d09525682ea2b408/.vscode/settings.json#L42
This has been annoying me for a while, we should either remove this or make this part of lint / prettier.

satish-ravi (approved):

src/recipients/RecipientPicker.tsx
silasbw: oh, i'll take a pass at updating eslint and/or prettier.

PR #3691, opened 2023-04-28
feat: in app review - redux and saga setup

MuckT:
### Description

Adds `inAppReviewLastInteractionTimestamp` to the app reducer.

- Successful Send

### Screenshot
<p align="center">

</p>

### Test plan

- Android Testing seems like it will require this to be merged and tested on the nightly build: [guidelines](https://github.com/MinaSamir11/react-native-in-app-review/blob/master/README.md#-android-guidlelines-and-notes).
- Tested locally on iOS on Alfajores.

### Related issues

- Fixes RET-15

### Backwards compatibility

Yes


src/appReview/actions.ts
```diff
@@ -0,0 +1,28 @@
+export enum Actions {
```
kathaypacific, line 1: why do you need an actions file when using redux toolkit? actions can be declared in the slice, i see you already declared some here `export const { inAppRatingSupported, setAppRated } = slice.actions`...so can we remove this file?

you can declare actions that don't introduce state change in the slice by just returning the state. https://github.com/valora-inc/wallet/blob/33cbe56c15c39a8a343f45fe3756908ab19cc129/src/consumerIncentives/saga.ts#L253 here is an example of triggering a saga from an action exported from a slice

src/appReview/saga.ts
```diff
@@ -0,0 +1,18 @@
...
+  yield call(initializeAppReview)
+}
+
+export function* setAppReview() {
+  const result = yield call(InAppReview.RequestInAppReview)
+  yield call(updateAppReview, result)
+}
+
+export function* appReviewSaga() {
+  yield takeLatest(Actions.INITIALIZE_APP_REVIEW, safely(initAppReview))
```
kathaypacific, line 16: i'm lost here. what do you want to do? it seems like you are taking the latest INITIALIZE_APP_REVIEW action, triggering the initAppReview saga which just dispatches the INITIALIZE_APP_REVIEW again? am i reading this right?

(by the way, we should use `call` to call functions, if you want to dispatch an action then we should use `put` instead)

src/appReview/slice.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export const slice = createSlice({
+  name: 'appReview',
+  initialState,
+  reducers: {
+    inAppRatingSupported: (state, action) => {
+      state.inAppRatingSupported = action.payload
+    },
+    setAppRated: (state, action) => {
```
kathaypacific, line 27: is this used somewhere?

src/appReview/saga.ts
```diff
@@ -0,0 +1,18 @@
...
+  yield takeLatest(Actions.UPDATE_APP_REVIEW, safely(setAppReview))
+}
```
kathaypacific, line 18: maybe i am missing something here. i thought the logic would be something like on app start, spawn a single saga that checks the previous time the user has been prompted in redux. if the time is null or above a threshold, display the review popup and update the prompted time in redux.

i guess i'm confused about this `appReviewSaga` and the actions. i was expecting just one saga that checks this, and one action for updating the timestamp (whether the user rated or not). why do we need  UPDATE_APP_REVIEW and INITIALIZE_APP_REVIEW?

src/appReview/slice.ts
```diff
@@ -0,0 +1,42 @@
...
+      state.appRated = action.payload
+      state.lastInteractionTimestamp = +new Date()
```
kathaypacific, line 29: i think we can do `Date.now()`?

kathaypacific (commented):
should we also add some analytics for if the user proceeds to rate or dismisses the dialog?

src/appReview/slice.ts
MuckT: This isn't use anywhere yet but before we dispatch a app rating action we'll want to check that in app rating is supported on the device via selector.

src/appReview/saga.ts
MuckT: > logic would be something like on app start, spawn a single saga that checks the previous time the user has been prompted in redux.

I could have this wrong, but we wait to initialize the app review saga until the user has completed onboarding.
Then after the user has completed a specified event, such as sending a transaction, we want to dispatch the app rating. This should happen only if the device supports app rating, the user hasn't rated previously, and they haven't seen the rating in n amount of time.

MuckT:
@kathaypacific I'll add the analytics in a PR branched off this one.

src/appReview/slice.ts
```diff
@@ -0,0 +1,42 @@
...
+      state.appRated = action.payload
+      state.lastInteractionTimestamp = Date.now()
+    },
+  },
+  extraReducers: (builder) => {
+    builder.addCase(REHYDRATE, (state, action: RehydrateAction) => ({
+      ...state,
+      ...getRehydratePayload(action, 'appReview'),
+    }))
+  },
```
jeanregisser, line 37: If there's no custom rehydration logic needed we can remove this. By default all reducers are rehydrated from persisted state at launch.

src/appReview/slice.ts
```diff
@@ -0,0 +1,42 @@
...
+export interface State {
+  inAppRatingSupported: boolean
+  appRated: boolean
+  lastInteractionTimestamp: number | null
+}
+
+const initialState: State = {
+  // Android version >= 21 and iOS >= 10.3
+  // https://github.com/MinaSamir11/react-native-in-app-review#usage
+  inAppRatingSupported: !!InAppReview.isAvailable,
```
jeanregisser, line 15: Why do we want to store this in redux?

MuckT: I opted to keep it in redux so we can do a quick return in the other functions if the in app review isn't available.

src/appReview/slice.ts
MuckT: Removed in a2d40f9bde8a242ba509db0333da7b149accad20!

src/appReview/saga.ts
MuckT: I've adjusted the PR a bit since this last review. Instead we are now listening to the action `Web3Actions.SET_ACCOUNT` for `initAppReview` and `SendActions.SEND_PAYMENT_SUCCESS` for `setAppReview`.

src/appReview/actions.ts
MuckT: Should be fixed now!

src/appReview/saga.ts
MuckT: So one fun thing I learned in implementing this is promise `InAppReview.RequestInAppReview()` doesn't actually return if the use rated or not. Just the fact that it was displayed.

src/appReview/slice.ts
kathaypacific: i think removing `initialized` and `inAppRatingSupported` from redux will simplify the logic, because then you'll also no longer need the `initAppReview` saga.

we'd only ever want to prompt someone to review the app if they've done something in the app before (i.e. dispatched some action, that you cannot dispatch until the wallet has been created already). am i missing a genuine scenario where the rating popup could be triggered during onboarding? 

i think the only variable that you need to persist in redux is the last app review interaction time + whether the user has already rated. on certain actions, the app rating saga can be spawned. this app rating saga will return early if 1) the OS level requirement isn't met, 2) if the user has rated before, 3) if the last interaction time doesn't meet the threshold. if none of these requirements are met, then the saga proceeds to launch the popup and record the last interaction time as the current time.

this app review saga should be triggered so rarely that you don't save much by checking the OS level and storing it in redux. whoever needs this information can do a real time check since it's a synchronous and fast check.  (i also think storing the `inAppRatingSupported` in redux opens a can of worms, since users can update their OS's. so you'd need to watch for that, which seems overcomplicated)

src/appReview/selectors.ts
```diff
@@ -0,0 +1,3 @@
+import { RootState } from 'src/redux/reducers'
+
+export const appReviewSelector = (state: RootState) => state.appReview
```
kathaypacific, line 3: why do we not have more specific selectors? components using `appReviewSelector` now will receive an update any time anything in this whole `appReview` object, which seems like a weird thing to want. it also doesn't fit the pattern we have in other selector files?

src/appReview/saga.ts
```diff
@@ -0,0 +1,47 @@
+import InAppReview from 'react-native-in-app-review'
+import { call, put, select, spawn, takeLatest, takeLeading } from 'redux-saga/effects'
+import { appReviewSelector } from 'src/appReview/selectors'
+import { setInitialized, setLastInteractionTimestamp } from 'src/appReview/slice'
+import { Actions as SendActions } from 'src/send/actions'
+import { safely } from 'src/utils/safely'
+import { ONE_DAY_IN_MILLIS } from 'src/utils/time'
+import { Actions as Web3Actions } from 'src/web3/actions'
+
+export function* initAppReview() {
```
kathaypacific, line 10: i elaborated on this in my other comment but i still don't understand the value of this saga, can you explain what fails if we delete this?

src/appReview/slice.ts
jeanregisser: +1 to what Kathy said.

We should only store the minimum necessary things in Redux.

MuckT:
@kathaypacific @jeanregisser - thanks for the great feedback! I've changed the pr so only `lastInteractionTimestamp` is stored in redux. I've also removed `initAppReview` and changed the selectors to return a primitive. There's a feature gate PR that fixes the E2E tests which might need to get merged into this first: #3704.

src/appReview/saga.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export function* setAppReview() {
+  // Quick return if the device does not support in app review
+  if (!InAppReview.isAvailable()) return
+
+  const lastInteractionTimestamp = yield select(lastInteractionTimestampSelector)
+  const now = Date.now()
+
+  // If the last interaction was less than a quarter year ago or null
+  if (lastInteractionTimestamp <= now - ONE_DAY_IN_MILLIS * 91) {
```
kathaypacific, line 17: nit: it'd be nice to have the threshold as a constant at the top of the file so that it's easy to find / update. so then you could have `if (now - lastInteractionTimestamp >= REVIEW_INTERVAL)`?

also, i think if apple only allows 3 review per year then the interval should be 122 days?

src/appReview/saga.ts
```diff
@@ -0,0 +1,31 @@
...
+import { call, put, select, spawn, takeLatest } from 'redux-saga/effects'
+import { lastInteractionTimestampSelector } from 'src/appReview/selectors'
+import { setLastInteractionTimestamp } from 'src/appReview/slice'
+import { Actions as SendActions } from 'src/send/actions'
+import { safely } from 'src/utils/safely'
+import { ONE_DAY_IN_MILLIS } from 'src/utils/time'
+
+export function* setAppReview() {
+  // Quick return if the device does not support in app review
+  if (!InAppReview.isAvailable()) return
```
kathaypacific, line 11: do you also want to include an early return if there is no wallet address? so that it is never displayed during onboarding?

src/appReview/slice.ts
```diff
@@ -0,0 +1,23 @@
+import { createSlice } from '@reduxjs/toolkit'
+
+export interface State {
+  lastInteractionTimestamp: number | null
```
kathaypacific, line 4: ah, now that there is only one variable i am wondering if we should have thrown this in the `app` reducer instead as something like "lastAppReviewPromptTimestamp"...in general i think small specific redux stores are fine, but i am struggling to think of any reasons why we'd add to this store in the future. "appReview" is almost too specific, so i'm not sure whether we should keep this. wdyt?

src/appReview/saga.ts
```diff
@@ -0,0 +1,31 @@
+import InAppReview from 'react-native-in-app-review'
+import { call, put, select, spawn, takeLatest } from 'redux-saga/effects'
+import { lastInteractionTimestampSelector } from 'src/appReview/selectors'
+import { setLastInteractionTimestamp } from 'src/appReview/slice'
+import { Actions as SendActions } from 'src/send/actions'
+import { safely } from 'src/utils/safely'
+import { ONE_DAY_IN_MILLIS } from 'src/utils/time'
+
+export function* setAppReview() {
```
kathaypacific, line 9: `requestInAppReview`? `promptAppStoreReview`? `askUserToReviewApp` (lol)?

kathaypacific (commented):
looks 🔥 

just tiny comments from me, although i do think we should think about whether it still makes sense to have a new redux slice here

src/appReview/slice.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const initialState: State = {
+  lastInteractionTimestamp: null,
+}
+
+export const slice = createSlice({
+  name: 'appReview',
+  initialState,
+  reducers: {
+    setLastInteractionTimestamp: (state, action) => {
```
jeanregisser, line 15: I know the codebase is not following this in a bunch of places, but it's better to model actions as events instead of setters.

Maybe?
```suggestion
   inAppReviewShown: (state, action) => {
```

src/appReview/slice.ts
MuckT: Back when it was three variables a small redux store was great, but now that we just have one it really does seem like a bit much. 🤔

src/appReview/slice.ts
MuckT: Fixed in commits/cfda34c959aa53f7095cdd15007b69e8b72be60e!

src/appReview/saga.ts
MuckT: Fixed in 52567a4d9d85377436362a213040ac010ba97813!

src/appReview/saga.ts
MuckT: Fixed in 12baef6524c0438ddbd1a4ec1f1c31c08be785df and tested in f88199429566a04db42eff25c6fdc58f95c8e298!

src/appReview/saga.ts
MuckT: App Store policy governs the actual display of a rating and review request view from [`requestReview(in:)`](https://developer.apple.com/documentation/storekit/skstorereviewcontroller/3566727-requestreview#discussion) [used by react-native-in-app-review](https://github.com/MinaSamir11/react-native-in-app-review#system-rating-and-review-prompts). Android seems to be [limited to about once per month](https://developer.android.com/guide/playcore/in-app-review#:~:text=Play%20Store%20mechanism.-,Quotas,not%20always%20display%20a%20dialog.), but can change anytime. As such once a quarter feels pretty unintrusive considering behavior I've seen in other apps.

MuckT: Adjusted the style of the check in abdc93177a7e5e509055a5f6d7d173109a639379!

src/appReview/slice.ts
MuckT: @kathaypacific to move this into the app reducer.

MuckT: Move to the app reducer in eaa797d8a4cc460f2f98a0d8dc1439e11a01ad59! The version with the slice, 668caf73ecf3146f71694de3177dc11b3cf5c523, is also passing CI.

src/app/saga.test.ts
```diff
@@ -62,12 +69,20 @@ import {
...
 jest.mock('src/analytics/ValoraAnalytics')
 jest.mock('src/sentry/Sentry')
 jest.mock('src/sentry/SentryTransactionHub')
+jest.mock('src/statsig')
+jest.mock('react-native-in-app-review', () => ({
+  RequestInAppReview: jest.fn(),
+  isAvailable: () => mockIsAvailable(),
+}))
+
+const mockIsAvailable = jest.fn()
```
kathaypacific, line 85:
```suggestion
const mockIsInAppReviewAvailable = jest.fn()
```

src/app/saga.test.ts
```diff
@@ -583,3 +598,81 @@ describe('appInit', () => {
...
+    ${oneDayAgo}             | ${false}    | ${'1 day ago'}   | ${true}     | ${null}
+    ${oneDayAgo}             | ${true}     | ${'1 day ago'}   | ${false}    | ${null}
+    ${null}                  | ${false}    | ${null}          | ${false}    | ${null}
+    ${oneQuarterAgo}         | ${false}    | ${'92 days ago'} | ${false}    | ${null}
+    ${oneDayAgo}             | ${false}    | ${'1 day ago'}   | ${false}    | ${null}
+  `(
+    `Should not show when Device Available: $isAvailable, Feature Gate: $featureGate, Last Interaction: $lastInteraction and Wallet Address: $walletAddress`,
+    async ({ lastInteractionTimestamp, isAvailable, featureGate, walletAddress }) => {
+      // Clear previous calls
+      jest.clearAllMocks()
```
kathaypacific, line 658: this is also not needed

src/app/saga.test.ts
```diff
@@ -583,3 +598,81 @@ describe('appInit', () => {
...
+      mockIsAvailable.mockReturnValue(isAvailable)
+
+      await expectSaga(requestInAppReview)
+        .withState(
+          createMockStore({
+            web3: { account: walletAddress },
+          }).getState()
+        )
+        .provide([[select(inAppReviewLastInteractionTimestampSelector), lastInteractionTimestamp]])
+        .dispatch({
```
kathaypacific, line 669: this is also not needed

src/app/saga.ts
```diff
@@ -106,6 +113,9 @@ const TAG = 'app/saga'
 //    (which will put an app into `background` state until dialog disappears).
 const DO_NOT_LOCK_PERIOD = 30000 // 30 sec
 
+// REVIEW_INTERVAL is the time between app review prompts
+const REVIEW_INTERVAL = ONE_DAY_IN_MILLIS * 91 // 91 days
```
kathaypacific, line 117: what's the benefit of having a 90 day window here? android users potentially get this prompt 4 times per year, but iOS users will get this on month 0, 3, 6, and then nothing for 6 months? why not have 120 days or so, so that the prompt is evenly spaced?

src/app/saga.test.ts
```diff
@@ -583,3 +598,81 @@ describe('appInit', () => {
...
+      mocked(getFeatureGate).mockReturnValue(true)
+      mockIsAvailable.mockReturnValue(true)
+      await expectSaga(requestInAppReview)
+        .withState(
+          createMockStore({
+            web3: { account: '0xTest' },
+          }).getState()
+        )
+        .provide([[select(inAppReviewLastInteractionTimestampSelector), lastInteractionTimestamp]])
+        .dispatch({
```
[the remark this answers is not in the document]
kathaypacific: instead, i think you can add an assertion that the interaction time was updated, with a `.put(...)` here

src/app/saga.ts
MuckT: Changed to 120 days in 6a154e6f519154f38ba1c3dfb3d77540ff961fa4!

src/app/actions.ts
```diff
@@ -157,6 +158,11 @@ export interface PushNotificationsPermissionChanged {
   enabled: boolean
 }
 
+export interface InAppReviewCalled {
```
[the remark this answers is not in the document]
MuckT: Fixed in a78eb53756c26ba562058b72fde52ac31743d135!

src/app/saga.test.ts
```diff
@@ -589,3 +601,72 @@ describe('appInit', () => {
...
+      await expectSaga(requestInAppReview, inAppReviewRequested(now))
+        .withState(
+          createMockStore({
+            web3: { account: '0xTest' },
+          }).getState()
+        )
+        .provide([[select(inAppReviewLastInteractionTimestampSelector), lastInteractionTimestamp]])
+        .run()
+
+      expect(InAppReview.RequestInAppReview).toHaveBeenCalledTimes(1)
```
kathaypacific, line 631: i may have commented in the other PR but i think it's important to check that the interaction timestamp is also updated here by checking that the correct action is dispatched. you can add a `.put()` before the `.run()` to do this :)

src/app/saga.test.ts
```diff
@@ -589,3 +601,72 @@ describe('appInit', () => {
...
+    ${fourMonthsAndADayAgo}  | ${false}    | ${'121 days ago'} | ${true}     | ${null}
+    ${oneDayAgo}             | ${false}    | ${'1 day ago'}    | ${true}     | ${null}
+    ${oneDayAgo}             | ${true}     | ${'1 day ago'}    | ${false}    | ${null}
+    ${null}                  | ${false}    | ${null}           | ${false}    | ${null}
+    ${fourMonthsAndADayAgo}  | ${false}    | ${'121 days ago'} | ${false}    | ${null}
+    ${oneDayAgo}             | ${false}    | ${'1 day ago'}    | ${false}    | ${null}
+  `(
+    `Should not show when Device Available: $isAvailable, Feature Gate: $featureGate, Last Interaction: $lastInteraction and Wallet Address: $walletAddress`,
+    async ({ lastInteractionTimestamp, isAvailable, featureGate, walletAddress }) => {
+      // Clear previous calls
```
kathaypacific, line 656: nit: i don't think this comment makes sense...do we need a comment..?

kathaypacific (approved):
LGTM! just a couple of nits on the tests that would be nice to discuss / address before we merge

src/app/saga.test.ts
MuckT: I'll give this another shot today. I was trying something similar yesterday to bump up the test coverage.

MuckT: Took a trying to add a `.put` before the `.run` based on the examples of sagas which have tests and it's not working. I've taken a look at the [codecov](https://app.codecov.io/gh/valora-inc/wallet/blob/tomm%2Fin-app-review/src/app/saga.ts) for `src/app/saga.ts` and still can't get it to work.

kathaypacific: i created a PR to show how it can work https://github.com/valora-inc/wallet/pull/3729

there were 2 things that this test was missing:
- not locking the date - since the action relies on `Date.now`, we need this value to be predictable
- the `RequestInAppReview` function is an async one that returns a value. you were incorrectly mocking this as `jest.fn()` which returns nothing, so the action was never dispatched

src/app/saga.test.ts
```diff
@@ -589,3 +601,72 @@ describe('appInit', () => {
...
+    ${oneDayAgo}             | ${false}    | ${'1 day ago'}    | ${false}    | ${'0xTest'}
+    ${oneDayAgo}             | ${true}     | ${'1 day ago'}    | ${true}     | ${null}
+    ${null}                  | ${false}    | ${null}           | ${true}     | ${null}
+    ${fourMonthsAndADayAgo}  | ${false}    | ${'121 days ago'} | ${true}     | ${null}
+    ${oneDayAgo}             | ${false}    | ${'1 day ago'}    | ${true}     | ${null}
+    ${oneDayAgo}             | ${true}     | ${'1 day ago'}    | ${false}    | ${null}
+    ${null}                  | ${false}    | ${null}           | ${false}    | ${null}
+    ${fourMonthsAndADayAgo}  | ${false}    | ${'121 days ago'} | ${false}    | ${null}
+    ${oneDayAgo}             | ${false}    | ${'1 day ago'}    | ${false}    | ${null}
+  `(
```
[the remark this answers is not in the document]
MuckT: Cleaned up the tests in cd194edeaee61b9baa6eb39dc3d8a0431750d623.

PR #3692, opened 2023-04-29
fix: adjust styles for no search result icon and text

MuckT:
### Description

Styles for the info icon in the no results search section were cluttered together. This uses a flex of 1 and 11 to create a 12 grid layout.

Before iOS / After iOS
<p align="center">

</p>

### Test plan

Tested locally in Android an iOS.

### Related issues

N/A

### Backwards compatibility

Yes


src/dappsExplorer/search/NoResults.tsx
```diff
@@ -17,18 +17,24 @@ function NoResults({ testID, searchTerm }: Props) {
...
   )
 }
 
 const styles = StyleSheet.create({
   iconContainer: {
     flex: 1,
+    alignSelf: 'flex-start',
+  },
+  searchTextContainer: {
+    flex: 11,
```
kathaypacific, line 37: if it's possible, i think it's nice to avoid flex ratios like this. 1:11 is pretty unintuitive and makes it more difficult to add new sibling components. could you achieve the same effect by having:
- viewContainer flex 1
- iconContainer no flex style declared, margin right 4 (or something)
- searchTextContainer flex 1
- remove the flex 10 on the `text` style

MuckT: This doesn't work with the wrap on long search results. I don't think the flex of 1:11 making a 12 column layout is too unintuitive, but I am happy to add a comment for what's going on.

<p align="center">

</p>

kathaypacific: i forgot to mention above, you will need to add a flexwrap on the text component to get it to wrap. the view container itself shouldn't wrap (i think that's what you're seeing there - the right hand side component gets big, but the parent says that it can wrap, so it became vertically aligned)

the following looks okay for me, maybe you want to try?
```
conContainer: {
    marginRight: 8, // or whatever this is supposed to be
  },
  searchTextContainer: {
    flex: 1,
  },
  searchedText: {
    color: Colors.dark,
    fontWeight: 'bold',
  },
  text: {
    ...fontStyles.xsmall,
    textAlignVertical: 'center',
    flexWrap: 'wrap',
  },
  viewContainer: {
    flex: 1,
    justifyContent: 'flex-start',
    flexDirection: 'row',
    gap: Spacing.Thick24,
    marginTop: Spacing.Thick24,
  },
```

kathaypacific: having said this, i think that the justifyContent and gap styles should be removed from `viewContainer`. flex-start is already the default style for justifyContent and gap doesn't work on the RN version we are using?

src/dappsExplorer/search/NoResults.tsx
```diff
@@ -17,17 +17,23 @@ function NoResults({ testID, searchTerm }: Props) {
...
+            <Text style={styles.searchedText} />
+          </Trans>
+        </Text>
+      </View>
     </View>
   )
 }
 
 const styles = StyleSheet.create({
+  /* Using a 1:11 flex on two elements will create a pseudo 12 column layout */
```
kathaypacific, line 32: lol do we need this comment? 🤣 we're not using flex 11 anymore, so it might be confusing why we need to know this

kathaypacific (approved):
👯‍♂️

PR #3698, opened 2023-05-03
chore: upgrade react native to 0.68

kathaypacific:
### Description

RN 68 contains some big changes for enabling the new architecture. There are a bunch of new files added to the android project, but these are all for the new architecture which this PR does not enable. They were copy + pasted directly from the upgrade helper with no additional changes (except to add in our app name).

Other changes to note:

### Test plan

- CI still runs
- Debug builds work locally
- Flipper still works

### Related issues

- Fixes RET-683

### Backwards compatibility

Y


jeanregisser (approved):
Awesome 🚀

PR #3699, opened 2023-05-03
feat: fetch and store positions

jeanregisser:
### Description

Fetch and store positions from plugins, if the `show_positions` feature gate is enabled.

Next up:
- show positions
- update total balance

### Test plan

- Added tests

### Related issues

- Fixes RET-723

### Backwards compatibility

Yes


kathaypacific (approved):
✨ looks great!!!!

PR #3700, opened 2023-05-03
fix(firebase): fix firebase messaging initialization

jh2oman:
### Description

fixes this issue on the nightly build https://valora-app.slack.com/archives/C04B61SJ6DS/p1683015075041129

### Test plan

Verified manually in the logs that firebase was successfully initialized on first app load and app reloads.

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

Yes

### DIFF
--- a/firebase.json
+++ b/firebase.json
@@ -1,5 +1,0 @@
-{
-  "react-native": {
-    "messaging_ios_auto_register_for_remote_messages": false
-  }
-}
--- a/src/account/saga.test.ts
+++ b/src/account/saga.test.ts
@@ -38,7 +38,6 @@ jest.mock('@react-native-firebase/app', () => ({
   app: jest.fn(() => ({
     messaging: () => ({
       getToken: jest.fn().mockResolvedValue('someToken'),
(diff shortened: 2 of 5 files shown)


firebase.json
```diff
@@ -1,5 +0,0 @@
-{
-  "react-native": {
-    "messaging_ios_auto_register_for_remote_messages": false
```
jh2oman, line 3: @jeanregisser mentioned on my last PR that doing this would move the Native prompt to Allow Notifications to the initial app opening, but I found that removing this file had no impact. The user is still prompted on the Home screen.

jeanregisser: Great!
Thanks yes I see the cloud messaging init waits for the [`VISIT_HOME`](https://github.com/valora-inc/wallet/blob/2c1fae9ddf2d4a16dc25bc11334fca7747574b1c/src/firebase/firebase.ts#L167) action now.
All good, sorry to have misled you.

jeanregisser (approved):
🙌

PR #3702, opened 2023-05-04
feat: show positions in the assets screen

jeanregisser:
### Description

This shows the positions in the assets screen.

This is only temporary until we have finalized designs.

### Test plan

- No tests added yet since this is throwaway code (mostly copied from the [poc](https://github.com/valora-inc/wallet/tree/jeanregisser/add-positions-poc))

 

### Related issues

- Fixes RET-723

### Backwards compatibility

Yes

### DIFF
--- a/src/dapps/saga.ts
+++ b/src/dapps/saga.ts
@@ -67,7 +67,7 @@ export function* handleOpenDapp(action: PayloadAction<DappSelectedAction>) {
 export function* handleFetchDappsList() {
   const dappsListApiUrl = yield select(dappsListApiUrlSelector)
   if (!dappsListApiUrl) {
-    Logger.warn(TAG, 'dappsListApiUrl not found')
+    Logger.warn(TAG, 'dappsListApiUrl not found, skipping dapps list fetch')
     return
   }
 
@@ -75,7 +75,7 @@ export function* handleFetchDappsList() {
   if (!address) {
     // the dapplist is fetched on app start, but for a new user who has not yet
     // created or restored a wallet the request will fail.
-    Logger.debug(TAG, 'Wallet address not found')
+    Logger.debug(TAG, 'Wallet address not found, skipping dapps list fetch')
     return
   }
 


src/positions/Positions.tsx
```diff
@@ -0,0 +1,147 @@
...
+  const balanceUsd =
+    position.type === 'contract-position'
+      ? new BigNumber(position.balanceUsd)
+      : new BigNumber(position.balance).multipliedBy(position.priceUsd)
+
+  return (
+    <View style={styles.tokenContainer}>
+      <View style={styles.row}>
+        <View style={{ flexDirection: 'row-reverse' }}>
+          {baseTokenImages.reverse().map((image, index) => (
```
dievazqu, line 48: Just curious, why do we need the `reverse()` here?

dievazqu (approved):
LGTM!

src/positions/Positions.tsx
jeanregisser: It's needed (along with `row-reverse`) because I wanted the first image on the left to be on top of the images on the right, with the overlap:
Screenshot 2023-05-05 at 12 20 09

By default for displaying `[MOO, CELO]`, with the default `row` layout we'd have:
Screenshot 2023-05-05 at 12 24 01
i.e. the images drawn last are displayed on top.
`CELO` is last so displayed on top of `MOO`

If we apply `row-reverse` we get this:
Screenshot 2023-05-05 at 12 24 38
Again `CELO` is last so displayed on top of `MOO`, but it started drawing them from the right.

So to get `MOO` on top of `CELO`, with `MOO` on the left, we need to `reverse()` the array.

src/positions/Positions.tsx
```diff
@@ -0,0 +1,147 @@
...
+// Temporary component to display positions until we have finalized designs
+export default function Positions() {
+  const positions = useSelector(positionsSelector)
+  const status = useSelector(positionsStatusSelector)
+
+  if (!getFeatureGate({ featureGateName: StatsigFeatureGates.SHOW_POSITIONS })) {
+    return null
+  }
+
+  if (positions.length === 0 && status === 'loading') {
```
kathaypacific, line 96: nit: should we always display the loading indicator at the top of the screen if the positions are loading, like the dapps screen? in case the previously loaded positions are quite outdated, would be nice to give visual feedback that they will be updated

src/positions/Positions.tsx
```diff
@@ -0,0 +1,147 @@
...
+    position.type === 'contract-position' ? undefined : new BigNumber(position.balance)
+  const balanceUsd =
+    position.type === 'contract-position'
+      ? new BigNumber(position.balanceUsd)
+      : new BigNumber(position.balance).multipliedBy(position.priceUsd)
+
+  return (
+    <View style={styles.tokenContainer}>
+      <View style={styles.row}>
+        <View style={{ flexDirection: 'row-reverse' }}>
```
kathaypacific, line 47: nit: use the stylesheet instead of inline? 🙈

kathaypacific (approved):
✨

src/positions/Positions.tsx
jeanregisser: yep 👍

PR #3703, opened 2023-05-04
chore(support): implement a new zendesk API based way to reach support

jh2oman:
### Description

* A new backend for sending support tickets (because the email method is not always reliable). 
* Whole feature is behind a Statsig Feature Gate
* UI needed to be updated slightly (with email and name field) which are required to make the API request.
* The binary files that were changed were all from creating the new ZENDESK_API_KEY secret

Simulator Screen Shot - iPhone 13 - 2023-05-04 at 15 42 43

https://user-images.githubusercontent.com/8432644/236537880-bd1c37f7-a769-47ce-94ed-afa4b43c0b02.mp4

### Test plan

Have tested manually and Lisa has confirmed that everything looks good on her end. We will also test in buggapalooza.
Unit tests

### Related issues

https://linear.app/valora/issue/ACT-660/support-logs-truncated-and-included-in-email-body

### Backwards compatibility

Yes, with feature gate to fallback if things go wrong


src/statsig/index.ts
```diff
@@ -84,7 +84,7 @@ export function getDynamicConfigParams<T extends Record<string, StatsigParameter
   }
 }
 
-export function getFeatureGate({ featureGateName }: { featureGateName: StatsigFeatureGates }) {
+export function getFeatureGate(featureGateName: StatsigFeatureGates) {
```
jh2oman, line 87: realized that since there was only one argument the object spread param was a bit redundant

jh2oman:
@MuckT I did not. I can test that out though!

jh2oman:
@MuckT Uploads worked at 5MB, 10MB, and 20MB

src/account/SupportContact.tsx
```diff
@@ -77,28 +91,40 @@ function SupportContact({ route }: Props) {
...
+          // Get the current months log file to attach as text if sendEmailWithNonNativeApp is used
+          attachments
+            ? attachments.find(
+                (attachment: { name: string }) => attachment.name === Logger.getCurrentLogFileName()
+              )?.path ?? false
+            : false
+        )
+      }
+      // Used to prevent flickering of the activity indicator on quick uploads
+      setTimeout(() => setInProgress(false), 1000)
```
cajubelt, line 122: this isn't really related to the new changes, but it seems like a Promise.all between this and the email would make sense here, so if the email takes more than a second by itself, this doesn't add an unnecessary delay

src/account/zendesk.ts
```diff
@@ -0,0 +1,176 @@
...
+        Authorization: `Basic ${Buffer.from(`${userEmail}/token:${ZENDESK_API_KEY}`).toString(
+          'base64'
+        )}`,
+      },
+      body: blob,
+    }
+  )
+  const resp = await uploadFileResponse.json()
+  if (uploadFileResponse.status >= 400) {
+    Logger.debug(
```
cajubelt, line 167: same here

src/account/zendesk.ts
```diff
@@ -0,0 +1,176 @@
...
+      TAG,
+      '_uploadFile returned response greater than 400',
+      uploadFileResponse.status,
+      JSON.stringify(resp)
+    )
+  }
+  const uploadToken = resp.upload?.token
+  return uploadToken
```
cajubelt, line 175:
```suggestion
  return resp.upload?.token
```
code golf

src/statsig/index.test.ts
```diff
@@ -106,20 +106,16 @@ describe('Statsig helpers', () => {
 
   describe('getFeatureGate', () => {
     it('returns default values if getting statsig feature gate throws error', () => {
-      mocked(Statsig.checkGate).mockImplementation(() => {
```
cajubelt, line 109: I thought `mocked(...)` was the preferred way of doing this

src/account/zendesk.ts
```diff
@@ -0,0 +1,176 @@
...
+        requester: {
+          email: userEmail,
+          name: userName,
+        },
+      },
+    }),
+  })
+  if (resp.status >= 400) {
+    const jsonResponse = await resp.json()
+    Logger.debug(
```
cajubelt, line 138: might want to log an error here

cajubelt (commented):
looks good overall, just left small comments

src/statsig/index.test.ts
jh2oman: weird, i think this was something odd going on with automerge

src/account/SupportContact.tsx
jh2oman: The timeout also has a second use b/c navigateBackAndToast is a bit slow and without the timeout the page briefly flashes the done-loading state before redirecting. I added a comment explaining.

cajubelt (approved):

PR #3705, opened 2023-05-05
feat(app-review): analytics

MuckT:
### Description

Adds the following analytics events for in app review:
- `app_review_impression`
- `app_review_error`

Also wraps `InAppReview.RequestInAppReview` in a try catch with logging on unexpected errors.

### Test plan

Unit tests updated.

### Related issues

- Fixes RET-15 

### Backwards compatibility

Yes


src/appReview/saga.ts
```diff
@@ -18,9 +23,17 @@ export function* setAppReview() {
 
   // If the last interaction was less than a quarter year ago or null
   if (lastInteractionTimestamp <= now - ONE_DAY_IN_MILLIS * 91) {
-    const result = yield call(InAppReview.RequestInAppReview)
-    // If the in app review was shown, update the last interaction timestamp
-    if (result) yield put(setLastInteractionTimestamp(now))
+    try {
+      const result = yield call(InAppReview.RequestInAppReview)
+      // If the in app review was shown, update the last interaction timestamp
+      if (result) {
```
kathaypacific, line 29: this made me wonder what to do when `result` is false...but reading the docs, it seems like `result` is always true for iOS, and true for Android _after_ they finish the review flow. am i understanding that right? if so, it seems like we should remove this if block - we always want to update the timestamp and dispatch the impression event

kathaypacific (commented):
Looks great! just one comment to check that we understand the return value of RequestInAppReview correctly :)

src/appReview/saga.ts
MuckT: It's not quite clear to me from the docs if result is always true for iOS and Android or an error. In [this example](https://github.com/MinaSamir11/react-native-in-app-review#usage) it looks like both was are used, but I also see that the updating timestamp is occurring outside the if block.

MuckT: After looking a bit more I think we're safe to remove the it block as result looks like it's either: `true` or an error.
- [react-native-in-app-review Android](https://github.com/MinaSamir11/react-native-in-app-review/blob/master/android/src/main/java/com/ibits/react_native_in_app_review/AppReviewModule.java#L68-L72)
- [react-native-in-app-review iOS](https://github.com/MinaSamir11/react-native-in-app-review/blob/master/ios/RNInAppReviewIOS.swift#L9C2-L23)

MuckT: Fixed in 3544d91aff763ed658c51a9bffa57f7521f1424d!

src/analytics/Properties.tsx
```diff
@@ -1211,6 +1212,13 @@ interface TokenBottomSheetEventsProperties {
   }
 }
 
+interface InAppReviewEventsProperties {
```
kathaypacific, line 1215: i think that this can be part of AppEventsProperties right?

src/app/saga.test.ts
```diff
@@ -547,61 +548,11 @@ describe('runCentralPhoneVerificationMigration', () => {
...
-      .put(setSupportedBiometryType(BIOMETRY_TYPE.TOUCH_ID))
-      .run()
-
-    expect(initializeSentry).toHaveBeenCalledTimes(1)
-    expect(ValoraAnalytics.init).toHaveBeenCalledTimes(1)
-    expect(initI18n).toHaveBeenCalledWith('en-US', true, '1')
-  })
-})
-
+// Keep this test before the app init test otherwise the error spec will fail
```
kathaypacific, line 551: this is not okay 🙈 we should never have to place unit tests in a specific order. if this is a requirement, it means that there are interdependencies between tests which is a bad thing for a unit test! in this case i think that we have not cleared up the mocks from the app init tests, and that has caused the next tests to fail...

kathaypacific (commented):
looking good :) seeing your comment in the test file got me curious. i think #3716 will enable you to remove the workaround

src/analytics/Properties.tsx
MuckT: Fixed in 24e81d31a615d6f02a97c876293433aee9eae3c6!

PR #3706, opened 2023-05-05
feat(statsig): refresh when app becomes active

satish-ravi:
### Description

Force a statsig refresh when app becomes active by updating the statsig user object with a new custom property, `loadTime`

### Test plan

Manually following the below steps

1. Open app and note variant for any exp (home screen actions control in below example)
2. Change experiment config by forcing a different variant (force home screen actions variant in below example)
3. background and reopen app and check the new variant loads

Also ensured logs don't show an "uninitialized" statsig value.

https://user-images.githubusercontent.com/5062591/236580431-a7474866-1cd9-4a57-8293-1b57ec2e8284.mp4

### Related issues

- Fixes ACT-747

### Backwards compatibility

Yes


src/app/actions.ts
```diff
@@ -36,6 +36,7 @@ export enum Actions {
   INVITE_LINK_CONSUMED = 'APP/INVITE_LINK_CONSUMED',
   HAPTIC_FEEDBACK_SET = 'APP/HAPTIC_FEEDBACK_SET',
   PUSH_NOTIFICATIONS_PERMISSION_CHANGED = 'APP/PUSH_NOTIFICATIONS_PERMISSION_CHANGED',
+  SET_STATSIG_LOAD_TIME = 'APP/SET_STATSIG_LOAD_TIME',
```
jeanregisser, line 39: I know the codebase is not following this in a bunch of places, but it's better to model actions as events instead of setters.

Maybe?
```suggestion
  STATSIG_LOADED = 'APP/STATSIG_LOADED',
```
And we can rename the rest after this.

src/app/saga.ts
```diff
@@ -106,6 +109,8 @@ const TAG = 'app/saga'
 //    (which will put an app into `background` state until dialog disappears).
 const DO_NOT_LOCK_PERIOD = 30000 // 30 sec
 
+const STATSIG_REFRESH_PERIOD = 10000 // 10 sec for testing. TODO change to 1h
```
jeanregisser, line 112: nit: easier to parse visually:
```suggestion
const STATSIG_REFRESH_PERIOD = 10_000 // 10 sec for testing. TODO change to 1h
```

jeanregisser (commented):
Nice! The approach looks good to me 👍

src/app/actions.ts
satish-ravi: Oh I didn't know this. Thanks! I actually ended up removing this state per @kathaypacific's suggestion to refresh always and not manage this state. This was also buggy, it doesn't reflect the correct value if the statsig user is patched with another field. We might end up invoking statsig more often, but this shouldn't cost us. Statsig continues to use previous values while this is refreshing.

src/app/saga.ts
```diff
@@ -415,6 +416,11 @@ export function* handleSetAppState(action: SetAppState) {
   if (requirePinOnAppOpen && isPassedDoNotLockPeriod && isAppActive) {
     yield put(appLock())
   }
+
+  if (isAppActive) {
+    // Force a statsig refresh by updating the user object
+    yield call(patchUpdateStatsigUser)
```
cajubelt, line 422: it seems like patchUpdateStatsigUser was created to work around an assumption we had about how Statsig updateUser was working that we later found to be false-- namely, that updateUser wipes the current user properties (we found this only to be the case when you update userId). I am wondering if we should refactor patchUpdateStatsigUser to make it more single-purpose to updating the userId, and call another function here that is also single-purpose. 

Relatedly, I find it pretty surprising that you can just call patchUpdateStatsigUser here and be done with it. Both that meta-observation and the quite sizable docstring on the function indicate to me that its behavior is rather opaque and non-intuitive at the moment, which further motivate a refactor in my view.

satish-ravi: > we found this only to be the case when you update userId
I don't think this is true unless I am missing something. AFAIK, updateUser requires the full object in all cases. The only difference is that if a userId is updated, it does a full reinitialization, whereas when other properties are updated, statsig returns cached values while this is still resolving, and if we call updateUser with no userId, it assumes a null user id.
But I agree on the method docstring, we could maybe rename it to maybe refreshStatsig and take optional user properties?

cajubelt (approved):
my previous comment was on a mistaken understanding of our findings with Statsig's updateUser. This looks ok, I commented offline about potentially moving the loadTime override out of getDefaultStatsigUser though

src/app/saga.test.ts
```diff
@@ -375,47 +376,52 @@ describe('handleOpenUrl', () => {
...
+    it('does nothing', async () => {
+      await expectSaga(handleSetAppState, setAppState('inactive'))
+        .provide([
+          [select(getLastTimeBackgrounded), 0],
+          [select(getRequirePinOnAppOpen), true],
+        ])
+        .not.put(appLock())
+        .not.call(patchUpdateStatsigUser)
+        .run()
+    })
```
kathaypacific, line 424: ah, these unit tests are such an improvement!!!! ✨

kathaypacific (approved):
awesome!! 🚀

PR #3711, opened 2023-05-05
chore: refactor settings class to functional component

kathaypacific:
### Description

This giant class was causing me all kinds of pains when i tried to implement the revoke phone number task in the sprint. Turning this into a functional component so that we can use hooks and reduce boilerplate.

No actual changes aside from adding an analytics event to track disabling analytics 😅

### Test plan

n/a

### Related issues

- Fixes RET-717 

### Backwards compatibility

Y


src/account/Settings.tsx
```diff
@@ -411,143 +365,139 @@ export class Account extends React.Component<Props, State> {
...
-            {!numberVerified && (
-              <SettingsItemTextValue title={t('confirmNumber')} onPress={this.goToConfirmNumber} />
-            )}
-            <SettingsItemTextValue
-              title={t('languageSettings')}
-              testID="ChangeLanguage"
-              value={currentLanguage?.name ?? t('unknown')}
-              onPress={this.goToLanguageSetting}
-            />
+  const currentLanguage = useSelector(currentLanguageSelector)
```
dievazqu, line 368: nit: Move this to the top of the function where all the selectors are.

dievazqu (approved):
LGTM!
The tests are failing, but I think it's because you have to mock the current language Selector. And maybe update the snapshots?

src/account/Settings.tsx
kathaypacific: ah, thank you!! i forgot about this

src/account/Settings.tsx
```diff
@@ -55,222 +58,167 @@ import {
...
+  const { sessions: walletConnectV2Sessions } = useSelector(selectSessionsV2)
+
+  const { v1, v2 } = useSelector(walletConnectEnabledSelector)
+  const supportedBiometryType = useSelector(supportedBiometryTypeSelector)
+  const shouldShowRecoveryPhraseInSettings = useSelector(shouldShowRecoveryPhraseInSettingsSelector)
+  const hapticFeedbackEnabled = useSelector(hapticFeedbackEnabledSelector)
+  const decentralizedVerificationEnabled = useSelector(decentralizedVerificationEnabledSelector)
+  const currentLanguage = useSelector(currentLanguageSelector)
+
+  const walletConnectEnabled = v1 || v2
```
kathaypacific, line 110: this change resulted in one of the snapshots updating. previously, this was only looking at whether v1 was enabled. i can't remember why it was like this, but this flag shows / hides the connected applications. idk why we'd want to hide them if v1 is off but v2 is on. so i added it

src/account/Settings.tsx
```diff
@@ -55,222 +58,167 @@ import {
...
+
+  const { sessions: walletConnectV1Sessions } = useSelector(selectSessionsV1)
+  const { sessions: walletConnectV2Sessions } = useSelector(selectSessionsV2)
+
+  const { v1, v2 } = useSelector(walletConnectEnabledSelector)
+  const supportedBiometryType = useSelector(supportedBiometryTypeSelector)
+  const shouldShowRecoveryPhraseInSettings = useSelector(shouldShowRecoveryPhraseInSettingsSelector)
+  const hapticFeedbackEnabled = useSelector(hapticFeedbackEnabledSelector)
+  const decentralizedVerificationEnabled = useSelector(decentralizedVerificationEnabledSelector)
+  const currentLanguage = useSelector(currentLanguageSelector)
```
kathaypacific, line 108: we were previously using the language from `i18n` directly to display the language, but it should really be from redux. we weren't mocking `i18n` properly in the tests before so "unknown" was the language in the snapshots. 

side note: this is why i dislike snapshot testing. some snapshots don't even make sense. but i couldn't find the energy to refactor the tests today

MuckT:
@kathaypacific it looks like the confirm account removal modal isn't displaying after completing the backup quiz.

kathaypacific:
> @kathaypacific it looks like the confirm account removal modal isn't displaying after completing the backup quiz.

@MuckT is there anything in this PR that could have caused that? i'm confused about what the flow should look like, and whether the behaviour already exists on main

PR #3713, opened 2023-05-05
chore: add new native bottom sheet component

kathaypacific:
### Description

This PR creates a component that wraps the native bottom sheet to make it easier to use (since it needs to some some crazy stuff to dynamically size its own height based on the content inside). Use it in the dapps screen. This will be used for the revoke phone number work next.

### Test plan

n/a

### Related issues

- Fixes RET-717

### Backwards compatibility

Y


dievazqu (approved):
LGTM

PR #3714, opened 2023-05-05
feat: include positions when showing the total balance

jeanregisser:
### Description

As the title says ;)

### Test plan

- Updated tests

**Wallet home before/after**

 

**Assets screen before/after**

  

### Related issues

- Fixes RET-723

### Backwards compatibility

Yes


src/tokens/TokenBalances.tsx
```diff
@@ -49,19 +51,23 @@ function TokenBalancesScreen({ navigation }: Props) {
   const walletAddress = useSelector(walletAddressSelector)
   const insets = useSafeAreaInsets()
 
+  const totalPositionsBalanceUsd = useSelector(totalPositionsBalanceUsdSelector)
+  const totalPositionsBalanceLocal = useDollarsToLocalAmount(totalPositionsBalanceUsd)
```
kathaypacific, line 55: any benefit to selecting the positions balances in usd and converting to local currency with this hook? the tokens selector uses local currency, should we also just return the local currency in the positions selector too?

kathaypacific (approved):
Woohoo!! 🥳

src/tokens/TokenBalances.tsx
jeanregisser: I'm gonna use this selector in other places and USD seemed the better.
My thinking was that the display in local currency should be handled by a component dedicated to this.
The tokens balance would eventually be refactored to return USD instead.

PR #3716, opened 2023-05-08
chore(test): fix app init test which persisted spy mock across tests

kathaypacific:
### Description

This issue surfaced in [this PR](https://github.com/valora-inc/wallet/pull/3705/files#diff-c409fcb6ca669d62d1c6f0d9ec9ce59237ec8be1a4ebf746019b2adfaf5594b1R551). Digging into this, i noticed that the mock value for `sentryNetworkErrorsSelector` in the app init tests was the incorrect type, which was causing the error mentioned in #3705. That's weird right, since it's a different test and the next test already has `jest.clearAllMocks()` in the `beforeEach`? The spy loggers were not included in the `jest.clearAllMocks()` (we'd have to individually clear them with `jest.clearMock`) which was causing the carry over mock. Idk why I had used a jest spy initially, but it wasn't necessary. 

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y


MuckT (approved):
🚀

PR #3717, opened 2023-05-08
feat: add revoke phone number to settings screen

kathaypacific:
### Description

As the title. This PR:
- updates the ToastWithCTA component to handle the CTA being on the bottom (previously the CTA and content lived side by side)
- add a settings item for revoking if the user is verified, remove this option from the debug menu
- add a bottom sheet to confirm the revoke
- add toast / logger feedback for revoke success / error
- add analytics

### Test plan

Added videos in the linear task.

I still need to change the PhoneNumberWithFlag text to be black and bolder, but this affects other components so i'll make a follow up PR. Also i think we can remove the decentralised revoke flow...

### Related issues

- Fixes RET-717

### Backwards compatibility

Y


src/analytics/Events.tsx
```diff
@@ -199,6 +202,10 @@ export enum PhoneVerificationEvents {
   phone_verification_input_help_continue = 'phone_verification_input_help_continue', // when the user presses continue on the help dialog to continue verification
 
   phone_verification_resend_message = 'phone_verification_resend_message', // when the user triggers a resend of the sms code
+
+  phone_verification_revoke_start = 'phone_verification_revoke_start', // when the revoke endpoint is called
+  phone_verification_revoke_success = 'phone_verification_revoke_success', // when the revoke endpoint returns successfully
+  phone_verification_revoke_error = 'phone_verification_revoke_error', // when the revoke endpoint returns an error
```
kathaypacific, line 208: i may have overdone it with the events. the `phone_verification` ones are related to calling the cloud function (which we can also get from the gcp logs), the `settings` ones are for button press. idk, happy to remove if anyone feels strongly but i thought it'd be easier to analyse the data if they're in the same place

src/navigator/types.tsx
```diff
@@ -334,7 +334,6 @@ export type StackParamList = {
   }
   [Screens.VerificationStartScreen]:
     | {
-        showSkipDialog?: boolean
```
kathaypacific, line 337: unused

src/components/BottomSheet.tsx
```diff
@@ -16,6 +16,7 @@ interface Props {
   buttonLabel: string
   buttonOnPress: () => void
   buttonType?: BtnTypes
+  buttonLoading?: boolean
```
jeanregisser, line 19: Side note, I think I missed the PR for this `BottomSheet` but slightly surprised it always has a button.
Seems this could have been grouped into an optional `button` element passed to it which allows for greater customization.

src/analytics/Events.tsx
jeanregisser: Seems fine 👍

jeanregisser (approved):
Awesome ☎️ ✂️ 🚀

locales/base/translation.json
```diff
@@ -27,6 +27,15 @@
     "body": "To switch Data Saver on and off repeatedly, you will need to restart the app and return to this screen.",
     "restart": "Restart to Switch"
   },
+  "revokePhoneNumber": {
+    "title": "Unlink Phone Number",
```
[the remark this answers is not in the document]
kathaypacific: yes i discussed this with Reed - wording is hard, "revoke" was not used here because it sounds technical and user unfriendly. we did consider "disconnect" because of the "connect" during onboarding but disconnecting a phone number makes it sound like you're physically disabling your phone plan :/ we ended up going with "unlink" but i'd be happy to discuss again. did you have any options in mind?

src/account/Settings.tsx
```diff
@@ -512,6 +557,33 @@ export const Account = ({ navigation, route }: Props) => {
...
+          <Text style={styles.revokeNumberWarningText}>{t('revokePhoneNumber.description')}</Text>
+        </View>
+      </BottomSheet>
+      <ToastWithCTA
+        showToast={showRevokeSuccess}
+        message={t('revokePhoneNumber.revokeSuccess')}
+        labelCTA={t('revokePhoneNumber.addNewNumberButton')}
+        ctaAlignment="bottom"
+        onPress={handleNavigateToVerifiedNumber}
+      />
```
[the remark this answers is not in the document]
kathaypacific: yeah good call out, i abstracted :D

src/components/BottomSheet.tsx
kathaypacific: yes good point...i abstracted this a bit hastily. i want to leave it like this for this PR and i'll discuss with Reed how we want to use the sheet going forward, and update the component. 

perhaps we should have a BottomSheet that just takes children, and also have an ActionBottomSheet that contains the title / description / button. or just have a BottomSheet that takes children....idk. but preferably not a BottomSheet that takes optional title / description / button / children, that feels like the first 3 props are useless...

PR #3726, opened 2023-05-10
feat: Jumpstart link handler

gnardini:
### Description

Adding a new deep link handler for the Wallet Jumpstart initiative. 
Initiative doc: https://docs.google.com/document/d/1mt86To5oO5MFADBktFU5ET_zuUOo7UmiXLJvZOe3_MY
Technical doc: https://docs.google.com/document/d/1tikRZcfysRuV2nlHfNtpwc-5ZR1v5wMKoMt6EbtgLfM

### Test plan

To test, you need to create a valid jumpstart deep link using the https://github.com/valora-inc/wallet-jumpstart repo and open it using `xcrun simctl openurl booted "celo://wallet/jumpstart/0x72008a99c60006d17bdd5df8c9dbc4e4a79fd8d44d800d39c724db7c22cb4d99"` (with your own link, this one is already used)

### Backwards compatibility

N/A


src/jumpstart/utils.ts
jeanregisser, line 1: Maybe rename the file to `jumpstartLinkHandler.ts`?

src/jumpstart/utils.ts
```diff
@@ -0,0 +1,104 @@
...
+  beneficiary: string
+  signature: string
+  sendTo: string
+  assetType: 'erc20' | 'erc721'
+}
+
+async function claimReward(rewardInfo: RewardInfo) {
+  const queryParams = new URLSearchParams({ ...rewardInfo }).toString()
+  const requestUrl = `${networkConfig.walletJumpstartUrl}?${queryParams}`
+  const response = await fetch(requestUrl, { method: 'POST' })
```
jeanregisser, line 98: Should we use `fetchWithTimeout`? Passing a custom timeout probably as the default is 15 secs.

src/jumpstart/utils.ts
jeanregisser, line 1: Can we add a test to this one? or too early for that?

jeanregisser (approved):
Awesome! 🚀

src/jumpstart/utils.ts
```diff
@@ -0,0 +1,104 @@
...
+        signature,
+        sendTo: userAddress,
+        assetType,
+      })
+    } catch (error: any) {
+      if (error.message === 'Already claimed') {
+        continue
+      }
+
+      if (error.message === 'execution reverted') {
```
silasbw, line 69: where does this message actually get generated from?  does the wallet jumpstart API somehow return it?

src/jumpstart/utils.ts
```diff
@@ -0,0 +1,104 @@
...
+          ? await jumpstart.methods.erc20Claims(beneficiary, index).call()
+          : await jumpstart.methods.erc721Claims(beneficiary, index).call()
+      if (info.claimed) {
+        throw new Error('Already claimed')
+      }
+
+      const messageHash = kit.web3.utils.soliditySha3(
+        { type: 'address', value: beneficiary },
+        { type: 'address', value: userAddress },
+        { type: 'uint256', value: '0' }
```
silasbw, line 48: did you mean to hardcode the index in the signature to `'0'`?

src/jumpstart/utils.ts
```diff
@@ -0,0 +1,104 @@
...
+) {
+  let index = 0
+  while (true) {
+    try {
+      const info =
+        assetType === 'erc20'
+          ? await jumpstart.methods.erc20Claims(beneficiary, index).call()
+          : await jumpstart.methods.erc721Claims(beneficiary, index).call()
+      if (info.claimed) {
+        throw new Error('Already claimed')
```
silasbw, line 42: is there a reason to `throw` instead of simply a `continue` (maybe hoisting the log message up from below)?

gnardini: Not really, mb, switched to straight `continue`

src/jumpstart/utils.ts
gnardini: This is the error that is thrown if you try to read an index that doesn't exist with `erc20Claims` or `erc721Claims`.

src/jumpstart/utils.ts
gnardini: I wanted not to, but I think CI is forcing me 😅

src/jumpstart/utils.ts
silasbw: ohhhhhhhh, that makes sense.

src/jumpstart/utils.ts
gnardini: Added the test 🎉

PR #3728, opened 2023-05-10
fix: ensure inviter and invitee is present in the same event

kathaypacific:
### Description

The firebase dynamic link (referral link) is consumed on app open, which is (mostly) before there is a wallet address created.  To make the association between invitee and inviter more reliably, add the inviter address to the verification success event.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y


jeanregisser (approved):
LGTM 👍

PR #3730, opened 2023-05-10
chore: use load balancer url for getting positions + pass right network

jeanregisser:
### Description

Title says it all ;)

### Test plan

- Waiting for https://github.com/valora-inc/plugins-api/pull/47 to confirm

### Related issues

- N/A

### Backwards compatibility

Yes

### DIFF
--- a/src/positions/saga.ts
+++ b/src/positions/saga.ts
@@ -1,4 +1,5 @@
 import { call, put, select, spawn, takeLeading } from 'redux-saga/effects'
+import { DEFAULT_TESTNET } from 'src/config'
 import {
   fetchPositionsFailure,
   fetchPositionsStart,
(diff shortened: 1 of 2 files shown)


kathaypacific (approved):

src/positions/saga.ts
```diff
@@ -21,9 +23,11 @@ const POSITIONS_FETCH_TIMEOUT = 45_000 // 45 seconds
 
 async function fetchPositions(walletAddress: string) {
   const response = await fetchWithTimeout(
-    // TODO: Use the final API URL once it's ready
-    'https://plugins-api-oaxbpxoaha-uc.a.run.app/balances?' +
-      new URLSearchParams({ network: 'celo', address: walletAddress }),
+    `${networkConfig.cloudFunctionsUrl}/hooks-api/getPositions?` +
```
silasbw, line 26: The existing pattern/convention is to explicitly define URLs in `src/web3/networkConfig` (for better or worse).  Can you do that instead (e.g., add a properties to the `networkConfigs`)?  And separately we can discuss changing the convention?

jeanregisser: Oh no, I've wanted to refactor that for ages. And I've already broken that [convention](https://github.com/valora-inc/wallet/blob/76971cf0cba23e8261553bcc10f30e923097426a/src/exchange/CeloNewsFeed.tsx#L45) 😅

PR #3720, opened 2023-05-09
feat(keyless-backup): add sign in with email screen

satish-ravi:
### Description

[Figma](https://www.figma.com/file/gGp6SVFWMbnSltkGzNzrpC/Wallet-Recovery?type=design&node-id=2698-22743&t=3w7xdC81dvYiLeye-0), depends on #3732

Just the UI, button click will be in a follow up

### Test plan

Manual

### Related issues

- Part of ACT-743

### Backwards compatibility

Yes

PR #3732, opened 2023-05-10
feat(Button): parameterize iconMargin and touchableStyle

satish-ravi:
### Description

Parameterizing `iconMargin` and `touchableStyle` will allow us to use the Button component for [this design](https://www.figma.com/file/gGp6SVFWMbnSltkGzNzrpC/Wallet-Recovery?type=design&node-id=2698-22743&t=3w7xdC81dvYiLeye-0). The main change is in src/components/Button.tsx, all others are snapshots.

### Test plan

Unit test. Will test the actual use of params in the next PR to create the email screen with button using these params. (see #3720)

### Related issues

- Part of ACT-743

### Backwards compatibility

Yes


jophish (approved):
Looked at `Button.tsx` and one or two snapshots -- looks good to me.

PR #3734, opened 2023-05-11
feat: update UI on token balances screen

kathaypacific:
### Description

This PR tweaks a few things:
- spacing on the screen to match [designs](https://www.figma.com/file/LZJAPf1CdLSRGujzQHCbXj/In-app-dapp-actions?type=design&node-id=386-3997&t=1xosKbu8Dhi99Q5P-0)
- use flatlist instead of scrollview
- update nft viewer banner design
- update percentage indicator svg's
- optionally render the asset balance on the screen header

Not included in this PR:
positions related things

### Test plan

https://github.com/valora-inc/wallet/assets/20150449/8f4a6a7d-ef3d-432f-a810-211cc859c7d5

### Related issues

- Fixes RET-734, RET-737

### Backwards compatibility

Y


src/tokens/AssetBalanceFiat.tsx
```diff
@@ -0,0 +1,68 @@
+import BigNumber from 'bignumber.js'
```
kathaypacific, line 1: this component looks similar to the one on the home screen but it will eventually have this [tool tip thing](https://www.figma.com/file/LZJAPf1CdLSRGujzQHCbXj/In-app-dapp-actions?type=design&node-id=225-2383&t=1xosKbu8Dhi99Q5P-0). that's why this is a new component

src/tokens/TokenBalances.tsx
```diff
@@ -127,6 +146,18 @@ function TokenBalancesScreen({ navigation }: Props) {
...
+  const handleMeasureBalanceHeight = (event: LayoutChangeEvent) => {
+    balanceHeightRef.current = event.nativeEvent.layout.height
+  }
+
+  const debouncedUpdateHeader = debounce((scrollOffset: number) => {
+    setShowBalanceInHeader(scrollOffset > balanceHeightRef.current)
+  }, HEADER_UPDATE_DEBOUNCE_TIME_MS)
+
+  const handleScroll = (event: { nativeEvent: NativeScrollEvent }) => {
+    debouncedUpdateHeader(event.nativeEvent.contentOffset.y)
```
kathaypacific, line 158: can't debounce at the `handleScroll` level because the synthetic scroll events are reused for performance reasons and would return null sometimes if debounced

src/tokens/AssetBalanceFiat.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  const totalBalanceLocal =
+    totalTokenBalanceLocal || totalPositionsBalanceLocal
+      ? new BigNumber(totalTokenBalanceLocal ?? 0).plus(totalPositionsBalanceLocal ?? 0)
+      : undefined
+
+  const balance = totalBalanceLocal?.toFormat(2) ?? new BigNumber(0).toFormat(2)
+  // TODO handle when fetching asset balances
+  const displayedBalance =
+    tokenFetchError || tokensAreStale || positionsStatus === 'error' ? ' –' : balance
```
dievazqu, line 42: Hey, not related directly to your feature, but I'm a bit confused about using `tokensAreStale`.
I saw we were already having this logic... But, this means that if any token is stale we are not showing the total balance, right? Even if it's a token that the user doesn't have?
I'm confused because I think we have a bunch of stale tokens in the list, but I'm not seeing this logic applied.

dievazqu: Oh, I think I know where the confusion is coming from...
I think the `stalePriceSelector` always returns `false`, because we are filtering the stale tokens [here](https://github.com/valora-inc/wallet/blob/fd340191e4bf884ad884e2ffad29f2713d50c101/src/tokens/selectors.ts#L88).
I'm wondering what we should do about this... It seems that we wanted to avoid showing the balance if we can't calculate the usd price of any of the tokens the user is holding... I think we are not doing that...
Since it's not related to your PR, I guess we can create a debt tech task for this.

jeanregisser, line 42: I think for now we should ignore errors fetching positions and still display what we have.
To match the balance display logic in the home feed.

src/tokens/TokenBalances.tsx
jeanregisser, line 158: Would be better to use an `Animated.Value` and update the opacity of the header, after a certain threshold.
This way no need to debounce and no arbitrary delay.

src/tokens/AssetBalanceFiat.tsx
jeanregisser: I think we should avoid differences and use the same component. Otherwise we risk having inconsistencies.

kathaypacific: thanks for this! i actually misunderstood what TokenBalance does, I thought that it was the whole component (with label and buttons etc) but it's not 😅 so i could reuse it

src/tokens/TokenBalances.tsx
```diff
@@ -127,6 +149,15 @@ function TokenBalancesScreen({ navigation }: Props) {
...
   }
 
+  const handleMeasureBalanceHeight = (event: LayoutChangeEvent) => {
+    balanceHeightRef.current = event.nativeEvent.layout.height
+  }
+
+  const handleScroll = (event: { nativeEvent: NativeScrollEvent }) => {
+    const opacityValue = event.nativeEvent.contentOffset.y > balanceHeightRef.current ? 1 : 0
+    headerOpacity.value = withTiming(opacityValue)
+  }
```
jeanregisser, line 159: We should use [`useAnimatedScrollHandler`](https://docs.swmansion.com/react-native-reanimated/docs/api/hooks/useAnimatedScrollHandler) along with `Animated.FlatList` otherwise the trigger won't be driven by the native driver.
i.e. it will be delayed if the JS thread is busy.

kathaypacific: thanks! i've updated this :D

jeanregisser (approved):

PR #3737, opened 2023-05-12
chore: upgrade to react native 0.69

kathaypacific:
### Description

The RN upgrades were light on the native side but the big change is React 18. There are some [breaking changes](https://blog.logrocket.com/upgrading-react-18-typescript/) which I had to fix with a mixture of manual updates and dependency upgrades. Of note:
- in the AppDelegate.mm file i had to move the `appStartedMillis` logic into the new `prepareInitialProps` function
- the `use_flipper` function has been moved inside `use_react_native`, the necessary flipper config is now passed through the input param for `use_react_native`
- in React 18, implicit children is no longer allowed. some components had `children` manually added to the prop defs
- annoying things happened with `Trans` and the output of `t` not being accepted as JSX elements anymore. so i upgraded react-i18n which fixed those issues, but the definition for the return type of `t` has changed. previously, it was `string | undefined` but now it is `string | null`. most of the manual changes were updating components to accept `null` instead of `undefined`.
- `TransProps` also changed, the definition was a little hard to use so i updated the invite modal components to take a string instead.
- the `@testing-library/react-hooks` lib was giving me typescript issues with the `wrapper`, it actually seems that this lib has been combined with `react-native-testing-library`. so i updated `react-native-testing-library` in order to resolve the TS errors.
- updated the svg lib from v12 to v13 because to get their TS definitions compatible with react 18. unfortunately v13 of the svg lib has some issues on android with animated svg's, i spent some time trying out the suggested fixes but couldn't get it to work. so i disabled the animation on android for now (see QRTabBar.tsx)
- react-native-camera is deprecated, but rather than changing out the lib right now i just patched it. ViewPropTypes is no longer an export of react native but it was an import which caused the app to crash on the QR screen. I just removed the import because proptypes don't affect runtime.
- changes in unit test / snapshots: empty text components are now rendered as `<Text />` rather than `<Text> </Text>`.  some new lines were also removed.
- the TokenBottomSheet and DAppsExplorerScreenSearch components use debounce, and i had a weird time with the unit tests for these. the upgraded testing lib started to behave weirdly with async tests (the tests would time out and not move on), removing the async seemed to fix this.

### Test plan

tested that debug builds run locally / flipper worked

### Related issues

- Fixes RET-683

### Backwards compatibility

y


ios/Podfile
```diff
@@ -24,9 +24,11 @@ require_relative "../node_modules/react-native/scripts/react_native_pods"
 require_relative "../node_modules/@react-native-community/cli-platform-ios/native_modules"
 
 platform :ios, '12.4'
-# install! 'cocoapods', :deterministic_uuids => false
+install! 'cocoapods', :deterministic_uuids => false
```
jeanregisser, line 27: Nothing to do but I wonder why `:deterministic_uuids => false` is now the default 🤔

jeanregisser (approved):
🚀

ios/Podfile
```diff
@@ -24,9 +24,11 @@ require_relative "../node_modules/react-native/scripts/react_native_pods"
...
 use_frameworks!
 
+production = ENV["PRODUCTION"] == "1"
```
jeanregisser, line 30: I wonder what this does.
Do you think we need to set this for release builds?

kathaypacific: i'll double check. i saw it being used in the use_react_native to trigger the use_flipper function...

kathaypacific:
jokes, this is a bigger upgrade than i thought it would be. from the react side 🙈 got some breaking TS changes....https://blog.logrocket.com/upgrading-react-18-typescript/

ios/Podfile
kathaypacific: the only place this is used if to enable flipper / the react dev pods. in the use_react_native script
```
if flipper_configuration.flipper_enabled && !production
    pod 'React-Core/DevSupport', :path => "#{prefix}/"
    use_flipper!(flipper_configuration.versions, :configurations => flipper_configuration.configurations)
  end
```

i don't really know where ENV is defined though, i assume that we don't need to do anything extra if the CI will run?

jeanregisser: yes it should be good. Thanks.

MuckT (approved):
For the tests not having to wrap fire.event in an act is great. I did notice the build time is a bit larger than `0.68`, but maybe that's to be expected: [Sample 0.68 build](https://github.com/valora-inc/wallet/actions/runs/4988912096/jobs/8932168503).

src/i18n/i18n.test.ts
```diff
@@ -61,7 +61,7 @@ describe('i18n', () => {
     it('only loads the selected language, but loads the default language when accessing a missing key', async () => {
       await i18n.changeLanguage('es-419')
       expect(i18n.t('someKey')).toEqual('¡Hola!')
-      expect(enLoaded).toBe(false)
+      expect(enLoaded).toBe(true) // i18n was initialised with en-US in the beforeEach
```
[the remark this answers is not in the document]
jeanregisser: I see, previously it didn't load the default language until a key was accessed.
I think the change is fine. Thanks.

src/qrcode/QRTabBar.tsx
```diff
@@ -43,11 +43,21 @@ export default function QRTabBar({
...
 
+  // using `animatedColor` with animated svg causes android crash since
+  // upgrading react-native-svg to v13. there are some suggested solutions
+  // linked below, but none that i could get to work after many attempts. since
+  // this is a relatively low impact feature, i'm going to leave it as is for
+  // now.
+  // https://github.com/software-mansion/react-native-svg/issues/1976
+  // https://github.com/software-mansion/react-native-reanimated/issues/3775
+  const color =
+    Platform.OS === 'ios' ? animatedColor : state.index === 0 ? colors.dark : colors.light
```
jeanregisser, line 59: 👍

jeanregisser (approved):
W00t! Amazing 🚀 🚀 🚀

PR #3740, opened 2023-05-15
feat(keyless-backup): add sign in with google

satish-ravi:
### Description

Implements button click to initiate google sign in flow. Result saved on redux but not used. Also encrypts .env.* files since they now include client ids. See [this thread](https://valora-app.slack.com/archives/C025V1D6F3J/p1684179136241829) for more context.

To be merged after #3741 

### Test plan

Manually by running the following steps

1. Enable backup locally by changing the feature flag to true (this line needs to change to true: https://github.com/valora-inc/wallet/blob/94fee8a/src/account/Settings.tsx#L374)
2. Build and run app, navigate to settings
3. Click on Keyless backup, click continue, then Sign In With Google and complete sign in
4. Ensure idToken is saved in redux (can check logs for googleSignInCompleted event or look in redux debugger in flipper)

iOS

https://github.com/valora-inc/wallet/assets/5062591/e432cac0-9992-4a03-94f7-3af7e12f777f

Android

https://github.com/valora-inc/wallet/assets/5062591/4c7874f4-8a7d-4909-9b31-ac53e8e0c062

### Related issues

- Fixes ACT-743

### Backwards compatibility

Yes

PR #3741, opened 2023-05-16
feat(keyless-backup): add redux slice and saga

satish-ravi:
### Description

Redux setup for keyless backup sign in with google flow. The actual sign in stuff will be implemented in #3740

### Test plan

Unit tests

### Related issues

- Part of ACT-743

### Backwards compatibility

Yes


src/keylessBackup/slice.ts
```diff
@@ -0,0 +1,41 @@
+import { PayloadAction, createSlice } from '@reduxjs/toolkit'
+
+export interface State {
+  google: {
+    loading: boolean
+    idToken: string | null
```
jeanregisser, line 6: Is it safe to persist `idToken`? The persisted state has no protection and is readable in clear text.

src/keylessBackup/slice.ts
```diff
@@ -0,0 +1,41 @@
...
+    error: string | null
```
jeanregisser, line 7: nit: to avoid representing impossible states (both loading and error at the same time)
```suggestion
    status: 'idle' | 'loading' | 'success' | 'error'
    idToken: string | null
```

src/keylessBackup/slice.ts
satish-ravi: The entire keylessBackup state is blacklisted here: https://github.com/valora-inc/wallet/pull/3741/files#diff-b3695b5171edbee1bb22dc8893e3517773bacdd8a51ba2a568e077f6e1e07940R29, so this wouldn't be persisted right?

jh2oman (approved):

satish-ravi:
Build is red because of codecov/patch. It just complains because tests are missing for trivial state changes in slices and saga setup.

PR #3746, opened 2023-05-17
feat(securesend): Remove Legacy Send screen from Secure Send flow

jophish:
### Description

For ACT-761.

Removes the `SendConfirmationLegacy` screen from the Secure Send flow. Currently, there are _no_ paths to the `ValidateRecipientAccount` screen that result in `SendConfirmationLegacy` being navigated to. As a result, there is no way for `SendConfirmationLegacy` to be reached in the Secure Send flow, which means that Secure Send specific  _features_ in the `SendConfirmationLegacy` screen (namely, having an "edit" button for the account address when there are multiple mappings) can be removed, since they will never be seen.

### Related issues

- Fixes ACT-761.

### Backwards compatibility

Yes.


jh2oman (approved):
One down!! 🎉

PR #3755, opened 2023-05-18
fix: revert encrypting env files

satish-ravi:
### Description

Reverting env encryption, which broke [release-automation](https://github.com/valora-inc/release-automation/actions/runs/5010075760/jobs/8979626618). Moved the google ouath client id to secrets.json, which doesn't need to be build specific, just alfajores/mainnet specific. This breaks iOS support for cloud backup features that are not released yet, which will be addressed later.

### Test plan

Manually by running through the sign in flow on Android, watch release-automation for tonight

### Related issues

- N/A

### Backwards compatibility

Yes


cajubelt (approved):
can you clarify in the description that it breaks iOS support for *cloud backup features that are not released yet*

PR #3763, opened 2023-05-19
feat: add info tooltip for total balance in assets screen

kathaypacific:
### Description

As the title

### Test plan

https://github.com/valora-inc/wallet/assets/20150449/7fc06279-eecf-4d39-ba1f-736e9060d62f

### Related issues

- Fixes RET-734

### Backwards compatibility

Y


src/components/TokenBalance.tsx
```diff
@@ -119,13 +120,35 @@ function useErrorMessageWithRefresh() {
...
+  }
 
   return (
     <View testID="AssetsTokenBalance">
-      <Text style={styles.totalAssets}>{t('totalAssets')}</Text>
+      <View style={styles.row}>
+        <Text style={styles.totalAssets}>{t('totalAssets')}</Text>
+        {showInfo && (
+          <TouchableOpacity
+            onPress={toggleInfoVisible}
```
dievazqu, line 137: I think it's a bit weird to dismiss the message only when tapping again in the info icon. Didn't you feel it weird?
I would expect that tapping anywhere would dismiss it.

src/components/TokenBalance.tsx
```diff
@@ -221,6 +244,21 @@ const styles = StyleSheet.create({
   totalAssets: {
     ...fontStyles.regular600,
     color: Colors.gray5,
+    marginRight: 4,
+  },
+  totalAssetsInfoContainer: {
+    position: 'absolute',
+    top: 24,
```
dievazqu, line 251: Related to my other comment, does this ensure the message appears below the info icon? Because, if somehow this is shown on top of the icon, the user won't be able to dismiss it.

dievazqu (approved):
Approving but I'm not sure about the dismissing logic.
But I'll leave that decision to you 😋

src/components/TokenBalance.tsx
kathaypacific: yeah this is a good point, it's not trivial to enable this on tap of _everywhere_ though so for now i've just added a handler to dismiss on tap of the header component

kathaypacific:
> What do you think of adding a little animation? Maybe a simple fade in/out?

yup good idea, i added. i think the default animation time is too slow though, i decreased it a bit (it will also affect the toast and legacy bottom sheet components but i really believe in the snappier animation....)

src/components/TokenBalance.tsx
```diff
@@ -119,14 +131,62 @@ function useErrorMessageWithRefresh() {
   }, [shouldShowError])
 }
 
-export function AssetsTokenBalance() {
+export function AssetsTokenBalance({ showInfo }: { showInfo: boolean }) {
   const { t } = useTranslation()
 
+  const [infoVisible, setInfoVisible] = useState(false)
+  const [shouldRenderInfoComponent, setShouldRenderInfoComponent] = useState(false)
```
kathaypacific, line 138: it's important to stop rendering the tooltip (instead of leaving it there with opacity 0) because it sits on top of the segmented control and will prevent the segmented control from receiving the tap action

src/components/TokenBalance.tsx
```diff
@@ -119,14 +131,62 @@ function useErrorMessageWithRefresh() {
...
-    <View testID="AssetsTokenBalance">
-      <Text style={styles.totalAssets}>{t('totalAssets')}</Text>
-      <TokenBalance singleTokenViewEnabled={false} />
-    </View>
+    <TouchableWithoutFeedback onPress={handleDismissInfo}>
+      <View testID="AssetsTokenBalance">
+        <View style={styles.row}>
+          <Text style={styles.totalAssets}>{t('totalAssets')}</Text>
+          {showInfo && (
+            <TouchableOpacity
```
jeanregisser, line 172: Ideally this would use `src/components/Touchable` to respect the platform convention (opacity on iOS and Ripple on Android). But we haven't been very consistent so far through the app.

jeanregisser (approved):

src/components/TokenBalance.tsx
kathaypacific: for this component i opted for the same behaviour on the home screen - i think TouchableOpacity works better here because the info icon is round. i did try with the Touchable component but the shadow is square and grey on android...

PR #3768, opened 2023-05-19
chore: dispatch in app review on successful swap and reward claim

MuckT:
### Description

Adds the additional trigger actions of `swapSuccess` and `claimRewardsSuccess` for In App Review. I opted to excluded dapp tx as the action in-app review would likely disrupt the dapp action flow and this is against [best practices](https://developer.apple.com/design/human-interface-guidelines/ratings-and-reviews).

> Avoid interrupting people while they’re performing a task or playing a game. Asking for feedback can disrupt the user experience and feel like a burden. Look for natural breaks or stopping points in your app or game where a rating request is less likely to be bothersome.

### Test plan

Tested locally on iOS.

### Related issues

- Fixes RET-15

### Backwards compatibility

Yes

### DIFF
--- a/src/app/saga.ts
+++ b/src/app/saga.ts
@@ -55,6 +55,7 @@ import {
   DYNAMIC_LINK_DOMAIN_URI_PREFIX,
   FETCH_TIMEOUT_DURATION,
 } from 'src/config'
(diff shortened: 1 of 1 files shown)


kathaypacific (approved):
🙏🏻

PR #3770, opened 2023-05-22
feat: add positions to assets screen

kathaypacific:
### Description

This PR adds UI for positions items:
- use SectionList for displaying the assets, since the positions need to be grouped by app
- add PositionItem component to display the positions
- the positions endpoint will return all the data that needs to be displayed (title, description, image)

### Test plan

(Note the green image backgrounds is just for this video, since we don't have images yet)

https://github.com/valora-inc/wallet/assets/20150449/246f17a0-eca8-4241-be9a-674f298c5c86

### Related issues

- Fixes RET-734

### Backwards compatibility

Y


src/redux/store.ts
```diff
@@ -123,6 +123,7 @@ export const setupStore = (initialState = {}, config = persistConfig) => {
           'goldToken',
           'stableToken',
           'send',
+          'positions',
```
jeanregisser, line 126: 👍

src/tokens/TokenBalances.tsx
jeanregisser, line 1: I think there's an opportunity to move these components to a new folder. Maybe `src/userAssets`? or something
We can do that later though.

test/RootStateSchema.json
jeanregisser, line 1: Ideally we should add a migration which deletes the existing positions data which didn't have the new required properties. Otherwise there's a risk of a crash because the new code accesses these new properties.

jeanregisser (approved):
Looks amazing 🤩 🚀

jeanregisser:
Small thing, the designs used UPPERCASE for dapp names:

Screenshot 2023-05-24 at 11 26 55

test/RootStateSchema.json
kathaypacific: true, i added a migration now

kathaypacific:
> Small thing, the designs used UPPERCASE for dapp names:
> 
> Screenshot 2023-05-24 at 11 26 55

thanks!! i missed this, i updated :D

PR #3778, opened 2023-05-23
chore(payment-request): Deprecate SendConfirmationLegacy's use in the payment request flow

jh2oman:
### Description

The only benefit of the old SendConfirmationLegacy screen was that it allowed users to edit which currency they want to send. Instead of that, payment requests now follow the following if/else rules to determine which currency to send. 

* If user has enough CUSD balance to fulfill, take them to SendConfirmation screen for CUSD
* Else, try the same thing for CEUR
* Else, red error banner saying they have insufficient cUSD or cEUR to complete the payment

https://github.com/valora-inc/wallet/assets/8432644/df35bda7-ed07-4438-b901-2c9299a1bba9

### Test plan

Added Unit tests. manually tested in mainnet.

### Related issues

[- Fixes #[issue number here]
](https://linear.app/valora/issue/ACT-764/payment-requests-sendconfirmationlegacy-=-sendconfirmation)

### Backwards compatibility

Can handle payment requests that originate from older app versions


src/paymentRequest/IncomingPaymentRequestListItem.tsx
```diff
@@ -63,26 +69,58 @@ export default function IncomingPaymentRequestListItem({ id, amount, comment, re
...
+      // Should never happen in producton
+      throw new Error('No token address found for cUSD')
+    }
+    // If the user has enough cUSD balance, pay with cUSD
+    // Else, try with cEUR
+    // Else, throw up an error banner
+    let transactionData: TransactionDataInput
+    const usdRequested = new BigNumber(paymentRequest.amount)
+
+    if (usdRequested.isLessThanOrEqualTo(cUsdTokenInfo.balance)) {
```
satish-ravi, line 89: this makes the assumption that 1 usd = 1 cusd right? should we just do the conversion similar to cEUR just to be safe

src/paymentRequest/IncomingPaymentRequestListItem.tsx
```diff
@@ -63,26 +69,58 @@ export default function IncomingPaymentRequestListItem({ id, amount, comment, re
...
+    const cEurTokenInfo = stableTokens.find((token) => token?.symbol === Currency.Euro)
+    if (!cUsdTokenInfo?.address || !cEurTokenInfo?.address) {
+      // Should never happen in producton
+      throw new Error('No token address found for cUSD')
+    }
+    // If the user has enough cUSD balance, pay with cUSD
+    // Else, try with cEUR
+    // Else, throw up an error banner
+    let transactionData: TransactionDataInput
+    const usdRequested = new BigNumber(paymentRequest.amount)
```
satish-ravi, line 87: is paymentRequest.amount always in usd? even if the local currency is different?

src/paymentRequest/IncomingPaymentRequestListItem.tsx
```diff
@@ -63,26 +69,58 @@ export default function IncomingPaymentRequestListItem({ id, amount, comment, re
...
-      reason: comment,
-      recipient: requester,
-      amount: new BigNumber(amount),
-      currency: Currency.Dollar,
-      type: TokenTransactionType.PayRequest,
-      firebasePendingRequestUid: id,
+    const cUsdTokenInfo = stableTokens.find((token) => token?.symbol === Currency.Dollar)
+    const cEurTokenInfo = stableTokens.find((token) => token?.symbol === Currency.Euro)
+    if (!cUsdTokenInfo?.address || !cEurTokenInfo?.address) {
+      // Should never happen in producton
```
satish-ravi, line 80:
```suggestion
      // Should never happen in production
```

src/paymentRequest/IncomingPaymentRequestListItem.tsx
```diff
@@ -63,26 +69,58 @@ export default function IncomingPaymentRequestListItem({ id, amount, comment, re
...
+      throw new Error('No token address found for cUSD')
```
satish-ravi, line 81:
```suggestion
      throw new Error('No token address found for cUSD or cEUR')
```

src/paymentRequest/IncomingPaymentRequestListItem.tsx
jh2oman: Yep thats how it was implemented :)

src/paymentRequest/IncomingPaymentRequestListItem.tsx
jh2oman: good idea

src/paymentRequest/IncomingPaymentRequestListItem.tsx
```diff
@@ -63,26 +69,58 @@ export default function IncomingPaymentRequestListItem({ id, amount, comment, re
...
+        comment: paymentRequest.comment,
+        recipient: requester,
+        inputAmount: new BigNumber(paymentRequest.amount),
+        tokenAmount: new BigNumber(paymentRequest.amount),
+        amountIsInLocalCurrency: false,
+        tokenAddress: cUsdTokenInfo.address,
+        paymentRequestId,
+      }
+    } else if (
+      cEurTokenInfo.usdPrice &&
```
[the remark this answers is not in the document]
jh2oman: It is just to make TS happy. There appears to be other places in the code where usdPrice is expected to be undefined. I'm not sure if that applies to the stablecoins but I'm hesitant to move the check to the 'should never happen in production' section for that reason

src/paymentRequest/IncomingPaymentRequestListItem.tsx
jh2oman: poorly named back then to just be 'amount'

src/paymentRequest/IncomingPaymentRequestListItem.tsx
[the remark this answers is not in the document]
satish-ravi: I see, so if these are undefined for some reason, we'd just end up throwing "not enough balance" right? Maybe worth double checking on if / when these can be undefined

satish-ravi (approved):

PR #3792, opened 2023-05-23
chore: add analytics for assets

kathaypacific:
### Description

As the title. We have a separate task for tracking positions in the super properties so i have not added that here.

### Test plan

n/a

### Related issues

- Fixes RET-734

### Backwards compatibility

Y


jeanregisser (approved):
🐑 it

PR #3796, opened 2023-05-24
chore: upgrade to react native 0.70

kathaypacific:
### Description

This is a relatively peaceful upgrade, the main difference is that hermes is now enabled on iOS (it was already enabled on Android).

The only trouble I had was building iOS debug, but luckily some helpful person on the internet had this issue and clearly explained [the fix](https://github.com/facebook/react-native/issues/32333#issuecomment-944159768) ✨

### Test plan

iOS and Android debug build still works. CI still runs.

### Related issues

- Fixes RET-683

### Backwards compatibility

Y


src/swap/SwapScreen.tsx
```diff
@@ -215,12 +215,7 @@ export function SwapScreenSection({ showDrawerTopNav }: { showDrawerTopNav: bool
   const handleShowTokenSelect = (fieldType: Field) => () => {
     ValoraAnalytics.track(SwapEvents.swap_screen_select_token, { fieldType })
     // ensure that the keyboard is dismissed before animating token bottom sheet
-    Keyboard.addListener('keyboardDidHide', onKeyboardDidHide(fieldType))
     Keyboard.dismiss()
-  }
-
-  const onKeyboardDidHide: any = (fieldType: Field) => {
-    Keyboard.removeListener('keyboardDidHide', onKeyboardDidHide)
```
kathaypacific, line 223: `removeListener` no longer exists on `Keyboard`

jeanregisser: I assume it's because `addListener` returns a callback to stop listening, right?
Don't we need to keep the initial logic?

kathaypacific: yes that's right. i opted for this logic change because idk how to reference the return value of `addListener` inside the callback of the `addListener`. i'm not sure that the existing logic is adding value anyway, i've attached videos of what i see on the nightly vs what i see on this branch. they look similar enough to me (the nightly build keyboard is not completely dismissed before the bottom sheet starts animating)

nightly:

https://github.com/valora-inc/wallet/assets/20150449/c735d466-6e13-4f00-9bcd-643ead6b9e04


this branch:

https://github.com/valora-inc/wallet/assets/20150449/e7cb9fac-3047-49fd-b661-c93a331cef29

kathaypacific:
> w00t, awesome! 🚀
> 
> It's great a bunch of frameworks are now linked statically. This should help with app startup time.
> 
> Also according to the Emerge tool analysis, download size is mostly stable, though we have the Hermes library added on iOS. I think this is also because of the switch from dynamic to static frameworks.
> 
> Do you think we can remove this too now?
> 
> https://github.com/valora-inc/wallet/blob/3c0dcbf6812d40adf62d2ba0e87ea36be1e00ce8/ios/Podfile#L9-L21

unfortunately I don't think we can remove the `$static_framework` workaround, the build does not succeed without this. the firebase docs recommend `use_frameworks! :linkage => :static` but `use_frameworks!` still doesn't work correctly with flipper...

src/swap/SwapScreen.tsx
jeanregisser: Great, thanks for double checking, I don't see much difference either.

jeanregisser (approved):
🐑 it

PR #3797, opened 2023-05-24
fix(tokens): ensure header component is above scrollbar on overscroll

kathaypacific:
### Description

Context https://valora-app.slack.com/archives/C029Z1QMD7B/p1684931810127659

### Test plan

https://github.com/valora-inc/wallet/assets/20150449/61937fb0-3ef7-4c40-b0ba-8744cbc8036e

### Related issues

- Fixes RET-732

### Backwards compatibility

Y

### DIFF
--- a/src/tokens/TokenBalances.tsx
+++ b/src/tokens/TokenBalances.tsx
@@ -312,6 +312,7 @@ function TokenBalancesScreen({ navigation, route }: Props) {
       <AnimatedSectionList
         contentContainerStyle={{
           paddingBottom: insets.bottom,
+          opacity: listHeaderHeight > 0 ? 1 : 0,
         }}
         // Workaround iOS setting an incorrect automatic inset at the top
         scrollIndicatorInsets={{ top: 0.01 }}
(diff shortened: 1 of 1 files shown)


src/tokens/TokenBalances.tsx
```diff
@@ -310,6 +312,7 @@ function TokenBalancesScreen({ navigation, route }: Props) {
       <AnimatedSectionList
         contentContainerStyle={{
           paddingBottom: insets.bottom,
+          opacity: listHeaderHeight > 0 ? 1 : 0,
```
kathaypacific, line 315: snuck this one in - this prevents the items being pushed down after the list header is measured

jeanregisser:
The video still shows a bug:
Screenshot 2023-05-24 at 18 16 10

We either need to put the header within the ScrollView (that won't be an option if we later on add the swipe to switch between assets / positions), or update `scrollIndicatorInsets`.

jeanregisser (approved):
👩‍🎨

PR #3799, opened 2023-05-24
feat(actions): Hide swap action when disabled

jophish:
### Description

For ACT-810. Hides the swap action for home screen actions experiment when it's disabled. This check uses the same flag that we use to [determine swap eligibility on the drawer navigator](https://github.com/valora-inc/wallet/blob/main/src/navigator/DrawerNavigator.tsx#L224).

### Test plan

Unit and manual tested.

### Related issues

- Fixes ACT-810.

### Backwards compatibility

Yes.

### DIFF
--- a/src/home/ActionsCarousel.test.tsx
+++ b/src/home/ActionsCarousel.test.tsx
@@ -9,18 +9,49 @@ import { navigate } from 'src/navigator/NavigationService'
 import { Screens } from 'src/navigator/Screens'
 import { CloseIcon } from 'src/navigator/types'
 import { mocked } from 'ts-jest/utils'
+import { MockStoreEnhanced } from 'redux-mock-store'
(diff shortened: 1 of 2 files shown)


src/home/ActionsCarousel.test.tsx
```diff
@@ -9,18 +9,49 @@ import { navigate } from 'src/navigator/NavigationService'
...
+    app: {
+      showSwapMenuInDrawerMenu: shouldShowSwapAction,
+    },
+  })
+}
 
 describe('ActionsCarousel', () => {
+  let store: MockStoreEnhanced<{}>
+  beforeEach(() => {
+    store = getStore(true)
```
satish-ravi, line 27: nit: `getStore(true)` is small enough that this beforeEach seems like an overkill 😄

satish-ravi (approved):

PR #3800, opened 2023-05-24
feat(nfts): add info carousel

MuckT:
### Description

- Added the dependency [`react-native-fast-image`](https://www.npmjs.com/package/react-native-fast-image) for quicker loading and better device caching when compared the [`Image`](https://reactnative.dev/docs/image) provided by react-native. Originally implemented using Image; however, after switching to react-native-fast-image the performance and caching improvements were astonishing.
- Added the `NftsInfoCarousel.tsx` screen
- Added nftScreens to `src/navigator/Navigator.tsx`.
- Adds `BackChevronStatic.tsx` existing chevron has some centering issues.
- Feature gate added in #3814
- `NftsLoadErrorScreen` covered in #3818

#### Video of Multiple Nfts in carousel

https://github.com/valora-inc/wallet/assets/26950305/24810640-c920-4727-be05-db949a0dd468

### Test plan

- [x] Tested locally on iOS
- [x] Tested in poor network conditions
- [x] Add unit tests for nft carousel
- [x] Address PR feedback

### Related issues

- Fixes RET-699

### Backwards compatibility

Yes


src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,367 @@
...
+      >
+        {loading && <ThumbnailImagePlaceholder />}
+      </FastImage>
+    </Touchable>
+  )
+}
+
+export default function NftsInfoCarousel({ route }: Props) {
+  const { nfts } = route.params
+  const [activeNft, setActiveNft] = useState((nfts && nfts[0]) ?? null)
```
kathaypacific, line 90: i think you can just do `nfts[0] ?? null` since `nft` is required

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,367 @@
...
+  }
+
+  // Some components that require parent state defined
+  const PlatformSpecificTopBarButtons = () => {
+    return (
+      <View style={styles.topBarButtonsContainerWrapper}>
+        <View style={styles.topBarButtonsContainer}>
+          {Platform.OS === 'ios' && (
+            <TopBarIconButton
+              onPress={() => navigateBack()}
```
kathaypacific, line 108: nit: `onPress={navigateBack}`

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,367 @@
...
+  const { t } = useTranslation()
+
+  const android_ripple = {
+    color: colors.gray2,
+    foreground: true,
+    borderless: true,
+  }
+
+  // Some components that require parent state defined
+  const PlatformSpecificTopBarButtons = () => {
```
kathaypacific, line 102: i also feel like this is a weird place for this component, this seems like it should be in the components folder. what's the diff between the two buttons? (like what happens if we use the `TopBarIconButton` on android, since we already do this pretty much across the app?

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,367 @@
...
+
+  const MainImagePlaceholder = () => {
+    return (
+      <SkeletonPlaceholder
+        borderRadius={8}
+        backgroundColor={colors.gray2}
+        highlightColor={colors.white}
+      >
+        <View
+          style={{
```
kathaypacific, line 137: use the stylesheet (there are also other places where we should remove the inline styles)

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,367 @@
...
+    return (
+      <View style={styles.nftImageCarouselContainer}>
+        <ScrollView
+          showsHorizontalScrollIndicator={false}
+          horizontal={true}
+          contentContainerStyle={styles.carouselScrollViewContentContainer}
+          style={styles.nftImageCarousel}
+          testID="NftsInfoCarousel/NftImageCarousel"
+        >
+          {/* Only display Nfts with metadata */}
```
kathaypacific, line 157: is it confusing that we're only displaying part of what the backend returns to the front end? i kind of assumed that we'd have a state that displays null metadata, we can still display the image using `media` when the metadata is unavailable?

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,367 @@
...
+          {nfts
+            .filter((nft) => nft.metadata)
+            .map((nft) => {
+              return (
+                <View
+                  key={`${nft.contractAddress}-${nft.tokenId}`}
+                  style={styles.nftPreviewImageSharedContainer}
+                  testID={`NftsInfoCarousel/NftThumbnail/${nft.contractAddress}-${nft.tokenId}`}
+                >
+                  <NftThumbnail nft={nft} setActiveNft={setActiveNft} activeNft={activeNft} />
```
kathaypacific, line 167: nit: i think it's nice for components to know as little as possible. `NftThumbnail` doesn't really need to know which nft is active, it just needs to know if it should display itself as active or not. so perhaps a nicer props interface is `isActive` and `onPress` rather than `activeNft / setActiveNft`

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,367 @@
...
+  )
+}
+
+const NftThumbnail = ({
+  nft,
+  setActiveNft,
+  activeNft,
+}: {
+  nft: Nft
+  setActiveNft: any
```
kathaypacific, line 58: remove `any`

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,367 @@
...
+  // Display error screen no Nfts provided
+  if (nfts.length === 0) {
+    return <NftsLoadErrorScreen />
+  }
+
+  return (
+    <SafeAreaView testID="NftsInfoCarousel">
+      <PlatformSpecificTopBarButtons />
+      <ScrollView>
+        {activeNft.metadata && (
```
kathaypacific, line 185: i think the render is a little hard to read because of all the conditions. it seems like we've made the assumption that we want to hide nfts with no metadata in the `NftImageCarousel`, so then we should carry that logic across the whole screen. fwiw i don't really love this and would prefer to show an explicit null metadata state. but if we wanted to have this metadata requirement, i think we should create a new variable that filters out nfts that we don't want to render immediately in `NftsInfoCarousel`, and use this in the render and child components. this means you can remove the sporadic checks and chain operators everywhere

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,367 @@
...
+                <Text style={styles.attributeTitle}>{attribute.trait_type}</Text>
+                <Text style={styles.attributeText}>{attribute.value}</Text>
+              </View>
+            ))}
+          </View>
+        )}
+        {/* This should be dynamic based on Network in the future. Always show as fallback */}
+        <View style={styles.sectionContainer}>
+          <Touchable
+            onPress={() =>
```
kathaypacific, line 244: i think generally it's a good idea to avoid inline functions for top level components, for performance but also we'll need to add analytics later so this function will grow in complexity

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,367 @@
...
+                width: variables.width,
+              },
+              isLoading && { borderRadius: 8 },
+            ]}
+            source={{
+              uri: activeNft.media.find((media) => media.raw === activeNft.metadata?.image)
+                ?.gateway,
+            }}
+            onLoad={(e) => {
+              setScaledHeight(
```
kathaypacific, line 200: what you are trying to do is layout the image so that you can determine the aspect ratio, then resize it to fit the screen right? this seems a bit complex to have a state variable and layout function...

could you do that with styles? like set the width as `variables.width` and use [resizeMode](https://reactnative.dev/docs/image#resizemode)?

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,367 @@
...
+          contentContainerStyle={styles.carouselScrollViewContentContainer}
+          style={styles.nftImageCarousel}
+          testID="NftsInfoCarousel/NftImageCarousel"
+        >
+          {/* Only display Nfts with metadata */}
+          {nfts
+            .filter((nft) => nft.metadata)
+            .map((nft) => {
+              return (
+                <View
```
kathaypacific, line 162: can the NftThumbnail provide the border radius, and we remove this View?

kathaypacific (commented):
look good... but this PR is still 50% larger than the maximum size we agreed on (600 new lines vs 400...) i tried to review thoroughly and not sure if i properly finished but i've exceeded the time i have so might have new comments when i return to this

src/nfts/NftsInfoCarousel.tsx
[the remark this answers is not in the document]
MuckT: @jeanregisser I can move this to `src/navigator/Headers.tsx`. I tried using `BackChevron.tsx` instead of `BackChevronStatic.tsx` as seen in the image below. I think fixing the BackChevron centering would be a good task in it's own PR and when that's done `BackChevronStatic.tsx` could be removed.

#### Using `BackChevron.tsx`
<p align="left">

</p>

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,367 @@
...
+  const aspectRatio = originalWidth / originalHeight
+  return targetWidth / aspectRatio
+}
+
+const NftsLoadErrorScreen = () => {
+  const { t } = useTranslation()
+  return (
+    <SafeAreaView
+      testID="NftsInfoCarousel/NftsLoadErrorScreen"
+      style={{ flex: 1, justifyContent: 'center', alignItems: 'center' }}
```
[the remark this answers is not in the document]
MuckT: This is just a placeholder for until https://github.com/valora-inc/wallet/pull/3818.

src/nfts/NftsInfoCarousel.tsx
MuckT: I've removed the filtering of Nfts with null metadata. To allow toggling nfts without metadata I had to add some conditional rendering the `NftThumbnail` and a placeholder with a info icon: https://github.com/valora-inc/wallet/pull/3800/commits/7dee7e1bfb5c58e1af48376810814b917175d4e0.

locales/base/translation.json
```diff
@@ -1722,5 +1722,12 @@
     "send": "Send",
     "swap": "Swap",
     "withdraw": "Withdraw"
+  },
+  "nftInfoCarousel": {
+    "viewOnCeloExplorer": "View on Celo Explorer",
+    "attributes": "Attributes",
+    "description": "Description",
+    "noNftsFound": "No NFTs found",
+    "nftImageLoadError": "There is an issue loading this Nft"
```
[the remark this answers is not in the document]
MuckT: `NFT`, fixed in https://github.com/valora-inc/wallet/pull/3800/commits/91c1266d801c09f89471dfa6b631bc57497aa3d7!

src/nfts/NftsInfoCarousel.tsx
MuckT: I removed most of the inline styles. The ones I left in https://github.com/valora-inc/wallet/pull/3800/commits/b6fc9b3d175c7155c6c3ad03a4007d42be9107a7 were because I started noticing a flicker on the main image after moving the `width` and `zIndex` to stylesheets 🤔

src/nfts/NftsInfoCarousel.tsx
MuckT: There's no support for height="auto" in RN, so I had to use this instead.

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,350 @@
...
+    )
+  }
+
+  // Display error screen no Nfts provided
+  if (nfts.length === 0) {
+    return <NftsLoadErrorScreen />
+  }
+
+  return (
+    <SafeAreaView edges={['top']} style={styles.safeAreaView} testID="NftsInfoCarousel">
```
kathaypacific, line 181: any reason why we don't want safe area on the other edges too? i'm always confused by this - when is it nicer for content to be displayed overlapping a bottom bar on iOS for example?

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,350 @@
...
+                `Error loading Nft image for Nft contractAddress: ${activeNft.contractAddress} tokenId: ${activeNft.tokenId}`
+              )
+            }
+            resizeMode={FastImage.resizeMode.contain}
+          >
+            {isLoading && <MainImagePlaceholder />}
+          </FastImage>
+        )}
+        {/* This could happen if the indexer is experiencing issues with particular nfts */}
+        {!activeNft.metadata && (
```
kathaypacific, line 214: nit: we could do `{activeNft.metadata ? (...) : (...)}` to make this easier to read?

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,350 @@
...
+          <View style={styles.nftImageLoadingErrorContainer}>
+            <Text style={{ color: colors.light }}>{t('nftInfoCarousel.nftImageLoadError')}</Text>
+          </View>
+        )}
+
+        {nfts.length > 1 && <NftImageCarousel />}
+        <View style={styles.sectionContainer}>
```
kathaypacific, line 221: do you think a structure like this is easier to read?
```
{activeNft.metadata && (
  <>
    {activeNft.metadata.title}
    {activeNft.metadata.description}
    {activeNft.metadata.attributes && CodeToRenderAttributes}
  </>
)}
```
the current structure with all the ternaries makes it a little hard to understand which elements are likely to be there or missing, since everything looks optional

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,350 @@
...
+  nftThumbnailSelected: {
+    height: 40,
+    width: 40,
+  },
+  nftThumbnailShared: {
+    borderRadius: 8,
+  },
+  nftThumbnailSharedContainer: {
+    borderRadius: 8,
+    marginRight: 8,
```
kathaypacific, line 314: nit: use `Spacing`

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,350 @@
...
+    overflow: 'hidden',
+  },
+  nftThumbnailUnSelected: {
```
kathaypacific, line 317:
```suggestion
  nftThumbnailUnselected: {
```

kathaypacific (commented):
☀️

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,350 @@
...
+    <Touchable
+      style={[
+        isActive ? styles.nftThumbnailSelected : styles.nftThumbnailUnSelected,
+        error && styles.errorThumbnail,
+      ]}
+      borderless={false}
+      onPress={onPress}
+      testID={`NftsInfoCarousel/NftThumbnail/${nft.contractAddress}-${nft.tokenId}`}
+    >
+      {error ? (
```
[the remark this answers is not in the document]
MuckT: I was confused why media was an array first at too, but it makes more sense given to context here: https://github.com/valora-inc/nft-indexer/pull/331. Right now it's just an array with one object but it could also contain entries for animation_url, video_url, mp4_url etc.

src/nfts/NftsInfoCarousel.tsx
MuckT: We have something similar with `src/home/WalletHome.tsx`. I tried without and it's a bit strange that all the content is scrolling except a sliver at the bottom of the screen.

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,350 @@
...
+          <View style={styles.sectionContainer}>
+            <Text style={styles.subSectionTitle}>{t('nftInfoCarousel.description')}</Text>
+            <Text style={styles.text}>{activeNft.metadata?.description}</Text>
+          </View>
+        )}
+        {activeNft.metadata?.attributes && (
+          <View style={styles.sectionContainer}>
+            <Text style={styles.subSectionTitle}>{t('nftInfoCarousel.attributes')}</Text>
+            {activeNft.metadata?.attributes.map((attribute, index) => (
+              <View key={index} style={styles.attributesContainer}>
```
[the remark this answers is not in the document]
MuckT: Looking at the [OpenSea's standard](https://docs.opensea.io/docs/metadata-standards#date-traits) it looks like there might be cases where there is no trait_type and value isn't unique so this might be the best we can do.
> If you don't want to have a trait_type for a particular trait, you can include just a value in the trait and it will be set as a generic property
```
{
  "value": "Happy"
}
```

src/icons/BackChevronStatic.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react'
```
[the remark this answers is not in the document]
jeanregisser: > A centering issue is mentioned in your description, can we fix the existing component and avoid a new component?

Ping on this.

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,369 @@
...
+  }
+
+  // Display error screen no Nfts provided
+  if (nfts.length === 0) {
+    return <NftsLoadErrorScreen />
+  }
+
+  return (
+    <SafeAreaView edges={['top']} style={styles.safeAreaView} testID="NftsInfoCarousel">
+      <ScrollView>
```
jeanregisser, line 187: Do we account for the bottom inset needed to fully view and interact with the last element in the ScrollView on devices with the home bar?

MuckT: @Jean would this be an Android device with a bottom bar?

jeanregisser: Just on iOS with the home bar.

MuckT: There is some marginBottom added to the explorer link via the `sectionContainerLast` style so that it displays above the bottom inset when scrolled similar to the readMoreButton in `src/exchange/CeloNewsFeed.tsx`. Is there a better way to do this?

### iPhone 16 Long Content
https://github.com/valora-inc/wallet/assets/26950305/ff2352c4-6619-4112-93dd-61493afddfbc

### iPhone 16 Short Content

https://github.com/valora-inc/wallet/assets/26950305/d308e02d-3989-4cd2-855a-7eeb93da7319

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,369 @@
...
+        </FastImage>
+      )}
+    </Touchable>
+  )
+}
+interface MainImagePlaceholderProps {
+  height: number
+}
+
+function MainImagePlaceholder({ height }: MainImagePlaceholderProps) {
```
kathaypacific, line 104: nit: i think that this component could be combined with ThumbnailImagePlaceholder, where it defaults to the thumbnail height unless a height is passed in?

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -0,0 +1,369 @@
...
+          <NftImageCarousel
+            nfts={nfts}
+            activeNft={activeNft}
+            handleOnPress={handleThumbnailPress}
+          />
+        )}
+        {/* Nft Details */}
+        {activeNft.metadata && (
+          <>
+            {activeNft.metadata?.name && (
```
kathaypacific, line 234: super nit: do you need all these chain operators? you check for metadata already, do you still need to do `metadata?.name` here (and also for other fields below?)

kathaypacific (approved):
LGTM, worth checking that @jeanregisser's comments are addressed before merging :)

src/nfts/NftsInfoCarousel.tsx
MuckT: There are some NFTs which do not have all metadata fields required by ERC-721. Unfortunately, [YourValoraPersona](https://explorer.celo.org/mainnet/token/0xc49F7d91b18731edC54eC5693aD43bb1dDf6553C/instance/48903/metadata) is an example that does not include a description field.

jeanregisser (approved):
🚀

src/nfts/NftsInfoCarousel.tsx
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/3800/commits/6a0419157af35e7f6b0374818be5bcdb6c1e356d!

PR #3806, opened 2023-05-25
chore(analytics): add properties on why a user selects CICO option

satish-ravi:
### Description

Add properties to make analysis easier, see ACT-665 for questions we want answered with simple queries. Mostly copied from #3044 

### Test plan

Unit tests

### Related issues

- Fixes ACT-665

### Backwards compatibility

N/A


src/fiatExchanges/types.ts
```diff
@@ -7,3 +10,16 @@ export enum SelectProviderExchangesLink {
   ExternalExchangesScreen = 'ExternalExchangesScreen',
   ExchangeQRScreen = 'ExchangeQRScreen',
 }
+
+export interface ProviderSelectionAnalyticsData {
```
jh2oman, line 14: It seems to me that `flow` would be another natural value to put in this set. Any reason why its left out?

src/fiatExchanges/SelectProvider.tsx
```diff
@@ -230,6 +241,17 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
...
   }
 
+  const analyticsData = getProviderSelectionAnalyticsData({
+    normalizedQuotes,
+    legacyMobileMoneyProviders,
+    exchangeRates,
+    tokenInfo,
+    centralizedExchanges: exchanges,
+    coinbasePayAvailable: coinbasePayVisible,
+    transferCryptoAmount: cryptoAmount.toFixed(2),
```
jh2oman, line 251: The .toFixed(2) would probably be more appropriate inside `getProviderSelectionAnalyticsData` for consistency

jh2oman (approved):

src/fiatExchanges/types.ts
satish-ravi: We don't need this for coinbase pay as it's cash in only, but I agree its good to include. I tried doing this, but it turns out to be a big refactor (causes cyclic imports on types, CICOFlow is defined in utils, ProviderSelectionAnalyticsData is in types). I think we can defer this until coinbase pay events need flow as well.

PR #3810, opened 2023-05-26
chore(swaps): add more analytics for debugging

kathaypacific:
### Description

We are still getting many errors that indicate that we are allowing the user to proceed with a swap amount greater than their balance. It's not clear why this still happens, because we have a (high) fee estimation in place already.

Ideally I'd like to have data to be able to say "oh yes, this swap was expected to fail because the swap amount was X, required fee was Y, and X + Y > the total balance". i think then we could go back and tweak the logic required for X, when we have a better idea of what X and Y are.

Sadly during my debugging today I think some errors are being thrown at the calculation for the gas fee 🫠 so we may not have the value for Y.

This PR adds data for 
- total "from" token balance at the time of executing the swap
- context id's used for the swap transactions so that we can also check the TransactionEvents which contains some information about gas / errors etc. If an error is thrown when calculating the gas fee, then we will be able to tell (from the lack of the fee event).
- add fee currency to the estimation and error TransactionEvents, to see if there's any correlation of token with errors.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y


src/analytics/Properties.tsx
```diff
@@ -713,6 +714,7 @@ interface TransactionEventsProperties {
   [TransactionEvents.transaction_exception]: {
     txId: string
     error: string
+    gasTokenAddress?: string
```
jeanregisser, line 717: For consistency:
```suggestion
    feeCurrencyAddress?: string
```

src/transactions/contract-utils.ts
```diff
@@ -73,10 +73,11 @@ interface EstimatedGas {
   type: SendTransactionLogEventType.EstimatedGas
   gas: number
   prefilled: boolean
+  gasTokenAddress: string | undefined
```
jeanregisser, line 76:
```suggestion
  feeCurrencyAddress: string | undefined
```

src/transactions/contract-utils.ts
```diff
@@ -73,10 +73,11 @@ interface EstimatedGas {
...
 }
 
-function EstimatedGas(gas: number, prefilled: boolean): EstimatedGas {
-  return { type: SendTransactionLogEventType.EstimatedGas, gas, prefilled }
+function EstimatedGas(gas: number, prefilled: boolean, gasTokenAddress?: string): EstimatedGas {
```
jeanregisser, line 79:
```suggestion
function EstimatedGas(gas: number, prefilled: boolean, feeCurrencyAddress?: string): EstimatedGas {
```

src/transactions/contract-utils.ts
```diff
@@ -109,10 +110,11 @@ function Failed(error: Error): Failed {
 interface Exception {
   type: SendTransactionLogEventType.Exception
   error: Error
+  gasTokenAddress?: string
 }
 
-function Exception(error: Error): Exception {
-  return { type: SendTransactionLogEventType.Exception, error }
+function Exception(error: Error, gasTokenAddress?: string): Exception {
```
jeanregisser, line 116:
```suggestion
function Exception(error: Error, feeCurrencyAddress?: string): Exception {
```

src/swap/saga.ts
```diff
@@ -69,9 +64,36 @@ export function* swapSubmitSaga(action: PayloadAction<SwapInfo>) {
...
+    action.payload.userInput.updatedField === Field.TO
+      ? ('buyAmount' as const)
+      : ('sellAmount' as const)
   const amount = action.payload.unvalidatedSwapTransaction[amountType]
 
+  const tokenBalances: TokenBalance[] = yield select(swappableTokensSelector)
+  const fromTokenBalance =
+    tokenBalances
+      .find((token) => token.address === sellTokenAddress)
+      ?.balance.shiftedBy(WEI_DECIMALS)
```
jeanregisser, line 77: We need to use the actual decimals used by the token.

jeanregisser (commented):
Great!

silasbw (approved):
:hot_pepper:

PR #3815, opened 2023-05-30
docs(android-emulator): update android version in WALLET.md

cajubelt:
### Description

docs said to use 29, which looks old. I updated them to use 31 since that is the target version in build.gradle

### Test plan

tried the updated docs and the app starts successfully on the emulator

### Related issues

na

### Backwards compatibility

na


MuckT (commented):
`Pixel_API_29_AOSP_x86_64` should also be changed in `package.json`, `e2e/README.md`, `e2e/conf/avd_conf.ini`. The `e2e/conf/ci_avd_conf.ini` can be removed as it is [no longer used](https://github.com/valora-inc/infrastructure/blob/ea7592be37b4d8cac018a4f2aadc60dd20312550/mac-stadium/components/install_android_emulators.sh#L20).

PR #3818, opened 2023-05-31
feat(nfts): nft loading error screen

MuckT:
### Description

An error screen as specified in figma when the Nft transaction details returns an empty array.

### Video Full Page Error iOS
https://github.com/valora-inc/wallet/assets/26950305/5cee4132-75e1-47a4-86bd-258bc3df3458

#### Screenshot Carousel Individual Nft image load error iOS
<p align="left">

</p>

### Test plan

- [x] Tested locally on iOS.
- [x] Tested locally on Android.
- [x] Unit tests added.
- [x] Tested with poor network conditions.

### Related issues

- #3800 
### Backwards compatibility

Yes


src/nfts/NftsInfoCarousel.tsx
```diff
@@ -173,8 +162,8 @@ export default function NftsInfoCarousel({ route }: Props) {
     )
   }
 
-  // Display error screen no Nfts provided
-  if (nfts.length === 0) {
+  // Catch All Error Screen
+  if (!nfts || nfts.length === 0) {
```
kathaypacific, line 166: when would `nfts` be null when it's required as a nav prop?

src/styles/colors.tsx
```diff
@@ -23,6 +23,7 @@ export enum Colors {
   gray1 = '#F8F9F9',
   light = '#FFFFFF',
   purple = '#D39CFE',
+  pink = '#FFC8C8',
```
kathaypacific, line 26: is this going to be part of the color palette officially? if it's for one time use, i think we should just keep it in the component styles

kathaypacific (approved):
looks good!

PR #3821, opened 2023-05-31
feat(analytics): track taps on assets

jeanregisser:
### Description

This tracks taps on assets.

Tapping doesn't do anything for the user, but allows us to see if users are doing it. Maybe expecting a detail screen.

### Test plan

- Added unit test

### Related issues

- Fixes RET-713

### Backwards compatibility

Yes


kathaypacific (approved):
👯‍♂️

PR #3826, opened 2023-06-01
refactor: use useAsyncCallback in fetching swap quote

kathaypacific:
### Description

As the title, just a refactor with no change in functionality.

### Test plan

Manually checked that everything works the same.

### Related issues

n/a

### Backwards compatibility

Y


src/swap/useSwapQuote.ts
```diff
@@ -20,88 +21,98 @@ const useSwapQuote = () => {
...
+      const response = await fetch(requestUrl)
+
+      if (!response.ok) {
+        throw new Error(await response.text())
+      }
 
-    requestUrlRef.current = requestUrl
+      return {
+        response,
+        requestParams: {
```
kathaypacific, line 73: from the `onSuccess` callback i couldn't figure out how to access the function arguments :/ open for suggestions on how better to do this part

src/swap/useSwapQuote.ts
```diff
@@ -20,88 +21,83 @@ const useSwapQuote = () => {
...
-    setFetchSwapQuoteError(false)
+  const refreshQuote = useAsyncCallback(
+    async (
+      fromToken: TokenBalance,
+      toToken: TokenBalance,
+      swapAmount: ParsedSwapAmount,
+      updatedField: Field
+    ) => {
+      if (!swapAmount[updatedField].gt(0)) {
+        setExchangeRate(null)
```
jeanregisser, line 38: setters should be done in `onSuccess` or `onError`, so they don't fire if the component has been unmounted.

jeanregisser (commented):
Thanks!

PR #3827, opened 2023-06-01
chore: refactor swaps max celo warning to be more generic

kathaypacific:
### Description

As the title, this is just a refactor to prep for the new warnings.

### Test plan

n/a

### Related issues

- Fixes RET-745

### Backwards compatibility

Y


locales/base/translation.json
```diff
@@ -1603,7 +1603,7 @@
     "swapTo": "TO",
     "insufficientFunds": "Insufficient {{token}} balance. Try a smaller amount or choose a different asset.",
     "fetchSwapQuoteFailed": "Sorry for the delay! It's taking longer than usual to get the best exchange rate. Please try again in a few minutes.",
-    "disclaimer": "Swaps powered by partners on Celo. \n<0>Learn More.</0>",
+    "disclaimer": "Swaps powered by partners on Celo \n<0>Learn more</0>",
```
kathaypacific, line 1606: this change was a request by Reed

src/swap/SwapScreen.tsx
```diff
@@ -265,9 +266,7 @@ export function SwapScreenSection({ showDrawerTopNav }: { showDrawerTopNav: bool
       }))
     }
 
-    if (fieldType === Field.FROM) {
-      setShowMaxSwapAmountWarning(false)
-    }
```
kathaypacific, line 270: we should really clear this warning when the "to" amount changes too, since that would also change the "from" amount

jeanregisser (approved):
👍

PR #3834, opened 2023-06-02
feat(quote): make settlement time dynamic

anthonydaoud:
### Description

Currently, the time estimation for a FiatConnect quote is set to fixed values. This PR makes it so that the time estimation is dynamic and based on data sent from the quote provider. 

SelectPaymentDynamicTime

### Test plan

Added unit tests and modified existing unit test

### Related issues

- Fixes ACT-366

### Backwards compatibility

Yes.


src/fiatExchanges/quotes/FiatConnectQuote.ts
```diff
@@ -152,12 +160,70 @@ export default class FiatConnectQuote extends NormalizedQuote {
...
+          settlementTime: SettlementTime.LESS_THAN_X_DAYS,
+          upperBound,
+        }
+      }
+
+      return {
+        settlementTime: SettlementTime.X_TO_Y_DAYS,
+        lowerBound,
+        upperBound,
+      }
```
cajubelt, line 187: it may not be worth the effort, but this does look almost identical between the two functions. There's probably a way to make it more DRY

cajubelt: I'm not that concerned about it though

src/fiatExchanges/quotes/FiatConnectQuote.ts
```diff
@@ -152,12 +160,70 @@ export default class FiatConnectQuote extends NormalizedQuote {
...
+    if (!fiatAccountInfo?.settlementTimeUpperBound) {
+      // payment method can only be bank or fc mobile money
+      return this.getPaymentMethod() === PaymentMethod.Bank
+        ? DEFAULT_BANK_SETTLEMENT_ESTIMATION
+        : DEFAULT_MOBILE_MONEY_SETTLEMENT_ESTIMATION
+    }
+
+    const daysToSettlementEstimation = (
+      lowerBound: number,
+      upperBound: number
```
cajubelt, line 174: would be nice if these parameter names included units

cajubelt: I'd probably opt for destructured assignment too, to prevent bugs where the order of inputs are swapped. it's an intuitive ordering as is, but it wouldn't hurt.

src/fiatExchanges/quotes/FiatConnectQuote.ts
```diff
@@ -152,12 +160,70 @@ export default class FiatConnectQuote extends NormalizedQuote {
...
+
+      return {
+        settlementTime: SettlementTime.X_TO_Y_DAYS,
+        lowerBound,
+        upperBound,
+      }
+    }
+
+    const hoursToSettlementEstimation = (
+      lowerBound: number,
```
cajubelt, line 191: same here with units in the parameter names

src/fiatExchanges/quotes/FiatConnectQuote.ts
```diff
@@ -152,12 +160,70 @@ export default class FiatConnectQuote extends NormalizedQuote {
...
+
+      return {
+        settlementTime: SettlementTime.X_TO_Y_HOURS,
+        lowerBound,
+        upperBound,
+      }
+    }
+
+    const lowerBound = fiatAccountInfo.settlementTimeLowerBound ?? '0'
+    const lowerBoundInHours = Math.ceil(parseInt(lowerBound) / SECONDS_IN_HOUR)
```
cajubelt, line 215:
```suggestion
    const lowerBoundInHours = fiatAccountInfo.settlementTimeLowerBound ? Math.ceil(parseInt(fiatAccountInfo.settlementTimeLowerBound) / SECONDS_IN_HOUR) : 0
```
code golf. you don't need the extra operations if it's falsey to begin with

src/fiatExchanges/quotes/constants.ts
```diff
@@ -19,7 +19,47 @@ export const DEFAULT_ALLOWED_VALUES: DefaultAllowedValues = {
...
+    }
+  | {
+      settlementTime: SettlementTime.LESS_THAN_X_DAYS
+      upperBound: number
+    }
+  | {
+      settlementTime: SettlementTime.X_TO_Y_DAYS
+      lowerBound: number
+      upperBound: number
+    }
```
cajubelt, line 50:
```suggestion
      settlementTime: SettlementTime.LESS_THAN_X_HOURS | SettlementTime.LESS_THAN_X_DAYS
      upperBound: number
    }
  | {
      settlementTime: SettlementTime.X_TO_Y_HOURS | SettlementTime.X_TO_Y_DAYS
      lowerBound: number
      upperBound: number
    }
```
code golf

cajubelt: btw LESS_THAN_ONE_HOUR doesn't seem necessary as its own case. it's just a special case of LESS_THAN_X_HOURS

src/fiatconnect/TransferStatusScreen.tsx
```diff
@@ -133,12 +133,31 @@ function SuccessOrProcessingSection({
...
+      case SettlementTime.X_TO_Y_HOURS:
+      case SettlementTime.X_TO_Y_DAYS:
+        return t(DESCRIPTION_STRINGS[settlementEstimation.settlementTime], {
+          lowerBound: settlementEstimation.lowerBound,
+          upperBound: settlementEstimation.upperBound,
+        })
+      default:
+        // Should not be reachable
+        return t('fiatConnectStatusScreen.success.baseDescription')
+    }
```
cajubelt, line 154:
```suggestion
  const getSettlementTimeString = ({ settlementTime, lowerBound, upperBound }: SettlementEstimation) => {
    return t(DESCRIPTION_STRINGS[settlementTime], { lowerBound, upperBound })
```
idk if this even compiles but if it does it would be nice. the switch seems more verbose than we might need

src/fiatExchanges/quotes/constants.ts
cajubelt: actually I'm realizing we probly need it so we can use a singular "hour" instead of "hours". bummer

cajubelt: might be worth leaving a code comment explaining that

cajubelt (commented):
looks close-- see my comment in slack about syncing with product teammates on this though

cajubelt (commented):
looks close-- see my comment in slack about syncing with product teammates on this though

src/fiatExchanges/quotes/FiatConnectQuote.test.ts
```diff
@@ -229,22 +233,155 @@ describe('FiatConnectQuote', () => {
   })
 
   describe('.getTimeEstimation', () => {
-    it('returns 1-3 days for bank account', () => {
+    it('returns default for bank account when no bounds are present', () => {
+      const quoteData = _.cloneDeep(mockFiatConnectQuotes[1]) as FiatConnectQuoteSuccess
+      quoteData.fiatAccount.BankAccount = {
+        ...quoteData.fiatAccount.BankAccount!,
```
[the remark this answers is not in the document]
anthonydaoud: This gives a typescript error, so I'll stick with the `!` operator

src/fiatExchanges/quotes/FiatConnectQuote.ts
```diff
@@ -152,12 +160,70 @@ export default class FiatConnectQuote extends NormalizedQuote {
...
+      // payment method can only be bank or fc mobile money
+      return this.getPaymentMethod() === PaymentMethod.Bank
+        ? DEFAULT_BANK_SETTLEMENT_ESTIMATION
+        : DEFAULT_MOBILE_MONEY_SETTLEMENT_ESTIMATION
+    }
+
+    const daysToSettlementEstimation = (
+      lowerBound: number,
+      upperBound: number
+    ): SettlementEstimation => {
```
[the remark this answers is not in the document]
anthonydaoud: I added the checks to the start of the `getTimeEstimation` function, which should ensure the days/hours to settlement estimation functions receive good inputs

src/fiatExchanges/quotes/FiatConnectQuote.ts
anthonydaoud: Was struggling to cleanly make it more DRY, so am sticking with this for now

src/fiatExchanges/quotes/FiatConnectQuote.ts
anthonydaoud: Good call, added both those suggestions

src/fiatExchanges/quotes/constants.ts
anthonydaoud: Yeah the singular hour is the reason for the extra enum key, added a comment

src/fiatconnect/TransferStatusScreen.tsx
anthonydaoud: It doesn't compile unfortunately, I agree that it's a bit verbose as-is

src/fiatExchanges/quotes/FiatConnectQuote.ts
```diff
@@ -152,12 +208,37 @@ export default class FiatConnectQuote extends NormalizedQuote {
...
+    const fiatAccountInfo = this.quote.fiatAccount[this.getFiatAccountType()]
+
+    const lowerBound = fiatAccountInfo?.settlementTimeLowerBound
+      ? parseInt(fiatAccountInfo?.settlementTimeLowerBound)
+      : 0
+    const upperBound = fiatAccountInfo?.settlementTimeUpperBound
+      ? parseInt(fiatAccountInfo?.settlementTimeUpperBound)
+      : 0
+    if (
+      !fiatAccountInfo?.settlementTimeUpperBound ||
```
jophish, line 221: I would imagine that after setting the `upperBound` and `lowerBound` variables, we should avoid directly accessing `fiatAccountInfo.settlmentTimeUpperBound` again. Can we just roll this first check into the `upperBound < 0` check by doing `upperBound <= 0` instead? It should be OK to keep the `lowerBound < 0` check as-is, since we're still capable of giving a non-default estimation when the provider doesn't return a lower bound.

src/fiatExchanges/quotes/FiatConnectQuote.ts
```diff
@@ -152,12 +208,37 @@ export default class FiatConnectQuote extends NormalizedQuote {
...
+      lowerBound < 0 ||
+      upperBound < 0 ||
+      lowerBound > upperBound
+    ) {
+      // payment method can only be bank or fc mobile money
+      return this.getPaymentMethod() === PaymentMethod.Bank
```
jophish, line 227: This is more of a nit, but I get a little nervous when assumptions are made about what type a value would be in a way that won't be caught by the TS compiler if that value were able to take on more types; e.g., once `getPaymentMethod` can plausibly return payment types other than `Bank` and `FiatConnectMobileMoney`, it's very probable that this line wouldn't get updated to reflect that.

Kind of a pain to deal with since `PaymentMethod` encompasses payment types that could _never_ be on a FiatConnectQuote, and I'm not sure the best way to do this such that TS will complain if a value was added to the `PaymentMethod` enum and _wasn't_ explicitly dealt with here. For now maybe add a TODO: to ensure that this gets updated once more payment types are possible?

src/fiatconnect/TransferStatusScreen.tsx
```diff
@@ -133,12 +133,31 @@ function SuccessOrProcessingSection({
...
+        return t(DESCRIPTION_STRINGS[settlementEstimation.settlementTime], {
+          upperBound: settlementEstimation.upperBound,
+        })
+      case SettlementTime.X_TO_Y_HOURS:
+      case SettlementTime.X_TO_Y_DAYS:
+        return t(DESCRIPTION_STRINGS[settlementEstimation.settlementTime], {
+          lowerBound: settlementEstimation.lowerBound,
+          upperBound: settlementEstimation.upperBound,
+        })
+      default:
```
jophish, line 151: Should we log a warning or something here if this should really never happen?

src/fiatconnect/TransferStatusScreen.tsx
```diff
@@ -133,12 +133,31 @@ function SuccessOrProcessingSection({
     | FiatExchangeEvents.cico_fc_transfer_success_view_tx
     | FiatExchangeEvents.cico_fc_transfer_processing_view_tx
 
+  const getSettlementTimeString = (settlementEstimation: SettlementEstimation) => {
+    switch (settlementEstimation.settlementTime) {
+      case SettlementTime.LESS_THAN_ONE_HOUR:
+        return t(DESCRIPTION_STRINGS[SettlementTime.LESS_THAN_ONE_HOUR])
```
jophish, line 139: nit: add a test to ensure this line gets covered?

src/fiatconnect/TransferStatusScreen.tsx
```diff
@@ -133,12 +133,31 @@ function SuccessOrProcessingSection({
     | FiatExchangeEvents.cico_fc_transfer_success_view_tx
     | FiatExchangeEvents.cico_fc_transfer_processing_view_tx
 
+  const getSettlementTimeString = (settlementEstimation: SettlementEstimation) => {
```
jophish, line 136: The semantics of this switch and the switch in `PaymentMethodSection.tsx` look identical to me, save for the actual translation strings that are mapped to by the `DESCRIPTION_STRINGS` and `SETTLEMENT_TIME_STRINGS` objects respectively. Could these be combined and parameterized by the translation string mapping?

src/fiatExchanges/quotes/FiatConnectQuote.ts
cajubelt: one option is to define a default that is not payment method-specific, and default to that if the payment method is not bank or mobile money

cajubelt (approved):

src/fiatExchanges/quotes/FiatConnectQuote.ts
anthonydaoud: Good call, updated the check

PR #3842, opened 2023-06-05
feat(nfts): add nfts to blockchain api query

MuckT:
### Description

Broken out backend work needed to support the components in #3800. 
- Added new Nft types in `src/nfts/types.ts`
- Updated the transaction query helper in `src/transactions/feed/queryHelper.ts` to query for Nfts.
- Added mock nfts for testing.

### Test plan

### Related issues

- #3800

### Backwards compatibility

Yes


test/values.ts
```diff
@@ -1261,3 +1262,54 @@ export const mockLegacyMobileMoneyProvider: LegacyMobileMoneyProvider = {
...
+    id: null,
+    image: 'https://example.com/3',
+    name: `${mockName}.fizzBuzz`,
+  },
+  ownerAddress: mockAccount,
+  tokenId: '3',
+  tokenUri: 'https://example.com/3',
+}
+
+export const mockNft3: Nft = {
```
kathaypacific, line 1310: it'd be nice to name these more semantically so that it's clearer what the difference between them are...like `mockNftNoMetadata` / `mockNftPartialMetadata` or something

kathaypacific (approved):

src/nfts/types.ts
```diff
@@ -0,0 +1,26 @@
...
+  contractAddress: string
+  tokenUri?: string | null
+  ownerAddress?: string | null
+  metadata: {
+    name: string
+    description: string
+    image: string
+    dna?: string | null
+    id?: number | null
+    date?: number | null
```
jeanregisser, line 22: What's this date?
Can we rename it to something more meaningful, for instance `createdAt` or `indexedAt`, etc?

MuckT: It's an optional field on the Nft metadata I've seen populated as in the examples below. I'd be hesitant to rename it `createdAt` or `indexedAt` as it doesn't exist on all Nfts. We could add a property indexedAt form the indexer returned a level up in the Nft object.
```
{
    ...
    "metadata": {
        "name": "Celostrial #6294",
        "description": "The worlds first nfETs",
        "image": "ipfs://Qme97ifAexrMDfjE3DZRMasWhe7D276uYsCGonyprSA2MJ/6294.png",
        "dna": "77c251ffa01132e1cf747afd6f7da93b3a37a6ae",
        "id": 6294,
        "date": 1644389403533,
        "attributes": [...]
    },
    "media": [...]
},
        ```

PR #3846, opened 2023-06-06
chore: add provider disclaimer to select payments screen

kathaypacific:
### Description

This PR adds a disclaimer on the payments screen which links to the fund account support article. I got a little carried away with some refactoring along the way, namely:
- use `Spacing` in styles instead of magic numbers
- reduce the amount of prop drilling (avoid passing a state setter to child components, as briefly mentioned discussed at brownbag)
- use `Trans` instead of concatenating strings (since concatenation may not work in the same order for all languages)

### Test plan

Updated unit tests. 

Updated copy below (the scenario is faked, the shorter copy will be displayed when users have all payment options available).

Simulator Screenshot - iPhone 14 Pro - 2023-06-06 at 12 34 26
Simulator Screenshot - iPhone 14 Pro - 2023-06-06 at 12 34 14

### Related issues

n/a

### Backwards compatibility

Y


silasbw (approved):
:artist: !

PR #3853, opened 2023-06-07
feat: add group size to big number format

satish-ravi:
### Description

We already specify a group separator from settings, but a group size is also needed to display the separator.

### Test plan

Manually, by having a balance > 1000 (or use a local currency so balance > 1000 in that currency), check if the group separator is displayed. Also by switching number format in iOS Settings (for some reason this isn't available in English, but available for Dutch / French, maybe due to iOS version)

Before (iOS): 

https://github.com/valora-inc/wallet/assets/5062591/6be86d8e-c397-4e2f-95ee-6bc17c04e0d1

After (iOS):

https://github.com/valora-inc/wallet/assets/5062591/73efe841-b439-41a1-a401-010f860b7587

Android:

| Before | After |
|--------|--------|
|  |  | 

### Related issues

- Fixes ACT-824

### Backwards compatibility

Yes


src/app/App.tsx
```diff
@@ -45,6 +45,7 @@ BigNumber.config({
   FORMAT: {
     decimalSeparator,
     groupSeparator: groupingSeparator,
+    groupSize: 3,
```
jeanregisser, line 48: I wonder if there's a weird locale that doesn't use 3.

jeanregisser (commented):
Great 👍 
Could we add a unit test for this?

satish-ravi:
@jeanregisser I am not sure if its worth adding a unit test, since it's just set on a global config. Any test on the TokenBalance component will just use the default config (since it doesn't import App.tsx).

src/app/App.tsx
satish-ravi: I guess there are locales that have other group sizes, e.g., for India, it's 3 for 1000s, then 2 for everything else. Although this would be a big refactor to support this, as react-native-localize doesn't support this.

jeanregisser:
I'm suggesting we cover this change with some kind of test, to avoid regressions in the future.
As we can see, the person who initially added this global change didn't cover it with a test which led to the `groupSeparator` thing to be actually useless.

Maybe we can add a test in App.test.tsx which just formats a BigNumber? Maybe not the best but better than nothing, no?

satish-ravi:
@jeanregisser added a test

jeanregisser (approved):
💯 🚀

PR #3854, opened 2023-06-07
fix(swaps): Fixing swap flow to support tokens with non default decimals.

dievazqu:
### Description

Some tokens such as WBTC, plastik, USDC, and many others don't use the Ethereum decimals (18).
In the first iteration, we assumed all the tokens had the default decimals.

This PR fixes that assumption, so now we can support any token. To maintain backward compatibility, we have created a new flag `isSwappableWithAnyDecimals` that enables the new feature.

### Test plan

In a iOS emulator + unit tests

### Related issues

- Fixes RET-712

### Backwards compatibility

Yes


MuckT (approved):
Tested locally on iOS and Android swapping CELO to plastiks and it's working great!

PR #3859, opened 2023-06-08
build(deps): disable hermes on ios

MuckT:
### Description

Disables Hermes that was enabled for iOS in #3796 due to issues with Detox.

### Test plan

- [ ] Test upgrading from Valora 1.59 if not included in a patch for release.

### Related issues

- Fixes RET-762

### Backwards compatibility

Yes


cajubelt (approved):

jeanregisser (approved):
Thanks for determining the source of the flakyness.
Have you seen folks discuss this on Detox's discord / GH?

I know @kathaypacific wanted to avoid this solution, but it's probably best in the short term.
I hope we can reenable it soon.

PR #3861, opened 2023-06-09
fix(qrcode): segmented control freeze on android

MuckT:
### Description

Sets the base value to `0.001` instead of `0`.

### Test plan

- [x] Tested locally on Android debug and release builds.

### Related issues

- Fixes ACT-821

### Backwards compatibility

Yes

### DIFF
--- a/src/components/SegmentedControl.tsx
+++ b/src/components/SegmentedControl.tsx
@@ -16,7 +16,8 @@ interface Props {
 }
 
 export default function SegmentedControl({ position, values, selectedIndex = 0, onChange }: Props) {
-  const [segmentWidth, setSegmentWidth] = React.useState(0)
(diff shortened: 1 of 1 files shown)


MuckT:
Closing as unfortunately this still has the same fail in CI on Android 😢 
```
FAIL e2e/src/QRScanner.spec.js (196.034 s)
  Given QR Scanner
    When opening QR scanner
      ✓ Then should display QR code (10050 ms)
      ✕ Then should be able to toggle camera (30216 ms)
      ✓ Then should be able to toggle to QR code (7785 ms)
      ✓ Then should be able to close QR code scanner (3253 ms)
    When 'scanning' QR
      ✕ Then should be able to handle Celo pay QR (41667 ms)
      ✕ Then should handle address only QR (29131 ms)
      ✎ todo Then should be able to wc QR
```
<p align="left">

</p>

PR #3874, opened 2023-06-13
fix(types): weaken requirements of blockchain-api responses as much as possible

cajubelt:
### Description

previously some types were defined with required fields that were optional in blockchain-api, which sets us up for unnecessary type errors at runtime. this PR weakens the reqs on blockchain-api as much as possible without actually changing application logic or breaking the typescript compilation. 

To fully resolve the out of sync types, we'll also need https://github.com/valora-inc/blockchain-api/pull/267 , which marks some fields as required in blockchain-api that are relied upon in the wallet and assumed to exist in the response from blockchain-api. 

### Test plan

compiled with typescript

### Related issues

na

### Backwards compatibility

yes. the changes to the wallet are just helping us avoid mistakenly assuming that optional fields are present on API responses. the changes to blockchain API are making fields that are relied on by the wallet required.


src/transactions/types.ts
```diff
@@ -131,8 +131,8 @@ export interface NftTransfer {
   transactionHash: string
   timestamp: number
   block: string
-  fees: Fee[]
-  nfts: Nft[]
+  fees?: Fee[]
+  nfts?: Nft[]
```
jophish, line 135: While we're at it, it might be good to add an optional `metadata?` field to NFT transfers to keep it in-sync with blockchain-api. Not needed though.

jophish (approved):

PR #3882, opened 2023-06-15
feat: update BottomSheet to be scroll enabled depending on content size

kathaypacific:
### Description

As the description

### Test plan

There should be no visual change for the existing components

https://github.com/valora-inc/wallet/assets/20150449/800fb29d-5779-425d-96e1-d699dd270162

### Related issues

- Fixes RET-698

### Backwards compatibility

Y

### DIFF
--- a/__mocks__/@gorhom/bottom-sheet.ts
+++ b/__mocks__/@gorhom/bottom-sheet.ts
@@ -1,0 +1,7 @@
+import { ScrollView } from 'react-native'
+
+module.exports = {
+  __esModule: true,
+  ...(jest.requireActual('@gorhom/bottom-sheet') as any),
+  BottomSheetScrollView: ScrollView,
+}
--- a/src/components/BottomSheet.tsx
+++ b/src/components/BottomSheet.tsx
@@ -14,7 +14,7 @@ interface Props {
   forwardedRef: React.RefObject<GorhomBottomSheet>
   title: string
   description?: string | null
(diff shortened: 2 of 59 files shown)


__mocks__/@gorhom/bottom-sheet.ts
```diff
@@ -0,0 +1,7 @@
+import { ScrollView } from 'react-native'
+
+module.exports = {
+  __esModule: true,
+  ...(jest.requireActual('@gorhom/bottom-sheet') as any),
+  BottomSheetScrollView: ScrollView,
```
kathaypacific, line 6: the BottomSheetScrollView has an internal utility called `getRefNativeTag` that "grab the native tag from the native host component. It only works when the ref is pointing to a native Host component" which is a real pain for unit tests. mocking it to be a RN scrollview seems harmless enough

dievazqu (approved):

src/components/BottomSheet.tsx
```diff
@@ -14,16 +14,13 @@ interface Props {
   forwardedRef: React.RefObject<GorhomBottomSheet>
   title: string
   description?: string | null
-  children?: React.ReactNode
+  children?: React.ReactNode | React.ReactNode[]
```
jeanregisser, line 17: nit: why is this needed?

jeanregisser (approved):
LGTM 👍

PR #3884, opened 2023-06-15
chore: tweak dapp favoriting callbacks

kathaypacific:
### Description

We need to differentiate between favourite from the dapps screen vs from the popular dapps sheet from an analytics perspective. I had to redo some plumbing because we had assumed that we can only favourite from one place.

This PR:
- update the dapp card to do the favourite action only, and accept callbacks (for analytics and other after effects) for favourite and unfavourite
- update the 4 variants of dapp screens and their favorites sections for this change 😵‍💫
- add favourite / unfavourite to the popular dapps bottom sheet 

### Test plan

Added unit tests.

https://github.com/valora-inc/wallet/assets/20150449/2b040d06-a55c-4886-a2da-73d5cf38d225

### Related issues

- Fixes RET-698

### Backwards compatibility

Y


locales/base/translation.json
```diff
@@ -1323,7 +1323,8 @@
     },
     "favoritedDappToast": {
       "message": "Added as a favorite",
-      "labelCTA": "View Favorites"
+      "labelCTA": "View Favorites",
+      "messageWithDappName": "{{dappName}} added as a favorite"
```
jeanregisser, line 1327: Do we really want a custom wording for this?

src/dappsExplorer/DappRankings.tsx
```diff
@@ -1,13 +1,14 @@
 import React, { RefObject, useEffect } from 'react'
 import { useTranslation } from 'react-i18next'
 import { StyleSheet, Text, View } from 'react-native'
+import Toast from 'react-native-simple-toast'
```
jeanregisser, line 4: Shouldn't we use our custom toast instead?

src/dappsExplorer/DappCard.tsx
```diff
@@ -29,31 +28,18 @@ interface Props extends DappCardContentProps {
...
-          dappId: dapp.id,
-          dappName: dapp.name,
-        }
-
     if (isFavorited) {
-      ValoraAnalytics.track(DappExplorerEvents.dapp_unfavorite, eventProperties)
       dispatch(unfavoriteDapp({ dappId: dapp.id }))
+      onUnfavoriteDapp?.(dapp)
     } else {
-      ValoraAnalytics.track(DappExplorerEvents.dapp_favorite, eventProperties)
```
jeanregisser, line 56: I'm wondering if we could pass the `section` (or other needed analytics props) so we can keep this in a single place?

src/dappsExplorer/DappRankings.tsx
kathaypacific: i checked with Reed about this shortcut and he was okay with it. the custom toast we have isn't styled in the way the designs were, and expects an action which this design doesn't have

locales/base/translation.json
kathaypacific: 😅 i could take it out but wasn't sure if there'd be languages where it'd make more sense for the dappName to be in the middle of the sentence or something

src/dappsExplorer/DappCard.tsx
kathaypacific: yeah this is a much better idea

jeanregisser (approved):

PR #3889, opened 2023-06-17
feat(nfts): add nft icon in transaction feed

MuckT:
### Description

Adds the first NFT as an icon in the transaction feed item for a NFT transfer.

#### Loaded / Image Load Error
<p align="left">

</p>

### Test plan

- [x] Tested locally on iOS
- [x] Tested in low network conditions

### Related issues

- Fixes RET-595

### Backwards compatibility

Yes


src/transactions/feed/NftFeedItem.tsx
```diff
@@ -12,12 +16,45 @@ import { Nft } from 'src/nfts/types'
...
+
+function NftIcon({ nft }: { nft: Nft }) {
+  const [loading, setLoading] = useState(true)
+  const [imageLoadingError, setImageLoadingError] = useState(false)
+  return imageLoadingError ? (
+    <View style={[styles.circleIcon, styles.errorCircleIcon]}>
+      <ImageErrorIcon size={30} testID="NftFeedItem/NftErrorIcon" />
+    </View>
+  ) : (
+    <FastImage
```
kathaypacific, line 44: i think there might be some way we can abstract this fastImage component, we have it in quite a few places now and they all do the same thing with slightly different styles. we can address this in a separate PR, but i think it'd be a nice improvement to handle the loading / error logic in a single place

kathaypacific (approved):
approving but we should add analytics before the release cut, can do in a separate PR if that's the plan

src/transactions/feed/NftFeedItem.tsx
```diff
@@ -39,15 +90,22 @@ function NftFeedItem({ transaction }: Props) {
...
         <View style={styles.descriptionContainer}>
           <Text style={styles.title}>
             {transaction.type === TokenTransactionTypeV2.NftReceived
-              ? t('receivedNft')
+              ? nfts.length > 1
+                ? t('receivedNfts')
+                : t('receivedNft')
+              : nfts.length > 1
+              ? t('sentNfts')
               : t('sentNft')}
```
jeanregisser, line 109: i18next supports [plurals](https://www.i18next.com/translation-function/plurals), we should use it instead.

jeanregisser (approved):
Awesome 🚀

src/transactions/feed/NftFeedItem.tsx
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/3889/commits/ff6263382977d369abebfeaaac86f1fd721f747a!

PR #3896, opened 2023-06-20
chore(nfts): analytics

MuckT:
### Description

Adds Analytics for the following:
- `nft_error_screen_open`
- `nft_image_load`
  - `tokenId` - string
  - `contractAddress` - string
  - `url?` - string
  - `origin` - `nftsInfoCarouselMain` | `nftsInfoCarouselThumbnail` | `transactionFeed`
  - `error` - boolean
  
Also adds a helper function `getGatewayUrl` used by `onImageLoad` function.

### Test plan

- [x] Tested locally with logs.
- [x] Unit tests updated - had some difficulty mocking `FastImage` so I tested the functions directly for the time being.

### Related issues

- Fixes RET-595

### Backwards compatibility

Yes


kathaypacific (commented):
Looks good but i think we should not merge a new utils file since we're trying to get away from them?

src/nfts/utils.ts
```diff
@@ -0,0 +1,34 @@
+import { NftEvents } from 'src/analytics/Events'
```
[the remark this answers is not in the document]
MuckT: I don't think placing shared functions in a different file really solves the issue of poorly thought out abstraction, but I'll make the change.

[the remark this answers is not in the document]
MuckT: Thought about this a bit and I don't think it makes sense to place in the `NftsInfoCarousel.tsx`. It'd probably be best to revisit this in a follow up when the shared image component is created.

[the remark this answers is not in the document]
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/3896/commits/ed2a4b7ec757306f376127547ca38816830a3e48!

kathaypacific (approved):

src/nfts/NftsLoadError.test.tsx
```diff
@@ -1,10 +1,18 @@
 import { fireEvent, render } from '@testing-library/react-native'
 import React from 'react'
+import { NftEvents } from 'src/analytics/Events'
+import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
 import { navigate } from 'src/navigator/NavigationService'
 import { Screens } from 'src/navigator/Screens'
 import NftLoadError from 'src/nfts/NftsLoadError'
 
+jest.mock('src/analytics/ValoraAnalytics')
```
jeanregisser, line 9: nit: this can be removed ValoraAnalytics is already [mocked by default](https://github.com/valora-inc/wallet/blob/5a982019f89ef552c8fc9e5f1b0fee7bdedd7c19/__mocks__/src/analytics/ValoraAnalytics.ts).

src/nfts/NftsInfoCarousel.test.tsx
```diff
@@ -124,4 +133,40 @@ describe('NftsInfoCarousel', () => {
...
+      error: false,
+    })
+    expect(ValoraAnalytics.track).toBeCalledTimes(1)
+    expect(Logger.info).toBeCalledWith(
+      NftOrigin.TransactionFeed,
+      `ContractAddress=${mockNftAllFields.contractAddress}, TokenId: ${mockNftAllFields.tokenId}, Loaded image from ${mockNftAllFields.media[0].gateway}`
+    )
+  })
+
+  it('onImageLoad error', () => {
```
jeanregisser, line 157: nit: can we give these dedicated `describe` blocks and have a more meaningful `it` strings (verb + description)?

src/analytics/Events.tsx
```diff
@@ -601,6 +601,11 @@ export enum AssetsEvents {
   tap_asset = 'tap_asset', // When a user taps on an asset
 }
 
+export enum NftEvents {
+  nft_error_screen_open = 'nft_error_screen_open', // When the high level error screen is mounted
```
jeanregisser, line 605: All screen mounts should already be captured. https://github.com/valora-inc/wallet/blob/69729e8213368db34bde7285eff523be2073d14c/src/navigator/NavigatorWrapper.tsx#L158

But apparently something broken https://linear.app/valora/issue/RET-423/[app-analytics]-global-screen-navigation-events
Would love for us to understand this, instead of manually doing this.

Nothing to change here though.

src/nfts/NftsInfoCarousel.test.tsx
MuckT: This is just a temporary place for these helper functions. They will be moved in a follow up PR.

PR #3902, opened 2023-06-22
feat(nfts): add nfts owned to redux

MuckT:
### Description

Adds users owned NFTs to redux to allow for image cache hits: enabled via feature flag.

### Test plan

- [x] Add Saga Tests
  - Succesful load
  - Error load
  - Feature flag enabled / disabled
- [x] Add redux migration test
- [x] Existing tests updated

### Related issues

- Fixes RET-764

### Backwards compatibility

Yes


src/nfts/saga.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  try {
+    const response = yield call(fetch, url, {
+      method: 'GET',
+      headers: {
+        'Content-Type': 'application/json',
+        Accept: 'application/json',
+      },
+    })
+    if (!response.ok) yield put(fetchNftsFailed({ error: 'Could not fetch NFTs' }))
```
dievazqu, line 36: Shouldn't we throw here? And the `catch` will handle it.
Maybe we can include some data from the failed response in the message. I think `response.status` and `response.statusText` are good to have.

Also, it's a good practice to always include the `{ }`.

MuckT: The structure was recommended in https://github.com/valora-inc/wallet/pull/3908#discussion_r1239871536.

kathaypacific: i think you should throw here, it will land in the catch block so you don't need to duplicate the error handling (calling `fetchNftsFailed` twice)

MuckT: Taking a look at this, wouldn't this make the error more generic? We wouldn't have a separate error for non ok response and a parse error. I am fine with this as long as we're ok losing this error granularity. @dievazqu

src/nfts/saga.ts
```diff
@@ -0,0 +1,54 @@
...
+  try {
+    const response = yield call(fetch, url, {
+      method: 'GET',
+      headers: {
+        'Content-Type': 'application/json',
+        Accept: 'application/json',
+      },
+    })
+    if (!response.ok) {
+      throw new Error('Could not fetch NFTs')
```
dievazqu, line 37: nit: This is fine, but maybe we can include some data of why it failed? At least the status code of the response?
I think it will make it easier for us if we have to debug.

dievazqu (approved):
LGTM! 👏

src/nfts/saga.ts
MuckT: fixed in https://github.com/valora-inc/wallet/pull/3902/commits/4573c5e25a234bce0dfe27bf72e1ff1e9cc12e0d!

PR #3903, opened 2023-06-22
refactor: abstract nft image and loading to one component

kathaypacific:
### Description

This PR creates one component that handles loading / success / error states of nft images.

### Test plan

Loading state (note that this is faked with a 2 second loading time, and the image appears under the loader but this wouldn't usually happen since the loader would disappear after the image loads)

https://github.com/valora-inc/wallet/assets/20150449/c12b79c6-8c18-4cdb-835b-3d4ac29a1735

Error state
Simulator Screenshot - iPhone 14 Pro - 2023-06-22 at 17 49 31

https://github.com/valora-inc/wallet/assets/20150449/5ec4e8ff-7ae7-498a-b1b0-f1e64743ee71

### Related issues

- Fixes RET-595

### Backwards compatibility

Y


src/nfts/NftsInfoCarousel.test.tsx
```diff
@@ -133,40 +131,4 @@ describe('NftsInfoCarousel', () => {
...
-      url: mockNftAllFields.media[0].gateway,
-      origin: NftOrigin.NftsInfoCarouselMain,
-      error: true,
-    })
-    expect(ValoraAnalytics.track).toBeCalledTimes(1)
-    expect(Logger.error).toBeCalledWith(
-      NftOrigin.NftsInfoCarouselMain,
-      `ContractAddress=${mockNftAllFields.contractAddress}, TokenId: ${mockNftAllFields.tokenId}, Failed to load image from ${mockNftAllFields.media[0].gateway}`
-    )
-  })
```
kathaypacific, line 171: i removed these tests because i removed these abstractions (we were unhappy with them anyway)

i decided not to add tests for these. what we actually want to test, is when FastImage loads or fails to load the image, that the correct analytics is dispatched. i'd argue that that is not what these tests were doing anyway, they are testing a function but not testing when the function is triggered. i considered mocking out FastImage to get it to throw errors etc but it seems too much like testing the internals of FastImage

src/nfts/NftImage.tsx
```diff
@@ -0,0 +1,135 @@
...
+      tokenId,
+      contractAddress,
+      url: imageUrl,
+      origin,
+      error,
+    })
+
+    if (error) {
+      Logger.error(
+        origin,
```
jeanregisser, line 82: Don't we want the error?
```suggestion
        error,
```

src/nfts/NftImage.tsx
jeanregisser, line 1: Should we add some simple tests?

src/nfts/NftImage.tsx
jeanregisser: ah now I see the error is not an error but a boolean 🤔

jeanregisser (commented):
Thanks!

src/nfts/NftImage.tsx
```diff
@@ -0,0 +1,135 @@
...
+  height,
+  width,
+  shouldAutoScaleHeight,
+  borderRadius,
+  origin,
+  ErrorComponent,
+  testID,
+}: Props) {
+  const [isLoading, setIsLoading] = useState(true)
+  const [error, setError] = useState(!nft.metadata)
```
jeanregisser, line 58: Looks like we should combine these 2 in a single `status` state to make it impossible to have it both in loading and error state.

kathaypacific: makes sense, i updated

src/nfts/NftImage.tsx
kathaypacific: i would only add tests for the loading and no metadata error state. to test the loading / success / error states of the actual fast image lib requires us to mock the lib (what comes out of the box is just `<FastImageView      testID="NftImage" />` on render in the testing env) which seems like not work the effort

src/nfts/NftImage.tsx
kathaypacific: i made this a string. i think it's helpful to differentiate between when the load fails because of lack of metadata vs bad image url. and would be good to measure the ratio of successful vs failed image loads so i kept the event unchanged (was tempted to only capture the failures for a while)

jeanregisser (approved):
🚀

PR #3904, opened 2023-06-22
chore: remove dapp filter and search experiment

kathaypacific:
### Related issues

- Fixes #[issue number here]


src/redux/migrations.ts
```diff
@@ -1135,4 +1135,22 @@ export const migrations = {
...
+        (dapp: Dapp | (Omit<Dapp, 'categories'> & { categoryId: string })) => ({
+          ...dapp,
+          categories: 'categories' in dapp ? dapp.categories : [dapp.categoryId],
+        })
+      ),
+      activeDapp: state.dapps.activeDapp
+        ? {
+            ...state.dapps.activeDapp,
+            categories: state.dapps.activeDapp.categories ?? [state.dapps.activeDapp.categoryId],
+          }
```
jeanregisser, line 1152: V1 categories can't all be directly mapped to V2 categories. Example with Moola:
https://github.com/valora-inc/dapp-list/blob/0645e9bd4ba7208df60573afffe1da8a675d3906/src/valora-dapp-list.json#L77-L78
Not entirely sure of the impact of this though.

I wonder if we should just clear the whole list and redownload a fresh list?

jeanregisser (commented):
Awesome!
Great to see all this added complexity gone 💯 
I'll let @MuckT confirm the changes since he was closer to them.

src/redux/migrations.ts
kathaypacific: the dapp list is fetched and refreshed on app start, as long as we assume that this service isn't going to fail / the user has internet connection then what we do here shouldn't matter too much. there are pros and cons like we discussed - 
1. if we clear the dapps: the recently used / favourited dapps on the homescreen will not appear since we have no dapps to show the information needed. but the migration code is simplified :D
2. if we don't clear dapps and apply this transform: we can potentially get this big "Lend, Borrow & Earn" filter if the dapp list isn't refreshed in time, but if the service fails for whatever reason then the user can continue to use the dapps as normal.

(there is some logic in the dapp list api that returns only the localised categories that are applicable to the response, but users who have the v1 response stored should also have the translation for the lend-borrow-earn category)

i lean towards keeping it like this since it's the option that lets users continue the journey if the dapplist service is down. but at the same time the service is pretty light and reliable so i'm also open to clearing the list 😅

jeanregisser (approved):

PR #3908, opened 2023-06-23
feat(nfts): gallery view

MuckT:
### Description

Adds the Gallery View for NFTs.

### Video iOS

https://github.com/valora-inc/wallet/assets/26950305/4b9dd0ee-6e6c-4878-8deb-8561cc86687a

### Test plan

- [x] Tested locally on iOS
- [ ] Tested locally on Android
- [ ] Unit tests added
  - Render
  - Error state
  - Pull to refresh

### Related issues

- Fixes RET-764

### Backwards compatibility

Yes


MuckT:
@kathaypacific ran into an issue using a custom loader with the pull to refresh and also opted to use the default similar to the transaction feed. Also I avoided duplicating the custom header; hopefully we can put this into `headers.tsx` and reuse it instead of reimplementing for each screen.

src/nfts/NftGallery.tsx
```diff
@@ -0,0 +1,129 @@
...
+  const error = useSelector(myNftsErrorSelector)
+  const nfts = useSelector(myNftsSelector)
+
+  function onRefresh() {
+    dispatch(fetchMyNfts())
+  }
+
+  // TODO: add analytics event for page load
+
+  if (error) {
```
kathaypacific, line 60: nit: if you move the error handling inside the main return block, you could save some code (don't have to declare the safe area / drawer top bar twice, you already check for error in the main block anyway)

src/nfts/NftGallery.tsx
```diff
@@ -0,0 +1,129 @@
...
+import { fetchMyNfts } from 'src/nfts/slice'
+import { Nft, NftOrigin } from 'src/nfts/types'
+import colors from 'src/styles/colors'
+import { Spacing } from 'src/styles/styles'
+import variables from 'src/styles/variables'
+
+// This is used instead of a gap between the images because the gap is not supported in RN yet.
+const imageSize = Math.min(variables.width / 2.5, variables.height / 4)
+
+function NftTouchable({ nft }: { nft: Nft }) {
```
kathaypacific, line 23: nit: this is my personal opinion but i don't really like having multiple components declared in the same file. it makes the file cluttered and difficult to read. so i think this abstraction is unneeded, it seems like we could have this code inside the main gallery component without complicating that component too much and reduce the amount of abstraction

src/nfts/saga.ts
```diff
@@ -0,0 +1,56 @@
...
+import { StatsigFeatureGates } from 'src/statsig/types'
+import Logger from 'src/utils/Logger'
+import { safely } from 'src/utils/safely'
+import { Actions } from 'src/web3/actions'
+import networkConfig from 'src/web3/networkConfig'
+import { walletAddressSelector } from 'src/web3/selectors'
+
+const TAG = 'NftsSaga'
+
+export function* handleFetchMyNfts() {
```
kathaypacific, line 13:
```suggestion
export function* handleFetchNfts() {
```

src/nfts/saga.ts
```diff
@@ -0,0 +1,56 @@
...
+  const showMyNftsInApp = getFeatureGate(StatsigFeatureGates.SHOW_IN_APP_NFT_GALLERY)
```
kathaypacific, line 14:
```suggestion
  const showNftGallery = getFeatureGate(StatsigFeatureGates.SHOW_IN_APP_NFT_GALLERY)
```

src/nfts/saga.ts
```diff
@@ -0,0 +1,56 @@
...
+      Accept: 'application/json',
+    },
+  })
+
+  if (response.ok) {
+    try {
+      const { result } = yield call([response, 'json'])
+      yield put(fetchMyNftsCompleted(result))
+    } catch (error) {
+      Logger.error(TAG, 'Could not parse NFTs response', error)
```
kathaypacific, line 41: nit: i think we'll be alerted from the backend if it returns an error, so it doesn't need to go to sentry too?

src/nfts/saga.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  const address = yield select(walletAddressSelector)
+  if (!address) {
+    Logger.debug(TAG, 'Wallet address not found, skipping NFTs list fetch')
+    return
+  }
+
+  const url = `${networkConfig.getNftsByOwnerAddressUrl}?address=${address}`
+
+  const response = yield call(fetch, url, {
```
kathaypacific, line 28: code golf:
```
try {
  yield call(fetch, ...)
  if (!response.ok) {
    // throw some kind of error
  }
  // do successful stuff
} catch (error) {
  // do error handling
}

```
benefit: only one catch block

src/nfts/slice.ts
```diff
@@ -0,0 +1,41 @@
+import { createSlice } from '@reduxjs/toolkit'
+import { Nft } from 'src/nfts/types'
+
+export interface FetchMyNftsCompleted {
+  nfts: Nft[]
+}
+
+export interface State {
+  myNftsLoading: boolean
```
kathaypacific, line 9: i think we should remove "my" from all the variable names? we don't have "fetchMyTokens" etc.

src/nfts/NftGallery.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+export default function NftGallery() {
+  const { t } = useTranslation()
+  const dispatch = useDispatch()
+  const loading = useSelector(myNftsLoadingSelector)
+  const error = useSelector(myNftsErrorSelector)
+  const nfts = useSelector(myNftsSelector)
+
+  function onRefresh() {
+    dispatch(fetchMyNfts())
```
kathaypacific, line 55: if we're thinking of NFTs as assets just like tokens and positions, i'd expect us to fetch nft's in the same place as we fetch tokens and positions (i.e. before users land on this screen - we can still do a fetch here to refresh the data, but if we also fetch on app start then we reduce the chance of users ending up on a blank loading state here)

src/nfts/slice.ts
MuckT: I wasn't sure if we'd eventually store other Nfts in redux, but we can migrate the redux state if needed. I'll remove all these _My_ prefixes.

src/nfts/NftGallery.tsx
MuckT: The fetch action is dispatched on spawn on the saga spawn similar to the `dappsSaga.ts`.

MuckT:
@kathaypacific something went sideway in this git history. I'll pull out the logic into a fresh PR. The saga / redux changes have been made in https://github.com/valora-inc/wallet/pull/3902.

PR #3911, opened 2023-06-23
refactor: abstract scroll aware header so that it can be reused

kathaypacific:
### Description

As the title, since it's coming up more in designs

### Test plan

Manually tested the token balances screen on iOS and still looks fine

### Related issues

n/a

### Backwards compatibility

Y


MuckT (approved):
Tested locally on iOS and Android and it appears the same as before.

PR #3912, opened 2023-06-23
feat(nfts): gallery view

MuckT:
### Description

Reworking #3908 of Adds the Gallery View for NFTs.

### iOS Screenshots

| One NFT   |  Three NFTs  | Error State | No NFTs |
|:-:|:-:|:-:|:-:|
|  |  |  | 
### Test plan

- [x] Tested locally on iOS
- [x] Unit tests added
  - Render
  - Error state
  - No NFTs

### Related issues

- Fixes RET-764

### Backwards compatibility

Yes


src/analytics/Properties.tsx
```diff
@@ -1296,6 +1296,7 @@ interface NftsEventsProperties {
     origin: NftOrigin
     error?: string
   }
+  [NftEvents.nft_gallery_screen_open]: undefined
```
kathaypacific, line 1299: what do you think about adding the number of nft's a user has to this event property? it might be interesting to understand if our users are holding many or few nft's?

src/nfts/NftGallery.tsx
```diff
@@ -0,0 +1,139 @@
...
+  function onRefresh() {
+    dispatch(fetchNfts())
+  }
+
+  useEffect(() => ValoraAnalytics.track(NftEvents.nft_gallery_screen_open), [])
+
+  return (
+    <SafeAreaView testID="NftGallery" style={styles.safeAreaContainer} edges={['top']}>
+      <DrawerTopBar middleElement={<Text>{t('nftGallery.title')}</Text>} />
+      <ScrollView
```
kathaypacific, line 42: could we use Flatlist instead for better performance, since the list of nft's could be quite long? there is a `numColumns` prop that you could use i think?

src/nfts/NftGallery.tsx
```diff
@@ -0,0 +1,139 @@
...
+    flexWrap: 'wrap',
+  },
+  lastImage: {
+    marginRight: variables.width / 2 - Spacing.Smallest8,
+  },
+  noNfts: {
+    ...fontStyles.regular,
+    color: colors.gray3,
+    textAlign: 'center',
+    marginTop: '32%',
```
kathaypacific, line 127: nit: why not use a `Spacing` constant instead?

src/nfts/NftsLoadError.tsx
```diff
@@ -27,7 +31,7 @@ export default function NftsLoadError() {
   }, [])
 
   return (
-    <SafeAreaView style={styles.safeArea} testID="NftsInfoCarousel/NftsLoadErrorScreen">
+    <SafeAreaView style={styles.safeArea} testID={testID ?? 'NftsInfoCarousel/NftsLoadErrorScreen'}>
```
kathaypacific, line 34: nit: if this component is reusable, i don't think it makes sense to have a default testID? i guess it's not very consequential here but it does feel like a trap for forgetting to pass this prop in and having the wrong testID assigned

src/nfts/selectors.ts
```diff
@@ -0,0 +1,7 @@
+import { RootState } from 'src/redux/reducers'
+
+export const nftsLoadingSelector = (state: RootState) => state.nfts.nftsLoading
+
+export const nftsErrorSelector = (state: RootState) => state.nfts.nftsError
```
kathaypacific, line 5: nit: we started doing this in the positions stuff, but it seems like it'd be nice to have a single state to represent status (e.g. status = 'loading' | 'error' | 'success' etc) to prevent the loading and error states from being simultaneously true. it also makes rendering a little easier. idk if you want to do that here, but wanted to call it out. changing the state would require a redux migration but you could also just derive this on the selector level

kathaypacific (commented):
looks great!!!! excited about this one!!

src/nfts/selectors.ts
MuckT: A follow up PR with a redux migration and adjusting this component seems like a good place to make this change. I agree having a single state to represent the status of the component makes a lot more sense to me.

src/nfts/NftGallery.tsx
MuckT: I don't think we have a spacing constant that is big enough. We could add one for '32%' as there are several places it's added as a margin / padding.

src/nfts/NftGallery.tsx
MuckT: Done in 1bbfdf6a67484f3d1145ade2c32cc2d20538c517!

src/analytics/Properties.tsx
kathaypacific: bump on this? :)

src/nfts/NftGallery.tsx
```diff
@@ -0,0 +1,132 @@
...
+    <SafeAreaView testID="NftGallery" style={styles.safeAreaContainer} edges={['top']}>
+      <DrawerTopBar middleElement={<Text>{t('nftGallery.title')}</Text>} />
+      {error ? (
+        <NftsLoadError testID="NftGallery/NftsLoadErrorScreen" />
+      ) : (
+        <FlatList
+          numColumns={2}
+          data={nfts}
+          refreshing={loading}
+          onRefresh={() => dispatch(fetchNfts())}
```
kathaypacific, line 48: can we make this a function if it will be reused in the RefreshControl too? (do we need both?)

src/nfts/NftGallery.tsx
```diff
@@ -0,0 +1,132 @@
...
+            <RefreshControl
+              refreshing={loading}
+              colors={[colors.greenUI]}
+              onRefresh={() => dispatch(fetchNfts())}
+            />
+          }
+          renderItem={({ item, index }) => (
+            <Touchable
+              borderless={false}
+              onPress={() => navigate(Screens.NftsInfoCarousel, { nfts: [item] })}
```
kathaypacific, line 60: do we need tracking on this press?

src/nfts/NftGallery.tsx
```diff
@@ -0,0 +1,132 @@
...
+              style={[
+                styles.touchableIcon,
+                index % 2 === 0
+                  ? { marginRight: Spacing.Regular16 } &&
+                    index === nfts.length - 1 &&
+                    styles.lastImage
+                  : { marginLeft: Spacing.Regular16 },
```
kathaypacific, line 67: i found this a `{ marginRight: Spacing.Regular16 } && index === nfts.length - 1 && styles.lastImage`  a little unexpected for a style. can we simplify? don't we only need to remove the right margin when index % 2 === 0?

src/nfts/NftGallery.tsx
```diff
@@ -0,0 +1,132 @@
...
+              ]}
+            >
+              <NftImage
+                nft={item}
+                testID="NftGallery/NftImage"
+                width={imageSize}
+                height={imageSize}
+                ErrorComponent={
+                  <View style={styles.errorView}>
+                    <ImageErrorIcon color="#C93717" />
```
kathaypacific, line 77: can we use the red from the colors constant?

src/nfts/NftGallery.tsx
```diff
@@ -0,0 +1,132 @@
...
+        />
+      )}
+    </SafeAreaView>
+  )
+}
+
+const styles = StyleSheet.create({
+  contentContainer: {
+    alignItems: 'center',
+    paddingBottom: '12%', // TODO: find a better way to do this
```
kathaypacific, line 102: why does this need to be a percentage? can we just use a Spacing constant?

src/nfts/NftGallery.tsx
MuckT: We could add it, but currently there are no requirements around it. We will have also abundance of information from the <NftImage /> component as it sends analytics for each image loaded.

src/nfts/NftGallery.tsx
MuckT: We use `#C93717` in `NftsInfoCarousel.tsx` also. I think it would be a bit strange if we displayed different colors of red. We could swap both, add this new color as a constant or keep it the way it is.

src/nfts/NftGallery.tsx
MuckT: I tried the largest spacing constant and the last image was still covered by the bottom bar.

<p align="center">

</p>

src/nfts/NftGallery.tsx
MuckT: Added a comment in 5a51564adb3e69c804800a184a43b2592b4d21ab to help explain why this was done.

MuckT:
@kathaypacific regarding '32%' there's some similar logic in `SwapExecuteScreen.tsx` and `NftsLoadError.tsx`. As I understand it using a precent is preferable to `Spacing.Large32` as it adjusts better for various screen sizes.

src/nfts/NftsLoadError.tsx
MuckT: Fixed in c85222bf3c67d3e2afbb5d5df8f3cabeead68fe0!

src/nfts/NftGallery.tsx
MuckT: There's some similar logic in `SwapExecuteScreen.tsx` and `NftsLoadError.tsx`. Using a precent was done as opposed to something like `Spacing.Large32` as it adjusts better for various screen sizes.

src/nfts/NftGallery.tsx
kathaypacific: maybe we could apply a bottom padding on the images themselves? i'm not sure, but it seems like something we can tidy up later

kathaypacific (approved):
i have some small comments remaining but don't think they're important enough to hold this up for the release. i'll merge but would be nice to loop back at some point on those :)

PR #3917, opened 2023-06-26
chore: add plumbing for dapp claimable rewards data

kathaypacific:
### Description

This PR:
- add `availableShortcutIds` to the expect positions type
- add `shortcuts` to the positions redux store
- add redux migration for both
- add saga to fetch and store shortcuts on app start
- add selector to get claimable positions with claimable tokens
- use this selector to decide whether to show claim action dapp card

### Test plan

n/a

### Related issues

- Fixes RET-733

### Backwards compatibility

Y


src/positions/types.ts
```diff
@@ -46,3 +53,17 @@ export interface ContractPosition extends AbstractPosition {
...
+  appId: string
+  name: string
+  description: string
+  networks: string[]
+  category?: 'claim'
+}
+
+export interface ClaimablePosition extends Omit<Position, 'availableShortcutIds' | 'tokens'> {
+  shortcutId: string
+  claimableTokens: TokenWithCategory[]
```
kathaypacific, line 68: i think this type makes it easier to display the claimable positions, but maybe there's a more elegant way to do this. since a position can have multiple "claim" shortcuts, and we need to display each "claim" shortcut as an item on the UI, my idea is to have a list of positions where each position has a "claim" shortcut and a list of claimable tokens. 

note that the claimable tokens can have nested tokens but the UI should only need to look on the top level for which tokens to display, so the selector moves all the nested claimable tokens to the top level

src/web3/networkConfig.ts
```diff
@@ -48,6 +48,7 @@ interface NetworkConfig {
   fetchAvailableSuperchargeRewardsV2: string
   resolveId: string
   getPositionsUrl: string
+  getPositionsShortcutsUrl: string
```
jeanregisser, line 51:
```suggestion
  getShortcutsUrl: string
```

src/positions/types.ts
jeanregisser: This looks good to me 👍 

I wonder if we want a direct item type which contains both the shortcuts info + claimableTokens, directly consumable by the list component that will display them?

src/positions/saga.ts
```diff
@@ -67,4 +88,5 @@ export function* watchFetchBalances() {
 
 export function* positionsSaga() {
   yield spawn(watchFetchBalances)
+  yield spawn(fetchShortcutsSaga)
```
jeanregisser, line 91: I wonder how often we should refetch these. Main problem is if the initial fetch fails.
Maybe for now we could do it everytime the balance needs to be fetched?

src/positions/selectors.ts
```diff
@@ -37,3 +42,43 @@ function sortByBalanceUsd(position1: Position, position2: Position) {
...
+  return [...claimableTokens, ...nestedTokens]
+}
+
+export const positionsWithClaimableRewardsSelector = createSelector(
+  [positionsSelector, claimableShortcutSelector],
+  (positions, shortcuts) => {
+    const claimablePositions: ClaimablePosition[] = []
+    positions.forEach((position) => {
+      const appShortcutIds = shortcuts
+        .filter((shortcut) => shortcut.appId === position.appId)
```
jeanregisser, line 67: For now we only want `claim` shortcuts.
In the future it's possible we'll have a position with a claimable token and a claim shortcut but also other shortcuts (deposit, withdraw, etc).


```suggestion
        .filter((shortcut) => shortcut.appId === position.appId && shortcut.category === 'claim')
```

jeanregisser (commented):
Awesome! 🚀

jeanregisser (approved):

src/web3/networkConfig.ts
kathaypacific: updated! `getShortcutsUrl` was also my first thought but i started to wonder if "shortcuts" is too vague 😅

src/positions/saga.ts
kathaypacific: ah yes i overlooked this, i only was thinking of not trying to fetch too many times. i added a new status to track the fetch in redux, so that we can retry every time the positions are fetched but skip if there was a previous successful request. seemed like not the highest priority work but also if i don't sneak it in now we might never find the motivation to update the logic....

src/positions/types.ts
kathaypacific: yeah that's a good idea. i updated, is this what you were thinking?

jeanregisser (approved):
Yihah! 🐑 it!

PR #3919, opened 2023-06-26
feat(SelectProvider): add info text on amount spent

satish-ravi:
### Description

Part 1 of the changes to select provider screen to show receive amount instead of fees. This includes integration with statsig feature gate and showing the "You'll pay/withdraw <amount> including fees" text. Showing the receive amount will be in a follow up

### Test plan

| Add | Withdraw |
|--------|--------|
|  |  | 

### Related issues

- Part of ACT-842

### Backwards compatibility

Yes


src/fiatExchanges/amount.tsx
```diff
@@ -0,0 +1,50 @@
+import BigNumber from 'bignumber.js'
```
satish-ravi, line 1: This is just moved from src/fiatconnect/ReviewScreen.tsx

jh2oman (approved):

PR #3924, opened 2023-06-29
fix(dapps): dismiss keyboard before external browser dapp open

MuckT:
### Description

Dismisses the keyboard on external browser dapp open.

### Test plan

Tested Locally on iOS

### Related issues

N/A

### Backwards compatibility

Yes


src/dapps/saga.ts
```diff
@@ -45,6 +46,7 @@ export function* handleOpenDapp(action: PayloadAction<DappSelectedAction>) {
       navigate(Screens.WebViewScreen, { uri: dappUrl })
     }
   } else {
+    Keyboard.dismiss() // Dismiss keyboard before opening dapp
```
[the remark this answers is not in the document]
MuckT: I moved it to the mount of `BottomSheetScrollView.tsx` as this covers cases where the user has triggered a dapp action from an external browser that was not navigated to in Valora.

src/components/BottomSheetScrollView.tsx
```diff
@@ -30,6 +30,11 @@ function BottomSheetScrollView({ handleContentLayout, containerStyle, testId, ch
     }
   }
 
+  // Dismiss keyboard on mount
+  useEffect(() => {
+    Keyboard.dismiss()
```
dievazqu, line 35: Just wondering, will this have conflicts if we want to implement some experience where the keyboard should automatically appear because there is some input field with auto focus in the bottom sheet?

Shouldn't we try to dismiss the keyboard in the specific action that triggers the dapp open flow?

dievazqu (approved):
I left a comment, but probably I'm thinking in a case that won't happen. So I'm fine merging this.

But It sounds a bit weird to dismiss the keyboard when the bottom sheet mounts, I feel they shouldn't be related.
Also, could some case happen where the bottom sheet is already mounted so the `useEffect` is not triggered (like if you open the bottom sheet twice)?

src/components/BottomSheetScrollView.tsx
MuckT: In general I think we should avoid any input field on the bottom sheet. This is similar to #3625 where the keyboard dismisses when the bottom sheet becomes visible. I haven't seen any instances where the `useEffect` is not triggered and the keyboard persists in testing.

PR #3925, opened 2023-06-29
fix(nft-thumbnail): add width to thumbnail

MuckT:
### Description

It looks like #3903 was merged without adding a width to the NFT thumbnail. Since we don't have the dimentions of the images prior to fetching this causes render issues on low latency networks.

#### iOS Before

https://github.com/valora-inc/wallet/assets/26950305/89dfc353-4b9e-4d66-8508-28439f84ba03

#### iOS After

https://github.com/valora-inc/wallet/assets/26950305/29e14c51-4ddf-450d-b67a-becdf00b1a05

### Test plan

Tested locally on iOS and with network link conditioner to simulate poor network conditions.

### Related issues

#3903

### Backwards compatibility

Yes


kathaypacific (approved):
thanks!!

PR #3923, opened 2023-06-28
test(e2e): Skip FC cash out tests for time being on android

jophish:
### Description

See [here](https://valora-app.slack.com/archives/C025V1D6F3J/p1687900991663059) for more context. In order to get builds to pass, temporarily disabling these tests while we fix the source of the issue.

### Test plan

N/A

### Related issues

N/A

### Backwards compatibility

Yes.

PR #3927, opened 2023-06-29
fix(cash-in-bottom-sheet): only show when home is in focus

MuckT:
### Description

Makes is so the cash in bottom sheet is only displayed when the `wallethome.tsx` is in focus and re-enables the fiat connect cash out tests: [Related Slack Thread](https://github.com/valora-inc/wallet/pull/3923#:~:text=See%20here%20for%20more%20context).

### Test plan

Tested locally on Android after the tests are enabled.

### Related issues

- #3923

### Backwards compatibility

Yes

### DIFF
--- a/src/home/WalletHome.tsx
+++ b/src/home/WalletHome.tsx
@@ -1,6 +1,6 @@
 import { useIsFocused } from '@react-navigation/native'
 import _ from 'lodash'
-import React, { useEffect, useRef } from 'react'
+import React, { useEffect, useMemo, useRef } from 'react'
 import { useTranslation } from 'react-i18next'
 import { RefreshControl, RefreshControlProps, SectionList, StyleSheet } from 'react-native'
 import Animated from 'react-native-reanimated'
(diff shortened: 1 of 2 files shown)


src/home/WalletHome.tsx
```diff
@@ -115,6 +117,9 @@ function WalletHome() {
   }
 
   const shouldShowCashInBottomSheet = () => {
+    if (!isFocused) {
+      return false
+    }
```
satish-ravi, line 122: one side effect of this change is that the bottom sheet will now appear every time the home screen gets back focus and the other conditions satisfy, whereas previously it appears only once on app load and then later only once when the balance goes from non zero to zero. I believe this is tracked by local state and this state is kept in memory as long as the conditions to show / not show never changes. With this change though, I believe we lose this state every time focus changes.

MuckT: 🤔 We could move the cash-in bottom sheet into redux state and dispatch it similar to to the in-app review.

MuckT: @satish-ravi I ended up wrapping the component in a useMemo which seems to do the trick in https://github.com/valora-inc/wallet/pull/3927/commits/500d92fff21640fd6991334543e706821bcecd87. I'll test a bit more and confirm things are WAI.

src/home/WalletHome.tsx
```diff
@@ -179,6 +155,39 @@ function WalletHome() {
...
+  )
+  const hasCelo =
+    coreTokenBalances
+      .find((token) => token.address === celoAddress)
+      ?.balance.isGreaterThan(CELO_TRANSACTION_MIN_AMOUNT) ?? false
+  const isAccountBalanceZero = hasStable === false && hasCelo === false
+
+  const WrappedCashInBottomSheet = useMemo(() => {
+    if (!isFocused) {
+      return null
```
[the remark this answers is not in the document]
jophish: I think we also want to ensure that the bottom sheet will only appear _a single time_ per app session; that is, I think the semantics should be:

* Show the bottom sheet _once_ per app session, on the _first occurrence_ of the following conditions:
  * Bottom sheet is enabled
  * User has zero balance
  * Home screen is focused
 
Namely, we want to avoid showing the bottom sheet multiple times in the situation where a user goes from zero balance -> nonzero balance -> zero balance, all within the same app session.

PR #3928, opened 2023-06-30
fix: add a test to assert validity of event names

silasbw:
And fix event names that were invalid.


cajubelt (approved):
thanks for finding this

kathaypacific (approved):
🙏🏻

PR #3411, opened 2023-01-31
fix: broken dapps support url

kathaypacific:
### Description

I forgot to update the branding hash after I added the new dapps support link, the CI was passing with [silent errors](https://github.com/valora-inc/wallet/actions/runs/4053160940/jobs/6973448494) in my branch, which i didn't notice. the CI correctly fails on [main](https://github.com/valora-inc/wallet/actions/runs/4053379964/jobs/6973931410).

### Test plan

n/a

### Related issues

- Fixes RET-550

### Backwards compatibility

Y

PR #3932, opened 2023-06-30
test(e2e): prevent empty test suite on android

MuckT:
### Description

#3923 added conditional test logic that creates an empty suite on Android which causes a test fail, but CI still passes. This PR skips those tests on android instead of conditionally excluding them on Android: [action run](https://github.com/valora-inc/wallet/actions/runs/5397315841/jobs/9802794612).

### Android Passing CI Run

```
FAIL e2e/src/FiatConnectTransferOut.spec.js
  ● Test suite failed to run

    Your test suite must contain at least one test.
```

### Test plan

Tested locally on Android

### Related issues

https://github.com/valora-inc/wallet/pull/3411

### Backwards compatibility

Yes

### DIFF
--- a/e2e/src/FiatConnectTransferOut.spec.js
+++ b/e2e/src/FiatConnectTransferOut.spec.js
@@ -5,7 +5,7 @@ import {
 import { MOCK_PROVIDER_BASE_URL, MOCK_PROVIDER_API_KEY } from 'react-native-dotenv'
 import { launchApp } from './utils/retries'
 
-describe('FiatConnect Transfer Out', () => {
+describe(':ios: FiatConnect Transfer Out', () => {
   // deliberately not doing onboarding in beforeEach, since we'll want to re-use accounts, use fresh accounts, etc for these tests
   beforeEach(async () => {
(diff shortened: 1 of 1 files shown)


e2e/src/FiatConnectTransferOut.spec.js
```diff
@@ -5,7 +5,7 @@ import {
 import { MOCK_PROVIDER_BASE_URL, MOCK_PROVIDER_API_KEY } from 'react-native-dotenv'
 import { launchApp } from './utils/retries'
 
-describe('FiatConnect Transfer Out', () => {
+describe(':ios: FiatConnect Transfer Out', () => {
```
satish-ravi, line 8: would we just move this prefix to the KYC test when we get this working on Android?

satish-ravi (approved):
Just to confirm, the failure message doesn't actually fail CI right?

e2e/src/FiatConnectTransferOut.spec.js
MuckT: Yep, that would work.

PR #3936, opened 2023-07-05
chore(e2e): Don't run QR Scanner e2e tests on Android

jh2oman:
### Description

Skip e2e QR scanner tests on android until we tackle some [tech debt with upgrading react-native-animated.](https://linear.app/valora/issue/ENG-76/[wallet]-update-react-native-reanimated-to-v3)

### Related issues

https://linear.app/valora/issue/ACT-821/update-qr-tab-bar-e2e-tests
https://linear.app/valora/issue/ENG-76/[wallet]-update-react-native-reanimated-to-v3

### DIFF
--- a/e2e/src/QRScanner.spec.js
+++ b/e2e/src/QRScanner.spec.js
@@ -1,7 +1,9 @@
 import { reloadReactNative } from './utils/retries'
 import { quickOnboarding, waitForElementId } from './utils/utils'
 
-describe('Given QR Scanner', () => {
+// Re-enable when react-native-reanimated is updated to v3
+// https://linear.app/valora/issue/ENG-76/[wallet]-update-react-native-reanimated-to-v3
+describe.skip('Given QR Scanner', () => {
   beforeAll(async () => {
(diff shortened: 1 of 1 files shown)


e2e/src/QRScanner.spec.js
```diff
@@ -1,7 +1,9 @@
 import { reloadReactNative } from './utils/retries'
 import { quickOnboarding, waitForElementId } from './utils/utils'
 
-describe('Given QR Scanner', () => {
+// Re-enable when react-native-reanimated is updated to v3
+// https://linear.app/valora/issue/ENG-76/[wallet]-update-react-native-reanimated-to-v3
+describe.skip('Given QR Scanner', () => {
```
satish-ravi, line 6: if it's an android only issue, I think we can do something like https://github.com/valora-inc/wallet/pull/3932/files#diff-b322e5ef789575e6edde98300f6e8afc87c6db925863e81bcd22d4d30e599f55R8 to disable this only for Android

satish-ravi (approved):

PR #3937, opened 2023-07-06
test(e2e): use impactMarket in dapp tests

MuckT:
### Description

Specifies [impactMarket](https://www.impactmarket.com/) as the dapp tested in the recent dapp flow. Bidali became the dapp at the top of list replacing Ubeswap when many dapps were disabled in https://github.com/valora-inc/dapp-list/pull/469 and due to it's unique flow the tests started failing.

### Test plan

Tested locally on iOS and both Android and iOS with CI.

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/e2e/src/usecases/DappListDisplay.js
+++ b/e2e/src/usecases/DappListDisplay.js
@@ -1,13 +1,21 @@
 import { fetchDappList, navigateToDappList } from '../utils/dappList'
 import { reloadReactNative } from '../utils/retries'
-import { getElementText, getElementTextList, sleep, waitForElementId } from '../utils/utils'
+import { getElementTextList, sleep, waitForElementId } from '../utils/utils'
 
 jestExpect = require('expect')
 
 export default DappListDisplay = () => {
   let dappList = null
+  let dappToTest = null
+
   beforeAll(async () => {
     dappList = await fetchDappList()
+    dappToTest = {
+      dapp: dappList.applications.find((dapp) => dapp.id === 'impactmarket'),
+      index: dappList.applications
+        .filter((dapp) => dapp[device.getPlatform() === 'ios' ? 'listOnIos' : 'listOnAndroid'])
+        .findIndex((dapp) => dapp.id === 'impactmarket'),
+    }
   })
 
   it('should show dapp info icon and subsequent modal', async () => {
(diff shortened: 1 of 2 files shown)


e2e/src/usecases/DappListDisplay.js
```diff
@@ -32,19 +40,19 @@ export default DappListDisplay = () => {
   })
 
   it('should show dapp bottom sheet when dapp is selected', async () => {
-    await element(by.id('DappCard')).atIndex(0).tap()
+    await element(by.id('DappCard')).atIndex(dappToTest.index).tap()
```
jeanregisser, line 43: Why not finding the index from `impactMarket` here?
I'm slightly concerned the `fetchDappList` order could be different since it doesn't pass the wallet address. And we've done experiments in the past changing the order based on it.

e2e/src/usecases/DappListRecent.js
```diff
@@ -28,7 +28,8 @@ export default DappListRecent = () => {
 
     // Scroll to first dapp or next after most recent dapp
     await scrollToDapp(startRecentDappCount + 1)
-    await element(by.id('DappCard')).atIndex(startRecentDappCount).tap()
+    // Tap impact Market as it is the second dapp in the list
```
jeanregisser, line 31: Why is it always the case?

jeanregisser (commented):
Thanks!

e2e/src/usecases/DappListDisplay.js
jeanregisser: Or why not incrementing the index until we get something that's not Bidali?
This way we don't have to rely too much on the fetchDappList data.

e2e/src/usecases/DappListRecent.js
MuckT: The e2e test wallet is set to always receive the control in Statsig: [console](https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/experiments/dapps_shuffle/setup), but a new dapp could be added that would take it's place.

jeanregisser: Thanks, would be nice to leave this as a comment in the code.

jeanregisser (approved):

e2e/src/usecases/DappListRecent.js
MuckT: Added a comment in https://github.com/valora-inc/wallet/pull/3937/commits/b9d876920c99651a0d320038e33893a282057731!

PR #3940, opened 2023-07-06
fix(nft-image): use unique key in dependency array and remove use memo

MuckT:
### Description

Nft images in the transaction feed could get into a persisted loading state on Android.

#### Android Before

https://github.com/valora-inc/wallet/assets/26950305/c867fb86-2cff-4dd4-bc56-06da9ebc766d

#### Android After

https://github.com/valora-inc/wallet/assets/26950305/af022787-6784-45f0-a97b-de4142035e89

### Test plan

Tested locally on Android 

### Related issues

- #3903

### Backwards compatibility

Yes

### DIFF
--- a/src/nfts/NftImage.tsx
+++ b/src/nfts/NftImage.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useState } from 'react'
+import React, { useEffect, useState } from 'react'
 import { View } from 'react-native'
 import FastImage from 'react-native-fast-image'
 import { NftEvents } from 'src/analytics/Events'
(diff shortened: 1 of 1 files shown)


dievazqu (approved):

src/nfts/NftImage.tsx
```diff
@@ -71,7 +68,7 @@ export default function NftImage({
       sendImageLoadEvent('No nft metadata')
       setStatus('error')
     }
-  }, [nft])
+  }, [`${nft.contractAddress}-${nft.tokenId}`])
```
jeanregisser, line 71: Could we do this instead?
```suggestion
  }, [imageUrl])
```

And use `imageUrl` in the condition inside?

MuckT: Weirdly enough image url isn't unique. Ipermatch for example has multiple NFTs with the same image url.

PR #3941, opened 2023-07-06
test(e2e): use nft viewer instead of impactMarket in dapp display tests

MuckT:
### Description

Follow up to https://github.com/valora-inc/wallet/pull/3937, where impactMarket was used in all the Dapp tests; however, that domain appears slow to load and is causing failures in our e2e test suites. Instead we should test again a domain we control such as the NFT viewer.

### Test plan

Tested locally on Android and in CI.

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/e2e/src/usecases/DappListDisplay.js
+++ b/e2e/src/usecases/DappListDisplay.js
@@ -1,7 +1,6 @@
 import { fetchDappList, navigateToDappList } from '../utils/dappList'
 import { reloadReactNative } from '../utils/retries'
-import { getElementTextList, sleep, waitForElementId } from '../utils/utils'
-import { scrollToDapp } from '../utils/dappList'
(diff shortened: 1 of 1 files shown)


silasbw (approved):

jophish (approved):

cajubelt (approved):

e2e/src/usecases/DappListDisplay.js
```diff
@@ -40,7 +40,9 @@ export default DappListDisplay = () => {
   })
 
   it('should show dapp bottom sheet when dapp is selected', async () => {
-    await element(by.id('DappCard')).atIndex(dappToTest.index).tap()
+    await sleep(2000)
```
dievazqu, line 43: Is this needed? Is this enough?
Sounds like a potential situation of a flaky test.

PR #3943, opened 2023-07-07
feat: enable hooks preview

jeanregisser:
### Description

This adds support for enabling hooks preview in Valora by scanning a QR code or opening a deep link.
It shows an alert to confirm hooks preview should be enabled.

There's a statsig feature gate to allow enabling hooks preview.

See also the PR which adds the banner when preview mode is enabled: https://github.com/valora-inc/wallet/pull/3946

### Test plan

- Updated tests
- Manually scanning the QR code from the [hooks developer server](https://github.com/valora-inc/hooks/pull/121):

 

### Related issues

- Fixes RET-750

### Backwards compatibility

Yes


dievazqu (approved):
Looks good!

Just wondering... Is it worth adding some metrics of the usage of this feature? I think we don't need them since this would be a feature for developer most likely, but I'm just mentioning it in case you find a use case for them.

jeanregisser:
Yep, I'm planning to add analytics in a separate PR 👍

PR #3951, opened 2023-07-12
feat: add claim shortcuts screen

kathaypacific:
### Description

As the title 
- add plumbing to navigate to a new claimable shortcuts screen
- UI for displaying claimable shortcuts
- the screen will only be navigatable when there are claimable shortcuts, so the screen does not cater for an empty or error state

Not included in this PR
- claiming reward
- analytics

### Test plan

Simulator Screenshot - iPhone 14 Pro - 2023-07-13 at 10 19 04

### Related issues

- Fixes RET-733

### Backwards compatibility

Y


src/styles/styles.ts
```diff
@@ -3,6 +3,7 @@ import { StyleSheet } from 'react-native'
 const BASE_UNIT = 8
 
 export enum Spacing {
+  Tiny4 = BASE_UNIT / 2,
```
kathaypacific, line 6: i was seeing 4 more and more in designs so pulled the trigger to add this

test/values.ts
```diff
@@ -1209,7 +1209,7 @@ export const mockPositions: Position[] = [
         category: 'claimable',
         decimals: 18,
         network: 'celo',
-        balance: '0.098322815093446616',
+        balance: '0.098322815093446616', // total USD value = priceUsd * balance = $0.00009
```
kathaypacific, line 1212: @jeanregisser is this a real scenario? i've got a claimable token on the `tokens` level, and another in the `tokens[0].tokens` level, so what would be displayed here is both these tokens as the reward?

src/dapps/DappShortcutsRewards.tsx
```diff
@@ -0,0 +1,172 @@
...
+  const localCurrencySymbol = useSelector(getLocalCurrencySymbol)
+  const localCurrencyExchangeRate = useSelector(localCurrencyExchangeRatesSelector)
+  const positionsWithClaimableRewards = useSelector(positionsWithClaimableRewardsSelector)
+
+  const handleClaimReward = (position: ClaimablePosition) => () => {
+    // do something
+  }
+
+  const renderItem = ({ item }: { item: ClaimablePosition }) => {
+    let claimableFiatValue = new BigNumber(0)
```
jeanregisser, line 34:
```suggestion
    let claimableValueUsd = new BigNumber(0)
```

src/dapps/DappShortcutsRewards.tsx
```diff
@@ -0,0 +1,172 @@
...
+                .map((token) => `${new BigNumber(token.balance).toFixed(5)} ${token.symbol}`)
+                .join(', ')}
+            </Text>
+            {claimableValueLocalCurrency && (
+              <Text style={styles.rewardFiatAmount}>
+                {`${localCurrencySymbol} ${claimableValueLocalCurrency.toFixed(
+                  claimableValueLocalCurrency.lt(0.01) ? 5 : 2
+                )}`}
+              </Text>
+            )}
```
jeanregisser, line 64: Could we use our existing `TokenDisplay` component instead?
I'd prefer we avoid custom formatting whenever possible.

test/values.ts
jeanregisser: Good question, this wouldn't happen for real with Ubeswap farms, but I guess it's possible it could happen with more complex vault / yield farming.

jeanregisser (commented):
Awesome!! 🚀 
Can't wait to see this live.

src/dapps/DappShortcutsRewards.tsx
kathaypacific: yes that makes sense, i made the change!

PR #3953, opened 2023-07-12
feat: add analytics events around hooks preview mode

jeanregisser:
### Description

Title says it all ;)

### Test plan

- Updated unit tests

### Related issues

- Fixes RET-750

### Backwards compatibility

Yes


kathaypacific (approved):
nice!!

a side thought i had, do you think it's worth adding a super property to indicate whether the app is in preview mode? it probably does not matter much, but i guess the builder behaviour in preview mode will be quite different compared to a normal user. it might be interesting to understand what builders are doing / exclude their activity if we need to?

jeanregisser:
Yes I was thinking about this too. Now you convinced me I should add it 👍

kathaypacific (approved):
⛴️

PR #3955, opened 2023-07-13
feat(wc): wip web3wallet migration sans redux sessions

MuckT:
### Related issues

- Fixes #[issue number here]


src/walletConnect/v2/saga.ts
```diff
@@ -345,21 +350,22 @@ export function* acceptSession({ session }: AcceptSession) {
...
     throw new Error('missing client')
   }
 
-  let sessionValue = client.session.values.find(
-    (value) => value.peer.publicKey === session.params.proposer.publicKey
-  )
-
+  let sessionValue: null | Web3WalletTypes.SessionProposal = null
   while (!sessionValue) {
+    // TODO: add correct type
```
kathaypacific, line 361: it looks like this should be `Record<string, SessionTypes.Struct>`?

src/walletConnect/v2/saga.ts
```diff
@@ -345,21 +350,22 @@ export function* acceptSession({ session }: AcceptSession) {
...
   if (!client) {
     // should not happen
     throw new Error('missing client')
   }
 
-  let sessionValue = client.session.values.find(
-    (value) => value.peer.publicKey === session.params.proposer.publicKey
-  )
-
+  let sessionValue: null | Web3WalletTypes.SessionProposal = null
```
kathaypacific, line 359: i think we should try to assign this instead of declaring as null - the current flow you have is
1) sessionValue = null
2) try to assign session value
3) wait 500ms 
4) repeat from 2 if 2 failed to assign a value

there will always be a 500ms wait in the above loop, even if step 2 passes the first time. while not terribly blocking, it's still an unnecessary delay...

kathaypacific (commented):
This is looking great, thanks for tackling this!! i had mainly small superficial comments for now.

for my understanding, from this PR it looks like the main difference between web3wallet and signClient is the session management - web3wallet provides a way to get the active sessions from the client itself so we no longer need to manage this in redux. is that right? are there any other key differences? 

i had two thoughts - 1) we should probably remove the sessions from redux if we don't need them anymore, and 2) i recall web3wallet doing something different with the auth signing where we might need to adjust the UI, i forget the details but i don't see any changes related to that here. is that planned for a follow up PR?

package.json
```diff
@@ -109,9 +109,10 @@
     "@sentry/types": "^7.29.0",
     "@th3rdwave/react-navigation-bottom-sheet": "^0.2.2",
     "@walletconnect/client": "^1.8.0",
-    "@walletconnect/react-native-compat": "^2.8.1",
-    "@walletconnect/sign-client": "^2.3.1",
+    "@walletconnect/react-native-compat": "^2.9.0",
+    "@walletconnect/sign-client": "^2.9.0",
```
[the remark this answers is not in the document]
MuckT: The e2e dapp for testing is using still using the sign client, but I can try moving this to dev dependencies.

src/walletConnect/saga.ts
```diff
@@ -28,6 +28,7 @@ export function* initialiseWalletConnect(uri: string, origin: WalletConnectPairi
   const walletConnectEnabled: boolean = yield call(isWalletConnectEnabled, uri)
 
   const [, , version] = uri.split(/[:@?]/)
+  // DEV - Add false to the if statement below to test on iOS emulator
```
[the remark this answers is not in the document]
MuckT: I was a bit confused as to why this was needed in iOS. Both `remoteConfigValuesDefaults.ts` and the remote configs for mainnet have `walletConnectV2Enabled` set to true. Checking in Firebase the default is false and 8% of users are getting that value - which seems strange.

src/walletConnect/v2/saga.ts
MuckT: Fixed in fade04c04e203e8cf7e4b7b142a760e4626f2a26!

MuckT:
@kathaypacific for this PR I tried to maintain feature parity while moving to web3wallet. Many of the [exciting features of web3wallet](https://github.com/WalletConnect/walletconnect-docs/blob/main/docs/web/web3wallet/wallet-usage.md#updating-a-session) such as: auth, wallet events and Android deeplinks are not included in this PR.

Web3wallet gives us session management, but unfortunately the 'Sessions.tsx' is still reading from redux. As such I've maintained the state in redux, but when we need the most up to date info we look to the client instead as in `getSessionFromRequest` and `handleAcceptRequest`.

Another difference is no equivalent to `SignClientTypes.EventArguments['session_update']` in web3wallet; however, it doesn't appear we've lost any utility by removing it.

> I recall web3wallet doing something different with the auth signing where we might need to adjust the UI...?

👀  we might need to adjust `handleAcceptRequest` for transaction signing - otherwise users might see a success message after a signing a transaction with a dapp.

package.json
```diff
@@ -78,7 +78,7 @@
     "@fiatconnect/fiatconnect-types": "^13.0.2",
     "@gorhom/bottom-sheet": "^4.4.7",
     "@json-rpc-tools/utils": "^1.7.6",
-    "@react-native-async-storage/async-storage": "^1.17.11",
+    "@react-native-async-storage/async-storage": "^1.19.0",
```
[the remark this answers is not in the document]
MuckT: It is not and it raises the minAndroid SDK to Android 6. Changed back to `1.17.11` in 52f5ec1a7f18b60e7bcf439b79b8ebef8f2b1179.

src/walletConnect/v2/saga.ts
```diff
@@ -107,18 +108,26 @@ function applyIconFixIfNeeded(
...
   if (!client) {
     Logger.debug(TAG + '@createWalletConnectChannel', `init start`)
 
-    client = yield call([SignClient, 'init'], {
-      logger: 'debug',
-      projectId: WALLET_CONNECT_PROJECT_ID,
-      relayUrl: networkConfig.walletConnectEndpoint,
+    // @ts-expect-error client is not assignable to type IWeb3Wallet
+    client = yield call([Web3Wallet, 'init'], {
+      core, // <- pass the shared `core` instance
```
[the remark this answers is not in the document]
MuckT: Changed in a8dda663a524e04b288ec1bc828d975717520af6!

silasbw:
@idelburgo-iomob wrote up this native test app: https://github.com/idelburgo-iomob/walletConnectIntegrationExample.  If it's not too much work, can you test this PR with it @MuckT ?

MuckT:
@silasbw I'll keep that native Android app for testing deeplinks, but that fix will be in a separate PR.

MuckT:
@jeanregisser we might want to revert 97dcf9bb36a4082c5b8f11fce8f245b98367df14 and 93b4435cbe7956d71e13d7df0f209dc5f03e975f prior to merging, but it looks as if our namespace handling is off.

MuckT:
I am occasionally seeing this error when attempt to connect to a dapp which already has an existing session. The session appears to pair correctly and follow up actions on the dapp are being handled correctly. The issue I am seeing now is that the timeout is still triggering after the successful action 🤔.

```
ERROR  [23:38:04.600Z] SM-G991U1 {"context": "core"} Pairing already exists: 2f5e3c5332ec900345e3a82430dc9e5311eadefbc31d79b3b89b0bf6db072d90
```

src/walletConnect/saga.ts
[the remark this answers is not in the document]
jeanregisser: It's probably old users where V2 is disabled there, no?

src/walletConnect/walletConnect.ts
```diff
@@ -73,6 +73,7 @@ export function* handleLoadingWithTimeout(origin: WalletConnectPairingOrigin) {
     sessionRequestReceivedV1: take(ActionsV1.SESSION_V1),
     sessionRequestReceivedV2: take(ActionsV2.SESSION_PROPOSAL_V2),
     actionRequestReceived: take(ActionsV1.PAYLOAD_V1),
+    actionRequestReceivedV2: take(ActionsV2.SESSION_PAYLOAD_V2),
```
jeanregisser, line 76: Good catch! This was causing real issues, right?

src/walletConnect/v2/saga.ts
```diff
@@ -67,14 +67,14 @@ import {
 import networkConfig from 'src/web3/networkConfig'
 import { getWalletAddress } from 'src/web3/saga'
 
-let client: SignClient | null = null
+export let client: IWeb3Wallet | null = null
```
jeanregisser, line 70: why exporting this now?

jeanregisser (commented):
Awesome! 🚀

src/walletConnect/walletConnect.ts
MuckT: Yep! This is likely why we had so many reports of timeouts despite not having many errors on transactions.

src/walletConnect/v2/saga.ts
MuckT: Fixed in ec3c8885e1dded575c1c79f62df8debba905775a!

src/walletConnect/saga.ts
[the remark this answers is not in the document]
MuckT: Old versions could be it. I did find it strange that despite the remote configs set to be enabled I was still getting a false value when testing a local build on iOS.

jeanregisser (approved):
🚀

PR #3957, opened 2023-07-13
feat: enable trigger shortcuts

kathaypacific:
### Description

This PR does 2 things (that made sense to test together)
1. add a saga for triggering a shortcut
2. add some redux state management to display available / claiming / claimed statuses on the rewards screen

The logic for 2 is a little more hefty than I originally planned, i may have overthought the scenario. What i wanted to cater for is:
1. a user claims a reward successfully
2. their balance and positions are refreshed in the background while the user remains on the rewards screen
3. they should see that their reward was claimed, but the positions may refresh and the claimable parts removed. so the screen essentially tracks all the data it was loaded with, and if any rewards go "missing" from the redux store then we mark it as "claimed".
4. to make this a little robust for rewards that can be continuously claimed (i.e. on successful claim, the reward remains but the reward amount changes) i made up a reward id that relies on a combination of reward amount and reward address.

### Test plan

https://github.com/valora-inc/wallet/assets/20150449/595bb6ae-26fc-426f-aede-12356929022f

### Related issues

- Fixes RET-733

### Backwards compatibility

Y

### DIFF
--- a/src/positions/saga.ts
+++ b/src/positions/saga.ts
@@ -12,6 +12,7 @@ import { HooksEnablePreviewOrigin } from 'src/analytics/types'
 import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
 import { ErrorMessages } from 'src/app/ErrorMessages'
 import { DEFAULT_TESTNET } from 'src/config'
+import { refreshAllBalances } from 'src/home/actions'
 import i18n from 'src/i18n'
 import {
   hooksApiUrlSelector,
@@ -250,6 +251,7 @@ export function* triggerShortcutSaga({ payload }: ReturnType<typeof triggerShort
       Toast.SHORT,
       Toast.BOTTOM
     )
+    yield put(refreshAllBalances())
   } catch (error) {
     yield put(triggerShortcutFailure(payload.id))
     // TODO customise error message when there are more shortcut types
--- a/src/positions/saga.ts
+++ b/src/positions/saga.ts
@@ -252,14 +252,14 @@ export function* triggerShortcutSaga({ payload }: ReturnType<typeof triggerShort
       Logger.debug(`${TAG}/triggerShortcutSaga`, 'Claimed reward successful', receipt)
     }
 
-    yield put(triggerShortcutSuccess(payload.positionAddress))
+    yield put(triggerShortcutSuccess(payload.id))
     Toast.showWithGravity(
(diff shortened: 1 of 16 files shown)


src/positions/saga.ts
```diff
@@ -188,6 +201,65 @@ export function* handleEnableHooksPreviewDeepLink(
...
+      )
+    }
+
+    yield put(triggerShortcutSuccess(payload.id))
+    Toast.showWithGravity(
+      i18n.t('dappShortcuts.claimRewardsScreen.claimSuccess'),
+      Toast.SHORT,
+      Toast.BOTTOM
+    )
+    yield put(refreshAllBalances())
```
jeanregisser, line 254: No need to do it now, but I would be great if the code refreshing balances would listen to `triggerShortcutSuccess` too to decide to refresh all balances.
Small step to avoid the many dispatch anti-pattern we have in the code base.

src/positions/saga.ts
```diff
@@ -188,6 +201,65 @@ export function* handleEnableHooksPreviewDeepLink(
...
+      applyChainIdWorkaround(transaction, yield call([kit.connection, 'chainId']))
+      const tx: CeloTx = yield call([normalizer, 'populate'], transaction)
+      const txo = buildTxo(kit, tx)
+
+      const receipt: CeloTxReceipt = yield call(
+        sendTransaction,
+        txo,
+        walletAddress,
+        newTransactionContext(TAG, 'Trigger shortcut')
+      )
```
jeanregisser, line 239: I think If there were really multiple transactions to send, this would wait unnecessarily for each of them to be included in different blocks. Instead of sending them all in parallel.

We don't need to address this now, just noting. Maybe we could add a TODO about this?

jeanregisser (approved):
Amazing work! 🚀 🚀 🚀 

One thing I'm debating is whether we want the explicit bottom sheet for signing like we initially designed this. Trading convenience for security. But arguably our current TX signing bottom sheet doesn't help that much with it, except for the explicit step and wording about Valora not being affiliated with the dapp asking to sign something.

Let's chat about this?

src/positions/saga.ts
kathaypacific: yeah good shout, updated

src/positions/saga.ts
kathaypacific: i always get confused about this topic, won't we run into nonce issues if we try to send multiple transactions at the same time? i added a TODO for now, but happy to address in a follow up

PR #3958, opened 2023-07-13
chore(recovery): Cleanup recovery experiment and remove forced backup

jophish:
### Description

For ACT-852. Removes all code related to the recovery phrase experiment in onboarding, and removes forced backup from the app!

### Test plan

Unit and manual tested.

### Related issues

- Fixes ACT-852.

### Backwards compatibility

Yes.


src/onboarding/steps.ts
```diff
@@ -232,18 +226,9 @@ export function _getStepInfo({ firstScreenInStep, navigator, dispatch, props }:
           } else if (choseToRestoreAccount) {
             popToScreen(Screens.Welcome)
             navigate(Screens.ImportWallet)
-          } else if (showRecoveryPhrase) {
+          } else {
             dispatch(initializeAccount())
             navigate(Screens.ProtectWallet)
-          } else if (skipVerification) {
```
satish-ravi, line 238: I guess we don't need these since the next step is always going to be protect wallet, but what was the scenario where verification is already skipped when the user is on the pincode / biometry screen?

src/backup/BackupIntroduction.tsx
```diff
@@ -91,28 +88,12 @@ interface AccountKeyStartProps {
...
-        .useNewBackupFlowCopy
-        ? {
-            title: t('introBackUpPhrase'),
-            body: t('introCompleteQuiz'),
-            primaryAction: t('continue'),
-          }
-        : {
-            title: t('introTitle'),
-            body: t('introBody'),
-            primaryAction: t('introPrimaryAction'),
```
satish-ravi, line 106: can these strings be removed from translation.json?

src/account/reducer.ts
satish-ravi, line 1: can you update the comments on the RecoveryPhraseInOnboardingStatus? Guess we can remove the variant references

src/analytics/Events.tsx
satish-ravi, line 1: Could we remove other unused events as well? I found `backup_delay`, but there could be others

satish-ravi (commented):
Looks good, just a few minor things

src/onboarding/steps.ts
jophish: I'm not totally certain.. I don't know if the `skipVerification` condition was here defensively for completeness or if it was actually reachable; either way since `showRecoveryPhrase` is effectively always `true` now, we can get rid of any conditions that come after it.

src/backup/BackupForceScreen.tsx
```diff
@@ -1,100 +0,0 @@
...
-    return () => BackHandler.removeEventListener('hardwareBackPress', backPressListener)
-  }, [])
-
-  return (
-    <SafeAreaView style={styles.container}>
-      <DevSkipButton onSkip={skipBackup} />
-      <ScrollView contentContainerStyle={styles.content}>
-        <Logo height={32} />
-        <Text style={styles.title}>{t('backupSetupTitle')}</Text>
-        <Text style={styles.body}>{t('backupSetupBody')}</Text>
```
jh2oman, line 62: nit: i think these translations can be removed too

src/backup/DelayButton.tsx
```diff
@@ -1,29 +0,0 @@
...
-    ValoraAnalytics.track(OnboardingEvents.backup_delay)
-    navigateHome()
-  }, [dispatch])
-
-  const { t } = useTranslation()
-
-  if (!shouldShowDelayButton) {
-    return null
-  }
-  return <TopBarTextButton onPress={onPressDelay} title={t('delayBackup')} />
```
jh2oman, line 28: and this translation.

src/home/NotificationBox.tsx
```diff
@@ -132,21 +129,15 @@ function useSimpleActions() {
 
   const actions: SimpleMessagingCardProps[] = []
   if (!backupCompleted) {
-    const { useNewBackupHomeCard } = getExperimentParams(
-      ExperimentConfigs[StatsigExperiments.RECOVERY_PHRASE_IN_ONBOARDING]
-    )
-    const text = useNewBackupHomeCard ? t('backupKeyNotification2') : t('backupKeyNotification')
```
jh2oman, line 138: `backupKeyNotification`too

src/onboarding/registration/ProtectWallet.tsx
```diff
@@ -90,53 +70,9 @@ function ProtectWallet({ navigation }: Props) {
...
-        onPress={onPressRecoveryPhrase}
-      />
-    )
-    if (showCloudBackupFakeDoor) {
-      const cloudBackupCard = (
-        <OnboardingCard
-          testId={'cloudBackupCard'}
-          key={'cloudBackupCard'}
-          title={t('protectWallet.cloudBackup.title')}
-          subtitle={t('protectWallet.cloudBackup.subtitle')}
```
jh2oman, line 129: I'm pretty sure all of `protectWallet.cloudBackup` too

src/onboarding/steps.ts
jh2oman: > but what was the scenario where verification is already skipped when the user is on the pincode / biometry screen?

`skipVerification` is actually a firebase remote config, not if the user has pressed 'skip' on the Verification screens. It was last used when decentralized phone verification went down for some time.

like joe said though, not necessary here anymore because ProtectWallet is always next

jh2oman (approved):
Nits with the translations, otherwise looks great!

PR #3960, opened 2023-07-18
fix: ensure dapp shortcut rewards value is refreshed

kathaypacific:
### Description

When a shortcut is claimed, we refetch the positions. When the positions data is updated, the claimable rewards could either disappear or the value could change (i.e. there is a new reward to claim). In the latter scenario, we need to also refresh the value of the reward displayed on the UI - this change makes the UI rely only on positions data in redux when available, and use the locally stored state only if the claimable position disappears.

### Test plan

https://github.com/valora-inc/wallet/assets/20150449/6a316091-f02c-49e9-9704-61479e7a5fb5

### Related issues

- Fixes RET-733

### Backwards compatibility

Y


jeanregisser (approved):
🚀

PR #3961, opened 2023-07-18
chore(deps): Use typed-redux-saga

jh2oman:
### Description

Using the [typed-redux-saga](https://github.com/agiledigital/typed-redux-saga) package to bring type checking to our sagas. This is in preparation for bumping our typescript version.

* First I used this [eslint rule](https://github.com/agiledigital/typed-redux-saga#eslint-rules) which adds the * to the yields and fixes all of the imports
* Then I went through and fixed all of the broken types we had (there were many 💀 )
* Most of it is very small nits, but i tried to add comments to highlight the more interesting parts.

### Test plan

Updated unit tests

### Related issues

Preparing for https://linear.app/valora/issue/ACT-784/typed-contracts-with-typechain


jh2oman:
I think codecov is having trouble understanding the yield* generator format

src/app/saga.ts
```diff
@@ -145,32 +145,34 @@ export function* appInit() {
...
   // This step is important if the user is offline and unable to fetch remote
   // config values, we can use the persisted value instead of an empty one
-  const sentryNetworkErrors = yield select(sentryNetworkErrorsSelector)
+  const sentryNetworkErrors = yield* select(sentryNetworkErrorsSelector)
   Logger.setNetworkErrors(sentryNetworkErrors)
 
-  const supportedBiometryType = yield call(Keychain.getSupportedBiometryType)
-  yield put(setSupportedBiometryType(supportedBiometryType))
+  const supportedBiometryType = yield* call(Keychain.getSupportedBiometryType)
+  if (supportedBiometryType) {
```
jh2oman, line 152: can be null

src/app/saga.ts
```diff
@@ -463,30 +465,37 @@ export function* runCentralPhoneVerificationMigration() {
...
-    const signedMessage = yield call(retrieveSignedMessage)
+    const signedMessage = yield* call(retrieveSignedMessage)
     if (!signedMessage) {
       Logger.warn(
         `${TAG}@runCentralPhoneVerificationMigration`,
         'No signed message was found for this user. Skipping CPV migration.'
       )
       return
     }
+    if (!phoneNumber) {
```
jh2oman, line 482: can be null

src/app/saga.ts
```diff
@@ -532,16 +541,16 @@ export function* requestInAppReview() {
   )
     return
 
-  const lastInteractionTimestamp = yield select(inAppReviewLastInteractionTimestampSelector)
+  const lastInteractionTimestamp = yield* select(inAppReviewLastInteractionTimestampSelector)
   const now = Date.now()
 
   // If the last interaction was less than a quarter year ago or null
-  if (now - lastInteractionTimestamp >= REVIEW_INTERVAL) {
+  if (!lastInteractionTimestamp || now - lastInteractionTimestamp >= REVIEW_INTERVAL) {
```
jh2oman, line 548: can be null

src/components/__snapshots__/CircleButton.test.tsx.snap
```diff
@@ -66,11 +66,7 @@ exports[`CircleButton renders correctly with minimum props 1`] = `
       xmlns="http://www.w3.org/2000/svg"
     >
       <path
-        d="M2.141,89.13c1.425,1.429,3.299,2.142,5.167,2.142c1.869,0,3.742-0.713,5.167-2.142l33.591-33.592L79.657,89.13
-          c1.426,1.429,3.299,2.142,5.167,2.142c1.867,0,3.74-0.713,5.167-2.142c2.854-2.854,2.854-7.48,0-10.334L56.398,45.205
```
jh2oman, line 70: these three snaps are the same, just different indentations.

src/firebase/firebase.ts
```diff
@@ -382,7 +382,7 @@ export async function fetchCoinbasePaySenders() {
 
 async function fetchListFromFirebase(path: string) {
   if (!FIREBASE_ENABLED) {
-    return []
+    return null
```
jh2oman, line 385: make the return type `EventChannel<NotUndefined> | null` which is more clear than an empty list

src/identity/selectors.ts
```diff
@@ -40,6 +40,6 @@ export const isBalanceSufficientForSigRetrievalSelector = createSelector(
   (tokens) => {
     const cusdBalance = tokens[Currency.Dollar]?.balance
     const celoBalance = tokens[Currency.Celo]?.balance
-    isBalanceSufficientForSigRetrieval(cusdBalance || 0, celoBalance || 0)
+    return isBalanceSufficientForSigRetrieval(cusdBalance || 0, celoBalance || 0)
```
jh2oman, line 43: not sure this was ever working...

src/import/saga.ts
```diff
@@ -70,13 +70,15 @@ export function* importBackupPhraseSaga({ phrase, useEmptyWallet }: ImportBackup
...
       try {
-        const { correctedPhrase, timeout } = yield race({
+        const { correctedPhrase, timeout } = yield* race({
           correctedPhrase: call(attemptBackupPhraseCorrection, normalizedPhrase),
           timeout: delay(MNEMONIC_AUTOCORRECT_TIMEOUT),
         })
         if (correctedPhrase) {
           Logger.info(TAG + '@importBackupPhraseSaga', 'Using suggested mnemonic autocorrection')
-          mnemonic = correctedPhrase
+          // there is a bug with 'race' in typed-redux-saga, so we need to hard cast the result
```
jh2oman, line 79: some annoying bugs with race & all

silasbw:
Nice :raised_hands: !

src/web3/contracts.ts
```diff
@@ -113,18 +113,18 @@ async function waitForContractKit(tries: number) {
 
 export function* getContractKit() {
   if (!contractKit) {
-    yield initContractKitLock.acquire()
+    yield* call([initContractKitLock, initContractKitLock.acquire])
```
satish-ravi, line 116:
```suggestion
    yield* call([initContractKitLock, 'acquire'])
```

src/web3/dataEncryptionKey.ts
```diff
@@ -51,53 +50,54 @@ import {
...
     walletToAccountAddress[normalizeAddressWith0x(walletAddress)] ?? walletAddress
-  const dek: string = yield call(accountsWrapper.getDataEncryptionKey, accountAddress)
-  yield put(updateAddressDekMap(accountAddress, dek || null))
-  return !dek ? null : hexToBuffer(dek)
+  const dek: string = yield* call(accountsWrapper.getDataEncryptionKey, accountAddress)
+  yield* put(updateAddressDekMap(accountAddress, dek || null))
+  return dek
 }
 
 export function* doFetchDataEncryptionKey(walletAddress: string) {
```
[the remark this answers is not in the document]
satish-ravi: Do the dependencies of this assume non null? If so, do they need to be updated to handle nulls?

satish-ravi (commented):
Looks mostly good. Nice job!

Few things:
- So with this PR we get typings for sagas, but lose code coverage because jest doesn't work with yield* right?
- I started leaving comments on removing explicit types for yield* call, but seems like its removed only in a few places, so feel free to ignore them. But eventually I guess we can remove the explicit types since they all should be inferred now
- Do we need to specify the type for `yield* take` always? 
- There are a few places where we do `as unknown as type` without any comments, are these the same as the bug you mentioned in some places?

src/web3/saga.ts
```diff
@@ -154,25 +154,26 @@ function* getAddress<T extends { address: string | null }>({
...
   while (true) {
-    const account = yield select(addressSelector)
+    const account = yield* select(addressSelector)
     if (account) {
       return account
     }
 
-    const actionEffect: T = yield take(action)
-    if (actionEffect.address) {
+    const actionEffect = yield* take(action)
```
jeanregisser, line 162: nit: can we do something like this instead?
```suggestion
    const actionEffect = (yield* take(action) as T)
```

jeanregisser (commented):
Epic work!! 
Can't express my gratitude enough ❤️ 

Left a couple of small comments.
Looks like we can drop a few more types, but no big deal.

Too bad about the coverage changing. It seems it's caused by the automatic replacement of `yield*` via the macros.

~~Maybe this will get fixed with an upgrade to jest? I'm not too hopeful though.~~
Update: see https://github.com/valora-inc/wallet/pull/3961#discussion_r1270757849

babel.config.js
```diff
@@ -34,6 +34,7 @@ module.exports = {
     '@babel/plugin-transform-flow-strip-types', // for a bug in FlatList caused by adding @babel/plugin-proposal-class-properties
     '@babel/plugin-proposal-class-properties', // for ethers
     '@babel/plugin-transform-private-methods', // for ethers
+    'macros',
```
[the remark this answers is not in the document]
jeanregisser: Why is it needed?
I looked at https://github.com/celo-tools/celo-web-wallet/blob/master/.babelrc and https://github.com/Uniswap/wallet/blob/main/apps/mobile/babel.config.js they both use `typed-redux-saga` but I can't yet find this there as well.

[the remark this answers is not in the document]
jeanregisser: It seems it may be unnecessary.
Not that both the projects listed above import like this:
`import { call, put } from 'typed-redux-saga'`
instead of 
`import { call, put } from 'typed-redux-saga/macro'` like we did in this PR.

Can we do the same?
Maybe this will help with the code coverage issue too?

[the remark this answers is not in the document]
jeanregisser: Answering myself, after re-reading the README for https://github.com/agiledigital/typed-redux-saga 😄 

Not entirely sure about the overhead of not using `macros` but it seems it would be quite [lite](https://github.com/agiledigital/typed-redux-saga/blob/master/src/index.js).
If this helps with code coverage, I think we should probably favor it.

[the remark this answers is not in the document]
jh2oman: Without macros all of our redux-saga-test-plan tests were failing with various issues. I couldn't get it to work. See branch here: https://github.com/valora-inc/wallet/pull/3991

jest.config.js
```diff
@@ -40,6 +40,7 @@ module.exports = {
   transform: {
     ...tsjPreset.transform,
     '^.+\\.(txt)$': require.resolve('./node_modules/react-native/jest/assetFileTransformer.js'),
+    '^.+\\.tsx?$': 'babel-jest',
```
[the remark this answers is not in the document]
jh2oman: Ah this is leftover from something I was trying when attempting to avoid using /macros. I'll remove

src/components/__snapshots__/CircleButton.test.tsx.snap
jh2oman: It was this which isn't needed. They'll go back to the way they were :) https://github.com/valora-inc/wallet/pull/3961/files#r1270625689

src/firebase/notifications.ts
```diff
@@ -101,25 +101,25 @@ export function* handleNotification(
...
 
   switch (message.data?.type) {
     case NotificationTypes.PAYMENT_REQUESTED:
-      yield call(handlePaymentRequested, message.data as unknown as PaymentRequest)
+      yield* call(handlePaymentRequested, message.data as unknown as PaymentRequest)
       break
 
     case NotificationTypes.PAYMENT_RECEIVED:
-      yield call(handlePaymentReceived, message.data as unknown as TransferNotificationData)
+      handlePaymentReceived(message.data as unknown as TransferNotificationData)
```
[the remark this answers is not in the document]
jh2oman: handlePaymentReceived is not an async function

src/identity/commentEncryption.ts
```diff
@@ -170,19 +174,19 @@ export function* checkTxsForIdentityMetadata({ transactions }: NewTransactionsIn
...
 
     const newIdentityData = findIdentityMetadataInComments(transactions, dataEncryptionKey)
 
-    const verifiedMetadata: IdentityMetadataInTx[] = yield call(
+    const verifiedMetadata = (yield* call(
       verifyIdentityMetadata,
       newIdentityData
-    )
-    yield call(updatePhoneNumberMappings, verifiedMetadata)
+    )) as unknown as IdentityMetadataInTx[]
```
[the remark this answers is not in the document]
jh2oman: I'll add a comment. this is one of the cases where the types are messed up because of a bug with 'race' in  verifyIdentityMetadata

src/web3/dataEncryptionKey.ts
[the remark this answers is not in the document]
jh2oman: no, the dependencies assumed it could be nulll

src/web3/saga.ts
jh2oman: I can move it up there, but we still have to cast as unknown first

src/app/saga.ts
jh2oman: will do

jh2oman:
> So with this PR we get typings for sagas, but lose code coverage because jest doesn't work with yield* right?

correct

> Do we need to specify the type for yield* take always?

No, only for channels that don't specify a type. Some, like the WalletConnectActions channel do that.

> There are a few places where we do as unknown as type without any comments, are these the same as the bug you mentioned in some places?

most of them are. I've updated with more comments

babel.config.js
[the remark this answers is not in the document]
jeanregisser: oh thanks, I found a way to make it work: https://github.com/valora-inc/wallet/pull/4002

Shown on a single test, but I think all test should be able to work similarly.

satish-ravi (approved):

babel.config.js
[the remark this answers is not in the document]
jh2oman: Awesome! Thanks @jeanregisser , I'll try to get it working on this PR.

satish-ravi (approved):
🎉

PR #3968, opened 2023-07-19
fix(wc): set namespace in wallet instead from dapp wc request

MuckT:
### Description

While working on the wc connect upgrade it became clear we are setting our allowed namespace based on the dapp request instead of from the wallet: [wc docs ](https://docs.walletconnect.com/2.0/ios/web3wallet/wallet-usage#-autonamespaces-builder-utility) on namespaces.

### Test plan

- [ ] Test all listed v2 dapps and make a list of non-compliant dapps.
- [ ] Test with working dapps on iOS and Android.

### Related issues

N/A

### Backwards compatibility

Yes, but some dapps will break until they update their required namespaces in their session requests.


kathaypacific:
Closing as it will be covered in the chain switch fix

PR #3970, opened 2023-07-20
chore(wc): remove v1 codepath

MuckT:
### Description

Removes most of the v1 codepath. ~~There are still some issues to resolve with the migration and deeplinks.~~ The redux state was migrated as follows:

```JavaScript
walletConnect: {
  v1: {
    'sessions': [],
    'pendingSessions': [],
    'pendingRequests': [],
  }
  v2: {
    'sessions': [],
    'pendingSessions': [],
    'pendingRequests': [],
  }
}
```

To

```JavaScript
walletConnect: {
  'sessions': [],
  'pendingSessions': [],
  'pendingRequests': []
}
```

### Test plan

- [ ] Test locally on iOS
- [ ] Test locally on Android

### Related issues

- Fixes RET-786

### Backwards compatibility

Yes


src/walletConnect/saga.ts
```diff
@@ -24,23 +22,21 @@ export function* isWalletConnectEnabled(uri: string) {
   return versionEnabled[version] ?? false
 }
 
-export function* initialiseWalletConnect(uri: string, origin: WalletConnectPairingOrigin) {
+export function* shouldInitialiseWalletConnect(uri: string, origin: WalletConnectPairingOrigin) {
```
jeanregisser, line 25: I feel the existing name is more appropriate.
It can fail and that's ok.
the `should` prefix seems to indicate it returns a boolean but doesn't do anything.

jeanregisser (commented):
Great! 👍 

What do you think of really separating removing v1 and moving v2 to the root in 2 different PRs?
Here I see some v2 things were moved but not all of them.

src/walletConnect/saga.ts
MuckT: I'll move this saga back down to the v2 file to have removing v1 and moving v2 to the root in different PRs. This should resolve the name conflict also.

src/walletConnect/saga.ts
```diff
@@ -24,23 +22,21 @@ export function* isWalletConnectEnabled(uri: string) {
...
-    case '1':
-      yield call(initialiseWalletConnectV1, uri, origin)
-      break
     case '2':
-      yield call(initialiseWalletConnectV2, uri, origin)
+      yield call(initialiseWalletConnect, uri, origin)
       break
+    case '1':
     default:
       throw new Error(`Unsupported WalletConnect version '${version}'`)
```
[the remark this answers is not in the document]
MuckT: I believe it will silently error currently.

MuckT:
@jeanregisser I ended up tackling the removal and migration all in this PR as it was only few changes away. I also felt a bit strange to be calling combine reducer on one reducer.

src/walletConnect/v2/saga.ts
jeanregisser, line 1: Let's merge this with the saga in the parent folder, no?

MuckT: Moved in b3ab27ae49511e4e75f488ec416f11a7a6789d48! There are still some instances of a `v2` suffix used for function names such as `initialiseWalletConnectV2`. This is because there is an another function in the file named `initialiseWalletConnect`. Should we rename these? Maybe `_initialiseWalletConnect` and `initialiseWalletConnect` respectively?

jeanregisser (approved):
🚀

PR #3977, opened 2023-07-21
feat: add confirmation request when claiming shortcut rewards

kathaypacific:
### Description

This PR adds confirmation bottom sheet to reward shortcuts screen. The confirmation bottom sheet displays the raw data for all the transactions that will be sent. On confirm, the transactions will be sent.
1. previously, we were fetching and executing the shortcut transactions with one action. now, we split them up so this PR adds an `executeShortcut` action to handle this.
2. we don't need to store the response of the `triggerShortcut` endpoint so i moved the logic for fetching the response into the screen component
3. on claim reward, mark the reward as "loading" - display the loader on the claim button of the reward with this
4. on confirm reward, mark the reward as "accepting" - display the loader on the bottom sheet with this
5. if the user does not proceed with confirming,  trigger the error action to make the reward claimable again.

A few things I thought about:
- why do we want to persist the claim state in redux? we should do this in case a user accepts the transaction and leaves and re-enters the screen. we want to keep displaying the correct loading state to prevent the user from trying to claim (probably unsuccessfully) again.

### Related issues

- Fixes RET-733

### Backwards compatibility

Y


src/components/DataFieldWithCopy.tsx
```diff
@@ -0,0 +1,77 @@
+import Clipboard from '@react-native-clipboard/clipboard'
```
kathaypacific, line 1: no interesting changes in this file, it was copied from ActionRequestPayload with the WC specific stuff removed

src/positions/slice.ts
```diff
@@ -3,7 +3,7 @@ import { REHYDRATE, RehydrateAction } from 'redux-persist'
 import { getRehydratePayload } from 'src/redux/persist-helper'
 import { Position, Shortcut } from './types'
 
-type Status = 'idle' | 'loading' | 'success' | 'error'
+type Status = 'idle' | 'loading' | 'success' | 'error' | 'accepting'
```
jeanregisser, line 6: Can we isolate this only for `triggeredShortcutsStatus`?

src/dapps/DappShortcutsRewards.tsx
```diff
@@ -1,34 +1,72 @@
...
+import {
+  denyExecuteShortcut,
+  executeShortcut,
+  RawShortcutTransaction,
+  triggerShortcut,
+  triggerShortcutFailure,
+} from 'src/positions/slice'
 import { ClaimablePosition } from 'src/positions/types'
-import Colors from 'src/styles/colors'
+import { default as colors, default as Colors } from 'src/styles/colors'
```
jeanregisser, line 31: Duplicated

src/dapps/DappShortcutsRewards.tsx
```diff
@@ -49,30 +87,91 @@ function DappShortcutsRewards() {
...
+  useEffect(() => {
+    if (
+      claimInProgress &&
+      (triggeredShortcutsStatus[claimInProgress.id] === 'success' ||
+        triggeredShortcutsStatus[claimInProgress.id] === 'error')
+    ) {
+      confirmBottomSheetRef.current?.close()
+      setClaimInProgress(null)
+    }
+  }, [triggeredShortcutsStatus])
```
jeanregisser, line 106:
```suggestion
  }, [claimInProgress, triggeredShortcutsStatus])
```

src/dapps/DappShortcutsRewards.tsx
```diff
@@ -49,30 +87,91 @@ function DappShortcutsRewards() {
...
+        dispatch(showError(ErrorMessages.SHORTCUT_CLAIM_REWARD_FAILED))
+      },
+    }
+  )
+
+  useEffect(() => {
+    if (triggerShortcutAsync.error && claimInProgress) {
+      dispatch(triggerShortcutFailure(claimInProgress.id))
+    }
+  }, [triggerShortcutAsync.error])
```
jeanregisser, line 141:
```suggestion
  }, [triggerShortcutAsync.error, claimInProgress])
```

src/dapps/DappShortcutsRewards.tsx
```diff
@@ -49,30 +87,91 @@ function DappShortcutsRewards() {
...
+
+  const handleConfirmClaimReward = (position: ClaimablePosition) => () => {
```
jeanregisser, line 143:
```suggestion
  const createConfirmClaimRewardHandler = (position: ClaimablePosition) => () => {
```

jeanregisser (commented):
Awesome!

I wonder if we could reduce some of the new complexity apparent in `DappShortcutsReward` by moving some of the new code in a new component using BottomSheet.

Also maybe keeping the fetch in a saga would limit the complexity too?
Now in the component we try to synchronize the fetch status with the overall trigger shortcut flow and it seems more error prone.

src/positions/saga.ts
```diff
@@ -214,23 +220,41 @@ export function* triggerShortcutSaga({ payload }: ReturnType<typeof triggerShort
...
 
-    Logger.debug(`${TAG}/triggerShortcutSaga`, 'Starting to claim reward(s)', data.transactions)
+    const triggeredShortcuts: TriggeredShortcuts = yield select(triggeredShortcutsStatusSelector)
+    // use JSON stringify / parse, otherwise the transaction fails with this
+    // error: 'Gas estimation failed: Could not decode transaction failure
+    // reason or Error: invalid argument 0: json: cannot unmarshal non-string
+    // into Go struct field TransactionArgs.chainId of type *hexutil.Big'
+    const shortcutTransactions = JSON.parse(
+      JSON.stringify(triggeredShortcuts[payload].transactions)
+    )
```
jeanregisser, line 252: Weird, now I wanna now what's happening :D

jeanregisser (approved):
🚀

src/positions/types.ts
```diff
@@ -60,11 +60,20 @@ export interface Shortcut {
   category?: 'claim'
 }
 
+export type ShortcutStatus =
+  | 'idle'
+  | 'loading'
+  | 'pendingAccept'
+  | 'accepting'
+  | 'acceptingInBackground'
```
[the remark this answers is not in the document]
kathaypacific: yes but just in the background, maybe there is a better way of doing it - if an approval takes a long time, we need to keep displaying the reward as loading on the claim button (on the screen itself) so we can keep it at the "accepting" state but we can't mark it as "idle" for example. if the approval will take several seconds and the user wants to initiate claiming another reward, they will end up with another reward in the "pendingAccept" state and it won't be clear which tx details we should display in the bottom sheet if we don't transition the first reward into some other state...

src/positions/saga.ts
kathaypacific: yes same!!! but in the interest of time i parked it for now :/ it's reproducible for me every time...

src/positions/types.ts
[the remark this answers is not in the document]
jeanregisser: Can't we use `accepting` for both?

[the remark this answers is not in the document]
kathaypacific: yes - maybe i overthought the problem. but if an approval can take 5 seconds, i assumed that in that time a user can dismiss the sheet and start claiming another reward. i used this "acceptingInBackground" to ensure that there can only be one tx in the "pendingAccept" or "accept" states, but i guess we could also achieve this by tracking the id of the reward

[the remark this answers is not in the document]
kathaypacific: oh i should add, some of the struggles comes from wanting to persist the bottom sheet during the "accepting" phase. now i'm questioning if that is actually a requirement :| should we just dismiss the bottom sheet after hitting the "allow" button, or persist the bottom sheet with the loading state until manually dismissed / the tx finishes?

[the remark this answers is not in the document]
jeanregisser: Good question, seems we could do both, if we dismiss right after tapping the allow button, it feels the UX would be better, but it's more likely the user will try too claim other rewards while the first claim is still happening.
Not sure if this could cause issues with nonces though.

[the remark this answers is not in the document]
kathaypacific: yeah, i recall that we intended the user to only have one in flight reward at a time. since the bottom sheet is dismissable though, it's not impossible for them to trigger another but having the bottom sheet persist might reduce the likelihood. having the persisting bottom sheet is also the pattern we use for wallet connect which is why my brain defaulted to that

PR #3982, opened 2023-07-24
fix: ensure drawer menu header padding is consistent

kathaypacific:
### Description

As the title. With the revoke phonenumber work, the top padding on the phone number was removed without testing this scenario.

### Test plan

Simulator Screenshot - iPhone 14 Pro - 2023-07-24 at 09 32 17

### Related issues

n/a

### Backwards compatibility

Y


jeanregisser (approved):
💅

PR #3987, opened 2023-07-24
refactor(verify): pull out a shared component so it can be reused with CAB

satish-ravi:
### Description

Pulls out parts of the PhoneVerificationCodeInputScreen so it can be reused with the phone verification for cloud backup. Kept the help header outside because it's part of the nav header and the CTA on the bottom sheet is different

### Test plan

Unit tests, manually. Left the unit tests for PhoneVerificationCodeInputScreen as is since they test the full verification flow with the hook.

### Related issues

- To reuse for act-759

### Backwards compatibility

Yes


src/verify/VerificationCodeInputWrapper.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+interface Props {
+  phoneNumber: string
+  verificationStatus: PhoneNumberVerificationStatus
+  setSmsCode: (code: string) => void
+  onResendSms?: () => void
+  onSuccess: () => void
+}
+
+function VerificationCodeInputWrapper({
```
kathaypacific, line 22:
```suggestion
function VerificationCodeInput({
```

src/verify/VerificationCodeInputWrapper.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+        verificationStatus={verificationStatus}
+        onSuccess={mockOnSuccess}
+        onResendSms={showResend ? mockOnResendSms : undefined}
+        setSmsCode={mockSetSmsCode}
+      />
+    )
+
+  beforeEach(() => {
+    jest.clearAllMocks()
+    jest.useFakeTimers()
```
kathaypacific, line 32: nit: i think this is already the default defined in jest_setup?

kathaypacific (approved):
👨🏼‍🎨

src/verify/VerificationCodeInputWrapper.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+      <VerificationCodeInputWrapper
+        phoneNumber={mockPhoneNumber}
+        verificationStatus={verificationStatus}
+        onSuccess={mockOnSuccess}
+        onResendSms={showResend ? mockOnResendSms : undefined}
+        setSmsCode={mockSetSmsCode}
+      />
+    )
+
+  beforeEach(() => {
```
jophish, line 30: Can we add a test to ensure the paste button appears if the clipboard contents are set appropriately (and that it doesn't appear if the contents are not valid)? We should be able to check the testID from [here](https://github.com/valora-inc/wallet/blob/main/src/components/ClipboardAwarePasteButton.tsx#L40).

jophish (approved):
Nice abstraction -- one comment about a potential test to add.

jophish (approved):
:tada:

PR #3990, opened 2023-07-25
chore: add analytics for dapp shortcuts

kathaypacific:
### Description

As the title

### Test plan

n/a

### Related issues

- Fixes RET-733

### Backwards compatibility

Y


src/analytics/Properties.tsx
```diff
@@ -1311,6 +1312,27 @@ interface BuilderHooksProperties {
...
+  rewardId: string
+  appName: string
+}
+interface DappShortcutsProperties {
+  [DappShortcutsEvents.dapp_shortcuts_rewards_screen_open]: {
+    numRewards: number
+  }
+  [DappShortcutsEvents.dapp_shortcuts_reward_claim_start]: DappShortcutClaimRewardEvent & {
+    rewardTokens: string[]
+    rewardAmounts: string[]
```
jeanregisser, line 1325: I forgot if arrays are easy to analyze in our analytics pipeline.

src/analytics/Properties.tsx
```diff
@@ -1311,6 +1312,27 @@ interface BuilderHooksProperties {
...
+    shortcutId: string
```
jeanregisser, line 1326: Can we also add `claimableValueUsd`?

src/analytics/Properties.tsx
```diff
@@ -1311,6 +1312,27 @@ interface BuilderHooksProperties {
   [BuilderHooksEvents.hooks_disable_preview]: undefined
 }
 
+interface DappShortcutClaimRewardEvent {
+  rewardId: string
+  appName: string
+}
```
jeanregisser, line 1318: Can we add the following properties?
- `appId`
- `network`
- `shortcutId`

jeanregisser (approved):
Awesome! 🚀 

Already approve with some small suggestions.

src/analytics/Properties.tsx
kathaypacific: i'm also not sure, i asked the question in the #data channel. i changed these to csv's to be safe, although i do see us using string[] elsewhere

src/analytics/Properties.tsx
kathaypacific: yes, done!

PR #3991, opened 2023-07-25
chore(typed-redux-saga): demonstrating without macros

jh2oman:
To demonstrate test errors while using typed-redux-saga instead of typed-redux-saga/macros


jeanregisser:
Closing now that it's been done in https://github.com/valora-inc/wallet/pull/3961 🎉

PR #3997, opened 2023-07-26
feat(swap): Adding swap time metrics

dievazqu:
### Description

This PR adds time metrics to the swap analytic event

### Test Plan
Tested with unit tests

### Related issues

- Fixes RET-777


src/swap/SwapReviewScreen.tsx
```diff
@@ -64,6 +64,8 @@ export function SwapReviewScreen() {
   const walletAddress = useSelector(walletAddressSelector)
   const celoAddress = useSelector(celoAddressSelector)
   const feeCurrency = useFeeCurrency() ?? celoAddress
+  const [quoteRequestAt, setQuoteRequestAt] = useState<number | undefined>()
+  const [quoteResponseAt, setQuoteResponseAt] = useState<number | undefined>()
```
kathaypacific: setting new state triggers a rerender. since that's not really needed here, should we use a `useRef` instead of a `useState`?

src/swap/saga.ts
```diff
@@ -124,30 +140,47 @@ export function* swapSubmitSaga(action: PayloadAction<SwapInfo>) {
...
+    const beforeApproveTimestamp = Date.now()
     yield call(
       sendApproveTx,
       sellTokenAddress,
       amountToApprove,
       allowanceTarget,
       swapApproveContext
     )
+    const afterApproveTimestamp = Date.now()
+    sendApprovalElapsedTimeInMs = afterApproveTimestamp - beforeApproveTimestamp
```
kathaypacific, line 152: why are we interested in this? i thought the interesting bit would be the time between when the quote response was received and beforeApproveTimestamp?

jeanregisser (approved):
Great! 🚀

src/swap/saga.ts
```diff
@@ -124,30 +140,47 @@ export function* swapSubmitSaga(action: PayloadAction<SwapInfo>) {
...
     Logger.debug(TAG, `Starting to swap execute for address: ${walletAddress}`)
 
+    const beforeSwapExecutionTimestamp = Date.now()
     yield call(
       handleSendSwapTransaction,
       { ...action.payload.unvalidatedSwapTransaction },
       swapExecuteContext
     )
+    const afterSwapExecutionTimestamp = Date.now()
+    sendSwapElapsedTimeInMs = afterSwapExecutionTimestamp - beforeSwapExecutionTimestamp
```
kathaypacific, line 165: also not sure why we want this 😅 i thought only quoteToTransactionElapsedTimeInMs would be interesting...

kathaypacific (commented):
looks good but not sure we need all of this?

src/swap/saga.ts
dievazqu: I wasn't sure... I thought that adding these won't harm, and maybe they are useful to see some issues.
Like if the `quoteToTransactionElapsedTimeInMs`, it would be great to understand where the bottleneck was.

I'm fine if you want me to remove them. We don't really care how much it took until we send the approval, since that should work always.

src/swap/saga.test.ts
```diff
@@ -115,6 +119,12 @@ describe(swapSubmitSaga, () => {
       fromTokenBalance: '10000000000000000000',
       swapApproveTxId: 'a uuid',
       swapExecuteTxId: 'a uuid',
+      quoteRequestTimestamp: mockStartTimestamp,
+      quoteRequestElapsedTimeInMs: 1000,
+      totalElapsedTimeInMs: expect.any(Number),
+      sendApprovalElapsedTimeInMs: expect.any(Number),
+      sendSwapElapsedTimeInMs: expect.any(Number),
+      quoteToTransactionElapsedTimeInMs: expect.any(Number),
```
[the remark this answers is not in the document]
dievazqu: I can give it try... We would have to mock `Date.now()` with different values, so I though that might be complicated, but let me see what I can do.

src/analytics/Properties.tsx
```diff
@@ -1203,6 +1203,14 @@ type SwapQuoteEvent = SwapEvent & {
   provider: string
 }
 
+export interface SwapTimeMetrics {
```
silasbw, line 1206: Some nits:
- the names aren't completely descriptive.  since you explained them in the commit body, it might be worth adding those explanations as comments
- I'm struggling with the "elapsed" times.  Sometimes they mean "time to make a specific call within this larger swap operation" and sometimes they mean "elapsed since the start of the swap operation".  Can we be consistent?  I'd suggest the latter (and then encourage us to use a profiling/APM tool for the former).

PR #3998, opened 2023-07-26
fix(wc): better compat with unsupported chains

jeanregisser:
### Description

This improves connecting to dapps which either don't specify the Celo chain in the required namespaces or have it only in the optional namespaces.

Curve, Toucan and Cred Protocol fall into this.
The typical issue was that the dapp was reporting the wallet couldn't switch network.
Specifically with Curve, the Ethereum network was selected in the dropdown and it was impossible to switch it.

Previously we were allowing any chain. Now we only only EVM chains, even if not Celo.
We then clearly warn if the dapp tries to perform an action on an unsupported chain, and reject it.

I've left a note to also update the connect experience to show a warning when the requested chain/method/event aren't supported. This will be done in a follow up PR.

### Test plan

- Updated tests
- Manually tested connecting to all listed dapps.

Example of the new warning when the chain is not supported (note: this isn't a real world example, I manually changed the values to produce this):

 

### Related issues

- Fixes RET-778

### Backwards compatibility

Yes.

### DIFF
--- a/src/components/Warning.tsx
+++ b/src/components/Warning.tsx
@@ -1,6 +1,5 @@
 import React from 'react'
-import { useTranslation } from 'react-i18next'
-import { StyleSheet, Text, View } from 'react-native'
+import { StyleProp, StyleSheet, Text, View, ViewStyle } from 'react-native'
 import AttentionIcon from 'src/icons/Attention'
 import Colors from 'src/styles/colors'
 import fontStyles from 'src/styles/fonts'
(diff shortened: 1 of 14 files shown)


src/components/Warning.tsx
```diff
@@ -27,7 +25,7 @@ export function Warning({ title, description, ctaLabel, onPressCta }: Props) {
 
       {ctaLabel && onPressCta && (
         <Text style={styles.learnMoreText} onPress={onPressCta}>
-          {t('swapScreen.maxSwapAmountWarning.learnMore')}
```
kathaypacific, line 30: omg lol oops

kathaypacific (approved):
🧙🏻

PR #4007, opened 2023-07-27
fix: try to reload nft images that failed to load on data refresh

kathaypacific:
### Description

As the title. Currently on pull to refresh, only the data is refreshed but the images states are not.

Also, fix pet peeve of mine...the error image and nft image had different border radiuses....

### Test plan

(Notice the broken images at the start of the video vs the end)

https://github.com/valora-inc/wallet/assets/20150449/3a74d9b2-33de-456f-acc9-56821526cd14

### Related issues

- Fixes RET-793

### Backwards compatibility

Y

### DIFF
--- a/src/nfts/NftGallery.tsx
+++ b/src/nfts/NftGallery.tsx
@@ -86,7 +86,7 @@ export default function NftGallery() {
                   </View>
                 }
                 origin={NftOrigin.NftGallery}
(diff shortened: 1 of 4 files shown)


jeanregisser (approved):
Great 💯

silasbw (approved):

PR #4009, opened 2023-07-27
chore(deps): Upgrade typescript to 4.7.4

jh2oman:
### Description

Almost the entire PR is handling errors from `try...catch` statements which are now type `unknown` but we often assume them to be type `Error`. I wrote a small helper function, `ensureError` to handle these cases.

4.7.X is the minimum version we need for typechain https://github.com/dethcrypto/TypeChain/tree/master/packages/target-ethers-v6

### Test plan

Updated all unit tests


test/RootStateSchema.json
satish-ravi, line 1: Are the changes here expected? Looks like we're losing a bunch of details from the schema

src/utils/Logger.ts
```diff
@@ -53,10 +54,11 @@ class Logger {
   error = (
     tag: string,
     message: string,
-    error?: Error,
+    unknownError?: unknown,
```
satish-ravi, line 57: Are there still places where Logger is called without ensureError?

jh2oman: Yes, over 100

test/RootStateSchema.json
jh2oman: These are not expected. I will investigate more to try and figure out why and if it matters

src/utils/ensureError.test.ts
```diff
@@ -0,0 +1,23 @@
+import { ensureError } from 'src/utils/ensureError'
+
+describe('ensureError', () => {
```
jophish, line 3: Can we add a test for when we for some reason can't stringify the value?

jophish (commented):
Just a general question -- is this a common/normal pattern to use? Was it ever the case that we were throwing a non-`Error` value in the first place? I guess I'm unclear as to why we have "lost" this type guarantee. Is there another way to access thrown values as `Error` objects without using a psuedo-type-guard like this every time?

jh2oman:
@jophish This is all necessary because previously catch(error) was type `any` and now it defaults to `unknown` as of typescript 4.4. more on that [here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html#defaulting-to-the-unknown-type-in-catch-variables---useunknownincatchvariables)

 I was inspired by [this article ](https://medium.com/with-orus/the-5-commandments-of-clean-error-handling-in-typescript-93a9cbdf1af5#:~:text=More%20importantly%3A%20before%20using%20a,block%20unknown%20instead%20of%20any%20.)for the ensureError solution. We would at minimum have to do something like `if(err instanceof Error)` everytime we access `err.message` but then what do we do in the else case? I thought it cleaner to add a helper like this.

we do in [at least one place](https://github.com/valora-inc/wallet/blob/main/src/identity/privateHashing.ts#L228) throw non-Error types on purpose

test/RootStateSchema.json
```diff
@@ -312,9 +350,7 @@
                     "type": "string"
                 },
                 "category": {
-                    "enum": [
```
jh2oman, line 315: change made from this PR https://github.com/YousefED/typescript-json-schema/pull/535

test/RootStateSchema.json
```diff
@@ -2,6 +2,45 @@
     "$schema": "http://json-schema.org/draft-07/schema#",
     "additionalProperties": false,
     "definitions": {
+        "ActiveDapp": {
```
jh2oman, line 5: Changes all seem to be re-aranging or from [this enum to const change](https://github.com/YousefED/typescript-json-schema/pull/535)

test/RootStateSchema.json
jh2oman: I upgraded typescript-json-schema and now the changes look much more reasonable

src/account/reducer.ts
```diff
@@ -18,15 +18,15 @@ export interface State {
...
+  startOnboardingTime?: number
   dismissedGetVerified: boolean
   dismissedGoldEducation: boolean
   acceptedTerms: boolean
   hasMigratedToNewBip39: boolean
-  choseToRestoreAccount: boolean | undefined
-  profileUploaded: boolean | undefined
-  recoveringFromStoreWipe: boolean | undefined
-  accountToRecoverFromStoreWipe: string | undefined
+  choseToRestoreAccount?: boolean
```
jh2oman, line 26: new versions of ts-json-schema-generator do not recognize the `| undefined`

satish-ravi (approved):

src/account/reducer.ts
jh2oman: The rootstate should fail as long as the base state for the field is undefined

PR #4013, opened 2023-07-31
fix: add header for nft image component

kathaypacific:
### Description

As the title

### Test plan

n/a

### Related issues

- Fixes RET-793

### Backwards compatibility

Y


jeanregisser (approved):
👍 

Maybe change the PR title to `fix` instead of `chore`?

PR #4016, opened 2023-07-31
chore: remove dapp help icon

MuckT:
### Description

Removes the dapp help icon from the dapp explorer screen. The QR scanner placement is a bit further left than what was previously done; however, this matches the placement of the QR scanner on the homescreen.

| iOS Header Before | iOS Header After |
| :---: | :---: |
|  | |

### Test plan

- Tested locally on iOS & Android
- Unit tests updated / moved
- E2E tests removed for this functionality

### Related issues

- Fixes RET-789

### Backwards compatibility

Yes


src/dappsExplorer/DAppsExplorerScreenSearchFilter.test.tsx
```diff
@@ -655,5 +658,30 @@ describe(DAppsExplorerScreenSearchFilter, () => {
...
+      mocked(getExperimentParams).mockReturnValue({
+        showQrScanner: true,
+      })
+
+      const { queryByTestId } = render(
+        <Provider store={defaultStore}>
+          <DAppsExplorerScreenSearchFilter />
+        </Provider>
+      )
+      expect(queryByTestId('DAppsExplorerScreen/QRScanButton')).toBeTruthy()
```
kathaypacific, line 672: nit: we should use the `getBy*` methods rather than `queryBy*` for positive assertions because the error messages are more helpful if the test fails

kathaypacific (approved):
i've approved but could we address the below before merge?

i'm okay with this change but it's not mentioned in the ticket to replace the info icon with the scan icon - could you check in with Ellen before merging to make sure she's fine with this change? 

also, the horizontal padding on the icons is triggering all of my OCD 🫠 the issue is on the home screen we have horizontal gutters of 16 and on the dapps screen we have 24. on the home screen, the drawer nav icon does not align vertically with the rest of the screen content. now on the dapps screen, the scanner icon does not align vertically with the rest of the content 💔 tackling consistent screen gutters seems out of scope but can we wrap the icon in a view that adds an 8px right padding here, so that we don't make this screen worse? this can just be a specific change in the dapps screen

MuckT:
@kathaypacific the scan icon is already present when enabled via Statsig, if disabled, no icon is displayed in the top right.
I agree the gutters on the homescreen and dapps page should match and that it's out of scope to fix it for this ticket.

Unfortunately adding a padding to just the dapp screen causes the QR scanner to appear to jump left when navigating from home -> dapps. This movement is distracting so opted to leave the placement consistent across the screens.

PR #4018, opened 2023-08-01
feat(cab): Add KeylessBackupProgress screen

jophish:
### Description

For ACT-763. Adds a progress screen for the actual backup/restore portion, after both auth methods are successful. See Figma designs [here](https://www.figma.com/file/gGp6SVFWMbnSltkGzNzrpC/Wallet-Recovery?type=design&node-id=3604-31351&mode=dev).

This adds in a new piece of Redux state to track the progress of the backup, and dispatches a (currently dummy) new action to kick off a saga to begin the backup process. This was all designed with the restore flow in mind, so should be able to be re-used for restore, if we wish.

### Test plan

Unit and manual tested. See below images:

2023-07-31-215926_829x1733_scrot
2023-07-31-215907_824x1737_scrot
2023-07-31-215842_823x1734_scrot

### Related issues

- Fixes ACT-763.

### Backwards compatibility

N/A.


src/analytics/Properties.tsx
```diff
@@ -227,6 +227,9 @@ interface KeylessBackupEventsProperties {
   [KeylessBackupEvents.cab_issue_valora_keyshare_start]: CommonKeylessBackupProps
   [KeylessBackupEvents.cab_issue_valora_keyshare_success]: CommonKeylessBackupProps
   [KeylessBackupEvents.cab_issue_valora_keyshare_error]: CommonKeylessBackupProps
+  [KeylessBackupEvents.cab_progress_completed_continue]: undefined
+  [KeylessBackupEvents.cab_progress_failed_later]: undefined
+  [KeylessBackupEvents.cab_progress_failed_manual]: undefined
```
satish-ravi, line 232: these are specific to setup and won't be reused for restore right?

src/keylessBackup/slice.ts
```diff
@@ -1,11 +1,13 @@
 import { PayloadAction, createSlice } from '@reduxjs/toolkit'
+import { KeylessBackupStatus, KeylessBackupFlow } from 'src/keylessBackup/types'
 
 export interface State {
   google: {
     status: 'idle' | 'loading' | 'success' | 'error'
     idToken: string | null
   }
   valoraKeyshare: string | null
+  backupStatus: KeylessBackupStatus | null
```
satish-ravi, line 10: is this a state that we'll want to persist and identify cab status on the rest of the app? Or is it just a temporary state to identify the status in backup status screen? If it's the former, I guess we'd have to persist and currently the entire keylessBackup state is blacklisted from redux-persist

src/keylessBackup/KeylessBackupProgress.tsx
```diff
@@ -0,0 +1,173 @@
...
+    }
+  }
+
+  const onPressLater = () => {
+    ValoraAnalytics.track(KeylessBackupEvents.cab_progress_failed_later)
+    navigateHome()
+  }
+
+  switch (keylessBackupStatus) {
+    case undefined:
```
satish-ravi, line 66: I am guessing this is to handle nulls? Maybe it's cleaner to just use a "not started" enum as the default?

src/keylessBackup/KeylessBackupProgress.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+  return getMockStackScreenProps(Screens.KeylessBackupProgress, {
+    keylessBackupFlow: KeylessBackupFlow.Setup,
+  })
+}
+
+describe('KeylessBackupProgress', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+  it('navigates to home on success', async () => {
```
satish-ravi, line 34: can we add a test for the in progress state and assert the right icon is rendered?

src/keylessBackup/KeylessBackupProgress.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+  const onPressContinue = () => {
+    ValoraAnalytics.track(KeylessBackupEvents.cab_progress_completed_continue)
+    navigateHome()
+  }
+
+  const onPressManual = async () => {
+    ValoraAnalytics.track(KeylessBackupEvents.cab_progress_failed_manual)
+    try {
+      const pinIsCorrect = await ensurePincode()
```
satish-ravi, line 51: why do we have to ensure pincode here?

src/keylessBackup/slice.ts
cajubelt: we want to know if a user has backed up yet for the side menu and in settings

src/analytics/Properties.tsx
jophish: Yeah, that's correct.

src/keylessBackup/KeylessBackupProgress.tsx
jophish: Sure, that seems cleaner to me.

src/keylessBackup/KeylessBackupProgress.tsx
jophish: This follows the same pattern that the backup quiz/info section currently follows when accessed via the drawer navigator.

src/keylessBackup/KeylessBackupProgress.tsx
```diff
@@ -0,0 +1,173 @@
...
+      if (pinIsCorrect) {
+        navigate(Screens.BackupIntroduction)
+      }
+    } catch (error) {
+      Logger.error(TAG, 'PIN ensure error', error)
```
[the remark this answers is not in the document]
jophish: No I don't believe this gets invoked in the case of an incorrect PIN; this is the same pattern we use elsewhere with ensurePincode (I tested by entering an incorrect PIN by attempting to access the recovery phrase section on the drawer nav; no error was logged)

src/redux/migrations.ts
```diff
@@ -1196,4 +1196,11 @@ export const migrations = {
     },
   }),
   138: (state: any) => state,
+  139: (state: any) => ({
+    ...state,
+    keylessBackup: {
+      ...state.keylessBackup,
+      backupStatus: null,
+    },
```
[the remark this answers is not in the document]
jophish: We do want to persist this, as Charlie mentioned; I'm not sure what you mean by "it will be a migration on a different key of state" however?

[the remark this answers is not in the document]
satish-ravi: oops, I made a comment with an assumption. So currently the entire `keylessBackup` state is blacklisted [here](https://github.com/valora-inc/wallet/blob/8a8b68d0c873979487a4aca41a968df96d54940e/src/redux/store.ts#L29), so this migration would just add a not started state every time. We have a couple of options if we want to persist this
1. Leave the blacklist as is and add this to a different portion of the state -- the above comment makes this assumption, hence I mention a migration on a different key on the state
2. Blacklist only specific fields in the keylessBackup state. This requires a bit of a refactor to our persist config / reducer setup. https://www.npmjs.com/package/redux-persist#nested-persists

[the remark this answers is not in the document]
jophish: Updated; removed the migration.

src/keylessBackup/slice.ts
jophish: Satish and I discussed this offline. We decided that this state would be ephemeral (session-only), and we'll add a _new_ piece of state to an existing, _persisted_ object in Redux to denote whether or not the user has completed CAB. This makes life a lot easier, since we won't have to be selective about which pieces of state in the `keylessBackup` object get persisted, which is something we have not done before.

src/keylessBackup/KeylessBackupProgress.tsx
satish-ravi: ah ok, seems like the ensure pincode could be self contained over copying this in a bunch of places. Not directly in scope of this PR I guess

satish-ravi (approved):

satish-ravi (commented):
A on green build

PR #4024, opened 2023-08-01
feat(ethers): Add helper function for calculating send fee info

jh2oman:
### Description

Wrapper function for [chooseTxFeeDetails](https://github.com/valora-inc/wallet/blob/main/src/transactions/send.ts#L114) using non-contractKit specific parameters

### Test plan

Added unit tests

### Related issues

https://linear.app/valora/issue/ACT-785/choosetxfeedetails

### Backwards compatibility

N/A


src/ethers/saga.ts
```diff
@@ -0,0 +1,42 @@
...
+  encryptedComment?: string
+}) {
+  const celoTx = yield* call(
+    buildSendTx,
+    tokenAddress,
+    amount,
+    recipientAddress,
+    encryptedComment || ''
+  )
+  // It may never be worth it for us to port this fee calculation logic to ethers since it is so celo-specific
```
cajubelt, line 29: does this comment assume we stay on contractkit forever for celo? contractkit is not going to be supported anymore soon so idk if that makes sense to assume

cajubelt (approved):

src/ethers/saga.ts
jh2oman: I'll rephrase, probably a good idea to never say 'never' anyways in these kinds of comments

PR #4025, opened 2023-08-02
feat(nfts): add support for videos in gallery

MuckT:
### Description

Adds support for video NFTs in app.

Some points of note:
- Headers are slightly different on Android and iOS to accommodate full screen play.
- Videos will play with sound if volume is > 0 on the device.

#### iOS Video Player
https://github.com/valora-inc/wallet/assets/26950305/5109a0f7-e5bd-4ed9-9d1b-912c76b8e9ea

#### Android Video Player
https://github.com/valora-inc/wallet/assets/26950305/b04d56f6-a6ab-4b8a-b650-533e8a57c2e9

### Test plan

- Tested locally on iOS and Android
- Unit tests updated

### Related issues

- Fixes RET-794

### Backwards compatibility

Yes


src/nfts/NftVideo.tsx
```diff
@@ -0,0 +1,131 @@
...
+          source={{
+            uri: videoUrl,
+            headers: {
+              origin: networkConfig.nftsValoraAppUrl,
+            },
+          }}
+          key={`${nft.contractAddress}-${nft.tokenId}-${reloadAttempt}`}
+          testID={testID}
+          style={{
+            marginTop: 42, // Otherwise the video controls are hidden behind the floating header :(
```
MuckT, line 116: This is a bit strange, but otherwise the video fullscreen control is overlapped by the floating header.

src/nfts/NftMedia.tsx
```diff
@@ -12,23 +14,30 @@ import variables from 'src/styles/variables'
...
   shouldAutoScaleHeight?: boolean
   ErrorComponent: React.ReactNode
 }
 
-function ImagePlaceholder({ height = 40, width, borderRadius = 0, testID }: ImagePlaceHolderProps) {
+function Placeholder({
+  testID,
+  width,
+  mediaType,
+  height = mediaType === 'video' ? DEFAULT_HEIGHT : 40,
```
jeanregisser, line 38: Why such a big difference between video and images for the height here?

src/nfts/NftMedia.tsx
```diff
@@ -58,64 +67,98 @@ export default function NftImage({
...
+            source={{
+              uri: videoUrl,
+              headers: {
+                origin: networkConfig.nftsValoraAppUrl,
+              },
+            }}
+            key={`${nft.contractAddress}-${nft.tokenId}-${reloadAttempt}`}
+            testID={testID}
+            style={{
+              height: DEFAULT_HEIGHT,
```
jeanregisser, line 146: No way to get the actual height of the video? Like we're doing with images?

src/nfts/NftsInfoCarousel.test.tsx
```diff
@@ -9,13 +9,14 @@ import { createMockStore, getMockStackScreenProps } from 'test/utils'
...
 jest.mock('src/utils/Logger')
+jest.mock('react-native-video', () => 'ReactNativeVideo')
 
 describe('NftsInfoCarousel', () => {
   beforeEach(() => {
     jest.clearAllMocks()
   })
 
-  it('renders correctly with one Nft', () => {
+  it('renders correctly with one Video Nft', () => {
```
jeanregisser, line 19: Why are we updating this test just for video?

src/nfts/NftsInfoCarousel.tsx
```diff
@@ -123,19 +124,35 @@ export default function NftsInfoCarousel({ route }: Props) {
...
+                <ImageErrorIcon />
+                <Text style={styles.errorImageText}>{t('nftInfoCarousel.nftImageLoadError')}</Text>
+              </View>
+            }
+            testID="NftsInfoCarousel/MainImage"
+            origin={NftOrigin.NftsInfoCarouselMain}
+            shouldAutoScaleHeight
+            mediaType={'image'}
+          />
+        )}
```
jeanregisser, line 155: nit: maybe combine the 2 since they share most of the props?

jeanregisser (commented):
Awesome! 🚀

src/nfts/NftsInfoCarousel.test.tsx
MuckT: The first test nft is now a video, we could add another test for a single image NFT, but the behavior in the carousel is very similar regardless of image or video.

src/nfts/NftMedia.tsx
MuckT: The video only is present in the main image section of the carousel, but the images are used as thumbnails several places and use a smaller default height of 40 in those places.

jeanregisser: I see, maybe it's fine, but my reaction was it could be surprising as a developer to get a different height when not specified and just the media type changes.

src/nfts/NftsInfoCarousel.test.tsx
jeanregisser: if the fact that it's a video doesn't really matter, maybe we shouldn't change the test names, no?

src/nfts/NftMedia.tsx
dievazqu: I think we should use the default height here. And force all the clients of this component to specify the height of 40 if they need to.

src/nfts/NftMedia.tsx
```diff
@@ -58,64 +67,98 @@ export default function NftImage({
...
   useEffect(() => {
-    // if the image failed to load before, try again when the user pulls to refresh
     if (status === 'error' && fetchingNfts) {
       setStatus('loading')
       setReloadAttempt((prev) => prev + 1)
     }
   }, [status, fetchingNfts])
 
-  function sendImageLoadEvent(error?: string) {
+  function sendLoadEvent(error?: string) {
```
dievazqu, line 99: nit: Maybe it's cleaner to split this function into a `sendSuccessfulEvent` and a `sendErrorEvent`.

src/nfts/NftMedia.tsx
MuckT: Makes sense. I checked an in most places we specify 40 where needed so I'll set it to `DEFAULT_HEIGHT` here.

src/nfts/NftMedia.tsx
MuckT: Fixed in 702a2aab1c054fcb4965d9381282a8dd77171926!

dievazqu (approved):
LGTM!

kathaypacific (commented):
oh no i'm sorry, i had written these comments and never submitted the review :( the second one is not too important but would prefer to use a stable version of this video dependency unless there's good reason to go with the alpha?

PR #4027, opened 2023-08-02
feat(notif): Adding notification bell

dievazqu:
### Description

Adding a notification bell in the top bar in the home screen, tapping on that will trigger an analytic event.
In future tasks we will navigate to the notification center when tapping on it.

### Test plan

Tested locally in the emulator and with unit tests.

### Related issues

- Fixes RET-804

### Backwards compatibility

Yes

### Screenshot
Simulator Screen Shot - iPhone 14 - 2023-08-02 at 15 04 14


src/components/NotificationBell.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export default function NotificationBell({ style, size, testID }: Props) {
+  const notifications = useNotifications()
+
+  const newNotifications = notifications.length > 0
+  const notificationMark = newNotifications ? colors.greenUI : undefined
+
+  const onPress = () => {
+    ValoraAnalytics.track(HomeEvents.notification_center_pressed, {
```
dievazqu, line 23: This is ok? Or should I create a `NotificationEvents`?

src/analytics/Events.tsx
```diff
@@ -42,6 +42,7 @@ export enum HomeEvents {
   view_token_balances = 'view_token_balances',
   view_nft_home_assets = 'view_nft_home_assets', // When "NFTs" is clicked in Home Assets Pages
   home_action_pressed = 'home_action_pressed',
+  notification_center_pressed = 'notification_center_pressed',
```
jeanregisser, line 45: Should we name it more precisely? Anticipating the case where there's several ways to reach the notification center and we may want to know the exact button that triggered it.
```suggestion
  notification_bell_pressed = 'notification_bell_pressed',
```

src/components/NotificationBell.tsx
jeanregisser: It's fine I think. But until we need this component in more than 1 place, can we keep it in `src/home`?
We usually move components to `src/components` when they are more generally useful.
But maybe that's the case for this one?

jeanregisser (approved):
Awesome! 🚀 

Maybe in the future, we could animate the dot on the bell? 💅 
Or even the bell swinging? 🧑‍🎨 🔔

src/analytics/Properties.tsx
```diff
@@ -172,6 +172,7 @@ interface HomeEventsProperties {
   [HomeEvents.view_token_balances]: { totalBalance?: string }
   [HomeEvents.view_nft_home_assets]: undefined
   [HomeEvents.home_action_pressed]: { action: HomeActionName }
+  [HomeEvents.notification_center_pressed]: { newNotifications: boolean }
```
kathaypacific, line 175:
```suggestion
  [HomeEvents.notification_center_pressed]: { hasNewNotifications: boolean }
```

src/components/NotificationBell.tsx
kathaypacific: agree - if a component is in the components folder then it should be generic (having analytics + navigation inside the onPress in this component makes it not generic / easy to reuse)

src/components/NotificationBell.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { HomeEvents } from 'src/analytics/Events'
+import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
+import NotificationBell from 'src/components/NotificationBell'
+
+const mockUseNotifications = jest.fn()
+
+jest.mock('src/analytics/ValoraAnalytics')
+
+jest.mock('src/home/NotificationBox', () => ({
+  useNotifications: () => mockUseNotifications(),
```
kathaypacific, line 12: it's a tricky balance game for how much mocking is too much mocking...i personally feel like mocking hooks is mocking too much when testing UI components. It makes certain assumptions about how the internals of the UI component and often the tests become less valuable because we are testing how the component uses the hook, rather than how the component is responding to data. if i am making a refactor in how the plumbing between redux and display state works (e.g. i delete `useNotification`), i expect the unit tests to help me catch bugs and only fail when the outcome changes unexpectedly. but here, the unit tests will fail because of the way it is mocked. as you can tell i have some strong feelings on this topic 😅 i would suggest refactoring this test to have a mockedProvider with notifications in redux, and making assertions based on that instead to make these tests more useful

dievazqu: I understand that unit tests might seem less valuable that functional tests, but they are actually more scalable? Changing the logic of one component / function shouldn't imply changing more than just 1 or 2 unit tests.

dievazqu: Also, I'm complaining here, because the `useNotification` hook implementation uses a lot of different redux states, and another hook, that uses a bunch more of states... All stuff that doesn't matter to this component. Mocking the redux state correctly sounds a lot of extra work that I think it doesn't belong to this component.

If mocking the redux state was just 1 or 2 fields, I might have allowed this change.

kathaypacific: i guess it depends how we think of a "unit"... i don't agree that if we break `useNotification`, that this test should still pass. at the same time, if i change the interface of `useNotification` without changing the outcome (e.g. it now accepts some arguments) ideally i'd like not to have to update these tests?

i'd also argue that it is a dangerous pattern to say it's okay to mock out the internals of a component because it is simple, because we're now thinking about the implementation details and it may not stay simple for long 😅 if we introduce more hooks into this component, then we will need to mock more hooks in these tests or write the tests in a different way. i think that's also not scalable in some sense.

to be clear i'm not proposing to add more tests for this component - i agree that if we want to exercise the useNotifications hook then we should do that separately. i'm just proposing that instead of mocking `mockUseNotifications.mockReturnValue(...)` in these tests, that we wrap the component in `<Provider />` with the state that we expect to generate the UI outcome (e.g. `notifications` in the mock store either has elements or it doesn't). do you find this approach more harmful than mocking the hook directly?

dievazqu: >  if i change the interface of useNotification without changing the outcome (e.g. it now accepts some arguments) ideally i'd like not to have to update these tests?

If you change an interface, I think it's ok to have to update the tests where that interface is used. It would be actually weird if you change an interface and nothing breaks.
That being said, I think we should include a `expect(useNotification).toHaveBeenCalledWith(params)` to test the interface even better if we are adding arguments for example.

> i'd also argue that it is a dangerous pattern to say it's okay to mock out the internals of a component because it is simple,

I don't see this as the internals of a component, since this function is shared with another component too. I see the hook as an external dependency of the component.

> if we introduce more hooks into this component, then we will need to mock more hooks in these tests or write the tests in a different way. i think that's also not scalable in some sense.

If we introduce more hooks we have to mock them (or mock the redux provider as you are suggesting, which requires around the same amount of effort), new code always means testing new flows, but with my approach, we don't have to test the same flow in multiple places... When we change the `useNotification` to support the new schema, I wouldn't want to change all the tests of the `NotificationBox`, `NotificationBell`, the upcoming `NotificationCenter`, and any other component that might want to know about notifications. But maybe I'm wrong... 😅 


I think I understand what you mean, we usually mock the redux provider directly, and we get to test the hook + the component together, which is more than an unit test in my opinion, but I see how that can be more valuable. Maybe I have to ignore the scalability in this case, since testing better is more important here?

> i think with the default store, there are some homecards already for the positive test case
If I only use this case, it means that if we change the `useNotification` hook, but the default store state is still valid we won't touch this test, which I thought that was what you wanted.

I think the different opinions come from whether we consider the hook part of the component or not. I think we shouldn't.
I think the hook should have its own test with all the expected test cases, and then the components that use it should mock the interface while testing.

But I will change this test to what you suggest.

dievazqu: @kathaypacific test updated! :)

kathaypacific: thanks for updating the test!

i think indeed these differences in opinion come from what we consider one unit to be. i don't think there is a silver bullet here - each approach comes with tradeoffs, which i feel we've debated extensively in this thread. i think consistency within a codebase takes top priority when there isn't an objectively "correct" solution / pattern, and i do think that mocking the Providers here is what we more commonly do. i'll leave a couple more opinions on why i prefer this approach, in case it convinces you 😅
- hooks are not pure functions - by mocking the return value, we miss all the stuff that happens within the react lifecycle. any changes in the hook that would cause consumers to have other hooks triggered (e.g. if `useNotification` rerenders, maybe it incorrectly triggers "useEffect" but by mocking this hook, we will miss the bug in "useEffect" because the mocked hook does not have the same lifecycles). so unless it is much much more complex, it is preferable to leave hooks alone and mock the surroundings that feed the hooks to have maximum confidence that everything works as expected.
- to add to this ^, any unit tests for complex screens (e.g. the swap screen) would be really complex if we tried to mock all the hooks because the hooks have lifecyles that are important for how the overall screen works. e.g. there is a hook to fetch quotes, and once the quotes come back, the screen should be refreshed. to coordinate this with manually mocking the hook is pretty hard work, and not a pattern that we use.
- for simple display components where we render some UI based on some state, any hooks involved in the process with the purpose of massaging the data to make it easier to render the UI can be thought of as an implementation detail. e.g. this NotificationBell component does not need to rely on useNotification, it can actually just do the logic inside the component. if there was another hook for "useSendBackgroundPushNotification" or something like that, with a different / side effect purpose, maybe it makes sense to mock that since it is quite separate :/ but this is a contrived example so it doesn't make sense either way 😅

src/home/NotificationBell.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+  })
+
+  it(`emits the analytics event on press when there are notifications`, () => {
+    const { queryByTestId, getByTestId } = render(
+      <Provider store={createMockStore({})}>
+        <NotificationBell testID={testId} />
+      </Provider>
+    )
+
+    expect(queryByTestId(testId)).toBeTruthy()
```
kathaypacific, line 49: nit: for positive assertions,[ we should use the `getBy*` methods ](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-query-variants-for-anything-except-checking-for-non-existence)

kathaypacific (approved):

PR #4031, opened 2023-08-03
chore(nfts): cleanup for gallery and assets banner

MuckT:
### Description

- Filters out NFTs without metadata from displaying in the gallery view
- Adjust banner NFT banner in assets:
   - Redirects to gallery if and adjusts text and icon if `SHOW_IN_APP_NFT_GALLERY` is enabled in Statsig.
   - Small fix for Android touchable ripple - previously was showing below banner.
 - Prevent ripple from extending outside of the border radius in gallery view.
   
| Android Assets Before | Android Assets After |
| :---: | :---: |
|  |  |

### Test plan

- Tested Locally on iOS & Android
- Unit tests updated for metadata filtering

### Related issues

N/A

### Backwards compatibility

Yes


src/tokens/TokenBalances.tsx
```diff
@@ -314,14 +317,17 @@ function TokenBalancesScreen({ navigation, route }: Props) {
...
+                </Text>
                 <View style={styles.nftBannerCtaContainer}>
                   <Text style={styles.nftBannerText}>{t('open')}</Text>
-                  <OpenLinkIcon color={Colors.greenUI} />
+                  {!shouldShowNftGallery && <OpenLinkIcon color={Colors.greenUI} />}
                 </View>
               </View>
             </Touchable>
           )}
+          <View style={{ height: Spacing.Thick24 }} />
```
jeanregisser, line 330: nit: avoid inline style?

src/nfts/NftGallery.tsx
```diff
@@ -61,39 +65,41 @@ export default function NftGallery() {
...
-                    )}
-                  </View>
-                }
-                origin={NftOrigin.NftGallery}
-                borderRadius={Spacing.Regular16}
-              />
-            </Touchable>
+            <View style={styles.touchableContainer}>
+              <Touchable
+                borderless={false}
```
jeanregisser, line 70: Looks like we should implement the ripple fix directly in `<Touchable />`. So it fixes this everywhere without us having to wrap each call site.
What do you think?

jeanregisser (approved):
Great! 🚀

PR #4033, opened 2023-08-03
fix: missing clevertap account and token in app

kathaypacific:
### Description

This PR loads secrets from secrets.json if it exists, and uses it to add the clevertap account id and token to the android apps at build time.

Example error I saw in my logs when running locally `08-03 09:16:40.527 16538 16538 I CleverTap: Account ID or Account token is missing from AndroidManifest.xml, unable to create default instance`. In step 2 of the [starter docs](https://developer.clevertap.com/docs/react-native-quick-start-guide#add-clevertap-credentials) it does tell us to add this, I'm not sure why we didn't or maybe the docs were updated after our initial implementation.

I'm not sure why this didn't matter before :/ Am checking this, suspect it's probably due to some changes in the way segment v2 / the new clevertap plugin works?

[Example of account creation events during invite rewards](https://us1.dashboard.clevertap.com/W6K-556-4K6Z/techno-detail.html?q=N4IgpgbiBcCMBMBWANOGBtAuqglh7IAzvqgI4C2A%2BgCYCGALmJQO47UDmY9lA9gA4xQ%2FGCADMIAL6oAZjHgAGeKPkA2eCtT05i5WoAcqAdCyGAHiRC0%2BRkyGqQLAJwtW%2BAOwuFzxgrREB1FQBpAFpERBUQgBYglQAtSSAA%3D%3D) - note the low number of Android events, when it should be > 4.1k (those events that did come through came from unaffected app versions). During the previous invite rewards iteration, we saw the [expected events distribution](https://us1.dashboard.clevertap.com/W6K-556-4K6Z/techno-detail.html?q=N4IgpgbiBcCMBMBWANOGBtAuqglh7IAzvqgA4kikAeFAhqedFqgCaQUBOdDAdhYTSYFaMEAHUAbAGkAtIkQSZAFikSAWiFQBHALYB9FrQAuYPQHccLAOZgjegPaNQj0QGYQAX1QAzGPAAM8K7%2BSvAAHKhGfoHBiIEeQAAA%3D%3D). The clevertap / segment upgrades happened between these 2 iterations, so the timing certainly seems suspicious.

### Test plan

- The error goes away 
- I can see my events in Clevertap
- I can build the app whether the secrets.json file exists or not

### Related issues

- Fixes RET-746

### Backwards compatibility

Y


jeanregisser (approved):
🚀

PR #4034, opened 2023-08-03
fix: clevertap ios push impression events

kathaypacific:
### Description

The issue is iOS impression events were not logged on Clevertap. We were missing some code in our implementation.

This PR:
- adds the code snippet from [the docs](https://developer.clevertap.com/docs/push-notifications-ios#push-impressions)
- adds some missing steps from [the setup docs](https://developer.clevertap.com/docs/react-native-push-notification#ios)

### Test plan

Successful impressions logged [here](https://us1.dashboard.clevertap.com/46W-4W9-4K6Z/campaigns/campaign/1691410759/report/stats/trend) and [here](https://us1.dashboard.clevertap.com/46W-4W9-4K6Z/campaigns/campaign/1691411108/report/stats/trend)

### Related issues

- Fixes RET-746

### Backwards compatibility

Y


src/firebase/firebase.ts
```diff
@@ -189,6 +189,8 @@ export function* initializeCloudMessaging(app: ReactNativeFirebase.Module, addre
   const isEmulator = yield* call([DeviceInfo, 'isEmulator'])
   // Emulators can't handle fcm tokens and calling getToken on them will throw an error
   if (!isEmulator) {
+    yield* call([CleverTap, 'registerForPush'])
+    yield* call([app.messaging(), 'registerDeviceForRemoteMessages'])
```
kathaypacific, line 193: i'd like to put this line back @jh2oman if that's okay with you - i'll test this in the nightly build tomorrow to see if it causes issues, but without this line i could not get the push notifications working locally. i looked into whether it could be a signing issue (the only difference i can think of) but i cannot see any reason why it would be :/

MuckT: There were some steps for testing pushes in https://github.com/valora-inc/wallet/pull/2012, that may or may not be useful.

MuckT (approved):
I am not 100% sure about the code in `ios/celo/AppDelegate.mm` as it differs a bit from the the [Clevertap iOS Push notifcation guide](https://developer.clevertap.com/docs/push-notifications-ios), but I am fine testing on the nightly build.

src/firebase/firebase.ts
jh2oman: We can see how the (nightly) build goes. I'm fine with merging. This line was causing issues when upgrading segment versions before so let's make sure we are also seeing all analytics events on the (nightly) build correctly

ios/NotificationService/Info.plist
```diff
@@ -2,6 +2,10 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>CleverTapAccountID</key>
+	<string>$(CLEVERTAP_ACCOUNT_ID)</string>
```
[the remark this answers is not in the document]
kathaypacific: we have the same account but the account id / token are different for each env (you can find them in the secrets.json file)

src/firebase/firebase.ts
kathaypacific: ty! i'll see how it goes on the nightly, and tweak if it does not work as expected

kathaypacific:
> I am not 100% sure about the code in `ios/celo/AppDelegate.mm` as it differs a bit from the the [Clevertap iOS Push notifcation guide](https://developer.clevertap.com/docs/push-notifications-ios), but I am fine testing on the nightly build.

the docs are not exactly what we want because we explicitly do not want to prompt the user for push permissions on app install (we want to delay this until after account creation), but i did consult the same docs when implementing this :)

PR #4036, opened 2023-08-04
fix: swappable tokens should have usd price

kathaypacific:
### Description

Currently we are showing all supported swap tokens in the swap flow, but the original intention was to exclude the tokens we don't have a usd price for. We should be referencing `swappableTokens` rather than `supportedTokens` in the swap screen.

### Test plan

https://github.com/valora-inc/wallet/assets/20150449/00faad28-c42b-4a9e-bc3d-aea1719cca23

### Related issues

n/a

### Backwards compatibility

Y

### DIFF
--- a/src/swap/SwapScreen.test.tsx
+++ b/src/swap/SwapScreen.test.tsx
@@ -104,9 +104,9 @@ const renderScreen = ({ celoBalance = '10', cUSDBalance = '20.456', showDrawerTo
           balance: celoBalance,
         },
         [mockTestTokenAddress]: {
+          // no usdPrice
           address: mockTestTokenAddress,
           symbol: 'TT',
-          priceFetchedAt: now,
           decimals: 18,
           imageUrl:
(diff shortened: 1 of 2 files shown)


dievazqu (approved):
LGTM!!

Now that we are showing the warning message using the estimated price impact, maybe it's ok to swap tokens even if we don't have a price for it.
However, I must say that, if we don't have a price for a token, probably it's a junk token.

Maybe we can allow users to swap this weird token in some "advanced" feature.

kathaypacific:
> LGTM!!
> 
> Now that we are showing the warning message using the estimated price impact, maybe it's ok to swap tokens even if we don't have a price for it. However, I must say that, if we don't have a price for a token, probably it's a junk token.
> 
> Maybe we can allow users to swap this weird token in some "advanced" feature.

Yes exactly - if we can't determine a price for a token, then it's probably not one that we should facilitate our users acquiring more of. Having said that, if our pricing service is down then swaps will also be down but we shouldn't optimise for that scneario 😅

PR #4040, opened 2023-08-04
feat(keyless-backup): exchange auth0 jwt for torus keyshare

cajubelt:
### Description

Takes the credential you get from auth0 when you sign in with google and exchanges it for a torus keyshare. I had this take place as soon as login with google finishes, rather than after phone verification, since Torus rejects jwt's that are over a minute old, and it could easily take users a minute to verify their phone numbers. This will also reduce wait times after users complete phone verification since the Torus network is kinda slow.

This will let us encrypt and decrypt the mnemonic for keyless backup and recovery.

Not released to users yet

### Test plan

manually tested two flows:
- successful login: logged torus keyshare to console

https://github.com/valora-inc/wallet/assets/7979215/75483aa6-9fb0-4f59-9fca-0fdc29bf44e1

- simulated failure to login: made sure failure screen appears at the end

https://github.com/valora-inc/wallet/assets/7979215/72929460-47bb-4334-8c5e-8a96b8ac33e0

I also verified that Android and iOS give us the same keys for the same email (!!)

### Related issues

- Fixes https://linear.app/valora/issue/ACT-823/exchange-auth0-oauth-token-for-torus-private-key

### Backwards compatibility

na


src/keylessBackup/KeylessBackupPhoneCodeInput.tsx
```diff
@@ -67,11 +64,6 @@ function KeylessBackupPhoneCodeInput({
         verificationStatus={verificationStatus}
         setSmsCode={setSmsCode}
         onSuccess={() => {
-          dispatch(
-            keylessBackupStarted({
```
cajubelt, line 71: this created a race condition where if the Torus network requests failed before this was dispatched, the failure status would get over-written.

it's also not where keyless backup really starts, so I moved it earlier in the flow

src/keylessBackup/saga.ts
```diff
@@ -1,14 +1,42 @@
...
+  payload: { idToken: jwt },
+}: ReturnType<typeof googleSignInCompleted>) {
+  // Note: this is done async while the user verifies their phone number.
+  try {
+    const torusPrivateKey = yield* call(getTorusPrivateKey, { verifier: 'valora-auth0', jwt })
+    yield* put(torusKeyshareIssued({ keyshare: torusPrivateKey }))
+  } catch (error) {
+    Logger.error(TAG, 'Error getting Torus private key from auth0 jwt', error)
+    yield* put(keylessBackupFailed()) // this just updates state for now. when the user reaches the
+    // KeylessBackupProgress screen (after phone verification), they will see the failure UI
```
cajubelt, line 27: I think this is what product/design wanted for what happens if a failure takes place with the Torus network, at least in our MVP. Maybe if this happens a lot we will think of some way to short circuit phone verification-- though if it happens a lot, we probably need to rethink the feature

src/keylessBackup/web3auth.ts
```diff
@@ -0,0 +1,57 @@
+import NodeDetailManager from '@toruslabs/fetch-node-details'
+import Torus from '@toruslabs/torus.js'
+import jwtDecode from 'jwt-decode'
+import Logger from 'src/utils/Logger'
+import { TORUS_NETWORK, TORUS_NETWORK_CONTRACT_ADDRESS, TORUS_SIGNER_BASE_URL } from 'src/config'
+
+const TAG = 'keylessBackup/torus'
+
+export async function getTorusPrivateKey({ verifier, jwt }: { verifier: string; jwt: string }) {
```
cajubelt, line 9: this function is alarmingly complex, and relies on versions of deps that are already out of date, with breaking changes having taken place since then. however, getting it to work took *significant* effort, and I want to unblock the rest of this feature. I suggest we use this for now and I can try to upgrade it before we release to users (so it's still ok if the private keys returned by the refactored version are inconsistent with this method, if that happens for some reason). Ideally we'd use web3auth's nodeJS client, but it didn't work last time I tried it.

cajubelt: I didn't unit test this for now because I'd rather just rewrite it than invest in mocks for behavior we don't really understand to begin with. But if anyone sees an opportunity for useful and meaningful unit test coverage I'm open to it.

satish-ravi: skipping unit test sounds good for now, especially if we are going to upgrade to a version with breaking changes. Maybe add a comment linking to a linear issue?

src/keylessBackup/KeylessBackupPhoneCodeInput.tsx
satish-ravi: I think this also requires a unit test update. There should be a test that checks navigation and if this action was dispatched

src/keylessBackup/WalletSecurityPrimer.test.tsx
```diff
@@ -36,5 +53,9 @@ describe('WalletSecurityPrimer', () => {
     )
     expect(navigate).toHaveBeenCalledTimes(1)
     expect(navigate).toHaveBeenCalledWith(Screens.SetUpKeylessBackup)
+    expect(mockDispatch).toHaveBeenCalledWith({
+      payload: { keylessBackupFlow: 'setup' },
+      type: 'keylessBackup/keylessBackupStarted',
+    })
```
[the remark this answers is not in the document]
cajubelt: thanks

src/keylessBackup/KeylessBackupPhoneCodeInput.tsx
cajubelt: thanks, I think I checked the wrong test file for this check. and the other test failures due to effectful imports hid this one

src/keylessBackup/web3auth.ts
cajubelt: done

jest_setup.ts
```diff
@@ -38,3 +38,8 @@ jest.mock('react-native/Libraries/Utilities/PixelRatio.js', () => ({
 
 // @ts-ignore
 global.__reanimatedWorkletInit = jest.fn()
+
+// avoid side effects from importing Torus dependencies while testing (doing so unmocks fetch, even for unrelated tests)
+// TODO may need to remove once we update to use latest Torus/Web3Auth API https://linear.app/valora/issue/ACT-876
+jest.mock('@toruslabs/torus.js', () => jest.fn())
+jest.mock('@toruslabs/fetch-node-details', () => jest.fn())
```
satish-ravi, line 45: Not a blocking comment, but I think all our global mocks are in the __mocks__ folder.

satish-ravi (approved):

jest_setup.ts
cajubelt: what about the other mocks in this file? they don't seem to be in the mocks folder

cajubelt: ..but I moved the mocks to __mocks__ anyway, seems like it might be a bit cleaner, and maybe there's some reason that isn't possible for some modules and that's why they're here in jset_setup, and I'm just not catching on to it..

PR #4041, opened 2023-08-05
fix(nfts): border radius for android ripple container

MuckT:
### Description

Moves much of the custom styles logic into the parent view from the touchable to fix the board for Android ripple effect. Also switches to use Flatlist's [ItemSeparatorComponent](https://reactnative.dev/docs/flatlist#itemseparatorcomponent) instead of margin bottom.

| Android Before | Android After |
| :---: | :---: |
|  |  |

#### Android Video
https://github.com/valora-inc/wallet/assets/26950305/40292736-90df-4f8d-91a5-0b0bf836a707

### Test plan

Tested locally on Android

### Related issues

Fixes: RET-795

### Backwards compatibility

Yes


kathaypacific (approved):
LGTM but is the title of the PR correct? it seems to be fixing the border radius for the last gallery image instead of creating android ripple right?

MuckT:
@kathaypacific adjusted to title to be a bit more specific. It's was actually the border radius and margins for all images, but last image being circular was just the easiest to show it on.

PR #4043, opened 2023-08-07
chore: remove @segment/sovran-react-native patch

satish-ravi:
### Description

Noticed a warning from patch-package that the patch file was on an older version, but turns out we can just remove this patch now that TS is upgraded

### Test plan

CI

### Related issues

N/A

### Backwards compatibility

N/A


cajubelt (approved):
🥳

PR #4045, opened 2023-08-08
feat(dapps-explorer): hide favorites section if empty

MuckT:
### Description

Hides the favorites section in the dapp explorer screen if empty. Previously we'd display a placeholder for no favorites and a no results section; however, this caused the search results and filter results to commonly display before the keyboard making search difficult. Additionally the Dapps title was moved into the header and the tagline removed.

| iOS Before | iOS After |
| :---: | :---: |
|  |  |
### Test plan

- Tested locally on iOS
- Unit tests updated

### Related issues

- Fixes RET-787

### Backwards compatibility

Yes


src/dappsExplorer/DAppsExplorerScreenSearchFilter.tsx
```diff
@@ -79,6 +82,27 @@ export function DAppsExplorerScreenSearchFilter() {
   const [favoriteResultsEmpty, setFavoriteResultsEmpty] = useState(false)
   const [allResultEmpty, setAllResultEmpty] = useState(false)
 
+  // Hide favorites when there are no favorites or no favorites matching the search term
+  useEffect(() => {
```
kathaypacific, line 86: could we simplify this logic?
```
const filteredFavorites = favoriteDappsWithCategoryNames.filter((dapp) =>
              dapp.categories.includes(selectedFilter)
            )
const searchResults = searchDappList(filteredFavorites, searchTerm)
setFavoriteResultsEmpty(searchDappList(filteredFavorites, searchTerm).length === 0)
```
slightly less performant but since favoriteDappsWithCategoryNames is not a very large list, the readability seems worth it?

if you're worried about performance, `favoriteResultsEmpty` could be a `useMemo` instead of a `useEffect`?

src/dappsExplorer/DAppsExplorerScreenSearchFilter.tsx
```diff
@@ -168,10 +199,6 @@ export function DAppsExplorerScreenSearchFilter() {
             }
             ListHeaderComponent={
               <>
-                <DescriptionView
```
kathaypacific, line 171: is it in the task to remove this?

kathaypacific (approved):

src/dappsExplorer/DAppsExplorerScreenSearchFilter.tsx
MuckT: Trying the code above I am getting the following:
`Error: Too many re-renders. React limits the number of renders to prevent an infinite loop.` 😕

src/dappsExplorer/DAppsExplorerScreenSearchFilter.tsx
MuckT: Checking with @nityarao if we want this, otherwise I am fine reverting this change.

kathaypacific: fwiw i think it should be the other way around 🙈 we should be asked for the changes and then make them, rather than make the changes and revert them if we don't need them....

PR #4050, opened 2023-08-08
feat(keyless-backup): Add crypto helper functions for encrypting passphrases

jh2oman:
### Description

Helper functions for implementing steps 5-7 of [this plan](https://www.notion.so/valora-inc/Draft-Cloud-Backup-Engineering-Proposal-7d5fadfb96cd44be972ee2c6a6dff583?d=8fdccb18c9c34a14a290606aab27978c#e7120d7bb3844f5c81d35cadcb931965) for Cloud Backup

### Test plan

Unit tests

### Related issues

https://linear.app/valora/issue/ACT-684/wallet-code-for-encryptingstoring-mnemonic


src/keylessBackup/encryption.ts
```diff
@@ -0,0 +1,91 @@
...
+ * Derives a 256-bit key using the HKDF method from two key shares then derives a secp256k1 pair from that.
+ *
+ * @param {string} keyshare1 - The first keyshare in utf8 format.
+ * @param {string} keyshare2 - The second keyshare in utf8 format.
+ * @returns {Buffer} A derived 256-bit key as a Buffer.
+ */
+export function getSecp256K1KeyPair(keyshare1: string, keyshare2: string): EC.KeyPair {
+  const derivedKey = deriveKeyFromKeyShares(keyshare1, keyshare2)
+
+  return ec.keyFromPrivate(derivedKey)
```
jh2oman, line 39: This always returns undefined in unit tests, so I believe this to be incorrect. I believe its because you can't just use any odd HKDF encryption key as a derivedKey for ec.keyFromPrivate, but I'm not sure how else I am supposed to get a secp256k1 key from the combined keyshare.

src/keylessBackup/encryption.ts
```diff
@@ -0,0 +1,91 @@
+import crypto from 'crypto'
+import { ec as EC } from 'elliptic'
```
nategraf, line 2: Heads up that the `elliptic` library has some _really_ nasty sharp edges. In particular it's `sign` method expects a hex-encoded string or array message _hash_ instead the message itself. What is really aweful though is that if you pass it a string that is not hex, it will simply drop the invalid hex characters, then truncate to 32-bytes and pretend it's a buffer. I would strongly recommend replacing it's  usage.

Instead I would recommend noble crypto's implementation of k256 (or ed25519 if you are not bound to k256) https://paulmillr.com/noble/

src/keylessBackup/encryption.ts
nategraf: If you do switch to noble secp256k1, you can use the `hashToPrivateKey: (hash: Hex)` [1] function. Looking at it's implementation, it accepts a hex string of at least 40 bytes, and produces a private key for you (Note that the "at least 40 bytes" is required to avoid bias in the resulting key, which would somewhat reduce it's strength).

I would recommend passing in 48 bytes simply because it's 32+16, which means the bias in the resulting key will be less that 2^-128, which is the gold standard for cryptography implementation.

[1] https://github.com/paulmillr/noble-secp256k1#utils

src/keylessBackup/encryption.ts
```diff
@@ -0,0 +1,91 @@
...
+ * Encrypts a passphrase using AES-256-GCM encryption with a key derived from two key shares.
+ *
+ * @param {string} keyshare1 - The first keyshare used to derive the encryption key.
+ * @param {string} keyshare2 - The second keyshare used to derive the encryption key.
+ * @param {string} passphrase - The passphrase to encrypt.
+ * @returns {string} The encrypted passphrase, formatted as `iv:encrypted:authTag`, all parts base64 encoded.
+ */
+export function encryptPassphrase(
+  keyshare1: string,
+  keyshare2: string,
```
nategraf, line 52: How are these keyshares encoded? Would it be possible to use a `Buffer` type instead? The reason I mention it is that it would be good to add some basic sanity check to this function such that it throws if they keys passed in are obviously bad. This would be helpful if something goes wrong in implementation and an `undefined` or empty key share gets passed in. I would recommend checking that the inputs are defined and have length of at least 16 bytes (so if this is a hex-encoded string, 32 characters).

nategraf (commented):
Looks good to me. Main suggestion is to use `noble-secp256k1` instead of `elliptic`, which has some nasty sharp edges that bit cLabs in the past. With that, there is a function you can use to get a key.

I would also recommend reviewing usages of elliptic elsewhere in the codebase, and considering swapping them out as well.

https://github.com/paulmillr/noble-secp256k1

src/keylessBackup/encryption.ts
```diff
@@ -0,0 +1,105 @@
...
+ * Derives a 256-bit key using the HKDF method from two key shares then derives a secp256k1 pair from that.
+ *
+ * @param {string} keyshare1 - The first keyshare in utf8 format.
+ * @param {string} keyshare2 - The second keyshare in utf8 format.
+ * @returns {Buffer} A derived 256-bit key as a Buffer.
+ */
+export function getSecp256K1KeyPair(keyshare1: Buffer, keyshare2: Buffer): Uint8Array {
+  // Creating two keys and concatonating them to form a 48 byte key since 40 is the minimum for hashToPrivateKey
+  const derivedKey1 = deriveKeyFromKeyShares(keyshare1, keyshare2)
+  const derrivedKey2 = deriveKeyFromKeyShares(keyshare1, keyshare2, 16)
```
nategraf, line 50: HDKF can generate longer outputs. Calling with this a length of 48 should work. Also, this will not give a psuedorandom 48-bytes string, the last 16 bytes will be equal to the first 16 bytes.
```suggestion
  const derivedKey1 = deriveKeyFromKeyShares(keyshare1, keyshare2, 48)
```

src/keylessBackup/encryption.ts
```diff
@@ -0,0 +1,105 @@
...
+): Buffer {
+  // sanity check
+  if (keyshare1.length < 16 || keyshare2.length < 16) {
+    throw new Error('Key shares must be at least 16 bytes long')
+  }
+
+  // Combining the keyshares
+  const combinedKeyShares = Buffer.concat([
+    keyshare1,
+    hkdf(keyshare2, bytes, {
```
nategraf, line 25: I would recommend fixing this step to a length of 32. This gives the maximum available security available with SHA-256, and prevents a bit of wasted work if the requested output is longer than 32 bytes.
```suggestion
    hkdf(keyshare2, 32, {
```

src/keylessBackup/encryption.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  let passphrase: string
+
+  beforeEach(() => {
+    keyshare1 = crypto.randomBytes(16)
+    keyshare2 = crypto.randomBytes(16)
+    passphrase = 'testPassphrase'
+  })
+
+  describe('deriveKeyFromKeyShares', () => {
+    it('should derive a buffer of length 32', () => {
```
nategraf, line 21: A test here for longer output lengths would be good

src/keylessBackup/encryption.test.ts
```diff
@@ -0,0 +1,66 @@
...
+      expect(buffer.length).toBe(32)
+    })
+
+    it('should produce different outputs for different keyshares', () => {
+      const buffer1 = deriveKeyFromKeyShares(keyshare1, keyshare2)
+      const buffer2 = deriveKeyFromKeyShares(keyshare2, keyshare1)
+      expect(buffer1.toString('hex')).not.toBe(buffer2.toString('hex'))
+    })
+    it('should throw if a keyshare is less than 16 bytes', () => {
+      expect(() => deriveKeyFromKeyShares(keyshare1, crypto.randomBytes(15))).toThrow()
```
cajubelt, line 33: same here with string literals

src/keylessBackup/encryption.test.ts
```diff
@@ -0,0 +1,66 @@
...
+} from './encryption'
+
+describe("Encryption utilities using Node's crypto and futoin-hkdf", () => {
+  let keyshare1: Buffer
+  let keyshare2: Buffer
+  let passphrase: string
+
+  beforeEach(() => {
+    keyshare1 = crypto.randomBytes(16)
+    keyshare2 = crypto.randomBytes(16)
```
cajubelt, line 16: I suggest using fixed constants here instead. Wherever possible, we should not use test inputs that vary to avoid inconsistent results.

src/keylessBackup/encryption.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    passphrase = 'testPassphrase'
```
cajubelt, line 17: might be worth using a more realistic test passphrase, like an actual twelve word mnemonic

src/keylessBackup/encryption.ts
```diff
@@ -0,0 +1,105 @@
...
+}
+
+/**
+ * Derives a 256-bit key using the HKDF method from two key shares then derives a secp256k1 pair from that.
+ *
+ * @param {string} keyshare1 - The first keyshare in utf8 format.
+ * @param {string} keyshare2 - The second keyshare in utf8 format.
+ * @returns {Buffer} A derived 256-bit key as a Buffer.
+ */
+export function getSecp256K1KeyPair(keyshare1: Buffer, keyshare2: Buffer): Uint8Array {
```
cajubelt, line 47: keypair sounds like it would return two things. but it looks like it just returns a private key

src/keylessBackup/encryption.ts
```diff
@@ -0,0 +1,107 @@
...
+ */
+export function decryptPassphrase(
+  keyshare1: Buffer,
+  keyshare2: Buffer,
+  encryptedData: string
+): string {
+  const derivedKey = deriveKeyFromKeyShares(keyshare1, keyshare2)
+  const [nonceBase64, encrypted, authTagBase64] = encryptedData.split(':')
+  const nonce = Buffer.from(nonceBase64, 'base64')
+  const authTag = Buffer.from(authTagBase64, 'base64')
```
jophish, line 99: What will happen in the (exceedingly unlikely) case that the authentication tag signals that the ciphertext has been tampered with? Does the crypto library throw?

src/keylessBackup/encryption.ts
```diff
@@ -0,0 +1,107 @@
+import * as secp from '@noble/secp256k1'
+import crypto from 'crypto'
+import hkdf from 'futoin-hkdf'
+
+/**
+ * Derives a 256-bit key using the HKDF method from two key shares.
+ *
+ * @param {string} keyshare1 - The first keyshare in utf8 format.
+ * @param {string} keyshare2 - The second keyshare in utf8 format.
+ * @param {number} outputKeyBytes - The number of bytes to output. Defaults to 32.
```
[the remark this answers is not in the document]
cajubelt: If I'm understanding correctly, we need a longer output for `getSecp256K1KeyPair` to get a secp private key from it, but when we use the helper just for encryption, there's no added benefit to using longer than 32.

See https://github.com/valora-inc/wallet/pull/4050#discussion_r1292409045 and https://github.com/valora-inc/wallet/pull/4050#discussion_r1292410081

src/keylessBackup/encryption.ts
```diff
@@ -0,0 +1,107 @@
...
+    hkdf(keyshare2, 32, {
+      // per advice from @nategraf: [using 32 here] gives the maximum available security available with SHA-256, and prevents a bit of wasted work if the requested output is longer than 32 bytes
+      salt: 'some fixed salt',
+      info: 'valora.keylessBackup.deriveKeyFromKeyShares',
+      hash: 'SHA-256',
+    }),
+  ])
+
+  // Using futoin-hkdf to derive a 256-bit key
+  return hkdf(combinedKeyShares, outputKeyBytes, {
```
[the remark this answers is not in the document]
cajubelt: updated

src/keylessBackup/encryption.ts
```diff
@@ -0,0 +1,107 @@
...
+  if (keyshare1.length < 16 || keyshare2.length < 16) {
+    throw new Error('Key shares must be at least 16 bytes long')
+  }
+
+  // Combining the keyshares
+  const combinedKeyShares = Buffer.concat([
+    keyshare1,
+    hkdf(keyshare2, 32, {
+      // per advice from @nategraf: [using 32 here] gives the maximum available security available with SHA-256, and prevents a bit of wasted work if the requested output is longer than 32 bytes
+      salt: 'some fixed salt',
```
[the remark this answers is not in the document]
cajubelt: sure

src/keylessBackup/encryption.ts
```diff
@@ -0,0 +1,107 @@
...
+ * @param {number} outputKeyBytes - The number of bytes to output. Defaults to 32.
+ * @returns {Buffer} A derived 256-bit key as a Buffer.
+ */
+export function deriveKeyFromKeyShares(
+  keyshare1: Buffer,
+  keyshare2: Buffer,
+  outputKeyBytes: number = 32
+): Buffer {
+  // sanity check
+  if (keyshare1.length < 16 || keyshare2.length < 16) {
```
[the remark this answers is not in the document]
cajubelt: I think it's worth doing. If we somehow get a weak keyshare from torus or valora's keyshare-issuing service, I'd rather we throw than potentially encrypt the mnemonic with a weak encryption key

src/keylessBackup/encryption.ts
cajubelt: yep https://nodejs.org/api/crypto.html#deciphersetauthtagbuffer-encoding

> If no tag is provided, or if the cipher text has been tampered with, [decipher.final()](https://nodejs.org/api/crypto.html#decipherfinaloutputencoding) will throw, indicating that the cipher text should be discarded due to failed authentication

src/keylessBackup/encryption.test.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  describe('encryptPassphrase', () => {
+    it('should encrypt a passphrase using two keyshares', () => {
+      const encrypted = encryptPassphrase(keyshare1, keyshare2, passphrase)
+      expect(encrypted).toBeTruthy()
+      expect(typeof encrypted).toBe('string')
+    })
+  })
+
+  describe('decryptPassphrase', () => {
```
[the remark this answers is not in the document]
cajubelt: added

src/keylessBackup/encryption.ts
```diff
@@ -0,0 +1,108 @@
...
+  encrypted += cipher.final('base64')
+  const authTag = cipher.getAuthTag()
+
+  return `${nonce.toString('base64')}:${encrypted}:${authTag.toString('base64')}`
+}
+
+/**
+ * Decrypts an encrypted passphrase using AES-256-GCM decryption with a key derived from two key shares.
+ *
+ * @
```
jophish, line 86: nit: remove this `@` :smile:

jophish (approved):
Looks great! :)

src/keylessBackup/encryption.ts
cajubelt:
```suggestion
```

PR #4051, opened 2023-08-08
fix(wc): check active sessions and extend existing pairing if present

MuckT:
### Description

A fix for the error from wallet connect `{ context: 'core' }, 'Pairing already exists: <pairingTopic>'`. This error was commonly occurring and could be contributing to session mismatches that result in the the following errors describe by @jeanregisser in https://github.com/orgs/WalletConnect/discussions/3291#discussioncomment-6608198.
 
```Bash
{ context: 'client' }, 'No matching key. session: 9a05b6099c6cc662ed39dd43795adec83ad64dfc6cfd7779ae1bc3d6470bb5c6'
{ context: 'client' }, [Error: No matching key. session: 9a05b6099c6cc662ed39dd43795adec83ad64dfc6cfd7779ae1bc3d6470bb5c6]
```

### Test plan

1. The `{ context: 'core' }, 'Pairing already exists: <pairingTopic>'` error is no longer seen.
2. Dapp action requests continue to populate after transacting with several dapps over two days.

### Related issues

Fixes: RET-818

### Backwards compatibility

Yes


index.js
```diff
@@ -15,6 +15,7 @@ import App from 'src/app/App'
 import * as Sentry from '@sentry/react-native'
 import 'react-native-gesture-handler'
 import { Text, TextInput } from 'react-native'
+import '@walletconnect/react-native-compat'
```
MuckT, line 18: This is how several other wallets import '@walletconnect/react-native-compat`: [search](https://github.com/search?q=%40walletconnect%2Freact-native-compat+%22index.js%22&type=code). 

[WalletConnect RN CLI Wallet](https://github.com/WalletConnect/react-native-examples/tree/main/wallets/rn_cli_wallet)
```JavaScript
// index.js
import './polyfills';

// polyfills.ts
import '@walletconnect/react-native-compat';
```

```

Comment (silasbw) [reply index.js:18]: is this directly related to the fix?

File: src/walletConnect/saga.ts [hunk 2]
```diff
@@ -183,9 +182,15 @@ function* createWalletConnectChannel() {
     }
   }) as EventChannel<WalletConnectActions>
 }
-
+/**
+ * It appears Actions.INITIALISE_PAIRING is emitting on both initial connections and on payload requests
```
silasbw, line 186: I'm not sure I understand this comment.  Can you rephrase it?  Specifically what/who is emitting?  Also, if what/who is emitting erroneously or unexpectedly, should we fix that instead?

src/walletConnect/saga.ts
```diff
@@ -196,9 +201,21 @@ function* handleInitialisePairing({ uri, origin }: InitialisePairing) {
...
-    Logger.debug(TAG + '@handleInitialisePairing', 'pair end')
+    // Check the client for an existing session with a pairingTopic equal to the uri topic
+    const activeSessions = client.getActiveSessions()
+    const foundSession = Object.keys(activeSessions).find((key) => {
+      return activeSessions[key].pairingTopic === topic
+    })
+
+    if (foundSession) {
+      Logger.debug(TAG + '@handleInitialisePairing', 'extending session start')
+      yield* call([client, 'extendSession'], { topic: foundSession })
```
silasbw, line 212: Q: isn't `foundSession` an instance of `Session` (vs `string` which is what I'd expect for `topic`)?

src/walletConnect/saga.ts
MuckT: It looks like some old assumptions were made on our end [here](https://github.com/valora-inc/wallet/blob/main/src/walletConnect/walletConnect.ts#L50C3-L53) in `src/walleconnect/walletconnect.ts` about what what is a wallet connect action vs connection request. I agree a fix here is probably more appropriate.

index.js
MuckT: When the import was in the `src/walletconnect/saga`, I was seeing more issues with pairings.

MuckT:
Closing in favor of #4062

PR #4054, opened 2023-08-09
fix: ensure nft videos are muted by default

kathaypacific:
### Description

As the title - ensure that videos can also be unmuted. https://github.com/react-native-video/react-native-video/blob/master/API.md#muted

### Test plan

Tested on iOS, in silent and not silent mode, by hard coding some video NFT's to load.

https://github.com/valora-inc/wallet/assets/20150449/4821597c-ec43-4021-b8b6-5ab77fb7c74a

### Related issues

- Fixes RET-811

### Backwards compatibility

Y

### DIFF
--- a/src/nfts/NftMedia.tsx
+++ b/src/nfts/NftMedia.tsx
@@ -157,6 +157,7 @@ export default function NftMedia({
             minLoadRetryCount={3}
             repeat={true}
             resizeMode={shouldAutoScaleHeight ? 'contain' : 'cover'}
+            muted={true}
           />
           {/* This is a hack to get the loading skeleton to overlay the media player while loading, nesting within the player doesn't work */}
           <View style={{ marginTop: -DEFAULT_HEIGHT }}>


src/nfts/NftMedia.tsx
```diff
@@ -157,6 +157,7 @@ export default function NftMedia({
             minLoadRetryCount={3}
             repeat={true}
             resizeMode={shouldAutoScaleHeight ? 'contain' : 'cover'}
+            muted={true}
```
MuckT, line 160: On iOS this works, unfortunately on Android this appears to remove all volume control and the media is always muted.

MuckT: `muted={Platform.OS === 'ios'}` will fix this for iOS, but I am not sure what to do about Android.

kathaypacific: i just noticed it doesn't quite work as expected on iOS too, i made some notes in the task. i think the problem with Android is the default controls this lib provides doesn't include anything for volume :/ seems like we need to do something extra to get volume control. i'll close this PR for now, seems non urgent and better to address both platforms at once

PR #4060, opened 2023-08-10
chore: add estimated USD value to swap analytics events

silasbw:
### Test plan

Verify properties in analytics backends.

### DIFF
--- a/src/analytics/Properties.tsx
+++ b/src/analytics/Properties.tsx
@@ -1257,6 +1257,7 @@ interface SwapEventsProperties {
       fromTokenBalance: string
       swapExecuteTxId: string
       swapApproveTxId: string
+      estimatedUsdValue?: number
     }
   [SwapEvents.swap_execute_error]: SwapQuoteEvent &
     SwapTimeMetrics & {
@@ -1264,6 +1265,7 @@ interface SwapEventsProperties {
       fromTokenBalance: string
       swapExecuteTxId: string
       swapApproveTxId: string
+      estimatedUsdValue?: number      
     }
   [SwapEvents.swap_learn_more]: undefined
   [SwapEvents.swap_price_impact_warning_displayed]: SwapEvent & {
--- a/src/analytics/Properties.tsx
+++ b/src/analytics/Properties.tsx
@@ -1265,7 +1265,7 @@ interface SwapEventsProperties {
       fromTokenBalance: string
       swapExecuteTxId: string
       swapApproveTxId: string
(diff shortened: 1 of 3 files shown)


src/analytics/Properties.tsx
```diff
@@ -1257,13 +1257,15 @@ interface SwapEventsProperties {
       fromTokenBalance: string
       swapExecuteTxId: string
       swapApproveTxId: string
+      estimatedUsdValue?: number
```
dievazqu, line 1260: nit: Do you think it makes sense to include this field in the `SwapQuoteEvent`?

src/swap/saga.ts
```diff
@@ -162,6 +181,7 @@ export function* swapSubmitSaga(action: PayloadAction<SwapInfo>) {
     ValoraAnalytics.track(SwapEvents.swap_execute_success, {
       ...defaultSwapExecuteProps,
       ...timeMetrics,
+      estimatedUsdValue,
```
dievazqu, line 184: nit: I think you can include this value in the `defaultSwapExecuteProps`

src/swap/saga.ts
```diff
@@ -79,6 +94,10 @@ export function* swapSubmitSaga(action: PayloadAction<SwapInfo>) {
     ? fromToken.balance.shiftedBy(fromToken.decimals).toString()
     : ''
 
+  const estimatedUsdValue = fromToken
```
dievazqu, line 97: What do you think about having 2 fields, one for the estimated usd value for the selling token, and another for the buying token?

Something like:
```
const estimatedUsdValueSellingToken = calculateEstimatedUsdValue({ tokenInfo: fromToken, tokenAmount: sellAmount })
const estimatedUsdValueBuyingToken = calculateEstimatedUsdValue({ tokenInfo: toToken, tokenAmount: buyAmount })
```

In theory, `fromToken`, `toToken`, `sellAmount` and `buyAmount` will be defined. So I don't know if we need to check them.
If they aren't defined we will have errors in other parts as well.

dievazqu (approved):
LGTM!

src/analytics/Properties.tsx
silasbw: yeah, probably.  I was going to chat with you about clarifying some of the other fields (e.g., `.usdValue`) before adding this one to `SwapQuoteEvent` (and by implication `swap_review_submit`).

src/swap/saga.ts
silasbw: :+1:

PR #4061, opened 2023-08-10
refactor: add timeout to all network requests related to CICO

satish-ravi:
### Description

Replaced the networkTimeoutSeconds remote config param with a [Statsig dynamic config](https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/dynamic_configs/wallet_network_timeout) that can hold different timeouts. Currently includes a default and a cico timeout. We could extend this to be used with other network requests in the future (there are a bunch of places where we have a hardcoded timeout value, like supercharge, jumpstart, etc.).

Also refactored the fetchWithTimeout method to use abort controller instead of a Promise.race (which can keep the network request open for a long time even after the timeout)

the networkTimeoutSeconds remote config will be removed in a follow up

### Test plan

Unit tests, manually by hardcoding a small timeout and ensuring the request aborts

### Related issues

- Fixes ACT-523

### Backwards compatibility

Yes


src/utils/fetchWithTimeout.ts
```diff
@@ -2,25 +2,14 @@ import { FETCH_TIMEOUT_DURATION } from 'src/config'
 
 export const fetchWithTimeout = async (
   url: string,
-  body: any | null = null,
+  options: RequestInit | undefined = undefined,
```
satish-ravi, line 5: this is just a rename, this is still backwards compatible

src/utils/fetchWithTimeout.ts
```diff
@@ -2,25 +2,14 @@ import { FETCH_TIMEOUT_DURATION } from 'src/config'
 
 export const fetchWithTimeout = async (
   url: string,
-  body: any | null = null,
+  options: RequestInit | null = null,
```
satish-ravi, line 5: this is just a rename, this is still backwards compatible

src/in-house-liquidity/client.ts
```diff
@@ -24,7 +25,9 @@ export const getClient = async (): Promise<SiweApiClient> => {
         sessionDurationMs: SESSION_DURATION_MS,
         loginUrl: `${networkConfig.inHouseLiquidityURL}/auth/login`,
         clockUrl: `${networkConfig.inHouseLiquidityURL}/clock`,
-        timeout: networkTimeoutSecondsSelector(store.getState()) * 1000,
```
cajubelt, line 27: I like that previously, the selector had "seconds" in the name. It makes me uneasy that we now use the word "timeout" to mean millis in some places and seconds in others

src/statsig/constants.ts
```diff
@@ -89,4 +89,11 @@ export const DynamicConfigs = {
       blockedNouns: [] as string[],
     },
   },
+  [StatsigDynamicConfigs.WALLET_NETWORK_TIMEOUT]: {
+    configName: StatsigDynamicConfigs.WALLET_NETWORK_TIMEOUT,
+    defaultValues: {
+      default: 15,
+      cico: 30,
```
cajubelt, line 96: these are really long. I doubt most users would wait 30 seconds for something before leaving the app

cajubelt: how about 10 for both?

cajubelt: or 10 for default and 15 for cico?

src/utils/fetchWithTimeout.ts
```diff
@@ -2,25 +2,14 @@ import { FETCH_TIMEOUT_DURATION } from 'src/config'
...
-    // Response should always be defined because `reject` throws an error
-    // but just satifying the linter with this check
-    if (!response) {
-      throw Error(`Request timed out after ${duration}ms`)
-    }
-    return response
-  } catch (error) {
-    throw error
-  }
+  const controller = new AbortController()
```
cajubelt, line 8: does fetchWithTimeout have any test coverage? this seems like the most significant change in the PR

src/statsig/constants.ts
satish-ravi: I've seen some getQuotes requests take 10+ seconds on my simulator. I suspect this would be higher for real users especially in different regions especially when there's a cold start. Although I think 15 should work without cold starts for most regions.
With regards to default, I just went with what the current default was on the fetchWithTimeout method. Although right now this default isn't used anywhere yet, so we could set it to 10.

src/in-house-liquidity/client.ts
satish-ravi: I can update the dynamic config to include seconds

src/utils/fetchWithTimeout.ts
satish-ravi: nope, I'll add some

src/in-house-liquidity/client.ts
cajubelt: I think you may have forgotten to rename the dynamic config

cajubelt: I went ahead and did it

cajubelt (approved):

jophish (approved):

PR #4067, opened 2023-08-11
refactor: move list margins to parent component

bakoushin:
### Description

I want to move bottom margins from individual card components to their parent components.
That way, we can reuse cards inside different layouts, with different margins per layout.
The gap between cards in the list would be specified by the list, rather than in the card.

Affected screens:
* List of incoming payment requests
* List of outgoing payment requests
* List of escrowed payments
* Notification box (in case there is only 1 payment and it is rendered directly within the box)

The changes are only in code organizing; the UI must remain visually same as before. 

### Test plan

Screens integrity is tested manually on iOS emulator.

### Related issues

- Facilitates RET-803

### Backwards compatibility

Yes


MuckT (approved):
Awesome the UI looks the same as before for me on iOS and Android! I am curious if you considered a FlatList with an ItemSeparatorComponent, but these lists are seldom long enough so that might be a premature optimization.

Small nit on the branch naming for the next PR. We usually use `<dev-name>/<feature>` to help keep things organized.

bakoushin:
Yep, `FlatList` seems more appropriate for these lists because of their dynamic nature. If they will not completely replaced by the notification feed (as proposed in RET-803), we should consider refactoring. Another possible option is to adopt [`rowGap`](https://reactnative.dev/blog/2023/01/12/version-071#simplifying-layouts-with-flexbox-gap) after migrating to RN 0.71 👀 

In the notification feed I will use `FlatList` for sure.

✅ Branch naming noted, will definitely do! Thanks!

PR #4068, opened 2023-08-11
feat(touchable): handle android overflow

MuckT:
### Description

Many of the touchable components on Android do not respect their border radius. This causes rounded icons to have a rectangular ripples. By default this rounded border container is disabled to preserve existing styles.

| Home Action Before | Home Action After | Dapp Card Before | Dapp Card After | 
| :----: | :----: | :----: | :----: |
|  |  |  |  |

### Test plan

- Tested locally on Android for correct behavior.
- Tested locally on iOS for regressions.

### Related issues

- Fixes RET-810

### Backwards compatibility

Yes


kathaypacific (commented):
yay thanks for tackling this!! i had a few thoughts, which maybe expand the scope here but i think it's worth discussing now if we want to address this topic more fully.
1. react-native-platform-touchable is archived since 2020 - should we switch it for [Pressable](https://reactnative.dev/docs/pressable) or the [gesture lib touchables](https://docs.swmansion.com/react-native-gesture-handler/docs/api/components/touchables/)?
2. i'm not sure if it was your intention to fix only the dapps related touchables, but there's still many examples in the app that don't have the correct ripple (e.g. the rewards pill on the drawer menu). not everyone tests on android and it's easy to miss a detail like the ripple, the approach now relies on the eng knowing that there is a `borderRadius` prop available and knowing when to set it. i wonder if a better approach is to make the `borderRadius` prop required (when the `borderless` prop is false or undefined?) so that you are forced to think about it. our design system doesn't really have sharp corners so i think setting a borderRadius will be necessary most of the time. this makes the scope of this work a bit bigger though

kathaypacific:
> @kathaypacific - For the first point we could maybe make a new touchable component and depreciate the existing one so all new code would use a different touchable, this seems like a better path and is similar to way the bottom sheet upgrades were handled.
> 
> Regarding the second point I started down that route, but there were 50+ snapshots that needed updating and at first glance several newer components without snapshot tests that needed to be updated: I imagine quite are few more impacted I didn't see. As such I went with the smaller change instead to keep the scope small; if we find a touchable that has a ripple that doesn't respect the border we can fix ad-hoc.

makes sense. my 2c here is we should always try to find solutions that make it harder for us to create new instances of a bug. the PR as it is now doesn't achieve that IMO because the prop that is required to fix the bug is optional and is easy to miss during development and reviews. so along that line of thinking, making a new component and deprecating the old one seems sensible and lowest lift. although i also feel that updating the existing component might not be a huge lift either but i could be missing something (it has the added benefit of fixing the problem all at once, i can imagine that we never get round to fully deprecating an old component) - lmk if you prefer the latter option, i'd be happy to work on it too because i love this kind of refactors!!

MuckT:
@kathaypacific I took a look at deprecating the current touchable component and creating a new one, but decided against it as that seems like its own stream of work. I took a look at adding a `isLegacy` prop and if present using the touchable without the view; however, for rectangular components we don't need the view wrapper and the `isLegacy === (borderRadius === 0)` logic seems more confusing then simply looking at the `borderRadius`. Ultimately I decided to avoid scope creep I'd keep it simple and just add some additional docs.

src/components/Touchable.tsx
```diff
@@ -1,15 +1,27 @@
...
 export interface Props extends TouchableWithoutFeedbackProps {
   borderless?: boolean
   children: React.ReactNode // must only have one direct child. see https://github.com/react-native-community/react-native-platform-touchable#touchable
+  borderRadius?: number
 }
 
-export default function Touchable({ borderless, ...passThroughProps }: Props) {
+/**
+ * @param borderless - If true, the touchable has a borderless ripple effect.
+ * @param borderRadius - Adds a border radius to the touchable and a parent view to prevent Android ripple overflow.
```
kathaypacific, line 13:
```suggestion
 * @param borderRadius - should be added if the touchable component has rounded corners, to prevent the ripple effect from going outside the component on tap (Android)
```

kathaypacific (approved):
i personally wouldn't notice the JS docs here (since I've used this component before, i'm likely to type really fast and miss the prompt, or copy + paste from another component) - so i still feel like this solution doesn't hit the mark fully for preventing similar bugs in the future. fine for now, i'll probably take a pass at removing react-native-platform-touchable since we've talked about it.

PR #4069, opened 2023-08-15
feat: notifications screen

bakoushin:
### Description

Add a new screen for notifications:
* new notifications screen in the stack navigator
* bell icon on the top right corner opens the notifications screen
* notification list, priorities, and ordering are the same as in existing homecards
* if there are no notifications, an empty state is displayed as per the designs

https://github.com/valora-inc/wallet/assets/2737872/741b3fe8-8be6-4027-96ad-8a127b323fc9

### Test plan

* unit tests are mostly copy-paste from those for homecards because the cards themselves still untouched
* tested manually with mock data in iOS and Android emulators

### Related issues

- Fixes RET-799
- Flattening currently grouped notifications to be fixed in RET-803

### Backwards compatibility

N/A


locales/base/translation.json
```diff
@@ -1137,6 +1137,7 @@
   "sent": "Sent",
   "received": "Received",
   "viewAll": "View All",
+  "youAreAllCaughtUp": "You're all caught up!",
```
kathaypacific, line 1140:
```suggestion
  "noNotificationsPlaceholder": "You're all caught up!",
```

kathaypacific: it's easier for the translators to translate accurately if the keys are named semantically :)

src/navigator/Navigator.tsx
```diff
@@ -577,6 +578,16 @@ const nftScreens = (Navigator: typeof Stack) => (
   </>
 )
 
+const notificationScreens = (Navigator: typeof Stack) => (
+  <>
+    <Navigator.Screen
```
kathaypacific, line 583: what do you think about putting this in the `generalScreens` instead?

src/home/NotificationCenter.tsx
```diff
@@ -0,0 +1,118 @@
...
+    />
+  )
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+  },
+  title: {
+    ...fontStyles.h1,
```
kathaypacific, line 96: nit: maybe this is not needed?

src/home/NotificationCenter.test.tsx
```diff
@@ -0,0 +1,498 @@
...
+  it('renders empty state when there is no notifications at all', () => {
+    const store = createMockStore({ ...storeDataNotificationsDisabled })
+    const { queryByTestId, queryByText } = render(
+      <Provider store={store}>
+        <NotificationCenter {...getMockStackScreenProps(Screens.NotificationCenter)} />
+      </Provider>
+    )
+
+    expect(queryByTestId('NotificationCenter/EmptyState')).toBeTruthy()
+    expect(queryByText('youAreAllCaughtUp')).toBeTruthy()
```
kathaypacific, line 118: nit: use the `getBy*` methods when making positive assertions for better error messages https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-query-variants-for-anything-except-checking-for-non-existence

src/home/NotificationCenter.test.tsx
```diff
@@ -0,0 +1,498 @@
...
+      },
+    })
+    const { getByTestId } = render(
+      <Provider store={store}>
+        <NotificationCenter {...getMockStackScreenProps(Screens.NotificationCenter)} />
+      </Provider>
+    )
+
+    const titleElement = getByTestId('IncomingPaymentRequestNotification/FAKE_ID_1/Title')
+    expect(getElementText(titleElement)).toBe('incomingPaymentRequestNotificationTitle, {}')
```
kathaypacific, line 160: is `incomingPaymentRequestNotificationTitle` meant to have an empty argument here?

src/home/NotificationCenter.tsx
```diff
@@ -0,0 +1,118 @@
...
+  return (
+    <View testID={`NotificationView/${item.id}`} key={item.id} style={styles.listItem}>
+      {item.element}
+    </View>
+  )
+}
+
+const keyExtractor = (item: Notification) => item.id
+
+const renderItemSeparator = () => <View style={styles.itemSeparator} />
```
kathaypacific, line 28: i usually see functions like these inside the component - i don't have a problem with them out here, but could you remind me of the difference / benefits? it's nice that these functions don't get recreated every re-render of the Notifications component, and they don't need anything specifically inside the Notifications component so there's no _need_ for them to live inside Notifications. are there other considerations?

src/home/NotificationCenter.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const renderEmptyState = () => (
+    <View testID="NotificationCenter/EmptyState" style={styles.emptyStateContainer}>
+      <ThumbsUpIllustration />
+      <Text style={styles.emptyStateLabel}>{t('youAreAllCaughtUp')}</Text>
+    </View>
+  )
+
+  const contentContainerStyle = { paddingBottom: safeAreaInsets.bottom + Spacing.Regular16 }
+
+  const notifications = useNotifications()
```
kathaypacific, line 74: super nit: keep all the hooks together near the top of the component?

kathaypacific (approved):
looks great!! already approved with a few small comments / questions 🙌🏼

src/home/NotificationCenter.test.tsx
bakoushin: cool! thanks! 😄

src/home/NotificationCenter.tsx
bakoushin: no, i don't know any new aspects besides that the functions don't get recreated every re-render.
i've noticed that having at least FlatList's `keyExtractor` outside the component is quite frequent among the codebase, so I considered it a common practice 😅 , for instance:
[CeloNewsFeed.tsx](https://github.com/valora-inc/wallet/blob/0d4dc3c04834ef5893b15342f186cf3a1f08bb0b/src/exchange/CeloNewsFeed.tsx#L29-L39)
[SelectLocalCurrency.tsx](https://github.com/valora-inc/wallet/blob/0d4dc3c04834ef5893b15342f186cf3a1f08bb0b/src/localCurrency/SelectLocalCurrency.tsx#L17-L19)
[SelectCountry.tsx](https://github.com/valora-inc/wallet/blob/0d4dc3c04834ef5893b15342f186cf3a1f08bb0b/src/onboarding/registration/SelectCountry.tsx#L17)
[Language.tsx](https://github.com/valora-inc/wallet/blob/0d4dc3c04834ef5893b15342f186cf3a1f08bb0b/src/language/Language.tsx#L25-L27)

i don't believe that there is a substantial performance win, so if placing these functions within the component will contribute to code readability and portability, i have no objections. what do you think?

src/home/NotificationCenter.tsx
bakoushin: yeah, `h1` is almost overwritten, the only property not being replaced is `color` 😅 

i've replaced h1 spreading with an evident `color` property

src/home/NotificationCenter.test.tsx
bakoushin: good catch! I've updated the mock state to have an address->name mapping here and in the sibling test for NotificationBox

src/home/NotificationCenter.tsx
bakoushin: I think, for readability it's worth not to tear them apart. 
grouped all inside the component 📦

PR #4075, opened 2023-08-17
feat(notifications): flatten payment request notifications

bakoushin:
### Description

Currently, we have one "summary" notification card that groups multiple incoming payment requests. On tap of this card, a user can enter a screen that shows a list with each individual request. 

It is a similar flow for outgoing payment requests and sent escrow invites.

In this PR each payment request is displayed directly on the notification feed, so there is no extra screen transition required to interact with individual notifications.

Within their priority group, payment requests are sorted in reverse chronological order (the most recent on top).

flatten-notifications

#### Minor fix

For the sake of consistency, the incoming payment request card is updated to display either a phone number or the address of the requester if no name is associated with the address. It is the same behavior as in outgoing payment requests.

### Test plan

* Tested manually on mocked data on iOS and Android emulators.
* Unit tests updated to ensure correct ordering of requests and correct render of incoming payment card

### Related issues

- Fixes RET-803

### Backwards compatibility

N/A


MuckT:
@bakoushin testing locally after I send a request I am unable to cancel it from the notification center; however, on previous app versions I am able to cancel from the home cards.

kathaypacific (approved):
looks great from my side!! i didn't test though, i see Tom raised an issue which would be nice to resolve before merge :)

bakoushin:
@MuckT good catch! thank you! I've fixed that. would you mind verifying it's fixed?

PR #4084, opened 2023-08-18
feat(swaps): Using estimated price impact to show warning message

dievazqu:
### Description

Using estimated price impact to show warning message instead of the fiat value.

### Test plan

Unit tests

### Related issues

- Fixes RET-809

### Backwards compatibility

Yes


locales/base/translation.json
```diff
@@ -1682,6 +1682,10 @@
       "title": "Before you swap",
       "body": "Swapping these assets may result in an unfavorable exchange rate. If you lower the amount, you may get a better rate."
     },
+    "missingPriceImpactWarning": {
+      "title": "Before you swap",
+      "body": "We couldn't estimate the price impact of this swap. Proceed carefully."
```
dievazqu, line 1687: Maybe there's a better wording for this?
This message shouldn't be shown normally, but there are some weird cases where this could happen.

locales/base/translation.json
```diff
@@ -1682,6 +1682,10 @@
       "title": "Before you swap",
       "body": "Swapping these assets may result in an unfavorable exchange rate. If you lower the amount, you may get a better rate."
     },
+    "missingPriceImpactWarning": {
+      "title": "Before you swap",
+      "body": "The price impact of this swap couldn't be estimated. Proceed carefully."
```
dievazqu, line 1687: Maybe there's a better wording for this?
This message shouldn't be shown normally, but there are some weird cases where this could happen.

kathaypacific: could we check with Ellen or Nitya before this is merged?

kathaypacific (commented):
looking good, a few small comments from me :)

src/swap/SwapScreen.tsx
```diff
@@ -182,19 +183,26 @@ export function SwapScreenSection({ showDrawerTopNav }: { showDrawerTopNav: bool
...
+      const priceImpact = exchangeRate.estimatedPriceImpact
+        ? new BigNumber(exchangeRate.estimatedPriceImpact).dividedBy(100)
+        : undefined
+
+      const isPriceImpactMissing = priceImpact === undefined
+
+      const priceImpactExceedsThreshold =
+        !!priceImpact && priceImpact.gte(priceImpactWarningThreshold)
+
+      setShowMissingPriceImpactWarning(isPriceImpactMissing)
```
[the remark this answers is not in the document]
kathaypacific: actually now that i say this, we will have 3 scenarios for price warnings in the future (if we plan to remove the fiat pricing requirement):

1. high price impact
2. null price impact
3. null fiat prices for either token

i wonder if 2 and 3 can be grouped, because there's not much user action that can be taken except to proceed with caution. perhaps we can design the copy / translation string keys for 2 to be generic and include the future scenario 3?

locales/base/translation.json
dievazqu: Yes, I'll check with Nitya!

src/swap/SwapScreen.tsx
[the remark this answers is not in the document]
dievazqu: Do we want to warn the user in `3`? If token prices aren't needed to swap, should the user care about them?
Maybe we want to use that data (e.g. missing token prices) as a argument to say that token is weird, but in reality, the missing token price could be an error our side (for example, the curve tokens that we are missing).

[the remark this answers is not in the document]
kathaypacific: yeah good question...i think that we should design assuming that our services work as expected, and we do have some work to support curve prices lined up. i think we also eventually want to remove the requirement that only tokens that we have prices for should be swappable since this is not what most wallets / dex's do, so this warning would be necessary. we can discuss this at our weekly this week - i just wanted to call it out here because i think it's worth thinking about a generically worded warning here that can be reused. having no price impact doesn't mean much for the user and we just want to provide an action point (which is to tread carefully, but this can be applied to other scenarios)

src/swap/SwapScreen.tsx
```diff
@@ -114,6 +113,46 @@ export function SwapScreenSection({ showDrawerTopNav }: { showDrawerTopNav: bool
...
+      ValoraAnalytics.track(SwapEvents.swap_price_impact_warning_displayed, {
+        toToken: exchangeRate.toTokenAddress,
+        fromToken: exchangeRate.fromTokenAddress,
+        amount: parsedSwapAmount[updatedField].toString(),
+        amountType: updatedField === Field.FROM ? 'sellAmount' : 'buyAmount',
+        priceImpact: priceImpact?.toString(),
+        provider: exchangeRate.provider,
+      })
+    }
+  }, [exchangeRate])
```
kathaypacific, line 154: nit: there is already a `useEffect` that is triggered on `exchangeRate` change, can we put this logic there?

src/swap/SwapScreen.tsx
```diff
@@ -114,6 +113,46 @@ export function SwapScreenSection({ showDrawerTopNav }: { showDrawerTopNav: bool
...
+    const isPriceImpactMissing = priceImpact === null
+    const priceImpactExceedsThreshold =
+      !!priceImpact && priceImpact.gte(priceImpactWarningThreshold)
+
+    return { isPriceImpactMissing, priceImpactExceedsThreshold, priceImpact }
+  }
+
+  const { isPriceImpactMissing, priceImpactExceedsThreshold } = useMemo(() => {
+    return calculatePriceImpactData()
+  }, [exchangeRate])
```
kathaypacific, line 134: ah, this is more complex than i imagined but i think you did this to save doing the priceImpactExceedsThreshold logic multiple times right? this is okay, what do you think about changing the type of exchangeRate.estimatedPriceImpact to be already a big number and divided by 100 or null when it reaches this component? so then your useEffect hook can dispatch analytics when `!exchangeRate.estimatedPriceImpact || exchangeRate.estimatedPriceImpact.gte(priceImpactWarningThreshold)` and your render can use `!exchangeRate.estimatedPriceImpact` and `exchangeRate.estimatedPriceImpact.gte(priceImpactWarningThreshold)` directly? it would save a little bit of overhead to not have the extra variables that you need to keep synchronised?

locales/base/translation.json
dievazqu: Updated text from [designs](https://www.figma.com/file/jRtYI8GiSvPAaClBnRLLpu/In-App-Swap?type=design&node-id=3870-2341&mode=design&t=qxsdpx6QJ35SdgJA-0) shared by nitya.

src/swap/SwapScreen.tsx
dievazqu: I did these changes, I think the code looks better. Thanks!

locales/base/translation.json
```diff
@@ -1682,6 +1682,10 @@
       "title": "Before you swap",
       "body": "Swapping these assets may result in an unfavorable exchange rate. If you lower the amount, you may get a better rate."
     },
+    "missingSwapImpactWarning": {
+      "title": "Before you swap",
+      "body": "We can't estimate the value of this swap. Valora is not responsible if you choose to proceed."
```
kathaypacific, line 1687:
```suggestion
      "body": "We can't estimate the value of this swap. {{appName}} is not responsible if you choose to proceed."
```

kathaypacific (approved):
🤌🏻

PR #4087, opened 2023-08-18
fix: upgrade RN (no jest)

kathaypacific:
### Related issues

- Fixes #[issue number here]


MuckT:
@kathaypacific I was able to build and test with Android 5 and it's WAI.

kathaypacific:
> @kathaypacific I was able to build and test with Android 5 and it's WAI.

that's great, then can i get an approval please?

MuckT:
> that's great, then can i get an approval please?
@kathaypacific #4088 should be merged first.

dievazqu (commented):
Looks like there might be a bug in the refactor of `src/import/saga.ts`?

The rest looks ok... I don't have the knowledge to review the changes in the `android` folder though...

src/import/saga.ts
```diff
@@ -242,21 +242,26 @@ function* attemptBackupPhraseCorrection(mnemonic: string) {
...
-        return true
-      }
-    }
-    requests = requests.filter((_, i) => balances[i] === undefined)
+
+  const getBalanceCEur = yield* fork(fetchTokenBalanceInWeiWithRetry, Currency.Euro, address)
+  const getBalanceCUsd = yield* fork(fetchTokenBalanceInWeiWithRetry, Currency.Dollar, address)
+  const getBalanceCelo = yield* fork(fetchTokenBalanceInWeiWithRetry, Currency.Celo, address)
+
+  const { balanceCEur, balanceCUsd, balanceCelo } = yield* race({
```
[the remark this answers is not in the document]
dievazqu: I guess we don't want to wait the 3 threads to finish if we already know the user has some balance, but I don't know how much we are saving with all this code... Actually, with the current version I think we might be re-running the `fetchTokenBalanceInWeiWithRetry` if they didn't finish at the same time...

I think I prefer to keep the code simpler and wait until the 3 threads finish... What do you think?

Also, this code looks already old, since it only checks for Celo, cUSD and cEUR...

I know this is not part of the RN upgrade, so maybe we should just leave as it was and fix it later? I think we should include other tokens here, this is use in the `importBackupPhraseSaga`, I think we the current code we might be missing some addresses that have other tokens...

android/build.gradle
```diff
@@ -25,7 +20,6 @@ buildscript {
     dependencies {
         classpath 'com.android.tools.build:gradle:7.2.1'
```
MuckT, line 21: Should this be `classpath("com.android.tools.build:gradle:7.3.1")`?

ios/Podfile
```diff
@@ -47,8 +53,7 @@ target "celo" do
     # Hermes is now enabled by default. Disable by setting this flag to false.
     # Upcoming versions of React Native may rely on get_default_flags(), but
     # we make it explicit here to aid in the React Native upgrade process.
-    # Disabled for now due to issues with Detox tests
-    :hermes_enabled => false,
+    :hermes_enabled => flags[:hermes_enabled],
```
MuckT, line 56: I'd argue we shouldn't enable hermes in this upgrade PR, but I also don't think it should block this from being merged. If we end up having an issue in CI we can revert similar to #3859.

ios/Podfile
MuckT, line 1: I see some comments in the upgrade tool around flipper in this file are we fine to exclude these changes?

>  If you are using a `react-native-flipper` your iOS build will fail when `NO_FLIPPER=1` is set.

patches/react-native-webview+11.6.5.patch
MuckT, line 1: This patch should be updated to use `13.3.1`.

test/RootStateSchema.json
```diff
@@ -903,6 +903,48 @@
             ],
             "type": "object"
         },
+        "CoreTypes.Metadata_1": {
```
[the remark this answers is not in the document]
MuckT: This seems suspect, do we have a duplicated state variable?

android/build.gradle
kathaypacific: yup, thanks! updated

ios/Podfile
kathaypacific: i'd like to enable hermes because it is a better user experience. i haven't noticed any issues with the CI so i'd like to leave it until we see issues

ios/Podfile
kathaypacific: yes we have a custom implementation for flipper, it is fine to omit those changes

patches/react-native-webview+11.6.5.patch
kathaypacific: thanks, good catch! updated

test/RootStateSchema.json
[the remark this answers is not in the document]
kathaypacific: this is not a duplicated state variable - the extra `_1` objects are type variables only.

this happened because there were duplicated walletconnect dependencies in the yarn.lock. i've resolved this

kathaypacific:
> > that's great, then can i get an approval please?
> 
> @kathaypacific should #4088 be merged first?

@MuckT i'd like an approval separately on this PR to confirm its contents before i merge #4088. that PR has over 8k new lines of changes, it will make it way more difficult to review if it is merged with this one. that's why i split these PRs up - even though the PRs are still large, the purpose of the PRs are specific :)

src/import/saga.ts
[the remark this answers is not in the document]
dievazqu: Yes!! This is even better and the code is so much simpler!!

dievazqu (approved):
LGTM! 🤞

PR #4088, opened 2023-08-18
fix: upgrade to react native 0.71 (with tests)

kathaypacific:
### Description

Notable changes:
- jest was updated to v29 - this is 3 major versions up so combined with the RN update, there are loads of snapshot changes that don't have any real impact. we should really get rid of snapshot tests for this reason.
- detox was upgraded to v20, i followed the migration guide [here](https://wix.github.io/Detox/docs/guide/migration/#200)
- ts-jest was updated to keep in sync with jest, but `ts-jest/utils` has been [removed](https://kulshekhar.github.io/ts-jest/docs/27.1/guides/test-helpers/). so there are many changes to update `mocked` to `jest.mocked`
- the way timers work changed in jest 29. to be able to override the global defaults, i'm using a `beforeAll` in the jest setup file before applying fake timers globally.
- replace `flushMicroservicesQueue` with `await act(() => {...})` - this fixed a lot of test flake and is the recommended way to await state changes and async effects

### Test plan

- unit tests work ✅ 
- e2e tests work ✅

### Related issues

- Fixes RET-792

### Backwards compatibility

Y


.github/workflows/e2e-pr.yml
```diff
@@ -3,6 +3,7 @@ on:
   pull_request:
     branches:
       - main
+      - kathy/RN-upgrade-71-2
```
kathaypacific, line 6: i'll remove this before i merge into main

ios/Podfile.lock
```diff
@@ -688,7 +688,7 @@ PODS:
     - RNFBApp
   - RNFS (2.20.0):
     - React-Core
-  - RNGestureHandler (2.9.0):
```
MuckT, line 691: Is this an expected change?

MuckT (approved):
Awesome work! 🚀 

I took a look at the test run duration and it looks like we have a significant improvement for both unit and e2e tests.

### Unit Tests
| Before | After | % Change |
| :---: | :---: | :---: |
| 4m 47s | 3m 33s | ~34% faster |

### E2E Tests (iOS 6 workers)
| Before  | After | % Change |
| :---: | :---: | :---: |
| 15m 41s | 12m 40s | ~24% faster |

src/RevokePhoneNumber.tsx
```diff
@@ -58,7 +58,13 @@ export const RevokePhoneNumber = ({ forwardedRef }: Props) => {
 
   const handleRevokePhoneNumber = async () => {
     ValoraAnalytics.track(SettingsEvents.settings_revoke_phone_number_confirm)
-    await revokeNumberAsync.execute()
+    try {
+      await revokeNumberAsync.execute()
+    } catch (error) {
+      // the error handler for revokeNumberAsync handles the error but it is
```
dievazqu, line 64: nit: Should we at least have a log message here?
Just in case we are ignoring an unexpected error?

dievazqu (approved):
LGTM 🙈

src/RevokePhoneNumber.tsx
kathaypacific: i can add a logger error here, but i'm not sure it has real user impact since the `onError` block is triggered. the error just keeps propagating up but i don't think there's much we can do about it

PR #4096, opened 2023-08-21
feat(keyless-backup): show/hide based on backup status

satish-ravi:
### Description

Add a state to track backup status and use it to show / hide nav item and show setup / delete on settings. The state is added on account because the entire keylessBackup state is not persisted 

### Test plan

Unit tests
Manually, by backing up and ensuring the drawer item is hidden and delete shows up

### Related issues

N/A

### Backwards compatibility

Yes


src/account/reducer.ts
```diff
@@ -16,7 +17,7 @@ export interface State {
   devModeClickCount: number
   photosNUXClicked: boolean
   pincodeType: PincodeType
-  backupCompleted: boolean
+  backupCompleted: boolean // recovery phrase backup
```
cajubelt, line 20: this seems redundant with keyless backup state, which contains the status of the backup. it also just seems non-intuitive to me to keep keyless backup state in two different places. Can you explain why you decided to add it here?

src/navigator/DrawerNavigator.tsx
```diff
@@ -224,9 +226,10 @@ export default function DrawerNavigator({ route }: Props) {
 
   const shouldShowNftGallery = getFeatureGate(StatsigFeatureGates.SHOW_IN_APP_NFT_GALLERY)
 
-  const shouldShowKeylessBackup = getFeatureGate(
-    StatsigFeatureGates.SHOW_CLOUD_ACCOUNT_BACKUP_SETUP
-  )
+  const cloudBackupGate = getFeatureGate(StatsigFeatureGates.SHOW_CLOUD_ACCOUNT_BACKUP_SETUP)
+  const anyBackupCompleted = backupCompleted || cloudBackupCompleted
+  const shouldShowWalletSecurity = !anyBackupCompleted && cloudBackupGate
+  const shouldShowRecoveryPhrase = !anyBackupCompleted && !cloudBackupGate
```
cajubelt, line 232: I think we should avoid "shouldShow" prefixes

src/account/reducer.ts
satish-ravi: the entire keylessBackup field is blacklisted so we don't persist sensitive info, so we need to track it in a separate place. It's possible to blacklist a subset of fields, but that requires a significant refactor in our reducer setup. 

> this seems redundant with keyless backup state

I believe we need two statuses because this will be true for existing users who have done the backup quiz, but we still want to show the "Set up" option for cloud backup. Although I guess if someone finishes cloud backup, maybe there shouldn't be an option to do the backup quiz at all?

src/navigator/DrawerNavigator.tsx
satish-ravi: fixed

src/account/reducer.ts
cajubelt: no I think there should still be an option

cajubelt: more backups is good

cajubelt: I left this comment on the wrong place, meant it to be on `cloudBackupCompleted`

cajubelt (approved):

src/account/reducer.ts
cajubelt: > the entire keylessBackup field is blacklisted so we don't persist sensitive info

I didn't realize this so I'm marking as resolved now

PR #4112, opened 2023-08-24
feat(nav-drawer): upgrade to v6 and enable freeze

MuckT:
### Description

Updates @react-navigation/drawer to v6 and enables freeze for performance improvements. Some prop changes were needed to maintain existing styles.

Freeze works as follows:
> ...Using Suspense mechanism introduced in React 17. The main use case of this library is to avoid unnecessary re-renders of parts of the app that are not visible to the user at a given moment. The frozen components are not unmounted when they are replaced with a placeholder view, so their React state and corresponding native view instances are retained during the freeze (DOM elements for react-dom or platform-native views for React Native apps) keeping things like scroll position, input state, or loaded images (for  components) unchanged. [source](https://github.com/software-mansion/react-freeze).

### Test plan

- Tested locally on Android 5.

### Related issues

N/A

### Backwards compatibility

Yes


MuckT:
Closing in favor of #4135

PR #4115, opened 2023-08-24
feat(multichain): Render multichain transfers on the home feed

jophish:
### Description

For ACT-887. This PR updates the wallet to render non-Celo transfers in the transaction feed on the home screen. This feature is locked behind a dynamic config parameter at the moment ([this one](https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/dynamic_configs/multi_chain_features), to be exact). This dynamic config also contains a parameter for separately controlling rendering of native tokens. For users who fail the feature gate, network requests will _not_ be made to fetch non-Celo transfers, and any previously cached non-Celo transfers will not be rendered.

The actual rendering of non-Celo transfers is kind of broken at the moment, but that's to be expected. In the video below, you can see the example transfer on Goerli Etherscan [here](https://goerli.etherscan.io/address/0xe192673ad298ff0b72dd1c8e074d34ddaafd9591) (compare the send address against the video).

Note that actually dealing with native assets (i.e., token-like thing that lacks an actual address) is being punted on for now, since it will require a larger overhaul to how we represent tokens in the wallet and throughout our entire stack.

### Test plan

Manual tested on the Goerli testnet. See video below.

https://github.com/valora-inc/wallet/assets/569401/878a193f-e1f4-4b62-9fc2-1360625a4537

### Related issues

- Fixes ACT-887.

### Backwards compatibility

Yes.


src/components/TokenDisplay.tsx
```diff
@@ -54,9 +54,6 @@ function TokenDisplay({
   style,
   testID,
 }: Props) {
-  if (tokenAddress ? currency : !currency) {
-    throw new Error('TokenDisplay must be passed either "currency" or "tokenAddress" and not both')
```
cajubelt, line 58: I'm lacking some context on why this was here to begin with. but if tokenAddress is now optional, you could still throw if both are provided

cajubelt: though just using discriminated union type for the params would be nicer, since it would result in a compile-time check for this rather than a runtime error

src/statsig/constants.ts
```diff
@@ -34,6 +34,7 @@ export const FeatureGates = {
   [StatsigFeatureGates.SHOW_NOTIFICATION_CENTER]: false,
   [StatsigFeatureGates.SHOW_CLOUD_ACCOUNT_BACKUP_SETUP]: false,
   [StatsigFeatureGates.SHOW_CLOUD_ACCOUNT_BACKUP_RESTORE]: false,
+  [StatsigFeatureGates.SHOW_MULTI_CHAIN_TRANSFERS]: false,
```
cajubelt, line 37: it might be worth using a dynamic config so we can release multichain features one by one if we wanted, and have them all live in the same place.

as soon as one feature gate makes it into an app version we'll need to keep it, so I do think it makes sense to be forward-thinking on this.

src/tokens/hooks.ts
```diff
@@ -9,9 +9,10 @@ import {
 import { convertLocalToTokenAmount, convertTokenToLocalAmount } from 'src/tokens/utils'
 import { Currency } from 'src/utils/currencies'
 
-export function useTokenInfo(tokenAddress: string) {
+export function useTokenInfo(tokenAddress: string | undefined) {
```
cajubelt, line 12:
```suggestion
export function useTokenInfo(tokenAddress?: string) {
```
code golf

src/transactions/feed/TransactionFeed.tsx
```diff
@@ -43,7 +48,7 @@ function TransactionFeed() {
     useFetchTransactions()
 
   const cachedTransactions = useSelector(transactionsSelector)
-
+  Logger.info(TAG, cachedTransactions)
```
cajubelt, line 51: did you mean to leave this here? should it be Logger.debug or removed?

src/transactions/feed/queryHelper.ts
```diff
@@ -60,70 +62,96 @@ const deduplicateTransactions = (
   return transactionsWithoutDuplicatedHash
 }
 
+export function getAllowedChains(): Array<Chain> {
+  return getFeatureGate(StatsigFeatureGates.SHOW_MULTI_CHAIN_TRANSFERS)
```
cajubelt, line 66: seems like another way of doing this could be to make the value of the feature gate a list of chains

cajubelt: it would give us more granular control over which chains are supported at the feature flag level

src/transactions/feed/queryHelper.ts
```diff
@@ -228,30 +331,44 @@ async function queryTransactionsFeed(
     },
     body: JSON.stringify({
       query: TRANSACTIONS_QUERY,
-      variables: { address, localCurrencyCode, afterCursor },
+      variables: { address, localCurrencyCode, chain, afterCursor },
```
cajubelt, line 334: I think it's fine to use the Chain enum you defined throughout most of the app, but at the API level I think we should use chain ID's agreed on here https://valora-app.slack.com/archives/C025V1D6F3J/p1689089907854609

cajubelt: I realize this will require an update in blockchain-api but I think it's worth it

cajubelt: I made an attempt and converting between testnet/mainnet-agnostic "chain names" and "chain id"'s [here](https://github.com/valora-inc/wallet/pull/4099/files) that you may find useful

cajubelt: in particular you can look [here](https://github.com/valora-inc/wallet/blob/4e6491825cba38d209e702f70f9e8573e98b7acd/src/utils/blockchains.ts) and [here](https://github.com/valora-inc/wallet/blob/4e6491825cba38d209e702f70f9e8573e98b7acd/src/config.ts#L103)

cajubelt: another way to do it could be to define chain ID's as new configuration parameters. Though I avoided it because I felt like it introduced a potential for bugs where your eth chain id and your celo chain id don't match (one is mainnet and one is testnet, for instance)

cajubelt (commented):
I gave this a quick look and should look more closely tomorrow

src/components/TokenDisplay.tsx
jophish: Using a discriminated union would be nicer, but it would mean we would need two new types e.g.

```ts
interface Address {
  kind: 'address',
  tokenAddress: string
}

interface Currency {
  kind: 'currency',
  currency: string
} 

type AddressOrCurrency = Address | Currency
```
which would require changes to a ton of files that use this component, so I think this is out of scope for this PR. This now throws an error when both are provided, however.

src/statsig/constants.ts
jophish: You mean [something like this](https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/dynamic_configs/multi_chain_feature_gates)? A dynamic config that contains all the feature gates that we want?

src/tokens/hooks.ts
jophish: :man_facepalming:

src/transactions/feed/TransactionFeed.tsx
jophish: Oops, should be removed

src/transactions/feed/queryHelper.ts
jophish: Yeah, added to the dynamic config linked above, instead of a feature gate.

src/transactions/feed/queryHelper.ts
jophish: Ah, interesting -- yeah, I'll add this into this PR, thanks for the references here.

jophish: Actually, this raises an interesting point -- blockchain-api already has deployments in mainnet and testnet envs, and thus we don't actually need to specify the _specific_ blockchain environment when parameterizing by chain in GraphQL calls. Our "formal" chain specifiers are environment specific, however. Updating blockchain-api to require us to use these environment-specific identifiers might be a little weird. For instance, when typing the request schema, what should the type of `chain` be? It seemingly depends on the environment it's deployed in, unless we want e.g. `"ethereum-goerli"` to be a statically valid (but semantically meaningless) chain to request against the mainnet blockchain-api deployment (which would presumably just throw at runtime).

src/components/TokenDisplay.tsx
cajubelt: I think you could do it with just this
```
interface Address {
  tokenAddress: string
}
interface Currency {
  currency: string
}
type AddressOrCurrency = Address | Currency
```
but yeah not that important

src/transactions/feed/queryHelper.ts
cajubelt: in my draft PR I suggested we use a test/mainnet-agnostic "chain name" enum that then maps to "chain id" depending on the env vars when it accesses the API's of various backend services. one option for dealing with chain id's in the backend services themselves is to do essentially the same thing in reverse, mapping chain ID's to chain names. We could throw errors if we get the "wrong" chain ID for the deployment of the service, e.g. a dev server receives a mainnet request. The upside of this is if we mis-configured our development/production app to hit the wrong backend, it would be less likely to silently failed (unless we also mis-configured which blockchain network to use, testnet vs mainnet). The downside is it's a bit more complex than just using chain names, for limited gains-- I'm not sure we've ever seen this kind of configuration error in the past.

Other alternatives are:
1. accept testnet or mainnet requests from all deployments (seems pointless to have multiple deployments at that point)
2. ditch the chain ID idea and use test/mainnet-agnostic chain names at the API level

I think this probably warrants discussion with the broader eng team

src/statsig/constants.ts
cajubelt: yes. you can probly get rid of the feature gate now

cajubelt: ok now that I'm looking at this with "fresh eyes" (a weekend.. lol) I'm realizing that my suggestion is something we haven't done before, which is to use dynamic configs to essentially namespace our feature flags. if you're into this idea then go for it but I could imagine some arguments against doing it that way-- that it's an abuse of dynamic configs for instance, that feature gates are simpler, etc.

cajubelt: the list of chains you set up is cool though

src/transactions/feed/queryHelper.ts
```diff
@@ -180,11 +225,19 @@ export function useFetchTransactions(): QueryHookResult {
...
+    // - We are not already loading
+    // - There exists at least one chain that has futher pages
+    // - EITHER we do not yet have enough TXs, OR NO chains whatsoever produced results
+    //    in the most recent round of fetching (which corresponds to case 2. above
+    //    occuring for all chains simaltaneously)
     const { transactions, pageInfo, hasTransactionsOnCurrentPage } = fetchedResult
     if (
       !loading &&
-      pageInfo?.hasNextPage &&
-      (transactions.length < MIN_NUM_TRANSACTIONS || !hasTransactionsOnCurrentPage)
```
cajubelt, line 187: I don't get why this is an OR, and it makes me uneasy to see the changes not understanding why it was done this way to begin with. Speaking of the original code only-- why would you fetch another page if you had already reached the minimum number of transactions to display? And now of the new changes-- are you sure that more transactions should be fetched only if *all* chains have no transactions on the current page?

src/transactions/feed/queryHelper.ts
```diff
@@ -210,16 +263,67 @@ export function useFetchTransactions(): QueryHookResult {
   }
 }
 
-async function queryTransactionsFeed(
-  address: string | null,
-  localCurrencyCode: string,
+function anyChainHasMorePages(pageInfo: { [key in Chain]?: PageInfo | null }): boolean {
+  return Object.values(pageInfo).reduce((acc, val) => acc || !!val?.hasNextPage, false)
```
cajubelt, line 267:
```suggestion
  return Object.values(pageInfo).some(({ hasNextPage }) => !!hasNextPage)
```
code golf

cajubelt: not sure if it compiles, but the main point is that "some" could be used here instead of reduce

src/transactions/feed/queryHelper.ts
```diff
@@ -210,16 +263,67 @@ export function useFetchTransactions(): QueryHookResult {
...
+}
+
+function anyChainHasTransactionsOnCurrentPage(hasTransactionsOnCurrentPage: {
+  [key in Chain]?: boolean
+}): boolean {
+  return Object.values(hasTransactionsOnCurrentPage).reduce((acc, val) => acc || val, false)
```
cajubelt, line 273: same here

src/transactions/feed/queryHelper.ts
```diff
@@ -210,16 +263,67 @@ export function useFetchTransactions(): QueryHookResult {
...
+}: {
+  address: string | null
+  localCurrencyCode: string
+  params: Array<{
+    chain: Chain
+    afterCursor?: string
+  }>
+}): Promise<{ [key in Chain]?: QueryResponse }> {
+  const results = await Promise.all(
+    params.map((params) =>
```
cajubelt, line 291: the callback's "params" argument should not shadow the variable name from outer scope, very confusing

src/transactions/feed/queryHelper.ts
jophish: My understanding was that the original intent of this hook is to cover two cases:

1. We're fetching initial transactions to populate the home feed, and we don't have enough yet -- keep fetching more pages until we have enough. 
2. We're fetching transactions _for any reason_ (whether through polling or an explicit request) and the current page that we got back from blockchain-api contains _no_ transactions at all, for whatever reason. This is indeed possible. Fetch the next page to "make up" for this empty page. For example, if a user swipes down to load more TXs and there are in fact more TXs available, but blockchain-api returns an intermediate empty page, the user shouldn't have to swipe again.

In both these cases, subsequent requests will only be made if more pages actually exist. In the move to multichain, nothing needs to really change for case 1. For case 2, what we're trying to avoid is _no_ transactions appearing despite more pages being available; for multichain, this occurs when no chains produce TXs on their most-recently-fetched page, but at least one chain has further pages remaining.

src/tokens/e2eTokens.ts
```diff
@@ -1,9 +1,9 @@
 import { StoredTokenBalances } from 'src/tokens/slice'
 
 // alfajores addresses
-const cUSD = '0x874069Fa1Eb16D44d622F2e0Ca25eeA172369bC1'
-const cEUR = '0x10c892A6EC43a53E45D0B916B4b7D383B1b78C0F'
-const CELO = '0xF194afDf50B03e69Bd7D057c1Aa9e10c9954E4C9'
+const cUSD = '0x874069Fa1Eb16D44d622F2e0Ca25eeA172369bC1'.toLowerCase()
+const cEUR = '0x10c892A6EC43a53E45D0B916B4b7D383B1b78C0F'.toLowerCase()
+const CELO = '0xF194afDf50B03e69Bd7D057c1Aa9e10c9954E4C9'.toLowerCase()
```
jophish, line 6: These changes were required to fix an e2e test that was failing to pass, though I'm still not quite sure why. TLDR, since these addresses were not normalized, they [were being stored in redux](https://github.com/valora-inc/wallet/blob/main/src/tokens/saga.ts#L266) with non-normalized keys. The `tokenAddress` fields returned in responses from blockchain-api, however, _are_ lowercase, which means that the `useTokenInfo` hook [here](https://github.com/valora-inc/wallet/blob/main/src/tokens/hooks.ts#L12) was failing to find the tokens, since the e2e tokens were placed into redux with non-normalized token addresses for keys.

From looking at blockchain-api, I don't believe that we ever switched from using non-normalized to normalized strings for token addresses at any point; I believe they've always been lowercased, so I'm not certain why this issue is appearing in this PR.

src/transactions/feed/queryHelper.ts
```diff
@@ -217,8 +217,8 @@ export function useFetchTransactions(): QueryHookResult {
   )
 
   useEffect(() => {
-    // this hook does 2 things:
-    // 1. ensures that we populate the entire screen with transactions on load
+    // this hook does 2 1:
+    // things. ensures that we populate the entire screen with transactions on load
```
cajubelt, line 221: this comment appears to have been messed up

cajubelt (approved):

cajubelt:
probly best to wait until network naming conventions conversation is resolved before merging, since this includes network names (including Celo) in querying blockchain api

src/statsig/constants.ts
satish-ravi: Would having the dynamic config make rolling this out really complex? We'd likely roll out different features in different versions and the dynamic config would end up something like
```js
if (version === 'x') return { featureA: true, featureB: ['a'], ...}
if (version === 'y') return { featureA: false, featureB: ['a', 'b'], ... }
...
return { defaults }
```
Also, we won't be able to use metrics when rolling out features unless we use a feature gate itself as a condition in dynamic config, although not sure how important that is here.

But I do like that this allows us to use more than booleans.

src/components/TokenDisplay.test.tsx
```diff
@@ -71,6 +69,27 @@ describe('TokenDisplay', () => {
...
     })
+    it('throws when neither currency nor tokenAddress are supplied', () => {
+      expect(() =>
+        render(
+          <Provider store={store()}>
+            <TokenDisplay showLocalAmount={false} amount={10} testID="test" />
+          </Provider>
+        )
+      ).toThrow()
+    })
```
satish-ravi, line 80: just to confirm, this is the same test as earlier just moved around right?

src/components/TokenTotalLineItem.tsx
```diff
@@ -12,7 +12,7 @@ import { LocalAmount } from 'src/transactions/types'
 
 interface Props {
   tokenAmount: BigNumber
-  tokenAddress: string
+  tokenAddress?: string
```
satish-ravi, line 15: should this be string | null? since blockchain API returns null and doesn't drop the field?

src/fiatExchanges/FiatExchangeAmount.test.tsx
```diff
@@ -19,7 +19,14 @@ const mockUseMaxSendAmount = jest.fn(() => mockMaxSendAmount)
 jest.mock('src/fees/hooks', () => ({
   useMaxSendAmount: () => mockUseMaxSendAmount(),
 }))
-jest.mock('src/statsig')
+jest.mock('src/statsig', () => {
+  return {
+    getFeatureGate: jest.fn(),
+    getDynamicConfigParams: jest.fn().mockReturnValue({
+      show_native_tokens: false,
+    }),
```
satish-ravi, line 27: Wonder if there's a way we can avoid including a mock related to an implementation detail of the TokenDisplay component here. Seems like its spread across multiple files, and this is something we'd likely forget to drop if say TokenDisplay no longer depends on this dynamic config any longer

src/redux/migrations.ts
```diff
@@ -1218,4 +1218,28 @@ export const migrations = {
...
+          return { ...tx, chain: Chain.Celo }
+        }
+      ),
+      transactions: (state.transactions.transactions as TokenTransaction[]).map((tx) => {
+        const __typename = // @ts-ignore
+          tx.__typename === 'TokenTransferV2'
+            ? 'TokenTransferV3' // @ts-ignore
+            : tx.__typename === 'NftTransferV2'
+            ? 'NftTransferV3'
+            : 'TokenExchangeV3'
```
satish-ravi, line 1236: Is it safe to change these in place? Or should we just drop them and let the app fetch and build this list again?

src/transactions/feed/queryHelper.ts
```diff
@@ -228,30 +332,44 @@ async function queryTransactionsFeed(
...
+      },
+    },
+  }
 }
 
 export const TRANSACTIONS_QUERY = `
-  query UserTransactions($address: Address!, $localCurrencyCode: String, $afterCursor: String) {
-    tokenTransactionsV2(address: $address, localCurrencyCode: $localCurrencyCode, afterCursor: $afterCursor) {
+  query UserTransactions($address: Address!, $localCurrencyCode: String, $afterCursor: String, $chain: Chain!) {
+    tokenTransactionsV3(address: $address, localCurrencyCode: $localCurrencyCode, afterCursor: $afterCursor, chain: $chain) {
```
satish-ravi, line 357: With this change, we can no longer make any backwards incompatible change to tokenTransactionsV3 right?

src/components/TokenDisplay.test.tsx
jophish: yep, that's correct

src/redux/migrations.ts
jophish: My assumption was that it was safe, since the rest of the app has been updated to handle these new typenames.

src/transactions/feed/queryHelper.ts
jophish: Yeah, that's correct -- if we want to be safe, we can re-use the feature gate that gates showing multichain transactions to switch between the v2 and v3 endpoints.

src/statsig/constants.ts
jophish: Yeah this is a good point -- I'll revert this back to using feature gates, since this seems like it adds more complexity than is worth it.

src/fiatExchanges/FiatExchangeAmount.test.tsx
jophish: yeah, it's not great, though with the refactor to using feature gates again, a lot of these have become redundant/simplified

PR #4116, opened 2023-08-24
fix: upgrade to react native 0.71 (notifications)

kathaypacific:
### Description

Android API level 33+ now requires the [push notification permission](https://developer.android.com/develop/ui/views/notifications/notification-permission#declare). By default, it is not allowed for an app to send notifications to an app. This PR adds some code to manually track if we have requested permissions.

I struggled to make Clevertap push notifications work, so I upgraded the clevertap sdk which seems to have fixed the problem. There were [breaking changes](https://developer.clevertap.com/docs/android-push#android-13-changes) with Android 13. I didn't spend too long to investigate what was causing issues with the old sdk, since we want to update dependencies anyway.

### Test plan

End to end tests Android
- can receive push notifications from Clevertap ✅ ([dashboard](https://us1.dashboard.clevertap.com/46W-4W9-4K6Z/campaigns/campaign/1692888205/report/stats/trend))
- can receive "payment received" notification ✅

End to end tests iOS
- can receive push notifications from Clevertap  🟡 (I haven't been able to test this, i think it is because of a mismatched APNS token between what Clevertap has configured and my local - i get the same APNS error on the main branch too, last time i worked on this i tested using the nightly build. i'm fairly confident that there should be no related changes in the RN upgrade that would affect this integration, so i would like to save time and test this after it is merged rather than getting my local env to work. clevertap is a bit of a black box and it is not easy to debug the cause of this APNS issue)
- can receive "payment received" notification  ✅

### Related issues

- Fixes RET-792

### Backwards compatibility

Y


dievazqu (approved):
LGTM!

jophish (approved):
:tada:

PR #4118, opened 2023-08-24
chore(wallet): Add viem wallet

jh2oman:
#### KeychainLock
* Manages locking/unlocking/updating password for multiple accounts
* Singleton

#### KeychainSigner
* Doesn't have actually much to do with the Keychain anymore since that functionality got moved to KeychainLock, may rename this to `LockableSigner`

#### KeychainWallet
* Small changes to use KeychainLock instead of KeychainSigner to manage locks

#### getLockableViemWallet
* uses the KeychainLock to block any WalletActions (currently just writeContract) if the account is locked

### Test plan

I've manually tested creating a new account, importing an existing account, changing pincode.

### Related issues

https://linear.app/valora/issue/ACT-882/switch-to-viem


src/utils/formatting.ts
```diff
@@ -11,16 +11,19 @@ export const getMoneyDisplayValue = (
   includeSymbol: boolean = false,
   roundingTolerance: number = 1
 ): string => {
+  const bigValue = new BigNumber(value)
   const decimals = CURRENCIES[currency].displayDecimals
   const symbol = CURRENCIES[currency].symbol
   // For stable currencies, if the value is lower than 0.01 we show an extra decimal point.
   // If the value is lower than 0.001, we just show <$0.001.
   const minValueToShow = Math.pow(10, -decimals - (currency === Currency.Celo ? 0 : 1))
-  if (value > 0 && value < minValueToShow) {
```
jh2oman, line 19: This code was comparing BigNumber.Value and number directly which is not allowed in the new typescript version

src/web3/KeychainSigner.ts
```diff
@@ -229,37 +59,8 @@ export class KeychainSigner implements Signer {
...
-   * @param newPassphrase - the new passphrase to use with the account
-   * @returns whether the update was successful
-   */
-  async updatePassphrase(oldPassphrase: string, newPassphrase: string) {
-    const privateKey = await getStoredPrivateKey(this.account, oldPassphrase)
-    if (!privateKey) {
-      return false
-    }
-    await storePrivateKey(privateKey, this.account, newPassphrase)
+    this.signer = new LocalSigner(privateKey)
```
jh2oman, line 62: still has the functionality of the new signer getting created on every unlock (not sure if thats actually needed)

src/web3/KeychainLock.ts
```diff
@@ -0,0 +1,244 @@
...
+  }
+  return `${ACCOUNT_STORAGE_KEY_PREFIX}${account.createdAt.toISOString()}--${normalizeAddress(
+    account.address
+  )}`
+}
+
+async function encryptPrivateKey(privateKey: string, password: string) {
+  return CryptoJS.AES.encrypt(privateKey, password).toString()
+}
+
```
jh2oman, line 46: Completely copy/pasted. I thought it now made more sense in this file than KeychainSigner

src/web3/contracts.ts
```diff
@@ -109,6 +116,30 @@ async function waitForContractKit(tries: number) {
...
+  const keys = yield* call(generateKeysFromMnemonic, mnemonic)
+  const privateKey = keys.privateKey
+  if (!privateKey) {
+    throw new Error('Failed to convert mnemonic to hex')
+  }
+  const wallet = getLockableViemWallet(
+    keychainLock,
+    chain,
+    normalizeAddressWith0x(privateKey) as Address
+  )
```
satish-ravi, line 137: This means that a viem wallet instance is created only when we have a private key right? Would we ever need a wallet instance when the account is not unlocked?

src/viem/getLockableWallet.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+  }
+
+  return createWalletClient({
+    chain,
+    transport: http(),
+    account,
+  }).extend((client: Client): Actions => {
+    return {
+      writeContract: (args) => {
```
satish-ravi, line 50: this is a predefined method on the client which were are overriding and unlockAccount is a custom action we're adding right? Maybe add a comment about it?

src/web3/contracts.ts
```diff
@@ -109,6 +116,30 @@ async function waitForContractKit(tries: number) {
...
+export function* getViemWallet(chain: Chain) {
+  if (viemWallets.has(chain)) {
+    return viemWallets.get(chain) as ViemWallet
+  }
+  const walletAddress = yield* select(walletAddressSelector)
+  const mnemonic = yield* call(getStoredMnemonic, walletAddress)
+  if (!mnemonic) {
+    throw new Error('No mnemonic found')
+  }
+  const keys = yield* call(generateKeysFromMnemonic, mnemonic)
```
satish-ravi, line 128: do we have to do this every time or could we use the getStoredPrivateKey method in KeychainLock?

src/web3/contracts.ts
```diff
@@ -109,6 +116,30 @@ async function waitForContractKit(tries: number) {
   return contractKit
 }
 
+export function* getViemWallet(chain: Chain) {
+  if (viemWallets.has(chain)) {
+    return viemWallets.get(chain) as ViemWallet
+  }
+  const walletAddress = yield* select(walletAddressSelector)
+  const mnemonic = yield* call(getStoredMnemonic, walletAddress)
```
satish-ravi, line 124: this would require a PIN right? does it open the pin screen automatically?

src/viem/getLockableWallet.ts
```diff
@@ -0,0 +1,58 @@
...
+        checkLock()
+        return writeContract(client, args)
```
satish-ravi, line 52: this is a little different from the existing wallet right? where a Signer instance has access to the private key only when the account is unlocked, whereas here the wallet instance always includes the account with the pk, but write methods are gated with a pin request?

src/web3/KeychainSigner.ts
satish-ravi, line 62: nit: rename signer to localSigner? since this class is also named KeychainSigner

src/web3/KeychainLock.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    })
+    await lock.unlock(mockAccount, 'password', 100)
+    expect(lock.isUnlocked(mockAccount)).toBe(true)
+  })
+  it('unlock returns false if the account has not been added', async () => {
+    expect(await lock.unlock(mockAccount, 'password', 100)).toBe(false)
+  })
+  it('unlock throws if has been added but private key is not in the keychain', async () => {
+    lock.addAccount({ address: mockAccount, createdAt: date })
+    await expect(async () => await lock.unlock(mockAccount, 'password', 100)).rejects.toThrow()
```
satish-ravi, line 66:
```suggestion
    await expect(lock.unlock(mockAccount, 'password', 100)).rejects.toThrow()
```

src/web3/contracts.ts
jh2oman: Correct.
I don't see anything in the import or create wallet flows that would require a wallet before its unlocked. 

The only other instance is when we load data from the Keychain. That is currently logic that lives in KeychainWallet and we will eventually have to de-couple from contractkit. When we do that de-coupling we shouldn't need an actual wallet before unlock, just some construct to track that a specific address has been loaded from Keychain.

src/web3/contracts.ts
jh2oman: we could go the list stored accounts route... thats probably cleaner!

jh2oman: You're right, it would be better to get from the keychain

src/web3/contracts.ts
jh2oman: yes it would, but refactoring to get the pin more directly

src/viem/getLockableWallet.ts
jh2oman: It is slightly different, but I believe the logic ends up being the same. 

One key is that in its current state the viem wallet doesn't even have signTransaction, signMessage, etc. functions. We'd need to add those alongside `writeContract` as another action. Essentially all wallet functionalities are 'Actions' in Viem https://viem.sh/docs/actions/wallet/introduction.html

src/web3/contracts.ts
```diff
@@ -109,6 +121,36 @@ async function waitForContractKit(tries: number) {
   return contractKit
 }
 
+// This code assumes that the account for walletAddress already exists in the Keychain
+// which is a responsibility currently handled by KeychainWallet
+export function* getViemWallet(chain: Chain) {
```
satish-ravi, line 126: can we add unit tests for this method?

src/web3/contracts.ts
```diff
@@ -109,6 +121,36 @@ async function waitForContractKit(tries: number) {
...
+  }
+  const password = yield* call(getPasswordSaga, walletAddress, false, true)
+  const privateKey = yield* call(getStoredPrivateKey, account, password)
+  if (!privateKey) {
+    throw new Error(`Private key not found for account ${walletAddress}`)
+  }
+  const wallet = getLockableViemWallet(
+    keychainLock,
+    chain,
+    normalizeAddressWith0x(privateKey) as Address
```
satish-ravi, line 147: Is the normalizeAddress required here? Seems like getStoredPrivateKey ensures this now with Jean's fix

src/web3/KeychainSigner.ts
```diff
@@ -239,58 +59,26 @@ export class KeychainSigner implements Signer {
...
 
-  /**
-   * Get the local signer. Throws if not unlocked.
-   */
-  protected get localSigner(): LocalSigner {
-    if (!this.isUnlocked()) {
-      this.unlockedLocalSigner = null
-    }
-    if (!this.unlockedLocalSigner) {
+  protected get unlockedLocalSigner(): LocalSigner {
```
satish-ravi, line 75: So we've flipped localSigner and unlockedLocalSigner? And should we keep the existing functionality of setting the localSigner instance variable to null if not unlocked? 
Also can we add some unit tests for at least this and the unlock method?

src/viem/saga.ts
```diff
@@ -0,0 +1,34 @@
...
+import { tokenAmountInSmallestUnit } from 'src/tokens/saga'
+import Logger from 'src/utils/Logger'
+import { publicClient } from 'src/viem'
+import { getViemWallet } from 'src/web3/contracts'
+import networkConfig from 'src/web3/networkConfig'
+import { call } from 'typed-redux-saga'
+import { getAddress } from 'viem'
+
+// example usage
+export function* send(
```
cajubelt, line 12: I think it was a really good idea to write this function to confirm that the changes work as expected and lend themselves to a good interface and devX for upcoming work. should it have a todo for plugging it into a main flow and a link to the GH issue where that work is going to get done? just so it doesn't get forgotten about and become dead code

cajubelt:
```suggestion
export function* send( // todo(ACT-786): build on this when implementing viem version of sendAndMonitorTransaction
```
maybe this would suffice

src/web3/KeychainSigner.ts
jh2oman: I flipped back the name, the switch was leftover from a previous iteration. KeychainWallet covers these methods pretty well so I don't think they are necessary to test.

satish-ravi (approved):

src/web3/KeychainSigner.ts
```diff
@@ -235,41 +55,15 @@ export class KeychainSigner implements Signer {
...
-
-  /**
-   * Updates the passphrase of an account
-   * @param oldPassphrase - the passphrase currently associated with the account
-   * @param newPassphrase - the new passphrase to use with the account
-   * @returns whether the update was successful
-   */
-  async updatePassphrase(oldPassphrase: string, newPassphrase: string) {
-    const privateKey = await getStoredPrivateKey(this.account, oldPassphrase)
+    const privateKey = await getStoredPrivateKey(this.account, passphrase)
```
jophish, line 62: code golf:
```ts
if (!(unlocked && privateKey)) {
  return false
}
```

src/web3/KeychainLock.ts
```diff
@@ -0,0 +1,254 @@
...
+    {
+      // Timestamp in milliseconds when the keychain was last unlocked
+      unlockTime?: number
+      // Number of seconds that the keychain was last unlocked for
+      unlockDuration?: number
+      account: KeychainAccount
+    }
+  > = new Map()
+
+  addAccount(account: KeychainAccount) {
```
jophish, line 198: If `addAccount` is called multiple times for the same account address and the account was previously unlocked, this will (in effect) lock the account prematurely -- is this a possibility?

src/web3/KeychainLock.ts
```diff
@@ -0,0 +1,254 @@
...
+  isUnlocked(address: string): boolean {
+    if (!this.locks.has(address)) {
+      return false
+    }
+    const { unlockTime, unlockDuration } = this.locks.get(address)!
+    if (unlockDuration === undefined || unlockTime === undefined) {
+      return false
+    }
+
+    if (unlockDuration === 0) {
```
jophish, line 229: Is `unlockDuration === 0` meant to be a special value meaning "unlock indefinitely"? Is this used anywhere/should we allow callers to unlock a signer forever? (And if it is required, should it be documented?)

src/web3/KeychainWallet.ts
```diff
@@ -15,16 +16,24 @@ export class KeychainWallet extends RemoteWallet<KeychainSigner> implements Unlo
...
+  constructor(
+    protected importMnemonicAccount: ImportMnemonicAccount,
+    protected lock: KeychainLock
+  ) {
     super()
   }
 
+  /**
+   * This function does the very critical job of loading in accounts from the Keychain, for instance when the user restarts their app.
+   * TODO: decouple this logic from contractKit and move it into an wallet agnostic place like KeychainLock
```
jophish, line 28:
```suggestion
   * TODO: decouple this logic from contractKit and move it into a wallet agnostic place like KeychainLock
```

src/web3/contracts.ts
```diff
@@ -109,6 +120,32 @@ async function waitForContractKit(tries: number) {
   return contractKit
 }
 
+// This code assumes that the account for walletAddress already exists in the Keychain
+// which is a responsibility currently handled by KeychainWallet
+export function* getViemWallet(chain: Chain) {
```
jophish, line 125: `getViemWallet` appears to require the user's account to be unlocked if this is the first time the Viem wallet is being fetched -- it doesn't appear that this is called during wallet initialization (will it eventually be?), so if there are places in the app that use the wallet but don't immediately require unlocked access, this will "greedily" request the user's PIN, right? I'm not immediately sure of places like that (maybe initializing FC SDKs?) and I imagine we'll eventually initialize Viem wallets along with our regular wallet (guaranteeing that the wallet is already unlocked), so perhaps this is a moot point.

jh2oman: Very good observation. Its correct that this component greedily unlocks. Satish and I were also thinking about places that the we need a wallet without unlocking and couldn't think of any. 

We also found a pretty easy path to decouple this in the future if we ever want to initialize the wallet without unlocking, so I decided to keep it this way for now and not over-engineer. We can always change it later if we cross that bridge

src/web3/KeychainLock.ts
jh2oman: Good point, i added it to the unlock function and this line

src/web3/KeychainLock.ts
jh2oman: It should be impossible from the KeychainSigner/Wallet side because they have checks before calling addAccount to see if the account is already in the keychain. I'll add a the same check here though to make sure that functionality is preserved.

src/web3/KeychainSigner.ts
jh2oman: not quite the same i think. If the account is locked, then i dont want to call the async `getStoredPrivateKey`

src/web3/KeychainLock.ts
jh2oman: Strike that, it seems it is already possible for this to be called twice, similarly it is possible for [addSigner](https://github.com/celo-org/celo-monorepo/blob/0dfd5730c19cb0b96cf05c7cff3e82d2f1f109cd/packages/sdk/wallets/wallet-base/src/wallet-base.ts#L59) to be called twice so I'm keeping the implementation as-is similar to that.

PR #4120, opened 2023-08-24
docs(analytics): compile-time enforcement of analytics docs

cajubelt:
### Description

Thinking of creating a new analytics event without documenting it and leaving your poor teammates in the dark about what it does? Think again! 😈 

I wrote a script that takes all the comments on Events.tsx and populates an object mapping analytics event types to descriptions of what they are. When an analytics event is added that isn't in the object, a type error results, which prevents us from merging the changes. 

Document your analytics events!!!

... does not attempt to backfill docs on all the analytics events we've done this for in the past.

### Test plan

typescript compiles

### Related issues

na

### Backwards compatibility

na

### DIFF
--- a/src/analytics/Events.tsx
+++ b/src/analytics/Events.tsx
@@ -28,7 +28,7 @@ export enum AppEvents {
 }
 
 export enum HomeEvents {
-  home_send = 'home_send',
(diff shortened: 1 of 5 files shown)


silasbw (approved):

jophish (approved):
this is awesome!

PR #4124, opened 2023-08-25
fix: ensure redux selectors are properly memoized

kathaypacific:
### Description

During the RN upgrade I noticed some errors in the test logs `Selector walletConnectEnabledSelector returned a different result when called with the same parameters. This can lead to new ref. Selectors that return a new reference (such as an object or an array) should be memoized: https://redux.js.org/usage/deriving-data-selectors#optimizing-selectors-with-memoization`. This PR fixes the selectors that were complained about.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y


MuckT (approved):
LGTM, tested locally on Android an WC is working as expected!

PR #4125, opened 2023-08-25
fix: add default path for jq in xcode script for m1 xcode users

kathaypacific:
### Description

Context: https://valora-app.slack.com/archives/C025V1D6F3J/p1692696050075289

Xcode seems to not be able to find `jq` in the PATH on m1 macs, this change should affect only m1 users.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y


scripts/xcode_scheme_build_pre_action.sh
```diff
@@ -16,7 +16,9 @@ eval "$(grep -v -e '^#' /tmp/.env-xcode | xargs -I {} echo export \'{}\')"
 
 # Now augment it with network specific "secrets" (not real secrets, just API keys)
 # See https://newbedev.com/how-to-convert-a-json-object-to-key-value-format-in-jq
-jq -r ".${DEFAULT_TESTNET}|to_entries|map(\"\(.key)=\(.value|tostring)\")|.[]" "${SRCROOT}/../secrets.json" >> /tmp/.env-xcode
+# Check if jq exists in the path or use the absolute path
+JQ_CMD=$(command -v jq || echo "/opt/homebrew/bin/jq")
```
silasbw, line 20: i know nothing about mac, but is `/opt/homebrew/bin` the default path to stuff homebrew installed executables?  why isn't it in PATH by default?

silasbw (approved):

scripts/xcode_scheme_build_pre_action.sh
kathaypacific: both @bakoushin and I have added to the PATH, but for some reason xcode does not read from here (e.g. when we build from the command line, there is no problem using `jq` but running the same command in xcode produces problems) - i spent a little time trying to figure out why but didn't find the root cause, and i figure that it's not critical functionality so okay enough to have a quick workaround

PR #4129, opened 2023-08-26
fix(deps): add dotenv as dev dependency for balance and fund task

MuckT:
### Description

The balance and fund task for the e2e test wallet was failing because dotenv was removed as a dev dependency. This PR adds it back; task [failing previously](https://github.com/valora-inc/wallet/actions/workflows/e2e-faucet-balance.yml).

### Test plan

- Tested locally
- Tested in CI

### Related issues

- Fixes [ENG-78](https://linear.app/valora/issue/ENG-78/fix-balance-and-fund-task-for-e2e-tests)

### Backwards compatibility

Yes

### DIFF
--- a/package.json
+++ b/package.json
@@ -251,6 +251,7 @@
     "babel-plugin-module-resolver": "^4.1.0",
     "depcheck": "^1.4.3",
     "detox": "^19.13.0",
+    "dotenv": "^16.3.1",
     "escape-string-regexp": "^1.0.5",
     "eslint": "^7.32.0",
     "eslint-plugin-import": "^2.23.3",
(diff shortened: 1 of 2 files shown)


MuckT:
@kathaypacific as it's only a dev dependency I don't think there's much harm in adding it.

jeanregisser (approved):
👍

kathaypacific:
> @kathaypacific as it's only a dev dependency I don't think there's much harm in adding it.
> 
> > Error: e2e/scripts/fund-e2e-accounts.ts(2,20): error TS7016: Could not find a declaration file for module 'dotenv'. '/home/runner/work/wallet/wallet/node_modules/dotenv/lib/main.js' implicitly has an 'any' type.

i'm not opposed to the general idea but the PR description says that we are putting back something that was wrongly removed. so i wanted to understand how it was wrongly removed / whether it's something we should fix before more dependencies become removed. however i wasn't able to find dotenv as a dependency when the CI job was passing, so i don't think the description is accurate. can we please update the PR description before we merge, since it serves as documentation?

PR #4135, opened 2023-08-28
feat(nav-drawer): upgrade to v6 and enable freeze

MuckT:
### Description

Updates @react-navigation/drawer to v6 and enables freeze for performance improvements. Some prop changes were needed to maintain existing styles and behavior.

Freeze works as follows:
> ...Using Suspense mechanism introduced in React 17. The main use case of this library is to avoid unnecessary re-renders of parts of the app that are not visible to the user at a given moment. The frozen components are not unmounted when they are replaced with a placeholder view, so their React state and corresponding native view instances are retained during the freeze (DOM elements for react-dom or platform-native views for React Native apps) keeping things like scroll position, input state, or loaded images (for  components) unchanged. [source](https://github.com/software-mansion/react-freeze).

| iOS Before | iOS After | Android Before | Android After |
| :---: | :---: | :---: | :---: |
|  | |  |  |

### Test plan

- Tested locally on Android & iOS

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/src/navigator/DrawerNavigator.tsx
+++ b/src/navigator/DrawerNavigator.tsx
@@ -1,7 +1,6 @@
 import {
   createDrawerNavigator,
   DrawerContentComponentProps,
-  DrawerContentOptions,
   DrawerContentScrollView,
 } from '@react-navigation/drawer'
 import {
@@ -80,7 +79,10 @@ const TAG = 'NavigationService'
 
 const Drawer = createDrawerNavigator()
 
-type CustomDrawerItemListProps = Omit<DrawerContentOptions, 'contentContainerStyle' | 'style'> & {
+type CustomDrawerItemListProps = Omit<
+  DrawerContentComponentProps,
+  'contentContainerStyle' | 'style'
+> & {
   state: DrawerNavigationState<ParamListBase>
   navigation: DrawerNavigationHelpers
   descriptors: DrawerDescriptorMap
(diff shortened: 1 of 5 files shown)


src/qrcode/QRScanner.tsx
```diff
@@ -69,6 +69,7 @@ export default function QRScanner({ onBarCodeDetected }: QRScannerProps) {
   }
 
   const submitModal = () => {
+    Keyboard.dismiss()
```
dievazqu, line 72: Is this change related to the freeze feature?

src/navigator/DrawerNavigator.tsx
```diff
@@ -149,15 +150,19 @@ function CustomDrawerItemList({
         label={drawerLabel !== undefined ? drawerLabel : title !== undefined ? title : route.name}
         icon={drawerIcon}
         focused={focused}
-        style={itemStyle}
+        labelStyle={[
+          fontStyles.regular,
+          { color: colors.dark, marginLeft: -20, fontWeight: 'normal' },
```
dievazqu, line 155: Should we move this style somewhere else?

dievazqu (approved):
LGTM

src/qrcode/QRScanner.tsx
MuckT: This is there was a failing e2e test in CI where the keyboard was not dismissed. I was unable to reproduce locally and this modal is only used in e2e tests so it seems fine to have this dismiss here.

PR #4136, opened 2023-08-28
chore(analytics): add analytics to select providers screen

cajubelt:
### Description

add information for figuring out how often and why few/no quotes appear for users trying to cash in

### Test plan

jest tests

### Related issues

- Fixes https://linear.app/valora/issue/ACT-900/select-provider-screen-additional-analytics

### Backwards compatibility

na

### DIFF
--- a/src/analytics/Properties.tsx
+++ b/src/analytics/Properties.tsx
@@ -538,6 +538,7 @@ interface EscrowEventsProperties {
     error: string
   }
 }
+
 interface SendEventsProperties {
   [SendEvents.send_scan]: undefined
   [SendEvents.send_select_recipient]: {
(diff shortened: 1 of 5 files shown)


src/fiatExchanges/SelectProvider.tsx
cajubelt, line 1: the diff on this file is very confusing, but the changes are simple:
- a new useEffect is added that publishes an analytics event once quotes are loaded
- a some analytics functions are hoisted up higher in the file to be usable in the useEffect hook, which needed to come before the `if (quotesLoading) { return spinner stuff } ` part to play by the rules of hooks

src/analytics/Properties.tsx
```diff
@@ -834,6 +835,22 @@ interface FiatExchangeEventsProperties {
...
   [FiatExchangeEvents.cico_providers_back]: { flow: CICOFlow }
+  [FiatExchangeEvents.cico_providers_fetch_quotes_result]: {
+    fiatType: LocalCurrencyCode
+    defaultFiatType: LocalCurrencyCode
+  } & Omit<ProviderSelectionAnalyticsData, 'transferCryptoAmount'> &
+    (
+      | {
+          flow: CICOFlow.CashOut
+          cryptoAmount: number // given by user
+          fiatAmount: undefined // exchange rate varies by provider
```
satish-ravi, line 846: Any reason to include an explicit undefined?

satish-ravi (approved):

src/analytics/Properties.tsx
cajubelt: yeah, otherwise I found that if I left the amount in it still passed the type check. it's misleading to consumers of the data to have a non null fiatAmount for cash outs (and cryptoAmount for cash ins) because in reality, that amount varies by provider. so any query that uses it would be buggy

PR #4138, opened 2023-08-29
chore(notifications): group tests by notification

bakoushin:
### Description

The proposal is to group tests by the notification they are testing for better readability / file navigation. Currently [NotificationCenter.test.tsx](https://github.com/valora-inc/wallet/blob/main/src/home/NotificationCenter.test.tsx) contains a flat list of 19 test cases for different notifications, and more to come. Better to split them into smaller meaningful groups.

| Before | After |
|--------|--------|
| Screenshot 2023-08-29 at 13 15 23 | Screenshot 2023-08-29 at 13 14 57 | 

Nothing is changed besides wrapping test cases into groups.

### Test plan

N/A

### Related issues

- Prepares for RET-801

### Backwards compatibility

Yes

### DIFF
--- a/src/home/NotificationCenter.test.tsx
+++ b/src/home/NotificationCenter.test.tsx
@@ -122,488 +122,506 @@ describe('NotificationCenter', () => {
     expect(getByText('noNotificationsPlaceholder')).toBeTruthy()
   })
 
-  it('renders reverify notification if decentrally verified and not CPV', () => {
-    const store = createMockStore({
-      app: {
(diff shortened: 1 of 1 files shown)


kathaypacific (approved):
LGTM! not sure what your plans are for the new tests but i wonder if the `describe`s should be less granular, e.g. `describe('supercharge notifications')` rather than describe`('specific supercharge scenario')`, but i'm also fine with how this is now

PR #4139, opened 2023-08-29
chore(notifications): add new analytics events (part 1)

bakoushin:
### Description

1. When the Notification Center is opened, a new analytics event `notification_center_opened` is dispatched, with a number of notifications currently visible to the user (`notificationsCount` field).

2. Analytics events for bundled notifications (e.g. supercharge) are dispatched with a new type `bundled_notification`. What was formerly in the `notificationType` field is now placed in the `notificationId` field to match remote notifications.

3. Impression analytics event with `notificationPosition` on screen is dispatched when a notification becomes visible for the first time. Notification is considered visible when it fully fits into the viewport within 0.5 seconds (to avoid dispatching impressions while user just scrolls really fast).

### Test plan

1. The first two changes are ensured by enhanced unit tests.
2. For the third change I was not able to make a unit test, because the event handler [`onViewableItemsChanged`](https://reactnative.dev/docs/0.70/flatlist#onviewableitemschanged) does not fire during test execution for some reason. I've opened the [issue in RN repo](https://github.com/facebook/react-native/issues/39191) seeking a feedback. Tested manually on iOS and Android emulators.

### Related issues

- Partially fixes RET-801

### Backwards compatibility

Not compatible, because it alters the fieldset of analytics events for bundled notifications.


src/home/NotificationCenter.test.tsx
```diff
@@ -623,5 +907,59 @@ describe('NotificationCenter', () => {
...
+          <NotificationCenter {...getMockStackScreenProps(Screens.NotificationCenter)} />
+        </Provider>
+      )
+
+      fireEvent.press(getByTestId('startSupercharging/CallToActions/dismiss/Button'))
+
+      expect(ValoraAnalytics.track).toHaveBeenLastCalledWith(HomeEvents.notification_select, {
+        notificationType: NotificationBannerTypes.bundled_notificaion,
+        selectedAction: NotificationBannerCTATypes.decline,
+        notificationId: BundledNotificationIds.supercharging,
```
MuckT, line 961: Should this be `BundledNotificationIds.start_supercharging`?

MuckT (approved):
Just one comment on the failing test. Otherwise this looks great. It might have been nice for to use it.each where possible, but I noticed the store is sufficiently different between the tests that this might have been more complicated and less easy to debug.

bakoushin:
Thank you for suggesting `it.each`! I am tempted to try it out, as you have noticed some repetition in the tests I added. However, this would change the way the tests are organized, from grouping them by notification type to grouping them by test type. Since the notifications are not completely generic, and many of them have their peculiarities, I would prefer to keep them grouped by notification type for now.

PR #4140, opened 2023-08-29
chore: update dapp open analytics props

kathaypacific:
### Description

As the title

### Test plan

Tested manually + unit tests

### Related issues

- Fixes RET-822

### Backwards compatibility

Y


src/dappsExplorer/useOpenDapp.tsx
```diff
@@ -22,6 +22,8 @@ const useOpenDapp = () => {
   const [selectedDapp, setSelectedDapp] = useState<ActiveDapp | null>(null)
   const dispatch = useDispatch()
 
+  const extraAnalyticsPropertiesRef = useRef<Record<string, string | number>>({})
```
kathaypacific, line 25: this is a bit hacky but i've left it here because i think this hook can be removed soon if we remove the confirmation bottom sheet. i haven't been able to confirm this with Ellen yet so i've left it here but i haven't optimised the solution for this bottom sheet for sure...

src/analytics/Properties.tsx
```diff
@@ -1189,7 +1189,11 @@ interface DappEventProperties extends DappProperties {
 
 interface DappExplorerEventsProperties {
   [DappExplorerEvents.dapp_impression]: DappEventProperties
-  [DappExplorerEvents.dapp_open]: DappEventProperties
+  [DappExplorerEvents.dapp_open]: DappEventProperties & {
+    activeFilter?: string
+    activeSearchTerm?: string
+    position?: number
```
kathaypacific, line 1195: i played around with various typescript types for this, to see if we can get static typechecking (e.g. for certain dapp sections, these 3 props are required but for other dapp sections, they are optional). it became complex and brittle, so i've settled for unit tests to cover the cases we care about

src/dappsExplorer/DAppsExplorerScreenSearchFilter.tsx
```diff
@@ -126,6 +126,14 @@ export function DAppsExplorerScreenSearchFilter() {
     ValoraAnalytics.track(DappExplorerEvents.dapp_screen_open)
   }, [])
 
+  const createOnPressDappHandler = (dapp: ActiveDapp, index: number) => () => {
+    onSelectDapp(dapp, {
+      position: 1 + index,
```
kathaypacific, line 131: i checked with Ellen if we should take into account the number of favorites when reporting the dapp's position https://linear.app/valora/issue/RET-822#comment-2ed0c83e for now, the answer is no because we care about the order to learn about our recommendations. the number / order of favorites that a user curates is not directly relevant here. would be nice to report that number too but that requires us to refactor the favourite search / filter results to be in this component rather than the Favorites component, but we also don't have a use case for that data yet so i haven't done it here

src/dappsExplorer/DAppsExplorerScreenSearchFilter.tsx
```diff
@@ -249,7 +257,7 @@ export function DAppsExplorerScreenSearchFilter() {
                           : t('dappsScreen.favoriteDapps').toLocaleUpperCase(language ?? 'en-US')}
                       </Text>
                       <FavoriteDappsSection
-                        onPressDapp={onSelectDapp}
+                        onPressDapp={createOnPressDappHandler}
```
jeanregisser, line 260: This feels wrong, we're not creating the handler here with this.

src/dappsExplorer/DAppsExplorerScreenSearchFilter.tsx
```diff
@@ -296,7 +306,7 @@ export function DAppsExplorerScreenSearchFilter() {
       {DappFavoritedToast}
       <DappRankingsBottomSheet
         forwardedRef={dappRankingsBottomSheetRef}
-        onPressDapp={onSelectDapp}
+        onPressDapp={createOnPressDappHandler}
```
jeanregisser, line 309: Same here. Did you mean something like this?
```suggestion
        onPressDapp={createOnPressDappHandler(...)}
```

src/dappsExplorer/DappRankingsBottomSheet.tsx
```diff
@@ -16,15 +16,11 @@ export function DappRankingsBottomSheet({
   onPressDapp,
 }: {
   forwardedRef: RefObject<BottomSheetRefType>
-  onPressDapp: (dapp: ActiveDapp) => void
+  onPressDapp: (dapp: ActiveDapp, index: number) => () => void
```
jeanregisser, line 19: Oh, now I understand, but it feels like too much indirection and surprising.
the `onXXX` props are usually just functions.

Can we do differently? and avoid this surprise.

src/dappsExplorer/FavoriteDappsSection.tsx
```diff
@@ -8,7 +8,7 @@ import { searchDappList } from 'src/dappsExplorer/searchDappList'
 
 interface Props {
   filterId: string
-  onPressDapp: (dapp: ActiveDapp) => void
+  onPressDapp: (dapp: ActiveDapp, index: number) => () => void
```
jeanregisser, line 11: Same here.

src/dappsExplorer/useOpenDapp.tsx
```diff
@@ -44,8 +46,14 @@ const useOpenDapp = () => {
     }
   }
 
-  const openDapp = (dapp: ActiveDapp) => {
-    ValoraAnalytics.track(DappExplorerEvents.dapp_open, getEventProperties(dapp))
+  const openDapp = (
+    dapp: ActiveDapp,
+    extraAnalyticsProperties: Record<string, string | number> = {}
```
jeanregisser, line 51: I saw your comment about having static types above and making it complex and brittle, but it feels we could at least have the 3 optional properties we want here.

src/dappsExplorer/useOpenDapp.tsx
jeanregisser: Ah another reason this "return component from hook" pattern feels off to me if we have to resort to this 😅

jeanregisser (commented):
Thanks!

Couple of comments on how it was implemented, but nice to see this.

jeanregisser (approved):

PR #4149, opened 2023-08-31
fix(assets-list): segmented control styles

MuckT:
### Description

Fixes the ripple overflow on Android and button styles. Also makes the minHeight the device height + the header to always enable the scrollable header.

#### Android Before

https://github.com/valora-inc/wallet/assets/26950305/b1c9c42a-cfa0-4a5e-b379-0f3696c3f492

#### Android After

https://github.com/valora-inc/wallet/assets/26950305/b9be624a-0372-4c9e-8c17-c13f9f46a6ac

### Test plan

Tested locally on Android & iOS

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/src/tokens/SegmentedControl.tsx
+++ b/src/tokens/SegmentedControl.tsx
@@ -26,6 +26,9 @@ function SegmentedControl({ values, selectedIndex, onChange }: Props) {
           style={[
             styles.button,
             { backgroundColor: index === selectedIndex ? Colors.greenUI : Colors.gray1 },
+            index === 0 && { borderBottomEndRadius: 0, borderTopEndRadius: 0 },
+            index === values.length - 1 && { borderBottomStartRadius: 0, borderTopStartRadius: 0 },
+            index !== 0 && index !== values.length - 1 && { borderRadius: 0 },
           ]}
           onPress={handleSelectOption(value, index)}
         >
@@ -42,6 +45,7 @@ function SegmentedControl({ values, selectedIndex, onChange }: Props) {
 
 const styles = StyleSheet.create({
   container: {
+    overflow: 'hidden',
     flexDirection: 'row',
     borderRadius: 100,
     backgroundColor: Colors.gray1,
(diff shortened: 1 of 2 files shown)


dievazqu (approved):

src/tokens/TokenBalances.tsx
```diff
@@ -342,6 +343,8 @@ function TokenBalancesScreen({ navigation, route }: Props) {
         contentContainerStyle={{
           paddingBottom: insets.bottom,
           opacity: listHeaderHeight > 0 ? 1 : 0,
+          // Allow header scroll even when not enough assets or positions
+          minHeight: variables.height + listHeaderHeight,
```
jeanregisser, line 347: Why do we need header scroll when everything is visible?

src/tokens/SegmentedControl.tsx
```diff
@@ -26,6 +26,11 @@ function SegmentedControl({ values, selectedIndex, onChange }: Props) {
           style={[
             styles.button,
             { backgroundColor: index === selectedIndex ? Colors.greenUI : Colors.gray1 },
+            // Round the left and right sides of the first and last buttons respectively
+            index === 0 && { borderBottomEndRadius: 0, borderTopEndRadius: 0 },
+            index === values.length - 1 && { borderBottomStartRadius: 0, borderTopStartRadius: 0 },
+            // Square corners, no border radius, for middle buttons
+            index !== 0 && index !== values.length - 1 && { borderRadius: 0 },
```
jeanregisser, line 33: This looks different from previously. Did you check with @nityarao that it's what we want?

src/tokens/TokenBalances.tsx
MuckT: This was more of a personal preference, but I agree it's not needed. Reverted in https://github.com/valora-inc/wallet/pull/4149/commits/7b161d7c1627b7c60d4f1cc4529dc2e761b2c6a2!

PR #4151, opened 2023-09-01
build(deps): polyfill big integer

MuckT:
### Description

Polyfills big integer needed for Viem / Ethers v6 due to lack of native BigInt support on iOS 12 and 13.

From Viem README:
> Browser native [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt), instead of large BigNumber libraries

### Test plan

Tested locally on iOS 13.7 with 1.65 release.

### Related issues

- Fixes ACT-888

### Backwards compatibility

Yes


MuckT:
It looks like hermes and BigInt might not play well together.

MuckT:
This is still failing locally on iOS 12 & 13.

PR #4153, opened 2023-09-04
chore(notifications): track notification positions in the list

bakoushin:
### Description

When user taps on notification CTA's, the position of the notification on the whole list is included in analytics event.

```
{
   notificationType: 'bundled_notification',
   selectedAction: 'accept',
   notificationId: 'start_supercharging',
   notificationPositionInList: 0 // <--- index of the notification in the list
}
```

### Test plan

* Updated unit tests
* Tested manually on iOS emulator

### Related issues

- Fixes RET-801

### Backwards compatibility

Yes


src/home/NotificationCenter.tsx
```diff
@@ -223,6 +258,8 @@ export default function Notifications({ navigation }: NotificationsProps) {
   )
 }
 
+export const listGapHeight = Spacing.Thick24
```
jeanregisser, line 261: Looks like this export is only needed for testing. 

Could you change it to:
```suggestion
// Exported for testing
export const _listGapHeight = Spacing.Thick24
```

jeanregisser (commented):
Oh this is quite clever to reuse the positions from `onViewableItemsChanged` and inject them using `Context`.

However I'm not entirely convinced it's worth the added complexity for adding the position on click.
Was there any problem with passing a `position` prop to each rendered item instead?
This feels more direct and could simplify the unit tests a bit.

But let me know if I missed something here 🙏

src/components/MessagingCard.tsx
```diff
@@ -3,14 +3,16 @@ import { StyleSheet } from 'react-native'
 import Card, { Props } from 'src/components/Card'
 import Colors from 'src/styles/colors'
 
+export const minHeight = 144
```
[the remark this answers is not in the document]
bakoushin: yes, we can.
it's used in the unit test for the NotifcationCenter, the same as the `listGapHeight` you've mentioned in another comment.
my initial intent with these exports was to use the original height/gap values during layout simulation in a test.
however, it is not strictly necessary as long as measurements/expectations within the test remain consistent.

src/home/NotificationCenter.tsx
bakoushin: I'm inclined to revert this as well. Instead, I'll provide heights for simulation directly in the test.

bakoushin:
> I'm not entirely convinced it's worth the added complexity for adding the position on click.
> Was there any problem with passing a position prop to each rendered item instead?

I agree. It happens to be kind of an overengineering. Thanks for pointing this out!
I wanted to avoid probably too deep props drilling. But it turns out it's not the case.

I've refactored the changes to use props instead of context.

src/home/NotificationCenter.tsx
```diff
@@ -161,6 +185,11 @@ export default function Notifications({ navigation }: NotificationsProps) {
   const handleViewableItemsChanged = useCallback(
     ({ viewableItems }: { viewableItems: ViewToken[] }) => {
       viewableItems.forEach(({ item }, index) => {
+        setNotificationPositions((prevNotificationPositions) => ({
+          ...prevNotificationPositions,
+          [item.id]: index,
+        }))
+
```
jeanregisser, line 192: I think we could avoid the extra state and the call to `setNotificationPositions` by directly consuming the `index` argument from the FlatList [renderItem](https://reactnative.dev/docs/flatlist#required-renderitem) method.

Added benefit is it avoids the re-render when items become visible.

But we'll have to wrap the `element` in a function call to pass the index.

What do you think?

bakoushin: I'm doubtful. The `renderItem` receives the index within the *whole* data array, while we want to get the index only within currently *visible* items. It's because we want to know where the item was on screen when an interaction happened. But maybe I'm missing something?

jeanregisser: oh I see, yeah we need to agree on what this position is 👍 

To me it was the position from the whole list, not just on screen.
We have several similar instances of tracking positions, and so far they were the positions from the whole data.

For instance: 
https://github.com/valora-inc/wallet/blob/fc46ae9cc2641545d15d9321b59f80076f9edaac/src/analytics/Properties.tsx#L1197
https://github.com/valora-inc/wallet/blob/fc46ae9cc2641545d15d9321b59f80076f9edaac/src/analytics/Properties.tsx#L1189

Seeing https://github.com/valora-inc/wallet/blob/fc46ae9cc2641545d15d9321b59f80076f9edaac/src/home/NotificationCenter.tsx#L162-L173
To me this should have been the position from the whole list too.

I guess it depends what question we want to answer.
We should check with @ellenwoconnor.

In all cases we should make sure this isn't ambiguous. Maybe finding names that convey this more clearly?

kathaypacific: ah sorry i'm late to the party. i think the position we want to track is relative to the whole data - we want to learn if users are scrolling to the bottom of the list, rather than which position on the screen is most effective :)

bakoushin:
I've updated the PR based on feedback received:
* in both types of events (`notification_impression` and `notification_select`) we send the index of the notification in **the whole rendered list** (instead of the position in the visible area)
* analytics property for position is named `notificationPositionInList` to convey more clear what kind of value it holds

bakoushin:
I refrained from changing property names in dapps analytics events @jeanregisser mentioned earlier (`position`, `horizontalPosition`) because that would be a breaking change and I'm not sure that just naming is a valid reason for it.

src/analytics/docs.ts
```diff
@@ -75,8 +75,8 @@ export const eventDocs: Record<AnalyticsEventType, string> = {
   [HomeEvents.drawer_navigation]: ``,
   [HomeEvents.drawer_address_copy]: ``,
   [HomeEvents.notification_scroll]: ``,
-  [HomeEvents.notification_impression]: ``,
-  [HomeEvents.notification_select]: ``,
+  [HomeEvents.notification_impression]: `when the notification appears on the user screen for the first time`,
+  [HomeEvents.notification_select]: `when user taps on notification button`,
```
kathaypacific, line 79:
```suggestion
  [HomeEvents.notification_select]: `when user taps on notification CTA(s)`,
```

src/components/CallToActionsBar.tsx
```diff
@@ -13,9 +13,10 @@ export interface CallToAction {
 interface Props {
   callToActions: CallToAction[]
   testID?: string
+  index?: number
```
kathaypacific, line 16: what do you think of `positionInNotificationList` instead of `index` so that it's easier to understand what the index is relative to?

src/home/NotificationCenter.tsx
```diff
@@ -134,6 +141,12 @@ export default function Notifications({ navigation }: NotificationsProps) {
 
   const notifications = useNotifications()
 
+  // Changing onViewableItemsChanged on the fly is not supported.
+  // This is a workaround to provide handleViewableItemsChanged with
+  // actual notifications array while keeping its dependecy list empty.
+  const notificationsRef = useRef<Notification[]>([])
+  notificationsRef.current = notifications
```
kathaypacific, line 148: maybe we should have a `useEffect` here that updates the ref when `notifications` changes? i think notifications can potentially change while the user is on the screen because the app talks to firebase rtdb directly

src/home/NotificationCenter.test.tsx
```diff
@@ -31,6 +32,10 @@ jest.mock('src/navigator/NavigationService', () => ({
   navigate: jest.fn(),
 }))
 
+jest.useFakeTimers()
```
kathaypacific, line 35: nit: fake timers are enabled by default in jest_config.ts so no need to declare it here

src/home/NotificationCenter.test.tsx
```diff
@@ -122,9 +128,39 @@ const mockcUsdWithoutEnoughBalance = {
   },
 }
 
+const layoutNotificationList = (screen: ReturnType<typeof render>) => {
```
kathaypacific, line 131: this is great!! could we add a comment to remind us that this function is needed to get the onViewableItems callback to fire? 🙏🏻

src/paymentRequest/IncomingPaymentRequestListItem.test.tsx
```diff
@@ -326,4 +332,54 @@ describe('IncomingPaymentRequestListItem', () => {
...
+    const { getByTestId } = render(
+      <Provider store={store}>
+        <IncomingPaymentRequestListItem {...props} />
+      </Provider>
+    )
+
+    const remindButton = getByTestId(
+      'IncomingPaymentRequestNotification/FAKE_ID_2/CallToActions/send/Button'
+    )
+    fireEvent.press(remindButton)
```
kathaypacific, line 351: i generally prefer using `getByText` rather than `getByTestId` wherever possible because it tests closer to what a user would see. this [article](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-querying-by-text) resonated with me, but of course it is just an opinion and we haven't been very consistent about this perspective in our projects

kathaypacific (approved):
nice!! i left a few superficial comments that i don't feel strongly about

src/home/NotificationCenter.tsx
bakoushin: That way we save the `notifications` value in a ref every time a component re-renders. It does that in particular when rtdb will send an update and the `useNotifications()` hook will be triggered with a new value for `notifications`. 

Wrapping ref assignment into a `useEffect` will scope ref update only to re-renders caused by `notifications` change. Looks like that way it makes the initial intent more clear, I will refactor it to `useEffect`. 

Thanks for the suggestion!

PR #4156, opened 2023-09-05
refactor: rename `localCurrencyToUsdSelector` -> `usdToLocalCurrencyRateSelector`

jeanregisser:
### Description

Previous name was a bit misleading.

Well, it's still misleading cause right now it's actually cUSD to local currency. But this is gonna change soon.

### Test plan

Unit tests pass

### Related issues

- Fixes RET-830

### Backwards compatibility

Yes


dievazqu:
Just to clarify, previously we were using that value correctly:
To convert usd to local currency we were multiplying the usd value for the `localCurrencyExchangeRate`.
The `localCurrencyToUsdSelector` name was misleading because it seemed that we had to divide instead of multiply, right?

dievazqu (approved):
LGTM!

PR #4157, opened 2023-09-05
feat(eth): Enable ETH in the Add/Withdraw flow

jophish:
### Description

For ACT-893. Enables ETH in the CICO flow's "currency select" screen, locked behind a feature gate (see in [Statsig](https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/gates/show_eth_in_cico)). It's a bit hacky still since we don't have information on the ETH asset in-app, causing the rendering to be messed up when listing ETH providers.

### Test plan

Unit and manual tested. See video below. A few notes:
* I've always had trouble getting the Ramp widget website to render correctly in my simulator's browser; copying the link to my desktop browser however, it renders with ETH pre-selected, as expected (see [here](https://app.ramp.network/?hostApiKey=zop7c2gfzpj8f3e68aqts8bp3mvj2mzyqpg2z7pk&userAddress=0xe192673ad298ff0b72dd1c8e074d34ddaafd9591&swapAsset=ETH_ETH&hostAppName=Valora&hostLogoUrl=https://storage.googleapis.com/celo-mobile-mainnet.appspot.com/images/valora-icon.png&fiatCurrency=USD&fiatValue=15&finalUrl=celo%3A%2F%2Fwallet%2Fcash-in-success%2FRamp&webhookStatusUrl=https://us-central1-celo-mobile-mainnet.cloudfunctions.net/rampWebhook&offrampWebhookV3Url=https://us-central1-celo-mobile-mainnet.cloudfunctions.net/rampWebhook&enabledFlows=ONRAMP,OFFRAMP&defaultFlow=ONRAMP))
* The `fetchProviders` URL that this is operating against is _mainnet_, not alfajores; Ramp does not return an ETH quote from their sandbox API in my region, for whatever reason, but does return one from their production API.

https://github.com/valora-inc/wallet/assets/569401/ae11d128-1b4a-437c-b045-f9d0df89ea60

### Related issues

- Fixes ACT-892.

### Backwards compatibility

Yes.


src/fiatExchanges/SelectProvider.tsx
```diff
@@ -187,6 +190,7 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
     asyncExchanges.loading ||
     selectFiatConnectQuoteLoading
 
+  Logger.info(TAG, asyncProviders.result?.externalProviders)
```
jh2oman, line 193: leftover from debugging? 

Could keep it, but should probably have a little extra context in the log.

src/fiatExchanges/FiatExchangeAmount.test.tsx
```diff
@@ -149,6 +153,23 @@ describe('FiatExchangeAmount cashIn', () => {
...
+      <Provider store={storeWithUSD}>
+        <FiatExchangeAmount {...mockScreenProps} />
+      </Provider>
+    )
+
+    fireEvent.changeText(tree.getByTestId('FiatExchangeInput'), '0')
+    expect(tree.getByTestId('FiatExchangeNextButton')).toBeDisabled()
+  })
+
+  it('disables the next button if the ETH amount is 0', () => {
```
cajubelt, line 168: seems like we should do this with an it.each since it largely copies some similar tests

src/fiatExchanges/FiatExchangeAmount.test.tsx
```diff
@@ -189,22 +212,41 @@ describe('FiatExchangeAmount cashIn', () => {
     fireEvent.changeText(tree.getByTestId('FiatExchangeInput'), '5')
     expect(tree.getByTestId('FiatExchangeNextButton')).not.toBeDisabled()
   })
+
+  it('enables the next button if the ETH amount is greater than 0', () => {
```
cajubelt, line 216: same here

src/fiatExchanges/FiatExchangeAmount.tsx
```diff
@@ -70,15 +71,15 @@ type Props = RouteProps
...
-  const { currency, flow } = route.params
+  const { currency, flow, network } = route.params
 
   const [showingInvalidAmountDialog, setShowingInvalidAmountDialog] = useState(false)
   const closeInvalidAmountDialog = () => {
     setShowingInvalidAmountDialog(false)
   }
   const [inputAmount, setInputAmount] = useState('')
   const parsedInputAmount = parseInputAmount(inputAmount, decimalSeparator)
-  const { address } = useTokenInfoBySymbol(currency)!
```
cajubelt, line 81: down with non-null assertion operators!

src/fiatExchanges/FiatExchangeAmount.tsx
```diff
@@ -122,7 +123,7 @@ function FiatExchangeAmount({ route }: Props) {
     dispatch(fetchExchangeRate())
   }, [])
 
-  if (!address) {
+  if (!address && currency !== CiCoCurrency.ETH) {
```
cajubelt, line 126: seems like this might need a todo to remove it right? once we have eth prices available to us?

src/fiatExchanges/FiatExchangeCurrency.tsx
```diff
@@ -114,6 +118,7 @@ function FiatExchangeCurrency({ route, navigation }: Props) {
     navigate(Screens.FiatExchangeAmount, {
       currency: selectedCurrency,
       flow: flow === FiatExchangeFlow.CashIn ? CICOFlow.CashIn : CICOFlow.CashOut,
+      network: selectedCurrency === CiCoCurrency.ETH ? Network.Ethereum : Network.Celo,
```
cajubelt, line 121: a more bug safe way to do this might be to create a mapping from cico currency to network and use type `Record<CiCoCurrency, Network>`, so we get a type error if we try to add a cicocurrency but forget to specify the correct network for this line

src/fiatExchanges/amount.tsx
```diff
@@ -14,7 +14,7 @@ export function CryptoAmount({
   currency: CiCoCurrency
   testID?: string
 }) {
-  const { address } = useTokenInfoBySymbol(currency)!
+  const { address } = useTokenInfoBySymbol(currency) || {}
```
cajubelt, line 17:
```suggestion
  const { address } = useTokenInfoBySymbol(currency) ?? {}
```

src/fiatconnect/consts.ts
```diff
@@ -34,7 +34,7 @@ export const FIATCONNECT_CURRENCY_TO_WALLET_CURRENCY: Record<FiatType, LocalCurr
   [FiatType.ZAR]: LocalCurrencyCode.ZAR,
 }
 
-export const WALLET_CRYPTO_TO_FIATCONNECT_CRYPTO: Record<CiCoCurrency, CryptoType> = {
+export const WALLET_CRYPTO_TO_FIATCONNECT_CRYPTO: Partial<Record<CiCoCurrency, CryptoType>> = {
```
cajubelt, line 37:
```suggestion
export const WALLET_CRYPTO_TO_FIATCONNECT_CRYPTO: Record<CiCoCurrency, CryptoType | null> = {
```

this way, we are still forced to consider whether a mapping should be added, but we can add "null" if no sensible mapping exists

src/fiatExchanges/FiatExchangeCurrency.tsx
jophish: Ah yeah, I like that idea -- added it in!

cajubelt (approved):

src/fiatExchanges/FiatExchangeAmount.test.tsx
```diff
@@ -103,41 +104,20 @@ describe('FiatExchangeAmount cashIn', () => {
...
-    expect(tree).toMatchSnapshot()
-  })
-
-  it('renders correctly with EUR as app currency', () => {
-    const mockScreenProps = getMockStackScreenProps(Screens.FiatExchangeAmount, {
-      currency: CiCoCurrency.cUSD,
-      flow: CICOFlow.CashIn,
-      network: Network.Celo,
+    jest.mocked(getFeatureGate).mockImplementation((gate) => {
+      return gate === StatsigFeatureGates.SHOW_NATIVE_TOKENS ? true : false
```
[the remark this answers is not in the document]
jophish: wow, this is one of the larger facepalms i've had in a while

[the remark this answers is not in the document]
jophish: perhaps something like this could work instead...

```suggestion
      return (gate === StatsigFeatureGates.SHOW_NATIVE_TOKENS ? (true ? true : false) : (false ? true : false))
```

PR #4162, opened 2023-09-06
feat: fetch and store USD to local currency rate

jeanregisser:
### Description

As the title says.

We were previously fetching and storing multiple rates, but only really using the cUSD to local.
blockchain-api was actually giving us the rate for USD to local, since it was assuming 1 cUSD = 1 USD.

The client behaves basically the same with regard to displaying amounts in the local currency. but how it does it is more correct, once we'll update blockchain-api to stop assuming 1 cUSD = 1 USD in the currency conversion code there.

### Test plan

- Unit tests pass

### Related issues

- Fixes RET-830

### Backwards compatibility

Yes.


src/localCurrency/saga.ts
```diff
@@ -54,23 +53,8 @@ export function* fetchLocalCurrencyRateSaga() {
...
-        {
-          [Currency.Dollar]: usdRate,
-          [Currency.Euro]: euroRate,
-          [Currency.Celo]: celoRate,
-        },
-        Date.now()
-      )
-    )
+    const usdToLocalRate = yield* call(fetchExchangeRate, LocalCurrencyCode.USD, localCurrencyCode)
+    yield* put(fetchCurrentRateSuccess(localCurrencyCode, usdToLocalRate, Date.now()))
```
jeanregisser, line 57: This is the changed fetch part.

src/localCurrency/reducer.ts
```diff
@@ -1,25 +1,20 @@
 import { Actions, ActionTypes } from 'src/localCurrency/actions'
 import { LocalCurrencyCode } from 'src/localCurrency/consts'
 import { getRehydratePayload, REHYDRATE, RehydrateAction } from 'src/redux/persist-helper'
-import { Currency } from 'src/utils/currencies'
 
 export interface State {
   isLoading: boolean
   error?: boolean
   preferredCurrencyCode?: LocalCurrencyCode
-  exchangeRates: { [token in Currency]: string | null }
```
jeanregisser, line 10: This is the changed storing part.

dievazqu (approved):
LGTM!

Issue #364, opened 2021-05-13
Added profile picture gets removed when user upgrading the build to V1.3.1 to V1.14.1


ValoraQA:
**Frequency:** 100%
**Repro on build version:** Test Flight build V1.14.0(58), Test Flight build V1.14.1(59),Test Flight build V1.13.1(57), , 
**Repro on devices:** iPhone SE (14.2),  iPhone 6+(12.4.5)
**Does not Repro on build version:** Android internal build V1.14.1
**Does not Repro on devices:** Samsung M30s (10), Vivo V11 pro (10)

**Testing Account:**  (680) 333 1978 / [Backup Key](https://docs.google.com/spreadsheets/d/1gXY1I4PBkT9f1uB3zxWgUUJJ1zjVeJcqcWLjiIls1f0/edit#gid=0&range=B9) 
**Pre-condition:** User should update the build to V1.3.1 to V1.14.1

  **Repro Steps:**

1. User should install & added profile picture in V1.13.1 build
2. Update it to V1.14.1
3. Observe profile picture gets removed

**Current Behavior:**  Added profile picture gets removed .
**Expected Behavior:** Users should be able to see the added profile picture after upgrading the build. 
 **Impact:** Bad user experience & USer get blocked to see profile picture.  
**Attachment:**  [ProfilePictureGetsRemoved.MP4](https://drive.google.com/file/d/1Rk78DaiJnFmoFKMXnzTRu6TNVwhHbD2h/view?usp=sharing)

Celoqa:
https://valora.testquality.com defect [D24](https://valora.testquality.com/project/12128/result/2413788/defect/7867).

Celoqa:
https://valora.testquality.com defect [D37](https://valora.testquality.com/project/12128/result/2418451/defect/7883).

ValoraQA:
https://valora.testquality.com defect [D127](https://valora.testquality.com/project/12128/result/2471316/defect/8213).

ValoraQA:
https://valora.testquality.com defect [D898](https://valora.testquality.com/project/12128/result/2613155/defect/9710).

PR #4163, opened 2023-09-06
docs: remove react-celo and update walletconnect v2 link

MuckT:
### Description

Removes outdated recommendation for dapps to use react-celo which is now end of life and instead recommends @celo/rainbowkit-celo and web3modal. Also replaces a dead link for WalletConnect Docs.

### Test plan

N/A

### Related issues

[react-celo end-of-life (EOL) process #364
](https://github.com/celo-org/react-celo/issues/364)

### Backwards compatibility

Yes


jeanregisser (approved):
💯

PR #4167, opened 2023-09-08
feat(viem): implement sendPayment

satish-ravi:
### Description

Part 1 of implementing sendPayment with viem. Includes fetching fee using a wrapper method to the existing chooseTxFeeDetails, creating the appropriate `transfer` contract method based on whether the token is a stable token or not and then simulating and then executing the method. Adding standby tx, waiting for receipt and confirming tx will be in a follow up PR

### Test plan

Unit tests, manually by setting the dummy flag to true and executing the send flow. Example transaction: https://explorer.celo.org/alfajores/tx/0x81bace7dbe2a426ff54884561da391bdb58db3ab6d63f3c26c30c72295edeef9

### Related issues

- Part of ACT-786

### Backwards compatibility

N/A


src/abis/IERC20.ts
satish-ravi, line 1: Copy of https://github.com/valora-inc/wallet/blob/bb8506945ac9020180e4813320922e47cd3a978d/src/abis/IERC20.json in TS, enables typed contract methods with viem

src/abis/StableToken.ts
satish-ravi, line 1: Copy of https://github.com/valora-inc/wallet/blob/bb8506945ac9020180e4813320922e47cd3a978d/src/abis/StableToken.json in TS

src/viem/saga.ts
```diff
@@ -1,35 +1,238 @@
...
 
-  return hash
+/**
+ * Helper function to call chooseTxFeeDetails for send transactions (aka
+ * transfer contract calls) using parameters that are not specific to contractkit
+ *
+ * @param options the getSendTxFeeDetails options
+ * @returns an object with the feeInfo compatible with viem
+ */
+export function* getSendTxFeeDetails({
```
satish-ravi, line 201: Copied mostly from #4024 which was reverted in #4098

src/viem/saga.ts
```diff
@@ -1,35 +1,238 @@
...
+}: {
+  context: TransactionContext
+  recipientAddress: string
+  amount: BigNumber
+  tokenAddress: string
   comment: string
-) {
+  feeInfo: FeeInfo
+}) {
   const wallet = yield* call(getViemWallet, networkConfig.viemChain.celo)
```
jh2oman, line 48: Don't we need to call wallet.unlockAccount as well? If its not the first time that `getViemWallet` is being called in the app then the wallet will not necessarily be unlocked

src/viem/saga.test.ts
```diff
@@ -0,0 +1,220 @@
...
+            feeInfo.gasPrice
+          ),
+          mockFeeDetails,
+        ],
+      ])
+      .returns(mockViemFeeInfo)
+      .run()
+  })
+
+  // TODO(any): remove this test once we've ensured gas and gasPrice are always BigNumbers
```
jh2oman, line 181: It seems like it might also be appropriate to put a similar comment by chooseTxFeeDetails.

And since this same theme of ensuring fee types will have TODO's in three places, can you make a ticket just to keep all the details in one place?

src/viem/index.ts
```diff
@@ -1,9 +1,10 @@
 import networkConfig, { Network } from 'src/web3/networkConfig'
 import { createPublicClient, http } from 'viem'
+import { celo, celoAlfajores } from 'viem/chains'
 
 export const publicClient = {
   [Network.celo]: createPublicClient({
-    chain: networkConfig.viemChain.celo,
+    chain: networkConfig.viemChain.celo as typeof celo | typeof celoAlfajores, // this type allows us to use feeCurrency on publicClient methods
```
cajubelt, line 7: can we just use this type on networkConfig (even if you have to cast there) instead of casting in multiple places?

src/viem/saga.ts
```diff
@@ -1,35 +1,238 @@
...
+    TAG,
+    'Transferring token',
+    context.description ?? 'No description',
+    context.id,
+    tokenAddress,
+    amount,
+    feeInfo
+  )
+
+  // TODO: add standby transaction
```
cajubelt, line 60: when's this going to get done?

src/viem/saga.ts
```diff
@@ -1,35 +1,238 @@
...
+    // something like
+    // SimulateContractReturnType<typeof stableToken.abi, 'transferWithComment'>
+    // | SimulateContractReturnType<typeof erc20.abi, 'transfer'>
+    // but TSC doesn't like it when passed to writeContract
+    const { request }: SimulateContractReturnType = yield* call(simulateContractMethod)
+
+    const hash = yield* call([wallet, 'writeContract'], request)
+    Logger.debug(TAG, 'Transaction successfully submitted. Hash:', hash)
+
+    // TODO: wait for receipt with a timeout, confirm tx and return
```
cajubelt, line 85: when's this going to get done?

src/viem/saga.ts
cajubelt: if it's part of a Linear issue can you reference it here? If it's not, but you personally plan to do it, can you add your name or username to the todo?

src/viem/saga.ts
satish-ravi: We currently unlock the contract kit wallet at the start of the send flow [here](https://github.com/valora-inc/wallet/blob/main/src/send/saga.ts#L312) and since locks are shared, an explicit unlock call on this wallet wasn't required. We can still call the unlock method on the viem wallet here but we'd probably need a wrapper similar [to this](https://github.com/valora-inc/wallet/blob/main/src/web3/saga.ts#L192) to get a password and then call unlock on the wallet. What do you think?

src/viem/saga.ts
```diff
@@ -1,35 +1,238 @@
...
+  encryptedComment?: string
+}) {
+  const celoTx = yield* call(
+    buildSendTx,
+    tokenAddress,
+    amount,
+    recipientAddress,
+    encryptedComment || ''
+  )
+  // TODO(any): port this logic over from contractkit to use viem
```
cajubelt, line 221: when?

src/viem/saga.test.ts
cajubelt: how do you suggest that we ensure gas/gasPrice are always BigNumber?

cajubelt: the todo is labeled "any", but I don't think anyone could pick this up-- as a non-implementer, I don't really know what's meant to be done here or how I would do it. or why it's desirable

cajubelt (commented):
left comments asking about all the todos

src/viem/saga.ts
jh2oman: I see, you're right then its unnecessary for now, but I still think we should do it. That code in sendPaymentSaga could change as we get off contractkit and we may be unaware of the side effects, this saga could also start being called somewhere else that doesn't unlock the account first. I think we'll have to create a wrapper unlock functioin like that eventually anyways and it would be prudent to do it here and now.

src/viem/saga.test.ts
jophish: Some additional context - we cache fees in Redux per token. When we write the fees to Redux, they're stored as BigNumbers. If there are no fees already present in the cache when we're fetching fees, we calculate a new fee and store it Redux; when we do this, any code that accesses this new fee shortly after it's written to Redux will access it as a BigNumber, however, on _subsequent_ accesses later on, the fee will be returned from Redux as a _string_. I think this has something to do with when/how Redux asynchronously serializes values to the store (we store the actual BN object, but Redux serializes it to a string when the store is actually updated) - some kind of race condition.

One potential solution to this is to just update the fee selector to deserialize the value in Redux (string or otherwise) to a BN object, though that's kind of a workaround to the bad practice of us _not_ serializing the BN before writing it to Redux in the first place.

src/viem/saga.ts
```diff
@@ -1,35 +1,238 @@
...
+        functionName: 'transferWithComment',
+        account: wallet.account,
+        args: [getAddress(recipientAddress), convertedAmount, encryptedComment || ''],
+        feeCurrency,
+        gas,
+        maxFeePerGas,
+      })
+  }
+
+  const { feeCurrency, gas, maxFeePerGas } = yield* call(getSendTxFeeDetails, {
```
jophish, line 166: Does passing `''` for `encryptedComment` here fundamentally change the results of this call? In other words, rather than having two calls to `getSendTxFeeDetails`, can we pull this out to a single call above the `if (isStablecoin(tokenInfo))` block above? I guess this complicates things further though, since we need to encrypt the comment... could we do something like the following above the `if` statement?

```ts
const encryptedComment = comment ? yield* call(
      encryptComment,
      comment,
      recipientAddress,
      userAddress: wallet.account.address,
      true
) : ''

const { feeCurrency, gas, maxFeePerGas } = yield* call(getSendTxFeeDetails, {
      recipientAddress,
      amount,
      tokenAddress,
      feeInfo,
      encryptedComment
})
```

src/viem/saga.ts
satish-ravi: I mentioned this in the PR description, its part of the same linear issue and there will be a followup PR. I can add my name to the TODO if you feel strongly about this.

src/viem/saga.ts
satish-ravi: same as above

src/viem/saga.ts
satish-ravi: I'll make an issue for this

src/viem/saga.test.ts
satish-ravi: I'll make an issue for this.

src/viem/index.ts
satish-ravi: The type on networkConfig is `ViemChain` so it can allow any viem chain. So I am not sure how we can specify the exact type there. Did you have anything specific in mind?

cajubelt: ...not really lol

src/viem/saga.ts
satish-ravi: @jophish I don't think its straightforward to update KeychainLock.unlock to not require a passphrase, I see examples where we pass different passphrases for different accounts with different passphrases in the same wallet. 
https://github.com/valora-inc/wallet/blob/4bfeab16689787c1fcc56c6c68d2f2dde02ff324/src/merchantPayment/hooks.ts#L43-L44
I guess we don't currently support this on ViemWallet, which has a 1:1 mapping between wallets and account instances which is why I believe the `unlockAccount` on ViemWallet just takes a passphrase and a duration and not the account address like the Celo wallet. I guess one option could be to have the ViemWallet's unlockAccount method to get the password and then pass it on but that seems like its something the Wallet class doesn't need to know about.

src/viem/saga.ts
satish-ravi: I've simplified this. Calling fee details with an empty comment should be fine for non stables. It will be ignored when a send tx is built as it conditionally uses the stable token or erc20 contract.

src/viem/saga.ts
```diff
@@ -1,35 +1,234 @@
...
+/**
+ * Gets a function that invokes simulateContract for the appropriate contract
+ * method based on the token. If the token is a stable token, it uses the
+ * `transferWithComment` on the stable token contract, otherwise the `transfer`
+ * method on the ERC20 contract
+ *
+ * @param options an object containing the arguments
+ * @returns a function that invokes the simulateContract method
+ */
+function* getSimulateContractMethod({
```
cajubelt, line 111: should this be called getSimulateTransfer? from the name alone, it sounds like it would be able to simulate other kinds of contract methods

cajubelt (approved):

PR #4171, opened 2023-09-09
fix(nav): app crash on procrastinate recovery

MuckT:
### Description

On iOS when a user would procrastinate during the recovery phrase setup flow the app would crash.

#### Video Before

https://github.com/valora-inc/wallet/assets/26950305/1f5cd7d6-cfee-41c5-a74a-633e4efd954c

#### Video After

https://github.com/valora-inc/wallet/assets/26950305/387adf69-79d7-4e0c-b10a-2f6591dec421

#### Tested with the following reproduction steps
1. Complete the onboarding flow with new account
2. After landing on the home screen navigate to ‘Recovery Phrase’
3. Navigate to this flow until you are viewing your recovery phrase.
4. Tap ‘Cancel’ and ‘Setup Later’ to exit the flow
5. Observe app crash.

### Related issues

N/A

### Backwards compatibility

Yes


jeanregisser (commented):
Great thanks! 👍 

Why does `navigateHome` crash though? I think we should fix that too so it doesn't come back in another case.

MuckT:
@jeanregisser the issue is we cannot call `navigateHome` from within a modal as `navigateHome` resets the navigation state; disposing of the currently active screen which appears to cause a null pointer dereference error within the modal. To test this if we can add a `setTimeout` prior to calling `navigationHome` and if the modal is dismissed prior to calling and there is no crash.

There's some [existing code](https://github.com/valora-inc/wallet/blob/main/src/send/saga.ts#L326-L330) that hints at this and it was confirmed through testing on the `Dialog` modals in `src/account/Settings.tsx`. I'll add some docs for navigateHome and modal usage.

src/navigator/NavigationService.ts
```diff
@@ -226,12 +226,20 @@ interface NavigateHomeOptions {
...
+ * If called from a modal make sure to pass fromModal: true. Otherwise it will cause a null pointer dereference and subsequent app crash
+ */
 export function navigateHome(options?: NavigateHomeOptions) {
   const { params } = options ?? {}
-  navigationRef.current?.reset({
-    index: 0,
-    routes: [{ name: Screens.DrawerNavigator, params }],
-  })
+  if (params?.fromModal) {
+    navigate(Screens.WalletHome)
```
jeanregisser, line 236: I think one issue is if at some point we call `navigateHome` from the onboarding flow from a modal, the stack won't be reset.

I don't yet have a great alternative proposal though.

But what about doing something similar to this when `fromModal` is set?

https://github.com/valora-inc/wallet/blob/5c0a7a5889a44a67a87c3df9117a7f8ec783a4fb/src/account/Settings.tsx#L296-L301

Not ideal I know but will reset the stack.

jeanregisser (commented):
Thanks for looking into the details and the new test 👍

e2e/src/usecases/NewAccountOnboarding.js
```diff
@@ -48,27 +74,33 @@ export default NewAccountOnboarding = () => {
...
+      .withTimeout(10 * 1000)
+    await element(by.text('Set Up Later')).tap()
 
-    await expect(element(by.id('AccountKeyWordsContainer'))).toBeVisible()
+    // App doesn't crash and we arrive at the home screen
+    await arriveAtHomeScreen()
+  })
+
+  it('Setup Recovery Phrase', async () => {
+    await reloadReactNative()
```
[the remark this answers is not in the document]
MuckT: We don't remove in https://github.com/valora-inc/wallet/pull/4171/commits/1eb8b95b57cddc2f4b68730253f81c6dd612efd1!

src/navigator/NavigationService.ts
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/4171/commits/99472a5afd2daeb274e2489ff2411236601894fa!

jeanregisser (approved):
Thanks!! 👍

PR #4173, opened 2023-09-11
feat: show P0 notifications only on the home screen

jeanregisser:
### Description

As the title says.

This ended up requiring very few changes.

Also for now P0 notifications are hidden from the notification center screen. I'll check this is what we want.

### Test plan

**When the notification center is enabled**

With a single notification marked with `showOnHomeScreen: true`:

 

With multiple notifications marked with `showOnHomeScreen: true`:

 

Note that this isn't a case we'll do on purpose. But since it allowed to reuse `<NotificationBox />` with minimal changes I left it this way. The highest priority notification is still displayed first.

**When the notification center is disabled**

 

This is same as before this PR.

### Related issues

- Fixes RET-800

### Backwards compatibility

Yes

### DIFF
--- a/src/components/SimpleMessagingCard.tsx
+++ b/src/components/SimpleMessagingCard.tsx
@@ -10,6 +10,7 @@ export interface Props {
   icon?: ImageSourcePropType | React.ReactNode
   callToActions: CallToAction[]
   priority: number
+  showOnHomeScreen?: boolean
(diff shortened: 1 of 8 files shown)


MuckT (approved):
👨‍🍳

bakoushin (approved):

PR #4175, opened 2023-09-12
feat(viem): implement sendAndMonitorTransaction

satish-ravi:
### Description

Implement the equivalent of the existing [sendAndMonitorTransaction](https://github.com/valora-inc/wallet/blob/955ea9dda0c8e9ca8b8e079097c74b307adf45f1/src/transactions/saga.ts#L151) method using contract kit. 

### Test plan

Unit tests, manually by enabling the dummy flag and running through the send flow

### Related issues

- Fixes ACT-786

### Backwards compatibility

N/A


src/transactions/actions.ts
```diff
@@ -59,6 +60,11 @@ export interface TransactionConfirmedAction {
   receipt: CeloTxReceipt
 }
 
+export interface TransactionConfirmedViemAction {
+  type: Actions.TRANSACTION_CONFIRMED_VIEM
+  txId: string
+}
```
satish-ravi, line 66: Made a separate action instead of reusing the existing one as the viem receipt isn't directly serializable without a custom serialization for big ints. The actual receipt isn't actually used anywhere from the redux action other than to check for success status, which is redundant since this action is only fired when the tx is a success.

cajubelt (approved):

PR #4176, opened 2023-09-12
chore(branding): Fallback to https if branding cloning via git protocol fails

bakoushin:
### Description

Currently Valora branding is synced only via `git` protocol, which leads to an error if someone is using `https` (e.g. authenticated using [GitHub CLI](https://cli.github.com)), although both ways are fine for cloning.

This adds a fallback to `https` protocol if cloning using `git` protocol fails.

### See also

[Slack thread](https://valora-app.slack.com/archives/C029Z1QMD7B/p1691490310877379)

### Test plan

Tested manually

### Related issues

N/A

### Backwards compatibility

Yes


bakoushin:
I'm worried this somehow caused decrease in build size
what do you think @MuckT?

scripts/sync_branding.sh
```diff
@@ -26,7 +26,13 @@ valora_branding_sha=89f86a0ce54d7d06afe914ea6f4a057affe09024
 if [[ "$branding" == "valora" ]]; then
   # prevents git from asking credentials
   export GIT_TERMINAL_PROMPT=0
-  if [[ ! -e branding/valora ]] && ! git clone git@github.com:valora-inc/valora-app-branding.git branding/valora ; then
+  remote_path=valora-inc/valora-app-branding.git
+  local_dir=branding/valora
+  if [[ ! -e $local_dir ]] && 
+    ( 
+      ! git clone git@github.com:$remote_path $local_dir &&
+      ! git clone https://github.com/$remote_path $local_dir
```
jeanregisser, line 34: nit: to avoid surprises with variables expansion
```suggestion
  if [[ ! -e "$local_dir" ]] && 
    ( 
      ! git clone "git@github.com:$remote_path" "$local_dir" &&
      ! git clone "https://github.com/$remote_path" "$local_dir" 
```

jeanregisser (approved):
Thanks! 👍 

Already approved with a suggestion.

scripts/sync_branding.sh
bakoushin: Thanks!

jeanregisser:
From the Emerge Tools size diff above, it seems the e2e test wasn't able to fetch the Valora branding.
Not sure why though, but it seems this change affects it too. 🤔

bakoushin:
I think it's because I messed with a logic (`||` instead of `&&`) 😅 
That is already fixed. Hope now the test will be fine.

PR #4178, opened 2023-09-12
build(ios): set os minimum deployment to 14.5

MuckT:
### Description

Removes support for iOS versions < 14.5.

### Test plan

Tested locally and in CI on iOS 14.5.

### Related issues

ACT-888

### Backwards compatibility

Yes for all users expect iOS uses < 14.5. Removes support for the following devices:

| Device | Release Date |
| --- | --- |
| [iPad Air (1st)](https://en.wikipedia.org/wiki/IPad_Air) | November 1, 2013 |
| [iPad Mini 2](https://en.wikipedia.org/wiki/IPad_Mini_2) | October 23, 2013 | 
| [iPad Mini 3](https://en.wikipedia.org/wiki/IPad_Mini_3) | October 22, 2014 |
| [iPhone 5S](https://en.wikipedia.org/wiki/IPhone_5S) | September 20, 2013 |
| [iPhone 6](https://en.wikipedia.org/wiki/IPhone_6) | September 19, 2014 |
| [iPod Touch (6th)](https://en.wikipedia.org/wiki/IPod_Touch_(6th_generation)) | July 15, 2015 |


ios/celo.xcodeproj/project.pbxproj
jeanregisser, line 1: What we want here is instead is update the project level setting: 

Screenshot 2023-09-20 at 09 32 15

Then the diff will look like this:
Screenshot 2023-09-20 at 09 32 49

The `celoTests` target isn't used, we could delete the setting so it would inherit from the project setting instead.

MuckT: @jeanregisser I think this should all be handled with the project settings now

jeanregisser (approved):
🚀

PR #4189, opened 2023-09-13
feat(payment): Add airtime method to select providers

finnian0826:
### Description

Select payment method screen will now show Airtime for users in KE and NG when Fonbnk quotes are available.

### Test plan

https://github.com/valora-inc/wallet/assets/140328381/149e0474-eb66-4672-959c-68fa37159035

https://github.com/valora-inc/wallet/assets/140328381/9fd42da4-5010-4688-b7f8-6301115d4049

### Backwards compatibility

Only shows when Fonbnk quote is received, which currently only happens on dev. If no Fonbnk quote is received nothing new happens.


src/fiatExchanges/quotes/ExternalQuote.ts
```diff
@@ -92,15 +93,28 @@ export default class ExternalQuote extends NormalizedQuote {
     return null
   }
 
+  getReqsSubtitle(): string | null {
+    return !isSimplexQuote(this.quote) &&
+      this.getPaymentMethod() === PaymentMethod.Airtime &&
+      this.quote.extraReqs
+      ? this.quote.extraReqs.mobileCarrier ?? ''
+      : this.getKycInfo()
```
cajubelt, line 101:
```suggestion
    return this.getPaymentMethod() === PaymentMethod.Airtime &&
      this.quote.extraReqs?.mobileCarrier ??
      : this.getKycInfo()
```
if extraReqs is truthy but mobileCarrier is not, seems like it should show kyc info and not an empty string

also, the simplex part doesn't seem relevant

cajubelt: one more thing-- we need it to say "MTN required", not just "MTN". The string should also be localized. I slacked you some information on how to accomplish that

src/fiatExchanges/quotes/ExternalQuote.ts
```diff
@@ -92,15 +93,28 @@ export default class ExternalQuote extends NormalizedQuote {
...
-      ? DEFAULT_BANK_SETTLEMENT_ESTIMATION
-      : DEFAULT_CARD_SETTLEMENT_ESTIMATION
+    // payment method can be bank, card or airtime
+    switch (this.getPaymentMethod()) {
+      case PaymentMethod.Bank:
+        return DEFAULT_BANK_SETTLEMENT_ESTIMATION
+      case PaymentMethod.Airtime:
+        return DEFAULT_AIRTIME_SETTLEMENT_ESTIMATION
+      default:
+        return DEFAULT_CARD_SETTLEMENT_ESTIMATION
```
cajubelt, line 116: seems like the default should be bank, since it's the most conservative of these

cajubelt: actually, this is better:
- define a constant with type Record<PaymentMethod, SettlementTimeEstimation>
- use that to map payment method to settlement time

that way, if we add a payment method but forget to update this, we get a compile-time error instead of silently defaulting to another payment method's time estimation

cajubelt (commented):
partial review since I'm seeing a couple things already and need to switch tasks

src/fiatExchanges/quotes/ExternalQuote.ts
finnian0826: I think the Simplex part is needed, I tried getting rid of it and `this.quote.extraReqs?` gives an error saying extraReqs does not exist in SimplexQuote. I think I could add an optional field to SimplexQuote for extraReqs which never gets used but otherwise I think I need this check.

src/fiatExchanges/quotes/ExternalQuote.ts
finnian0826: Settlement times are only defined for Bank, Card and Airtime, however PaymentMethod also includes Coinbase, MobileMoney and FiatConnectMobileMoney, should I have a default settlement for these other payment methods? Otherwise the map doesn't work because for key type PaymentMethod there are missing values.

src/fiatExchanges/SelectProvider.test.tsx
```diff
@@ -186,6 +186,7 @@ describe(SelectProviderScreen, () => {
         [PaymentMethod.MobileMoney]: false,
         [PaymentMethod.Coinbase]: false,
         [PaymentMethod.FiatConnectMobileMoney]: false,
+        [PaymentMethod.Airtime]: false,
```
cajubelt, line 189: on line 219 of this file there is a test `shows the provider sections` that should be updated to add a check for airtime

cajubelt (approved):
looks close! nice jo

src/fiatExchanges/quotes/ExternalQuote.ts
cajubelt: (discussed offline)

src/fiatExchanges/quotes/ExternalQuote.ts
cajubelt: ok I didn't realize the simplex check was a type guard

PR #4190, opened 2023-09-13
feat(token-icon): add dynamic token icon component

MuckT:
### Description

Adds a component for dynamic rendering of token icons where native tokens, CELO & ETH, render as before and non native tokens, stCELO etc...,  render with a small child icon of the network they are on. This is accomplished by displaying the icon if `networkIconUrl` is populated on a token: this option was chosen as it allows the most backend configurability. 

#### Example Screenshots - not used on these pages
| iOS Tokens | iOS Positions | Android Tokens | Android Postions |
| --- | --- | --- | --- |
|  |  |  |  |

### Test plan

- Tested locally on iOS
- Unit tests added

### Related issues

ACT-914

### Backwards compatibility

Yes


src/tokens/AssetItem.tsx
```diff
@@ -42,7 +43,7 @@ export const PositionItem = ({ position }: { position: Position }) => {
       onPress={onPress}
     >
       <View style={styles.row}>
-        <Image source={{ uri: position.displayProps.imageUrl }} style={styles.tokenImg} />
+        <TokenIcon token={position} viewStyle={styles.tokenImgContainer} />
```
satish-ravi, line 46: The new asset icon with the logo is applicable only to the new Assets page, we want to leave the existing Asset list and Dapp positions as is.

src/components/TokenIcon.tsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react'
+import { StyleProp, StyleSheet, View, ViewStyle } from 'react-native'
+import FastImage from 'react-native-fast-image'
```
satish-ravi, line 3: What's the benefit of this over just regular Image? Seems like this package hasn't been active for close to a year

src/components/TokenIcon.tsx
```diff
@@ -0,0 +1,75 @@
...
+export default function TokenIcon({ token, viewStyle, testID }: Props) {
+  const isToken = 'symbol' in token
+  const tokens = useSelector(tokensListSelector)
+
+  // Always defaults to Celo as token.network is not currently defined
+  // Switch statement for easier future expansion
+  let networkImage
+  switch (token.network) {
+    default:
+    case 'Celo':
```
satish-ravi, line 30: would be nice to avoid the switch here with hardcoded network token names.. Maybe this logic can go in the cloud function / blockchain api route to get tokens and the wallet directly can access an optional "subscript logo url" (could use a better name) which can be used here? That allows us flexibility on switching this logic too in the future without a new version release.

src/components/TokenIcon.tsx
MuckT: [FastImage](react-native-fast-image
) has some better caching as is more performant than the react native image component. Since we already use it in the NFTMedia component it doesn't seem like much of a risk to add here.

src/components/TokenIcon.tsx
MuckT: I agree that it would be nice to this hardcoded switch. I do worry storing another key value for every entry might be get quite large when we already have the network to use, but if that's unfounded it certainly seems easiest.

Maybe this is a good case for a small dynamic config with some basic network information dynamically: this would be network and not chain specific. Also because its just general information it would not need to be app version specific as was the concern in https://github.com/valora-inc/wallet/pull/4115#discussion_r1309727228.

```JSON
{
  "celo": {
    "nativeSymbol": "CELO",
    "iconUrl": "https://example.com/celo-icon.png",
    "foundationWebsite": "https://example.com",
  },
  "ethereum": {
    "nativeSymbol": "ETH",
    "iconUrl": "https://example.org/icon.png",
    "foundationWebsite": "https://example.org",
  }
}
```

src/components/TokenIcon.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+import TokenIcon from 'src/components/TokenIcon'
+import { createMockStore } from 'test/utils'
+import { mockCeloAddress, mockCusdAddress, mockTokenBalances } from 'test/values'
+
+// Setting up the mock token balances with expected additional values
+mockTokenBalances[mockCusdAddress].networkIconUrl =
+  'https://raw.githubusercontent.com/valora-inc/address-metadata/main/assets/tokens/CELO.png'
+
+const CELO_TOKEN = mockTokenBalances[mockCeloAddress]
+const CUSD_TOKEN = mockTokenBalances[mockCusdAddress]
```
satish-ravi, line 13:
```suggestion
const CELO_TOKEN = mockTokenBalances[mockCeloAddress]
const CUSD_TOKEN = {
  ...mockTokenBalances[mockCusdAddress],
  networkIconUrl: 'https://raw.githubusercontent.com/valora-inc/address-metadata/main/assets/tokens/CELO.png'
```
so this doesn't change this globally.

satish-ravi (approved):

PR #4191, opened 2023-09-14
chore: make notifications event properties consistent

kathaypacific:
### Description

This PR makes the notification id and type consistent for the notification_impression and notification_select events. After this change, we have 3 types of notifications:
1. all bundled notifications (e.g. supercharge, verify phone number, backup) will have the same `notificationId` and `notificationType`
2. all user generated notifications (e.g. incoming / outgoing payment requests) will have the same `notificationType` and `notificationId`, except the `notificationId` will also contain the unique identifier of the request. The format is `type/id`.
3. all remote notifications will have `notificationType` = `"remote_notification"` and the `notificationId` as whatever is set in admin dashboard.

Summary of changes:
1. expect notification type and id as event properties for both `notification_select` and `notification_impression`
2. when generating the list of notifications, we were hard coding an `id` for all bundled and user generated notifications. This was a different value to what is used in the analytics events, and is used as the `id` in the impression event. e.g. the `notification_impression` event had properties `{"notificationId":"incomingPaymentRequest"}` while `notification_select` had `{"notificationType":"incoming_tx_request","selectedAction":"pay"}`. i've removed these arbitrarily hard coded id's and replaced them with the values in the notification types enum. i guess this is a breaking change for the data but i don't see how the data before was analysable anyway. i'll let Chris know about the change in id value just in case.
3. move the notification enums into the types file to avoid introducing new cyclical dependencies.

### Test plan

Tested manually and in unit tests. Tested the events with and without the notification center.

Without notification center:
Screenshot 2023-09-14 at 14 44 44

With notification center:
Screenshot 2023-09-14 at 14 45 45

### Related issues

- Fixes RET-845

### Backwards compatibility

Y


src/components/SimpleMessagingCard.tsx
```diff
@@ -12,6 +13,7 @@ export interface Props {
   priority: number
   showOnHomeScreen?: boolean
   testID?: string
+  type: NotificationType
```
bakoushin, line 16: I wonder why we introduce this prop? Looks like it is not used within the component.

bakoushin (approved):
Thanks, it's great!

Please look at one little question about potentially redundant code. 

Also, I wonder do we really need to send unique user-generated ids for payment requests and escrows to analytics? Wouldn't it just clutter the data?

src/components/SimpleMessagingCard.tsx
kathaypacific: ah good question 😅 https://github.com/valora-inc/wallet/blob/58869ca6a0ebe70ab332691ebd85efaa3a871860/src/home/NotificationBox.tsx#L136 <- here is the culprit, and i was too lazy to untangle the types. i've updated now in https://github.com/valora-inc/wallet/pull/4191/commits/188bfdaecb77ea7cb24bf2ffa83425b26f909482, so hope this looks less weird

kathaypacific:
> Thanks, it's great!
> 
> Please look at one little question about potentially redundant code.
> 
> Also, I wonder do we really need to send unique user-generated ids for payment requests and escrows to analytics? Wouldn't it just clutter the data?

yeah good question...i actually had to put the user generated id's there so that we have unique id's to track the impression events in the notification list https://github.com/valora-inc/wallet/blob/58869ca6a0ebe70ab332691ebd85efaa3a871860/src/home/NotificationCenter.tsx#L171

i think it looks ugly, but i thought that it's okay because we can ignore the unique id's during data analysis by querying using regex rather than exact string match. thinking more about it now, it might make mixpanel harder to use :/ i'll take another pass to see if i can think of a better solution

kathaypacific:
> > Thanks, it's great!
> > Please look at one little question about potentially redundant code.
> > Also, I wonder do we really need to send unique user-generated ids for payment requests and escrows to analytics? Wouldn't it just clutter the data?
> 
> yeah good question...i actually had to put the user generated id's there so that we have unique id's to track the impression events in the notification list
> 
> https://github.com/valora-inc/wallet/blob/58869ca6a0ebe70ab332691ebd85efaa3a871860/src/home/NotificationCenter.tsx#L171
> 
> i think it looks ugly, but i thought that it's okay because we can ignore the unique id's during data analysis by querying using regex rather than exact string match. thinking more about it now, it might make mixpanel harder to use :/ i'll take another pass to see if i can think of a better solution

darn, i'm torn. ideally any `id` should be unique, so that's why i like having the user generated id in there. but i agree, it's not helpful to send the id in analytics and probably makes it harder to analyse data using the mixpanel UI (i couldn't find an easy way to make an "OR" condition with exact + fuzzy match...) 

so the 2 options we have are to have what we want in analytics in the code (i.e. accept that the `id` for some notifications are not unique, can't use them for react keys etc.) or keep id's unique in the code and filter out the parts we don't want before sending to analytics. i kind of prefer the latter because it's more explicit about what we are trying to achieve, but it's also more brittle because we need to remember to add this logic to every relevant new event 😮‍💨

bakoushin:
It looks like we are not reusing the unique ids from the list in analytics events, but declare them separately.
What if we just skip adding unique part?

bakoushin:
I forgot about  `notification_impression`: here we are reusing the id.
But since we decided to duplicate `notificationType` into  `notificationId` we could implement this in the code as well.

kathaypacific:
hey @bakoushin i've been mulling this over for a few days and think the best thing to do is leave it like this. are you fine with this?

the wallet should expect that a notification id can uniquely identify a notification. so i think it makes sense to append the user generated id to the `notificationId` for user generated notifications. the question would be whether we should have special logic before dispatching analytics events to remove the unique id's which don't add value for analytics. we could do this but it seems brittle to remember to do this for new events or for refactors, and i'd say it's worth the effort if passing the unique id to analytics makes data analysis very difficult. it seems like that's not the case after [checking with Chris](https://valora-app.slack.com/archives/C02DY3RK79T/p1695043765828009), so i'm inclined to keep the code simple. wdyt?

bakoushin:
@kathaypacific thanks for checking that with Chris! My whole concern was about analytics being usable.
Since having unique ids makes no problem for him, having analytics ids consistent with app keys is fine!

PR #4192, opened 2023-09-14
feat(networks): Use `networkId` at the API level

jophish:
### Description

Currently in draft state; the actual API changes required for this to work have not been merged yet. This depends on changes to the APIs in `cloud-functions` and `blockchain-api`.

* `cloud-functions` PR implementing the required changes: [here](https://github.com/valora-inc/cloud-functions/pull/129)
* `blockchain-api` PR implementing the required changes: TBD

For some context, we want to get this change in before the release cut. If we ever want our backend services to be able to speak to multiple different blockchain environments, _without this change_, we'll need a bit of a hacky workaround. (e.g., if we want production GCP  blockchain-api to be able to fetch data from `celo-alfajores` or `ethereum-sepolia`.) If we do not make this change, and want to have this kind of support in the future, we'll need to update `blockchain-api` and `fetchProvider` in a backwards compatible way in order to not break intermediate app versions. This sort of BC change would be somewhat hacky - there are two obvious ways we could do it:

* Have the existing `network` parameter in these APIs expanded or "overloaded" to recognize _both_ env-agnostic and env-specific network identifiers (with env-agnostic identifiers "falling back" to the service's deploy environment to resolve the correct env-specific identifier).
* Adding another, optional, env-specific parameter to these endpoints, so they can field requests both for env-agnostic identifiers and env-specific identifiers. This would mean it would be possible to call these endpoints with: 
  * no network identifier at all (falling back onto celo as default chain + deploy env)
  * an env-specific identifier
  * an env-agnostic identifier (falling back onto deploy env)
  * both an env-specific and env-agnostic identifier (undefined behavior?)

Since we've only just really begun hooking together these multi-chain backend APIs with the wallet, it doesn't make a ton of sense to introduce the potential for a hacky-ish fix down the road when we can obviate the need for it now.

### Test plan

Unit tests updated; will manual test on Alfajores once the `cloud-functions` and `blockchain-api` PRs have been merged and deployed.

### Related issues

N/A

### Backwards compatibility

Yes.

### DIFF
--- a/src/send/saga.test.ts
+++ b/src/send/saga.test.ts
@@ -63,6 +63,23 @@ jest.mock('src/transactions/types', () => {
   }
 })
 
+jest.mock('src/web3/networkConfig', () => {
+  const originalModule = jest.requireActual('src/web3/networkConfig')
+  return {
+    __esModule: true,
+    ...originalModule,
+    default: {
+      ...originalModule.default,
+      networkToNetworkId: {
+        celo: 'celo-alfajores',
+        ethereum: 'ethereuim-sepolia',
+      },
+      supportedNetworkIds: ['celo-alfajores', 'ethereum-sepolia'],
+      defaultNetworkId: 'celo-alfajores',
+    },
+  }
+})
+
 const mockContext = { id: 'mock' }
 mockNewTransactionContext.mockReturnValue(mockContext)
 
--- a/src/send/saga.test.ts
+++ b/src/send/saga.test.ts
@@ -74,7 +74,6 @@ jest.mock('src/web3/networkConfig', () => {
         celo: 'celo-alfajores',
         ethereum: 'ethereuim-sepolia',
       },
-      supportedNetworkIds: ['celo-alfajores', 'ethereum-sepolia'],
       defaultNetworkId: 'celo-alfajores',
     },
   }
--- a/src/send/saga.test.ts
+++ b/src/send/saga.test.ts
@@ -23,7 +23,7 @@ import { getERC20TokenContract, getStableTokenContract } from 'src/tokens/saga'
 import { addStandbyTransaction } from 'src/transactions/actions'
 import { sendTransactionAsync } from 'src/transactions/contract-utils'
 import { sendAndMonitorTransaction } from 'src/transactions/saga'
-import { TokenTransactionTypeV2, TransactionStatus, Network } from 'src/transactions/types'
+import { TokenTransactionTypeV2, TransactionStatus, NetworkId } from 'src/transactions/types'
 import {
   getConnectedAccount,
   getConnectedUnlockedAccount,
@@ -305,7 +305,7 @@ describe(sendPaymentSaga, () => {
       .put(
         addStandbyTransaction({
           context: mockContext,
-          network: Network.Celo,
+          networkId: NetworkId['celo-alfajores'],
           type: TokenTransactionTypeV2.Sent,
           comment: sendAction.comment,
           status: TransactionStatus.Pending,
--- a/src/transactions/feed/queryHelper.ts
+++ b/src/transactions/feed/queryHelper.ts
@@ -12,7 +12,7 @@ import { vibrateSuccess } from 'src/styles/hapticFeedback'
 import { fetchTokenBalances } from 'src/tokens/slice'
 import { updateTransactions } from 'src/transactions/actions'
 import { transactionHashesSelector } from 'src/transactions/reducer'
-import { TokenTransaction, Network } from 'src/transactions/types'
+import { TokenTransaction, Network, NetworkId } from 'src/transactions/types'
 import Logger from 'src/utils/Logger'
 import config from 'src/web3/networkConfig'
 import { walletAddressSelector } from 'src/web3/selectors'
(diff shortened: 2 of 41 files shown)


src/web3/networkConfig.ts
```diff
@@ -167,6 +173,12 @@ const CAB_ISSUE_VALORA_KEYSHARE_MAINNET = `${CLOUD_FUNCTIONS_MAINNET}/issueValor
 const networkConfigs: { [testnet: string]: NetworkConfig } = {
   [Testnets.alfajores]: {
     networkId: '44787',
+    networkToNetworkId: {
+      [Network.Celo]: NetworkId['celo-alfajores'],
+      [Network.Ethereum]: NetworkId['ethereum-sepolia'],
+    },
+    supportedNetworkIds: [NetworkId['celo-alfajores'], NetworkId['ethereum-sepolia']],
```
jophish, line 180: These two fields - `networkToNetworkId` and `supportedNetworkIds` - may seem a bit redundant. If we only ever "support" the networks that the wallet knows about (via the `Network` enum), then we can just use `Object.values(networkToNetworkId)` in place of `supportedNetworkIds`. `networkToNetworkId` however provides type safety (forcing us to update it when a new `Network` value is added, which is good), so this would mean that without `supportNetworkIds`, we couldn't update the `Network` enum without immediately supporting that network (or remove support for a network without modifying the `Network` enum).

src/web3/networkConfig.ts
```diff
@@ -3,17 +3,20 @@ import { OdisUtils } from '@celo/identity'
...
+  mainnet as ethereum,
+  sepolia as ethereumSepolia,
+} from 'viem/chains'
 
 export enum Testnets {
   alfajores = 'alfajores',
   mainnet = 'mainnet',
 }
 
-export enum Network {
```
jophish, line 13: removing this in favor of the other `Network` enum which is used more widely.

src/transactions/feed/queryHelper.ts
```diff
@@ -344,16 +361,16 @@ async function queryChainTransactionsFeed({
...
         }),
       },
     },
   }
 }
 
 export const TRANSACTIONS_QUERY = `
-  query UserTransactions($address: Address!, $localCurrencyCode: String, $afterCursor: String, $network: Network!) {
-    tokenTransactionsV3(address: $address, localCurrencyCode: $localCurrencyCode, afterCursor: $afterCursor, network: $network) {
+  query UserTransactions($address: Address!, $localCurrencyCode: String, $afterCursor: String, $networkId: NetworkId!) {
```
jophish, line 372: N.B to @cajubelt -- I'm assuming that the parameter for the blockchain-api endpoint is going to be called `networkId` with an enum name of `NetworkId` in the GraphQL schema.

src/transactions/feed/queryHelper.ts
```diff
@@ -43,6 +43,13 @@ export interface QueryResponse {
   }
 }
 
+const networkIdToGraphQlParam: Record<NetworkId, string> = {
```
cajubelt, line 46: could also just use `networkId.replaceAll('-', '_')`

src/web3/networkConfig.ts
cajubelt: don't we filter supportedNetworkIds with the remote config before using them anyway?

cajubelt: I'm not sure I see the purpose of supportedNetworkIds still

cajubelt (commented):
main things are mocking networkConfig and questions about supportedNetworkIds

src/transactions/feed/queryHelper.ts
```diff
@@ -62,10 +69,10 @@ const deduplicateTransactions = (
   return transactionsWithoutDuplicatedHash
 }
 
-export function getAllowedNetworks(): Array<Network> {
+export function getAllowedNetworkIds(): Array<NetworkId> {
   return getFeatureGate(StatsigFeatureGates.SHOW_MULTI_CHAIN_TRANSFERS)
-    ? Object.values(Network)
-    : [Network.Celo]
+    ? config.supportedNetworkIds
```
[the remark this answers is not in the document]
jophish: I can ditch the `supportedNetworkIds` field, it's not crucial and is prolly of dubious value.

src/web3/networkConfig.ts
jophish: the naming is wonky. `supportedNetworkIds` is supposed to encompass "all" the networks we support for multi-chain, assuming multi-chain features are enabled. it's not really important though; i've updated the code to just use `Object.values(networkToNetworkId)` instead.

src/send/saga.test.ts
```diff
@@ -63,6 +63,23 @@ jest.mock('src/transactions/types', () => {
...
+  return {
+    __esModule: true,
+    ...originalModule,
+    default: {
+      ...originalModule.default,
+      networkToNetworkId: {
+        celo: 'celo-alfajores',
+        ethereum: 'ethereuim-sepolia',
+      },
+      supportedNetworkIds: ['celo-alfajores', 'ethereum-sepolia'],
```
cajubelt, line 77: is this needed anymore? looks like you removed it from networkConfig.ts

cajubelt (approved):
🥳

src/send/saga.test.ts
jophish: good catch, ty

PR #4194, opened 2023-09-15
feat(viem): use in send flow based on feature gate

satish-ravi:
### Description

Update send flow to use viem. Also updates the Send e2e test to run with both viem and contract kit by using statsig overrides through launch arguments.

### Test plan

- Manually by setting up overrides
- E2E test 

### Related issues

- Fixes ACT-787

### Backwards compatibility

Yes


src/send/saga.test.ts
```diff
@@ -42,14 +44,15 @@ import {
...
+  mockQRCodeRecipient,
   mockQrCodeData,
   mockQrCodeData2,
-  mockQRCodeRecipient,
   mockRecipientInfo,
   mockTransactionData,
 } from 'test/values'
 
 jest.mock('@celo/connect')
+jest.mock('src/statsig/')
```
jeanregisser, line 55: Is the trailing slash required?

jeanregisser (approved):
Awesome! 🚀 
I love that you added an e2e test already to make sure we don't have a blind spot there.

jeanregisser:
I noticed the following [caveat](https://github.com/iamolegga/react-native-launch-arguments#caveats) for `react-native-launch-arguments`:

> Due to React Native https://github.com/facebook/react-native/issues/37518, on Android, the module force-waits for the Android activity to reach the [RESUMED state](https://developer.android.com/guide/components/activities/activity-lifecycle#alc), before moving foward with native-modules initialization completion. While commonly the wait is scarce (the activity is already in the resumed state at this point), until the RN issue is fixed, the module may introduce delays in app loading time in some edge cases.

Hopefully this is negligible 👀

src/send/saga.test.ts
satish-ravi: nope, removed

PR #4200, opened 2023-09-15
fix: make transaction migrations idempotent

jophish:
### Description

A couple recent transaction migrations that we added were not idempotent, which introduced an issue when downgrading and re-upgrading app version. Basically, the v145 migration updated stored transactions by translating their `__typename` field to a new format. If a user upgraded from v145 (thus triggering this translation), then downgraded to v144, the transactions would still retain their v145 translated `__typename` field. This did not crash the app (I believe), however when upgrading to v145 _again_, another translation would be attempted, which translated `__typename` to incorrect values (since the translation is not idempotent).

This would never affect actual prod installs of the wallet, but it _can_ affect alfajores/nightly/local dev builds.

This fix simplifies the v145 migration, and another migration that follows, to simply wipe the cached transactions from Redux.

### Related issues

- Fixes #[issue number here]


satish-ravi (approved):

jeanregisser (commented):
I'm not entirely sure we should optimize for the downgrade use case for non production builds.

The downside of this is production users will briefly see their transaction feed empty while it fetches it again.

Are we fine with this?

jophish:
Closing this out re: @jeanregisser's reasoning -- this PR was mostly opportunistic anyhow, since it piggybacked off of @kathaypacific's patch.

PR #4203, opened 2023-09-18
ci: update sample wallet address

kathaypacific:
### Description

We want to create some tests around CPV and verified the existing 12 word seed phrase wallet against a new number but this broke the existing test since it's expecting a different set of onboarding steps. This PR creates a new wallet for the 12 word seed phrase test and saves the existing wallet for the upcoming CPV tests.

Context: https://valora-app.slack.com/archives/C02E2FE98P2/p1695025073879669

### Test plan

I've funded the account with 5 of each celo native token.

### Related issues

n/a

### Backwards compatibility

Y


bakoushin (approved):
LGTM!

PR #4207, opened 2023-09-19
feat: add new token balance item

MuckT:
### Description

Adds a new `tokenBalanceItem.tsx` component to be used in the new token assets screens.

#### Screenshots Samples on Existing Pages for Demo Only
| iOS | Android |
| ----- | ----- | 
|  |  |

### Test plan

- [x] Tested locally on iOS
- [x] Unit tests added

### Related issues

- Fixes ACT-912

### Backwards compatibility

Yes


src/tokens/TokenBalanceItem.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+import React from 'react'
+import { Provider } from 'react-redux'
+import { TokenBalance } from 'src/tokens/slice'
+import { TokenBalanceItem } from 'src/tokens/TokenBalanceItem'
+import { createMockStore } from 'test/utils'
+import { mockCusdAddress } from 'test/values'
+
+let mockTokenInfo: TokenBalance
+
+describe('TokenBalanceItem', () => {
```
satish-ravi, line 12: can we add a test for when the local currency amount is `--`?

kathaypacific (commented):
could we please add some padding between the token name and the token balance? same for the network name and fiat balance. in the screenshot, the ellipsis of the long token name directly touches the token balance 🙈

src/tokens/TokenBalanceItem.tsx
```diff
@@ -0,0 +1,118 @@
...
+    <Touchable>
+      <View style={styles.container}>
+        <TokenIcon token={token} viewStyle={styles.marginRight} />
+        <View style={styles.textContainer}>
+          <View style={styles.line}>
+            <Text numberOfLines={1} style={styles.label}>
+              {token.name}
+            </Text>
+            {/* Consider moving formatValueToDisplay out of TokenDisplay */}
+            <Text numberOfLines={1} style={styles.amount}>
```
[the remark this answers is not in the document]
kathaypacific: oh i see Satish commented something similar. i guess my main concern is that the logic for displaying token prices / rounding / handling stale and null prices should be handled only once, so if it doesn't make sense to use an existing component, then this should be pulled out to a new component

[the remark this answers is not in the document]
MuckT: Pulled out the display logic into it's own component that just takes a token balance in https://github.com/valora-inc/wallet/pull/4207/commits/3dea146417ce7acd02511bf7d20d127f72c1faa5. For now it's living in the same file, but it can be moved in the future if needed.

src/tokens/TokenBalanceItem.test.tsx
MuckT: Added in https://github.com/valora-inc/wallet/pull/4207/commits/b0f64cda6fc9512fd4580084c08f6cb74bfc925e!

src/tokens/TokenBalanceItem.tsx
```diff
@@ -0,0 +1,118 @@
...
+            {showAmount ? (
+              <Text numberOfLines={1} style={styles.subAmount}>
+                {localCurrencySymbol}
+                {formatValueToDisplay(amountInLocalCurrency.absoluteValue())}
+              </Text>
+            ) : (
+              <Text numberOfLines={1} style={styles.subAmount}>
+                --
+              </Text>
+            )}
```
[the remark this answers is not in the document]
MuckT: Move it to it's own component in https://github.com/valora-inc/wallet/pull/4207/commits/3dea146417ce7acd02511bf7d20d127f72c1faa5. It lives in the same file currently, but can be moved easily.

src/tokens/slice.ts
```diff
@@ -17,6 +17,8 @@ export interface BaseToken {
   isSwappable?: boolean
   minimumAppVersionToSwap?: string
   networkIconUrl?: string
+  bridge?: string
+  networkName?: string
```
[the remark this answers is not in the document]
MuckT: A static map in the wallet could be a problem if there's a future fork of a chain such as [BTC](https://www.coingecko.com/en/coins/bitcoin) > [BCH](https://www.coingecko.com/en/coins/bitcoin-cash). Adding it to  address-metadata seems easiest and most configurable.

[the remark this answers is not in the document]
satish-ravi: Controlling this through the backend makes sense to me. Can you leave a todo comment with reference to ACT-923 that should be made required?

src/tokens/TokenBalanceItem.tsx
```diff
@@ -0,0 +1,118 @@
...
+import TokenIcon from 'src/components/TokenIcon'
+import Touchable from 'src/components/Touchable'
+import { getLocalCurrencySymbol, usdToLocalCurrencyRateSelector } from 'src/localCurrency/selectors'
+import useSelector from 'src/redux/useSelector'
+import colors from 'src/styles/colors'
+import { typeScale } from 'src/styles/fonts'
+import { Spacing } from 'src/styles/styles'
+import { TokenBalance } from 'src/tokens/slice'
+
+const TokenBalanceText = (token: TokenBalance) => {
```
satish-ravi, line 15: nit: maybe TokenBalanceInLocalAmount? I'd also suggest exporting this now and have specific tests for this component since we need it in ACT-915, but I can also take it up as part of that issue.

src/analytics/Properties.tsx
```diff
@@ -1341,6 +1341,13 @@ interface AssetsEventsProperties {
         description: string
         balanceUsd: number
       }
+    | {
+        assetType: 'token'
+        tokenId: string
+        title: string // Example: 'cUSD'
+        description: string
+        balanceUsd: number
```
satish-ravi, line 1349: can we include `networkId` and `address` explicitly? It is part of the `tokenId` but the value is meant to be opaque.

src/tokens/TokenBalanceItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+import TokenIcon from 'src/components/TokenIcon'
+import Touchable from 'src/components/Touchable'
+import { getLocalCurrencySymbol, usdToLocalCurrencyRateSelector } from 'src/localCurrency/selectors'
+import useSelector from 'src/redux/useSelector'
+import colors from 'src/styles/colors'
+import { typeScale } from 'src/styles/fonts'
+import { Spacing } from 'src/styles/styles'
+import { TokenBalance } from 'src/tokens/slice'
+
+const TokenBalanceInLocalCurrency = (token: TokenBalance) => {
```
satish-ravi, line 17: I think this should be merged with the refactored TokenDisplay component on #4234 which supports tokenIds. Could be in a follow up after that is merged.

src/tokens/TokenBalanceItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+          </View>
+          <View style={styles.line}>
+            {token.networkName ? (
+              <Text numberOfLines={1} style={styles.subLabel} testID="NetworkLabel">
+                {t('assets.networkName', { networkName: token.networkName })}
+              </Text>
+            ) : (
+              <View />
+            )}
+            {TokenBalanceInLocalCurrency(token)}
```
satish-ravi, line 75:
```suggestion
            <TokenBalanceInLocalCurrency token={token} />
```

src/tokens/TokenBalanceItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+      <View style={styles.container}>
+        <TokenIcon token={token} viewStyle={styles.marginRight} />
+        <View style={styles.textContainer}>
+          <View style={styles.line}>
+            <Text numberOfLines={1} style={[styles.label, styles.marginRight]}>
+              {token.name}
+            </Text>
+            <Text numberOfLines={1} style={styles.amount}>
+              {formatValueToDisplay(token.balance)} {token.symbol}
+            </Text>
```
satish-ravi, line 65: This can also use the refactored TokenDisplay component right?

satish-ravi (commented):
Looking good. Few minor things

src/analytics/Properties.tsx
MuckT: It looks like address is possibly undefined currently.

cajubelt: still seems worth adding, for instance if we wanted to join on an existing table that uses only addresses

src/tokens/TokenBalanceItem.tsx
```diff
@@ -0,0 +1,129 @@
...
+  token: TokenBalance
+  onPress?: () => void
+  containerStyle?: ViewStyle
+}
+
+export const TokenBalanceItem = ({ token, onPress, containerStyle }: Props) => {
+  const { t } = useTranslation()
+
+  return (
+    <Touchable onPress={onPress}>
```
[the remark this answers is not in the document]
MuckT: Conditional rendering added in https://github.com/valora-inc/wallet/pull/4207/commits/97b669f166e403044a81902fbf8d734bcb9b4f62!

src/tokens/TokenBalanceItem.test.tsx
```diff
@@ -0,0 +1,104 @@
...
+      </Provider>
+    )
+
+    expect(getByText('Celo Dollar')).toBeTruthy()
+    expect(getByText('10.00 cUSD')).toBeTruthy()
+    expect(getByText('₱13.30')).toBeTruthy()
+    expect(getByTestId('NetworkLabel')).toBeTruthy()
+  })
+
+  it('tracks data about the asset when tapped', () => {
```
satish-ravi, line 72: This test is for actually checking if the component is touchable if onPress is passed and invokes it on clicking right? And not specific to analytics. Could we rename this test?

src/tokens/TokenBalanceItem.tsx
```diff
@@ -0,0 +1,122 @@
...
+            >
+              {t('assets.bridge', { bridge: token.bridge })}
+            </Text>
+          )}
+        </View>
+      </View>
+    )
+  }
+
+  return typeof onPress !== undefined ? (
```
satish-ravi, line 74: nit:
```suggestion
  return onPress ? (
```

src/tokens/TokenBalanceItem.tsx
```diff
@@ -0,0 +1,122 @@
...
+    </Touchable>
+  ) : (
+    <Content token={token} containerStyle={containerStyle} />
+  )
+}
+
+const styles = StyleSheet.create({
+  container: {
+    marginHorizontal: Spacing.Thick24,
+    marginBottom: Spacing.Large32,
```
satish-ravi, line 86: Not super important since the styles are overridable, but could the marginBottom be split evenly to the top and the bottom? This would also allow us to re-use default styles on the Token Details screen and also the touchable area is split evenly instead of favoring one side.

MuckT: This might run into issues with vertical margin collapse where marginTop and marginBottom of adjacent items merge together, but I'll give it a try.

src/shared/conts.ts
```diff
@@ -0,0 +1,6 @@
+export const NETWORK_NAMES: Record<string, string> = {
```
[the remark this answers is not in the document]
MuckT: Using the enum is a lot safer and already caught a misspelling. Fixed in https://github.com/valora-inc/wallet/pull/4207/commits/64a98e1cfcfeb1c1dfd936857868c14f8f121f3c.

src/tokens/TokenBalanceItem.tsx
MuckT: Tested locally and it appears react native doesn't have this same issue that web does. Fixed in https://github.com/valora-inc/wallet/pull/4207/commits/465dff7375c3e506f1bffd0d6645276158755d19!

src/shared/conts.ts
```diff
@@ -0,0 +1,8 @@
+import { NetworkId } from 'src/transactions/types'
+
+export const NETWORK_NAMES: Record<NetworkId, string> = {
+  ['celo-alfajores']: 'Celo Alfajores',
```
satish-ravi, line 4: nit: Let's use the enum keys here. I think this currently checks the enum values. E.g.,
```suggestion
  [NetworkId['celo-alfajores']]: 'Celo Alfajores',
```
Feels a bit redundant mainly because the keys are also in kebab case, if it were not kebab case it would've been something like `[NetworkId.celoAlfajores]`

src/tokens/TokenBalanceItem.tsx
```diff
@@ -0,0 +1,123 @@
...
+  ) : (
+    <Content token={token} containerStyle={containerStyle} />
+  )
+}
+
+const styles = StyleSheet.create({
+  container: {
+    marginHorizontal: Spacing.Thick24,
+    marginTop: Spacing.Regular16,
+    marginBottom: Spacing.Regular16,
```
satish-ravi, line 87: nit:
```suggestion
    marginVertical: Spacing.Regular16,
```

satish-ravi (approved):

PR #4210, opened 2023-09-19
feat(tokens): Use new token metadata source/schema and fetch from blockchain-api

jophish:
### Description

For ACT-909. Updates the wallet to fetch the new format of token metadata from the new REST endpoint on blockchain-api and user balances for all available chains (if multi-chain balances are enabled via statsig). As soon as the new blockchain-api REST endpoint is updated to return non-Celo token metadata, multi-chain balances should work out-of-the-box in-app. In short, there are two functional changes in this PR:

* Token balances are fetched for all supported chains, provided an override has been set in Statsig for the `fetch_multi_chain_balances` feature gate.
* We no longer fetch token metadata directly from Firebase, but instead the new blockchain-api endpoint.
  * Internal (Redux) state has been updated to reflect the schema changes in the token metadata.

The second bullet above required some rejiggering of types throughout the repo, and a lot of mechanical changes to tests throughout the app. Existing places in code that depend on a token's address being available have been updated to use the "new" (but identical to the "old") `WithAddress` suffixed types. New code which supports multi-chain features should begin using the types that make address optional. Currently, there are no places in code that take advantage of this new format; we'll need to add a selector that returns the "new format" data from Redux, typed correctly.

Note that this change requires a migration; I have specifically opted _not_ to muck about with trying to migrate stored token information in Redux (re: the trouble caused by attempting to do so with transfers), and instead opted to wipe the stored token info. This means that users will need to refresh their balances upon updating.

### Test plan

Unit and manual tested.

### Related issues

- Fixes ACT-909.

### Backwards compatibility

Yes.


test/schemas.ts
```diff
@@ -2548,6 +2559,27 @@ export const v148Schema = {
   },
 }
 
+export const v149Schema = {
```
jophish, line 2562: Note that this "migration" is only for _test_ data and is not an _actual_ migration. A ton of our tests implicitly rely on `tokenBalances` being present (duh) and fail in unusual ways if it's empty. An actual migration would be a bit more onerous and error prone (re: the recent issue with transfer migrations) so it's an "empty" migration as stated in the PR description.

cajubelt:
please don't merge this until the blockchain api endpoint includes prices. I don't want to create a race condition between the completion of Jacob's blockchain-api pricing issue and this PR

src/statsig/constants.ts
```diff
@@ -37,6 +37,7 @@ export const FeatureGates = {
   [StatsigFeatureGates.SHOW_MULTI_CHAIN_TRANSFERS]: false,
   [StatsigFeatureGates.SHOW_NATIVE_TOKENS]: false,
   [StatsigFeatureGates.SHOW_ETH_IN_CICO]: false,
+  [StatsigFeatureGates.FETCH_MULTI_CHAIN_BALANCES]: false,
```
cajubelt, line 40: I have some concerns about the naming of this, and possibly the usage of it; I haven't gotten that far yet.

I'd guess that we'll want to continue to add chains, rather than revert to a single chain some time after shipping Ethereum. If that's the case, there will be some cleanup after shipping at 100%. Likely we wouldn't get to this cleanup until we're trying to ship an additional chain, at which point we would replace the set of feature gates for the chain that was rolled out with the next chain we want to ship. 

This gate is named as if we'll continue to add chains with a single feature gate, maintaining the option to revert back to a single chain. I just don't see that happening.

cajubelt: same concern applies to the show_multi_chain_transfers feature flag

cajubelt: I have a separate, less significant concern about the "fetch" part of the feature gate name-- almost makes it sounds like there will be some other feature gate for whether to actually show the multichain balances.

cajubelt (commented):
partial review for up to statsig/constants.ts

src/tokens/saga.ts
```diff
@@ -233,25 +238,47 @@ interface UserBalancesResponse {
...
+    })
+  )
+  return userBalances.reduce(
+    (acc, response) => acc.concat(response.data.userBalances.balances),
+    [] as FetchedTokenBalance[]
+  )
+}
+
+export async function getTokensInfo(): Promise<StoredTokenBalances> {
+  const response = await fetchWithTimeout(networkConfig.getTokensInfoUrl)
```
cajubelt, line 274: I think codecov is probably right that this could use a unit test

src/tokens/saga.ts
```diff
@@ -233,25 +238,47 @@ interface UserBalancesResponse {
...
+        },
+        fetchPolicy: 'network-only',
+        errorPolicy: 'all',
+      })
+    })
+  )
+  return userBalances.reduce(
+    (acc, response) => acc.concat(response.data.userBalances.balances),
+    [] as FetchedTokenBalance[]
+  )
```
cajubelt, line 270:
```suggestion
  return userBalances.flatMap(
    (response) => response.data.userBalances.balances
  )
```
code golf

src/tokens/saga.ts
```diff
@@ -233,25 +238,47 @@ interface UserBalancesResponse {
...
   address: string
 ): Promise<FetchedTokenBalance[]> {
-  const response = await apolloClient.query<UserBalancesResponse, { address: string }>({
-    query: gql`
-      query FetchUserBalances($address: Address!) {
-        userBalances(address: $address) {
-          balances {
-            tokenAddress
-            balance
+  const chainsToFetch = getFeatureGate(StatsigFeatureGates.FETCH_MULTI_CHAIN_BALANCES)
```
cajubelt, line 241: this function should probably have a unit test

src/tokens/slice.ts
```diff
@@ -33,23 +37,42 @@ export interface StoredTokenBalance extends BaseToken {
...
 export interface TokenBalance extends BaseToken {
   balance: BigNumber
   usdPrice: BigNumber | null
   lastKnownUsdPrice: BigNumber | null
   historicalUsdPrices?: HistoricalUsdPrices
 }
 
+// The "WithAddress" suffixed types are legacy types, for places in the wallet
+// that require an address to be present. As we move to multichain, (where address
+// is not guaranteed,) existing code should be updated to use the "address optional" types.
```
cajubelt, line 53: should @deprecated be used in case people don't read this?

src/web3/networkConfig.ts
```diff
@@ -70,6 +71,9 @@ interface NetworkConfig {
 const CLOUD_FUNCTIONS_STAGING = 'https://api.alfajores.valora.xyz'
 const CLOUD_FUNCTIONS_MAINNET = 'https://api.mainnet.valora.xyz'
 
+const BLOCKCHAIN_API_STAGING = 'https://blockchain-api-dot-celo-mobile-alfajores.appspot.com'
```
cajubelt, line 74: I appreciate the effort to decouple a test environment from the alfajores testnet, but I don't know if I've used the "staging" environment name elsewhere at Valora. How about "dev" or "development"?

test/RootStateSchema.json
```diff
@@ -4273,6 +4276,9 @@
                 "isCoreToken": {
                     "type": "boolean"
                 },
+                "isNative": {
+                    "type": "boolean"
```
cajubelt, line 4280: I think isNative is optional

src/tokens/saga.ts
```diff
@@ -233,25 +238,47 @@ interface UserBalancesResponse {
...
+      })
+    })
+  )
+  return userBalances.reduce(
+    (acc, response) => acc.concat(response.data.userBalances.balances),
+    [] as FetchedTokenBalance[]
+  )
+}
+
+export async function getTokensInfo(): Promise<StoredTokenBalances> {
```
jh2oman, line 273: I think i found a problem with the existing `StoredTokenBalance` type. `usdPrice` can actually be undefined and is quite often for tokens that we don't have any price data for. The current `StoredTokenBalance` type says its always there. We should probably fix this now

jophish: Ah, yeah -- that's a good callout. Interestingly, `historicalUsdPrice` is already optional. I'll wait until Jean's branch is merged into this one before making this change, in case there are downstream consumers whose access patterns need to be modified in response to the types changing.

test/RootStateSchema.json
jophish: It is optional -- it's marked as such in the `BaseToken` type.

src/web3/networkConfig.ts
jophish: I was following the `CLOUD_FUNCTIONS_STAGING` example from above (and elsewhere in this file). The file itself is somewhat inconsistent between `STAGING` and `ALFAJORES`, so I opted for `STAGING` here rather than introducing a new convention.

src/tokens/saga.ts
jophish: ohh sick -- wouldn't have thought to use `flatMap`!

src/tokens/saga.ts
jophish: Actually, after talking with Jean, I think we're gonna do:
* Merge this PR into main
* Merge Jean's PR into main
* Make a new PR afterwards making `priceUsd` optional

I'll get a PR out for that once both of our PRs have been merged.

src/statsig/constants.ts
jophish: Interesting -- this is a good point. So, you're basically saying that, steady state, in production, we'll support some number of chains, while more chains could be "in development" s.t. we want access to them on dev builds/wallets, but don't necessarily want them to be available in production?

That use case above is _kind of_ what I had in mind with the `supportedChains` concept from an earlier PR; e.g., the wallet maintains a "full" list of chains that it knows about, some _subset_ of which are available in production. Overriding a feature gate in Statsig would allow the user to see the "full" list of chains in-app. Both the full list of chains, and the subset of "prod-available" chains, could be configured remotely by a dynamic config in Statsig. I think this would also mean we don't need new feature gates for each chain we want to roll out; we'd just add it to the prod-available subset when we're ready. (Actually, we could probably do this with a single dynamic config that contains prod-available chains, and specify overrides/targeting rules to return an expanded list with chains that are in development -- thinking out loud here.)

This of course assumes that we don't want finer granularity in multi-chain features, e.g., we wouldn't be able to turn on/off token balance or transfer feed support for an arbitrary chain. I guess we could follow the above pattern per-feature, if we wanted to.

jophish: Discussed this offline; a more comprehensive solution will come in a followup PR; will keep this as-is for now.

cajubelt (approved):

MuckT (approved):

src/statsig/index.ts
```diff
@@ -142,7 +142,7 @@ export function setupOverridesFromLaunchArgs() {
     const { statsigGateOverrides } = LaunchArguments.value<ExpectedLaunchArgs>()
     if (statsigGateOverrides) {
       Logger.debug(TAG, 'Setting up gate overrides', statsigGateOverrides)
-      statsigGateOverrides.split(',').forEach((gateOverride) => {
+      statsigGateOverrides.split(',').forEach((gateOverride: string) => {
```
satish-ravi, line 145: is this change required? `split`'s type should guarantee `string[]` as the return type right?

src/tokens/saga.ts
satish-ravi, line 241: Reading the PR title, I assumed this PR was just updating the token metadata source. This seems like a big change that should also go in the PR title. Can we update the title to reflect that? (Would be ideal if fetching multi chain balances is in its own PR)

src/tokens/AssetItem.tsx
satish-ravi, line 1: For the TokenBalanceItem here to remain backwards compatible per this plan https://valora-app.slack.com/archives/C029Z1QMD7B/p1695234205410489?thread_ts=1695147118.671939&cid=C029Z1QMD7B, should this PR also read the separate bridge field and append to the name? Although the blockchain-api still currently reads from firebase, so it doesn't return a `bridge` field separately and still includes the bridge info in the name field itself. So we should either switch the blockchain-api to return the new format before the wallet starts reading it, or deal with the backwards compatibility when we switch blockchain-api to stop reading from firebase. 
cc @cajubelt

src/statsig/index.ts
jophish: I'm not sure why this was needed; my TS compiler was yelling at me without it.

src/tokens/AssetItem.tsx
jophish: Yeah, this is something @cajubelt and I discussed. I was planning on adding that to this PR. I believe all that needs to be done for this here is just updating the old selectors that ensure `address` is present to construct the name dynamically based off of the presence of a `bridge` field in the Redux state.

src/tokens/saga.ts
jophish: Yeah, title has been updated; I agree that this likely should have been split out into separate PRs; this has gotten a bit bigger than anticipated.

PR #4226, opened 2023-09-22
feat(assets): implement new assets screen

satish-ravi:
### Description

Adds a new Assets screen based on the existing TokenBalances screen. Made a new screen to avoid adding more conditionals to the existing screen, which already has a bunch of flags which have been rolled out. 

[Figma](https://www.figma.com/file/OE6RFzqXAMqU9ygHtNYyof/Assets-Page-(Multichain)?node-id=2099%3A16300&mode=dev)

### Test plan

Manually by enabling the dummy flag

iOS:

https://github.com/valora-inc/wallet/assets/5062591/78115b9b-7ead-4388-be6a-511295fa8404

Android:

https://github.com/valora-inc/wallet/assets/5062591/2f9a43c2-93e9-4ec6-ac40-a98f241ae314

Large font size:

No Positions:

### Related issues

- Fixes ACT-911

### Backwards compatibility

Yes


src/tokens/Assets.tsx
satish-ravi, line 1: This is mostly copied from TokenBalances.tsx, use this diff for easier review: https://gist.github.com/satish-ravi/0e91976b020970fc8cfec23a89ae6cd3

src/tokens/Assets.tsx
```diff
@@ -0,0 +1,425 @@
...
+  const tokensAreStale = useSelector(stalePriceSelector)
+  const insets = useSafeAreaInsets()
+
+  const positions = useSelector(positionsSelector)
+  const showPositions = getFeatureGate(StatsigFeatureGates.SHOW_POSITIONS)
+  const displayPositions = showPositions && positions.length > 0
+
+  const dappShortcutsEnabled = getFeatureGate(StatsigFeatureGates.SHOW_CLAIM_SHORTCUTS)
+  const positionsWithClaimableRewards = useSelector(positionsWithClaimableRewardsSelector)
+  const showClaimRewards =
```
jophish, line 99: Is this meant to render on _both_ the assets and dapp positions tabs?

src/tokens/Assets.tsx
```diff
@@ -0,0 +1,425 @@
...
+          <Text style={styles.positionSectionHeaderText}>
+            {section.appName.toLocaleUpperCase()}
+          </Text>
+        </View>
+      )
+    }
+    return null
+  }
+
+  const renderAssetItem = ({ item }: { item: TokenBalance | Position }) => {
```
jophish, line 258: I understand this logic is largely copied, but this isn't the easiest thing to wrap my head around (having `item` being a union and then disambiguating with a type guard).

satish-ravi: yeah I didn't find this easy to read as well. The renderSectionHeader function also does something similar. The item being union type starts from this type here https://github.com/valora-inc/wallet/blob/0e5eb6cf6a7786570aba3dcc6cc861eefc6eb800/src/tokens/Assets.tsx#L63-L65

So even if we make two separate list instances with specific functions, the renderAssetItem function's `item` type will be of the union type. I attempted making the `AnimatedSectionList` a generic type but the changes were getting large, so I retained the existing code. One thing we could do is make two explicit `AnimatedSectionList` with specific types, but let me know if you have other ideas

src/tokens/Assets.tsx
satish-ravi: Confirmed that we want to keep this on the assets and dapp positions page

src/tokens/Assets.tsx
```diff
@@ -0,0 +1,442 @@
...
+
+  const scrollPosition = useSharedValue(0)
+  const footerPosition = useSharedValue(0)
+  const handleScroll = useAnimatedScrollHandler<{ prevScrollY: number }>(
+    {
+      onScroll: (event, ctx) => {
+        const scrollY = event.contentOffset.y
+        scrollPosition.value = scrollY
+
+        function clamp(value: number, min: number, max: number) {
```
jophish, line 125: I think lodash [has a clamp method](https://www.geeksforgeeks.org/lodash-_-clamp-method/) built in.

src/tokens/Assets.tsx
```diff
@@ -0,0 +1,442 @@
...
+    scrollPosition,
+    startFadeInPosition: nonStickyHeaderHeight - HEADER_OPACITY_ANIMATION_START_OFFSET,
+    animationDistance: HEADER_OPACITY_ANIMATION_DISTANCE,
+  })
+
+  const handleMeasureNonStickyHeaderHeight = (event: LayoutChangeEvent) => {
+    setNonStickyHeaderHeight(event.nativeEvent.layout.height)
+  }
+
+  const handleMeasureListHeaderHeight = (event: LayoutChangeEvent) => {
```
jophish, line 206: These "measure" callbacks are only called _once_ right? Their purpose is to just set an initial value in state for the section heights?

src/tokens/Assets.tsx
```diff
@@ -0,0 +1,442 @@
...
+    setListHeaderHeight(event.nativeEvent.layout.height)
+  }
+
+  const handleMeasureListFooterHeight = (event: LayoutChangeEvent) => {
```
jophish, line 210: This "footer" is not always visible -- just when a user can claim rewards, right? Can you share a video of the footer being shown?

src/tokens/Assets.tsx
```diff
@@ -0,0 +1,442 @@
...
+export enum AssetTabType {
+  Tokens = 0,
+  Collectibles = 1,
+  Positions = 2,
+}
+
+// offset relative to the bottom of the non sticky header component, where the
+// screen header opacity animation starts
+const HEADER_OPACITY_ANIMATION_START_OFFSET = 44
+// distance in points over which the screen header opacity animation is applied
```
jophish, line 80: By 'points' do you mean 'pixels' here?

src/tokens/Assets.tsx
```diff
@@ -0,0 +1,442 @@
...
+  const totalPositionsBalanceLocal = useDollarsToLocalAmount(totalPositionsBalanceUsd)
+  const totalBalanceLocal = totalTokenBalanceLocal?.plus(totalPositionsBalanceLocal ?? 0)
+
+  const [nonStickyHeaderHeight, setNonStickyHeaderHeight] = useState(0)
+  const [listHeaderHeight, setListHeaderHeight] = useState(0)
+  const [listFooterHeight, setListFooterHeight] = useState(0)
+
+  const scrollPosition = useSharedValue(0)
+  const footerPosition = useSharedValue(0)
+  const handleScroll = useAnimatedScrollHandler<{ prevScrollY: number }>(
```
jophish, line 119: It seems like this just handles scroll as normal but takes into account the potential existence of a footer -- what happens without this handler?

src/tokens/Assets.tsx
```diff
@@ -0,0 +1,442 @@
...
+        ctx.prevScrollY = clampedScrollY
+      },
+      onBeginDrag: (event, ctx) => {
+        ctx.prevScrollY = event.contentOffset.y
+      },
+    },
+    [listFooterHeight]
+  )
+
+  const animatedListHeaderStyles = useAnimatedStyle(() => {
```
jophish, line 148: This just disables the shadow if `nonStickyHeaderHeight` is 0 and sets a shadow if it's visible? What triggers that change to happen?

src/tokens/Assets.tsx
```diff
@@ -0,0 +1,442 @@
...
+    return {
+      transform: [
+        {
+          translateY: footerPosition.value,
+        },
+      ],
+    }
+  }, [footerPosition.value])
+
+  useScrollAwareHeader({
```
jophish, line 187: I've never used `useScrollAwareHeader` before -- what purpose does this serve? Basically to just have a header that hides when the user scrolls?

jophish (commented):
Ignore my previous comments -- just looked at your diff this time :laughing:

src/tokens/Assets.tsx
```diff
@@ -0,0 +1,442 @@
...
+      },
+      onBeginDrag: (event, ctx) => {
+        ctx.prevScrollY = event.contentOffset.y
+      },
+    },
+    [listFooterHeight]
+  )
+
+  const animatedListHeaderStyles = useAnimatedStyle(() => {
+    if (nonStickyHeaderHeight === 0) {
```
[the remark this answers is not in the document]
satish-ravi: Because the segmented control was visible only when `displayPositions` was true whereas now the TabBar is always visible and we want it to be sticky

satish-ravi:
> Ignore my previous comments -- just looked at your diff this time 😆

No worries, these are still good questions, maybe @kathaypacific, who looks like was the original author, can help answer?

jophish (approved):
LGTM!

src/tokens/Assets.tsx
```diff
@@ -0,0 +1,443 @@
...
+        <Touchable
+          testID="Assets/TabBarItem"
+          key={value}
+          onPress={handleSelectOption(value, index)}
+          style={styles.tabBarTouchable}
+        >
+          <Text
+            style={[index === selectedIndex ? styles.tabBarItemSelected : styles.tabBarItem]}
+            numberOfLines={1}
+            ellipsizeMode="tail"
```
MuckT, line 386: nit: "tail" is the default: [docs](https://reactnative.dev/docs/text#ellipsizemode).

MuckT (approved):

src/navigator/Navigator.tsx
```diff
@@ -615,6 +626,7 @@ export function MainStackScreen() {
       {generalScreens(Stack)}
       {swapScreens(Stack)}
       {nftScreens(Stack)}
+      {assetScreens(Stack)}
```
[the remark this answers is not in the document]
jeanregisser: They are all part of the same stack.

PR #4230, opened 2023-09-27
test(e2e): use real tokens list

jeanregisser:
### Description

Now that https://github.com/valora-inc/wallet/pull/4210 has been merged we can now fetch the real tokens list.

This should give us more confidence as it exercises more of the real code (including backend endpoint).

### Test plan

- e2e tests pass

### Related issues

- N/A

### Backwards compatibility

Yes


bakoushin (approved):

jeanregisser:
This change actually uncovered an old bug with pay deep links. Investigating...

PR #4231, opened 2023-09-27
fix: do not show swap warning when swap pair is not selected

kathaypacific:
### Description

Noticed this when testing.

### Test plan

Before:
Simulator Screenshot - iPhone 14 Pro - 2023-09-27 at 15 51 02

After:
Simulator Screenshot - iPhone 14 Pro - 2023-09-27 at 15 50 39

### Related issues

n/a

### Backwards compatibility

Y


bakoushin (approved):
LGTM!

PR #4232, opened 2023-09-27
fix: pay deeplinks showing and sending incorrect amounts

jeanregisser:
### Description

This fixes a regression with `celo://wallet/pay?...` deeplinks which happened when we switched to support multi token a long time ago.

This got uncovered working on https://github.com/valora-inc/wallet/pull/4230, blocking the e2e tests there.

### Test plan

Only the unit tests needed to be updated here.
The e2e tests couldn't uncover this bug because of the hardcoded cUSD price there ([1 cUSD is 1 USD](https://github.com/valora-inc/wallet/blob/2e91e5b8749114ea77fa46d7f969e2ad0223b050/src/tokens/e2eTokens.ts#L18)), and the local currency selected being USD (maybe we should change that?).

With the following deeplink:
```
celo://wallet/pay?address=0xC0509a7dcc69a0B28c7Ca73feD2FF06b9d59E5b9&amount=0.1&currencyCode=USD&token=cUSD&displayName=TestFaucet&comment=sending+usd:+0.1+to+my+wallet
```

The expected effect is that the equivalent of 0.1 USD in cUSD will be sent.
Since this is on Alfajores, where 1 cUSD is 1.5 USD, the sent cUSD amount should be 0.1 / 1.5 = 0.0667.

Note that the screenshots were taken on a device where the local currency is PHP. So it's expected to see most values shown in PHP.

**before/after**

With the following deeplink:
```
celo://wallet/pay?address=0xC0509a7dcc69a0B28c7Ca73feD2FF06b9d59E5b9&amount=0.1&token=cUSD&displayName=TestFaucet&comment=sending+usd:+0.1+to+my+wallet
```
The expected effect is that the equivalent of 0.1 PHP (local currency, since `currencyCode` isn't set explicitly) in cUSD will be sent.
Since this is on Alfajores, where 1 USD is 56.931999 PHP and 1 cUSD is 1.5 USD, the sent cUSD amount should be 0.1 / 56.931999 / 1.5  = 0.0012.

Again, the screenshots were taken on a device where the local currency is PHP. So it's expected to see most values shown in PHP.

Please also note that the comment should be "sending local: 0.1 to my wallet" to be correct and less confusing.

**before/after**

Final note, there seem to be another bug with the local amount at the top + fee not adding up to the total displayed. This isn't something new though. We can dig and fix this separately.

### Related issues

- N/A

### Backwards compatibility

Pay deeplinks amounts will now work as [documented](https://github.com/valora-inc/wallet/blob/main/docs/deeplinks.md).

### DIFF
--- a/src/send/utils.test.ts
+++ b/src/send/utils.test.ts
@@ -41,10 +41,6 @@ describe('send/utils', () => {
     it('should navigate to SendAmount screen when no amount nor token is sent', async () => {
       await expectSaga(handleSendPaymentData, mockData, false, undefined, undefined)
         .withState(createMockStore({}).getState())
-        .provide([
-          [matchers.call.fn(fetchExchangeRate), '1'],
-          [select(usdToLocalCurrencyRateSelector), '1'],
-        ])
         .run()
       expect(navigate).toHaveBeenCalledWith(
         Screens.SendAmount,
@@ -66,10 +62,6 @@ describe('send/utils', () => {
         undefined
       )
         .withState(createMockStore({}).getState())
-        .provide([
-          [matchers.call.fn(fetchExchangeRate), '1'],
-          [select(usdToLocalCurrencyRateSelector), '1'],
-        ])
         .run()
       expect(navigate).toHaveBeenCalledWith(
         Screens.SendAmount,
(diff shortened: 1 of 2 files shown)


MuckT (approved):
Thanks for finding and fixing this 🧑‍🎨

PR #4234, opened 2023-09-27
feat(multichain): Enable multi-chain balances

jophish:
### Description

For ACT-910. Enables rendering of multi-chain balances in the wallet, if enabled for the user in Statsig. This PR introduces a host of new selectors that allow the wallet to access the new token metadata, indexed by `tokenId`. Where necessary for this PR, our existing selectors (i.e., with their existing names) have had their implementations changed to use `tokenId`-based selectors, and the old implementations have been moved to `WithAddress`-infixed selectors, whose usage is marked deprecated.

I've also given the same treatment to our `TokenDisplay` component. Existing usages of `TokenDisplay` have been changed to be `LegacyTokenDisplay` and deprecated, and `TokenDisplay` now accepts `tokenId` instead of `address`. While the implementation is pretty similar, some offline discussion suggested we pull this out into its own component rather than trying to stuff more functionality into the old `TokenDisplay` in a backwards-compatible way.

A note on selectors/hooks: from here on out, individual features throughout the app will potentially want to support different sets of networks. In order to support this, the "new" versions of selectors are actually created on-the-fly via higher order functions which accept a list of `NetworkId`s to filter on. Since this pattern is slightly wonky (passing paramters to selectors), it is actually _not_ exposed directly to components. Instead, it is wrapped inside of _hooks_. New components should use these hooks rather than accessing the higher-order selectors directly. This is called out in a comment in the `src/tokens/selectors.ts` file.

### Test plan

Manual and unit tested.

### Related issues

- Fixes ACT-910.

### Backwards compatibility

Yes.


src/tokens/hooks.ts
```diff
@@ -4,17 +4,34 @@ import useSelector from 'src/redux/useSelector'
...
+ * @deprecated use useTokenInfo and select using tokenId
+ */
+export function useTokenInfoByAddress(tokenAddress?: string | null) {
   // TODO: Make this work with native tokens lacking a tokenAddress
   const tokens = useSelector(tokensByAddressSelector)
   return tokenAddress ? tokens[tokenAddress] : undefined
 }
 
+export function useTokenInfo(tokenId: string) {
+  return useSelector(tokensByIdSelector)[tokenId]
```
MuckT, line 24: We try to avoid chaining off selectors, but I can't remember exactly why, so this should be something like:

```TypeScript
export function useTokenInfo(tokenId: string) {
  const tokens = useSelector(tokensByIdSelector)
  return tokens[tokenId]
}

File: src/tokens/saga.ts [hunk 2]
```diff
@@ -17,9 +17,10 @@ import { FeeInfo } from 'src/fees/saga'
 import { SentryTransactionHub } from 'src/sentry/SentryTransactionHub'
 import { SentryTransaction } from 'src/sentry/SentryTransactions'
 import { Actions } from 'src/stableToken/actions'
+
```
MuckT, line 20: nit: extra space

src/components/NonNativeTokenDisplay.tsx
MuckT, line 1: Since this display can technically display CELO and ETH, would a better name maybe be `LegacyTokenDisplay`?

MuckT (approved):

MuckT:
@jophish should we wait to merge until after https://github.com/valora-inc/wallet/pull/4236 is merged to avoid issues with the release cut tomorrow?

jophish:
@MuckT This luckily doesn't need to make it into the release cut since it's a new feature and not a fix, so there's no rush with this one.

src/tokens/selectors.ts
```diff
@@ -50,12 +80,22 @@ export const tokensByAddressSelector = createSelector(
   }
 )
 
-export const tokensListSelector = createSelector(tokensByAddressSelector, (tokens) => {
+export const tokensListSelector = createSelector(tokensByIdSelector, (tokens) => {
```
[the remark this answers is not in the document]
cajubelt: looks like you renamed the old version before re-using the same name... leaving this comment here in case, though

src/tokens/hooks.ts
MuckT: I don't see anywhere else in the repo we chain off the useSelector hook. Maybe it's a style choice, but this isn't a pattern we've used in the repo previously.

src/tokens/selectors.ts
```diff
@@ -24,7 +29,32 @@ export type CurrencyTokens = {
...
+export const tokensByIdSelector = createSelector(
+  (state: RootState) => state.tokens.tokenBalances,
+  (storedBalances) => {
+    const tokenBalances: TokenBalances = {}
+    for (const storedState of Object.values(storedBalances)) {
+      if (!storedState || storedState.balance === null) {
+        continue
+      }
+      const priceUsd = new BigNumber(storedState.priceUsd)
+      const tokenPriceUsdIsStale =
```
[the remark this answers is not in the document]
jophish: I believe that's the intent, yes -- this behavior is carried over from the previous selector, but if we don't care about it, it should be fine to remove.

src/tokens/TokenBalances.tsx
```diff
@@ -271,7 +271,14 @@ function TokenBalancesScreen({ navigation, route }: Props) {
     return null
   }
 
-  const renderAssetItem = ({ item }: { item: TokenBalanceWithAddress | Position }) => {
+  const keyExtractor = (item: TokenBalance | Position) => {
+    if (assetIsPosition(item)) {
+      return item.address
```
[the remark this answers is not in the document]
jophish: I've kept positions out of scope for this -- I've added in TODO comments where positions selectors are used to ensure that they're updated once positions support multi-chain.

src/tokens/hooks.ts
jophish: I'm not sure I understand the difference here besides style -- I've updated the usage to match your recommendation here regardless.

src/components/NonNativeTokenDisplay.tsx
jophish: Yeah... `LegacyTokenDisplay` is a way better name -- will change it throughout!

src/components/LegacyTokenDisplay.test.tsx
```diff
@@ -0,0 +1,282 @@
...
+    return createMockStore({
+      localCurrency: {
+        usdToLocalRate: '0.10',
+        preferredCurrencyCode: LocalCurrencyCode.BRL,
+        fetchedCurrencyCode: LocalCurrencyCode.BRL,
+      },
+      tokens: {
+        tokenBalances: {
+          ['0xusd']: {
+            address: '0xusd',
```
MuckT, line 29: nit: use `mockCusdTokenId` from `test/values.ts`

src/components/LegacyTokenDisplay.tsx
```diff
@@ -0,0 +1,113 @@
...
+    throw new Error(
+      'LegacyTokenDisplay must be passed either "currency" or "tokenAddress" and not both'
+    )
+  } else if (tokenAddress && currency) {
+    throw new Error(
+      'LegacyTokenDisplay must be passed tokenAddress, currency, or nethier, but not both'
+    )
+  }
+
+  const tokenInfoFromAddress = useTokenInfoByAddress(tokenAddress!)
```
MuckT, line 73: `tokenAddress!` > `tokenAddress`

MuckT (approved):

src/components/TokenDisplay.tsx
```diff
@@ -25,29 +20,9 @@ interface Props {
...
-  const exponent = value?.e ?? 0
-  if (exponent >= 0) {
-    return DEFAULT_DISPLAY_DECIMALS
-  }
-
-  return Math.abs(exponent) + 1
-}
-
-// Formats |value| so that it shows at least 2 significant figures and at least 2 decimal places without trailing zeros.
-export function formatValueToDisplay(value: BigNumber) {
```
satish-ravi, line 38: this should eventually exist in this file right? When LegacyTokenDisplay usages are removed and cleaned up

src/tokens/hooks.ts
```diff
@@ -4,19 +4,73 @@ import useSelector from 'src/redux/useSelector'
...
 
-export function useTokenInfoBySymbol(symbol: string) {
-  const tokens = useSelector(tokensListSelector)
+export function useTokensWithUsdValue(networkIds: NetworkId[]) {
+  return useSelector(tokensWithUsdValueSelectorWrapper(networkIds))
+}
+
+export function useTotalTokenBalance(networkIds: NetworkId[]) {
+  return useSelector(totalTokenBalanceSelectorWrapper(networkIds))
+}
```
satish-ravi, line 34: I wonder if this should take networkIds or just filter by network ids supported for balance. One the one hand its nice that all of the functions here take a networkId, but is there a use case for when we need use different network ids for total token balance in different parts of the app? Also, I guess we'd need some wrapper for getting tokens eligible for send / swap / CICO which can also wrap the statsig logic and doesn't need to take network ids

src/tokens/TokenBalances.tsx
```diff
@@ -47,28 +47,29 @@ import fontStyles from 'src/styles/fonts'
 import { getShadowStyle, Shadow, Spacing } from 'src/styles/styles'
```
satish-ravi, line 47: probably not necessary to change now, but I guess we could just do this change in the new Assets screen, since this is anyway going to be deprecated as part of the assets redesign work. I can copy these changes to #4226

src/analytics/selectors.ts
```diff
@@ -16,12 +16,12 @@ import {
...
   totalPositionsBalanceUsdSelector,
 } from 'src/positions/selectors'
-import { coreTokensSelector, tokensWithTokenBalanceSelector } from 'src/tokens/selectors'
+import { coreTokensSelector, tokensWithTokenBalanceAndAddressSelector } from 'src/tokens/selectors'
 import { sortByUsdBalance } from 'src/tokens/utils'
 import { mtwAddressSelector, rawWalletAddressSelector } from 'src/web3/selectors'
 
 const tokensSelector = createSelector(
-  [tokensWithTokenBalanceSelector, coreTokensSelector],
+  [tokensWithTokenBalanceAndAddressSelector, coreTokensSelector],
```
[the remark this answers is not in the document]
jophish: Yeah, good call -- I'll write an issue for this.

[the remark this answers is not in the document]
jophish: https://linear.app/valora/issue/ACT-935/placeholder-update-analytics-selectors-for-multichain

src/tokens/utils.ts
```diff
@@ -114,3 +118,9 @@ export function convertTokenToLocalAmount({
 
   return tokenAmount.multipliedBy(tokenPriceUsd).multipliedBy(usdToLocalRate)
 }
+
+export function getSupportedNetworkIdsForTokenBalances(): NetworkId[] {
+  return getFeatureGate(StatsigFeatureGates.FETCH_MULTI_CHAIN_BALANCES)
+    ? Object.values(networkConfig.networkToNetworkId)
```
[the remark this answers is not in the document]
jophish: Do you mean network ids? The dynamic config will contain a list of supported network ids for balances, right?

src/tokens/hooks.ts
jophish: Yeah, I think this is probably a good idea -- I've updated the implementation. We'll still need to get the supported networks from Statsig directly from the components, because the components that use `useTotalTokenBalance` _also_ use other hooks that _do_ require a list of network IDs (e.g. `useTokensWithUsdValue`). Despite this, I think there's no harm in having the Statsig logic within `useTotalTokenBalance` as, it's very possible we use this elsewhere in the app, and this lowers the risk of passing the "wrong" network IDs.

jophish: I've also added this treatment to `useTokensWithTokenBalance` as well, since I don't imagine we'll want _individual_ token balances to appear for all chains but not be reflected in _total_ balance, or vice versa.

src/tokens/utils.ts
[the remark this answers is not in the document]
satish-ravi: oh yeah, I meant network ids here.

satish-ravi (approved):

PR #4235, opened 2023-09-27
feat(recipient): Remove v from recipients/contacts

finnian0826:
### Description

Remove the V icon from recipients/contacts who are on Valora.

Reason for the change: [Linear issue](https://linear.app/valora/issue/ACT-782/remove-the-v-icon-on-contacts-recipients), [Notion discussion](https://www.notion.so/valora-inc/Recipient-screen-Valora-V-behavior-658afd84498345d38027fc43df67397d)

### Test plan

Before change:
v_icon_before

After change:
v_icon_after


src/recipients/RecipientItem.tsx
```diff
@@ -33,13 +32,6 @@ function RecipientItem({ recipient, onSelectRecipient, loading }: Props) {
           </Text>
           {recipient.name && <Text style={styles.phone}>{getDisplayDetail(recipient)}</Text>}
         </View>
-        <View style={styles.rightIconContainer}>
-          {loading ? (
```
satish-ravi, line 37: I think we still need the loading icon here. We want to show it when looking up whether a phone number is connected to an address. Can you confirm this by typing a phone number on the search box and tapping it?

cajubelt: @satish-ravi are you suggesting just using the loading prop to show a spinner that turns into a blank space once the lookup is complete? why do we need that?

cajubelt: oh ic, a bottom sheet comes up once the lookup completes successfully

cajubelt: and an error banner appears if it fails

satish-ravi: yeah, I think it also navigates to an invite screen if its a phone number not connected to Valora

jeanregisser:
Thanks! 👍 

Even though it's a small change, could you add some info to help folks not familiar with this (me? 😄) why we're making this change? Could be a linear task, slack convo, or just some text in the PR. 🙏

cajubelt (approved):

PR #4237, opened 2023-09-27
fix(tokens): Make priceUsd optional

jophish:
### Description

This field can be undefined, so updating the schema to match.

### Related issues

- Fixes #[issue number here]


satish-ravi (approved):

jeanregisser (approved):
I was wondering if we should return 0 / NaN if we don't have a price, to maybe simplify some of the consuming code.
But it seems undefined is ok.

PR #4241, opened 2023-09-28
fix: upgrade react-navigation-bottom-sheet and remove deprecated hook

kathaypacific:
### Description

This PR upgrades the `@th3rdwave/react-navigation-bottom-sheet` lib to the latest, and removes usage of the deprecated `useBottomSheetDynamicSnapPoints ` hook. The implementation can now be replaced by the [`enableDynamicSizing ` prop](https://gorhom.github.io/react-native-bottom-sheet/props/#enabledynamicsizing).

This simplifies our implementation for the bottom sheet screens and components, and we can remove the previously necessary patch.

### Test plan

There should be no change for the user.

Tested the WalletConnect request screens, and the revoke CPV bottom sheet component on Android and iOS.

### Related issues

- Fixes RET-721
- Part of RET-871

### Backwards compatibility

Y


src/navigator/Navigator.tsx
```diff
@@ -750,13 +732,12 @@ function RootStackScreen() {
     <RootStack.Navigator
       screenOptions={{
         backdropComponent: renderBackdrop,
-        handleHeight: animatedHandleHeight,
-        snapPoints: animatedSnapPoints,
-        contentHeight: animatedContentHeight,
+        enableDynamicSizing: true,
+        snapPoints: ['CONTENT_HEIGHT'],
```
kathaypacific, line 736: unfortunately, there is a default snap point of 66%. this means that it is always possible for the bottom sheet to lock into this position. `enableDynamicSizing` adds to the existing snap points instead of overriding them. if we do not set a snap point (or set it to another value), it means that bottom sheet screens will resize to the content, but be able to lock to this arbitrary height too. it looked weird in the wallet connect requests, where you can drag the sheet to lock at 66% : 

https://github.com/valora-inc/wallet/assets/20150449/c7aa76cc-1cf5-4d38-bc36-e5a90eb6686b

to override this behaviour i'm using CONTENT_HEIGHT instead but unfortunately under the hood, this library still uses the deprecated `useBottomSheetDynamicSnapPoints` hook so we still see dev warnings. i think this is the best approach for now, and wait for there to be an update for the lib.

bakoushin: Would you mind adding a short comment near this prop for our future selves?

src/components/BottomSheet.tsx
```diff
@@ -36,14 +29,12 @@ const BottomSheet = ({ forwardedRef, title, description, children, testId }: Pro
     <GorhomBottomSheet
       ref={forwardedRef}
       index={-1}
-      snapPoints={animatedSnapPoints}
-      handleHeight={animatedHandleHeight}
-      contentHeight={animatedContentHeight}
+      enableDynamicSizing
```
bakoushin, line 32: Do we want do add `snapPoints: ['CONTENT_HEIGHT'],` here in consistency with [Navigator.tsx](https://github.com/valora-inc/wallet/pull/4241/files#diff-37c5052bf8f3b1e84a2a8b2c37c9da7689b68089261c702b8031dc7f1db77b83R736)?

bakoushin (approved):
LGTM! 🙌 

I left few small non-blocking comments/questions.

src/navigator/Navigator.tsx
kathaypacific: yes, added!

src/components/BottomSheet.tsx
kathaypacific: no, the default snap point is implemented in the `@th3rdwave/react-navigation-bottom-sheet` library only. under the hood, `@th3rdwave/react-navigation-bottom-sheet` uses `@gorhom/bottom-sheet` but there are no default snap points required in `@gorhom/bottom-sheet`. the BottomSheet component uses the `@gorhom/bottom-sheet` lib directly.

PR #4242, opened 2023-09-28
feat: allow Eth selection in the send flow

MuckT:
### Description

Allows the selection of Eth in the send flow and the use of the Max button while preserving functionality of the existing Celo send flow.

#### Video

https://github.com/valora-inc/wallet/assets/26950305/2ce014b9-8d89-49a8-bbd7-f8471a81a3c9

### Test plan

- Tested locally on iOS by adding a wallet address to `Prod Multichain Testers` segment in Statsig.
- Unit tests updated.

### Related issues

- Fixes ACT-903

### Backwards compatibility

Yes


src/analytics/Properties.tsx
```diff
@@ -638,17 +638,18 @@ interface SendEventsProperties {
     error: string
   }
   [SendEvents.token_dropdown_opened]: {
-    currentTokenAddress: string
+    currentTokenId: string
```
cajubelt, line 641: @MuckT what do you think about adding the token id column and changing the token address columns to string | null, rather than dropping token address columns? That way if there are dashboards that group by token address, they will continue to have some utility in helping us analyze everything except ETH sends, rather than immediately breaking

cajubelt: thinking more about analytics, I think it's very likely that we'll want to group by networkId too

cajubelt: so it would be good to add that as well

src/fees/hooks.ts
```diff
@@ -43,9 +43,60 @@ export function usePaidFees(fees: Fee[]) {
...
+  }
+  return balance.minus(feeEstimate)
+}
+
+// Returns the maximum amount a user can send, taking into account gas fees required for the transaction
+// also optionally fetches new fee estimations if the current ones are missing or out of date
+/**
+ * @deprecated use useMaxSendAmount instead
+ */
+export function useMaxSendAmountByAddress(
```
cajubelt, line 99: It looks like this largely duplicates `useMaxSendAmount`. That's not ideal for the following reasons:
- makes our code harder to understand, because there is more code to read through in total to understand how the two hooks work
- makes our code less safe from bugs, because if a bug crops up, it now needs to be solved in 2 different places

In general, I suggest challenging yourself to never copy/paste code from another part of the codebase. When you do that, chances are you're missing an opportunity to create some helper function that would be useful in two places. And since you're not writing each line, you might miss some detail that will lead to bugs.

In this case, you could refactor useMaxSendAmountByAddress to just be a wrapper for your new implementation, shrinking the implementation down to just 2 lines
```
  const tokenInfo = useTokenInfoByAddress(tokenAddress)
  return useMaxSendAmount(tokenInfo?.tokenId, feeType, shouldRefresh)
```

However, this would require `useMaxSendAmount` to have an optional `tokenId`, which in turn would require `useTokenInfo` to make `tokenId` optional. I think this is going to be useful in the end anyway, since porting all the usages of `useTokenInfoByAddress` (24 right now) over to `useTokenInfo` without making the input optional sounds like unnecessary scope. I'll let @jophish comment if he had some intention to make inputs of hooks required more often, but it seems cumbersome to me to require that, given that you often want to use hooks by other hooks, which being governed by rules of hooks (no conditional evaluation in particular), make it hard to guarantee that inputs will be truthy in general

cajubelt: alternatively, you could just update the 2 usages of `useMaxSendAmountByAddress` to use your new version, since both usages appear to already have easy access to a tokenId. However, I think you will run into the same issue of `tokenId` being required in both cases. More evidence that `useTokenInfo` needs `tokenId` to be optional IMO

cajubelt (changes requested):
reviewed thru src/fees/hooks and needed to go offline

src/fees/hooks.ts
cajubelt: I created this visual on the general strategic choice. When there's this much duplication between the celo-specific implementation and the multichain one, I don't see option 2 as viable. https://www.figma.com/file/iiPCJG2cA18ysZrLAZI80d/Multichain-refactor-strategies?type=whiteboard&node-id=0%3A1&t=157ni2SZLUNGpni9-1

src/navigator/Headers.tsx
```diff
@@ -216,7 +217,7 @@ export function HeaderTitleWithBalance({
   )
 }
 
-export function HeaderTitleWithTokenBalance({
+export function HeaderTitleWithTokenBalanceByAddress({
```
cajubelt, line 220: why is this necessary? it has the same input types as the new version. When I checked out your branch and swapped the only usage of this out for HeaderTitleWithTokenBalance, no type errors resulted. Am I missing something?

src/send/SendAmount/index.tsx
```diff
@@ -61,14 +64,42 @@ type Props = RouteProps
...
+    localAmount,
+    tokenAmount,
+    usdAmount: usdAmount && convertToMaxSupportedPrecision(usdAmount),
+  }
+}
+
+/**
+ * @deprecated Use useInputAmounts instead
+ */
+export function useInputAmountsByAddress(
```
cajubelt, line 94: please make these more DRY

src/tokens/hooks.ts
```diff
@@ -93,7 +118,20 @@ export function useLocalToTokenAmount(
...
+    tokenAmount,
+    tokenInfo,
+    usdToLocalRate,
+  })
+}
+
+/**
+ * @deprecated use useLocalToTokenAmount
+ */
+export function useTokenToLocalAmountByAddress(
```
cajubelt, line 134: please make these more DRY

src/tokens/hooks.ts
```diff
@@ -80,7 +92,20 @@ export function useTokenInfoByCurrency(currency: Currency) {
...
+    localAmount,
+    tokenInfo,
+    usdToLocalRate,
+  })
+}
+
+/**
+ * @deprecated use useLocalToTokenAmount
+ */
+export function useLocalToTokenAmountByAddress(
```
cajubelt, line 108: DRY

src/tokens/hooks.ts
```diff
@@ -106,7 +144,18 @@ export function useTokenToLocalAmount(
...
+  if (!tokenInfo?.priceUsd) {
+    return null
+  }
+  return amount.multipliedBy(tokenInfo.priceUsd)
+}
+
+/**
+ * @deprecated use useAmountAsUsd
+ */
+export function useAmountAsUsdByAddress(amount: BigNumber, tokenAddress: string) {
```
cajubelt, line 158: DRY

src/tokens/hooks.ts
```diff
@@ -67,6 +70,15 @@ export function useTokenInfo(tokenId: string) {
   return tokens[tokenId]
 }
 
+/**
+ * @deprecated only used in the old CICO send bar and should be removed when that is updated to multi-chain support
+ */
+export function useTokenInfoBySymbol(symbol: string) {
```
cajubelt, line 76: I'm not seeing any usages of this anymore, can we just get rid of it? Also, as written it looks like it includes ethereum tokens with no feature flag gating that...

src/send/Send.tsx
```diff
@@ -60,7 +65,11 @@ function Send({ route }: Props) {
   const allRecipients = useSelector(phoneRecipientCacheSelector)
   const recentRecipients = useSelector((state) => state.send.recentRecipients)
 
-  const tokensWithBalance = useSelector(tokensWithTokenBalanceAndAddressSelector)
+  const supportedNetworkIds = getSupportedNetworkIdsForSend()
+  const tokensWithBalance = useTokensWithTokenBalance()
```
cajubelt, line 69: looks like useTokensWithTokenBalance filters by the network id's that are supported for token balances first. then you apply a filter here by network id's that are supported for send. I don't think that's how our feature flags should work-- seems like we should be able to turn on/off different multichain features independently of each other, and not have a hardcoded dependency where the send feature flag depends on the balances feature flag

cajubelt: what about a hook like `useTokensForSend` that just uses the send feature flag, instead of a combination of both?

src/analytics/Properties.tsx
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/4242/commits/f9d9276183d26c2e7d9795422fbca647bf580498!

src/navigator/Headers.tsx
MuckT: Removed in https://github.com/valora-inc/wallet/pull/4242/commits/330bbe245ffe9a8a89a2185d0cbb2ac89d9e50ae!

src/tokens/hooks.ts
MuckT: Removed in https://github.com/valora-inc/wallet/pull/4242/commits/b5235c47a65fcd2b31f2ba79ad547854a651039c!

src/fees/hooks.ts
MuckT: Updating the two usages of `useMaxSendAmountByAddress` looks like it would require updating `FiatExchangeAmount.tsx` and `SwapScreen.tsx` which means using tokenId in the swap flow which impacts RET-872. It seems easier to simply create a new `useMaxSendAmount` function which takes tokenId can be used by swaps when ready and migrate the flows to that new function as we add multi-chain support @jh2oman.

src/send/Send.tsx
MuckT: I think that makes sense, in the case where supportedNetworks for send was `celo-mainnet` & `ethereum-mainnet`, but `tokensWithBalance` only included tokens from `celo-mainnet` it would be fine because only tokens on `celo-mainnet` would have balances available for send. For requests Celo stables are only available so there should be no impact.

MuckT: Fixed in https://github.com/valora-inc/wallet/pull/4242/commits/775caebe434df7f28a1c32807065ccfe5e835413!

src/fees/hooks.ts
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/4242/commits/d6786bd16d6cf578a3a1de8c45e1d57432028718!

src/send/SendAmount/useTransactionCallbacks.ts
```diff
@@ -64,9 +64,9 @@ function useTransactionCallbacks({
       inputAmount: inputIsInLocalCurrency ? localAmount! : tokenAmount,
       tokenAmount,
       amountIsInLocalCurrency: inputIsInLocalCurrency,
-      tokenAddress: transferTokenAddress,
+      tokenAddress: tokenInfo!.address!,
```
cajubelt, line 67: please update the types for the analytics events instead of using non-null assertions like this. the first one could even cause a runtime error, which would be unnecessary, particularly considering it's just assembling data for analytics

src/send/SendAmount/useTransactionCallbacks.ts
```diff
@@ -78,7 +78,7 @@ function useTransactionCallbacks({
       localCurrencyExchangeRate,
       localCurrency: localCurrencyCode,
       localCurrencyAmount: localAmount?.toString() ?? null,
-      underlyingTokenAddress: transferTokenAddress,
+      underlyingTokenAddress: tokenInfo!.address!,
```
cajubelt, line 81: same here...

src/send/SendAmount/useTransactionCallbacks.ts
```diff
@@ -64,9 +64,9 @@ function useTransactionCallbacks({
       inputAmount: inputIsInLocalCurrency ? localAmount! : tokenAmount,
       tokenAmount,
       amountIsInLocalCurrency: inputIsInLocalCurrency,
-      tokenAddress: transferTokenAddress,
+      tokenAddress: tokenInfo!.address!,
     }),
-    [recipient, tokenAmount, transferTokenAddress]
+    [recipient, tokenAmount, transferTokenId]
```
cajubelt, line 69:
```suggestion
    [recipient, tokenAmount, tokenInfo]
```

cajubelt: I think using transferTokenId in places like this is relying on the assumption that tokenInfo will always be static relative to tokenId. If we added something to tokenInfo in the future like price or balance, that would no longer be the case, and we could get bugs (if we also used that dynamic property in the callback).

src/send/SendAmount/useTransactionCallbacks.ts
```diff
@@ -103,7 +103,7 @@ function useTransactionCallbacks({
 
   const feeType = FeeType.SEND
   const estimateFeeDollars =
-    useSelector(getFeeEstimateDollars(feeType, transferTokenAddress)) ?? new BigNumber(0)
+    useSelector(getFeeEstimateDollars(feeType, tokenInfo!.address!)) ?? new BigNumber(0)
```
cajubelt, line 106: another non-null assertion operator. please remove it and deal with the case where tokenInfo is null instead.

also, it's weird to me that this useTransactionCallbacks hook has been refactored to accept tokenId as a parameter, but then uses token addresses under the hood, breaks with a runtime error if an address does not exist, and seems to assume every token's address is a Celo address (here, when getting estimated fees). 

Unfortunately, it looks like refactoring `getFeeEstimateDollars` to use token ID's is going to take some effort. But leaving the hook in this half-refactored state seems very risky

src/tokens/hooks.ts
```diff
@@ -107,11 +114,22 @@ export function useLocalToTokenAmount(
...
 ): BigNumber | null {
   const tokenInfo = useTokenInfoByAddress(tokenAddress)
+  return useLocalToTokenAmount(localAmount, tokenInfo?.tokenId)
+}
+
+export function useTokenToLocalAmount(
+  tokenAmount: BigNumber,
+  tokenId: string | undefined
+): BigNumber | null {
+  const tokenInfo = useTokenInfo(tokenId!)
```
cajubelt, line 132:
```suggestion
  const tokenInfo = useTokenInfo(tokenId)
```

src/tokens/hooks.ts
```diff
@@ -120,14 +138,33 @@ export function useTokenToLocalAmount(
...
+export function useTokenToLocalAmountByAddress(
+  tokenAmount: BigNumber,
+  tokenAddress?: string
+): BigNumber | null {
   const tokenInfo = useTokenInfoByAddress(tokenAddress)
+  return useTokenToLocalAmount(tokenAmount, tokenInfo?.tokenId)
+}
+
+export function useAmountAsUsd(amount: BigNumber, tokenId: string | undefined) {
+  const tokenInfo = useTokenInfo(tokenId!)
```
cajubelt, line 153:
```suggestion
  const tokenInfo = useTokenInfo(tokenId)
```

src/send/SendAmount/useTransactionCallbacks.ts
cajubelt: your Linear issue says actually issuing the send and doing fee estimation for ethereum is out of scope. maybe you could avoid changing this hook at all.. you'd need to throw an error from SendAmount if "review" is pressed and the network id is not celo or address was undefined, and leave a todo to pick it up in a later issue in the epic, but that's probably better than silently failing like this imo

src/components/TokenBottomSheet.tsx
```diff
@@ -23,47 +25,60 @@ export enum TokenPickerOrigin {
...
-          />
-          <LegacyTokenDisplay
-            style={styles.currencyBalance}
-            amount={tokenInfo.balance}
-            tokenAddress={tokenInfo.address}
-            showLocalAmount={false}
-            testID={`${tokenInfo.symbol}Balance`}
-          />
-        </View>
+        {tokenInfo.tokenId ? (
```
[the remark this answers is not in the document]
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/4242/commits/aa0877eb68f51b1a8d60707f1b5b0e6c1efce893!

src/home/SendBar.tsx
```diff
@@ -25,8 +25,8 @@ export default function SendBar({ selectedCurrency, skipImport }: Props) {
   const onPressSend = () => {
     navigate(Screens.Send, {
       skipContactsImport: skipImport,
-      defaultTokenOverride: tokenInfo?.address,
-      forceTokenAddress: !!tokenInfo?.address,
+      defaultTokenOverride: tokenInfo?.tokenId,
```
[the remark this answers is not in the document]
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/4242/commits/c5cb1c96b0de8c61737b1aa25fd2e9e5966a9e7f!

src/fees/hooks.ts
```diff
@@ -43,30 +43,29 @@ export function usePaidFees(fees: Fee[]) {
...
       (feeType === FeeType.SWAP && balance.gt(0)) ||
       !feeEstimate ||
       feeEstimate.error ||
       feeEstimate.lastUpdated < Date.now() - ONE_HOUR_IN_MILLIS
     ) {
-      dispatch(estimateFee({ feeType, tokenAddress }))
+      dispatch(estimateFee({ feeType, tokenAddress: tokenInfo.address }))
     }
-  }, [tokenAddress, shouldRefresh])
+  }, [tokenId, shouldRefresh])
```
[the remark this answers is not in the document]
MuckT: I had to change this to tokenInfo?.tokenId since tokenInfo is an object was created on render using it in the dependency array causes fee's to be re-estimated ever render.

src/send/SendAmount/useTransactionCallbacks.ts
MuckT: `Send Review` refers  to `src/send/SendConfirmation.tsx` which doesn't correctly render at present; `src/send/SendAmount/index.tsx` is the previous screen in the flow. It seems like updating this screen work will take keeping the updates to useTransactionCallbacks.ts and also updating `TransactionDataInput` type. Would it be better to revert the changes to the `useTransactionCallbacks.ts` and punt on the display for `src/send/SendConfirmation.tsx` or should I keep the update to `useTransactionCallbacks.ts` and also update `TransactionDataInput`? Granted `TransactionDataInput` is used a lot of places.

src/send/SendAmount/useTransactionCallbacks.ts
MuckT: Fixed in https://github.com/valora-inc/wallet/commit/83caddecdff9f0e497f29222f60eb4fb2a1c5fd8!

src/send/SendAmount/useTransactionCallbacks.ts
MuckT: Similar issue to https://github.com/valora-inc/wallet/pull/4242#discussion_r1351131006.

src/send/SendAmount/useTransactionCallbacks.ts
MuckT: Ended up making getFeeEstimateDollars handle a possible null address in https://github.com/valora-inc/wallet/pull/4242/commits/8d29f198c011df360d22438e7346ea2a0fe72936!

src/send/SendAmount/useTransactionCallbacks.ts
cajubelt: bumping this

src/fees/selectors.ts
```diff
@@ -9,12 +10,16 @@ export function getFeeInTokens(feeInWei: BigNumber.Value | null | undefined) {
...
 export const feeEstimatesSelector = (state: RootState) => state.fees.estimates
 
-export function getFeeEstimateDollars(feeType: FeeType | null, tokenAddress: string) {
+export function getFeeEstimateDollars(
+  feeType: FeeType | null,
+  tokenInfo: TokenBalance | undefined
+) {
   return (state: RootState) => {
-    if (!feeType) {
+    // TODO: Handle cases where address is null ex: Ethereum
```
cajubelt, line 18:
```suggestion
    // TODO(ACT-922): Handle cases where address is null ex: Ethereum
```
is [ACT-922](https://linear.app/valora/issue/ACT-922/ethereum-fee-estimation) when you envisioned this happening?

when you leave a TODO can you try to leave some indication of when or by whom it should get done?

cajubelt (commented):
looks close!

cajubelt (approved):
🥳 🥳 🥳

PR #4253, opened 2023-10-03
feat(token-icon): parameterize size and add fallback icon

satish-ravi:
### Description

The Token details screen requires a smaller version of the token icon with the network icon, so parameterized the TokenIcon component to take a size parameter. Also, added a default icon as fallback since `imageUrl` for tokens can potentially be undefined, there are at least 10 tokens on mainnet at this time.

### Test plan

Unit tests & manual

Small image:

small-token

Default token image:

default-token

### Related issues

Part of ACT-915

### Backwards compatibility

Yes


MuckT (approved):
LGTM. Tested on iOS and Android.

PR #4254, opened 2023-10-04
feat(swap): make from token an optional prop to swap screen

satish-ravi:
### Description

When we navigate to the swap screen from the [token details page](https://www.figma.com/file/OE6RFzqXAMqU9ygHtNYyof/Assets-Page-(Multichain)?node-id=2099%3A15029&mode=dev), we want to default the swap from token to the token being navigated from.

### Test plan

Unit tests, manually by setting this option via the home action swap navigation

### Related issues

- Part of ACT-915

### Backwards compatibility

Yes


src/swap/SwapScreen.test.tsx
```diff
@@ -775,24 +801,6 @@ describe('SwapScreen', () => {
     expect(queryByTestId('TTTouchable')).toBeFalsy()
   })
 
-  it('should be able to hide top drawer nav when parameter is set', () => {
```
satish-ravi, line 778: showDrawerTopBar is no longer an option. No need to test this.

kathaypacific (approved):
LGTM!

PR #4260, opened 2023-10-04
fix: dismiss keyboard on open and close bottom sheet

kathaypacific:
### Description

This PR fixes / tweaks a few things that were all small and related:
1. any open keyboards should be dismissed when the bottom sheet is launched and dismissed. i can't think of when this logic should _**not**_ be applied, so i added it to the bottom sheet so that it doesn't have to be duplicated in all consumers.
2. scrolling lists inside the bottom sheet should not preserve the last scrolled position on relaunch. since the bottom sheet is not unmounted on dismiss, manually reset the scroll when it is closed.
3. allow the bottom sheet scroll view to handle taps when the keyboard is enabled (i.e. token selection can happen without first dismissing the keyboard)
4. there was some logic to dismiss the keyboard in the BottomSheetScrollView introduced in https://github.com/valora-inc/wallet/pull/3924, i removed this to put it inside the BottomSheet instead. reading the PR description though, it seems like the logic needs to happen when opening dapps so I added it to the `useOpenDapp` hook instead.
5. in the swap screen, make the token bottom sheet taller to give more screen real estate to the list
6. in the swap screen, the missing price impact warning shouldn't be displayed unless both tokens are selected
7. turn off autocorrect and spellcheck on the token bottom sheet search keyboard. this hides the suggestions bar on top of the iOS keyboard, but also users are likely to be searching for token names that autocorrect would try to "fix". note that i didn't try to disable suggestions on the android keyboard because there seems to be a shortcuts bar on the keyboard either way, see recording below.

### Test plan

(Example on an Android device that has a top notch, the notch isn't shown well in the recording)

https://github.com/valora-inc/wallet/assets/20150449/521d5e57-d1c2-49cf-9fe4-cbd458b60a9d

### Related issues

- RET-871

### Backwards compatibility

Y


src/swap/SwapScreen.tsx
```diff
@@ -225,14 +225,11 @@ export function SwapScreen({ route }: Props) {
 
   const handleShowTokenSelect = (fieldType: Field) => () => {
     ValoraAnalytics.track(SwapEvents.swap_screen_select_token, { fieldType })
-    // ensure that the keyboard is dismissed before animating token bottom sheet
-    Keyboard.dismiss()
     setSelectingToken(fieldType)
-    tokenBottomSheetRef.current?.snapToIndex(0)
-  }
 
-  const handleCloseTokenSelect = () => {
```
kathaypacific, line 234: this was redundant, since the selecting token is already cleared when the token is selected in `handleSelectToken`

src/components/BottomSheet.tsx
```diff
@@ -47,6 +48,18 @@ const BottomSheet = ({
...
+  const handleAnimate = (fromIndex: number, toIndex: number) => {
+    if (toIndex === -1 || fromIndex === -1) {
+      Keyboard.dismiss()
+    }
+  }
+
+  const handleClose = () => {
+    onClose?.()
+    scrollViewRef.current?.scrollTo({ y: 0, animated: false })
+  }
```
jeanregisser, line 61: Could we add comments as to why we're doing these? what's in the PR description.
It helps a lot while reading the code.

src/components/SearchInput.tsx
```diff
@@ -14,7 +14,13 @@ export default function SearchInput({ style, ...passThroughProps }: Props) {
       <View style={styles.iconContainer}>
         <Search />
       </View>
-      <TextInput {...passThroughProps} inputStyle={styles.input} testID="SearchInput" />
+      <TextInput
+        {...passThroughProps}
+        inputStyle={styles.input}
+        autoCorrect={false}
+        spellCheck={false}
```
jeanregisser, line 21: Also here a comment would help as to why we don't want these.

src/swap/SwapScreen.tsx
```diff
@@ -225,14 +225,11 @@ export function SwapScreen({ route }: Props) {
...
-    setSelectingToken(null)
+    requestAnimationFrame(() => {
+      tokenBottomSheetRef.current?.snapToIndex(0)
+    })
```
jeanregisser, line 232: We could add a comment as to why we need `requestAnimationFrame`.

src/swap/SwapScreen.tsx
```diff
@@ -257,7 +254,9 @@ export function SwapScreen({ route }: Props) {
     }
 
     setSelectingToken(null)
-    tokenBottomSheetRef.current?.close()
+    requestAnimationFrame(() => {
+      tokenBottomSheetRef.current?.close()
+    })
```
jeanregisser, line 259: ditto

jeanregisser (approved):
Awesome improvements! 
Thank you! 🚀 🤩

kathaypacific:
> Awesome improvements! Thank you! 🚀 🤩

TY @jeanregisser i've added comments on the places you suggested!

PR #4265, opened 2023-10-05
feat(multichain): Use new dynamic config for multichain features

jophish:
### Description

For ACT-931. Replaces three of our existing feature gates with a single Statsig Dynamic Config (available [here](https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/dynamic_configs/multi_chain_features)).

The `show_native_tokens` feature gate is a straggler here in terms of multichain config. This was added as part of a quick and dirty hack to get non-Celo transfers showing up in the home feed a while back, and ought to be removed. I'll remove this in a followup PR.

### Test plan

Manual and unit tested.

### Related issues

- Fixes ACT-931

### Backwards compatibility

Yes.


src/transactions/feed/TransactionFeed.test.tsx
```diff
@@ -18,6 +18,11 @@ import { mockCusdAddress } from 'test/values'
 
 jest.mock('src/statsig', () => ({
   getFeatureGate: jest.fn(() => false),
+  getDynamicConfigParams: jest.fn(() => ({
```
satish-ravi, line 21: Noticed these in a bunch of test files, are the mocks on these test files required because there's a nested dependency to this dynamic config? And is the getFeatureGate mock now required? I wonder if we should consider creating a shared mock under `__mocks__` similar to how navigate / analytics is mocked so we don't need this everywhere

src/statsig/constants.ts
```diff
@@ -91,4 +89,12 @@ export const DynamicConfigs = {
       cico: 30,
     },
   },
+  [StatsigDynamicConfigs.MULTI_CHAIN_FEATURES]: {
+    configName: StatsigDynamicConfigs.MULTI_CHAIN_FEATURES,
+    defaultValues: {
+      enableCico: [networkConfig.defaultNetworkId],
+      showBalances: [networkConfig.defaultNetworkId],
+      showTransfers: [networkConfig.defaultNetworkId],
```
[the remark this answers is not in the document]
jophish: Sure, I can change this to `showCico`.

src/tokens/TokenBalances.test.tsx
```diff
@@ -211,6 +214,9 @@ describe('TokenBalancesScreen', () => {
   })
 
   it('renders the correct information in positions', () => {
+    jest.mocked(getDynamicConfigParams).mockReturnValue({
+      showBalances: [NetworkId['celo-alfajores'], NetworkId['ethereum-sepolia']],
+    })
```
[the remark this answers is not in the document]
jophish: Whoops, thanks for this catch -- they do not; I've updated this to match the diff for Assets.test.tsx.

src/transactions/feed/TransactionFeed.test.tsx
jophish: That's correct, re: nested dependency. And yes, the `getFeatureGate` mock is still required here. And interesting -- I haven't messed around with shared mocks before. How would that work? Would there be a mock for the entirety of `src/statsig` that would automatically be included in all tests? What if we want to override this mock (which will certainly be the case)?

src/fiatExchanges/FiatExchangeCurrency.tsx
```diff
@@ -104,7 +107,10 @@ function FiatExchangeCurrency({ route, navigation }: Props) {
   const { flow } = route.params
 
   const [selectedCurrency, setSelectedCurrency] = useState<CiCoCurrency>(CiCoCurrency.cUSD)
-  const showEth = getFeatureGate(StatsigFeatureGates.SHOW_ETH_IN_CICO)
+  // TODO: Update this to actually respect all possible networkIds correctly
```
satish-ravi, line 110: is this in a follow up PR or is there a ticket for this?

src/transactions/feed/TransactionFeed.test.tsx
satish-ravi: yes it would mock `src/statsig` for all tests. We can still setup test specific mocks as we do normally and it will override this mock. In any case its outside the scope of this PR, but something to consider if we have more of these shared mocks.

satish-ravi (approved):

src/fiatExchanges/FiatExchangeCurrency.tsx
jophish: https://linear.app/valora/issue/ACT-938/update-cico-flow-to-correctly-respect-enabled-tokens-from-statsig

PR #4266, opened 2023-10-05
feat(assets): add token details screen

satish-ravi:
### Description

[Figma](https://www.figma.com/file/OE6RFzqXAMqU9ygHtNYyof/Assets-Page-(Multichain)?node-id=2099%3A15029&mode=dev)

### Test plan

Unit tests, manual by enabling dummy feature flag

Showing all actions with swap enabled:

https://github.com/valora-inc/wallet/assets/5062591/5041860d-d332-49e2-97cc-8abcaf214e7c

Showing actions with swap disabled (withdraw is hidden behind more):

https://github.com/valora-inc/wallet/assets/5062591/0d84a6ad-bde8-4a41-8b25-4aa67dac950e

### Related issues

- Fixes ACT-915

### Backwards compatibility

Yes


cajubelt:
just from the videos-
- the bottom sheet when "more" is clicked isn't in the demo video
- the "withdraw" icon looks kinda close to the edge. it's not in the designs, so this is pretty subjective

very cool though!

src/tokens/hooks.ts
```diff
@@ -61,6 +67,51 @@ export function useTokenPricesAreStale(networkIds: NetworkId[]) {
   }
 }
 
+export function useSendableTokens() {
+  // TODO: get this from statsig
```
cajubelt, line 71: is there an issue where this will be done? or is it something you plan to do in a followup PR?

src/tokens/hooks.ts
```diff
@@ -61,6 +67,51 @@ export function useTokenPricesAreStale(networkIds: NetworkId[]) {
...
+  const networkIdsForSend = [networkConfig.defaultNetworkId]
+  const tokens = useSelector(tokensListSelectorWrapper(networkIdsForSend))
+  return tokens.filter((tokenInfo) => tokenInfo.balance.gt(TOKEN_MIN_AMOUNT))
+}
+
+export function useSwappableTokens() {
+  const appVersion = DeviceInfo.getVersion()
+  // TODO: get this from statsig
```
cajubelt, line 79: same here

src/tokens/hooks.ts
```diff
@@ -61,6 +67,51 @@ export function useTokenPricesAreStale(networkIds: NetworkId[]) {
...
+    .filter(
+      (tokenInfo) =>
+        tokenInfo.isSwappable ||
+        (tokenInfo.minimumAppVersionToSwap &&
+          !isVersionBelowMinimum(appVersion, tokenInfo.minimumAppVersionToSwap))
+    )
+    .sort(tokenCompareByUsdBalanceThenByName)
+}
+
+function isCiCoToken(tokenSymbol: string): tokenSymbol is CiCoCurrency {
```
cajubelt, line 92: what do you think about abandoning the "CiCo" casing pattern? seems like proper camelCase would just be CICO

src/tokens/hooks.ts
```diff
@@ -61,6 +67,51 @@ export function useTokenPricesAreStale(networkIds: NetworkId[]) {
...
+  return Object.values(CiCoCurrency).some((cicoSymbol) => cicoSymbol === tokenSymbol)
+}
+
+export function useCashInTokens() {
```
cajubelt, line 96: see in cases like this one we capitalize the i, just not when we abbreviate. doesn't make sense to me

MuckT:
@satish-ravi I got an app crash when navigating around the tabs

src/tokens/TokenDetails.tsx
```diff
@@ -0,0 +1,325 @@
...
+  const cashOutTokens = useCashOutTokens()
+  const isSwapEnabled = useSelector(isAppSwapsEnabledSelector)
+
+  const actions = [
+    {
+      name: TokenDetailsActionName.Send,
+      text: t('tokenDetails.actions.send'),
+      iconComponent: QuickActionsSend,
+      onPress: () => {
+        navigate(Screens.Send, { defaultTokenOverride: token.address! })
```
cajubelt, line 143: I recently saw a refactor [here](https://github.com/valora-inc/wallet/pull/4242) by @MuckT that changes the meaning of defaultTokenOverride from address to tokenId, but the type is still string. Some faint alarm bell went off when I saw that, and I'm realizing now that I was worried that one of you could merge the changes without seeing this change and we'd end up with a bug. @MuckT what do you think should be done to prevent this? Maybe the `defaultTokenOverride` parameter name could be changed (in Tom's PR) so we get a compilation error if someone makes this mistake

src/tokens/TokenDetails.tsx
```diff
@@ -0,0 +1,325 @@
...
+        !!swappableTokens.find((tokenInfo) => tokenInfo.tokenId === token.tokenId) &&
+        token.balance.gt(TOKEN_MIN_AMOUNT),
+    },
+    {
+      name: TokenDetailsActionName.Add,
+      text: t('tokenDetails.actions.add'),
+      iconComponent: QuickActionsAdd,
+      onPress: () => {
+        navigate(Screens.FiatExchangeAmount, {
+          currency: token.symbol as CiCoCurrency,
```
cajubelt, line 165: this type cast seems to rely on the internal logic of `useCashInTokens`, which could change and cause a bug here...

cajubelt: same with the following one

src/tokens/TokenDetails.tsx
```diff
@@ -0,0 +1,325 @@
...
+  const moreAction = {
+    name: TokenDetailsActionName.More,
+    text: t('tokenDetails.actions.more'),
+    iconComponent: QuickActionsMore,
+    onPress: () => {
+      // TODO(ACT-917): open bottom sheet
+    },
+  }
+
+  const actionButtons = actions.length > 3 ? [actions[0], actions[1], moreAction] : actions
```
cajubelt, line 196: nit: magic numbers. probly better to put 3 into `MAX_BUTTONS_PER_ROW`

src/tokens/TokenDetails.tsx
```diff
@@ -0,0 +1,325 @@
...
+
+  return (
+    <View
+      style={[
+        styles.actions,
+        { gap: actionButtons.length === 3 ? Spacing.Smallest8 : Spacing.Regular16 },
```
cajubelt, line 202: magic numbers

cajubelt (approved):
main concerns are a type cast and a potential conflict with Tom's PR-- can you work with him to resolve it?

nothing major so I'm approving also

satish-ravi:
@cajubelt
- the more bottom sheet is out of scope for this and will be covered in a separate ticket. 
- I've been chatting w/ Nitya on the Withdraw button, I think she's ok with this since for most users swap would be enabled and withdraw will only pop up on the bottom sheet. This would also be a problem for all buttons in other languages if there are longer names, but I think we can hold off on optimizing it for all languages now.

src/tokens/TokenDetails.tsx
MuckT: Happy to change it from `defaultTokenOverride` to `defaultTokenIdOverride` in #4242.

src/tokens/TokenDetails.tsx
satish-ravi: fixed

src/tokens/TokenDetails.tsx
satish-ravi: fixed

src/tokens/TokenDetails.tsx
satish-ravi: added an explicit condition on the navigate function

src/tokens/hooks.ts
satish-ravi: todo addressed after #4265 was merged

src/tokens/hooks.ts
satish-ravi: I ended up changing to Cico to be consistent with the feature gate `showCico`

src/tokens/TokenDetails.tsx
satish-ravi: Left a TODO, so this can be addressed depending on which one merges first

cajubelt: I think Tom's suggestion of changing the variable name is safer, since it doesn't depend on merge order-- in either case, a build error should take place if one is merged without the other taking the changes into account yet

cajubelt (approved):

PR #4267, opened 2023-10-05
feat: add nfts to new assets redesign

MuckT:
### Description

Adds Nft Collections to the new assets page. Nft items are displayed as groups in a section list instead of a flat list as was previously done in https://github.com/valora-inc/wallet/pull/3912. Now accepts an arbitrary number for `NUM_OF_NFTS_PER_ROW` constant.

| 2 Nfts Per Row | 3 Nfts Per Row | Error | No Nfts |
|  |  |  |  |

### Test plan

- [x] Tested locally on iOS and Android
- [x] Unit tests updated

### Related issues

- Fixes Act-918

### Backwards compatibility

Yes


MuckT:
I am seeing this warning in development mode: `flexWrap: 'wrap' is not supported with the 'VirtualizedList' components. Consider using 'numColumns' with 'FlatList' instead.` which seems concerning. There is some discussion about it in https://github.com/facebook/react-native/issues/18079. Alternatively to ignoring this warning I could render one item which would be a FlatList containing all of the Nfts, but that also seems rather hacky.

src/tokens/Assets.tsx
```diff
@@ -268,18 +286,91 @@ function AssetsScreen({ navigation, route }: Props) {
...
+      <View
+        testID="NftItem"
+        style={[
+          styles.nftsTouchableContainer,
+          // For even indexes, add right margin; for odd indexes, add left margin.
+          // If the index is even and it's the last image, add a right margin to left-align the image in the last row.
+          index % 2 === 0
+            ? { marginRight: Spacing.Regular16 } &&
+              index === nfts.length - 1 &&
+              styles.nftsLastImage
```
satish-ravi, line 310: Not sure I understand this. Is the right margin added for even indices if its not the last item? It seems like `index === nfts.length - 1` would be false and this would ultimately result in a `false`. The right margin seems to be applied only for the last item (from `styles.nftsLastImage`) which also seems like it shouldn't  make any difference since the last row will have only one item in this case.

src/tokens/Assets.tsx
```diff
@@ -249,12 +265,14 @@ function AssetsScreen({ navigation, route }: Props) {
       ? [{ data: tokenItems }]
       : activeTab === AssetTabType.Positions
       ? positionSections
+      : nfts.length
+      ? [{ data: nfts }]
```
satish-ravi, line 269: nit: might be worth just doing an if else or switch and avoid the nested ternary.

satish-ravi:
> I am seeing this warning in development mode: `flexWrap: 'wrap' is not supported with the 'VirtualizedList' components. Consider using 'numColumns' with 'FlatList' instead.` which seems concerning. There is some discussion about it in [facebook/react-native#18079](https://github.com/facebook/react-native/issues/18079). Alternatively to ignoring this warning I could render one item which would be a `FlatList` containing all of the Nfts, but that also seems rather hacky.

This does seem concerning, given that the implication is not completely understood. One another way to implement this would be to make a section be 2 items and the render item function could render a single row with 2 items, which also seems a bit hacky, but maybe will make some of the styling cleaner.

src/tokens/Assets.tsx
MuckT: If the last image is an even index we only need to apply the right margin. If the last image is an odd index we need to apply special style to align it left, otherwise we get cases where the last image is in the center of the two columns.

src/tokens/Assets.tsx
```diff
@@ -55,20 +64,39 @@ import { TokenBalance } from 'src/tokens/slice'
 import { getSupportedNetworkIdsForTokenBalances, sortByUsdBalance } from 'src/tokens/utils'
 
 const DEVICE_WIDTH_BREAKPOINT = 340
+const NUM_OF_NFTS_PER_ROW = 2
+
+const nftImageSize =
+  (variables.width - Spacing.Regular16 * (NUM_OF_NFTS_PER_ROW + 1)) / NUM_OF_NFTS_PER_ROW
```
satish-ravi, line 70: per designs looks like the border on edges is 24, and border between images is 16

src/tokens/Assets.tsx
```diff
@@ -306,10 +414,16 @@ function AssetsScreen({ navigation, route }: Props) {
...
-        }}
+        contentContainerStyle={[
+          {
+            paddingBottom: insets.bottom,
+            opacity: listHeaderHeight > 0 ? 1 : 0,
+          },
+          activeTab === AssetTabType.Collectibles && styles.nftsContentContainer,
+          activeTab === AssetTabType.Collectibles && nftsError
+            ? { alignItems: 'center' }
+            : { paddingLeft: Spacing.Regular16 },
```
satish-ravi, line 425: Same as above comment, this should be 24, maybe this should go in nftsContentContainer, since the noNftsText also requires 24 margin instead of 16

satish-ravi (approved):
Approved once horizontal padding / margin is fixed to match designs

PR #4272, opened 2023-10-06
chore: remove merchant payments feature

kathaypacific:
### Description

Essentially this PR reverts #3376.

I've also removed the translations because I think it is unlikely that we will want to push over the air translations for this feature.

### Test plan

n/a

### Related issues

- Fixes RET-852

### Backwards compatibility

Y


jeanregisser (approved):
This feels good!! ✂️ 🚀

PR #4275, opened 2023-10-06
feat(transfers): Remove old feature flag, support tokenId in transfer feed and CICO

jophish:
### Description

This PR cleans up an old hacky feature flag we had laying around, and in the process adds support for native tokens to appear _correctly_ in the transfer feed. This also has the effect of refactoring some of our CICO code to support tokenIds as well, in order to use the new `TokenDisplay` component in the case of `ETH` being selected.

### Test plan

Unit and manual tested. See video below for ETH information showing up where it previously was not (namely: transfer feed items, transfer feed details page, CICO header balance). The second video shows the existing behavior, where ETH is kinda broken all over the place.

https://github.com/valora-inc/wallet/assets/569401/c7931d12-87a2-42a6-80de-516a18a343b8

https://github.com/valora-inc/wallet/assets/569401/b2bbb8e3-cee1-467b-85da-d7ce3ab0690c

### Related issues

N/A

### Backwards compatibility

Yes.


src/components/TokenTotalLineItem.tsx
```diff
@@ -4,15 +4,17 @@ import { Trans, useTranslation } from 'react-i18next'
 import { StyleSheet, Text } from 'react-native'
```
jophish, line 4: Note that this file is the only one that's not "fully" refactored, since it is used in many places outside the scope of this PR. Option 1 [here](https://www.figma.com/file/iiPCJG2cA18ysZrLAZI80d/Multichain-refactor-strategies?type=whiteboard&node-id=0-1&t=PCFrhpA7ruyapZgH-0) would be too much scope, and this doesn't fit neatly into options 2 or 3 -- perhaps this is a fourth option, which is "make an existing thing backwards compatible"? Though you could perhaps consider this to be a variant of option 3, since it remains DRY by not duplicating the component outright.

cajubelt: @jophish to me this looks like a pretty standard case where we would want option 3. How about creating a new TokenTotalLineItem that uses tokenId to identify tokens, and a LegacyTokenTotalLineItem (or "byAddress" version, whatever) that wraps it by getting token info from the token address, then getting the token id from there?

cajubelt: the effort here to make this component take either a token address and a tokenId here seems to have some downsides
- practically speaking, there is no easy way for consumers of this component to see that tokenId is the "preferred" way of using it. There isn't even a jsdoc we can use to show it's deprecated. We would rely on people looking into the actual implementation and seeing your todo there.
- this is going to sound kind of fuzzy, but it's how I think about it so here goes. at a high level, the function now serves two purposes and has two fairly distinct execution pathways depending on the inputs provided. to keep code simple, it's better to keep functions single-purpose

src/components/TokenTotalLineItem.tsx
```diff
@@ -37,13 +45,23 @@ export default function TokenTotalLineItem({
...
         textStyle={fontStyles.regular600}
         amount={
-          <LegacyTokenDisplay
-            amount={tokenAmount.plus(feeInToken ?? 0)}
-            tokenAddress={tokenAddress}
-            localAmount={localAmount}
-            hideSign={hideSign}
-            testID="TotalLineItem/Total"
-          />
+          tokenId ? (
```
cajubelt, line 48: if you were to keep the implementation you have-- which I still think you probly shouldn't-- what's the point of these usages of LegacyTokenDisplay? Aren't you guaranteed to have a tokenId by this point from `tokenInfo`?

src/components/TokenTotalLineItem.tsx
jophish: Yeah -- I think I agree. The change to split this into a `Legacy` component that wraps the new `tokenId`-based one was actually quite small; I should have gone with it from the start I think. Being able to `@deprecate` the old components is actually pretty valuable I think, since allows us to quickly see at a glance where in the app we still need to update.

cajubelt (approved):

PR #4278, opened 2023-10-06
refactor: make tokensByAddressSelector more DRY

cajubelt:
### Description

a toy example of where we can use strategy 3 for [this doc](https://www.figma.com/file/iiPCJG2cA18ysZrLAZI80d/Multichain-refactor-strategies?type=whiteboard&node-id=4-181&t=E6ENxluAmHAfiloW-0) (vertical refactor with DRY helper) and cut down on duplicated code as we do multichain refactors

obviously nothing earth-shattering here, but I'd like us to keep a lookout for opportunities to do more of this to avoid adding duplication and complexity as we add multichain features

### Test plan

existing coverage makes sure the bridge names are in the token name and ethereum is filtered out, so I think we're good

### Related issues

na

### Backwards compatibility

yes

### DIFF
--- a/src/localCurrency/hooks.test.tsx
+++ b/src/localCurrency/hooks.test.tsx
@@ -21,6 +21,7 @@ function createStore(usdToLocalRate: string | null = '2') {
       tokenBalances: {
         'celo-alfajores:0xcUSD': {
           networkId: NetworkId['celo-alfajores'],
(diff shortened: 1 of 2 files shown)


jophish (approved):
Great change -- thank you for cleaning this up!

PR #4289, opened 2023-10-10
chore(assets): Add feature gate for asset page redesign

finnian0826:
### Description

Add a feature gate to show if the new asset page redesign should be shown, replace Satish's dummy gate that always returned false.

Feature gate: https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/gates/show_asset_page_redesign

### Test plan

With feature gate returnin
feature_gate_false
g false:

With feature gate set to true for dev environment:

https://github.com/valora-inc/wallet/assets/140328381/627f4fb4-67c2-42a5-8de4-e67075cc6323

### Related issues

N/A

### Backwards compatibility

Yes, just updates existing boolean to use feature gate.


src/tokens/utils.ts
```diff
@@ -127,8 +127,7 @@ export function getSupportedNetworkIdsForTokenBalances(): NetworkId[] {
 }
 
 export function showAssetDetailsScreen() {
-  // TODO(ACT-919): get from feature gate
-  return false
+  return getFeatureGate(StatsigFeatureGates.SHOW_ASSET_PAGE_REDESIGN)
```
satish-ravi, line 130: I think we've generally favored dropping the wrapper dummy function whenever a feature gate is implemented. Although I don't feel too strongly on this. At the least let's keep the util function and feature gate names the same

src/components/TokenBalance.test.tsx
```diff
@@ -98,6 +94,12 @@ describe('FiatExchangeTokenBalance and HomeTokenBalance', () => {
         },
       })
 
+      jest
```
MuckT, line 97: I think we can define this globally in the file and then only use it where needed in the tests to reduce duplication.

cajubelt (approved):
🥳

PR #4304, opened 2023-10-12
fix: back button hit area too small

jeanregisser:
### Description

Increased back button hit area, which was too small and almost impossible to hit.

I'm not too happy about the manual padding added, I would have preferred the [hitSlop](https://github.com/valora-inc/wallet/blob/ede52965f035ce263c38e83839be64c7412cb3bb/src/navigator/TopBarButton.tsx#L47) to work, but somehow it doesn't.
We can investigate this more later. But for now this should be good enough.

### Test plan

**before/after**

Showing with the ripple effect:
 

Note: it would be nice for the ripple effect to not be cut off on the left, but this would mean moving the button more toward the right, and then not be aligned anymore with the content. We can discuss that with design later on.

### Related issues

- See Slack [thread](https://valora-app.slack.com/archives/C025V1D6F3J/p1696953791346979?thread_ts=1696932462.287309&cid=C025V1D6F3J)

### Backwards compatibility

Yes

### DIFF
--- a/src/components/BackButton.tsx
+++ b/src/components/BackButton.tsx
@@ -3,6 +3,7 @@ import { StyleSheet, View } from 'react-native'
 import BackChevron, { Props as BackChevronProps } from 'src/icons/BackChevron'
 import { navigateBack } from 'src/navigator/NavigationService'
 import { TopBarIconButton, TopBarIconButtonProps } from 'src/navigator/TopBarButton'
+import { Spacing } from 'src/styles/styles'
 
 type Props = Omit<TopBarIconButtonProps, 'icon'> & BackChevronProps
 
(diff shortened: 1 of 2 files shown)


src/components/BackButton.tsx
```diff
@@ -26,6 +27,13 @@ const styles = StyleSheet.create({
     justifyContent: 'center',
     alignItems: 'center',
   },
+  button: {
+    // Quick hack to workaround hitSlop set for the internal Touchable component not working
+    // I tried removing the parent view, but it didn't help either
+    paddingHorizontal: 16,
+    paddingVertical: 12, // vertical padding slightly smaller so the ripple effect isn't cut off
```
kathaypacific, line 34: should we use `Spacing`?

kathaypacific (approved):
TY! LGTM, curious did you find out which change introduced this regression?

jeanregisser:
I don't know what caused this change. But it seems it was already problematic for some time. I installed 1.67 and it was already difficult to hit.

PR #4306, opened 2023-10-12
feat(sends): Support sending native/ERC20 non-Celo assets using Viem

jophish:
### Description

For ACT-904. Adds support for e2e sends on Ethereum, including both the native ETH asset and arbitrary ERC20 tokens.

A couple of notes:
* Supporting non-Celo sends requires _two_ separate ways of submitting TXs to the blockchain. For ERC20 sends, where we definitely have a contract address, we simply re-use the existing contract simulation/request logic, extending it to Viem clients for other chains. For native sends, we don't have a contract address available, so we must simulate the send using a `call`, followed by a `sendTransaction`.
* Since we don't have Ethereum fee estimation, fees have been made optional for Ethereum for the time being, while still being required for all sends on Celo.

### Test plan

Unit and manual tested. Note that in the videos below, the standby transaction is added correctly, but since I'm sending the assets to my own address, blockchain-api sees that transaction as containing exactly 1 send and 1 receive, so classifies it as a swap in the TX feed once the TX has been confirmed.

#### Native ETH Send

https://github.com/valora-inc/wallet/assets/569401/306d0fca-1721-4f83-a613-5c38db8931b0

#### ERC20 USDC Send

https://github.com/valora-inc/wallet/assets/569401/7d2973ba-355e-4ff4-b8c9-54074dc8e586

### Related issues

- Fixes ACT-904

### Backwards compatibility

Yes.


src/viem/saga.ts
```diff
@@ -74,53 +82,117 @@ export function* sendPayment({
...
+      // TODO (ACT-922): Fill in fee estimation for native token transfers (non-Celo)
+      const callMethod = async () =>
+        await publicClient[network].call({
+          account: wallet.account,
+          to: getAddress(recipientAddress),
+          data: '0x00',
+          value: convertedAmount,
+        })
+      Logger.info(TAG, 'Call completed successfully')
+      const callData = yield* call(callMethod)
```
jophish, line 171: I've not seen `callData` end up as anything other than an empty object. The [call docs](https://viem.sh/docs/actions/public/call.html) mark `data` as a required field, but I've been able to omit `data` without TS yelling at me. I'm not convinced that this is how we ought to simulate native sends, or if it's possible(???) using Veim?

src/viem/saga.ts
```diff
@@ -146,30 +218,45 @@ function* getTransferSimulateContract({
...
-    tokenAddress,
-    feeInfo,
-    encryptedComment: encryptedComment || '',
-  })
+  const network = getNetworkFromNetworkId(tokenInfo?.networkId)
+  if (!network) {
+    throw new Error('invalid network for transfer')
+  }
+
+  // TODO (ACT-922): Use real fee estimation for Ethereum
```
jophish, line 246: Since we don't yet have fee estimation for Ethereum, I'm punting on it for this ticket, and allowing non-Celo sends to work without fees being available. This is kind of a hack, and will need to be reverted once ACT-922 is complete.

src/viem/saga.ts
```diff
@@ -74,53 +82,117 @@ export function* sendPayment({
...
+          value: convertedAmount,
+        })
+      Logger.info(TAG, 'Call completed successfully')
+      const callData = yield* call(callMethod)
+      Logger.info(TAG, `calldata: ${JSON.stringify(callData, null, 4)}`)
+      yield* call(unlockAndAddStandby)
+
+      const sendNativeTxMethod = function* () {
+        try {
+          const hash = yield* call([wallet, 'sendTransaction'], {
```
jophish, line 177: So, I've seen some very weird behavior here. On Sepolia, this reliably results in a 503 from the RPC node, with no real indication as to what the actual error is. I've tried using different Sepolia RPC nodes (by using a custom URL in the `transport` field for the Ethereum client in `src/viem/index.ts`) but received similar errors. 

The address that I've been using for testing _also_ has a balance on Goerli. Out of curiosity, I reconfigured the Ethereum client to use the Goerli chain with the default RPC node, and it actually worked, and successfully sent a TX [here](https://goerli.etherscan.io/tx/0xe0fdffc41dc13b6b24f9dd7c19fc9751d961fe40e750920549c854c9be611b99). This is really confusing to me -- I'm not sure what/if there's a meaningful difference between these two testnets.

src/viem/saga.ts
```diff
@@ -200,7 +287,7 @@ function* getTransferSimulateContract({
   })
 
   return () =>
-    publicClient.celo.simulateContract({
+    publicClient[network].simulateContract({
```
jophish, line 290: This should in theory "just work" for non-Celo ERC20 tokens (i.e., those with an address). I haven't been able to test this though, since my test wallet only has ETH.

src/send/saga.ts
```diff
@@ -257,19 +267,22 @@ function* sendPayment(
...
   usdAmount,
   comment,
   recipient,
-  feeInfo,
   fromModal,
+  feeInfo,
 }: SendPaymentAction) {
   try {
     yield* call(getConnectedUnlockedAccount)
+    Logger.info(TAG, 'starting tx')
```
cajubelt, line 279: should these be debug log level?

src/send/saga.ts
```diff
@@ -288,6 +301,7 @@ export function* sendPaymentSaga({
     yield* put(sendPaymentSuccess(amount))
     SentryTransactionHub.finishTransaction(SentryTransaction.send_payment)
   } catch (e) {
+    Logger.info(TAG, e)
```
cajubelt, line 304: same here

src/tokens/saga.ts
```diff
@@ -201,23 +205,33 @@ export function* fetchTokenBalancesSaga() {
   }
 }
 
-export function* tokenAmountInSmallestUnit(amount: BigNumber, tokenAddress: string) {
-  const tokens: TokenBalance[] = yield* select(tokensListWithAddressSelector)
-  const tokenInfo = tokens.find((token) => token.address === tokenAddress)
+export function* tokenAmountInSmallestUnit(amount: BigNumber, tokenId?: string) {
+  const networkIds = Object.values(networkConfig.networkToNetworkId)
```
cajubelt, line 209: we should probably have a constant somewhere for ALL_NETWORK_IDS

src/tokens/saga.ts
```diff
@@ -201,23 +205,33 @@ export function* fetchTokenBalancesSaga() {
...
+  const tokens: TokenBalance[] = yield* select((state) => tokensListSelector(state, networkIds))
+  const tokenInfo = tokens.find((token) => token.tokenId === tokenId)
```
cajubelt, line 211:
```suggestion
  const tokenInfo = yield* getTokenInfo(tokenId)
```

src/tokens/utils.ts
```diff
@@ -33,7 +33,11 @@ export function sortByUsdBalance(token1: TokenBalance, token2: TokenBalance) {
 }
 
 export function isStablecoin(token: TokenBalance | undefined) {
-  return token?.isCoreToken && token.symbol !== 'CELO'
+  return (
+    token?.isCoreToken &&
+    token.symbol !== 'CELO' &&
+    token.networkId === networkConfig.networkToNetworkId[Network.Celo]
+  )
```
cajubelt, line 40: there is an `isStableCoin` property now in address-metadata, can we just use that?

cajubelt: oh interesting, this function is used to see if it's a contract that supports transfer comments. So misleading! That's not a property of stablecoins, it's a property of a subset of Celo contracts...the function should be renamed to make that more clear

src/viem/saga.ts
```diff
@@ -135,7 +199,7 @@ export function* sendPayment({
  */
 function* getTransferSimulateContract({
   wallet,
-  tokenAddress,
+  tokenId,
```
cajubelt, line 202: why refactor this to take token Id's if it should never be called for native assets? should it take token address and network id instead?

cajubelt: seems like requiring those two fields would give us static validation instead of runtime errors, which would be preferable

cajubelt (commented):
very cool!

src/tokens/utils.ts
jophish: Yeah -- it's pretty weird. I've updated the function name to reflect this usage.

src/viem/saga.ts
jophish: Yeah, after going back and forth on this, I've updated this to take in a `tokenAddress` and a `tokenInfo` object, which gives us everything we need, and means we don't have to use the deprecated `byAddress` hooks to get the full token info object.

src/analytics/Properties.tsx
```diff
@@ -552,7 +552,8 @@ interface SendEventsProperties {
         localCurrencyAmount: string | null
         tokenAmount: string
         tokenSymbol: string
-        tokenAddress: string
+        tokenAddress?: string
```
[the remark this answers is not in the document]
jophish: good catch -- added a networkId field here and adjusted the types.

src/web3/utils.ts
```diff
@@ -93,3 +95,10 @@ export function buildTxo(kit: ContractKit, tx: CeloTx): CeloTxObject<never> {
     },
   }
 }
+
+export function getNetworkFromNetworkId(networkId?: NetworkId): Network | undefined {
```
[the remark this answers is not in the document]
jophish: Hm, not entirely sure I understand. We still need the env-agnostic concept of `Network` throughout the app, since certain features (like these sends) care about the actual blockchain, not whether or not we're on test/mainnet. Without using `Network`, we'd need to explicitly check e.g. `if(networkId === NetworkId['ethereum-mainnet'] || networkId === NetworkId['ethereum-sepolia'])`.

src/web3/networkConfig.ts
```diff
@@ -68,8 +68,12 @@ interface NetworkConfig {
     [key in CiCoCurrency | Currency]: string
   }
   celoTokenAddress: string
+  alchemyRpcUrl: string
```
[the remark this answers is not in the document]
jophish: Oh, that's a good point. Will fix this.

src/viem/saga.ts
```diff
@@ -277,15 +360,15 @@ export function* sendAndMonitorTransaction({
...
     ValoraAnalytics.track(TransactionEvents.transaction_hash_received, {
       ...commonTxAnalyticsProps,
       txHash: hash,
     })
     yield* put(addHashToStandbyTransaction(context.id, hash))
-    const receipt = yield* call([publicClient.celo, 'waitForTransactionReceipt'], { hash })
+    const receipt = yield* call([publicClient[network], 'waitForTransactionReceipt'], { hash })
     ValoraAnalytics.track(TransactionEvents.transaction_receipt_received, commonTxAnalyticsProps)
-    return receipt
+    return receipt as unknown as TransactionReceipt
```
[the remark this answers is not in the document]
jophish: Sure, added a comment here.

src/viem/getLockableWallet.ts
```diff
@@ -9,13 +9,25 @@ import {
...
 import { privateKeyToAccount } from 'viem/accounts'
-import { writeContract } from 'viem/actions'
+import { writeContract, sendTransaction } from 'viem/actions'
+import { viemTransports } from 'src/viem'
+import networkConfig from 'src/web3/networkConfig'
+import { Network } from 'src/transactions/types'
 
 const TAG = 'viem/getLockableWallet'
 
+export function getTransport(chain: Chain): Transport {
```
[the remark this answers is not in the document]
jophish: Yes, it's definitely a bit strange.. we have to do a reverse lookup since we create the viem wallet using `Chain`, which is the _value_ in the `networkConfig.viemChain` mapping, but `Network` is used to key the `viemTransports` mapping. We unfortunately can't use `Chain` to key this transport mapping, since it's not a simple type. Yeah -- this should be simpler the viem wallet creation uses `Network`/`NetworkId` instead.

src/tokens/saga.ts
```diff
@@ -206,9 +206,10 @@ export function* fetchTokenBalancesSaga() {
 }
 
 export function* tokenAmountInSmallestUnit(amount: BigNumber, tokenId?: string) {
-  const networkIds = Object.values(networkConfig.networkToNetworkId)
-  const tokens: TokenBalance[] = yield* select((state) => tokensListSelector(state, networkIds))
-  const tokenInfo = tokens.find((token) => token.tokenId === tokenId)
+  if (!tokenId) {
```
cajubelt, line 209: I dislike when a function allows null/undefined in the function signature, but then throws a runtime error if it's actually called with null/undefined. It requires you to open up and read through every function to know if there is real null safety. (harder to understand, less safe from bugs)

it looks like all the consumers of this function have an entire `tokenInfo` object at their disposal.. maybe the "decimals" field from that could be taken as a parameter instead? And maybe throw one level higher if tokenInfo is null in `buildAndSendTx`? (the only place this is used where `tokenInfo` can currently be null)

src/viem/saga.ts
```diff
@@ -199,36 +201,33 @@ export function* sendPayment({
...
+  tokenAddress,
+  tokenInfo,
   amount,
   recipientAddress,
   comment,
   feeInfo,
 }: {
   wallet: ViemWallet
+  tokenAddress: string
+  tokenInfo: TokenBalance
```
cajubelt, line 213: seems like you could use `TokenBalanceWithAddress` here

cajubelt: given that there are ongoing use cases for dealing with tokens that have addresses differently (like right here), we may not ever want to deprecate that type

src/viem/saga.ts
```diff
@@ -199,36 +201,33 @@ export function* sendPayment({
  */
 function* getTransferSimulateContract({
   wallet,
-  tokenId,
+  tokenAddress,
+  tokenInfo,
```
cajubelt, line 205: this raises the rather uncomfortable possibility that tokenAddress will be different from the address in tokenInfo

cajubelt (commented):
looks close!

src/tokens/saga.ts
jophish: Yeah, you're totally right -- using `decimals` as an arg feels way cleaner to me.

src/viem/saga.ts
jophish: Fortunately, using `TokenBalanceWithAddress` means we can safely get rid of this parameter.

src/viem/saga.ts
jophish: Yeah, this is a good call. I agree we may not want to deprecate this type, but I wonder how we would differentiate between cases where its usage actually _is_ deprecated (i.e., somewhere we truly want to transition over to using the new types) vs this use, which is actually valid.

src/web3/networkConfig.ts
```diff
@@ -66,6 +66,14 @@ interface NetworkConfig {
     [key in CiCoCurrency | Currency]: string
   }
   celoTokenAddress: string
+  alchemyEthereumRpcUrl: string
+}
+
+const ALCHEMY_ETHEREUM_RPC_URL_STAGING = 'https://eth-sepolia.g.alchemy.com/v2/'
+const ALCHEMY_ETHEREUM_RPC_URL_MAINNET = 'https://eth-mainnet.g.alchemy.com/v2/'
+
+export type blockExplorerUrls = {
```
cajubelt, line 75: kinda weird to me how there is a lowercase-b version of this type and an uppercase-B version, which are similar but a little different

cajubelt (approved):

PR #4308, opened 2023-10-13
chore: remove request feature

cajubelt:
### Description

very few people (0.5% of active users per month) use this, and it would take time to add safeguards to prevent it from being used for phishing. Silas and Denisse asked us to remove the feature citing infosec risks of keeping it around.

### Test plan

ci

manually tested send flow, still works

https://github.com/valora-inc/wallet/assets/7979215/c5414e2c-fd52-40a5-87da-80987603e8da

### Related issues

fixes https://linear.app/valora/issue/ACT-939/remove-requests-feature-from-wallet

### Backwards compatibility

versions including this change will not receive, or be able to send, payment request notifications. Additionally, they will not be able to consume payment request QR codes built for an also-unused merchant payments feature.


src/redux/migrations.ts
```diff
@@ -1331,4 +1331,17 @@ export const migrations = {
...
+      ...state,
+      app: {
+        ...state.app,
+        activeScreen: Object.values(Screens).includes(state.app.activeScreen)
+          ? state.app.activeScreen
+          : 'Main', // since PaymentRequestConfirmation screen was deleted
+      },
+    }
+    delete output.paymentRequest // doing this instead of output = {..._.omit(state, 'paymentRequest')} because _.omit removes underscore-prefixed properties too, like the required _persist key
+    return output
```
jeanregisser, line 1345: wow that's weird 🤔

jeanregisser (approved):
Awesome! 🚀 🚀 🚀 
Damn this feature had so many lines of code.

Probably already on your radar, but we can then remove https://github.com/valora-inc/cloud-functions/blob/main/src/paymentRequests too.

PR #4309, opened 2023-10-13
feat: more actions token bottom sheet

MuckT:
### Description

Adds a more actions token bottom sheet if the more actions button is present. Moves actions to a hook called `useActions` that takes a token as a parameter. Moves `onPressCicoAction` to a shared function that is exported. It appears we will likely need a new feature flag for swaps as there is currently a coupling of the swap display in the drawer and this component which seem unrelated.

| iOS | Android |
| --- | --- |
|  |  |

### Test plan
- [x] Tested locally on iOS
- [x] Unit tests added

### Related issues

- ACT-917

### Backwards compatibility

Yes


src/tokens/TokenDetails.tsx
```diff
@@ -136,7 +136,21 @@ function PriceInfo({ token }: { token: TokenBalance }) {
...
+export const onPressCicoAction = (token: TokenBalance, flow: CICOFlow) => {
+  const tokenSymbol = token.symbol
+  // this should always be true given that we only show Add / Withdraw if a
+  // token is CiCoCurrency, but adding it here to ensure type safety
+  if (isCicoToken(tokenSymbol)) {
+    navigate(Screens.FiatExchangeAmount, {
+      currency: tokenSymbol,
+      tokenId: token.tokenId,
+      flow,
+      network: Network.Celo, // TODO: use networkId from token
```
MuckT, line 148: I see we're using the old Network enum here. Should this be updated in the fiatExchange to use networkId?

src/tokens/TokenDetails.tsx
```diff
@@ -136,7 +136,21 @@ function PriceInfo({ token }: { token: TokenBalance }) {
...
+    })
+  }
+}
+
+export const useActions = (token: TokenBalance) => {
```
MuckT, line 153: Made this a hook that takes a token and returns an array of actions, curious to hear others opinions of if this is needed / good practice.

src/tokens/TokenDetails.tsx
```diff
@@ -145,24 +159,11 @@ function Actions({ token }: { token: TokenBalance }) {
   const isSwapEnabled = useSelector(isAppSwapsEnabledSelector)
```
MuckT, line 159: We might need a new feature gate for this as currently the only time swaps would display in more actions for a supported token is when swaps are also displayed in the hamburger drawer.

src/tokens/TokenDetails.tsx
satish-ravi: I think this would result in re-computing the list of actions in both the actions component and the bottom sheet. One option might be to compute the list of eligible actions in the Action component behind a use memo and then pass a list of `TokenDetailActionName`s as a prop to the bottom sheet screen. And then maybe we can also have a map of `TokenDetailActionName` to `TokenDetailsAction` which can be used to lookup other properties for rendering

src/tokens/TokenDetails.tsx
satish-ravi: Why do we need a new feature gate? I believe the remote config, even though is called `showSwapMenuInDrawerMenu`, is actually just to used to turn on / off swap action in the home screen actions. We no longer show swap in the drawer menu.

src/tokens/TokenDetails.tsx
```diff
@@ -136,7 +136,21 @@ function PriceInfo({ token }: { token: TokenBalance }) {
   )
 }
 
-function Actions({ token }: { token: TokenBalance }) {
+export const onPressCicoAction = (token: TokenBalance, flow: CICOFlow) => {
+  const tokenSymbol = token.symbol
+  // this should always be true given that we only show Add / Withdraw if a
+  // token is CiCoCurrency, but adding it here to ensure type safety
+  if (isCicoToken(tokenSymbol)) {
```
satish-ravi, line 143: there's an open question on where Withdraw should navigate right?

src/tokens/TokenDetails.tsx
MuckT: `showSwapMenuInDrawerMenu` if set to false shows swap actions in the home screen action carousel. Then in `src/tokens/TokenDetails.tsx` we only display swaps as an action if `showSwapMenuInDrawerMenu` is true. If we ever want to display swaps in the home actions and in the token actions. Aside from that I'd argue that setting `showSwapMenuInDrawerMenu` to false to make it display in the home carousel is not very clear based on the remote config name. This could be addressed in a follow up PR.

MuckT: Got the double negatives mixed up 🤦. You're right this is essentially just a swaps enabled feature gate.

src/tokens/TokenDetails.tsx
MuckT: This is still an open question @nityarao responded in [Figma](https://www.figma.com/file/OE6RFzqXAMqU9ygHtNYyof/Assets-Page-(Multichain)?node-id=2490%3A11056&mode=dev#579471570).

src/tokens/TokenDetailsMoreActions.tsx
```diff
@@ -0,0 +1,80 @@
...
+}) {
+  const { t } = useTranslation()
+  const token = useTokenInfo(tokenId)
+  if (!token) throw new Error(`Token ${tokenId} not found`)
+
+  return (
+    <BottomSheet
+      forwardedRef={forwardedRef}
+      title={t('tokenDetails.moreActions')}
+      testId={'TokenDetailsMoreActions'}
```
satish-ravi, line 32: can we set a `titleStyle` based on the designs? I believe the bottom sheet defaults to one of the old font styles

src/tokens/TokenDetailsMoreActions.tsx
```diff
@@ -0,0 +1,80 @@
...
+  actions,
+  tokenId,
+}: {
+  forwardedRef: RefObject<BottomSheetRefType>
+  actions: TokenDetailsAction[]
+  tokenId: string
+}) {
+  const { t } = useTranslation()
+  const token = useTokenInfo(tokenId)
+  if (!token) throw new Error(`Token ${tokenId} not found`)
```
satish-ravi, line 26: Since this is now just a component and not a screen, we can just pass the token itself directly instead of tokenId and doing the lookup.

MuckT: fixed in https://github.com/valora-inc/wallet/pull/4309/commits/d001e30e9196c60e10130c64debc8c1219ece3df!

src/tokens/TokenDetailsMoreActions.tsx
MuckT: Fixed in https://github.com/valora-inc/wallet/pull/4309/commits/d001e30e9196c60e10130c64debc8c1219ece3df!

src/tokens/TokenDetailsMoreActions.test.tsx
```diff
@@ -0,0 +1,179 @@
...
+import { Screens } from 'src/navigator/Screens'
+import { onPressCicoAction } from 'src/tokens/TokenDetails'
+import TokenDetailsMoreActions from 'src/tokens/TokenDetailsMoreActions'
+import { StoredTokenBalance, TokenBalance } from 'src/tokens/slice'
+import { TokenDetailsAction, TokenDetailsActionName } from 'src/tokens/types'
+import { Network, NetworkId } from 'src/transactions/types'
+import { createMockStore } from 'test/utils'
+import { mockCeloAddress, mockCeloTokenId } from 'test/values'
+
+jest.mock('src/statsig', () => ({
```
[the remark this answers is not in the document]
MuckT: Removed in https://github.com/valora-inc/wallet/pull/4309/commits/d001e30e9196c60e10130c64debc8c1219ece3df!

src/tokens/TokenDetails.tsx
```diff
@@ -183,30 +195,45 @@ function Actions({ token }: { token: TokenBalance }) {
...
+
+function Actions({
+  token,
+  bottomSheetRef,
+}: {
+  token: TokenBalance
+  bottomSheetRef: React.RefObject<BottomSheetRefType>
+}) {
+  const { t } = useTranslation()
+  const actions = useActions(token)
```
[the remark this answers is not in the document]
MuckT: 🤦 yes we can. Fixed in https://github.com/valora-inc/wallet/pull/4309/commits/d001e30e9196c60e10130c64debc8c1219ece3df!

src/tokens/TokenDetailsMoreActions.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+const store = createMockStore({
+  tokens: {
+    tokenBalances: {
+      [mockCeloTokenId]: mockStoredCeloTokenBalance,
+    },
+  },
+  app: {
+    showSwapMenuInDrawerMenu: true,
+  },
+})
```
satish-ravi, line 56: seems like these can also be removed since there's no redux dependency now

src/tokens/TokenDetails.tsx
satish-ravi: I left a comment on the thread. This can be fixed in a follow up PR

satish-ravi (approved):

PR #4310, opened 2023-10-13
feat(send): Create new send confirmation screen

finnian0826:
### Description

Creates new send confirmation screen which shows amount in token currency (with conversion shown in fiat currency).

Feature gate: https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/gates/show_new_send_confirmation_screen

### Test plan

Updated/added unit tests.

New screen for CELO send:
new_screen_celo

New screen for cUSD send:
new_screen_cusd

### Related issues

https://linear.app/valora/issue/ACT-944/create-new-send-confirmation-screen

### Backwards compatibility

Maintains the old behavior, new behavior is behind a feature gate.


src/components/TokenTotalLineItem.tsx
```diff
@@ -63,9 +77,9 @@ export default function TokenTotalLineItem({
         }
         amount={
           <TokenDisplay
-            amount={tokenAmount}
+            amount={tokenAmount.plus(newSendScreen ? feeInToken ?? 0 : 0)}
             tokenId={tokenId}
-            showLocalAmount={false}
+            showLocalAmount={!showLocalAmount}
```
MuckT, line 82: This is confusing is it correct that the show showLocalAmount prop takes the opposite of showLocalAmount?

finnian0826: Yeah this part was a bit confusing but the logic is correct. There are two totals being displayed here, the actual total, and the converted total. We want these to be opposite (i.e. if one is in local amount, the other should be in crypto amount). The `showLocalAmount` param is for the Total, and this part of code is for the exchange rate where we want to do the opposite.

I could create an additional variable `const showLocalAmountForConversion = !showLocalAmount` to make it more clear, or I could add a comment if that would be sufficient.

finnian0826: I added in a comment that explains this

src/components/TokenDisplay.tsx
```diff
@@ -68,7 +70,15 @@ function TokenDisplay({
 
   const amountToShow = showLocalAmount ? amountInLocalCurrency : new BigNumber(amount)
 
-  const sign = hideSign ? '' : amountToShow.isNegative() ? '-' : showExplicitPositiveSign ? '+' : ''
+  const sign = hideSign
```
satish-ravi, line 73: I think we should keep the approx symbol and +/- signs separate. This doesn't seem like the right abstraction, and it creates dependencies between the two options (e.g., if `hideSign` and `showApprox` are both true, `hideSign` takes precedence).

src/components/TokenTotalLineItem.tsx
satish-ravi: It seems like this component displays both a local currency amount and a fiat amount, and it looks like `showLocalAmount` actually means whether local amount is displayed in the first right row? Maybe this can be named better. Also, it looks like the new props added to this component are essentially only sent on the SendConfirmation page and seems to be derived from `newSendScreen` feature gate. If we want to keep this component generic enough, we should having a param called `newSendScreen`, which is specific to a usage of this component. Maybe its just simpler to make a new version of TokenTotalLineItem for the new send confirmation screen?

src/fees/saga.ts
satish-ravi, line 1: Should we hold off on all the fee related changes here and just use the new fee stuff per ACT-955? cc @cajubelt

src/send/SendConfirmation.tsx
```diff
@@ -233,7 +269,10 @@ function SendConfirmation(props: Props) {
     )
   }
 
-  const allowComment = isStablecoin(tokenInfo)
+  const allowComment = newSendScreen
```
satish-ravi, line 272: This would end up allowing comments for any celo token right? I think only celo stables support comments, so this line can just stay as is (we should rename the fn though)

src/send/SendConfirmation.tsx
```diff
@@ -360,6 +408,11 @@ const styles = StyleSheet.create({
     paddingVertical: 8,
     ...fontStyles.largeNumber,
   },
+  amountSubscript: {
+    ...fontStyles.regular,
```
satish-ravi, line 412: this is one of the deprecated font styles right? should we use one of the new font styles?

src/send/SendConfirmation.tsx
finnian0826: The [linear issue](https://linear.app/valora/issue/ACT-944/create-new-send-confirmation-screen) says it should appear if the selected token is on the Celo network, but maybe they meant to keep the current behavior but only for Celo network. I can change it to do that.

MuckT (approved):

src/send/SendConfirmation.tsx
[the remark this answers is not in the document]
finnian0826: Yeah the linear issue was ambiguous, I am updated to just allow comments for stables on Celo

src/send/SendConfirmation.tsx
```diff
@@ -36,18 +37,22 @@ import { noHeader } from 'src/navigator/Headers'
 import { navigate } from 'src/navigator/NavigationService'
 import { Screens } from 'src/navigator/Screens'
 import { StackParamList } from 'src/navigator/types'
-import { getDisplayName, Recipient, RecipientType } from 'src/recipients/recipient'
+import { Recipient, RecipientType, getDisplayName } from 'src/recipients/recipient'
 import useSelector from 'src/redux/useSelector'
 import { useInputAmountsByAddress } from 'src/send/SendAmount'
```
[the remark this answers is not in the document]
finnian0826: I think there will need to be other updates to get this to work for ETH sends, for example the props for the SendConfirmation component take a tokenAddress, so it seems like that will have to be refactored to take a tokenId, but I don't think that is in scope for [ACT-944](https://linear.app/valora/issue/ACT-944/create-new-send-confirmation-screen)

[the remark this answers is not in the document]
finnian0826: Ahh okay looks like Joe takes care of this in #4306

src/send/SendConfirmation.tsx
```diff
@@ -97,6 +102,8 @@ function SendConfirmation(props: Props) {
     },
   } = props.route.params
 
+  const newSendScreen = getFeatureGate(StatsigFeatureGates.SHOW_NEW_SEND_CONFIRMATION_SCREEN)
```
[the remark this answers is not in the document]
finnian0826: that existing feature gate only filters for test users, whereas the one I added also has a filter where if USE_VIEM_FOR_SEND is disabled, then it is also disabled.

I could also add that logic to that existing feature gate, but I'm not sure if we want to disable the bottom sheet when viem is not being used?

src/components/TokenTotalLineItem.tsx
finnian0826: @satish-ravi I just pushed some changes to clean up TokenTotalLineItem. There are really only two pieces that are new, the `showApprox` and the different exchange rate string. It seems like it would be a bit overkill to make a different component for this since other than 2 lines they are exactly the same. 

I also think it makes sense to have a param for showing local amount (I renamed it to `showLocalAmountForTotal` to clear up confusion), since then when using this component you can decide which order you want (fiat for total and crypto below or vice versa). 

However if these changes make it confusing or there is a better way to implement a new component without as much duplication let me know and i can do that.

src/fees/saga.ts
```diff
@@ -29,12 +29,17 @@ const TAG = 'fees/saga'
...
 export const SWAP_CELO_FEE_ESTIMATE_MULTIPLIER = 3 * SWAP_FEE_ESTIMATE_MULTIPLIER
 
-export interface FeeInfo {
+export type FeeInfo = {
   fee: BigNumber
   gas: BigNumber
   gasPrice: BigNumber
-  feeCurrency: string | undefined
-}
+} & (
```
cajubelt, line 36: I know I probly suggested this discriminated union trick, but it seems like it forced you to use a bunch of clunky ternaries, which will slow things down later if we want to make feeCurrency optional even when feeTokenId is defined, for instance, to support native assets that don't have an address. (since you have things like `if (feeCurrency && feeTokenId) { ... }` in a few places just to avoid type errors, and each of those would need to be redone)

cajubelt: might be better to just use `feeCurrency?: string; feeTokenId?: string`

src/send/SendConfirmation.test.tsx
```diff
@@ -320,7 +360,9 @@ describe('SendConfirmation', () => {
     expect(queryByTestId('commentInput/send')).toBeFalsy()
   })
 
-  it('doesnt show the comment for non core tokens', () => {
+  // TODO: Need to make this test for new and old version
```
cajubelt, line 363: same here

cajubelt (commented):
looks close!

src/send/SendConfirmation.tsx
```diff
@@ -97,6 +101,8 @@ function SendConfirmation(props: Props) {
     },
   } = props.route.params
```
satish-ravi, line 102: Seems like this screen still accepts tokenAddress as a param. Should this be updated to accept tokenId instead?
Also it accepts an `inputAmount` (can be either token or local currency amount based on the `amountIsInLocalCurrency` param) and `tokenAmount` but in the new flow the user is only allowed to enter token amounts, so some of these params seem redundant in the new flow. 
I think we should just make a new screen for the new send flow with just the required params and tokenId. There maybe some duplication but it will avoid a bunch of boolean based rendering and make cleanup easier when the new flow is rolled out to 100%.

src/components/TokenTotalLineItem.test.tsx
```diff
@@ -160,4 +165,30 @@ describe('TokenTotalLineItem', () => {
       expect(getElementText(getByTestId('TotalLineItem/Subtotal'))).toEqual('10.00 cUSD')
     })
   })
+
+  describe('When called from new send screen', () => {
```
satish-ravi, line 169: nit: can we avoid the new send screen naming?

src/components/FeeDrawer.tsx
```diff
@@ -40,8 +36,8 @@ export default function FeeDrawer({
   feeHasError,
   totalFee,
   testID,
-  currencyInfo,
   showLocalAmount,
+  tokenId,
```
jophish, line 40: Are `FeeDrawer` and `LegacyFeeDrawer` an example of where the "Legacy" version could be a wrapper over the new, `tokenId` version? e.g., `LegacyFeeDrawer` could continue to accept its current parameters, but internally, get a `tokenId` from the given `currency` and pass it through to `FeeDrawer`?

src/components/LegacyFeeDrawer.tsx
```diff
@@ -0,0 +1,183 @@
...
+  totalFee?: BigNumber
+  testID?: string
+  currencyInfo?: CurrencyInfo
+  showLocalAmount?: boolean
+}
+
+/**
+ * @deprecated use FeeDrawer instead
+ */
+export default function LegacyFeeDrawer({
```
jophish, line 34: I'm assuming that this file is completely unchanged, and just a rename? (which GH doesn't pick up since `FeeDrawer.tsx` still exists)

src/send/SendConfirmation.test.tsx
```diff
@@ -183,7 +199,26 @@ describe('SendConfirmation', () => {
...
+    const feeComponent = getByTestId('feeDrawer/SendConfirmation/totalFee')
+    expect(getElementText(feeComponent)).toEqual('0.004 CELO')
+
+    const totalComponent = getByTestId('TotalLineItem/Total')
+    expect(getElementText(totalComponent)).toEqual('~1.02 cUSD')
+  })
+
+  // TODO: Need to make for old and new
+  it('renders correctly for send payment confirmation with cEUR fees (old UI)', async () => {
+    jest.mocked(getFeatureGate).mockReturnValue(false)
```
jophish, line 221: nit: you can set a default mock for `getFeatureGate` in the `jest.mock('src/statsig')` line at the top of this file so you don't have to manually mock the return value in quite as many tests.

src/send/SendConfirmation.tsx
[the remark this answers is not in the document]
jophish: Yeah I think this whole new send flow should prolly use two feature flags: one for the new recipient screen, and one for the rest, e.g. `SHOW_NEW_RECIPIENT_SCREEN` and  `SHOW_NEW_SEND_FLOW` (prolly there are better names), with the following relations:

* `SHOW_NEW_RECIPIENT_SCREEN` implies `SHOW_NEW_SEND_FLOW`
* `SHOW_NEW_SEND_FLOW` implies `USE_VIEM_FOR_SEND`

assuming that `USE_VIEM_FOR_SEND` still exists.

src/components/LegacyFeeDrawer.tsx
finnian0826: yep

src/components/FeeDrawer.tsx
finnian0826: I don't think so, `LegacyFeeDrawer` uses a weird `currency` param that is hard coded in it's use cases but to different values, and it also uses the (now) deprecated `CurrencyDisplay` component which does some weird stuff behind the scenes.

src/send/SendConfirmation.tsx
finnian0826: I think making a new screen might be difficult to coordinate, @jophish has changes made to `SendConfirmation` [here](https://github.com/valora-inc/wallet/pull/4306/files#diff-b79fcaffc5d8f13b96f758016a7d2d6ee2a3f180d4b785b0c112793f6cfc7cf2) in #4306 which adds in a tokenId param as well as other changes for multichain, so I'm not sure that it would be easy to create a new screen and make sure it has my changes as well as the necessary multichain changes from Joe.

satish-ravi: From a quick look at the changes on Joe's PR, it seems like there'd be a messy merge conflict with the changes you have currently. You could just create a new screen with the new UI that just directly supports the multichain params, and we also get the benefit of dropping unnecessary params for the new screen and makes it easier to cleanup once the new stuff is rolled out. But I'll leave it up to you since you're closer to the details

jophish: Yeah this is kind of tricky. I think it depends on how onerous the merge conflict is? I think I'd probably prefer keeping the screens as one, since if we separate them out, we'd have to duplicate the multichain changes from my PR in your new screen, like you said. Alternatively, if there's a way to refactor stuff so that we minimize duplication, that could be nice, but that may not be possible/end up being even more clutter than having two separate screens.

cajubelt (approved):
looks good, remember the last AC on the issue which can be done in a separate PR:

> If [ACT-943](https://linear.app/valora/issue/ACT-943/error-states-for-enter-amount-screen) completed already, update Review button on Send Amount screen to navigate to this page.

finnian0826:
> looks good, remember the last AC on the issue which can be done in a separate PR:
> 
> > If [ACT-943](https://linear.app/valora/issue/ACT-943/error-states-for-enter-amount-screen) completed already, update Review button on Send Amount screen to navigate to this page.

@cajubelt it looks like that is [already taken care](https://github.com/valora-inc/wallet/blob/main/src/send/SendEnterAmount.tsx#L131) of since all the changes were done to the original screen, so should be good to go

PR #4320, opened 2023-10-16
feat(viem): switch WalletConnect request handler to Viem

bakoushin:
### Description

As a step towards multichain, this PR proposes switching WalletConnect requests handling to [Viem](https://viem.sh/).

Since Viem is a new library not carrying over some legacy features, we have to handle a few deprecations:

#### 1. Celo legacy transaction type

Viem doesn't support the serialization of Celo legacy transactions. It can serialize only the CIP-42 and later CIP-64 (once it lands in the Viem codebase). Those transactions are sent by several dApps in the ecosystem, namely impactMarket, Moola Market and HaloFi ([details](https://www.notion.so/valora-inc/dApps-transaction-types-sent-over-WalletConnect-fc502944ca704201adcc3adc850decce)).

To keep some extent of backward compatibility, legacy transaction requests will be automatically (and silently) upgraded to CIP-42 by omitting the `gasPrice` parameter in favor of auto-filled `maxFeePerGas`.

#### 2. `personal_decrypt` action
* Not supported by Viem
* Not listed in the [WalletConnect docs](https://docs.walletconnect.com/advanced/rpc-reference/ethereum-rpc#eth_sendtransaction)
* Never used in the last 12 months according to [Mixpanel data](https://mixpanel.com/project/2786968/view/3321744/app/insights#g48do5cCn4cJ)

### Misc

It was found that `gasLimit` can be used in transaction requests as a misnomer for `gas` (e.g. in [our tests](https://github.com/valora-inc/wallet/blob/e38c03588b7beb744b00dbc301006f3b0451200c/e2e/src/usecases/WalletConnectV2.js#L49)). To handle this, during request normalization `gasLimit` value is assigned to `gas` if the latter is absent.

### Test plan

Updated unit tests and e2e tests.

Tested manually on [Mento dApp](https://app.mento.org/) on Alfajores.

### Related issues

- Fixes [RET-870](https://linear.app/valora/issue/RET-870/[wallet]-switch-walletconnect-flows-to-use-viem)

### TODO

- RET-900: WalletConnect demo dApp currently can't deserialize CIP-42 transactions and must be upgraded
- RET-905: inject `feeCurrency` if it is not provided in a transaction request

### Backwards compatibility

Y

Potentially breaking `eth_signTransaction` and `eth_sendTransaction` Viem paths are kept under the feature flag for now to not break current dApps experience due to missing `feeCurrency` injection.

The `personal_decrypt` action was not used in the last 12 months.

### Feature flag

The `eth_signTransaction` and `eth_sendTransaction` logic is kept intact unless a feature flag `use_viem_for_walletconnect_transactions` is enabled to ensure backward compatibility until the `feeCurrency` injection is in place (see RET-905).


src/viem/getLockableWallet.test.ts
jeanregisser, line 1: Great! 👍

test/values.ts
```diff
@@ -422,6 +423,13 @@ export const mockWallet: UnlockableWallet = {
   computeSharedSecret: jest.fn(),
 }
 
+export const mockViemWallet: Partial<ViemWallet> = {
+  prepareTransactionRequest: jest.fn(),
+  signTransaction: jest.fn(),
+  signTypedData: jest.fn(),
+  signMessage: jest.fn(),
+}
```
jeanregisser, line 431: Looks like we should consolidate this with `__mocks/src/web3__` to avoid confusion?
i.e. having one default mockViemWallet, no?

src/walletConnect/saga.ts
```diff
@@ -531,6 +531,13 @@ function* handleAcceptRequest({ request }: AcceptRequest) {
       ...defaultTrackedProperties,
       error: e.message,
     })
+
+    // Notify the client about the error if possible
+    const { topic, id } = request
+    if (client && topic && id) {
+      const response = formatJsonRpcResult(id, { error: e })
+      yield* call([client, 'respondSessionRequest'], { topic, response })
+    }
```
jeanregisser, line 540: Great!  
Could we wrap this in a `try / catch` otherwise if an exception is thrown `handlePendingStateOrNavigateBack` will never be called.

src/walletConnect/request.ts
```diff
@@ -109,12 +187,35 @@ export function* handleRequest({ method, params }: { method: string; params: any
...
+    case SupportedActions.eth_signTypedData:
+      return (yield* call([wallet, 'signTypedData'], JSON.parse(params[1]))) as string
+    case SupportedActions.personal_sign: {
+      const data = { message: { raw: params[0] } } as SignMessageParameters
+      return (yield* call([wallet, 'signMessage'], data)) as string
+    }
+    case SupportedActions.eth_sign: {
+      const data = { message: { raw: params[1] } } as SignMessageParameters
+      return (yield* call([wallet, 'signMessage'], data)) as string
+    }
```
jeanregisser, line 200: Could we also wrap these in `useViem`?
I'd prefer it controlled all WC actions, this way we keep the old implementation in case something goes wrong.

src/viem/getLockableWallet.ts
```diff
@@ -51,6 +64,29 @@ export default function getLockableViemWallet(
       // For instance we will later need signTransaction which we can add here by
       // importing the signTransaction action and blocking it with the checkLock function
       // Introduction to wallet actions: https://viem.sh/docs/actions/wallet/introduction.html
+      prepareTransactionRequest: (args) => {
+        checkLock()
+        return prepareTransactionRequest(
+          client,
+          args
+        ) as Promise<PrepareTransactionRequestReturnType>
+      },
```
jeanregisser, line 73: Are we using this one in this PR?
Since it doesn't yet correctly work for Celo, I'd prefer we don't include it for now and leave a note about it?

src/walletConnect/request.ts
```diff
@@ -4,35 +4,84 @@ import { ContractKit } from '@celo/contractkit'
...
+          tx = rawTx
+        } else {
+          tx = yield* call(normalizeTransaction, wallet, rawTx)
+        }
+
+        // Convert hex values to numeric ones for Viem
+        // TODO: remove once Viem allows hex values as quanitites
+        const formattedTx: any = yield* call(formatTransaction, tx)
+
+        // TODO: inject fee currency for Celo when it's ready
```
jeanregisser, line 61: Yesss! You can follow https://github.com/valora-inc/wallet/pull/4378 too about this.

src/walletConnect/request.ts
```diff
@@ -4,35 +4,84 @@ import { ContractKit } from '@celo/contractkit'
...
+
+        const txCountParams: GetTransactionCountParameters = {
+          address: account as Address,
+          blockTag: 'pending',
+        }
+        const nonce = yield* call(getTransactionCount, wallet, txCountParams)
```
jeanregisser, line 67: nit: only do this if the `nonce` isn't already passed?

e2e/src/usecases/WalletConnectV2.js
```diff
@@ -93,283 +109,312 @@ export default WalletConnect = () => {
...
-        chainId: 'eip155:44787',
-        request: {
-          method: 'personal_sign',
-          params,
-        },
-      })
-      .then((result) => {
-        signature = result
+      it('Then is able to establish a session', async () => {
+        // TODO: restore initial starting logic after feeCurrecny estimation is ready
```
jeanregisser, line 185:
```suggestion
        // TODO: restore initial starting logic after feeCurrency estimation is ready
```

e2e/src/usecases/WalletConnectV2.js
```diff
@@ -93,283 +109,312 @@ export default WalletConnect = () => {
...
+          .then((result) => {
+            signedTx = result
+          })
+
+        await waitFor(element(by.text(`${dappName} would like to sign a Celo transaction.`)))
+          .toBeVisible()
+          .withTimeout(15 * 1000)
+        await verifySuccessfulTransaction('Sign transaction', tx)
+
+        // TODO: keep only Viem branch after feeCurrecny estimation is ready
```
jeanregisser, line 263:
```suggestion
        // TODO: keep only Viem branch after feeCurrency estimation is ready
```

e2e/src/usecases/WalletConnectV2.js
```diff
@@ -93,283 +109,312 @@ export default WalletConnect = () => {
...
+        if (web3Library === 'viem') {
+          // TODO: assert transaction signer address if once Viem could provide it
```
jeanregisser, line 265:
```suggestion
          // TODO: assert transaction signer address once Viem could provide it
```

e2e/src/usecases/WalletConnectV2.js
```diff
@@ -93,283 +109,312 @@ export default WalletConnect = () => {
...
+          const recoveredTx = parseTransactionCelo(signedTx)
+          jestExpect(recoveredTx.nonce).toEqual(hexToNumber(tx.nonce))
+          jestExpect(recoveredTx.to).toEqual(tx.to)
+          jestExpect(recoveredTx.gas).toEqual(BigInt(tx.gas))
+          jestExpect(recoveredTx.value).toEqual(BigInt(tx.value))
+        } else {
+          const [recoveredTx, recoveredSigner] = recoverTransaction(signedTx)
```
jeanregisser, line 272: `recoverTransaction` only works with legacy transactions? Or would it help to update `@celo/...` libs?

jeanregisser (commented):
OMG, epic work! 🚀 
Super happy to see this.

Small things here and there, let me know.

src/walletConnect/request.test.ts
```diff
@@ -106,33 +120,101 @@ describe(handleRequest, () => {
...
+            },
+          ],
+        })
+          .provide([
+            [matchers.call.fn(getViemWallet), mockViemWallet],
+            [matchers.call.fn(getTransactionCount), mockTransactionCount],
+            [matchers.call.fn(estimateFeesPerGas), mockEstimateFeePerGas],
+          ])
+          .withState(state)
+          .call(formatTransaction, {
```
kathaypacific, line 203: is it possible to asset that the mockViemWallet signTransaction method was called with these fields instead? not sure if it's possible with the big ints :/ but it would be nice since that's the end result (the current test could still pass if we incorrectly changed parts of the function after `formatTransaction` was called correctly)

src/walletConnect/request.ts
```diff
@@ -4,35 +4,84 @@ import { ContractKit } from '@celo/contractkit'
...
+          formattedTx.feeCurrency
+        )
+        const txRequest = {
+          ...formattedTx,
+          ...(formattedTx.nonce === undefined ? { nonce } : {}),
+          ...(formattedTx.maxFeePerGas === undefined ? { maxFeePerGas } : {}),
+          ...(formattedTx.maxPriorityFeePerGas === undefined ? { maxPriorityFeePerGas } : {}),
+        }
+
+        return (yield* call([wallet, 'signTransaction'], txRequest)) as string
```
kathaypacific, line 80: the logic in this function looks good to me, although i wonder if there's a way to avoid all the `any`'s? it's an annoying question but i couldn't help but think that anyone touching this has to be really confident in the changes being made, because there's not much that typescript is going to do to help you here 😅 maybe it's unavoidable or would add too much complexity, i didn't think this suggestion through too deeply but just wanted to make a note 😅

kathaypacific (commented):
looks great!!!! i don't have much feedback here, just similar thoughts as Jean that we should wrap useViem in all the WC methods to be able to have full control over the viem rollout :)

src/walletConnect/request.test.ts
bakoushin: I've planned this test block with a limited scope, just to ensure that normalization works correctly when it should.

    Viems's `signTransaction()` must only be called with `bigint` values. Calling it with string hex values will result in an error. That's why currently `formatTransaction` must precede any further action with transaction request. Ideally, Viem should handle formatting itself under the hood.
    
    I'm doubtful that checking the whole code of `sign_transaction` action end to end can add any value, because most of the functions inside are mocked -- they are from Viem or other helpers. The only place where we are executing our own meaningful code is normalization, which is covered here.

    I think the real e2e test are better suited to check that we still can sign a valid transaction.

    What do you think?

e2e/src/usecases/WalletConnectV2.js
bakoushin: Yes, if we update Celo libs, it possibly should work with both legacy and CIP-42.

But the update is not looking straightforward, I think it's better to make a separate PR for that.

src/walletConnect/request.test.ts
kathaypacific: yes i agree with all of that :)

src/walletConnect/request.ts
bakoushin: I wish I knew that way 😭

Unfortunately I can't figure out a proper type orchestration for Viem+Celo. It's too smart for me.

bakoushin: The problem seems to be global: https://valora-app.slack.com/archives/C029Z1QMD7B/p1698657194032799

My proposal it to keep `any`s for now and refactor once we find the solution for proper typing. Probably along with [RET-905](https://linear.app/valora/issue/RET-905/[walletconnect]-add-feecurrency-injection-back)

src/walletConnect/request.ts
```diff
@@ -4,33 +4,134 @@ import { ContractKit } from '@celo/contractkit'
...
+        if (rawTx.__skip_normalization) {
+          delete rawTx.__skip_normalization
+          tx = rawTx
+        } else {
+          tx = yield* call(normalizeTransaction, wallet, rawTx)
+        }
+
+        // Convert hex values to numeric ones for Viem
+        // TODO: remove once Viem allows hex values as quanitites
+        const formattedTx: any = yield* call(formatTransaction, tx)
```
kathaypacific, line 60: can the bulk of this logic be shared between sendTransaaction and signTransaction? (the only difference i can see, aside from the obvious wallet method that's called at the very end, is that we don't respect `__skip_normalization` for sendTransaction) is it intentional that the logic will diverge between these 2 methods at some point so we haven't abstracted the logic for formatting the tx?

kathaypacific (approved):
looks good to me ✨

src/walletConnect/request.ts
bakoushin: probably we don't need `__skip_normalization` at all, wdyt @jeanregisser? why someone could want to force push forward a wrong transaction?

e2e/src/usecases/WalletConnectV2.js
jeanregisser: oh yes, a separate PR is definitely warranted :D

__mocks__/src/web3/contracts.ts
```diff
@@ -45,6 +45,9 @@ export async function getWeb3Async() {
 
 const mockViemWallet = {
   account: { address: '0x1234' },
+  signTransaction: jest.fn().mockReturnValue('0x123456789'),
+  signTypedData: jest.fn().mockReturnValue('0x123456789'),
+  signMessage: jest.fn().mockReturnValue('0x123456789'),
```
jeanregisser, line 50: Should we add `sendTransaction` and `sendRawTransaction` here too?

src/viem/getLockableWallet.test.ts
jeanregisser, line 1: Could we also cover `sendRawTransaction` here?
I added it and skipped the tests because I knew you already had similar tests in this PR.
Can be done in a separate PR if you prefer.

src/walletConnect/request.ts
jeanregisser: Yes sounds good to remove it 👍

jeanregisser (approved):
🚀

src/viem/getLockableWallet.test.ts
bakoushin: It depends 😅 

The tests here are checking if the wallet lock works correctly.
Although, [sendRawTransaction()](https://github.com/valora-inc/wallet/blob/4f832a9370c71c2a85f27c2a639a4cd2f62ba323/src/viem/getLockableWallet.ts#L80-L82) is not currnetly checking the lock. Is that intended? If not, I can fix it & add to tests as well. Otherwise, it should't be covered here (and apparently don't have to be extended at all).

src/walletConnect/request.ts
bakoushin: great, I've abstracted all shared logic about preparing tx request into a separate function

src/viem/getLockableWallet.test.ts
jeanregisser: It's intentional, it doesn't do anything that needs an unlocked account.

PR #4322, opened 2023-10-16
feat(swap): new fee currency selection logic

jeanregisser:
### Description

This implements the new proposed [fee selection logic](https://www.notion.so/valora-inc/Fee-currency-selection-logic-4c207244893748bd85e23b754334f42d). 

Essentially, it creates the fully determined TXs (with gas, maxFeePerGas, etc) to use for swapping once we have a quote from our providers.

It shows warnings when TXs can't be created. See the screenshots below.

The goal is to cover all edge cases and allow the user to see a successful swap in all possible scenarios. No matter what `feeCurrency` the user has.

There will be PRs on top to use the created TXs for the rest of the flow, until sending once the user approves.

### Test plan

- Added unit tests

- When the user tries to send the full balance (or close to it) for a `feeCurrency`, and has no other `feeCurrency` to cover the gas cost:

This is a rough UI proposal, we can/should tweak the wording and design.

- When the user has no `feeCurrency` to cover for the gas cost

Again this is a rough UI proposal, we can/should tweak the wording and design.

### Related issues

- Fixes RET-881

### Backwards compatibility

Yes (under a new feature gate).


src/swap/useSwapQuote.ts
```diff
@@ -85,10 +170,81 @@ const useSwapQuote = () => {
...
+      }
+
+      // We can decrease the swap from amount to pay for gas,
+      // We'll ask the user if they want to proceed
+      return {
+        ...baseQuoteResult,
+        type: 'need-decrease-swap-amount-for-gas',
+        maxGasCost: result[1],
+        feeCurrency: result[0],
+      } satisfies QuoteResultNeedDecreaseSwapAmountForGas
```
jeanregisser, line 245: This is the main logic implemented.

Note: I haven't tried to break it down into smaller pieces. Although it's tailored for swaps, the logic should be able to be refactored to be more generic.

src/swap/SwapScreen.tsx
```diff
@@ -203,6 +210,36 @@ export function SwapScreen({ route }: Props) {
...
+      }
+
+      const quoteResultType = exchangeRate.type
+      switch (quoteResultType) {
+        case 'need-decrease-swap-amount-for-gas': // fallthrough on purpose
+        case 'not-enough-balance-for-gas':
+          quoteResultReviewBottomSheetRef.current?.snapToIndex(0)
+          break
+        case 'possible':
+          // TODO: show review screen with the possible transactions
```
jeanregisser, line 232: The review screen will use the fully determined TXs to show the maximum gas cost expected. 

And then once the user approves, they would simply be signed and sent to the network, no other logic needed related to fees.

src/swap/useSwapQuote.ts
```diff
@@ -85,10 +170,81 @@ const useSwapQuote = () => {
...
+          ...baseSwapTx,
+          maxFeePerGas: gasPrice,
+          feeCurrency: !feeCurrency.isNative ? (feeCurrency.address as Address) : undefined,
+        }
+
+        const maxGasCost = getMaxGasCost([approveTx, swapTx])
+        const maxGasCostInDecimal = new BigNumber(maxGasCost.toString()).shiftedBy(
+          -feeCurrency.decimals
+        )
+        maxGasCosts.push([feeCurrency, maxGasCostInDecimal])
```
dievazqu, line 200: nit: What do you think about doing `maxGasCosts.push({feeCurrency, maxGasCostInDecimal})` instead?
So then, when we use these values we do something like `result.feeCurrency` instead of `result[0]`?

jeanregisser: Yep much better 👍

src/swap/useSwapQuote.ts
```diff
@@ -85,10 +170,81 @@ const useSwapQuote = () => {
...
+        // TODO: this is Celo specific, make it generic
+        const gasPrice = BigInt((await getGasPrice(feeCurrency.address ?? undefined)).toFixed())
+
+        const approveTx = {
+          ...baseApproveTx,
+          maxFeePerGas: gasPrice,
+          feeCurrency: !feeCurrency.isNative ? (feeCurrency.address as Address) : undefined,
+          // We could estimate this ourselves,
+          // but for now we'll just use the value from the quote
+          gas: BigInt(quote.approveTransaction.gas),
```
kathaypacific, line 187: can we get this another way? i thought that we would be removing this `approveTransaction` in the work Jacob is doing on the backend...

src/swap/useSwapQuote.ts
```diff
@@ -85,10 +170,81 @@ const useSwapQuote = () => {
           : null,
       }
 
-      return updatedExchangeRate
+      const baseApproveTx = createApproveTransaction(updatedField, quote.unvalidatedSwapTransaction)
```
kathaypacific, line 173:
```suggestion
      const baseApproveTx = createBaseApproveTransaction(updatedField, quote.unvalidatedSwapTransaction)
```

kathaypacific (commented):
looking good, makes sense 👨🏼‍🎨

src/fees/hooks.ts
```diff
@@ -104,3 +105,35 @@ export function useMaxSendAmountByAddress(
...
   return useMaxSendAmount(tokenInfo?.tokenId, feeType, shouldRefresh)
 }
+
+/**
+ *
+ */
+export function useFeeCurrencies(networkId: NetworkId) {
+  const networkTokens = useSelector((state) => tokensListSelector(state, [networkId]))
+
+  const result = networkTokens
```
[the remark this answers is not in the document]
jeanregisser: Yep I was hesitating between useMemo or a memoized selector.

src/swap/useSwapQuote.ts
```diff
@@ -85,10 +170,81 @@ const useSwapQuote = () => {
...
+      const baseSwapTx = createSwapTransaction(quote.unvalidatedSwapTransaction)
+
+      let maxGasCosts: Array<[TokenBalance, BigNumber]> = []
+      for (const feeCurrency of feeCurrencies) {
+        // TODO: this is Celo specific, make it generic
+        const gasPrice = BigInt((await getGasPrice(feeCurrency.address ?? undefined)).toFixed())
+
+        const approveTx = {
+          ...baseApproveTx,
+          maxFeePerGas: gasPrice,
```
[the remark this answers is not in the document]
jeanregisser: Yes I'm a bit unsure, viem implementation makes you think it's needed. I'll check again with Aaron.

src/swap/useSwapQuote.ts
```diff
@@ -18,10 +23,90 @@ interface ExchangeRate {
...
+  | QuoteResultNeedDecreaseSwapAmountForGas
+  | QuoteResultNotEnoughBalanceForGas
+
+function createApproveTransaction(
+  updatedField: Field,
+  unvalidatedSwapTransaction: SwapTransaction
+) {
+  const { guaranteedPrice, sellTokenAddress, buyAmount, sellAmount, allowanceTarget, from } =
+    unvalidatedSwapTransaction
+  const amountType: string =
```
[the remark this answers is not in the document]
jeanregisser: Yes it's turned off at the UI level, but the rest of the code supports it and I didn't want to diverge from it.
Should we still remove it? Probably should be done separately.

src/swap/useSwapQuote.ts
jeanregisser: Yes I'll check with him.

src/swap/QuoteResultReviewBottomSheet.tsx
```diff
@@ -0,0 +1,133 @@
...
+      )
+    case 'not-enough-balance-for-gas':
+      return (
+        <QuoteResultNotEnoughBalanceForGasBottomSheet forwardedRef={forwardedRef} quote={quote} />
+      )
+    case 'possible':
+      return null
+    default:
+      // To catch any missing cases at compile time
+      const assertNever: never = quoteType
```
jh2oman, line 43: neat!

src/swap/useSwapQuote.ts
jh2oman: can confirm, going away :)

src/swap/useSwapQuote.ts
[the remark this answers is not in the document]
jeanregisser: Yes checked and added it.

src/swap/useSwapQuote.ts
```diff
@@ -1,27 +1,238 @@
...
+    updatedField === Field.TO ? ('buyAmount' as const) : ('sellAmount' as const)
+
+  const amountToApprove =
+    amountType === 'buyAmount'
+      ? BigInt(new BigNumber(buyAmount).times(guaranteedPrice).toFixed(0, 0))
+      : BigInt(sellAmount)
```
jeanregisser, line 68: I copied what we already had in the saga, but we said we should use the amount that was entered for the approval.
To mitigate the security risk.
I'll change this separately. cc @dievazqu

src/swap/useSwapQuote.ts
```diff
@@ -70,25 +282,38 @@ const useSwapQuote = () => {
       const swapPrice = useGuaranteedPrice
         ? quote.unvalidatedSwapTransaction.guaranteedPrice
         : quote.unvalidatedSwapTransaction.price
+      const price =
```
cajubelt, line 285: it is a little unnerving for me to see `swapPrice` and `price`, which are named very similarly, have different units if updatedField is not FROM

src/swap/useSwapQuote.ts
```diff
@@ -1,27 +1,238 @@
...
+  } satisfies PreparedTransactionsNeedDecreaseSwapAmountForGas
 }
 
 const useSwapQuote = () => {
   const walletAddress = useSelector(walletAddressSelector)
   const useGuaranteedPrice = useSelector(guaranteedSwapPriceEnabledSelector)
-  const [exchangeRate, setExchangeRate] = useState<ExchangeRate | null>(null)
+  const [exchangeRate, setExchangeRate] = useState<QuoteResult | null>(null)
+
+  // TODO use the networkId from the fromToken
```
cajubelt, line 234: same questions as the other todo. In general, I worry that todo's are easily forgotten about

src/swap/useSwapQuote.ts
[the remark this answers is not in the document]
cajubelt: removing sounds good if it's not accessible by users, though it also sounds like you'd want to remove it completely rather than just in this function

src/swap/useSwapQuote.ts
```diff
@@ -1,27 +1,238 @@
...
+
+  const amountToApprove =
+    amountType === 'buyAmount'
+      ? BigInt(new BigNumber(buyAmount).times(guaranteedPrice).toFixed(0, 0))
+      : BigInt(sellAmount)
+
+  const data = encodeFunctionData({
+    abi: erc20.abi,
+    functionName: 'approve',
+    args: [allowanceTarget as Address, amountToApprove],
```
cajubelt, line 73: instead of casting to Address everywhere could we use `getAddress`?

src/swap/useSwapQuote.ts
```diff
@@ -1,27 +1,238 @@
...
+  return approveTx
+}
+
+function createBaseSwapTransaction(unvalidatedSwapTransaction: SwapTransaction) {
+  const { from, to, value, data, gas } = unvalidatedSwapTransaction
+
+  const swapTx: TransactionRequestCIP42 & { gas: bigint } = {
+    from: from as Address,
+    to: to as Address,
+    value: BigInt(value ? value : 0),
```
cajubelt, line 93:
```suggestion
    value: BigInt(value ?? 0),
```
code golf

src/swap/useSwapQuote.ts
```diff
@@ -1,27 +1,238 @@
...
+  type: 'not-enough-balance-for-gas'
+  feeCurrencies: TokenBalance[]
+}
+
+export type PreparedTransactionsResult =
+  | PreparedTransactionsPossible
+  | PreparedTransactionsNeedDecreaseSwapAmountForGas
+  | PreparedTransactionsNotEnoughBalanceForGas
+
+function createBaseApproveTransaction(
```
cajubelt, line 56: what makes something a "base" transaction?

src/swap/useSwapQuote.ts
```diff
@@ -1,27 +1,238 @@
...
+    to: sellTokenAddress as Address,
+    data,
+    type: 'cip42',
+    amountToApprove,
+  }
+
+  return approveTx
+}
+
+function createBaseSwapTransaction(unvalidatedSwapTransaction: SwapTransaction) {
```
cajubelt, line 87: what makes a transaction "unvalidated"?

src/swap/useSwapQuote.ts
```diff
@@ -1,27 +1,238 @@
...
+  const { from, to, value, data, gas } = unvalidatedSwapTransaction
+
+  const swapTx: TransactionRequestCIP42 & { gas: bigint } = {
+    from: from as Address,
+    to: to as Address,
+    value: BigInt(value ? value : 0),
+    data: data as Hex,
+    gas: BigInt(gas), // This isn't entirely accurate
```
cajubelt, line 95: > This isn't entirely accurate

what does that mean?

src/swap/useSwapQuote.ts
```diff
@@ -1,27 +1,238 @@
...
+    )
+
+    const approveTx = {
+      ...baseApproveTx,
+      maxFeePerGas,
+      maxPriorityFeePerGas,
+      feeCurrency: feeCurrencyAddress,
+    }
+
+    // TODO maybe cache this? and add static padding when using non-native fee currency
```
cajubelt, line 143: who's going to do this? is there a Linear issue tracking the work that can be linked to here? If not, how are we going to remember to do it?

cajubelt: isn't the static padding part done in line 170?

src/swap/useSwapQuote.ts
```diff
@@ -1,27 +1,238 @@
...
+    }
+
+    const swapTx = {
+      ...baseSwapTx,
+      maxFeePerGas,
+      maxPriorityFeePerGas,
+      feeCurrency: feeCurrencyAddress,
+      // We assume the provided gas value is with the native fee currency
+      // If it's not, we add the static padding
+      gas: !feeCurrency.isNative ? baseSwapTx.gas + BigInt(STATIC_GAS_PADDING) : baseSwapTx.gas,
```
cajubelt, line 170:
```suggestion
      gas: baseSwapTx.gas + BigInt(feeCurrency.isNative ? 0 : STATIC_GAS_PADDING),
```
code golf

src/swap/useSwapQuote.ts
```diff
@@ -1,27 +1,238 @@
...
+  for (const tx of txs) {
+    if (!tx.gas || !tx.maxFeePerGas) {
+      throw new Error('Missing gas or maxFeePerGas')
+    }
+    maxGasCost += BigInt(tx.gas) * BigInt(tx.maxFeePerGas)
+  }
+  return maxGasCost
+}
+
+async function prepareTransactions(
```
cajubelt, line 113: wdyt about adding an explicit return type to this function? to make it clearer what comes out, and potentially make type errors easier to debug if we make breaking changes to it by accident

src/swap/useSwapQuote.ts
```diff
@@ -1,27 +1,238 @@
...
+  updatedField: Field,
+  unvalidatedSwapTransaction: SwapTransaction,
+  price: string,
+  feeCurrencies: TokenBalance[]
+) {
+  const baseApproveTx = createBaseApproveTransaction(updatedField, unvalidatedSwapTransaction)
+  const baseSwapTx = createBaseSwapTransaction(unvalidatedSwapTransaction)
+
+  const maxGasCosts: Array<{ feeCurrency: TokenBalance; maxGasCostInDecimal: BigNumber }> = []
+  for (const feeCurrency of feeCurrencies) {
```
cajubelt, line 124: this function is pretty long and complex. could the stuff in the for loop be pulled out into a helper function to make it easier to read, and to write tests for?

src/swap/useSwapQuote.ts
```diff
@@ -70,25 +282,38 @@ const useSwapQuote = () => {
...
+        price,
+
         provider: quote.details.swapProvider,
         estimatedPriceImpact: estimatedPriceImpact
           ? new BigNumber(estimatedPriceImpact).dividedBy(100)
           : null,
       }
 
-      return updatedExchangeRate
+      if (shouldPrepareTransactions) {
```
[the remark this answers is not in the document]
jeanregisser: I agree with this, however here it's a temporary adjustment until we remove the feature flag.
This isn't meant to stay like this. I do think it's the right place to have it to avoid duplicating the hook.
Maybe not ideal, but the alternative I was seeing didn't make sense in this context.

src/swap/useSwapQuote.ts
jeanregisser: Yes, we can probably improve that later. Also caused by a feature flag.

src/swap/useSwapQuote.ts
jeanregisser: Yes fair enough, we have work queued up for multichain support. This will be changed in a follow up PR.

src/swap/useSwapQuote.ts
jeanregisser: Oh interesting, but I see it does a checksum which isn't needed here:
https://github.com/wagmi-dev/viem/blob/5c95fafceffe7f399b5b5ee32119e2d78a0c8acd/src/utils/address/getAddress.ts

Maybe we could introduce our own helper? We can do this in a separate PR.

src/swap/useSwapQuote.ts
jeanregisser: It's related to this: https://www.notion.so/valora-inc/Fee-currency-selection-logic-4c207244893748bd85e23b754334f42d?pvs=4#8b7c27d31ebf4fca981f81e9411f86ee

I can add the link as a comment.

src/swap/useSwapQuote.ts
jeanregisser: Yeah naming is hard. This was suggested by Kathy, just that's it's a partial transaction which doesn't have all fields we want here.
Do you have another suggestion?

src/swap/useSwapQuote.ts
jeanregisser: This is the naming used by our API, it means it hasn't been fully checked and may revert.

src/swap/useSwapQuote.ts
jeanregisser: > who's going to do this? is there a Linear issue tracking the work that can be linked to here? If not, how are we going to remember to do it?

I am, this PR is just the beginning of more changes I'm doing.

jeanregisser: > isn't the static padding part done in line 170?

Line 170 is for the swap TX, this is for the approval.

src/swap/useSwapQuote.ts
jeanregisser: Definitely agree. This code is at the "make it work" stage and I haven't tried to break it down differently yet. I know we probably want to make it more generic, but until we have other examples to better guide this, I preferred to wait and see.

src/swap/useSwapQuote.ts
```diff
@@ -1,27 +1,241 @@
...
+      if (e instanceof EstimateGasExecutionError) {
+        // Likely too much gas was needed
+        Logger.warn(
+          'SwapScreen@useSwapQuote',
+          `Couldn't estimate gas for approval with feeCurrency ${feeCurrency.symbol} (${feeCurrency.tokenId}), trying next feeCurrency`,
+          e
+        )
+        continue
+      }
+      throw e
```
kathaypacific, line 163: any reason why we want to throw here? i thought continuing to the next fee currency would be better than exiting from getting the quote entirely

kathaypacific (approved):
nice tests for the swap screen! this logic looks good to me, and makes sense that we want to wait before abstracting to make this logic more generic. if i were nitpicking i'd suggest moving all the helper functions out of the `useSwapQuote` file, since that was a little hard to navigate (some fees stuff live there and some inside the viem folder)

src/swap/useSwapQuote.ts
cajubelt: thanks

src/swap/useSwapQuote.ts
cajubelt: yeah I noticed the checksum...it does seem like that could trip us up since in Valora code we always assume lower cased addresses

src/swap/useSwapQuote.ts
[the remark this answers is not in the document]
cajubelt: ok, maybe it would be helpful to leave a comment to that effect?

cajubelt (approved):

src/swap/useSwapQuote.ts
jeanregisser: Yes this isn't supposed to happen. Trying the next fee currency isn't the appropriate action.

PR #4327, opened 2023-10-17
chore: remove legacy standby transactions

kathaypacific:
### Description

This PR removes `stableTokenTransferLegacySaga` which is spawned from the root saga. However, it waits for the `Actions.TRANSFER` action before executing any logic. This action is not fired anymore, so this PR removes this saga plus all the unused code paths for adding standby transactions. Also removes code paths relating to the unused CalculateFee component.

### Test plan

CI

### Related issues

Fixes RET-894

### Backwards compatibility

Y


src/tokens/saga.ts
```diff
@@ -140,63 +133,6 @@ export async function createTokenTransferTransaction(
...
-        timestamp: Math.floor(Date.now() / 1000),
-        address: recipientAddress,
-      })
-    )
-
-    try {
-      const account: string = yield* call(getConnectedUnlockedAccount)
-
-      const currencyAddress: string = yield* call(getCurrencyAddress, currency)
-      const tx = yield* call(createTokenTransferTransaction, currencyAddress, {
```
jh2oman, line 177: pull on the `createTokenTransferTransaction` string too! I was able to follow it all the way to this unused component that we can delete https://github.com/valora-inc/wallet/blob/main/src/fees/CalculateFee.tsx lots more dead code!

kathaypacific: TY, i didn't notice this! I have removed :D

jh2oman (approved):

PR #4335, opened 2023-10-18
feat(viem): add helpers to get fees per gas with a feeCurrency

jeanregisser:
### Description

This is needed for now to correctly craft CIP42 TXs with viem, which have correct `maxFeePerGas` / `maxPriorityFeePerGas` based on the passed `feeCurrency`.
See https://valora-app.slack.com/archives/C029Z1QMD7B/p1697640184208049

Extracted from https://github.com/valora-inc/wallet/pull/4322

### Test plan

- Added unit tests

### Related issues

N/A

### Backwards compatibility

Yes


src/viem/estimateFeesPerGas.ts
```diff
@@ -0,0 +1,31 @@
+import { Address, Client, hexToBigInt } from 'viem'
+
+// Custom function that can be removed once it's supported in viem
+// See https://github.com/wagmi-dev/viem/discussions/914
+export async function estimateFeesPerGas(client: Client, feeCurrency?: Address) {
+  const [gasPrice, maxPriorityFeePerGas] = await Promise.all([
+    getGasPrice(client, feeCurrency),
+    getMaxPriorityFeePerGas(client, feeCurrency),
+  ])
+  const maxFeePerGas = gasPrice + maxPriorityFeePerGas
```
dievazqu, line 10: I find this weird...
What does `maxFeePerGas` mean? What's the difference between `maxPriorityFeePerGas`?
I don't understand why adding the `gasPrice` works here.

jeanregisser: Yes I was confused by these myself at first.

The total maximum cost of a EIP-1559 TX (roughly equivalent to CIP-42) is `gas * maxFeePerGas`.
Previously it was `gas * gasPrice`, which older TX types.

`gasPrice` is the minimum required to get the transaction included in a block. It is set by the network and fluctuates from block to block.

`maxPriorityFeePerGas` is the part that will go to the validator to incentivize the TX to be included up.

See also: https://docs.alchemy.com/docs/maxpriorityfeepergas-vs-maxfeepergas

bakoushin: That's a very good question!

The `maxFeePerGas` and `maxPriorityFeePerGas` are the new parameters introduced in EIP-1559/CIP-42 to replace `gasPrice` for fee calculations. Basically the formula for the fee now is more complex.

I think this is a replica of internal Viem logic, but I wonder too, why it is relying on the `gasPrice` and not on the `baseFeePerGas` block property?

jeanregisser: > I think this is a replica of internal Viem logic, but I wonder too, why it is relying on the gasPrice and not on the baseFeePerGas block property?

Viem uses `baseFeePerGas` to calculate the final `maxFeePerGas`.

https://github.com/wagmi-dev/viem/blob/5c95fafceffe7f399b5b5ee32119e2d78a0c8acd/src/actions/public/estimateFeesPerGas.ts#L154-L156

But we can't do this on Celo, because it depends on the `feeCurrency`.
I don't know all the underlying details but using `eth_getGasPrice` is what is [recommended on Celo](https://docs.celo.org/developer/viem#gas-price).

bakoushin: Oh, I see. Thanks for bringing this out.

Other questions remain:
* Does this way of getting `maxFeePerGas` apply to the situations when we don't have `feeCurrency` provided? 
* Why we don't multiply this value in the same way as it is done with `baseFeePerGas`? Is it already well done under the hood?

But I think it's better to discuss it with Celo devs.

bakoushin: Also, in the Celo docs example they don't sum up the value they get from `eth_gasPrice` with `maxPriorityFeePerGas`.

jeanregisser: Ok @bakoushin and I had a interesting convo with cLabs and long story short for our goal here, to be able to display the max total cost, it's recommended to do `maxFeePerGas = gasPrice`

See details in this [thread](https://valora-app.slack.com/archives/CNJ7KTHQU/p1697717100995909?thread_ts=1697647756.662059&cid=CNJ7KTHQU) if you're interested. I've also updated the code to explain this.

src/viem/estimateFeesPerGas.test.ts
```diff
@@ -0,0 +1,27 @@
...
+        expect(params).toBeUndefined()
+        if (method === 'eth_gasPrice') return '0x64' // 100 in hex
+        throw new Error(`Unknown method ${method}`)
+      }),
+    }
+    const fees = await estimateFeesPerGas(client as any)
+    expect(fees).toEqual({ maxFeePerGas: BigInt(100), maxPriorityFeePerGas: undefined })
+  })
+
+  it('when fee currency is specified, it should return the correct fees per gas', async () => {
```
bakoushin, line 16: Nit about test name: I think it should return the correct fees per gas, regardless of fee currency presence.

bakoushin (approved):
Awesome!  🥷

Thanks for nailing it!

src/viem/estimateFeesPerGas.test.ts
jeanregisser: oh exactly, this was only to highlight the different usage with the extra parameter.
Do you have another proposition?

bakoushin: oh no, I just didn't noticed the `expect` in the middle of `client.request` mock!
now it makes sense!

PR #4342, opened 2023-10-19
refactor: make standby transaction type similar to token transactions

kathaypacific:
### Description

1. Update the schema of `standbyTransaction` to match `TokenTransaction`. The existing type is a flat object that contains properties that describes a token transfer (i.e. token sent or received) . If we do not update this type, we would need to add more optional properties to the flat object to describe swaps and other types of transfers we want to support in the future.
2. Add `status` to all transaction items. For now we will add the "confirmed" status to any transactions received from blockchain-api (based on how the standby transactions are resolved, this was the implicit assumption anyway). Transaction items will need a status to be displayed in the home feed and tx details screen soon.

### Test plan

There should be no change to how the app works.

### Related issues

- Fixes RET-894

### Backwards compatibility

Y


jeanregisser (approved):
Great! 🚀 
Makes it easier to capture all the needed context. 👍

PR #4343, opened 2023-10-19
feat(token-bottom-sheet): add param to use TokenBalanceItem

satish-ravi:
### Description

Updates the existing token bottom sheet component to take a TokenOptionComponent parameter to render either the legacy token option or the new token balance item. Also parameterizes titleStyle so we can use the new font types

### Test plan

Unit tests, manually 

| iOS | Android |
|--------|--------|
|  |  | 

### Related issues

- Fixes ACT-942

### Backwards compatibility

Yes


src/components/TokenBottomSheet.tsx
```diff
@@ -93,6 +96,8 @@ function TokenBottomSheet<T extends TokenBalance>({
   tokens,
   searchEnabled,
   title,
+  titleStyle,
+  useTokenBalanceItem,
```
cajubelt, line 100: it's not obvious to me what it actually means to use the token balance item component vs TokenOption. Is TokenBalanceItem preferable for new code? 

I just left a similar [comment](https://github.com/valora-inc/wallet/pull/4322#discussion_r1366180095) on Jean's PR here-- I think we should avoid using boolean parameters whenever we can. In this case, you could write a new version of TokenBottomSheet, and a legacy version that shares most of the code besides the TokenBalanceItem/TokenOption part.

cajubelt (approved):
🥳

PR #4349, opened 2023-10-20
feat: add UI for pending transactions in homefeed

kathaypacific:
### Description

This PR:
- tweak the homefeed items order so that standby tx's are at the top of the homefeed
- create common component to handle the tx feed item image (including pending and failed states), use this for transfers and swaps

### Test plan

(I hacked the send saga to navigate back to the homescreen early for this video, this change is not included in this PR)

https://github.com/valora-inc/wallet/assets/20150449/850b4438-7afc-4831-9ea5-a6cff74e451c

### Related issues

- Fixes RET-894

### Backwards compatibility

Y


src/transactions/feed/TransactionFeedItemImage.tsx
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  if (__typename === 'TokenTransferV3') {
+    return <ContactCircle recipient={props.recipient} size={AVATAR_SIZE} />
+  }
+
+  Logger.error(
+    'TransactionFeedItemImage',
+    `Could not render image for transaction for transaction type ${__typename} and status ${status}`
+  )
```
jeanregisser, line 43: Maybe only in dev mode? this could become very noisy fast.

src/transactions/feed/TransactionFeedItemImage.tsx
```diff
@@ -0,0 +1,47 @@
...
+type Props =
+  | {
+      status: TransactionStatus
+      __typename: 'TokenExchangeV3'
+    }
+  | {
+      status: TransactionStatus
+      __typename: 'TokenTransferV3'
+      recipient: Recipient
+    }
```
jeanregisser, line 20: I think it may be the first time we really use the `__typename` field.
Do we still wanna use `__typename` here for the prop name? It it gives the feeling it's a private thing that shouldn't be used.

Also don't we want NFTs here too?

src/transactions/feed/TransferFeedItem.tsx
```diff
@@ -41,10 +41,10 @@ function TransferFeedItem({ transfer }: Props) {
   return (
     <Touchable testID="TransferFeedItem" disabled={false} onPress={openTransferDetails}>
       <View style={styles.container}>
-        <ContactCircle
-          style={{ alignItems: 'flex-start' }}
+        <TransactionFeedItemImage
           recipient={recipient}
-          size={AVATAR_SIZE}
```
jeanregisser, line 47: Remove `AVATAR_SIZE` const above since now unused?

src/transactions/utils.ts
```diff
@@ -1,33 +1,44 @@
...
 // Sections are hidden if they have no items.
-export function groupFeedItemsInSections<T extends { timestamp: number }>(items: T[]) {
+export function groupFeedItemsInSections(
+  standbyTransactions: TokenTransaction[],
+  confirmedTransactions: TokenTransaction[]
+) {
   const sectionsMap: {
     [key: string]: {
-      data: T[]
+      data: TokenTransaction[]
```
jeanregisser, line 17: Just curious, ee can remove the paramterized type because we remove the legacy code?

jeanregisser (approved):
Awesome! 🚀

src/transactions/feed/TransactionFeedItemImage.tsx
kathaypacific: actually i think this scenario should never happen, every combination of the expected input props should be covered in the if statements above...i initially had just a `return null // this shouldn't happen!` but decided to log an error just in case it does. if you're not strongly opposed, i'd like to leave this here

src/transactions/feed/TransactionFeedItemImage.tsx
kathaypacific: omg yeah i felt uneasy about this too! the underscores are really awkward, i renamed the prop to `transactionType`

kathaypacific: i didn't put NFTs here for a few reasons...i'm lazy and because we don't yet have any paths where an NFT transaction could be pending or failed (since you can only receive) then the NFTs don't really benefit from this component. also there's a bit of logic about showing the NFT image (rendering the image vs fallback etc) which would pollute this component a bit

src/transactions/utils.ts
kathaypacific: yes i think so! :D

PR #4352, opened 2023-10-20
fix: rely on blockchain-api only for confirmed transactions

kathaypacific:
### Description

Problem: after sending, the pending transaction disappears for a few seconds before the new confirmed transaction appears. This is weird can causes the transaction feed to jump around.

Root cause: currently the sequence of events after sending is
1. the tx feed displays standby transactions in the "pending" state + the transactions in the `transactions` store
2. a pending tx is added to the store when the tx is submitted to the network
3. the `ADD_HASH_TO_STANDBY_TRANSACTIONS` action adds the tx hash to the pending tx in the store when we receive the tx receipt. Also at this time, a `TRANSACTION_CONFIRMED` action marks the pending transaction as "completed"
4. since we only display standby transactions in the "pending" state, this transaction disappears
5. it will reappear when the next poll to blockchain-api picks up this transaction and gives it to the wallet

I think this is a bit overly complex, and there becomes 2 sources of truth for which transactions are confirmed. I propose that we expect the blockchain-api to be the source of truth for confirmed transactions, and only when we get a transaction from there do we mark the pending transaction as confirmed. the tradeoff is that transactions will appear as "pending" for longer, since we poll blockchain-api every 10 seconds for the feed. but maintaining this elaborate logic seems really overkill for this 10 second optimisation that will be for the Celo network only...

### Test plan

Tested manually, but there should be no change to the user experience. Pending transactions will transition to being completed (maybe taking longer than before).

https://github.com/valora-inc/wallet/assets/20150449/81693e08-240a-4608-adb3-26385d054215

### Related issues

- Fixes RET-894

### Backwards compatibility

Y


bakoushin:
Thanks for handling it! Having a single source of truth is good! 📖 

But strangely enough, after switching to your branch and rebuilding the app the described behavior persisted (tx disappears and reappears).

kathaypacific:
> Thanks for handling it! Having a single source of truth is good! 📖
> 
> But strangely enough, after switching to your branch and rebuilding the app the described behavior persisted (tx disappears and reappears).

hey @bakoushin would you mind checking again? i went through the code and can't see why the disappear / reappear behaviour would persist, i also tested locally a few times and _it works on my machine_ 🙈

bakoushin:
@kathaypacific
damn! I mistakenly pulled the wrong branch that day 😖
the behavior is indeed as expected on the right branch.
so sorry about causing you trouble!

bakoushin (approved):

jeanregisser (changes requested):
Please no 😢 
Blockchain-api can be unreliable.

PR #4365, opened 2023-10-23
fix: swaps using CELO sometimes not working

kathaypacific:
### Description

When the swap quote comes from 0x, the buy token is "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee". We need to turn this into something we recognise to get the related token info.

This PR does a few things:
1. given the sell and buy tokens from the swap quote, get the token id's and use those to get the token info
2. map "0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee" to the native token. it'd be nice to avoid this in the wallet but we'll need similar logic for ETH.
3. use the sell and buy tokens directly in the swap saga (instead of the buy and sell token addresses from the swap quotes)
4. create the standby swap transaction before submitting the swap, and manually add the tx hash.
5. fix typo for variable name in reducer

### Test plan

(This quote wasn't from 0x, I tried many times but couldn't get one from 0x. So this shows the regression scenario)
https://github.com/valora-inc/wallet/assets/20150449/7d619789-3305-4757-9612-b1fd92051ce4

### Related issues

- Fixes RET-894

### Backwards compatibility

Y

### DIFF
--- a/src/swap/SwapReviewScreen.tsx
+++ b/src/swap/SwapReviewScreen.tsx
@@ -13,9 +13,10 @@ import { maxSwapSlippagePercentageSelector } from 'src/app/selectors'
 import BackButton from 'src/components/BackButton'
 import Button, { BtnSizes } from 'src/components/Button'
 import Dialog from 'src/components/Dialog'
-import CustomHeader from 'src/components/header/CustomHeader'
 import LegacyTokenDisplay from 'src/components/LegacyTokenDisplay'
+import { formatValueToDisplay } from 'src/components/TokenDisplay'
 import Touchable from 'src/components/Touchable'
+import CustomHeader from 'src/components/header/CustomHeader'
 import { useFeeCurrency } from 'src/fees/hooks'
 import InfoIcon from 'src/icons/InfoIcon'
 import { noHeader } from 'src/navigator/Headers'
@@ -28,11 +29,10 @@ import { swapUserInputSelector } from 'src/swap/selectors'
 import { swapStart } from 'src/swap/slice'
 import { FetchQuoteResponse, Field } from 'src/swap/types'
 import { celoAddressSelector, tokensByAddressSelector } from 'src/tokens/selectors'
-import { divideByWei } from 'src/utils/formatting'
 import Logger from 'src/utils/Logger'
+import { divideByWei } from 'src/utils/formatting'
 import networkConfig from 'src/web3/networkConfig'
 import { walletAddressSelector } from 'src/web3/selectors'
-import { formatValueToDisplay } from 'src/components/TokenDisplay'
 
 const TAG = 'SWAP_REVIEW_SCREEN'
 const initialUserInput = {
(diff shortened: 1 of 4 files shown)


src/swap/SwapReviewScreen.tsx
```diff
@@ -129,7 +121,7 @@ export function SwapReviewScreen() {
     async () => {
       if (!shouldFetch) return
       const params = {
-        buyToken: toCeloWorkaround(toToken),
+        buyToken: toToken,
```
kathaypacific, line 124: this is handled on the backend already

src/swap/saga.ts
```diff
@@ -90,9 +114,23 @@ export function* swapSubmitSaga(action: PayloadAction<SwapInfo>) {
...
+        sellTokenAddress === nativeTokenAddressForSwapProviders ? undefined : sellTokenAddress
+      )
+    ]
+  const toToken =
+    tokens[
+      getTokenId(
+        networkConfig.defaultNetworkId,
+        buyTokenAddress === nativeTokenAddressForSwapProviders ? undefined : buyTokenAddress
+      )
+    ]
```
jeanregisser, line 133: Could/should we use `action.payload.userInput.{from|to}Token` to avoid having to know about `0xeeeee...` here?

jeanregisser (approved):
Great!

src/swap/saga.ts
kathaypacific: actually this is a great question - i wondered the same, but i started wondering if it is a risk? e.g. if the token addresses are incorrect in the response, then we potentially execute something different to what we display? maybe this is not something we should worry about?

jeanregisser: I'm asking myself the same question for the viem flow.

dievazqu: I think we should rely on the backend response.
Using the UI data as source of truth sounds risky.

jeanregisser: I guess since the real thing happening is with the `data` field from the transaction, the risk of executing something different that what is displayed is there no matter if we use the user input or the `sellTokenAddress` or `buyTokenAddress`, or not.

jh2oman: I actually disagree with using the (swaps) backend as a source of truth here. The swaps cloud functions are centered around the providers and 0x and openocean use different addresses for celo native tokens.

kathaypacific: after thinking about it some more, i tend to agree that we should trust what comes from the backend. it simplifies this swaps logic but also more broadly it is a rabbit hole to start doubting the responses that our services provide... we should indeed validate the `data` field if we want to deal with the security topic but that would require our swaps providers to verify their contracts (i have requested this from the 0x team but didn't get a response :( ) since the buy and sell token addresses from the response here does not affect the outcome of the swap, i'll make the change to not read them

PR #4366, opened 2023-10-23
fix: re-enable sagas that were meant to be triggered on new tx feeds

kathaypacific:
### Description

Context: https://valora-app.slack.com/archives/C025V1D6F3J/p1697617792013689

This PR re-enables the sagas that were incorrectly no longer dispatched on the `NEW_TRANSACTIONS_IN_FEED` action:
- remove the old `NEW_TRANSACTIONS_IN_FEED` and replace with `UPDATE_TRANSACTIONS`
- the sagas triggered by `NEW_TRANSACTIONS_IN_FEED` were updated to handle the new Transaction type
- some unused apollo types were removed

### Test plan

Since I didn't change the content of the sagas that are meant to be triggered, I did not test the logic in the sagas. I only tested that the homefeed can display new transactions from blockchain-api as normal.

### Related issues

N/A

### Backwards compatibility

Y


jeanregisser (approved):
Thanks! 👍

PR #4367, opened 2023-10-23
chore(ios): update config for Sonoma / xcode 15

jh2oman:
### Description

iOS wasn't building for me on Xcode 15 on Sonoma until I made these changes based on comments in this [xcode support thread](https://developer.apple.com/forums/thread/730974)

### Test plan

Make sure it works for Xcode 14 as well


jeanregisser (approved):
Thanks! 🙌

MuckT (approved):

PR #4368, opened 2023-10-23
feat: use viem for the swap flow

jeanregisser:
### Description

This updates the full swap flow to use viem when `USE_VIEM_FOR_SWAP` is enabled.

The updated swap quote with the prepared transaction(s) is now passed to the review screen.
Pull-to-refresh is disabled in that case (but I guess most users didn't know they could do that), and that screen is going away soon anyway.

Some things aren't entirely ideal (like having the `unvalidatedSwapTransaction` included in the quote), but it made fitting into the existing review / confirmation / saga code easier.

I'll make another PR to update the analytics to track whether we use viem or contractkit, and add props to measure reverts due to hitting the gas limit so we can better adjust it. See https://github.com/valora-inc/valora-rest-api/pull/697

### Test plan

- Updated tests
- Manually tested swaps in different configurations

### Related issues

- Fixes RET-881

### Backwards compatibility

Yes


src/swap/useSwapQuote.ts
jeanregisser, line 1: You can mostly ignore the refactored which happened here.
Eventually this PR will be rebased on top of https://github.com/valora-inc/wallet/pull/4378

src/swap/SwapReviewScreen.tsx
jeanregisser, line 1: This screen is going away. So I didn't bother adding tests here.

src/swap/saga.ts
```diff
@@ -247,8 +250,182 @@ export function* swapSubmitSaga(action: PayloadAction<SwapInfo>) {
...
+
+    const beforeSwapExecutionTimestamp = Date.now()
+    quoteToTransactionElapsedTimeInMs = beforeSwapExecutionTimestamp - quoteReceivedAt
+
+    const txHashes: Hash[] = []
+    for (const preparedTransaction of preparedTransactions.transactions) {
+      const signedTx = yield* call([wallet, 'signTransaction'], {
+        ...preparedTransaction,
+        nonce: nonce++,
+      } as any)
```
jeanregisser, line 362: Casting here is also because of `typed-redux-saga` erasing parameterized types and causing errors.
I have local branch where I tried to solve this, I'll try to finish it.

src/swap/useSwapQuote.ts
```diff
@@ -67,42 +74,132 @@ function createBaseApproveTransaction(
...
+  const transactions: TransactionRequestCIP42[] = []
+
+  const feeCurrencyAddress = getFeeCurrencyAddress(feeCurrency)
+  const { maxFeePerGas, maxPriorityFeePerGas } = await estimateFeesPerGas(
+    publicClient.celo,
+    feeCurrencyAddress
+  )
+
+  for (const [index, baseTx] of baseTransactions.entries()) {
+    // We can only truly estimate the gas for the first transaction
```
cajubelt, line 129: seems like there are cases this might not be true, like if we wanted to use this for estimating gas for registering the DEK and doing a c-stable transfer

src/swap/SwapScreen.tsx
```diff
@@ -210,6 +210,16 @@ export function SwapScreen({ route }: Props) {
   const handleReview = () => {
     ValoraAnalytics.track(SwapEvents.swap_screen_review_swap)
 
+    const userInput = {
+      toToken: toToken!.address,
+      fromToken: fromToken!.address,
```
cajubelt, line 215: is there a way we could avoid the non-null assertion operators here? Even if it's just throwing a descriptive error message if either is null, instead of a type error, to make our lives a bit easier if we have to debug it?

src/swap/SwapScreen.tsx
```diff
@@ -230,7 +240,12 @@ export function SwapScreen({ route }: Props) {
           preparedTransactionsReviewBottomSheetRef.current?.snapToIndex(0)
           break
         case 'possible':
-          // TODO: show review screen with the possible transactions
+          // TODO: we want to remove the need to use redux, but for now keeping it
```
cajubelt, line 243: when's it going to get removed?

src/swap/saga.ts
```diff
@@ -247,8 +250,182 @@ export function* swapSubmitSaga(action: PayloadAction<SwapInfo>) {
...
+
+    const receipt = yield* call([publicClient.celo, 'waitForTransactionReceipt'], {
+      hash: swapTxHash,
+    })
+    Logger.debug('Got swap transaction receipt', receipt)
+    if (receipt.status !== 'success') {
+      throw new Error(`Swap transaction reverted: ${receipt.transactionHash}`)
+    }
+
+    const timeMetrics = getTimeMetrics()
```
cajubelt, line 401: could we try to use Sentry for time metrics instead of this DIY stuff?

src/swap/saga.ts
```diff
@@ -247,8 +250,182 @@ export function* swapSubmitSaga(action: PayloadAction<SwapInfo>) {
...
+    navigate(Screens.SwapExecuteScreen)
+
+    const wallet = yield* call(getViemWallet, networkConfig.viemChain.celo)
+    if (!wallet.account) {
+      // this should never happen
+      throw new Error('no account found in the wallet')
+    }
+
+    const preparedTransactions = action.payload.quote.preparedTransactions
+    if (preparedTransactions?.type !== 'possible') {
```
cajubelt, line 336: would be nice if we had static type guarantees for this, instead of having to check it at runtime

src/swap/saga.ts
```diff
@@ -247,8 +250,182 @@ export function* swapSubmitSaga(action: PayloadAction<SwapInfo>) {
...
+    yield* put(
+      addStandbyTransaction({
+        context: swapExecuteContext,
+        __typename: 'TokenExchangeV3',
+        networkId: networkConfig.defaultNetworkId,
+        type: TokenTransactionTypeV2.SwapTransaction,
+        inAmount: {
+          value: valueToBigNumber(sellAmount)
+            .multipliedBy(guaranteedPrice)
+            .shiftedBy(-toToken.decimals),
```
cajubelt, line 382: I'm confused.. should this be "fromToken"? Seems like you put "fromToken" in and get "toToken" out for a swap, right? But this does the opposite..

src/swap/useSwapQuote.ts
```diff
@@ -67,42 +74,132 @@ function createBaseApproveTransaction(
...
+  }
+}
 
-  return swapTx
+function getFeeCurrencyAddress(feeCurrency: TokenBalance) {
+  return !feeCurrency.isNative ? (feeCurrency.address as Address) : undefined
 }
 
-function getMaxGasCost(txs: TransactionRequestCIP42[]) {
+async function tryEstimateTransactions(
```
cajubelt, line 116: happy to see this getting broken down more

src/swap/useSwapQuote.ts
```diff
@@ -67,42 +74,132 @@ function createBaseApproveTransaction(
...
+    publicClient.celo,
+    feeCurrencyAddress
+  )
+
+  for (const [index, baseTx] of baseTransactions.entries()) {
+    // We can only truly estimate the gas for the first transaction
+    if (index === 0) {
+      const tx = await tryEstimateTransaction(
+        baseTx,
+        feeCurrency,
```
cajubelt, line 133: maybe feeCurrencyAddress should be passed in instead? since it has already been computed

src/swap/useSwapQuote.ts
jeanregisser: true, maybe this could be an option we pass? `{ isDependantTransactions: boolean }`

src/swap/SwapReviewScreen.tsx
```diff
@@ -86,7 +105,22 @@ export function SwapReviewScreen() {
     return estimatedCeloFeeAmount.dividedBy(feeCurrencyPriceUsd).multipliedBy(celoPriceUsd)
   }
 
-  const estimatedFeeAmount = estimateFeeAmount()
+  function getEstimatedMaxFee() {
```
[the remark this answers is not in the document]
jeanregisser: Yes, but this screen is going away.

src/swap/saga.ts
jeanregisser: Yes it will eventually.

src/swap/saga.ts
jeanregisser: Maybe, but there are existing dashboards using this. I'll take a note.
Also I remember we were having difficulty using/analyzing anything metrics related with Sentry.

src/swap/useSwapQuote.ts
jeanregisser: yes but then I wanted to have more info in the logged error (id + symbol).

src/swap/SwapScreen.tsx
jeanregisser: Should be toward the end of the refactor for [multichain support](https://linear.app/valora/project/multichain-swaps-e7ec5b514232).

src/swap/saga.ts
jeanregisser: Yes I was confused at first too. I double checked and `inAmount` is what the user account received, which is the swap `toToken`.

src/swap/useSwapQuote.ts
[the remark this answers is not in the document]
jeanregisser: ahah yes, I was counting to base this on the refactor in https://github.com/valora-inc/wallet/pull/4378,
But it seems we're going this other way way around for now 😅

src/swap/SwapScreen.tsx
jeanregisser: yep! 👍

kathaypacific: i think Jacob has work in flight to update this screen to use tokenId which will fix this problem, so i don't think it's worth doing here

src/swap/saga.ts
kathaypacific: these time metrics aren't used to measure performance. we wanted to be able to analyse swap errors with some data about how long swaps were taking (to answer the question of are users simply just waiting too long to submit the swap after receiving the quote) - it's much easier to do this analysis in Mixpanel if we have all the data there

src/swap/saga.ts
kathaypacific: the in and out is relative to the wallet here 😅 it is a bit confusing, but this seems right. the net incoming amount to the wallet is the sellAmount (fromToken) * price (this logic assumes that we are using the sell flow, where we've specified the sell token amount only)

src/swap/useSwapQuote.ts
```diff
@@ -67,42 +74,132 @@ function createBaseApproveTransaction(
...
-  })
-
-  const approveTx: TransactionRequestCIP42 & { amountToApprove: bigint } = {
-    from: from as Address,
-    to: sellTokenAddress as Address,
-    data,
-    type: 'cip42',
-    amountToApprove,
+  // Approve transaction if the sell token is ERC-20
+  if (allowanceTarget !== zeroAddress && fromToken.address) {
```
kathaypacific, line 78: do we want to add a check for how much is already approved, so we don't need to do this and save the user some gas cost if they've previously done a bigger swap using the same allowanceTarget?

src/swap/useSwapQuote.ts
```diff
@@ -67,42 +74,132 @@ function createBaseApproveTransaction(
...
+        maxPriorityFeePerGas
+      )
+      if (!tx) {
+        return null
+      }
+      transactions.push(tx)
+    } else {
+      if (!baseTx.gas) {
+        throw new Error(
+          'When multiple transactions are provided, all transactions but the first one must have a gas value'
```
kathaypacific, line 144: sorry remind me why the first transaction doesn't need to have a gas value?

kathaypacific (approved):
i walked through the flows and think that they make sense, but i don't have enough brain power left over to offer any detailed improvement ideas because there's a lot going on here. i think it looks good, would be nice to get this one in to unblock the other swaps work and we can iterate as we go :)

src/swap/useSwapQuote.ts
jeanregisser: That's a great idea! Keeping a note for a follow up PR.

src/swap/useSwapQuote.ts
jeanregisser: Because we can estimate it ourselves and be more accurate.

src/swap/SwapScreen.tsx
jeanregisser: oh yes that's right, it's why I didn't touch it either.

jeanregisser:
> i walked through the flows and think that they make sense, but i don't have enough brain power left over to offer any detailed improvement ideas because there's a lot going on here. i think it looks good, would be nice to get this one in to unblock the other swaps work and we can iterate as we go :)

ah yes sorry it got a bit bigger, I thought the `useSwapQuote` refactor would have eventually been done outside of this one, but it didn't happen.

src/viem/getLockableWallet.ts
jeanregisser, line 1: I saw @bakoushin added tests for similar methods in https://github.com/valora-inc/wallet/pull/4320/files#diff-37bbda7bd611274a2b7b50cc627bc821f61407768c34698f9505e2f16ad1f422
So I didn't cover that here.

src/swap/useSwapQuote.ts
cajubelt: could we decide whether to estimate gas for each transaction by whether "gas" is already present on the baseTransaction? Seems like if it's not defined, we definitely need to estimate it, and if it is defined, the consumer of the function probly wouldn't expect it to be over-written

cajubelt (approved):

src/swap/useSwapQuote.ts
jeanregisser: yes great idea, and if it's missing and we estimate and it's dependant transactions, the estimate will fail anyway!

PR #4369, opened 2023-10-23
feat: add new send flow feature gate

MuckT:
### Description

Adds a feature gate `use_new_send_flow` in [Statsig](https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/gates/use_new_send_flow) and uses it to determine if the token bottom sheet should be displayed after selecting a recipient in the send flow. When users without a balance attempt to send they will land on the existing send page after selecting a recipient - previously they would get an empty bottom-sheet. Should be used to direct users to the new send flow once ACT-941 is done.

### Test plan

- [x] Tested locally on iOS
- [x] Unit tests updated

### Related issues

ACT-945

### Backwards compatibility

Yes


src/send/Send.tsx
```diff
@@ -127,7 +129,9 @@ function Send({ route }: Props) {
 
     // Only show currency picker once we know that the recipient is verified,
     // and only if the user is permitted to change tokens.
-    if (defaultTokenIdOverride) {
+    if (defaultTokenIdOverride || getFeatureGate(StatsigFeatureGates.USE_NEW_SEND_FLOW)) {
```
satish-ravi, line 132: the gate can just go in a separate `if` block right? E.g.,
```ts
if (getFeatureGate(StatsigFeatureGates.USE_NEW_SEND_FLOW)) {
  navigate(new send screen) // for now could be old screen
}
else if (defaultTokenOverride) {
  navigate(old send amount)
}
// show bottom sheet
```

MuckT: Currently they both go to the same place, but I think having the separate if matches the A/C better: fixed in e553865f470b40710327a1724e59ab2c1cd8af3a!

satish-ravi (approved):

PR #4371, opened 2023-10-24
fix: show completed standby transactions in the homefeed

kathaypacific:
### Description

There are a few things going on in this PR to achieve the end result:
1. update the TransactionFeed component to show transactions using redux instead of this hook OR redux logic. i'm not sure why it was like this, but the useFetchTransactions hook already saves transactions to redux so i can't see why we can't just rely on this single source of truth.
2. the transactions selector should select from the `transactions` store (where new tx's from blockchain-api are stored) and combine that with completed tx's in the standby tx store. sort by time.
3. update the `StandbyTransaction` type to make step 2 easier: when a tx is confirmed, use the tx receipt to make the standby tx data as similar as possible to what comes back from blockchain-api from the reducer. i didn't bother trying to construct the fees object because i can't see the fee currency from the receipt. this means that between the tx confirmed and blockchain-api catching up, no fees can be shown in the tx details screen.
4. i removed the TRANSACTION_CONFIRMED_VIEM action because it seems to do the same thing as TRANSACTION_CONFIRMED. rather than rewriting the same reducer twice, i just combined the actions.
5. ensure that the `UPDATE_TRANSACTIONS` reducer stores the new transactions without overwriting the old transactions.

### Related issues

- Fixes RET-894

### Backwards compatibility

Y

### DIFF
--- a/src/transactions/feed/queryHelper.ts
+++ b/src/transactions/feed/queryHelper.ts
@@ -25,7 +25,6 @@ const MIN_NUM_TRANSACTIONS = 10
 export interface QueryHookResult {
   loading: boolean
   error: Error | undefined
-  transactions: TokenTransaction[]
   fetchingMoreTransactions: boolean
   fetchMoreTransactions: () => void
 }
@@ -238,7 +237,6 @@ export function useFetchTransactions(): QueryHookResult {
   return {
     loading,
     error,
-    transactions: fetchedResult.transactions,
     fetchingMoreTransactions,
     fetchMoreTransactions: () => {
       if (!fetchedResult.pageInfo) {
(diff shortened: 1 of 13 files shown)


src/transactions/reducer.ts
```diff
@@ -141,21 +148,11 @@ export const reducer = (
   }
 }
 
-function mapForContextId(
```
kathaypacific, line 144: i removed this because the `any` type on the `mapping` makes it very hard to work with.

src/transactions/feed/queryHelper.ts
```diff
@@ -157,10 +142,7 @@ export function useFetchTransactions(): QueryHookResult {
           }
           // If there are new transactions update transactions in redux and fetch balances
           if (hasNewTransaction) {
-            batch(() => {
```
kathaypacific, line 160: not directly related but we no longer need to manually dispatch the token balances fetch, this is done in the home saga already

kathaypacific:
Closing in favour of #4380

PR #4373, opened 2023-10-24
feat: update tx details screen

bakoushin:
### Description

Transaction details screen displays whether the tx is completed / pending / failed.

It also has a button linking to a block explorer for more details.
In the bottom it has the link doing the exact same thing - never miss the tx details!
Works for both Celo and Ethereum.
 
If transaction is failed, instead of a block explorer button a retry button is displayed (for swap, token send tx only). Depending on tx type it takes user to:
* Swap → Swap screen
* Send → Contact selection screen

Note: currently, details of the failed transaction (recipient, amounts) aren't copied. It only navigates to the flow start.

Taps on details / retry buttons, and the block explorer links invoke the appropriate analytics events.

| Completed | Pending | Failed |
|--------|--------|--------|
|  Completed | Pending | Failed | 

### Misc

* DefaultAvatar replaced with a generic user avatar instead of infamous "big nose guy"
* Some icons reused and updated according the last designs to bring a grain of consistency
* Swap tx details screen appearance adjusted to closely mimic the last design

### Follow up

* Switching the default Celo explorer from https://explorer.celo.org to https://celoscan.io will be in a separate PR.

### Test plan

* Tested manually on Alfajores/Sepolia
* Updated unit tests

### Related issues

- Fixes RET-890

### Backwards compatibility

Y


locales/base/translation.json
```diff
@@ -1694,8 +1694,8 @@
     "swapFrom": "Swap from",
     "swapTo": "Swap to",
     "rate": "Rate",
-    "estimatedFee": "Estimated Fee",
-    "viewOnExplorer": "View on CeloExplorer"
```
jeanregisser, line 1698: Are older client stilling using this string? If so we should keep it for backward compatibility.

src/icons/ExclamationCircleIcon.tsx
```diff
@@ -1,13 +1,17 @@
 import * as React from 'react'
 import { Path, Svg } from 'react-native-svg'
 import Colors from 'src/styles/colors'
+interface Props {
```
jeanregisser, line 4: nit:
```suggestion

interface Props {
```

src/transactions/feed/TransactionDetailsScreen.tsx
```diff
@@ -104,23 +116,96 @@ function TransactionDetailsScreen({ navigation, route }: Props) {
...
       content = <SwapContent exchange={transaction as TokenExchange} />
+      retryHandler = () => navigate(Screens.SwapScreenWithBack)
       break
   }
 
+  const transactionNetwork = networkConfig.networkIdToNetwork[transaction.networkId]
+  const openBlockExplorerHandler = transactionNetwork
+    ? () => {
+        navigate(Screens.WebViewScreen, {
+          uri: `${networkConfig.blockExplorerBaseTxUrl[transactionNetwork]}${transaction.transactionHash}`,
```
jeanregisser, line 127: nit: we're probably doing this all over the place, but I think we should join url paths using a function to properly handle cases where the string include `/` or not.

We have one such util from `@celo/utils`, but we could as well add our own. Since we want to eventually get rid of contractkit.

```suggestion
          uri: appendPath(networkConfig.blockExplorerBaseTxUrl[transactionNetwork]}, transaction.transactionHash),
```

src/transactions/feed/TransactionDetailsScreen.tsx
```diff
@@ -104,23 +116,96 @@ function TransactionDetailsScreen({ navigation, route }: Props) {
...
+        })
+      }
+    : undefined
+
+  const primaryActionHanlder =
+    transaction.status === TransactionStatus.Failed ? retryHandler : openBlockExplorerHandler
+
   return (
-    <ScrollView contentContainerStyle={styles.contentContainer}>
-      <SafeAreaView style={styles.content}>{content}</SafeAreaView>
```
jeanregisser, line 112: Why removing `SareAreaView`? Doesn't this cause issue scrolling the full content on devices with the home bar at the bottom (iOS)?

Especially when the font size this increased by the user.

jeanregisser (commented):
Awesome work! 🚀 🚀 

Left a couple of small questions.

locales/base/translation.json
bakoushin: Good catch, thanks! Restored it.

src/transactions/feed/TransactionDetailsScreen.tsx
bakoushin: What if we import an NPM library for that, like [url-join](https://www.npmjs.com/package/url-join) to not reinvent the wheel?
What do you think @jeanregisser?

src/analytics/Properties.tsx
```diff
@@ -1381,6 +1381,13 @@ interface DappShortcutsProperties {
   [DappShortcutsEvents.dapp_shortcuts_reward_tx_rejected]: DappShortcutClaimRewardEvent
 }
 
+interface TransactionDetailsProperties {
+  [TransactionDetailsEvents.transaction_details_tap_details]: undefined
```
kathaypacific, line 1385: should we add the transaction type and status to all these events? so we can learn if the tx details screen is most helpful for certain transaction types or statuses :)

src/styles/colors.tsx
```diff
@@ -36,6 +36,15 @@ export enum Colors {
...
   lightGreen = '#E6F7EC',
 
+  successDark = '#137211',
+  successLight = '#F1FDF1',
+
+  warningDark = '#9C6E00',
+  warningLight = '#FFF9EA',
+
+  errorDark = '#C93717',
+  errorLight = '#FBF2F0',
```
kathaypacific, line 46: just checking that you've aligned with Nitya on these colors and their semantics? e.g. errors will now have a dark and light variation in future designs and they will be those values? all good if you've done this, but otherwise i think it will be confusing to use these in the future (e.g. if we need a `errorMediumDark`, `errorMediumDarkerDark` 🙈) and would recommend to either use existing colors in the closest shade / using one off hard coded colors in the components where you need them / naming these like `gray1` etc. if we don't have a clear purpose for them.

src/transactions/feed/TransactionDetailsScreen.tsx
bakoushin: Oh, I missed that. Thanks for the suggestion! Will bring `SafeAreaView` back.

src/web3/networkConfig.ts
```diff
@@ -232,6 +244,10 @@ const networkConfigs: { [testnet: string]: NetworkConfig } = {
     celoExplorerBaseTokenUrl: CELO_EXPLORER_BASE_TOKEN_URL_ALFAJORES,
     celoExplorerBaseTxUrl: CELO_EXPLORER_BASE_TX_URL_ALFAJORES,
     celoExplorerBaseAddressUrl: CELO_EXPLORER_BASE_ADDRESS_URL_ALFAJORES,
+    blockExplorerBaseTxUrl: {
+      [Network.Celo]: CELO_EXPLORER_BASE_TX_URL_ALFAJORES,
```
kathaypacific, line 248: this is fine...but if you did `[NetworkId['celo-alfajores']]: CELO_EXPLORER_BASE_TX_URL_ALFAJORES` instead then you don't need `networkIdToNetwork`? i'm not sure, all these mappings make me confused

kathaypacific: also, maybe we can change `CELO_EXPLORER_BASE_URL_ALFAJORES` and `CELO_EXPLORER_BASE_URL_MAINNET` to celo scan? :D

kathaypacific (commented):
This is looking great :) just a few small comments

src/styles/colors.tsx
bakoushin: I took these names from Figma because it looks like they are part of some library designers are using 😅 
But I will check with Nitya if are they stable as a color system and can be reusable by name or not.

src/analytics/Properties.tsx
bakoushin: Sure, that's a great idea!

src/web3/networkConfig.ts
bakoushin: thanks for the feedback about confusion! it makes sense.
I refactored it a bit to make more straightforward.

as for moving to CeloScan, would you mind if I bring that in a separate PR immediately after this one is merged?
it has to deal with other references and translations too, and I would like to have those changes clear, not messed up with all changed in here.

src/transactions/feed/TransactionDetailsScreen.tsx
bakoushin: since we are concatenating some appendix with base url, this perfectly fits to built-in [URL API](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL), switched to it

dievazqu (commented):
I'm a bit concern about the UI component changes. It seems some components might be impacted by these changes, and I don't know if we want that.

src/icons/Checkmark.tsx
```diff
@@ -22,12 +22,12 @@ export default class Checkmark extends React.PureComponent<Props> {
...
         width={this.props.width}
         height={this.props.height}
-        viewBox="0 0 32 32"
+        viewBox="0 0 16 16"
         fill="none"
         testID={this.props.testID}
       >
         <Path
-          d="M8.414 15.086a2 2 0 10-2.828 2.828l2.828-2.828zM12.5 22l-1.414 1.414a2 2 0 002.8.029L12.5 22zm13.885-10.557a2 2 0 00-2.77-2.886l2.77 2.886zm-20.8 6.471l5.5 5.5 2.83-2.828-5.5-5.5-2.83 2.828zm8.3 5.529l12.5-12-2.77-2.886-12.5 12 2.77 2.886z"
+          d="M13.667 4.886 6.124 12.43 2.667 8.972l.886-.886 2.57 2.564L12.78 4l.887.886Z"
```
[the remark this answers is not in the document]
bakoushin: I intentionally updated current shared icons/components according to new designs: if we got a new look for `Checkmark`, let's consider that look should be applied everywhere.

Otherwise, we shouldn't use shared elements and implement everything for each screen separately.

Or pollute shared folders with Checkmark2, Checkmark3, CheckmarkWithBlowsAndWhistles etc.

Let me check with Nitya to see if she is okay with my proposed approach. If not, I will revert those in favor of dedicated icons for those new designs.

src/transactions/feed/TransactionDetailsScreen.tsx
```diff
@@ -104,23 +118,109 @@ function TransactionDetailsScreen({ navigation, route }: Props) {
...
+      : undefined
+
+  const primaryActionHanlder =
+    transaction.status === TransactionStatus.Failed ? retryHandler : openBlockExplorerHandler
+
   return (
-    <ScrollView contentContainerStyle={styles.contentContainer}>
-      <SafeAreaView style={styles.content}>{content}</SafeAreaView>
+    <ScrollView contentContainerStyle={styles.container}>
+      <SafeAreaView edges={['right', 'bottom', 'left']}>
```
jeanregisser, line 143: I'm not entirely sure `left` and `right` are needed. Probably already handled by the wrapper navigator.

jeanregisser (approved):
LGTM 👍 

Approving for my part of the review, but we should address the other reviews before merging.

src/web3/networkConfig.ts
```diff
@@ -70,6 +70,14 @@ interface NetworkConfig {
   celoTokenAddress: string
 }
 
+export type blockExplorerUrls = {
+  [key in NetworkId]: { baseTxUrl: string }
+}
+
+export type networkIdToNetwork = {
```
kathaypacific, line 77: nit: i think it is conventional to use Capital letters for types?

kathaypacific (approved):
🥳 yay looks great!!

src/components/HorizontalLine.tsx
```diff
@@ -1,18 +1,20 @@
...
+import { Colors } from 'src/styles/colors'
+import { Spacing } from 'src/styles/styles'
 
-export default function HorizontalLine() {
-  return <View style={styles.container} />
+export interface Props {
+  color?: Colors
+}
+
+export default function HorizontalLine({ color = Colors.gray2 }: Props) {
```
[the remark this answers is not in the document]
bakoushin: I've checked with @nityarao yesterday. She confirmed that intent is to globally update to `gray2`.

src/components/HorizontalLine.tsx
```diff
@@ -1,18 +1,20 @@
...
 }
 
 const styles = StyleSheet.create({
   container: {
+    height: 1,
     width: '100%',
-    borderStyle: 'solid',
-    borderTopWidth: 1,
-    borderTopColor: colors.gray1,
-    marginTop: 10,
```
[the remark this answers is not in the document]
bakoushin: I agree, in addition to color changes I introduced also the spacing changes between elements. Although the differences are subtle, it potentially may break the existing layouts, which we obviously don't want.

I will revert all `HorizontalLine` changes back, except for color (this was approved by @nityarao).

For new designs I will use a new component `RowDivider` as it is in the designs.

src/icons/Checkmark.tsx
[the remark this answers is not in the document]
bakoushin: I've checked with @nityarao yesterday and she approved that we can update icons globally. 

Specifically for the `Checkmark`, I have updated it as per designs: https://www.figma.com/file/erFfzHvSTm5g1sjK6jWyEH/Working-Design-System-Components?type=design&node-id=585-17&mode=design&t=fihuWlOHIEwfJsx1-0

Note that sizes remain intact.

src/icons/ClockIcon.tsx
```diff
@@ -17,9 +17,9 @@ export default class ClockIcon extends React.PureComponent<Props> {
 
   render() {
     return (
-      <Svg width={this.props.width} height={this.props.height} viewBox="0 0 512 512" fill="none">
+      <Svg width={this.props.width} height={this.props.height} viewBox="0 0 16 16" fill="none">
         <Path
-          d="M232 120C232 106.7 242.7 96 256 96C269.3 96 280 106.7 280 120V243.2L365.3 300C376.3 307.4 379.3 322.3 371.1 333.3C364.6 344.3 349.7 347.3 338.7 339.1L242.7 275.1C236 271.5 232 264 232 255.1L232 120zM256 0C397.4 0 512 114.6 512 256C512 397.4 397.4 512 256 512C114.6 512 0 397.4 0 256C0 114.6 114.6 0 256 0zM48 256C48 370.9 141.1 464 256 464C370.9 464 464 370.9 464 256C464 141.1 370.9 48 256 48C141.1 48 48 141.1 48 256z"
```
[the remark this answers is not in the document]
bakoushin: I've checked with @nityarao yesterday and she approved that we can update icons globally. 

Specifically for the `ClockIcon`, I have updated it as per designs: 
https://www.figma.com/file/erFfzHvSTm5g1sjK6jWyEH/Working-Design-System-Components?type=design&node-id=566-17&mode=design&t=fihuWlOHIEwfJsx1-0

Note that sizes remain intact.

src/icons/ArrowRight.tsx
```diff
@@ -1,19 +1,16 @@
...
 
 interface Props {
-  color?: colors
+  color?: Colors
+  size?: number
 }
 
-const ArrowRight = ({ color = colors.greenUI }: Props) => (
-  <Svg width={18} height={18} fill="none">
+const ArrowRight = ({ color = Colors.greenUI, size = 24 }: Props) => (
```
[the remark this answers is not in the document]
bakoushin: Yes, that's the wrong one. Reverted it.

What I really wanted is `ArrowRightThick`, updated it to have custom size and color.

src/transactions/feed/TransactionFeed.tsx
```diff
@@ -31,6 +31,20 @@ function TransactionFeed() {
   const { loading, error, transactions, fetchingMoreTransactions, fetchMoreTransactions } =
     useFetchTransactions()
 
+  const pedningTxs = new Set([
```
dievazqu, line 34: nit: typo `pedningTxs`

dievazqu (approved):
LGTM

PR #4374, opened 2023-10-24
fix: ensure all transactions are saved correctly in redux

kathaypacific:
### Description

Context: https://valora-app.slack.com/archives/C025V1D6F3J/p1698145522738719

When new transactions are fetched, we should store them alongside the previously fetched ones instead of overwriting them.

### Test plan

Homefeed works as previously, the `UPDATE_TRANSACTIONS` action is only dispatched when new transactions are received.

Before this change, after a new transaction is received or sent. Note that `UPDATE_TRANSACTIONS` is dispatched every 10 seconds (poll rate), but I am doing nothing with the wallet during this time.
Screenshot 2023-10-25 at 12 09 59

After this change - note that the `UPDATE_TRANSACTIONS` is dispatched once when a new transaction is registered, and not more than that.
Screenshot 2023-10-25 at 12 07 54

After this change - nothing too exciting here, but this video shows that new transaction items continue to appear correctly (the receive scenario).
https://github.com/valora-inc/wallet/assets/20150449/706769a0-2f17-4005-9c9b-33131224c6fc

### Related issues

Fixes RET-894

### Backwards compatibility

Y


src/transactions/utils.ts
```diff
@@ -47,3 +48,17 @@ export function groupFeedItemsInSections(
...
     }))
 }
+
+export const deduplicateTransactions = (
+  existingTxs: TokenTransaction[],
+  incomingTxs: TokenTransaction[]
+) => {
+  const currentHashes = new Set(existingTxs.map((tx) => tx.transactionHash))
+  const transactionsWithoutDuplicatedHash = existingTxs.concat(
+    incomingTxs.filter((tx) => !isEmpty(tx) && !currentHashes.has(tx.transactionHash))
```
jh2oman, line 58: It looks like if incomingTxs has duplicates within itself then those will not be caught by this method. Is that a case we need to cover?

kathaypacific: that's a good point....i copied and paste this function so i didn't think too closely on this. if there are duplicates in `incomingTxs`, that should be a bug with blockchain-api which we can fix on the fly. the duplicate transaction would also be removed the next time this reducer is run (so i guess the impact of this issue depends on how often the user has new tx's)....i can add a filter to remove duplicates from the incoming txs

kathaypacific: i updated this @jh2oman :)

src/transactions/utils.ts
```diff
@@ -1,3 +1,4 @@
+import { uniqBy } from 'lodash'
```
jeanregisser, line 1: nit: I know we're already using lodash in a bunch of places, but maybe if we had an implementation not using it, maybe we shouldn't touch it so it's easier to eventually get rid of lodash as a whole?

jeanregisser (commented):
> When new transactions are fetched, we should store them alongside the previously fetched ones instead of overwriting them.

Are we sure about that? I think we had this behavior in the past and this caused faulty data coming from blockchain-api to never be cleared from redux when fetching the latest data.

src/transactions/utils.ts
kathaypacific: this is news to me, i thought lodash was a good lib. why do we want to move away? i would have needed to introduce more logic to this function to address Jacob's comment above. i already find this function a little hard to read, so lodash seemed like the more readable and efficient option here. do we have a list of dependencies that we want to move away from?

jeanregisser: oh yes this is really a side comment.
Maybe we can bring this up at the next eng team meeting? lodash is perfectly fine, but there's a bunch of things it does which are now unnecessary with modern JS.

https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore

kathaypacific: yeah, we can discuss. anyway, i've removed lodash here in case

kathaypacific:
> > When new transactions are fetched, we should store them alongside the previously fetched ones instead of overwriting them.
> 
> Are we sure about that? I think we had this behavior in the past and this caused faulty data coming from blockchain-api to never be cleared from redux when fetching the latest data.

@jeanregisser does https://github.com/valora-inc/wallet/pull/4374/commits/4b3653b10965d71d3b8791b31525401ba249fa5e help to address these concerns? TBH i wasn't aware of faulty data coming from blockchain-api

kathaypacific:
i think i have a better way to solve this problem, closing this

PR #4375, opened 2023-10-24
fix: tagTxsWithProviderInfo saga failures

kathaypacific:
### Description

`txHashesToProvider` can return null, which causes this `tagTxsWithProviderInfo` saga to fail. no real user impact but i saw it during dev.
Simulator Screenshot - iPhone 14 Pro - 2023-10-24 at 14 51 40

### Test plan

this error no longer persists.

### Related issues

n/a

### Backwards compatibility

y


src/fiatExchanges/saga.ts
```diff
@@ -135,14 +135,14 @@ export function* tagTxsWithProviderInfo({ transactions }: UpdateTransactionsActi
     Logger.debug(`${TAG}@tagTxsWithProviderInfo`, `Checking ${transactions.length} txs`)
 
     const providerLogos: ProviderLogos = yield* select(providerLogosSelector)
-    const txHashesToProvider: TxHashToProvider = yield* call(fetchTxHashesToProviderMapping)
+    const txHashesToProvider: TxHashToProvider | null = yield* call(fetchTxHashesToProviderMapping)
```
jh2oman, line 138: Can you instead fix the typing in `fetchTxHashesToProviderMapping`?

jh2oman (changes requested):

src/fiatExchanges/saga.ts
kathaypacific: yes good one, done

jh2oman (approved):

PR #4378, opened 2023-10-25
refactor(viem): extend prepareTransactions for send use case

cajubelt:
### Description

extends the prepareTransactions helper @jeanregisser created for swaps in #4322 for re-use in the send flow

a lot of the code is just moved over from `useSwapQuote.ts`. I tried leaving comments indicating what I changed and what I just moved over

intended as a starting point for lazy fee calculations with viem (instead of calculating fees for sends with contractkit and saving to redux)

### Test plan

jest tests

### Related issues

### Backwards compatibility

na


src/swap/SwapScreen.tsx
```diff
@@ -481,8 +481,14 @@ export function SwapScreen({ route }: Props) {
         <PreparedTransactionsReviewBottomSheet
           forwardedRef={preparedTransactionsReviewBottomSheetRef}
           preparedTransactions={exchangeRate.preparedTransactions}
-          onAcceptDecreaseSwapAmountForGas={({ decreasedSwapAmount }) => {
-            handleChangeAmount(updatedField)(decreasedSwapAmount.toString())
+          onAcceptDecreaseSwapAmountForGas={({ decreasedSpendAmount }) => {
+            handleChangeAmount(updatedField)(
+              // ensure units are for the asset whose amount is being selected by the user
```
cajubelt, line 486: I moved this out of `prepareTransactions` since it is specific to sends, and I thought it made the helper easier to reason about for its outputs to have consistent units

src/viem/prepareTransactions.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    networkId: NetworkId['celo-mainnet'],
+    tokenId: 'celo-mainnet:0xspend',
+    symbol: 'SPEND',
+    name: 'Spend token',
+  }
+  const mockPublicClient = { estimateGas: jest.fn() }
+  beforeAll(() => {
+    publicClient[Network.Celo] = mockPublicClient as any
+  })
+  it('not enough balance for gas', async () => {
```
cajubelt, line 54: I wrote this unit test for debugging. Will follow up with more unit tests later.

src/viem/prepareTransactions.ts
```diff
@@ -0,0 +1,155 @@
...
+          publicClient.celo,
+          feeCurrencyAddress
+        )
+        const transaction = {
+          ...baseTransaction,
+          maxFeePerGas,
+          maxPriorityFeePerGas,
+          feeCurrency: feeCurrencyAddress,
+        }
+        if (baseTransaction.gas) {
```
cajubelt, line 81: this is the main difference between the generalized `prepareTransactions` and its previous form that was specific to swaps. It lets the consumer tell this helper whether gas should be estimated with viem, or a pre-calculated value should be used (as in the case of swaps, where you want to check with viem for "approve" but use a precalculated string for "swap")

src/viem/prepareTransactions.ts
```diff
@@ -0,0 +1,155 @@
...
+      // No balance, try next fee currency
+      continue
+    }
+    const feeCurrencyAddress = !feeCurrency.isNative ? (feeCurrency.address as Address) : undefined
+    const transactions: (TransactionRequestCIP42 | null)[] = await Promise.all(
+      baseTransactions.map(async (baseTransaction) => {
+        const { maxFeePerGas, maxPriorityFeePerGas } = await estimateFeesPerGas(
+          publicClient.celo,
+          feeCurrencyAddress
+        )
```
jeanregisser, line 74: We can move this out of this map as it doesn't change for a given feeCurrency.

src/viem/prepareTransactions.ts
jeanregisser: Yeah this is great. One thing is I think we shouldn't process the base transactions in parallel.
And only estimate the first one if gas is not set. 
As they can be dependant transactions, like for swaps.
I had started doing this in my local branch for swaps just yesterday :D

jeanregisser (commented):
Love this ❤️

src/viem/prepareTransactions.ts
jeanregisser: This is what I was proposing about this specific point: https://github.com/valora-inc/wallet/pull/4368/commits/327adddbf846b41be9e8ffc8042ef028fa0484c3#diff-c9b55f4aa0b75764dddb8dac1eb9aa7df00d6195311c998a0df0a95bfd726eacR116-R162

src/viem/prepareTransactions.ts
```diff
@@ -0,0 +1,206 @@
...
+        maxPriorityFeePerGas,
+        // Don't include the feeCurrency field if not present.
+        // See https://github.com/wagmi-dev/viem/blob/e0149711da5894ac5f0719414b4ecc06ccaecb7b/src/chains/celo/serializers.ts#L164-L168
+        ...(feeCurrencyAddress && { feeCurrency: feeCurrencyAddress }),
+        // We assume the provided gas value is with the native fee currency
+        // If it's not, we add the static padding
+        gas: baseTx.gas + BigInt(feeCurrency.isNative ? 0 : STATIC_GAS_PADDING),
+      })
+    }
+  }
```
jeanregisser, line 132: Like you said in https://github.com/valora-inc/wallet/pull/4368#discussion_r1372395257
> seems like there are cases this might not be true, like if we wanted to use this for estimating gas for registering the DEK and doing a c-stable transfer

I proposed to add an option to control this https://github.com/valora-inc/wallet/pull/4368#discussion_r1372726574
> true, maybe this could be an option we pass? { isDependantTransactions: boolean }

But maybe we could do that later as an optimization? if we really want/need this.

src/send/hooks.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Prepare a transaction for sending an ERC-20 token.
+ *
+ * Works for ERC-20 and native transfers.
+ *
+ * NOTE: does not include fees for registering a DEK before sending a c-stable, since that is a different transaction.
+ *
+ */
+export async function prepareSendERC20Transaction(
```
jeanregisser, line 16: nit: should this be outside of `hooks` if it isn't a hook?

jeanregisser (commented):
Looking great! 🚀

src/viem/prepareTransactions.ts
cajubelt: thanks. in my efforts to keep my PR in sync with yours I ended up copying over things that I questioned on your PR lol. I updated it

src/send/hooks.ts
cajubelt: I moved it to prepareTransactions, but then it felt weird having prepareSwapTransactions in a hooks module as well, so I moved that to prepareTransactions too

src/viem/prepareTransactions.ts
```diff
@@ -0,0 +1,314 @@
...
+
+  return {
+    type: 'need-decrease-spend-amount-for-gas',
+    maxGasCost: adjustedMaxGasCost,
+    feeCurrency: result.feeCurrency,
+    decreasedSpendAmount: maxAmount,
+  } satisfies PreparedTransactionsNeedDecreaseSpendAmountForGas
+}
+
+function createBaseSwapTransactions(
```
cajubelt, line 210: just moved and renamed

src/viem/prepareTransactions.ts
```diff
@@ -0,0 +1,314 @@
...
+        return null
+      }
+      transactions.push(tx)
+    }
+  }
+
+  return transactions
+}
+
+export async function prepareTransactions({
```
cajubelt, line 139: this is the main change (making this work for transfer as well as swap)

src/viem/prepareTransactions.ts
```diff
@@ -0,0 +1,314 @@
...
+  const transactions: TransactionRequestCIP42[] = []
+
+  const feeCurrencyAddress = getFeeCurrencyAddress(feeCurrency)
+  const { maxFeePerGas, maxPriorityFeePerGas } = await estimateFeesPerGas(
+    publicClient.celo,
+    feeCurrencyAddress
+  )
+
+  for (const baseTx of baseTransactions) {
+    if (baseTx.gas) {
```
cajubelt, line 108: this line was changed. previously we just estimated gas for the first tx, now makes gas estimation possible for any of the transaction inputs that don't already have a gas field

src/viem/prepareTransactions.ts
```diff
@@ -0,0 +1,314 @@
...
+    maxGasCost += BigInt(tx.gas) * BigInt(tx.maxFeePerGas)
+  }
+  return new BigNumber(maxGasCost.toString())
+}
+
+function getFeeCurrencyAddress(feeCurrency: TokenBalance) {
+  return !feeCurrency.isNative ? (feeCurrency.address as Address) : undefined
+}
+
+async function tryEstimateTransaction({
```
cajubelt, line 53: copied with tweak (address and symbol given as params, instead of TokenBalanceWithAddress)

src/viem/prepareTransactions.ts
```diff
@@ -0,0 +1,314 @@
...
+  for (const tx of txs) {
+    if (!tx.gas || !tx.maxFeePerGas) {
+      throw new Error('Missing gas or maxFeePerGas')
+    }
+    maxGasCost += BigInt(tx.gas) * BigInt(tx.maxFeePerGas)
+  }
+  return new BigNumber(maxGasCost.toString())
+}
+
+function getFeeCurrencyAddress(feeCurrency: TokenBalance) {
```
cajubelt, line 49: copied

src/viem/prepareTransactions.ts
```diff
@@ -0,0 +1,314 @@
...
+  type: 'not-enough-balance-for-gas'
+  feeCurrencies: TokenBalance[]
+}
+
+export type PreparedTransactionsResult =
+  | PreparedTransactionsPossible
+  | PreparedTransactionsNeedDecreaseSpendAmountForGas
+  | PreparedTransactionsNotEnoughBalanceForGas
+
+export function getMaxGasCost(txs: TransactionRequestCIP42[]): BigNumber {
```
cajubelt, line 38: moved with tweak (return BigNumber instead of bigint)

src/viem/prepareTransactions.ts
```diff
@@ -0,0 +1,326 @@
...
+    maxGasCost += BigInt(tx.gas) * BigInt(tx.maxFeePerGas)
+  }
+  return new BigNumber(maxGasCost.toString())
+}
+
+export function getFeeCurrencyAddress(feeCurrency: TokenBalance) {
+  return !feeCurrency.isNative ? (feeCurrency.address as Address) : undefined
+}
+
+export async function tryEstimateTransaction({
```
cajubelt, line 53: mainly the same, besides parameters (symbol and address instead of the whole TokenBalanceWithAddress object)

src/viem/prepareTransactions.ts
jeanregisser: This looks good in most cases.

I'm trying to anticipate incorrect usage of this function.
Let's say someone calls it with multiple dependant TXs (like for swaps) and doesn't set the `gas` for any of them.
The first estimate will work, and then for the next TX it should likely fail (because it needed the first TX to have been executed), but then will just try the next fee currency and fail again.
And eventually it will return a `not-enough-balance-for-gas` result.
Which isn't correct.

But now I'm thinking we could maybe improve how we handle `if (e instanceof EstimateGasExecutionError) {` and look at the exact error message instead and if it's really about not enough gas, we can indeed try the next fee currency, otherwise we must abort and throw that we can't estimate (i.e. the dependant TXs case without gas set)

I'm happy to try this in a follow up PR not to hold this one more than necessary.
What do you think?

jeanregisser (commented):
Awesome! 👍 

I proposed to add a few more tests in https://github.com/valora-inc/wallet/pull/4398

src/viem/prepareTransactions.test.ts
```diff
@@ -0,0 +1,332 @@
...
+import erc20 from 'src/abis/IERC20'
+
+jest.mock('src/viem/estimateFeesPerGas')
+jest.mock('viem')
+
+describe('prepareTransactions module', () => {
+  const mockFeeCurrencies: TokenBalanceWithAddress[] = [
+    {
+      address: '0xfee1',
+      balance: new BigNumber(100), // 10k wei, 100.0 decimals
```
jeanregisser, line 26: nit: should we use `units` instead? `wei` is really 10^-18 ETH.
Otherwise the step to name functions or variables with `wei` as a suffix / prefix isn't far. And may really confuse us.

Or has `wei` been really used interchangeably to talk about the smallest unit of any token?

src/viem/prepareTransactions.ts
cajubelt: it does sound like in the ideal case, we would handle it more intelligently when gas estimation fails because the transaction failed for reasons unrelated to gas (in this case, because the transferFrom amount wasn't approved yet for the swap contract)

src/viem/prepareTransactions.ts
cajubelt: moved them back

src/viem/prepareTransactions.ts
cajubelt: if you're willing to give it a shot I'd appreciate it. I do think being able to handle independent transfers will be helpful for DEK registration

src/viem/prepareTransactions.test.ts
cajubelt: that's a good point, I made the same mistake in `prepareERC20TransferTransaction`. wei is the smallest unit of eth, and I've seen it used as the smallest units of other tokens that use 18 decimal places, but I don't think it applies across the board for tokens that use different numbers of decimal places.

cajubelt: I renamed and added docs to `prepareERC20TransferTransaction`

jeanregisser (approved):
🙌

PR #4386, opened 2023-10-25
feat(send): Add new send recipient screen

jophish:
### Description

For ACT-949. Adds in a new initial recipient selection screen on the send flow, gated behind the [this new Statsig feature gate](https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/gates/use_new_recipient_screen).

### Test plan

#### User has no previous transactions

https://github.com/valora-inc/wallet/assets/569401/bf590ec8-3ccd-495c-a704-06359d7bd963

#### User has previous transactions

https://github.com/valora-inc/wallet/assets/569401/3abe0407-d3f2-4e97-8805-7c6c562a1d04

### Related issues

- Fixes ACT-949.

### Backwards compatibility

Yes.


src/components/TextInput.tsx
```diff
@@ -75,9 +75,9 @@ export class CTextInput extends React.Component<Props, State> {
         <RNTextInput
           ref={forwardedRef}
           style={[
+            inputStyle ?? fontStyles.regular,
```
jophish, line 78: This is still a bit of a head-scratcher for me. Basically, this was never respecting font styling overrides. If `fontStyles.regular` was included _anywhere_ (either above or below `inputStyle`) the font styling in `inputStyle` would never be respected. This fix required a handful of snapshots to be updated. An alternative solution could have been to add a new param to this component such that existing snapshots wouldn't need to be updated, but that felt pretty hacky, considering we already have at least two styling props on the component.

locales/base/translation.json
```diff
@@ -1860,6 +1860,41 @@
       "withdraw": "Transfer tokens to a bank account, mobile money, gift card etc."
     }
   },
+  "sendRedesign": {
```
cajubelt, line 1863: "sendRedesign" is funky naming because it only really makes sense right now, as we are updating the designs on the send flow. Within a few months this should be the only send flow that exists, making the "redesign" part quickly outdated, and if/when send designs are updated again, it will be actively confusing. I would even prefer "send2" or "sendV2", since at least we can increment the version number as designs evolve

This wouldn't be such a big deal if it was just on the translation file, but it appears to be all over the PR, from filenames to analytics events, components, screens, variable names, ...

src/components/TextInput.tsx
cajubelt: might make sense to put this in its own PR

src/send/SendRedesign.tsx
```diff
@@ -0,0 +1,277 @@
...
+
+  const [searchQuery, setSearchQuery] = useState('')
+  const throttledSearch = throttle((searchInput: string) => {
+    setSearchQuery(searchInput)
+  }, SEARCH_THROTTLE_TIME)
+
+  const [shouldImportContacts, setShouldImportContacts] = useState(false)
+
+  const recentRecipients = useSelector((state) => state.send.recentRecipients)
+  const shouldShowGetStarted = !recentRecipients.length
```
cajubelt, line 144: nit: "should" prefix is redundant

src/send/SendRedesign.tsx
```diff
@@ -0,0 +1,277 @@
...
+  const defaultCountryCode = useSelector(defaultCountryCodeSelector)
+
+  const { recipientVerificationStatus, recipient, setSelectedRecipient } =
+    useFetchRecipientVerificationStatus()
+
+  const dispatch = useDispatch()
+
+  useAsync(async () => {
+    if (shouldImportContacts) {
+      // TODO: Currently if a user does not have a phone number linked, this silently fails
```
cajubelt, line 155:
```suggestion
      // TODO(ACT-949): Currently if a user does not have a phone number linked, this silently fails
```

src/send/SendRedesign.tsx
```diff
@@ -0,0 +1,277 @@
...
+      if (permissionGranted) {
+        dispatch(importContacts())
+      }
+      setShouldImportContacts(false)
+    }
+  }, [shouldImportContacts])
+
+  const onPressContacts = () => {
+    ValoraAnalytics.track(SendEvents.send_redesign_invite)
+    setShouldImportContacts(true)
```
cajubelt, line 166: should this just do the contacts import directly, instead of toggling state which triggers the useAsync hook which does the import?

cajubelt: not sure the extra layer is useful, since shouldImportContacts isn't used for anything else

src/send/SendRedesign.tsx
```diff
@@ -0,0 +1,277 @@
...
+    })
+  }
+
+  const buildSections = (): Section[] => {
+    const sections = [{ key: t('sendRedesign.recents'), data: recentRecipients }].filter(
+      (section) => section.data.length > 0
+    )
+
+    return sections
+  }
```
cajubelt, line 182:
```suggestion
  const sections: Section[] = []
  if (recentRecipients.data.length) {
    sections.push({ key: t('sendRedesign.recents'), data: recentRecipients })
  }
```
code golf. requires you to not call "buildSections" later.. but that doesn't appear to be necessary.

cajubelt (approved):

locales/base/translation.json
satish-ravi: +1 to renaming this. maybe `SendSelectRecipient` is an option?

src/components/TextInput.tsx
satish-ravi: if styles.input comes later than inputStyle, wouldn't this mean that padding stuff can never be overridden? It seems like some of the usages of the inputStyle is overriding the padding. A few examples:
- https://github.com/valora-inc/wallet/blob/1b382a974557a7e51aaf21b91073613bb31740cf/src/swap/SwapAmountInput.tsx#L82
- https://github.com/valora-inc/wallet/blob/1b382a974557a7e51aaf21b91073613bb31740cf/src/components/SearchInput.tsx#L17

Also, +1 to making these changes on a separate PR

src/statsig/types.ts
```diff
@@ -35,6 +35,7 @@ export enum StatsigFeatureGates {
   SHOW_ASSET_DETAILS_SCREEN = 'show_asset_details_screen',
   RESTRICT_SUPERCHARGE_FOR_CLAIM_ONLY = 'restrict_supercharge_for_claim_only',
   USE_VIEM_FOR_SWAP = 'use_viem_for_swap',
+  USE_NEW_RECIPIENT_SCREEN = 'use_new_recipient_screen',
```
satish-ravi, line 38: How does this feature gate interact with the `USE_NEW_SEND_FLOW` gate? Are they independent? Or if the new recipient screen is shown, should the new send flow always be shown?

locales/base/translation.json
jophish: Yeah -- I like `SendSelectRecipient`! Adjusted naming throughout.

src/statsig/types.ts
jophish: Yeah, `USE_NEW_RECIPIENT_SCREEN` should imply the rest of the new send flow being used; we can set that up in Statsig, I believe.

PR #4387, opened 2023-10-25
chore(deps): upgrade crypto-js from 4.1.1 to 4.2.0

MuckT:
### Description

Upgrades [crypto-js](https://github.com/brix/crypto-js) following the release of a [new security advisory](https://github.com/brix/crypto-js/security/advisories/GHSA-xwcq-pm8m-c4vf).

### Test Plan

- [x] Test that account made on a prior version handle the upgrade without issue.

### Related issues

N/A

### Backwards compatibility

Yes


yarn.lock
```diff
@@ -7339,11 +7339,16 @@ crypto-js@^3.1.9-1:
   resolved "https://registry.yarnpkg.com/crypto-js/-/crypto-js-3.1.9-1.tgz#fda19e761fc077e01ffbfdc6e9fdfc59e8806cd8"
   integrity sha1-/aGedh/Ad+Af+/3G6f38WeiAbNg=
 
-crypto-js@^4.0.0, crypto-js@^4.1.1:
+crypto-js@^4.0.0:
   version "4.1.1"
   resolved "https://registry.yarnpkg.com/crypto-js/-/crypto-js-4.1.1.tgz#9e485bcf03521041bd85844786b83fb7619736cf"
   integrity sha512-o2JlM7ydqd3Qk9CA0L4NL6mTzU2sdx96a+oOfPu8Mkl/PK51vSyoi8/rQ8NknZtk44vq15lmhAj9CIAGwgeWKw==
 
+crypto-js@^4.2.0:
```
cajubelt, line 7347:
```suggestion
crypto-js@^4.0.0, crypto-js@^4.2.0:
```
I think this should fix the vulnerability warning

cajubelt (commented):
ci fails due to a vulnerable version of crypto-js still in the lockfile. Also, @types/crypto-js needs to be bumped too

cajubelt:
also, dependabot opens these automatically fyi https://github.com/valora-inc/wallet/pull/4385

yarn.lock
MuckT: It looks like there are not updated types yet; the latest I see from `@types/crypto-js` is `4.1.3`

cajubelt (approved):

kathaypacific:
hey @MuckT i went ahead and merged #4385 instead to unblock the CI. i see that you added a yarn resolution for the crypto-js version in this PR, which i wasn't sure we wanted (we usually save the resolutions for indirect dependencies only)

jeanregisser:
I think we must use the resolution as `@celo/utils` blocks it to 3.x
It will be harmless as it doesn't even use it internally.
Also it's been [removed since then](https://github.com/celo-org/celo-monorepo/commit/5cfd16214ca7ef7a7ff428c7d397933b3e1eeb51#diff-423c3037cc6f08ccc85d77b6bdfb68b38622438db3c710d11c21827ba35a82f9).

jeanregisser (approved):
Thanks!! 💯 

Once we upgrade `@celo/utils` we can even remove the resolution.

PR #4388, opened 2023-10-26
fix: ensure transactions for all networks are correctly saved in redux

kathaypacific:
### Description

Context https://valora-app.slack.com/archives/C025V1D6F3J/p1698145522738719

As the title. Currently we are writing and overwriting the transactions redux store constantly. The issue is happening because blockchain-api returns transactions per network, and we save these transactions for a certain network (replacing the existing transactions) to redux. If there is more than one network enabled, the result is constant overwriting since we are reading the transaction hashes in the redux store to decide whether there are new transactions....

The fix:
- make redux schema reflect how the data that we receive, since it acts like a cache layer
- use selectors to transform the data into what we want for display

### Test plan

Before:

https://github.com/valora-inc/wallet/assets/20150449/5ac66014-5dc1-45e9-9ab3-8ba9e12d6e54

After, app start without any cached tx's:

https://github.com/valora-inc/wallet/assets/20150449/10497e11-c806-41ec-bba6-7423d7a4d532

After, receiving 2 transactions on different networks. Note the timestamps of the actions dispatched (only once, when the transaction is returned for the first time).

https://github.com/valora-inc/wallet/assets/20150449/ff696f46-62ef-40f6-b7b9-5db4f99eff7a

### Related issues

n/a

### Backwards compatibility

Y


src/identity/commentEncryption.test.ts
```diff
@@ -237,7 +237,10 @@ describe(checkTxsForIdentityMetadata, () => {
     const mockAccountsWrapper = {
       getWalletAddress: jest.fn((address) => address),
     }
-    await expectSaga(checkTxsForIdentityMetadata, updateTransactions(transactions))
+    await expectSaga(
+      checkTxsForIdentityMetadata,
```
kathaypacific, line 241: we can probably skip this saga for non Celo networks...right?

jeanregisser: Right 💯

jeanregisser (commented):
Sorry before looking more at the details, I have this question: we are we making a request per network?
Shouldn't we include the list of networks we want in a single request?

jeanregisser:
Answering myself, I looked at https://github.com/valora-inc/wallet/blob/500dc0941e0f2a937d2eeab56514c13cefd5aea5/src/transactions/feed/queryHelper.ts#L110-L165

I'm a bit concerned by the approach and the impact it can have with pagination.
It seems we can get (temporary) gaps in the feed because of that, when pulling transactions.
I mean we would have similar problems server side, but could hide this complexity there.

Thinking of this problem:
User transacts on both network A and B, however only transacts on network B seldomly.
when we query both A and B for the first page, we could get something like this:
- A: [yesterday, last week, last week - 1]
- B: [last year, last year - 1]

And we would simply combine them in our feed, putting all of B's TXs at the end of the feed. Since we order in reverse chronological order.
Then when we scroll to the bottom, we'd get more of A and more of B, pushing B's TXs again at the end.
This doesn't sound great.

Maybe @jophish @cajubelt you thought about this?

bakoushin (approved):
Looks great! 

I left some micro-optimization/naming nits

kathaypacific:
> Answering myself, I looked at
> 
> https://github.com/valora-inc/wallet/blob/500dc0941e0f2a937d2eeab56514c13cefd5aea5/src/transactions/feed/queryHelper.ts#L110-L165
> 
> I'm a bit concerned by the approach and the impact it can have with pagination. It seems we can get (temporary) gaps in the feed because of that, when pulling transactions. I mean we would have similar problems server side, but could hide this complexity there.
> 
> Thinking of this problem: User transacts on both network A and B, however only transacts on network B seldomly. when we query both A and B for the first page, we could get something like this:
> 
> * A: [yesterday, last week, last week - 1]
> * B: [last year, last year - 1]
> 
> And we would simply combine them in our feed, putting all of B's TXs at the end of the feed. Since we order in reverse chronological order. Then when we scroll to the bottom, we'd get more of A and more of B, pushing B's TXs again at the end. This doesn't sound great.
> 
> Maybe @jophish @cajubelt you thought about this?

are you proposing that we revisit the way the wallet is calling blockchain-api @jeanregisser ? if not, i think this PR is still needed either way right? we can address this problem scenario you raised on blockchain-api side somehow (e.g. return pages that are of similar time ranges per network? i'm not sure how easy this is / if it is possible though)

src/transactions/reducer.ts
```diff
@@ -121,7 +126,10 @@ export const reducer = (
 
       return {
         ...state,
-        transactions: action.transactions,
+        transactions: {
+          ...state.transactions,
+          [action.networkId]: action.transactions,
```
cajubelt, line 131: seems like an alternative to this whole refactor would have been to only overwrite the values with the same network id and leave the rest. I don't think this is more expensive because you're already copying the previous state which is linear complexity (same as a filter)

cajubelt (approved):
thanks for fixing. note my comment on an alternative. agree with Alex's comments.

src/transactions/reducer.ts
kathaypacific: oh interesting....can you expand on the benefits of that approach over this one? i think i'm open to it, but i'm not sure that i see the advantage aside from less code changes. one thing i like about the current approach is that we keep the "cached" data on the client the same as what it is on the server side (easier to debug, also easier visualise what changed in flipper - reordering the whole transactions array makes seeing the actual changes really difficult)

test/schemas.ts
```diff
@@ -2755,6 +2755,20 @@ export const v163Schema = {
...
+export const v164Schema = {
+  ...v163Schema,
+  _persist: {
+    ...v163Schema._persist,
+    version: 164,
+  },
+  transactions: {
+    ...v163Schema.transactions,
+    transactions: {
+      [networkConfig.defaultNetworkId]: v163Schema.transactions.transactions,
```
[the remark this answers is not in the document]
kathaypacific: this was a shortcut 😅 for the schemas test we only had celo transactions. this is only used for tests i believe - the "real" stuff is in the migrations and i added a more detailed test there

src/transactions/reducer.ts
```diff
@@ -179,10 +187,29 @@ export const knownFeedTransactionsSelector = (state: RootState) =>
 export const recentTxRecipientsCacheSelector = (state: RootState) =>
   state.transactions.recentTxRecipientsCache
 
-export const transactionsSelector = (state: RootState) => state.transactions.transactions
+export const transactionsByNetworkSelector = (state: RootState) => state.transactions.transactions
```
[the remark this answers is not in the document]
kathaypacific: TY, this makes sense! i've updated

src/transactions/feed/queryHelper.ts
```diff
@@ -149,14 +149,15 @@ export function useFetchTransactions(): QueryHookResult {
           )
           // Compare the new tx hashes with the ones we already have in redux
           for (let i = 0; i < nonEmptyTransactions.length; i++) {
-            if (!transactionHashes.includes(nonEmptyTransactions[i].transactionHash)) {
+            const knownTransactionHashes = transactionHashesByNetwork[networkId] ?? []
+            if (!knownTransactionHashes?.includes(nonEmptyTransactions[i].transactionHash)) {
```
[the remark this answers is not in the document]
kathaypacific: TY for the great suggestions, i've added them all (although i don't think a type guard is needed, `transactionHashesByNetworkIdSelector` will always return an object even if it has no entries so seems safe to look up the network id on it without a check)

src/transactions/feed/queryHelper.ts
```diff
@@ -147,16 +147,17 @@ export function useFetchTransactions(): QueryHookResult {
           const nonEmptyTransactions = returnedTransactions.filter(
             (returnedTransaction) => !isEmpty(returnedTransaction)
           )
+          const knownTransactionHashes = transactionHashesByNetwork[networkId]
           // Compare the new tx hashes with the ones we already have in redux
-          for (let i = 0; i < nonEmptyTransactions.length; i++) {
-            if (!transactionHashes.includes(nonEmptyTransactions[i].transactionHash)) {
+          for (const tx of nonEmptyTransactions) {
+            if (!knownTransactionHashes || knownTransactionHashes.has(tx.transactionHash)) {
```
bakoushin, line 153: Should we check if tx hash in _not_ in known hashes?

```suggestion
            if (!knownTransactionHashes || !knownTransactionHashes.has(tx.transactionHash)) {
```

src/transactions/feed/queryHelper.ts
```diff
@@ -101,7 +101,6 @@ export function useFetchTransactions(): QueryHookResult {
   })
 
   const [fetchingMoreTransactions, setFetchingMoreTransactions] = useState(false)
-  let hasNewTransaction = false
```
kathaypacific, line 104: this variable shouldn't be global, we should decide if each individual response from blockchain-api has new transactions that need to be cached. so moved to the relevant scope

src/transactions/feed/queryHelper.ts
kathaypacific: lol yup, updated

bakoushin:
🚀

src/transactions/reducer.ts
cajubelt: I approved anyway because I don't think the benefit is large. 

One upside I can think of is that it keeps state more similar to what the user ends up seeing in their feed. It also limits the scope of the changes, like you mentioned, and might mean a bit less code to maintain. Given that we don't have a use case right now of showing transactions separately per blockchain, it doesn't seem necessary to store them separately, even if we do query for them separately. Given a choice between "matching what the user sees" and "matching how we fetch the data" I tend to opt for the former when choosing how to represent and store data. 

All that being said, you did it this way, and I think the difference between both options are minor enough to make it author's choice. Just wanted to voice the alternative while I was noticing it

kathaypacific: yes that makes sense, thanks for elaborating! i agree generally storing data in the format that the client needs to use it is the best option. with selectors we can get the best of both worlds here and hide the complexity in there before it reaches the display components while minimising the amount of data processing that goes into the cached data

PR #4389, opened 2023-10-26
fix: add confirmed state for swap standby transaction

kathaypacific:
### Description

As the title.

It's not nice that we're dispatching 2 consecutive actions here, but this will change in the future.

### Test plan

Swap transactions should be already confirmed before the next blockchain-api poll.

https://github.com/valora-inc/wallet/assets/20150449/6ab8e4b5-dba8-444b-ae25-68189cda0c42

### Related issues

- Fixes RET-894

### Backwards compatibility

Y

### DIFF
--- a/src/swap/saga.test.ts
+++ b/src/swap/saga.test.ts
@@ -41,7 +41,7 @@ const contractKit = {
 }
 
 jest.mock('src/transactions/send', () => ({
-  sendTransaction: jest.fn(() => ({ transactionHash: '0x123' })),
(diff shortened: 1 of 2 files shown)


jeanregisser (approved):
Good, in the future we'll listen to the status so we don't need to do this individually?

I'll apply the same to https://github.com/valora-inc/wallet/pull/4368

kathaypacific:
> Good, in the future we'll listen to the status so we don't need to do this individually?
> 
> I'll apply the same to #4368

yes i'm hoping that @dievazqu can create a saga that will handle fetching the receipts for all pending transactions with the failed tx's work

PR #4390, opened 2023-10-26
chore: remove redundant add tx hash to standby tx action

kathaypacific:
### Description

As the title https://valora-app.slack.com/archives/C02898GN22V/p1698310629394759?thread_ts=1698304008.767299&cid=C02898GN22V

### Test plan

n/a

### Related issues

RET-894

### Backwards compatibility

Y


jeanregisser (approved):
Awesome! 🚀

src/viem/saga.ts
```diff
@@ -282,7 +281,6 @@ export function* sendAndMonitorTransaction({
       ...commonTxAnalyticsProps,
       txHash: hash,
     })
-    yield* put(addHashToStandbyTransaction(context.id, hash))
```
jeanregisser, line 285: Ah damn this was a valid use case for it. 
Until we refactor things.
Sorry I read that too fast.

PR #4394, opened 2023-10-27
feat(send): add validation for send amount screen

satish-ravi:
### Description

Adds the error states for send enter amount screen. The prepare transaction stuff is mocked based on the upcoming implementation in #4378 

### Test plan

Unit tests, manual

| Above balance | Max amount | No balance for gas |
|--------|--------|--------|
|  |  |  | 

### Related issues

- Fixes ACT-943

### Backwards compatibility

Yes


MuckT:
@satish-ravi it seems like input doesn't handle comma's which are used for decimals in Europe. PR #3074 fixed a similar issue on the swaps screen. I also saw it attempted to convert the input amount to a number when only adding `.` resulting in a `NaN` displayed.

satish-ravi:
> @satish-ravi it seems like input doesn't handle comma's which are used for decimals in Europe. PR #3074 fixed a similar issue on the swaps screen. I also saw it attempted to convert the input amount to a number when only adding `.` resulting in a `NaN` displayed.

thanks @MuckT, addressed in a4bd260

MuckT (approved):
LGTM to me and the native keyboard is working great 👨‍🍳

PR #4395, opened 2023-10-27
fix: add hash to standby tx for viem send flow

kathaypacific:
### Description

As the title. Related to https://github.com/valora-inc/wallet/pull/4390/files#r1373425723

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y


jeanregisser (commented):
Thanks, we should put back the call in `src/transactions/sagas` too I think
https://github.com/valora-inc/wallet/pull/4390/files#diff-5cf428b03df54b744bb38d9098a2081b1d402d578427631eeb50909a6b0fb5eb

kathaypacific:
> Thanks, we should put back the call in `src/transactions/sagas` too I think https://github.com/valora-inc/wallet/pull/4390/files#diff-5cf428b03df54b744bb38d9098a2081b1d402d578427631eeb50909a6b0fb5eb

why is that? it seems like `transactionHash` and `receipt` will come back at the same time, so there is no time between dispatching the `addHashToStandbyTransaction` and `transactionConfirmed` actions right? am i missing something?

jeanregisser:
> > Thanks, we should put back the call in `src/transactions/sagas` too I think https://github.com/valora-inc/wallet/pull/4390/files#diff-5cf428b03df54b744bb38d9098a2081b1d402d578427631eeb50909a6b0fb5eb
> 
> why is that? it seems like `transactionHash` and `receipt` will come back at the same time, so there is no time between dispatching the `addHashToStandbyTransaction` and `transactionConfirmed` actions right? am i missing something?

There it's happening in a retry loop, so the hash could change if it estimates again (and some values change) when retrying.

kathaypacific:
> > > Thanks, we should put back the call in `src/transactions/sagas` too I think https://github.com/valora-inc/wallet/pull/4390/files#diff-5cf428b03df54b744bb38d9098a2081b1d402d578427631eeb50909a6b0fb5eb
> > 
> > 
> > why is that? it seems like `transactionHash` and `receipt` will come back at the same time, so there is no time between dispatching the `addHashToStandbyTransaction` and `transactionConfirmed` actions right? am i missing something?
> 
> There it's happening in a retry loop, so the hash could change if it estimates again (and some values change) when retrying.

sorry i'm walking through the code but still don't think i understand :( are you saying that the `transactionHash` and `receipt.transactionHash` returned by `sendTransactionPromises` can be different? i guess i'm not seeing where the hash could change, can you point me to that part?

jeanregisser:
No worries, I'm comparing to what was removed in https://github.com/valora-inc/wallet/pull/4390:

Screenshot 2023-10-27 at 15 15 31

The way it was done, `sendTransactionPromise` was inside `sendTxMethod`, along with `addHashToStandbyTransaction` so can be called multiple times, and the returned receipt could be different.
This is because `sendTransactionPromises` can potentially estimate things again and lead to a new hash.

In brief the whole PR should be reverted (except the fee fix, but maybe that's a reason to not include side fixes, as it makes it harder to revert single changes).

kathaypacific:
> No worries, I'm comparing to what was removed in #4390:
> 
> Screenshot 2023-10-27 at 15 15 31
> The way it was done, `sendTransactionPromise` was inside `sendTxMethod`, along with `addHashToStandbyTransaction` so can be called multiple times, and the returned receipt could be different. This is because `sendTransactionPromises` can potentially estimate things again and lead to a new hash.
> 
> In brief the whole PR should be reverted (except the fee fix, but maybe that's a reason to not include side fixes, as it makes it harder to revert single changes).

okay i reverted this change because it doesn't hurt but honestly i still don't understand what the difference between having this and not. it seems like `transactionConfirmed` would be dispatched right after and add the tx hash, and i don't see the scenario where not dispatching this `addHashToStandbyTransaction` would break something. maybe we can walk through together at our 1:1 next week.

jeanregisser:
Yes probably requires a video chat :D
What if `transactionConfirmed` is never dispatched because it throws before?

jeanregisser (approved):
👍

PR #4397, opened 2023-10-27
feat: switch to CeloScan

bakoushin:
### Description

Switching default block explorer to [CeloScan](https://celoscan.io):

* faster than [Celo Explorer](https://explorer.celo.org/)
* similar UI as Etherscan (which is used for Ethereum txs)

#### Exception for NFTs

Celo Explorer is kept for NFTs since CeloScan is currently missing a visual representation of an NFT:

* [on Celo Explorer](https://explorer.celo.org/mainnet/token/0x965FCdd4Eb4662863894613D719a8c034E116c3f/inventory)
* [on CeloScan](https://celoscan.io/token/0x965fcdd4eb4662863894613d719a8c034e116c3f#inventory)

### Test plan

* Tested manually on Alfajores (send/receive/swap)
* Updated unit tests

### Related issues

- Fixes RET-890

### Backwards compatibility

Y


src/web3/networkConfig.ts
```diff
@@ -71,7 +71,11 @@ interface NetworkConfig {
 }
 
 export type blockExplorerUrls = {
```
kathaypacific, line 73: could we make this Capital case for types? i think i commented on something similar before, sorry if i missed a reply there

src/web3/networkConfig.ts
```diff
@@ -180,11 +184,29 @@ const CELO_EXPLORER_BASE_ADDRESS_URL_MAINNET = `${CELO_EXPLORER_BASE_URL_MAINNET
...
+const ETHERSCAN_BASE_TX_URL_SEPOLIA = `${ETHERSCAN_BASE_URL_SEPOLIA}/tx/`
+const ETHERSCAN_BASE_TX_URL_MAINNET = `${ETHERSCAN_BASE_URL_MAINNET}/tx/`
 
-const ETH_EXPLORER_BASE_TX_URL_SEPOLIA = `${ETH_EXPLORER_BASE_URL_SEPOLIA}/tx/`
-const ETH_EXPLORER_BASE_TX_URL_MAINNET = `${ETH_EXPLORER_BASE_URL_MAINNET}/tx/`
+const ETHERSCAN_BASE_ADDRESS_URL_SEPOLIA = `${ETHERSCAN_BASE_URL_SEPOLIA}/address/`
+const ETHERSCAN_BASE_ADDRESS_URL_MAINNET = `${ETHERSCAN_BASE_URL_MAINNET}/address/`
+
+const ETHERSCAN_BASE_TOKEN_URL_SEPOLIA = `${ETHERSCAN_BASE_URL_SEPOLIA}/token/`
+const ETHERSCAN_BASE_TOKEN_URL_MAINNET = `${ETHERSCAN_BASE_URL_MAINNET}/token/`
```
kathaypacific, line 209: i'm wondering if there's any benefit to having these urls (excluding the actual base url) declared here and then used in `blockExplorerUrls`, would it be more readable to create these directly inside `blockExplorerUrls`?

src/web3/networkConfig.ts
```diff
@@ -180,11 +184,29 @@ const CELO_EXPLORER_BASE_ADDRESS_URL_MAINNET = `${CELO_EXPLORER_BASE_URL_MAINNET
 const CELO_EXPLORER_BASE_TOKEN_URL_ALFAJORES = `${CELO_EXPLORER_BASE_URL_ALFAJORES}/token/`
 const CELO_EXPLORER_BASE_TOKEN_URL_MAINNET = `${CELO_EXPLORER_BASE_URL_MAINNET}/token/`
```
kathaypacific, line 185: should we remove these and the blockscout related variables in the network config object? i think only NFT details still should go through Blockscout (although even that i'm doubting because of [this](https://valora-app.slack.com/archives/CL7BVQPHB/p1698335014761419?thread_ts=1696001827.200189&cid=CL7BVQPHB) - i recall @MuckT saying that CeloScan does not display NFT metadata as well as Blockscout but it might still be worth using CeloScan instead)

kathaypacific (commented):
Yay great!! I think there are still a few lingering variables that can be removed, including a reference to Blockscout in config.ts :)

src/web3/networkConfig.ts
bakoushin: Sure, thanks! The other one was fixed, but this one I missed.

src/web3/networkConfig.ts
bakoushin: I kept these variables only for NFTs, not really sure that switch to CeloScan brings better user experience over Blockscout for NFT users.

As far as I understood the slack discussion you've mentioned, another block explorer is cooked. Maybe it's too early to switch now?

kathaypacific: yes i'm also not sure what we want to do with the NFTs right now, so let's leave it. it seems like we can remove `celoExplorerBaseTxUrl` and `celoExplorerBaseAddressUrl` though? and also `BLOCKSCOUT_BASE_URL` from config.ts?

kathaypacific (approved):
LGTM!

PR #4399, opened 2023-10-27
fix: face verification using internal web browser

kathaypacific:
### Description

The face verification on GoodDollar does not currently work using the internal web browser on Android. It seems like to enable using the camera using react-native-webview, we need to set `mediaPlaybackRequiresUserAction={false}`. However, the [documentation](https://github.com/react-native-webview/react-native-webview/blob/fd61a320579fda6b57989a81460e03ec2d59a7ca/docs/Reference.md?plain=1#L292) for `mediaPlaybackRequiresUserAction` says it is a ```Boolean that determines whether HTML5 audio and video requires the user to tap them before they start playing. The default value is `true`.``` I hesitate to make this a global change because generally on mobile you don't want audio and video to auto play...

I ended up creating a dynamic config for this so that we can specify which urls and OS's we will allow this for.

### Test plan

Face verification should work (although specifically for Valora's side, the camera works for face verification)

Before:

https://github.com/valora-inc/wallet/assets/20150449/22ee5473-a6e8-4024-af2c-a714868b78fe

After:

https://github.com/valora-inc/wallet/assets/20150449/bc2d7a7f-5f3b-47d8-a940-044ebd870035

### Related issues

- Fixes RET-820

### Backwards compatibility

Y


src/statsig/constants.ts
```diff
@@ -102,4 +102,15 @@ export const DynamicConfigs = {
...
     },
   },
+  [StatsigDynamicConfigs.DAPP_WEBVIEW_CONFIG]: {
+    configName: StatsigDynamicConfigs.DAPP_WEBVIEW_CONFIG,
+    defaultValues: {
+      ios: {
+        disableMediaPlaybackRequiresUserAction: [] as string[],
+      },
+      android: {
+        disableMediaPlaybackRequiresUserAction: [] as string[],
```
jeanregisser, line 112: Could we control ios vs android directly in statsig instead?
So we have a simple config to read.

src/webview/WebViewScreen.tsx
```diff
@@ -196,6 +203,7 @@ function WebViewScreen({ route, navigation }: Props) {
           setCanGoForward(navState.canGoForward)
           handleSetNavigationTitle(navState.url, navState.title, navState.loading)
         }}
+        mediaPlaybackRequiresUserAction={!disableMediaPlaybackRequiresUserActionUrls.includes(uri)}
```
jeanregisser, line 206: Maybe we should only check the origin?

In that case I'd recommend we name the config `disabledMediaPlaybackRequiresUserActionOrigins` or `originsForDisableMediaPlaybackRequiresUserAction`

jeanregisser (commented):
It's interesting the WebView doesn't follow the default of the browser.
Anyway it seems good that we make it a really special case 👍

src/statsig/constants.ts
kathaypacific: ah, TY for suggesting this. i forgot we could do this 😅 updated

kathaypacific: the config is here https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/dynamic_configs/dapp_webview_config

src/webview/WebViewScreen.tsx
kathaypacific: yes that's a good idea, i've updated

jeanregisser (approved):
🚀

PR #4400, opened 2023-10-27
fix: app crash on swap confirmed tx item tap

kathaypacific:
### Description

I think this fix got lost in some merge conflicts, it was introduced in https://github.com/valora-inc/wallet/pull/4390/files#diff-3ade5aeec77d2dbc8ab35eab85cf5c3063663ee12d23d67909ca5871c312c739R80.

### Test plan

Make a swap -> go to the home screen and tap on the new homefeed item. The app should not crash and should display the tx details screen with no fees.

### Related issues

n/a

### Backwards compatibility

Y


jeanregisser (approved):
👍

PR #4401, opened 2023-10-27
feat: remove default from token in swap screen

kathaypacific:
### Description

As the title.

Most of the changes were in the tests - i created a helper to do the token selection rather than copy + paste the same 3 lines of code everywhere (since tests all need to firstly choose a default token now)

### Test plan

Simulator Screenshot - iPhone 14 Pro - 2023-10-27 at 17 25 49

### Related issues

Part of RET-865

### Backwards compatibility

Y


dievazqu (approved):

bakoushin (approved):
🎯

PR #4403, opened 2023-10-27
fix(prepare-transactions): more precise handling of gas estimation errors

cajubelt:
### Description

handles concern raised by @jeanregisser [here](https://github.com/valora-inc/wallet/pull/4378#discussion_r1374351997)

To summarize, we needed to avoid swallowing errors that had nothing to do with insufficient funds for gas under the assumption that they were stemming from insufficient funds to pay for gas. The case Jean noticed was that if we forgot to include a "gas" field in transactions that depend on some prior, not-yet-submitted transaction (like a swap approval), we could get a gas estimation error here that would result in a nonsensical output to the user-- specifically, an indicator that they need more gas to perform a swap (even though that has no relation with the real source of the error). 

### Test plan

- jest tests. unit and integration. includes 2 regression tests
- simulated insufficient gas errors and approval errors manually and verified they are differentiated by this check

### Related issues

na

### Backwards compatibility

na


satish-ravi (approved):

jeanregisser (approved):
Perfect! 💯 

Thank you 🙌

PR #4404, opened 2023-10-27
fix(text-input): Allow overriding default font

jophish:
### Description

Needed in order to get correct styling in #4386. [Context](https://github.com/valora-inc/wallet/pull/4386#discussion_r1373301452). The snapshot updates don't change the appearance of those components

### Related issues

- Fixes #[issue number here]


src/components/TextInput.tsx
```diff
@@ -75,9 +75,9 @@ export class CTextInput extends React.Component<Props, State> {
         <RNTextInput
           ref={forwardedRef}
           style={[
+            inputStyle ?? fontStyles.regular,
```
MuckT, line 78: If the inputStyle contains `paddingVertical`, `paddingHorizontal`, `lineHeight`, `flex` value it looks like it'll be overwritten by the later `styles.input`.  I am able to overwrite the styles with the previous code if I pass `inputStyle={fontStyles.xsmall500}` or `inputStyle={{...fontStyles.xsmall500}}` to a `<TextInput ...>` such as the ones seen in `src/account/SupportContact.tsx`

PR #4405, opened 2023-10-27
fix(swaps): pending failed swaps never removed from home feed

bakoushin:
### Description

As the title goes. If swap is failed, it still live in home feed in pending status for ever.
Context: https://valora-app.slack.com/archives/C04B61SJ6DS/p1698428109804219?thread_ts=1698426762.876189&cid=C04B61SJ6DS

This fix ensures that pending swap is removed from feed if swap is failed. 

### Reproduction

On Mainnet ensure you don't have enough fee currency and execute the swap, so it fails.

### Test plan

* Tested manually on Mainnet.
* Updated unit test

### Backwards compatibility

Y


MuckT (approved):
👍

jeanregisser (approved):

kathaypacific (approved):
TY for fixing this!! i 🙏🏻

PR #4406, opened 2023-10-27
fix(deps): downgrade WalletConnect

MuckT:
### Description

It appears from [Sentry](https://valora-inc.sentry.io/discover/results/?cursor=0%3A250%3A0&field=title&field=release&field=environment&field=user.display&field=timestamp&name=Error%3A+Unable+to+find+WalletConnect+session+matching+topic+dfe8ce6bc3a89058f7a292b37d934e6f2bbe944a9bc15...&project=1250733&query=issue%3ACELO-MOBILE-EJR&sort=-timestamp&statsPeriod=90d&yAxis=count%28%29) that since v1.68.0 we've had increasing numbers of users hit issues with Wallet connect. This PR reverts upgrades to those dependencies to versions previously used.

### Test plan

TBD

### Related issues

N/A

### Backwards compatibility

Yes


jeanregisser (commented):
Thanks!
This downgrade sounds ok for the short term. Though it is worrying if we can't upgrade with confidence.

We should chat about what we should do longer term.

yarn.lock
```diff
@@ -4962,29 +4962,51 @@
...
+    "@walletconnect/utils" "^2.9.0"
     events "^3.3.0"
     isomorphic-unfetch "^3.1.0"
 
-"@walletconnect/core@2.10.2", "@walletconnect/core@^2.10.1":
-  version "2.10.2"
-  resolved "https://registry.yarnpkg.com/@walletconnect/core/-/core-2.10.2.tgz#a1bf6e3e87b33f9df795ce0970d8ddd400fdc8a3"
-  integrity sha512-JQz/xp3SLEpTeRQctdck2ugSBVEpMxoSE+lFi2voJkZop1hv6P+uqr6E4PzjFluAjeAnKlT1xvra0aFWjPWVcw==
+"@walletconnect/core@2.10.0":
+  version "2.10.0"
```
jeanregisser, line 4985: Can we have one single version of this package?

yarn.lock
```diff
@@ -5157,10 +5179,22 @@
...
+  dependencies:
+    "@walletconnect/events" "^1.0.1"
+    "@walletconnect/heartbeat" "1.2.1"
+    "@walletconnect/jsonrpc-types" "1.0.3"
+    "@walletconnect/keyvaluestorage" "^1.0.2"
+    "@walletconnect/logger" "^2.0.1"
+    events "^3.3.0"
+
+"@walletconnect/utils@2.10.0":
+  version "2.10.0"
```
jeanregisser, line 5195: Same for this one?

MuckT: I was able to resolve this modifying the lock file directly, but I am not sure that's ideal. I believe @kathaypacific had to do something similar in the past with the WalletConnect dependencies.

jeanregisser (approved):

MuckT:
@jeanregisser I wonder if we should clear out the sessions and pending requests in a redux migration.

PR #4407, opened 2023-10-30
fix: prevent pending transaction from being stuck

kathaypacific:
### Description

In addition to #4405, I think there are still some scenarios where a user could have a stuck pending transaction.
1. in the contractKit flows, if the transaction receipt has false status (transaction failed) then no state transaction happens (the tx remains in pending state)
2. if for whatever reason a transaction hash is not added to a standby transaction, we cannot really do anything with them in in subsequent app sessions as we lose the execution context. i think even if we fetched all pending tx's for a wallet, it'd be difficult to match the pending tx to the standby tx. so we should just remove them in the next app session.

### Related issues

- Fixes #[issue number here]


kathaypacific:
after some more investigation i don't think this PR is needed with the current implementation.
1. it seems like we throw an error if a transaction is confirmed as failed, so no failed transactions come through the `transactionConfirmed` reducer.
2. i think this is so edge casey that it's probably not needed and can wait for a full address with the failed transactions work.

PR #4408, opened 2023-10-30
feat(transactions): Watch pending transactions

dievazqu:
### Description

This PR adds a saga that "watches" periodically for the pending standby transactions, in order to find if the transaction has been reverted or it was successful.

#### Considerations
* This only works for pending transactions with a transaction hash, we have some flows that create a standby transaction without a hash, those transactions aren't watchable with this PR.
* Also, there are some flows that remove the standby transaction when catching an error. We have to modify those flows if we want to show the transaction in the feed.

### Test plan

Tested in a emulator where I manually tweak the code to create standby transactions with a transaction hash of a failed transaction.
Unit tests.

### Related issues

- Fixes RET-896

### Backwards compatibility

Yes

### Screen recording

https://github.com/valora-inc/wallet/assets/12504748/4c0a385d-d3c3-4354-93bf-5316da760d02


src/icons/AlertWithBackground.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react'
```
kathaypacific, line 1: instead of adding a new icon, can we use the existing Alert icon and draw the pink background with css? i feel like our icons folder is becoming massive and difficult to navigate, also it seems reasonable to assume that if the Alert icon needs to be updated then we'd also expect that change to affect the homefeed icon so i was thinking they should be the same component?

src/send/saga.ts
```diff
@@ -151,6 +151,7 @@ export function* buildAndSendPayment(
     feeInfo
   )
 
+  // TODO: We need to add the transaction hash in order to be able to watch if the transaction fails.
```
kathaypacific, line 154: how should this TODO be addressed? i think it's not possible to get the tx hash ahead of time sometimes, which is why i think we should remove any pending tx's that don't have tx hashes at the start of new app sessions

src/swap/saga.ts
```diff
@@ -68,6 +68,7 @@ function* handleSendSwapTransaction(
   const tx: CeloTx = yield* call(normalizer.populate.bind(normalizer), rawTx)
   const txo = buildTxo(kit, tx)
 
+  // TODO: We need to add the transaction hash in order to be able to watch if the transaction fails.
```
kathaypacific, line 71: same as my other comment

src/transactions/feed/TransactionFeed.tsx
```diff
@@ -33,7 +34,12 @@ function TransactionFeed() {
 
   const cachedTransactions = useSelector(transactionsSelector)
   const allPendingTransactions = useSelector(pendingStandbyTransactionsSelector)
-  const allConfirmedStandbyTransactions = useSelector(completedStandbyTransactionsSelector)
+  const allFailedTransactions = useSelector(failedStandbyTransactionsSelector)
+  const allCompletedStandbyTransactions = useSelector(completedStandbyTransactionsSelector)
+  const allConfirmedStandbyTransactions = [
```
kathaypacific, line 39: it seems like we want to treat successful and failed completed transactions the same right? so is there any need to create a separate failed tx's selector? (should we just add failed tx's to the `completedStandbyTransactionsSelector`?)

src/transactions/saga.ts
```diff
@@ -222,7 +227,65 @@ function* watchAddressToE164PhoneNumberUpdate() {
...
+      `Found error when trying to fetch status for transaction with hash: ${transactionHash}`,
+      e
+    )
+    return
+  }
+}
+
+export function* watchPendingTransactions() {
+  while (true) {
+    const pendingStandbyTransactions: StandbyTransaction[] = yield* select(
```
kathaypacific, line 277: nit: we don't need to add the type here, it should already be inferred by the `select`

src/transactions/saga.ts
```diff
@@ -222,7 +227,65 @@ function* watchAddressToE164PhoneNumberUpdate() {
...
+      watchablePendingTransactionSelector
```
kathaypacific, line 278: i'm wondering if we can just use the normal pending tx's selector, since we check for the presence of the tx hash in the watchPendingTransaction saga anyway?

src/transactions/saga.ts
```diff
@@ -222,7 +227,65 @@ function* watchAddressToE164PhoneNumberUpdate() {
...
+    )
+    for (const transaction of pendingStandbyTransactions) {
+      yield* fork(watchPendingTransaction, transaction)
```
kathaypacific, line 281: nit: we're not really watching the pending tx here, we're just getting the tx receipt right? maybe `getTransactionReceipt` will do?

src/transactions/saga.ts
```diff
@@ -222,7 +227,65 @@ function* watchAddressToE164PhoneNumberUpdate() {
...
+
+  if (!transactionHash) {
+    Logger.warn(TAG, `Transaction ${transaction.context.id} has no hash, skipping`)
+    return
+  }
+
+  // TODO: Remove this constraint when our viem client supports other networks
+  const supportedNetworkdIds = [NetworkId['celo-mainnet'], NetworkId['celo-alfajores']]
+  if (!supportedNetworkdIds.includes(networkId)) {
+    Logger.warn(TAG, `Transaction ${transaction.context.id} is not on the celo network, skipping`)
```
kathaypacific, line 241: can we log the `networkId` here too?

src/transactions/types.ts
```diff
@@ -34,10 +34,20 @@ export type CompletedStandbyTransaction = (
   context: TransactionContext
 }
 
+export type FailedStandbyTransaction = (
```
kathaypacific, line 37:
```suggestion
export type FailedStandbyTransaction = (
  | Omit<PendingStandbyTransfer, 'status'>
  | Omit<PendingStandbySwap, 'status'>
) & {
  status: TransactionStatus.Failed
}
```
code golf

src/viem/saga.ts
```diff
@@ -91,6 +91,7 @@ export function* sendPayment({
     // unlock account before executing tx
     yield* call(unlockAccount, wallet.account.address)
 
+    // TODO: We need to add the transaction hash in order to be able to watch if the transaction fails.
```
kathaypacific, line 94: do we need this comment? adding the tx hash was reinstated in #4395

src/transactions/reducer.ts
```diff
@@ -79,6 +80,24 @@ export const reducer = (
...
+      const { txId } = action
+
+      return {
+        ...state,
+        standbyTransactions: state.standbyTransactions.map(
+          (standbyTransaction): StandbyTransaction => {
+            if (standbyTransaction.context.id === txId) {
+              return {
+                ...standbyTransaction,
+                status: TransactionStatus.Failed,
```
kathaypacific, line 93: we should also update the timestamp here

kathaypacific (commented):
looking good :) it wasn't mentioned in the PR description but will there be follow up PRs? a few things that i can think of that aren't addressed yet in this PR:
- displaying failed transactions when the send / swap sagas's fail and removeStandbyTransaction is dispatched (you mention this in the description, but we do need to address it)
- the "failed transaction" subtitle on the homefeed item (i think currently for a failed send transaction, it might still say "payment sent")
- adding support for ethereum once #4306 is merged

i'm also wondering if 10 seconds is a good polling interval for slower chains, or if we should have a mapping that allows us to change this per network?

src/transactions/saga.ts
dievazqu: Hmm, what would you mean by "watching"? `getTransactionReceipt` sounds like it should return the receipt of the tx without any side effects, but that's not the case.
I'll change the name anyways, but I'm not sure if it makes a difference

src/transactions/reducer.ts
```diff
@@ -137,6 +156,15 @@ export const reducer = (
 export const standbyTransactionsSelector = (state: RootState) =>
   state.transactions.standbyTransactions
 
+export const watchablePendingTransactionSelector = createSelector(
+  [standbyTransactionsSelector],
```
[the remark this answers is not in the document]
dievazqu: Sorry, I don't want to use `pendingStandbyTransactionsSelector` because it fills the `transactionHash` with `''`.
I'm gonna create a new type for the "watchable" transactions.

src/send/saga.ts
dievazqu: This is a good question. I'm not sure how, but as far as I understand the transaction hash can be obtained before sending the transaction to the blockchain.
I think we should do the effort / investigation to understand if this is feasible in all cases, and if so, we should force the pending transactions to have a transaction hash.

dievazqu:
@kathaypacific Thanks for the review! The PR is ready for another review!

Regarding the things that aren't addressed in this PR:

> displaying failed transactions when the send / swap sagas's fail and removeStandbyTransaction is dispatched (you mention this in the description, but we do need to address it)

Yes, I guess I left that because I have some questions...
We create the standby transaction before sending it (and without the transaction hash in many cases), when the transaction fails we have 2 scenarios:
* The transaction fails before even reaching the blockchain (the client fails when sending the transaction), in this case there is no failed transaction in the blockchain, so I guess we should remove the standby transaction from the feed?
* The transaction was sent to the blockchain, but it was reverted there. In this case, we should get the transaction hash and start watching it until we get the status. This case is complicated to replicate and test, since the client has to run the transaction ok but the problem must happen after the transaction was sent.

> adding support for ethereum once https://github.com/valora-inc/wallet/pull/4306 is merged

Should I branch from that PR to add support for ethereum? Or just wait until that PR is merged?
The send and swap flows in celo are synchronic, so it was really painful to test this functionality. Using ethereum flows should be easier to test.

> i'm also wondering if 10 seconds is a good polling interval for slower chains, or if we should have a mapping that allows us to change this per network?

Yeah, that's a good point. 10 seconds sounded enough for this PR, but when we add support for new chains we can implement some mapping 👌

dievazqu:
Finally, I decided to use this standard:

We should add the transaction hash to the standby tx as soon as we send the tx to the blockchain.
So, if there is an error, we should check if the standby tx has a hash, if it doesn't have a hash, it means the error happened in our client, so the transaction never reached the blockchain, we should remove it from the feed. If it has a hash, we should watch it.

One thing that we still have to address:
I think it's possible for a transaction to have a hash but never reach the blockchain (for several reasons, one for example if the gas cost is low so no miner is picking the tx), so it might be possible to have a tx forever in the standby txs list which is very inefficient. Maybe we should also add a time constraint, something like, only watch txs that happened in the last 2 hours. What do you think?

src/transactions/saga.ts
```diff
@@ -225,7 +234,60 @@ function* watchAddressToE164PhoneNumberUpdate() {
   )
 }
 
+function* getTransactionReceipt(
```
kathaypacific, line 237: could we add a test for this? 🙏🏻

src/transactions/saga.ts
```diff
@@ -225,7 +234,60 @@ function* watchAddressToE164PhoneNumberUpdate() {
...
+      yield* fork(getTransactionReceipt, transaction, network)
+    }
+
+    yield* delay(WATCHING_DELAY_BY_NETWORK[network])
+  }
+}
+
+export function* watchPendingTransactions() {
+  const supportedNetworks = Object.keys(publicClient) as Network[]
+  for (const network of supportedNetworks) {
```
kathaypacific, line 284: could we reduce the number of layers / spawned sagas? watchPendingTransactions could handle all networks (like it can select all pending tx's that have a hash, and for each, spawn the `getTransactionReceipt` saga with the network fromthe transaction itself?

src/transactions/saga.ts
```diff
@@ -157,8 +167,7 @@ export function* sendAndMonitorTransaction<T>(
     return { receipt: txReceipt }
   } catch (error) {
     Logger.error(TAG + '@sendAndMonitorTransaction', `Error sending tx ${context.id}`, error)
-    yield* put(removeStandbyTransaction(context.id))
-    yield* put(transactionFailed(context.id))
+    yield* put(removeStandbyTransactionWithoutHash(context.id))
```
kathaypacific, line 170: hmm i think we need to distinguish between the 2 scenarios you mention - 
1. if the transaction has been mined and failed (i.e. the user has paid some gas fee) then we should show it on the tx feed. the intention is to explain to the user why their balance changed. This is less important on Celo since the gas fees are negligible, but is important for Ethereum.
2. if an error was thrown before the transaction made it to the blockchain, then we should remove the standby transaction since a user has not incurred any losses. in that case we should probably display an error banner or some kind of error, but we can add that later.

from that perspective, i don't think this is the correct change. it is indeed not a straightforward change to meet those 2 conditions so i'd recommend reverting this change and tackling it in a follow up PR?

src/transactions/transferFeedUtils.ts
```diff
@@ -216,6 +216,10 @@ export function useTransferFeedDetails(transfer: FeedTokenTransfer) {
     subtitle = t('confirmingTransaction')
   }
 
+  if (transfer.status === TransactionStatus.Failed) {
```
kathaypacific, line 219: maybe we can also add a test for this?

src/transactions/types.ts
```diff
@@ -34,11 +34,24 @@ export type CompletedStandbyTransaction = (
...
+  context: TransactionContext
+}
+
 export type StandbyTransaction =
+  | FailedStandbyTransaction
   | PendingStandbySwap
   | PendingStandbyTransfer
   | CompletedStandbyTransaction
 
+export type BaseStandbyTransaction =
```
kathaypacific, line 51: any reason for moving this here? i had it originally in the actions file because it is a type that is used specifically only for actions (and not intended as a useful type elsewhere). we can also rename it to be like `BaseStandbyTransactionActionPayload` or something if you want to keep it here

kathaypacific (commented):
looking good, i tried to answer some of your questions in my comments but let's chat over slack or at our 1:1 if anything is unclear! :)

src/transactions/saga.ts
kathaypacific: i was thinking of "watching" like it is actively watching / waiting / listening for the transaction receipt. but if i understand correctly, we only use the `getTransactionReceipt` RPC method to see if one exists. this method doesn't wait until one exists, it just makes a fetch and acts accordingly to the result.

src/send/saga.ts
kathaypacific: it's my understanding that for the swap flow using viem, we have the tx hash before sending the tx because we prepare the transactions upfront. but for the contract kit flows and the viem send flow, we do not. i'm not sure that it's possible to get this with contractkit upfront (and tbh not sure that it is worth the effort since we want to move away from it - so it seems like timing wise, it's not going to be worth investigating this topic for this task?)

src/transactions/types.ts
dievazqu: Sorry, I just felt this was a better place, but I'm fine rolling back this change

src/transactions/feed/detailContent/SwapContent.tsx
```diff
@@ -60,7 +60,7 @@ export default function SwapContent({ exchange }: Props) {
           )} ${toTokenSymbol}`}
         </Text>
       </View>
-      {exchange.fees[0] && (
+      {exchange.fees && exchange.fees[0] && (
```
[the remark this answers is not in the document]
dievazqu: Not really, now I think all transactions have the `fees` set. But I was thinking that adding this is harmless?
Anyways, I'm removing it :)

src/transactions/saga.ts
dievazqu: Hmm, if we want to have a delay configurable by chain, we need the flows to be separated, right?
What's the problem with creating a saga per chain for this?

src/transactions/saga.ts
dievazqu: I added some test for `watchPendingTransactions` which covers this method. Let me know if that's enough :)

src/transactions/reducer.ts
```diff
@@ -6,6 +6,7 @@ import { RootState } from 'src/redux/reducers'
 import { Actions, ActionTypes } from 'src/transactions/actions'
```
kathaypacific, line 6: can we also make the necessary changes to persist standby transactions across app sessions? https://github.com/valora-inc/wallet/blob/645f51785d23acd16e6d8e45ef09b2595d68d978/src/transactions/reducer.ts#L63-L70

src/transactions/saga.ts
```diff
@@ -225,7 +233,62 @@ function* watchAddressToE164PhoneNumberUpdate() {
...
+
+function* watchPendingTransactionsInNetwork(network: Network) {
+  while (true) {
+    yield* call(internalWatchPendingTransactionsInNetwork, network)
+    yield* delay(WATCHING_DELAY_BY_NETWORK[network])
+  }
+}
+
+function* watchPendingTransactions() {
+  const supportedNetworks = Object.keys(publicClient) as Network[]
```
kathaypacific, line 284: i started thinking about how this affects most users once shipped, we will be spawning a new saga every few seconds that checks for ethereum pending tx's when they're guaranteed to have none. the impact is probably small, but it seems like something we can already optimise for by collecting the enabled networks in the multichain dynamic config? since pending tx's are only enabled for send and swaps, we could already know which networks we need to use

kathaypacific (commented):
looking great, i just had a few small comments which i hope will be easy peasy :)

src/transactions/send.ts
```diff
@@ -249,6 +250,8 @@ export function* sendTransaction(
       gasPrice,
       nonce
     )
+    const hash: string = yield transactionHash
+    yield* put(addHashToStandbyTransaction(context.id, hash))
```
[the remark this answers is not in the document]
dievazqu: Hmm, yeah, I think you are right, we don't need this.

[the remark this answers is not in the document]
dievazqu: I guess, then we don't need this either: https://github.com/valora-inc/wallet/blob/68c0d6a1894c921876cef6cf3ed1ca1484bdfc90/src/transactions/saga.ts#L142

src/transactions/saga.ts
kathaypacific: yes! except we want the union, not the intersection :D

src/transactions/types.ts
```diff
@@ -34,7 +34,16 @@ export type CompletedStandbyTransaction = (
   context: TransactionContext
 }
 
+export type FailedStandbyTransaction = (
```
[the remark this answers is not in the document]
kathaypacific: lol naming is hard 😅 i agree, i think this is okay. it would be misleading if we had "success" in the name but completed and confirmed both seem correct to me

src/transactions/saga.ts
dievazqu: The union between send flow + swap flow, but the intersection with viem.
We can't process networks that viem doesn't support

src/transactions/send.ts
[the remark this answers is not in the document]
dievazqu: 🙈

dievazqu:
@kathaypacific Sorry for the long PR with a lot of comments! And thanks for all the time you put into reviewing this!
I really appreciate it!
I hope everything is ok now! 🙏

kathaypacific (approved):
yay great!! 🚀

PR #4409, opened 2023-10-30
fix: incorrect pending swap values

kathaypacific:
### Description

This PR fixes incorrect logic for pending swap tx values when the swap pair has different number of decimals. For example, the swap to value for a cEUR -> USDT on the pending transaction was too large by a factor of 12. 

### Test plan

Tested manually

### Related issues

Fixes RET-894

### Backwards compatibility

Y


jeanregisser (approved):
💯

PR #4412, opened 2023-10-30
fix: do not persist pending transactions across app sessions

kathaypacific:
### Description

As the title - we currently don't have a way to resolve these pending transactions across app sessions. 

### Test plan

n/a

### Related issues

- Fixes RET-894

### Backwards compatibility

Y

### DIFF
--- a/src/transactions/reducer.ts
+++ b/src/transactions/reducer.ts
@@ -55,9 +55,18 @@ export const reducer = (
 ): State => {
   switch (action.type) {
     case REHYDRATE: {
+      const persistedState: State = getRehydratePayload(action, 'transactions')
       return {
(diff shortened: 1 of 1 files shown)


jeanregisser (approved):
Sounds good! 👍

bakoushin (approved):
The only thing I'm concerned with that it could be hard to remember to get rid of it once peding txs can be resolved correctly.

kathaypacific:
> The only thing I'm concerned with that it could be hard to remember to get rid of it once peding txs can be resolved correctly.

yes indeed, i added an A/C to RET-896 to make sure we don't forget FYI @dievazqu

PR #4413, opened 2023-10-30
feat(balances): Add button to hide balances on home screen

finnian0826:
### Description

Adds a button to hide users balances on the home screen. When a user taps the little eye icon next to their total balance, the balance becomes `XX.XX`, the transaction feed balances are hidden, and the eye icon turns to a crossed out eye icon. When it is pressed again it goes back to normal.

### Test plan

https://github.com/valora-inc/wallet/assets/140328381/c2e3ce34-bd8f-432e-92f0-00a244a702dc

### Related issues

N/A

### Backwards compatibility

Yes


src/components/TokenBalance.tsx
```diff
@@ -87,20 +92,21 @@ function TokenBalance({
...
-            {localCurrencySymbol}
-            {totalTokenBalanceLocal?.toFormat(2) ?? '-'}
+            {!hideBalance && localCurrencySymbol}
+            {hideBalance ? 'XX.XX' : totalTokenBalanceLocal?.toFormat(2) ?? '-'}
           </Text>
           <Text style={styles.tokenBalance}>
-            {formatValueToDisplay(tokenBalance)} {tokensWithUsdValue[0].symbol}
+            {!hideBalance && formatValueToDisplay(tokenBalance)}{' '}
+            {!hideBalance && tokensWithUsdValue[0].symbol}
           </Text>
```
satish-ravi, line 101: nit: you can just skip rendering this Text component if hideBalances is true right?
```
{!hideBalance &&
(<Text style={styles.tokenBalance}>
  {formatValueToDisplay(tokenBalance)} {tokensWithUsdValue[0].symbol}
</Text>)
}
```

src/components/TokenBalance.tsx
```diff
@@ -203,7 +209,13 @@ export function AssetsTokenBalance({ showInfo }: { showInfo: boolean }) {
   )
 }
 
-export function HomeTokenBalance() {
+export function HomeTokenBalance({
+  buttonOnPress,
+  hideBalance = false,
+}: {
+  buttonOnPress?: () => void
```
satish-ravi, line 216: nit: rename to be more descriptive. maybe `eyeIconOnPress`?
Also should the props be required instead of optional? Seems like the eye icon is rendered always and there needs to be an onPress passed.

src/home/WalletHome.tsx
```diff
@@ -56,6 +58,12 @@ function WalletHome() {
   const showNotificationCenter = getFeatureGate(StatsigFeatureGates.SHOW_NOTIFICATION_CENTER)
   const showNotificationSpotlight = showNotificationCenter && canShowNotificationSpotlight
 
+  const [hideBalances, setHideBalances] = useState(false)
+  const eyeballOnPress = () => {
+    setHideBalances(!hideBalances)
+    ValoraAnalytics.track(HomeEvents.hide_balances)
```
satish-ravi, line 64: should there be separate props for hide / show? Or at least include it as a prop?

src/icons/EyeIcon.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react'
+import Svg, { Path } from 'react-native-svg'
+import Colors from 'src/styles/colors'
+
+export interface Props {
+  height?: number
```
satish-ravi, line 6: maybe call this size since this controls both height and width?

src/icons/HiddenEyeIcon.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react'
+import Svg, { ClipPath, Defs, G, Path } from 'react-native-svg'
+import Colors from 'src/styles/colors'
+
+export interface Props {
+  height?: number
```
satish-ravi, line 6: same comment as above

src/icons/EyeIcon.tsx
```diff
@@ -0,0 +1,27 @@
...
+  color?: Colors
+}
+
+function EyeIcon({ color, height }: Props) {
+  return (
+    <Svg height={height} width={height} fill="none" testID="EyeIcon">
```
satish-ravi, line 12: I think this would need a viewBox param set to `0 0 24 24`. Otherwise I don't think the `height` prop makes any difference. Can you test this out by passing a different height than 24?

src/icons/HiddenEyeIcon.tsx
```diff
@@ -0,0 +1,31 @@
...
+  color?: Colors
+}
+
+function EyeIcon({ color, height }: Props) {
+  return (
+    <Svg height={height} width={height} fill="none" testID="HiddenEyeIcon">
```
satish-ravi, line 12: same comment as above

src/home/WalletHome.tsx
```diff
@@ -56,6 +58,16 @@ function WalletHome() {
   const showNotificationCenter = getFeatureGate(StatsigFeatureGates.SHOW_NOTIFICATION_CENTER)
   const showNotificationSpotlight = showNotificationCenter && canShowNotificationSpotlight
 
+  const [hideBalances, setHideBalances] = useState(false)
+  const eyeIconOnPress = () => {
```
satish-ravi, line 62: can we get some test coverage for this? Should be fine to do some test with the HomeTokenBalance and TxFeed components to ensure balances are hidden / shown when the icon is clicked.

src/home/WalletHome.tsx
```diff
@@ -56,6 +58,16 @@ function WalletHome() {
...
+    setHideBalances(!hideBalances)
+    if (hideBalances) {
+      ValoraAnalytics.track(HomeEvents.hide_balances)
+    } else {
+      ValoraAnalytics.track(HomeEvents.show_balances)
+    }
```
satish-ravi, line 68:
```suggestion
    ValoraAnalytics.track(hideBalances ? HomeEvents.hide_balances : HomeEvents.show_balances)
```

src/home/WalletHome.tsx
```diff
@@ -56,6 +58,16 @@ function WalletHome() {
   const showNotificationCenter = getFeatureGate(StatsigFeatureGates.SHOW_NOTIFICATION_CENTER)
   const showNotificationSpotlight = showNotificationCenter && canShowNotificationSpotlight
 
+  const [hideBalances, setHideBalances] = useState(false)
```
satish-ravi, line 61: Also, I just realized this. Should this state be persisted across sessions? E.g., if someone wants to have their balances hidden when Valora is launched

finnian0826: I'm not sure it makes sense to persist it. In that case a user would have to decide before they leave the app that next time they open it they want it to be hidden, which I don't know how likely that is. We could have the initial state be true so that it is hidden by default?

MuckT: +1 persisting this setting in redux. This could be handled similar to `hapticFeedbackEnabled` in the app state.

src/components/TokenBalance.tsx
```diff
@@ -87,20 +92,22 @@ function TokenBalance({
         <Image source={{ uri: tokensWithUsdValue[0].imageUrl }} style={styles.tokenImg} />
         <View style={styles.column}>
           <Text style={style} testID={'TotalTokenBalance'}>
-            {localCurrencySymbol}
-            {totalTokenBalanceLocal?.toFormat(2) ?? '-'}
-          </Text>
-          <Text style={styles.tokenBalance}>
-            {formatValueToDisplay(tokenBalance)} {tokensWithUsdValue[0].symbol}
+            {!hideBalance && localCurrencySymbol}
+            {hideBalance ? 'XX.XX' : totalTokenBalanceLocal?.toFormat(2) ?? '-'}
```
MuckT, line 96: This should be dependent on the decimal separator similar to the balance. If the decimal seperator is a `,` it should be `XX,XX` instead.

https://github.com/valora-inc/wallet/assets/26950305/52f9432b-9009-460a-a644-79995769d5c5

src/components/TokenBalance.tsx
```diff
@@ -228,6 +246,13 @@ export function HomeTokenBalance() {
 
   const [infoVisible, setInfoVisible] = useState(false)
 
+  const hideBalance = useSelector(hideBalancesSelector)
+
+  const eyeIconOnPress = () => {
+    dispatch(switchHideBalancesState())
+    ValoraAnalytics.track(hideBalance ? HomeEvents.hide_balances : HomeEvents.show_balances)
```
satish-ravi, line 253: wouldn't this fire the opposite event from the action?

src/components/TokenBalance.tsx
```diff
@@ -87,20 +97,26 @@ function TokenBalance({
         <Image source={{ uri: tokensWithUsdValue[0].imageUrl }} style={styles.tokenImg} />
         <View style={styles.column}>
           <Text style={style} testID={'TotalTokenBalance'}>
-            {localCurrencySymbol}
-            {totalTokenBalanceLocal?.toFormat(2) ?? '-'}
-          </Text>
-          <Text style={styles.tokenBalance}>
-            {formatValueToDisplay(tokenBalance)} {tokensWithUsdValue[0].symbol}
+            {(overrideHideBalance || !hideBalance) && localCurrencySymbol}
```
satish-ravi, line 100: I find the overrideHideBalance prop and these booleans expressions hard to read. It seems like only HomeTokenBalance needs to honor the hideBalance state. So could we just use the selector there and TokenBalance could just take a prop on whether to hideBalance or not?

src/components/TokenBalance.tsx
```diff
@@ -87,20 +97,26 @@ function TokenBalance({
...
     )
   } else {
     return (
       <Text style={style} testID={'TotalTokenBalance'}>
-        {localCurrencySymbol}
-        {totalBalanceLocal?.toFormat(2) ?? new BigNumber(0).toFormat(2)}
+        {(overrideHideBalance || !hideBalance) && localCurrencySymbol}
+        {hideBalance && !overrideHideBalance
+          ? 'XX' + decimalSeparator + 'XX'
+          : totalBalanceLocal?.toFormat(2) ?? new BigNumber(0).toFormat(2)}
```
satish-ravi, line 119: can we make this DRY, seems like you could have a var outside the if's e.g., 
```
const balanceDisplay = hideBalance ? `XX${decimalSeparator}XX` : totalBalanceLocal?.toFormat(2)
```
and then do 
```
balanceDisplay ?? `-`
```
and 
```
balanceDisplay ?? new BigNumber(0).toFormat(2)
```
in the two places

Also it's better to use template literals over string concatenation

src/components/TokenBalance.tsx
```diff
@@ -87,20 +97,26 @@ function TokenBalance({
...
-            {totalTokenBalanceLocal?.toFormat(2) ?? '-'}
-          </Text>
-          <Text style={styles.tokenBalance}>
-            {formatValueToDisplay(tokenBalance)} {tokensWithUsdValue[0].symbol}
+            {(overrideHideBalance || !hideBalance) && localCurrencySymbol}
+            {hideBalance && !overrideHideBalance
+              ? 'XX' + decimalSeparator + 'XX'
+              : totalTokenBalanceLocal?.toFormat(2) ?? '-'}
           </Text>
+          {!hideBalance && (
```
satish-ravi, line 105: seems like overrideHideBalance is not honored here. Another reason to drop it and just have a single boolean prop for this function.

src/redux/migrations.ts
```diff
@@ -1416,4 +1416,5 @@ export const migrations = {
       },
     }
   },
+  166: (state: any) => state,
```
satish-ravi, line 1419: This should add the `hideBalances` state to app. You can see other migrations above for examples.

src/transactions/feed/TransactionFeed.test.tsx
satish-ravi, line 1: This just seems like an import sort change and doesn't seem related to the PR. Can we revert?

src/app/actions.ts
```diff
@@ -38,6 +38,7 @@ export enum Actions {
   PUSH_NOTIFICATIONS_PERMISSION_CHANGED = 'APP/PUSH_NOTIFICATIONS_PERMISSION_CHANGED',
   IN_APP_REVIEW_REQUESTED = 'APP/IN_APP_REVIEW_REQUESTED',
   NOTIFICATION_SPOTLIGHT_SEEN = 'APP/NOTIFICATION_SPOTLIGHT_SEEN',
+  SWITCH_HIDE_BALANCES_STATE = 'APP/SWITCH_HIDE_BALANCES_STATE',
```
satish-ravi, line 41: naming nit: maybe `TOGGLE_HIDE_BALANCES` or `HIDE_BALANCES_TOGGLED`?

src/app/actions.ts
```diff
@@ -169,6 +170,10 @@ export interface NotificationSpotlightSeen {
   type: Actions.NOTIFICATION_SPOTLIGHT_SEEN
 }
 
+interface HideBalances {
```
satish-ravi, line 173: use the same name as action name here, similar to other actions

src/app/reducers.ts
```diff
@@ -99,6 +100,7 @@ const initialState = {
   pushNotificationsEnabled: false,
   inAppReviewLastInteractionTimestamp: null,
   showNotificationSpotlight: false,
+  hideBalances: false,
```
satish-ravi, line 103: maybe this can be called `hideHomeBalances`? since its specific to home screen?

src/components/TokenBalance.test.tsx
```diff
@@ -17,6 +18,7 @@ import { createMockStore, getElementText } from 'test/utils'
 import { mockPositions, mockTokenBalances } from 'test/values'
 
 jest.mock('src/statsig')
+jest.mock('src/analytics/ValoraAnalytics')
```
satish-ravi, line 21: we have a global mock for this. I don't think you need it.

src/components/TokenBalance.test.tsx
```diff
@@ -652,6 +654,44 @@ describe('FiatExchangeTokenBalance and HomeTokenBalance', () => {
...
+  it('tracks analytics event when eye icon is pressed', async () => {
+    const store = createMockStore(defaultStore)
+
+    const tree = render(
+      <Provider store={store}>
+        <HomeTokenBalance />
+      </Provider>
+    )
+    fireEvent.press(tree.getByTestId('EyeIcon'))
+    expect(ValoraAnalytics.track).toHaveBeenCalledTimes(1)
```
satish-ravi, line 693: can we assert that the right event is called?

src/transactions/feed/TransactionFeed.test.tsx
finnian0826: I have tried undoing the change manually but when I hit save it then re-arranges the order again. Is there another way that I can revert and save but without it auto arranging the imports?

src/components/TokenBalance.tsx
```diff
@@ -257,13 +278,19 @@ export function HomeTokenBalance() {
...
+      <View style={styles.row}>
+        <TokenBalance
+          style={
+            getFeatureGate(StatsigFeatureGates.SHOW_ASSET_DETAILS_SCREEN)
+              ? styles.totalBalance
+              : styles.balance
+          }
+          hideBalance={hideBalance}
+        />
+        <Touchable onPress={eyeIconOnPress}>
```
MuckT, line 290: nit: add `hitSlop={variables.iconHitslop}` to this touchable.

MuckT (approved):
There's one comment remaining on test assertions and a nit but otherwise LGTM! 🎨

PR #4414, opened 2023-10-30
test(e2e): add assets screen tests

satish-ravi:
### Description

Add tests for the new assets screen for wallet with balance and zero balance

### Test plan

CI
Recording of iOS e2e run: https://drive.google.com/file/d/1txuw_rZbMQ2GyyD_TB7Kl-He_ZDerBuL/view?usp=sharing

### Related issues

- Fixes ACT-934

### Backwards compatibility

N/A


e2e/src/usecases/Assets.js
```diff
@@ -0,0 +1,174 @@
...
+
+async function validateAddFlow(tokenSymbol) {
+  // check the header includes the appropriate token symbol
+  await waitFor(element(by.id('HeaderTitle')))
+    .toHaveText(`Add ${tokenSymbol}`)
+    .withTimeout(10 * 1000)
+  await element(by.id('BackChevron')).tap()
+}
+
+const makeAssetsTests =
```
cajubelt, line 29: could this be done with a describe.each instead? my IDE doesn't let me run tests individually in the UI if they are wrapped in a custom function like this

cajubelt: so it makes development and debugging a bit more cumbersome

cajubelt (approved):

PR #4415, opened 2023-10-30
refactor(fees): use viem for fees on new send flow enter amount screen

cajubelt:
### Description

calculate fees with viem on the enter amount screen

### Test plan

jest tests

manual tests:

https://github.com/valora-inc/wallet/assets/7979215/88453b30-dc84-4751-8ff1-8e1bf8ae257d

### Related issues

### Backwards compatibility

na (new send flow not released yet)


src/send/usePrepareSendTransactions.test.ts
cajubelt, line 1: @satish-ravi it looks like you had prepped this for integration tests that included prepareTransactions. I went a different direction and mocked the hook, since it made the tests in here easier to reason about (making them more like proper unit tests). If integration tests failed, one might have to dig through 3 different layers to find the failure reason (the component, the hook, and the prepareTransactions helper functions).

That said, I wavered quite a bit on this decision and am pretty willing to go in and change it back.

src/send/SendEnterAmount.test.tsx
cajubelt, line 1: @satish-ravi it looks like you had prepped this for integration tests that included prepareTransactions. I went a different direction and mocked the hook, since it made the tests in here easier to reason about (making them more like proper unit tests). If integration tests failed, one might have to dig through 3 different layers to find the failure reason (the component, the hook, and the prepareTransactions helper functions).

That said, I wavered quite a bit on this decision and am pretty willing to go in and change it back. One of the downsides of mocking the hook the way I did was that I wasn't able to verify that calling refreshPreparedTransactions updated the fees shown, for instance

src/send/SendEnterAmount.tsx
```diff
@@ -278,30 +290,35 @@ function SendEnterAmount({ route }: Props) {
...
-                  showLocalAmount={false}
-                  style={styles.feeInCrypto}
-                />
-              </View>
-              <TokenDisplay
-                tokenId={`${token.networkId}:native`}
-                amount={0.005}
-                style={styles.feeInFiat}
-              />
+              {showFeeAmounts ? (
```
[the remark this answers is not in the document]
cajubelt: whoops! added one... it ended up needing some extra logic, which I put into a new component to make it easier to unit test. more complicated than I expected.

test/values.ts
```diff
@@ -511,6 +511,26 @@ export const mockTokenBalances: Record<string, StoredTokenBalance> = {
   },
 }
 
+export const mockCeloTokenBalance: TokenBalance = {
```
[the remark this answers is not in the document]
cajubelt: done

src/send/SendEnterAmount.tsx
```diff
@@ -37,38 +43,59 @@ import { NETWORK_NAMES } from 'src/shared/conts'
...
-  return {
-    type: 'possible' as const,
-  }
+const TAG = 'SendEnterAmount'
+
+const FETCH_UPDATED_TRANSACTIONS_DEBOUNCE_TIME = 250
+
+function FeeLoading() {
+  return (
+    <>
```
satish-ravi, line 65: is the wrapper `<> </>` required?

src/send/SendEnterAmount.tsx
```diff
@@ -37,38 +43,59 @@ import { NETWORK_NAMES } from 'src/shared/conts'
...
+        <Text style={styles.feeInCrypto}>{'≈ '}</Text>
+        <ActivityIndicator size="small" color={Colors.gray4} />
+      </View>
+    </>
+  )
+}
+
+function FeePlaceholder({ feeTokenSymbol }: { feeTokenSymbol: string }) {
+  return (
+    <>
```
satish-ravi, line 76: same question as above

src/send/SendEnterAmount.tsx
```diff
@@ -37,38 +43,59 @@ import { NETWORK_NAMES } from 'src/shared/conts'
...
+      </Text>
+    </>
+  )
+}
+
+function FeeAmount({ feeTokenId, feeAmount }: { feeTokenId: string; feeAmount: BigNumber }) {
+  return (
+    <>
+      <View testID="SendEnterAmount/FeeInCrypto" style={styles.feeInCryptoContainer}>
+        <Text style={styles.feeInCrypto}>~</Text>
```
satish-ravi, line 88: Finn added a `showApprox` prop to TokenDisplay on the send confirmation screen updates PR, so we can drop this Text and the wrapper view and just use that.

src/send/usePrepareSendTransactions.ts
```diff
@@ -0,0 +1,108 @@
...
+    if (result?.type === 'possible') {
+      setFeeCurrency(result.feeCurrency)
+      setFeeAmount(
+        getMaxGasCost(result.transactions).dividedBy(
+          new BigNumber(10).exponentiatedBy(result.feeCurrency.decimals)
+        )
+      )
+    } else {
+      setFeeCurrency(undefined)
+      setFeeAmount(undefined)
```
satish-ravi, line 37: these are for the two warn states right? Should we still set these so a fee + warning is displayed instead of the placeholder?

src/send/usePrepareSendTransactions.ts
```diff
@@ -0,0 +1,108 @@
...
+) {
+  const [prepareTransactionsResult, setPrepareTransactionsResult] = useState<
+    PreparedTransactionsResult | undefined
+  >()
+  const [feeCurrency, setFeeCurrency] = useState<TokenBalance | undefined>()
+
+  const [feeAmount, setFeeAmount] = useState<BigNumber | undefined>()
+  const prepareTransactions = useAsyncCallback(prepareSendTransactionsCallback, {
+    onError: (error) => {
+      Logger.error(TAG, `prepareTransactionsOutput: ${error}`)
```
satish-ravi, line 89: I noticed this error is thrown with fee section just showing a spinner when I choose CELO and enter an amount close to max. I think we'd want to pick another fee currency and return tx is possible if there's balance for them (I did have cUSD, cEUR balance) or return a max-amount response? But maybe this is out of scope for this PR since it seems like an issue with the `prepareERC20TransferTransaction` function

cajubelt: yes I noticed that too. I want to handle it separately.

src/send/usePrepareSendTransactions.ts
cajubelt: if there's not enough balance for gas, it's unclear which fee currency a gas cost should be shown for. Extra, potentially cumbersome logic would be needed in prepareTransactions to handle that.

for the 'need-decrease-spend-amount-for-gas' case, sure

src/send/usePrepareSendTransactions.ts
```diff
@@ -0,0 +1,113 @@
...
+
+// just exported for testing
+export function _getOnSuccessCallback({
+  setFeeCurrency,
+  setPrepareTransactionsResult,
+  setFeeAmount,
+}: {
+  setFeeCurrency: (token: TokenBalance | undefined) => void
+  setPrepareTransactionsResult: (result: PreparedTransactionsResult | undefined) => void
+  setFeeAmount: (amount: BigNumber | undefined) => void
```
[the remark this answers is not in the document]
cajubelt: yeah makes sense, updated

src/send/SendEnterAmount.tsx
```diff
@@ -37,38 +43,55 @@ import { NETWORK_NAMES } from 'src/shared/conts'
...
+const TAG = 'SendEnterAmount'
+
+const FETCH_UPDATED_TRANSACTIONS_DEBOUNCE_TIME = 250
+
+function FeeLoading() {
+  return (
+    <View testID="SendEnterAmount/FeeLoading" style={styles.feeInCryptoContainer}>
+      <Text style={styles.feeInCrypto}>{'≈ '}</Text>
+      <ActivityIndicator size="small" color={Colors.gray4} />
+    </View>
```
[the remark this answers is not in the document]
cajubelt: updated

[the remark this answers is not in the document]
cajubelt: I'm not sure why, but the designs for this component have a skeleton placeholder, and the linear issue mentions an activity indicator. Seems like the designs are the better source of truth...

satish-ravi (approved):

src/send/usePrepareSendTransactions.ts
```diff
@@ -0,0 +1,71 @@
...
+      })
+    }
+    // TODO(ACT-955): case where we need to register DEK too
+  }
+  // TODO(ACT-956): non-ERC20 native asset case
+}
+
+export function usePrepareSendTransactions(
+  prepareSendTransactionsCallback = _prepareSendTransactionsCallback // just for testing
+) {
```
jeanregisser, line 53: nit: I'm not entirely sold on exposing this. I get that it may make testing easier, but also in a way it adds unnecessary details to the interface of the hook which may confuse consumers. 🤔 

Or maybe we should prefix these parameters with `_` and add jsdoc comments so it's clear?

jeanregisser (approved):
👍

src/send/SendEnterAmount.tsx
```diff
@@ -37,37 +37,86 @@ import { NETWORK_NAMES } from 'src/shared/conts'
...
+/**
+ * Given prepared transactions, get the fee currency and amount in decimals
+ *
+ * Exported only for testing
+ *
+ * @param prepareTransactionsResult
+ */
+export function _getFeeCurrencyAndAmount(
+  prepareTransactionsResult: PreparedTransactionsResult | undefined
+): { feeAmount: BigNumber | undefined; feeCurrency: TokenBalance | undefined } {
```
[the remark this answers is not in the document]
cajubelt: moved it to prepareTransactions.ts

src/send/usePrepareSendTransactions.ts
cajubelt: I always feel like there is a faustian bargain between writing a proper unit test and avoding "just for testing" DI. I feel gross in either case. Most of the time, I try to move the helper function into its own module so I can mock it and have it both ways (no ugly DI, and single-purpose tests). But in this case, the prepare send transactions callback really did not make sense to put in its own module.

cajubelt: I took your suggestion with the jsdoc and parameter rename

src/send/SendEnterAmount.tsx
```diff
@@ -37,37 +37,86 @@ import { NETWORK_NAMES } from 'src/shared/conts'
...
+  } else if (prepareTransactionsResult?.type === 'need-decrease-spend-amount-for-gas') {
+    feeCurrency = prepareTransactionsResult.feeCurrency
+    feeAmountSmallestUnits = prepareTransactionsResult.maxGasCost
   }
   return {
-    type: 'possible' as const,
+    feeAmount:
+      feeAmountSmallestUnits &&
+      feeCurrency &&
+      feeAmountSmallestUnits.dividedBy(new BigNumber(10).exponentiatedBy(feeCurrency.decimals)),
```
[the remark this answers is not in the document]
cajubelt: nice. thanks

src/send/usePrepareSendTransactions.ts
cajubelt: now that usePrepareSendTransactions has slimmed down, I am less averse to testing it without mocking _prepareSendTransactionsCallback. updated.

PR #4417, opened 2023-10-31
chore(deps): add WalletConnect to renovate ignore list

MuckT:
### Description

Adds WalletConnect dependencies to [renovate's ignoreDeps list](https://docs.renovatebot.com/configuration-options/#ignoredeps).

### Test plan

N/A

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -66,4 +66,14 @@
       groupName: 'prettier',
     },
   ],
+
+  // A list of dependencies to be ignored by Renovate - "exact match" only
+  ignoreDeps: [
+    '@walletconnect/react-native-compat',
+    '@walletconnect/utils',
+    '@walletconnect/web3wallet',
+    '@walletconnect/legacy-types',
+    '@walletconnect/sign-client',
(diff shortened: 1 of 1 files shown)


renovate.json5
```diff
@@ -66,4 +66,14 @@
       groupName: 'prettier',
     },
   ],
+
+  // A list of dependencies to be ignored by Renovate - "exact match" only
+  ignoreDeps: [
```
satish-ravi, line 71: One implication of this is that we'll never get renovate PRs. If this is going to be temporary this is fine. Another option is to just set a package rule for these deps to have automerge turned off. That way we'd get PRs but they won't be automerged by renovate, but someone would need to remember to close those PRs to unblock others and not merge them, which doesn't seem ideal either.

satish-ravi (approved):

renovate.json5
MuckT: Hopefully this is temporary just until ACT-962 completed 🤞.

PR #4418, opened 2023-11-01
feat: add UI for swap transaction details

kathaypacific:
### Description

This PR just adds the new UI, and a dummy feature flag that prevents it from being shown. The plumbing for getting the fee estimate particularly will be a big ish change so breaking this part out to keep PRs small.

### Test plan

https://github.com/valora-inc/wallet/assets/20150449/f40e9217-19d7-41dc-9b82-b8e8e985a386

### Related issues

Related to RET-865

### Backwards compatibility

Y


locales/base/translation.json
```diff
@@ -1923,5 +1931,9 @@
       "title": "You need more {{feeTokenSymbol}} for gas",
       "description": "Add {{feeTokenSymbol}} to cover your gas fees or adjust the amount you are sending"
     }
+  },
+  "networkName": {
+    "ethereum": "Ethereum",
+    "celo": "Celo"
```
jeanregisser, line 1937: We don't already have this anywhere else?
cc @cajubelt @satish-ravi

src/swap/SwapTransactionDetails.tsx
```diff
@@ -0,0 +1,98 @@
...
+import Touchable from 'src/components/Touchable'
+import InfoIcon from 'src/icons/InfoIcon'
+import colors from 'src/styles/colors'
+import { typeScale } from 'src/styles/fonts'
+import { Spacing } from 'src/styles/styles'
+import { Network } from 'src/transactions/types'
+
+interface Props {
+  network: Network
+  networkFee: BigNumber
```
jeanregisser, line 16: This will probably evolve as it's calculated asynchronously and we'll need a skeleton placeholder I think.

jeanregisser (commented):
Great!

Are we intentionally leaving out tests here for a future PR?

kathaypacific:
> Great!
> 
> Are we intentionally leaving out tests here for a future PR?

yes this component just displays some inputs without much logic so i don't think it's worth testing if it's using fake data. i intend to test that the correct values are displayed in the swap screen directly later

src/swap/SwapTransactionDetails.tsx
kathaypacific: yes i will also address this later, since we need the skeleton loader elsewhere on the screen

locales/base/translation.json
kathaypacific: I couldn't find it in the translations, i suppose we can use a hard coded value if we prefer that. idk if blockchain names can be translated (i imagined that they could?)

locales/base/translation.json
```diff
@@ -1689,6 +1689,14 @@
       "title": "Insufficient gas token balance",
       "description": "You don't have enough {{feeCurrencies}} to cover gas fees. Please add {{feeCurrencies}} to your wallet to complete this swap.",
       "dismissButton": "Dismiss"
+    },
+    "transactionDetails": {
+      "networkFee": "{{networkName}} Network Fee",
+      "swapFee": "{{appName}} Fee",
+      "swapFeeWaived": "Free",
+      "slippagePercentage": "Slippage Tolerance",
+      "networkFeeInfo": "You are swapping on the {{networkName}} Network.\n\nThis estimated network fee is required to conduct a transaction on the {{networkName}} Network. This fee secures your transaction and is paid to network validators.",
```
bakoushin, line 1698: > paid to network validators

Does this claim holds true given that on Celo 80% of base fee is going to a [Carbon Offset Fund](https://blog.celo.org/ultragreen-money-c677e7508abb)?

And on Ethereum the base fee is simply burned.

I wonder if carbon offestting thing can be appealing to some people using Celo (e.g. in the Netherlands some sellers tend to emphasize their environmental impact in their product marketing)?

That having said, I think the most neutral phrasing could probably be like "This fee secures your transaction and is keeps the network operational".

bakoushin (approved):

locales/base/translation.json
jeanregisser: > Does this claim holds true given that on Celo 80% of base fee is going to a [Carbon Offset Fund](https://blog.celo.org/ultragreen-money-c677e7508abb)?

Small nit: 80% is burned, 20% goes to the Carbon Offset Fund.

Screenshot 2023-11-03 at 11 26 51

kathaypacific: TY for picking this up, i copied what was in designs without thinking about it. I modified the copy to your suggestion @bakoushin - i'll check with Nitya and Laura if we want chain specific messaging. if that's the case i'll follow up with a separate PR

locales/base/translation.json
kathaypacific: ah, i found what you mean https://github.com/valora-inc/wallet/blob/main/src/shared/conts.ts#L3 - i updated to use this because it's already there but i'm not sure that this is better than having it in the transaction file (i can imagine for languages like Chinese, there would be a translation for these network names).

PR #4422, opened 2023-11-01
feat(import-token): Add `Import Token` button, feature flagged

denvalora:
### Description

Adds `Import Token` button in the Assets screen when `show_import_tokens_flow` feature flag is on.

I'll add PRs for the Import Token screens afterwards.

### Test plan

Screenshots of before (feature flag is not set or is `false`) and after (flag is `true`).

| Before | After | After (Larger Text) |
| -- | -- | -- |
| image | image | image |

### Related issues

- Starts RET-893

### Backwards compatibility

This adds a button for now for a new flow, no breaking changes.


src/components/ImportTokenButton.tsx
```diff
@@ -0,0 +1,57 @@
...
+      icon={<Times />}
+    />
+  )
+}
+
+const styles = StyleSheet.create({
+  title: {
+    ...typeScale.labelMedium,
+    color: Colors.greenUI,
+    paddingRight: 8,
```
bakoushin, line 55: nit: we can also use a [Spacing.Smallest8](https://github.com/valora-inc/wallet/blob/05e659e3305a9847cff46c0259b19104bc64cbe8/src/styles/styles.ts#L7C7-L7C7)

src/tokens/Assets.tsx
```diff
@@ -483,6 +484,14 @@ function AssetsScreen({ navigation, route }: Props) {
...
 AssetsScreen.navigationOptions = {
   ...headerWithBackButton,
+  headerRight: HeaderRight,
+}
+
+function HeaderRight() {
+  if (getFeatureGate(StatsigFeatureGates.SHOW_IMPORT_TOKENS_FLOW)) {
+    return <ImportTokenButton />
+  }
+  return <></>
```
bakoushin, line 494: what is the advantage of using this over `null`?

src/components/ImportTokenButton.tsx
```diff
@@ -0,0 +1,57 @@
+import * as React from 'react'
```
bakoushin, line 1: What is the advantage of having this custom component over generic RN [Button](https://reactnative.dev/docs/button)?

bakoushin:
Congrats with your first PR! 🎉

Left some questions here and there.

src/components/ImportTokenButton.tsx
```diff
@@ -0,0 +1,57 @@
...
+  function onPressCancel() {
+    if (onCancel) {
+      onCancel()
+    } else {
+      navigateBack()
+    }
+  }
+
+  const { t } = useTranslation()
+  return buttonType !== 'icon' ? (
```
cajubelt, line 33: is this ever going to be icon? looks like it's just in one place in the designs, and it's text

src/components/ImportTokenButton.tsx
```diff
@@ -0,0 +1,57 @@
...
+  buttonType?: 'text' | 'icon'
+}
+
+export default function ImportTokenButton({
+  eventName,
+  onCancel,
+  style,
+  buttonType = 'text',
+}: Props) {
+  function onPressCancel() {
```
cajubelt, line 24: looks like you just copied the CancelButton component. calling the onPress handler "onPressCancel" doesn't make sense if the button isn't a cancel button.

src/components/ImportTokenButton.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { navigateBack } from 'src/navigator/NavigationService'
+import { TopBarIconButton, TopBarTextButton } from 'src/navigator/TopBarButton'
+import { Colors } from 'src/styles/colors'
+import { typeScale } from 'src/styles/fonts'
+
+interface Props {
+  onCancel?: () => void
+  style?: StyleProp<TextStyle>
+  eventName?: AnalyticsEventType
+  buttonType?: 'text' | 'icon'
```
cajubelt, line 15: these don't make sense to me.
1. what does "onCancel" even mean when you are importing a token? This button is only used in one place-- does it even need a configurable onPress prop?
2. Again, the button's only in one place in the designs-- why add a style prop?
3. For the same reason, why does it need a configurable analytics event name?
4. Same reason, why allow it to be text or icon if it's only going to be text?

src/components/ImportTokenButton.tsx
cajubelt: by the way, if you find yourself copy/pasting entire files like this, you should probably look for something you can reuse instead, like Alex's suggestion on using a generic RN button

src/components/ImportTokenButton.tsx
cajubelt, line 1: we should probably reserve the "components" directory for buttons that are more reusable and less single-purpose like this one

cajubelt (changes requested):
Unless this is a draft PR we need you to include tests

src/tokens/Assets.tsx
```diff
@@ -483,6 +484,14 @@ function AssetsScreen({ navigation, route }: Props) {
 
 AssetsScreen.navigationOptions = {
   ...headerWithBackButton,
+  headerRight: HeaderRight,
+}
+
+function HeaderRight() {
+  if (getFeatureGate(StatsigFeatureGates.SHOW_IMPORT_TOKENS_FLOW)) {
+    return <ImportTokenButton />
```
kathaypacific, line 492: i don't think you need a new component at all, you can directly reuse `TopBarTextButton`. [here](https://github.com/valora-inc/wallet/blob/9fa5c4d87373843306815be395f4f54c36bd8fa3/src/webview/WebViewScreen.tsx#L95-L100) is an example

locales/base/translation.json
```diff
@@ -122,6 +122,7 @@
   "editProfile": "Edit Profile",
   "confirmNumber": "Connect phone number",
   "cancel": "Cancel",
+  "importToken": "Import Token",
```
kathaypacific, line 125: can you please check with large device fonts enabled that the layout is okay? the assets screen has a scroll aware header and i think we may need to rethink this copy

src/tokens/Assets.tsx
kathaypacific: agree, returning null or undefined is preferable i think

locales/base/translation.json
denvalora: | Standard | Larger Text | Larger Accessibility |
| -- | -- | -- |
| image | image | image |

src/tokens/Assets.test.tsx
```diff
@@ -336,6 +336,21 @@ describe('AssetsScreen', () => {
...
+    jest.mocked(getFeatureGate).mockReturnValue(true)
+    const store = createMockStore(storeWithPositionsAndClaimableRewards)
+
+    const component = (
+      <Provider store={store}>
+        <MockedNavigator component={AssetsScreen} />
+      </Provider>
+    )
+    const { getByTestId } = render(component)
+    fireEvent.press(getByTestId('Assets/ImportTokenButton'))
```
bakoushin, line 349: Nit: we aim to use `getByTestId` as little as possible unless there is a strong reason (e.g., testing which text is displayed).

In this case, if we use the `getByText` selector, the test behavior would be closer to the actual user experience.

src/tokens/Assets.test.tsx
```diff
@@ -336,6 +336,21 @@ describe('AssetsScreen', () => {
     expect(navigate).toHaveBeenCalledWith(Screens.DappShortcutsRewards)
   })
 
+  it('clicking Import Token opens a screen', () => {
+    jest.mocked(getFeatureGate).mockReturnValue(true)
```
bakoushin, line 340: It could make sense to test the reverse state as well (when the feature flag is not enabled).

src/tokens/Assets.tsx
```diff
@@ -572,6 +587,11 @@ const styles = StyleSheet.create({
     ...typeScale.labelMedium,
     color: Colors.dark,
   },
+  topBarTextItem: {
```
bakoushin, line 590: Is it intentional that the style for `TextButton` is named `TextItem`? I'm doubtful that the `greenUI` color could be reusable for regular text :sweat_smile:

src/tokens/Assets.tsx
```diff
@@ -414,6 +415,20 @@ function AssetsScreen({ navigation, route }: Props) {
...
+    navigation.setOptions({
+      headerRight: () =>
+        getFeatureGate(StatsigFeatureGates.SHOW_IMPORT_TOKENS_FLOW) ? (
+          <TopBarTextButton
+            testID="Assets/ImportTokenButton"
+            onPress={navigateBack}
+            title={t('importToken')}
+            style={styles.topBarTextItem}
+          />
+        ) : null,
```
bakoushin, line 428: Here is a neat trick widely used: we can use [`&&` operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND) instead of ternary + null. Makes code a bit less noisy. 
```suggestion
        getFeatureGate(StatsigFeatureGates.SHOW_IMPORT_TOKENS_FLOW) && (
          <TopBarTextButton
            testID="Assets/ImportTokenButton"
            onPress={navigateBack}
            title={t('importToken')}
            style={styles.topBarTextItem}
          />
        ),
```

locales/base/translation.json
kathaypacific: hmm yeah i guess the definition of what is acceptable layout is subjective, if Nitya is okay with this then fine. but i personally think that we'd be better off with just "Import" so that it doesn't push the (more important) asset balance off center

src/tokens/Assets.test.tsx
kathaypacific: agree with this - there is more than 1 feature gate used in this Assets component so we should really be mocking when `getFeatureGate(StatsigFeatureGates.SHOW_IMPORT_TOKENS_FLOW)` returns true and false

kathaypacific (commented):
looking really close! :)

src/tokens/Assets.tsx
```diff
@@ -414,6 +415,20 @@ function AssetsScreen({ navigation, route }: Props) {
...
+    })
+  })
```
bakoushin, line 430: Should we provide a dependency array?

src/tokens/Assets.tsx
denvalora: Got distracted by `tabBarItem` as it sounds close to `topBar...`. Fixed.

TopBarTextButton actually uses `colors.greenUI` by default. Removed here due to repeating it.

src/tokens/Assets.tsx
```diff
@@ -414,6 +415,19 @@ function AssetsScreen({ navigation, route }: Props) {
     return items
   }, [t, displayPositions])
 
+  useLayoutEffect(() => {
```
kathaypacific, line 418: nit: while the order in which we declare variables and hooks don't affect the outcome at runtime, components can get long and difficult to read so it helps to have a predictable order. generally i think we're using the pattern of declaring state and hooks at the top of the component, and handlers / variables after those. can we please move this function to live closer to the other hooks?

src/tokens/Assets.test.tsx
```diff
@@ -336,6 +342,37 @@ describe('AssetsScreen', () => {
...
+      </Provider>
+    )
+    const { queryByText } = render(component)
+    const button = queryByText('assets.importToken')
+
+    expect(button).toBeNull()
+  })
+
+  it('clicking Import Token opens a screen', () => {
+    mockEnabledFeatureFlags(StatsigFeatureGates.SHOW_IMPORT_TOKENS_FLOW)
```
kathaypacific, line 361:
```suggestion
    jest
      .mocked(getFeatureGate)
      .mockImplementation(
        (gate: StatsigFeatureGates) => gate === StatsigFeatureGates.SHOW_IMPORT_TOKENS_FLOW
      )
```
code golf. your implementation works, but there are other tests that use this other pattern. i think it is quite readable (it is nice to only see the override variables needed for the test, rather than having to also declare the default state as in the test above). unless there is a clear benefit, i think predictability and consistency are the most important principles so let's reuse the pattern? remember to add `jest.mocked(getFeatureGate).mockReset()` to the `beforeEach` so that the mocks are cleared from previous tests with this approach

kathaypacific (approved):
looks good, i'm approving but left 2 small comments which we should address before merging 🙏🏻

PR #4425, opened 2023-11-02
feat: add network fee to swap screen

kathaypacific:
### Description

This PR:
- add network fee in fiat currency to swap tx details display
- enable swap tx details display on the swap input screen permanently

Not included in this PR:
- skeleton loading state for the network fee

For the network fee value, I tried to emulate the logic used in the swap review screen but since we are axing the contractkit flow soon, i didn't want to copy over too much extra logic.
- for the viem swap flow, use the fees in the prepared transactions. if the transaction isn't possible, default to "-" (the user can't proceed with the swap anyway). for ease of use, i added the total fee to the prepared transaction, since it is already there.
- for the contractkit flow, i made an approximation using the data provided by the quote itself and assuming that the fee currency is Celo (rather than bringing over all the fee currency calculation logic). since gas fees are low, and we are displaying it in fiat value, i think the difference will be negligable.
- use the TokenDisplay component to show the fiat value of the network fee, rather than calculating it upfront.

### Test plan

Using the viem and contractKit flows, verify that the network fee value displayed in this PR is the same as what is displayed in the swap review screen. **Note that there are some performance issues on this screen - I can see the same issues on main so will address those separately, they were not introduced by changes in this PR.** At the time of recording, our swap providers are having performance issues so the videos are very long :( 

ContractKit: notice the fee currency in fiat in the first screen is the same as the fee currency in token units since i have EUR local currency + cEUR as the fee currency. 

https://github.com/valora-inc/wallet/assets/20150449/a9638538-4205-434d-8de7-196609772b9e

Viem: notice the fee currency in fiat is half as much as the token value in the swap review screen (1 CELO is roughly 0.5EUR)

https://github.com/valora-inc/wallet/assets/20150449/f41bcbd9-dc82-4839-964d-bf9c47d9a781

### Related issues

- Related to RET-865

### Backwards compatibility

Y


src/viem/prepareTransactions.ts
```diff
@@ -16,6 +16,8 @@ import erc20 from 'src/abis/IERC20'
 interface PreparedTransactionsPossible {
   type: 'possible'
   transactions: TransactionRequestCIP42[]
+  feeCurrency: TokenBalance
+  maxGasCostInDecimal: BigNumber
```
jeanregisser, line 20: I think we should have a helper function to get these directly from the prepared transactions, so it remains the single source of truth.
It would probably use `getMaxGasCost` under the hood.
Related: https://github.com/valora-inc/wallet/pull/4415#discussion_r1380576424

@cajubelt did that already in https://github.com/valora-inc/wallet/pull/4415/files#diff-0412194130424f92cdf96a29da9526a23b02f6ab75c78a899bc934e9c8f1ba08R102

jeanregisser (commented):
Approach looks good to me 👍

src/viem/prepareTransactions.ts
jeanregisser: Hum, maybe it's fine to attach (some of ?) these to `PreparedTransactionsResults` after all 🤔

src/swap/SwapReviewScreen.tsx
```diff
@@ -111,9 +111,9 @@ export function SwapReviewScreen({ route }: Props) {
       return new BigNumber(0)
     }
 
-    return getMaxGasFee(quote.preparedTransactions.transactions)
-      .shiftedBy(-feeCurrencyToken.decimals)
-      .times(feeCurrencyToken.priceUsd)
```
kathaypacific, line 116: we are showing the network fee in token units on the review screen, so this is incorrect...

jeanregisser: Damn, good catch, I need to fix this then for today's release 😄

kathaypacific: if we merge this then we don't need to fix it separately? :D

src/viem/prepareTransactions.ts
```diff
@@ -9,18 +10,18 @@ import { publicClient } from 'src/viem/index'
...
 } from 'viem'
-import stableToken from 'src/abis/StableToken'
 
 const TAG = 'viem/prepareTransactions'
 
 export interface PreparedTransactionsPossible {
   type: 'possible'
   transactions: TransactionRequestCIP42[]
   feeCurrency: TokenBalance
+  maxGasFeeInDecimal: BigNumber
```
jeanregisser, line 24: Would be nice to use the same property name here and in
https://github.com/valora-inc/wallet/blob/41eae0760b73f3ac7b5072907bf852813e7ba3e8/src/viem/prepareTransactions.ts#L28

src/swap/SwapScreen.tsx
```diff
@@ -55,13 +58,43 @@ const DEFAULT_SWAP_AMOUNT: SwapAmount = {
 
 type Props = NativeStackScreenProps<StackParamList, Screens.SwapScreenWithBack>
 
+function getNetworkFee(quote: QuoteResult | null, networkId?: NetworkId) {
+  if (quote && quote.preparedTransactions?.type === 'possible') {
+    return {
+      networkFee: quote.preparedTransactions.maxGasFeeInDecimal,
+      feeTokenId: quote.preparedTransactions.feeCurrency.tokenId,
+    }
+  }
```
jeanregisser, line 67: We could use `getFeeCurrencyAndAmount` here, it would allow displaying a fee when for the `need-decrease-spend-amount-for-gas` type too.
https://github.com/valora-inc/wallet/blob/41eae0760b73f3ac7b5072907bf852813e7ba3e8/src/viem/prepareTransactions.ts#L368

jeanregisser (commented):
Great! 🎉

src/viem/prepareTransactions.ts
kathaypacific: yes good idea...this wasn't there when i started implementing this PR. naming is hard, and i'm not sure that `InDecimal` is crystal clear but it seems like we have 1 gas fee value in the minimum token units and another after shifting by the token decimals. `maxGasFee` there is actually in decimals too :/ so I thought it was more clear to rename `maxGasFee` to `maxGasFeeInDecimal`. wdyt?

src/swap/SwapScreen.tsx
kathaypacific: updated!

src/swap/SwapScreen.tsx
```diff
@@ -58,13 +62,35 @@ const DEFAULT_SWAP_AMOUNT: SwapAmount = {
 
 type Props = NativeStackScreenProps<StackParamList, Screens.SwapScreenWithBack>
 
+function getNetworkFee(quote: QuoteResult | null, networkId?: NetworkId) {
+  // TODO remove this block once we have viem enabled for everyone. this block
+  // only services the contractKit (Celo) flow. the fee will be displayed in
+  // fiat value and will be very low, since it's an approximation anyway we'll
+  // simplify this logic to just use the fees from the quote.
+  if (quote && !quote.preparedTransactions) {
```
jeanregisser, line 70: nit: code golf
```suggestion
  if (!quote?.preparedTransactions) {
```

jeanregisser (approved):
🎉

src/swap/SwapScreen.tsx
dievazqu: If `quote` is null we don't want to enter in the if block.
`!quote?.preparedTransactions` will return true if quote is null, right?

kathaypacific: yes, i need the quote here but no prepared transactions :)

PR #4426, opened 2023-11-02
chore: rename "gas cost" -> "gas fee"

jeanregisser:
I started using the "gas cost" terminology in https://github.com/valora-inc/wallet/pull/4368 which is also mentioned a few times in https://ethereum.org/en/developers/docs/gas
But I think we should use "gas fee" to be more clear maybe? Which is referred more commonly in that same doc.

https://github.com/valora-inc/wallet/pull/4402#discussion_r1376416923


kathaypacific (approved):
💅🏼

PR #4430, opened 2023-11-03
feat(ci): Replace depcheck with knip, add dead code regression to CI

jophish:
### Description

This PR does two things:
* Removes depcheck, and replaces our depcheck CI step with knip instead, using only its dependency checking
* Adds in a new regression CI step using knip, which basically runs all of knip's checks against the PR branch and main, and for each check, ensures that the PR branch has no more errors than main.

See a successful run of the regression step [here](https://github.com/valora-inc/wallet/actions/runs/6741018767/job/18324961630); you can see an overview of the diff between PR branch and main near the end of the main output.

(re: the knip CI script -- knip doesn't expose a programmatic API; it's only available as a CLI tool, so it required a bit of hacky regex'ing to parse the results)

### Test plan

CI

### Related issues

N/A

### Backwards compatibility

Yes.


.github/scripts/checkKnipRegression.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+function compareKnipResults(
+  mainKnipResults: Record<string, number>,
+  branchKnipResults: Record<string, number>
+): boolean {
+  let shouldFail = false
+  for (const [category, mainProblems] of Object.entries(mainKnipResults)) {
+    const branchProblems = branchKnipResults[category] ?? 0
+    console.log(`${category} Main: ${mainProblems} Branch: ${branchProblems}`)
```
jeanregisser, line 27: nit: what do you think of using [`console.table`](https://developer.mozilla.org/en-US/docs/Web/API/console/table) to have a slightly more readable summary?

jeanregisser (approved):
Awesome! ✂️ 🚀 
This flagged so many things we can possibly remove 🙌 

Should we also delete `.depcheckrc`?

.github/scripts/checkKnipRegression.ts
jophish: oh this is pretty handy! will do.

.github/scripts/checkKnipRegression.ts
```diff
@@ -0,0 +1,60 @@
+import * as $ from 'shelljs'
+
+const SECTION_REGEX = /[a-zA-Z ]+\([0-9]+\)/
+const CATEGORY_REGEX = /[a-zA-Z ]+/
+const PARENS_REGEX = /\(|\)| /
+
+function parseKnipOutput(knipOutput: string): Record<string, number> {
```
cajubelt, line 7: would be cool if this had a test that just included some output you've seen from knip

.github/scripts/checkKnipRegression.ts
```diff
@@ -0,0 +1,60 @@
...
+      const category = (line.match(CATEGORY_REGEX) as RegExpMatchArray)[0]
+      result[category] = parseInt(tokens[tokens.length - 2])
+    }
+  }
+  return result
+}
+
+function compareKnipResults(
+  mainKnipResults: Record<string, number>,
+  branchKnipResults: Record<string, number>
```
cajubelt, line 22: nit: if these were an object it would be easier to avoid bugs where you include the right params but in the wrong order

.github/scripts/checkKnipRegression.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+  }
+  console.table(rows)
+  return shouldFail
+}
+
+if (process.env.GITHUB_EVENT_NAME === 'pull_request') {
+  const branchName = process.env.GITHUB_HEAD_REF
+
+  const branchKnipOutput = $.exec('yarn knip --no-gitignore').stdout.trim()
```
cajubelt, line 44: nit: should the no-gitignore just be in the knip config?

.github/workflows/check.yml
```diff
@@ -94,7 +94,7 @@ jobs:
           key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock', 'patches/*.patch') }}
       - run: yarn || yarn --network-concurrency 1
       - run: yarn check-licenses
-  depcheck:
+  knip-depcheck:
```
cajubelt, line 97: maybe instead of replacing depcheck immediately we could keep depcheck and run knip depcheck as a non-required CI step for a while to see if it's working?

cajubelt: seems possible that we'll notice some categories we don't care about, like unused enum members or something, and we'll want to update the knip config. would be nice if we weren't blocked with our other work every time we find something like that

cajubelt (commented):
very cool!

cajubelt (approved):
actually my comments from before aren't really blocking

cajubelt:
would be cool to publish this as a standalone github action that we can use for other repos and maybe get community contributions to

PR #4432, opened 2023-11-03
fix(prepare-transactions): only throw error if transfer value exceeds balance

cajubelt:
depends on #4415 which should be merged first

### Description

We should only throw errors in `prepareTransactions` if the amount the user is trying to spend exceeds their balance, not if gas plus the value they're sending exceeds balance. (In the latter case, we want to swallow the error and give a result with type "need-decrease-spend-amount-for-gas" or "not-enough-balance-for-gas".)

Previously, we were getting errors for both cases in the send flow.

### Test plan

added and updated jest tests

with the fix, selecting an amount that is less than the balance but still too high:
simulator_screenshot_DE10EB16-55B8-4CEF-96B6-743A5DAB5C5A

without the fix, selecting the same amount in the same wallet:
simulator_screenshot_932AD76B-0B45-4230-A090-CDD3F18DCD40

### Related issues

### Backwards compatibility

na


src/viem/prepareTransactions.test.ts
```diff
@@ -177,7 +227,7 @@ describe('prepareTransactions module', () => {
       const result = await prepareTransactions({
         feeCurrencies: mockFeeCurrencies,
         spendToken: mockFeeCurrencies[1],
-        spendTokenAmount: new BigNumber(100_000),
+        spendTokenAmount: new BigNumber(9_000),
```
jeanregisser, line 230: This isn't more than the max amount for `mockFeeCurrencies[1]` no?
Maybe we should update the test description then? since we can't try to prepare more than the spend balance anymore.
Or even remove it?

src/viem/prepareTransactions.ts
```diff
@@ -153,6 +191,11 @@ export async function prepareTransactions({
   decreasedAmountGasFeeMultiplier: number
   baseTransactions: (TransactionRequestCIP42 & { gas?: bigint })[]
 }): Promise<PreparedTransactionsResult> {
+  if (spendTokenAmount.isGreaterThan(spendToken.balance.shiftedBy(spendToken.decimals))) {
+    throw new Error(
+      `Cannot prepareTransactions for amount greater than balance. Amount: ${spendTokenAmount}, Balance: ${spendToken.balance}, Decimals: ${spendToken.decimals}`
+    )
+  }
```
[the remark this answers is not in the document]
cajubelt: right

src/viem/prepareTransactions.test.ts
cajubelt: yeah it doesn't look like this test adds useful coverage anymore given the other tests that we have and the expected behavior now. I removed it

jeanregisser (approved):
🚀

PR #4436, opened 2023-11-05
feat(import-token): Create a Notification component

denvalora:
### Description

This is based on the [Warning](https://github.com/valora-inc/wallet/blob/main/src/components/Warning.tsx) component.
- Extended to display Informational and Error notifications.
- Made `title` property optional.
- Made styling consistent with the design in [Figma](https://www.figma.com/file/l6cnlCx2Ht2YSgnbBdzZph/App-Style-Guide?type=design&node-id=311-245&mode=dev). The updated screens all show rounded corners.

Storybook demo of the component is available in [2e6da03](/valora-inc/wallet/tree/2e6da03). `yarn` to install packages, `yarn storybook` to start the Storybook which previews the component in different variations.

### Test plan

Tested:
- if CTA buttons appear,
- whether onPress calls the passed functions, 
- and whether the rendered matches the design in these scenarios (note: _headline = title_):

| Headline | Description | Actions |
| -- | -- | -- |
| ✅ | ✅ | ✅ |
| ❌ | ✅ | ✅ |
| ❌ | ✅ | ❌ |
| ✅ | ✅ | ❌ |

#### SendEnterAmount
| Before | After | Latest |
| -- | -- | -- |
| image | image | image |
| image | image | image |

#### SwapScreen
| Before | After | Latest |
| -- | -- | -- |
| image | image | image |
| image | image | image |
| image | image | image |

#### ActionRequest
| Before | After | Latest |
| -- | -- | -- |
| image | image | image |

#### SessionRequest
| Before | After | Latest |
| -- | -- | -- |
| image | image | image |

### Related issues

- `InLineNotification` with `Severity.Informational` will be used in RET-893

### Backwards compatibility

No issues. `Warning` component is removed, and all uses are replaced with this updated component. The styling differs from `Warning` component but looks fine.


src/components/InLineNotification.tsx
```diff
@@ -0,0 +1,158 @@
...
+    paddingVertical: Spacing.Tiny4,
+  },
+  ctaLabel2: {
+    ...fontStyles.small600,
+    paddingHorizontal: Spacing.Smallest8,
+    paddingVertical: Spacing.Tiny4,
+  },
+})
+
+const severityStyles: Record<Severity, CustomStyles> = {
```
kathaypacific, line 137: this works, but i think we may not need so much complexity / flexibility. providing a way to change many styles for these elements implies that we can customise the component a lot between the different severity levels (e.g. i can maybe add a paddingVertical to the container in Information, but not in Warning) - i think that this is flexibility that we actually don't want, to keep the component looking consistent between the severity levels. any customised style props should be applied to all variations.

from the design, it looks like the only thing that changes between the severity levels is color. so i think we could define a simple map that provides a primary and secondary color to each severity, and just add inline styles inside the component (e.g. instead of ` <Text
                style={[defaultStyles.ctaLabel2, severityStyle.ctaLabel2]}
                onPress={onPressCta2}
              >` we could have ` <Text
                style={[defaultStyles.ctaLabel2, { color: inLineNotificationColors[severity].primary }]}
                onPress={onPressCta2}
              >` . i'd argue that this makes the component easier to read, since the level of customisation is right there. what do you think?

src/components/InLineNotification.tsx
```diff
@@ -0,0 +1,158 @@
...
+        </View>
+      )}
+
+      <View style={defaultStyles.bodyContainer}>
+        <View style={[defaultStyles.row, defaultStyles.bodyRow]}>
+          <View style={defaultStyles.attentionIcon}>
+            {!title && <AttentionIcon color={severityStyle.attentionIcon.color} />}
+          </View>
+          <Text style={defaultStyles.bodyText}>{description}</Text>
+        </View>
```
kathaypacific, line 62: i think that the structure can be simplified - the way you have it here, you need to declare the icon twice. but it seems like the icon should always be there. to simplify you could have a structure like
```
<View style={style.row}>
  <Icon>
  <View>
    {title ?? <Text {...} />}
    {description}
    <View style={styles.row}>
      {ctas}
    </View>
  </View>
</View>
```
alternatively you could have a row that has the icon on the left, title and description on the right, and a second row for buttons

src/components/InLineNotification.tsx
```diff
@@ -0,0 +1,158 @@
...
+  bodyContainer: {
+    gap: Spacing.Smallest8,
+  },
+  row: {
+    flexDirection: 'row',
+    gap: Spacing.Smallest8,
+    alignItems: 'center',
+  },
+  bodyRow: {
+    alignItems: 'flex-start',
```
kathaypacific, line 104: this is already the default value, do you need this?

src/components/InLineNotification.tsx
```diff
@@ -0,0 +1,158 @@
...
+  },
+  ctaRow: {
+    paddingVertical: Spacing.Tiny4,
+    paddingHorizontal: Spacing.Smallest8,
```
kathaypacific, line 108: are you sure you want padding horizontal here? doesn't that mean the right hand side of the button would be indented on the right compared to the description (i.e. the content doesn't line up on the right)?

src/styles/colors.tsx
```diff
@@ -39,6 +39,9 @@ export enum Colors {
   successDark = '#137211',
   successLight = '#F1FDF1',
 
+  informationDark = '#0768AE',
```
kathaypacific, line 42: we shouldn't be adding new colors to the stylesheet unless it exists in the design system. there is already an informational and onboardingBlue using this hex value. in figma it looks like it says onboardingBlue when I inspect the value, so please use that instead

src/styles/colors.tsx
```diff
@@ -39,6 +39,9 @@ export enum Colors {
...
+  informationFaint = '#E8F8FF',
```
kathaypacific, line 43: for this color i'd just go with `onboardingLightBlue`

src/components/InLineNotification.test.tsx
```diff
@@ -0,0 +1,43 @@
+import { fireEvent, render } from '@testing-library/react-native'
+import * as React from 'react'
+import InLineNotification, { Severity } from 'src/components/InLineNotification'
+
+describe(InLineNotification, () => {
+  it('does not render CTA when onPress function is not provided', async () => {
+    const { findByText } = render(
```
kathaypacific, line 7: you should be able to use `getByText` here - this component isn't doing anything async before rendering. you'll get better errors with `getByText` too.

kathaypacific, line 7: i think you should also add an assertion that the title and description can be found

kathaypacific (commented):
looking good. from the screenshots, the padding bottom doesn't look like it matches the padding top when there are no cta buttons, could we please look into that?

src/components/InLineNotification.tsx
denvalora: >can customise the component a lot between the different severity levels 
Originally I meant it for customising colours, e.g. if CTA buttons may need to have different colors. If it's needed in the future, we can come back. I agree on keeping things consistent with Severity (or Type/Level).

src/components/InLineNotification.tsx
denvalora: It's not necessary for the text component but for the row's View component. It groups both the icon's View and Text components together.

If the description is spread across 2 lines or more, then the icon would be vertically centered. This would not look great. Here's a preview:

image

src/styles/colors.tsx
denvalora: I think it's better to try to reuse the colors if possible, although I think informationFaint is easier to see/read in this case. I'll keep informationFaint.

| Colors.informationFaint | Colors.onboardingLightBlue
| -- | -- |
| image | image |

kathaypacific: i think my main point here is that we are trying to make our design system in code the same as what exists in Figma since we have RET-913 coming up. i suggest using an existing color if it does no harm, but if you think that it's not a good alternative then we should check with Nitya if this color belongs in the design system and align on the name she will call it in Figma going forward (i think on last inspection, i did not see a name for it).

src/styles/colors.tsx
denvalora: ## Primary colour
[Icon](https://www.figma.com/file/l6cnlCx2Ht2YSgnbBdzZph/App-Style-Guide?type=design&node-id=311-269&mode=design&t=RCaFcCeXk1L63S33-4) uses `background: #0768AE;` (Onboarding/Blue) matching [Info-Dark](https://www.figma.com/file/erFfzHvSTm5g1sjK6jWyEH/Working-Design-System-Components?type=design&node-id=213-4&mode=design&t=qWFasC1gqiWu4zR0-4)

We already had informational and onboardingBlue. I'll rename `informational` to `informationDark`.

src/components/InLineNotification.tsx
denvalora: That's correct. I think it's more visually pleasing to keep this padding, helps to have a focus on the button options. Figma design matches this as well.

src/components/InLineNotification.tsx
denvalora: I've changed this part a bit to define the icon just once and remove some redundant View components.

I've went with a solution where the title and description are in separate rows. I chose that because we vertically center the icon with the title.

src/styles/colors.tsx
denvalora: The [Figma design](https://www.figma.com/file/erFfzHvSTm5g1sjK6jWyEH/Working-Design-System-Components?type=design&node-id=213-5&mode=design&t=oTFPVxN7Czm77Rb3-4) for Typography & Color does have `#E8F8FF` color, it's called Info-Light.

I've mentioned to Nitya that we don't have the Info-Light color, but we have `onboardingLightBlue` color `#D6EAF3`. She has updated the design, which references Info-Light color and mentioned I can this new color to the app.

To keep it consistent, I've renamed it to `infoLight` in the code (and also renamed the other color to `infoDark`, it was used in only one place).

src/components/InLineNotification.tsx
bakoushin: I'm missing the word `color` either in the name of the object (e.g. `severetyColors`) or in property names (e.g. `primaryColor`, `secondaryColor`).

src/components/InLineNotification.tsx
```diff
@@ -0,0 +1,148 @@
...
+  ctaLabel,
+  onPressCta,
+  ctaLabel2,
+  onPressCta2,
+  testID,
+}: Props) {
+  const severityStyle = severityStyles[severity]
+  const renderCtaLabel = (
+    label?: string | null,
+    onPress?: (event: GestureResponderEvent) => void,
```
bakoushin, line 45: Is this the same `onPress` we get from top-level component props? I'm confused why the typings are different.

src/components/InLineNotification.tsx
```diff
@@ -0,0 +1,148 @@
...
+        <View style={[defaultStyles.row, defaultStyles.ctaRow]}>
+          {renderCtaLabel(ctaLabel, onPressCta, severityStyle.primary)}
+          {renderCtaLabel(ctaLabel2, onPressCta2, severityStyle.primary)}
+        </View>
+      )}
+    </View>
+  )
+}
+
+const defaultStyles = StyleSheet.create({
```
kathaypacific, line 87:
```suggestion
const styles = StyleSheet.create({
```

src/components/InLineNotification.tsx
```diff
@@ -0,0 +1,148 @@
...
+  iconLessRow: {
+    paddingLeft: Spacing.Large32,
+  },
+  attentionIcon: {
+    padding: Spacing.Tiny4,
+    width: Spacing.Thick24,
+    height: Spacing.Thick24,
+  },
+  titleText: {
+    ...fontStyles.small600,
```
kathaypacific, line 117: we should avoid adding new deprecated styles. can we find a close match with something from `typeScale`? maybe `labelSmall` and set a `fontWeight` if you need to?

src/components/InLineNotification.tsx
kathaypacific: is it a hard requirement that the icon is vertically centered if the title exceeds one line? there's no example in the Figma file, and the example with the multi-line description + no title shows the icon is not vertically centered. so if it is okay to keep this pattern so that the icon is always aligned to the top, i have an idea to simplify the styles of this component. i tried to describe this [here](https://github.com/valora-inc/wallet/pull/4436#discussion_r1387808143) but i think i was not clear, i made a [branch](https://github.com/valora-inc/wallet/compare/denvalora/notifications...kathy/notification-alternative?expand=1) to show you my idea - lmk what you think :)

kathaypacific (commented):
this is looking great, just left a few code golfy questions

src/components/InLineNotification.tsx
denvalora: It comes from `node_modules/react-native/Libraries/Text/Text.d.ts`
```ts
onPress?: ((event: GestureResponderEvent) => void) | undefined;
```

src/components/InLineNotification.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+        ctaLabel={'Action 1'}
+        onPressCta={jest.fn()}
+        ctaLabel2={'Action 2'}
+      />
+    )
+
+    expect(getByText('Title')).toBeTruthy()
+    expect(getByText('Description')).toBeTruthy()
+    expect(getByText('Action 1')).toBeTruthy()
+    expect(() => getByText('Action 2')).toThrow()
```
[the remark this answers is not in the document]
denvalora: Added also `queryByText` into the destructuring assignment.

src/components/InLineNotification.tsx
denvalora: I've cherry picked your commit and added it to this branch.

## Before
| Description | Description + CTA | Short title + description | Title + description | Title + description + CTA |
| -- | -- | -- | -- | -- |
| image | image | image | image | image |


## After
| Description | Description + CTA | Short title + description | Title + description | Title + description + CTA |
| -- | -- | -- | -- | -- |
| image | image | image | image | image |

src/components/InLineNotification.tsx
bakoushin: I mean, why the typings for `onPressCta`/`onPressCta2` are different from this one?

src/components/InLineNotification.tsx
denvalora: Changed to the namings `CustomColors`, `severityColor`, `severityColors` in https://github.com/valora-inc/wallet/pull/4436/commits/3c25faf35a9a558099c54822172b446e10aaaaf3

src/components/InLineNotification.tsx
denvalora: I've edited my answer above to clarify, basically that's how `Warning` component was. While `Text` component provides `GestureResponderEvent` as argument, we don't really need it, nor use it. Previously `onPressCta` would ignore `event`.

I've updated the code now to match the typing from `Text`, so we can process `GestureResponderEvent` parameter in a function passed to `onPressCta` should we need to in the future.

kathaypacific: we could also have `onPress?: () => void`, right?

kathaypacific (approved):
LGTM ✨

PR #4438, opened 2023-11-06
fix: don't use non serializable fields in swap actions

jeanregisser:
### Description

This allows using Flipper again with the viem swap flow.
I shouldn't have used non serializable fields in the first place.

### Test plan

- Updated tests

### Related issues

- See Slack [thread](https://valora-app.slack.com/archives/C025V1D6F3J/p1698416818574259)

### Backwards compatibility

Yes


src/viem/preparedTransactionSerialization.ts
```diff
@@ -0,0 +1,65 @@
+// Helper functions making prepared transactions serializable
+// so they can be used in redux actions (or even stores).
+import { TransactionRequestCIP42 } from 'node_modules/viem/_types/chains/celo/types'
+
+const bigIntProps = ['value', 'gas', 'maxFeePerGas', 'maxPriorityFeePerGas', 'gatewayFee'] as const
```
jeanregisser, line 5: I used a known list for now. Because we need to know which properties to convert back from `string` to `bigint` in `getPreparedTransactions`.
This is fine as TypeScript will error if there's another `bigint` not listed.
The alternative was to store the keys converted to `string`. But maybe we'll eventually need that, once we use a more generic `TransactionRequest` for both Celo and Ethereum?

src/viem/preparedTransactionSerialization.ts
```diff
@@ -0,0 +1,65 @@
...
+
+type BigIntProps = (typeof bigIntProps)[number]
+
+type MapBigIntToString<T, K extends keyof T> = {
```
cajubelt, line 9: nit: the naming's a little weird here because this type doesn't actually deal with bigint's, or even the keys we expect to have bigint values, it seems more general than that. Maybe it could be called MapSomePropsToString, something like that

src/viem/preparedTransactionSerialization.ts
```diff
@@ -0,0 +1,65 @@
...
+  const result = { ...obj } as MapBigIntToString<T, K>
+
+  for (const key of bigIntKeys) {
+    const value = obj[key]
+    if (typeof value === 'bigint') {
+      result[key] = value.toString() as any
+    }
+  }
+
+  return result
```
cajubelt, line 26:
```suggestion
  return bigIntKeys.reduce(
    (acc, key) => {
      const value = obj[key]
      return {
        ...acc,
        [key]: typeof value === 'bigint' ? value.toString() : value,
      }
    },
    { ...obj } as MapBigIntToString<T, K>
  )
```
I was looking for a way to avoid the "as any" type cast, looks like this would work

src/swap/SwapReviewScreen.tsx
```diff
@@ -228,9 +229,19 @@ export function SwapReviewScreen({ route }: Props) {
 
     // New flow for viem
     if (quote && userInput) {
+      if (quote?.preparedTransactions?.type !== 'possible') {
```
cajubelt, line 232: could we guarantee this with static typing instead of runtime errors?

kathaypacific (approved):
TY for this!! looks good on my side

src/viem/preparedTransactionSerialization.ts
jeanregisser: Right, I had started with checking bigint types and eventually changed it, I'll rename.

src/viem/preparedTransactionSerialization.ts
jeanregisser: I'll check, but does it work because TS is getting lost somehow? 😄

src/swap/SwapReviewScreen.tsx
jeanregisser: Yes probably, this screen is going away and we'll improve that.

src/viem/preparedTransactionSerialization.ts
jeanregisser: Ok I did a couple of tests,
It does seem like TS is lost because of the spread and hence doesn't complain.
But it doesn't make the code safer.

See this: https://www.typescriptlang.org/play?#code/C4TwDgpgBAsghmAQgSwOYEkB2wAqB7AZWACdlNUAeHAGigGkoIAPYCTAEwGcoBrCEPADMoOAHxQAvFADeAKChQA2gAUoZXvyEiAugC4oq5qw7cGAfiicSZVFH04V22QF9ZswQFdMAY2DI8mFAAtggoGNic+ESk5JxUtAxGbFwaAsJiABTyUHgARgBW9tTZuWhYwHT8nPrEEHDsAQA2IPSKTgCU+vBIZdhR1uTx9OJyCrXAHsSBpeEVVQB0tewe3hBZCgoZcN7etHwg7ZIj2RtQ3gFWUABucI0e0FJ5+Yr7TqdjEBNTMifv8--bXa-U4vfh6KCgSBaG53B4SKQAchmZGACKgFkwEAA7rAEBlDvoYfdaAB6EkiAhQBoQTiYBHAM54IJgRpwMi0XIeBnIbm0+lQUioAAWwGBUFcp2cxVO0ig-3mT3FUDg3G6YXK-RilBow2y7RcQA

PR #4439, opened 2023-11-06
chore(viem-fees): get fee estimates for case where token supports comments

cajubelt:
### Description

Get fee estimates for c-stables transferWithComment method in select amount screen.

Note that since DEK registration does not actually take place before the send transaction, even if the DEK is not registered on-chain yet, DEK registration fees are not included. More context on DEK's [here](https://valora-app.slack.com/archives/C02KBT0DAHJ/p1699301693722619)

### Test plan

jest tests

tested manually, looks like this:
simulator_screenshot_F4C3AAF7-9E4C-4431-8DEF-E79FEBA1A01C

### Related issues

### Backwards compatibility

na


src/send/usePrepareSendTransactions.ts
```diff
@@ -18,32 +19,30 @@ export async function _prepareSendTransactionsCallback({
...
-      })
+    const transactionParams = {
+      fromWalletAddress: walletAddress,
+      toWalletAddress: recipientAddress,
+      sendToken: token,
+      amount: BigInt(tokenAmountInSmallestUnit(amount, token.decimals)),
+      feeCurrencies,
+    }
+    if (tokenSupportsComments(token)) {
+      return prepareTransferWithCommentTransaction(transactionParams)
```
satish-ravi, line 42: this won't include comments because we only request it on the next screen right? Can fees change based on the length of the comment?

satish-ravi (approved):

src/send/usePrepareSendTransactions.ts
cajubelt: yeah they can... I'd like to follow up on that in a separate PR since I think this one has a reasonable behavior for now

cajubelt: it's a good point though

cajubelt: one thing we could do is set a maximum length on comments on the next screen, and on this screen use a dummy comment that is of the max length, which is the most pessimistic scenario for gas fee estimation

cajubelt: it's a little tricky because we then encrypt the comment, so we can't just use the same constant in both places

PR #4441, opened 2023-11-07
fix(swap): prevent extra rerenders after fetching swap quote

kathaypacific:
### Description

I noticed a weird flicker on the swap screen.
https://github.com/valora-inc/wallet/assets/20150449/c1334b07-8733-46b1-a56b-5164ccaed7df

It is because the `refreshQuote` is invoked and the loading state is activated after the swap amount changes _and_ after the quote is fetched, but it is only part way through the `refreshQuote` function that we decide that we don't need to fetch the quote for the latter scenario.

The solution in this PR is to prevent calling `refreshQuote` when we're not supposed to, rather than try to catch this from inside the `refreshQuote` function.

### Test plan

No more flicker.

https://github.com/valora-inc/wallet/assets/20150449/c573a7a7-a841-43ed-b961-b107e9c73e55

### Related issues

n/a

### Backwards compatibility

Y


jeanregisser (approved):
Thanks! I wonder if there is also another problem where the `onSuccess` of `refreshQuote` might get called when not relevant anymore 🤔 
Since they aren't being cancelled.

jeanregisser:
> Thanks! I wonder if there is also another problem where the `onSuccess` of `refreshQuote` might get called when not relevant anymore 🤔 Since they aren't being cancelled.

I guess `react-async-hook` likely "cancels" (as in stop subscribing to) the current promise in flight whenever we call `.execute()` 
I'll double check to be sure.

dievazqu (approved):

kathaypacific:
> > Thanks! I wonder if there is also another problem where the `onSuccess` of `refreshQuote` might get called when not relevant anymore 🤔 Since they aren't being cancelled.
> 
> I guess `react-async-hook` likely "cancels" (as in stops subscribing to) the current promise in flight whenever we call `.execute()` I'll double check to be sure.

TY for raising this! i did think about this too during the implementation, i modified the swap amount where the quote fetch was kicked off, and during the refresh i changed the swap amount again. the result was the "to" amount remains in the dark grey loading state, but it does change twice. so it seems like the first request is not cancelled, but that also seems like an acceptable experience for now (we could also improve the loading state of the "to" amount rather than cancelling the first request)

PR #4442, opened 2023-11-07
fix(prepare-transactions): add `throwOnSpendTokenAmountExceedsBalance` option

jeanregisser:
### Description

I realized a bit too late that the changes in https://github.com/valora-inc/wallet/pull/4432 broke the swap flow where we want to prepare transactions when the spend amount is greater than the balance.
This is because we want users to be able to try any amount to see the quote (and associated fee).

### Test plan

- Updated tests

### Related issues

- N/A

### Backwards compatibility

Yes

### DIFF
--- a/src/swap/useSwapQuote.ts
+++ b/src/swap/useSwapQuote.ts
@@ -103,6 +103,8 @@ export async function prepareSwapTransactions(
     spendTokenAmount: new BigNumber(amountToApprove.toString()),
     decreasedAmountGasFeeMultiplier: DECREASED_SWAP_AMOUNT_GAS_FEE_MULTIPLIER,
     baseTransactions,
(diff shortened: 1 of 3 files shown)


dievazqu (approved):

cajubelt (approved):
to avoid the code smell of a boolean flag, another way of doing this would be to throw a specific error type and swallow that error type in useSwapQuote

jeanregisser:
> to avoid the code smell of a boolean parameter, another way of doing this would be to throw a specific error type and swallow that error type in useSwapQuote

Yes, here at least I could go with a named option, not the worst, though in general yes I agree we should avoid them.
I thought about exposing another function like `prepareTransactionsWithoutCheckingSpendBalance` or something. But preferred the option for now. Can revisit later of course.

PR #4443, opened 2023-11-07
fix(viem): Adding standby tx after sending

dievazqu:
### Description

Adding the standby transaction only after the transaction has been sent to the blockchain in the viem flow.

### Test plan

Run locally in a emulator and unit tests

### Related issues

- Fixes RET-896

https://github.com/valora-inc/wallet/assets/12504748/9e9e5a0e-c4f9-4503-bb58-07865bc77a2c


kathaypacific (approved):
this seems good to me, would be good to get @jeanregisser 's eyes on this too. last time i missed some details in https://github.com/valora-inc/wallet/pull/4395 🙈

src/viem/saga.ts
```diff
@@ -85,15 +81,17 @@ export function* sendPayment({
...
     // This will never happen, but Typescript complains otherwise
     if (!wallet.account) {
       throw new Error('no account found in the wallet')
     }
 
     // unlock account before executing tx
     yield* call(unlockAccount, wallet.account.address)
+  }
 
+  const addPendingStandbyTransaction = function* (hash: string) {
```
jeanregisser, line 94: In theory the `addStandbyTransaction` action could be called mutliple times, because it's wrapped in

https://github.com/valora-inc/wallet/blob/ec4623ecc279f2d609d9dfb1c40f442d67fa231f/src/transactions/send.ts#L272

However, right now this `TX_NUM_TRIES` is set to 1.

I think just to be safe, we should update the reduce to replace any standby TX which has the same hash or context id

https://github.com/valora-inc/wallet/blob/b99b375d0a52c2c28ee2bddeaff4c3f529e501eb/src/transactions/reducer.ts#L72-L83

What do you think?

jeanregisser (commented):
Thanks!

I'm slightly confused by the example video because it shows the swap flow, but right now it doesn't use `src/view/saga.ts`, which was changed in this PR.

Did I miss something?

src/viem/saga.ts
dievazqu: Sounds good!

dievazqu:
> Thanks!
> 
> I'm slightly confused by the example video because it shows the swap flow, but right now it doesn't use `src/viem/saga.ts`, which was changed in this PR.
> 
> Did I miss something?

@jeanregisser Sorry, I should have mentioned that I have tweaked the code in the video so it uses the viem flow and also throws an error because I'm skipping the approval step.

PR #4444, opened 2023-11-07
fix(fees): set tip on Celo from `eth_maxPriorityFeePerGas`

jeanregisser:
### Description

We initially got the [recommendation](https://valora-app.slack.com/archives/CNJ7KTHQU/p1697718182062359?thread_ts=1697647756.662059&cid=CNJ7KTHQU) to only set `maxFeePerGas` to the result of `eth_gasPrice` and that the difference between it and the base fee would be the tip distributed to the validator.

However after testing, we noticed that if we leave `maxPriorityFeePerGas` to `undefined` (or if we don't specify it), the final serialized tx has is set to `0`.
Resulting to `0` tip being effectively distributed to the validator. Which doesn't look right for validators 😄 

So here we bring back the original implementation I was originally [proposing](https://github.com/valora-inc/wallet/pull/4335/commits/ec11ae7f4c2d50e93eee7122a407d58b08c08386) where we set it to the result of `eth_maxPriorityFeePerGas`. Similar to what [viem does on Ethereum](https://github.com/wagmi-dev/viem/blob/291e9ba2abc1d298d77e5a66f91f85bc54a3d31b/src/actions/public/estimateFeesPerGas.ts#L144-L162).

Note: we'll wait for the confirmation from cLabs before merging this though.

### Test plan

- Updated tests

### Related issues

- N/A

### Backwards compatibility

Yes

### DIFF
--- a/src/viem/estimateFeesPerGas.test.ts
+++ b/src/viem/estimateFeesPerGas.test.ts
@@ -6,11 +6,12 @@ describe(estimateFeesPerGas, () => {
       request: jest.fn(async ({ method, params }) => {
         expect(params).toBeUndefined()
         if (method === 'eth_gasPrice') return '0x64' // 100 in hex
+        if (method === 'eth_maxPriorityFeePerGas') return '0xa' // 10 in hex
         throw new Error(`Unknown method ${method}`)
       }),
     }
     const fees = await estimateFeesPerGas(client as any)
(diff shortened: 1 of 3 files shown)


cajubelt (approved):

dievazqu (approved):

bakoushin (approved):
💯

PR #4445, opened 2023-11-07
fix(standby-txs): Removing standby txs removal

dievazqu:
### Description

Removing the removal of standby transactions when there is an error, so we can display the failed transaction in the feed.
Pending transactions without a hash will disappear after new sessions. 

This PR aims to unblock the internal release of multichain, a future PR will come with a better handling of failed txs.

### Related issues

- Fixes RET-896

https://github.com/valora-inc/wallet/assets/12504748/8a7ff374-fbab-4760-8a4d-23232a658ae3


src/swap/saga.ts
```diff
@@ -556,6 +552,10 @@ export function* swapSubmitPreparedSaga(action: PayloadAction<SwapInfoPrepared>)
       })
     )
 
+    if (swapTxReceipt.status !== 'success') {
+      throw new Error(`Swap transaction reverted: ${swapTxReceipt.transactionHash}`)
```
dievazqu, line 556: Keeping this to show the correct UI for failures and analytics.

src/viem/saga.ts
```diff
@@ -388,6 +384,9 @@ export function* sendAndMonitorTransaction({
         status: receipt.status === 'success',
       })
     )
+    if (receipt.status === 'reverted') {
+      throw new Error('transaction reverted')
+    }
```
jeanregisser, line 389: This change here affects the `transaction_confirmed` analytics event, just double checking, is this all ok?

jeanregisser (approved):
🚀

src/viem/saga.ts
dievazqu: Oh, that's true, I'm gonna move the analytics after this to keep with the current logic.

kathaypacific (commented):
i think this works but if there is actually an execution error then we will not remove the pending transaction right? so the pending tx would persist on the homefeed until the user clears their redux state with a new app session. this seems okay but i think it would be more correct to remove the pending tx if there was an actual runtime error? perhaps we can address this later

PR #4383, opened 2023-10-25
chore: remove ramp cash in bottom sheet

MuckT:
### Description

Removes `CashInBottomSheet.tsx` and related code. This feature was already disabled 100%, but was still used in e2e to prevent regressions such as https://github.com/valora-inc/wallet/issues/3043.

### Test plan

- [x] Unit tests updated
- [x] E2E tests updated

### Related issues

- ACT-886

### Backwards compatibility

Yes

PR #4446, opened 2023-11-08
revert: cash in bottom sheet removal

MuckT:
### Description

Reverts the changes from #4383

We want to keep it for now, to be able to switch it back on in the future. See Slack [thread](https://valora-app.slack.com/archives/C02898GN22V/p1699383234679679?thread_ts=1698272444.022329&cid=C02898GN22V).

### Test plan

- [x] Redux migrations updated
- [x] Existing e2e tests restored
- [x] Existing unit tests restored

### Related issues

N/A

### Backwards compatibility

Users should not have an issue upgrading from 1.70.4 to the next release; however, developers will need a fresh install due to the changes in the redux migrations.


src/redux/migrations.ts
```diff
@@ -1416,5 +1412,5 @@ export const migrations = {
       },
     }
   },
-  166: (state: any) => ({ ...state, app: { ...state.app, hideHomeBalances: false } }),
+  165: (state: any) => ({ ...state, app: { ...state.app, hideHomeBalances: false } }),
```
jeanregisser, line 1415: We shouldn't touch past migrations. Only fix forward.

jeanregisser (changes requested):

src/redux/migrations.ts
jeanregisser: It's an issue if we use a key for something, then remove the feature without removing the key from the state and then reuse the same key name for a new thing, that could have a different type.

MuckT: Fixed in 5f66a56f259a5a96ed337bd3a0dbfa283744ed3d!

src/statsig/constants.ts
```diff
@@ -53,6 +53,7 @@ export const ExperimentConfigs = {
     defaultValues: {
       onboardingNameScreenEnabled: true,
       chooseAdventureEnabled: false,
+      cashInBottomSheetEnabled: true,
```
cajubelt, line 56: this should not have been removed to begin with

cajubelt: even if it wasn't used in wallet code anymore, we need to give defaults for all experiment params when we query for them from statsig

cajubelt (approved):

src/statsig/constants.ts
kathaypacific: i thought the cash in bottom sheet is a remote config and had nothing to do with statsig....why do we need this here? i couldn't find anything related to cash in bottom sheet in statsig

jeanregisser (approved):

PR #4447, opened 2023-11-08
fix: ensure swap review screen displays gas fee in correct units

kathaypacific:
### Description

This PR fixes mismatching units for gas fee on the swap review screen. The unit should be in token units rather than fiat.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y


jeanregisser (approved):
Thanks for fixing my bug 🚀 🙏 

I guess I got confused as I saw it was using `priceUsd` in the contract-kit case.

PR #4450, opened 2023-11-08
chore: Update fiat exchange flow to use token IDs

finnian0826:
### Related issues

- Fixes #[issue number here]

### DIFF
--- a/src/fiatExchanges/SelectProvider.tsx
+++ b/src/fiatExchanges/SelectProvider.tsx
@@ -54,15 +54,15 @@ import colors from 'src/styles/colors'
 import fontStyles from 'src/styles/fonts'
 import { Spacing } from 'src/styles/styles'
 import variables from 'src/styles/variables'
-import { useTokenInfoWithAddressBySymbol } from 'src/tokens/hooks'
-import { Network } from 'src/transactions/types'
+import { useTokenInfo } from 'src/tokens/hooks'
+import { NetworkId } from 'src/transactions/types'
 import Logger from 'src/utils/Logger'
 import { CiCoCurrency } from 'src/utils/currencies'
 import { navigateToURI } from 'src/utils/linking'
-import networkConfig from 'src/web3/networkConfig'
 import { currentAccountSelector } from 'src/web3/selectors'
 import {
   CICOFlow,
+  CloudFunctionDigitalAsset,
   FiatExchangeFlow,
   LegacyMobileMoneyProvider,
   PaymentMethod,
@@ -72,7 +72,6 @@ import {
   filterLegacyMobileMoneyProviders,
   filterProvidersByPaymentMethod,
   getProviderSelectionAnalyticsData,
-  resolveCloudFunctionDigitalAsset,
 } from './utils'
 
 const TAG = 'SelectProviderScreen'
@@ -90,9 +89,8 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
   const dispatch = useDispatch()
   const {
     flow,
-    selectedCrypto: digitalAsset,
+    tokenId,
     amount: { crypto: cryptoAmount, fiat: fiatAmount },
-    network,
   } = route.params
   const userLocation = useSelector(userLocationDataSelector)
   const account = useSelector(currentAccountSelector)
@@ -103,7 +101,7 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
   const fiatConnectQuotesError = useSelector(fiatConnectQuotesErrorSelector)
   const selectFiatConnectQuoteLoading = useSelector(selectFiatConnectQuoteLoadingSelector)
   const usdToLocalRate = useSelector(usdToLocalCurrencyRateSelector)
-  const tokenInfo = useTokenInfoWithAddressBySymbol(digitalAsset)
+  const tokenInfo = useTokenInfo(tokenId)
 
   const { t } = useTranslation()
   const coinbasePayEnabled = useSelector(coinbasePayEnabledSelector)
@@ -114,12 +112,12 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
     dispatch(
       fetchFiatConnectQuotes({
         flow,
-        digitalAsset,
+        digitalAsset: CiCoCurrency[tokenInfo?.symbol],
         cryptoAmount,
         fiatAmount,
       })
     )
-  }, [flow, digitalAsset, cryptoAmount])
+  }, [flow, tokenInfo?.name, cryptoAmount])
 
   useEffect(() => {
     if (fiatConnectQuotesError) {
@@ -129,7 +127,10 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
 
   const asyncExchanges = useAsync(async () => {
     try {
-      const availableExchanges = await fetchExchanges(userLocation.countryCodeAlpha2, digitalAsset)
+      const availableExchanges = await fetchExchanges(
+        userLocation.countryCodeAlpha2,
+        tokenInfo?.symbol
+      )
 
       return availableExchanges
     } catch (error) {
@@ -149,11 +150,11 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
           userLocation,
           walletAddress: account,
           fiatCurrency: localCurrency,
-          digitalAsset: resolveCloudFunctionDigitalAsset(digitalAsset),
+          digitalAsset: CloudFunctionDigitalAsset[tokenInfo?.symbol],
           fiatAmount,
           digitalAssetAmount: cryptoAmount,
           txType: flow === CICOFlow.CashIn ? 'buy' : 'sell',
-          networkId: networkConfig.networkToNetworkId[network],
+          networkId: tokenInfo?.networkId,
         }),
         fetchLegacyMobileMoneyProviders(),
       ])
@@ -162,7 +163,7 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
         rawLegacyMobileMoneyProviders,
         flow,
         userLocation.countryCodeAlpha2,
-        digitalAsset
+        tokenInfo?.symbol
       )
       return { externalProviders, legacyMobileMoneyProviders }
     } catch (error) {
@@ -298,7 +299,8 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
           analyticsData={analyticsData}
         />
       ))}
-      {network === Network.Celo && (
+      {(tokenInfo?.networkId === NetworkId['celo-mainnet'] ||
+        tokenInfo?.networkId === NetworkId['celo-alfajores']) && (
         <LegacyMobileMoneySection
           providers={legacyMobileMoneyProviders || []}
           digitalAsset={digitalAsset}
@@ -336,7 +338,7 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
   )
 }
 
-function AmountSpentInfo({ flow, selectedCrypto, amount }: Props['route']['params']) {
+function AmountSpentInfo({ flow, tokenId, amount }: Props['route']['params']) {
   const localCurrency = useSelector(getLocalCurrencyCode)
   return (
     <View style={styles.amountSpentInfo} testID="AmountSpentInfo">
--- a/src/fiatExchanges/SelectProvider.tsx
+++ b/src/fiatExchanges/SelectProvider.tsx
@@ -62,7 +62,6 @@ import { navigateToURI } from 'src/utils/linking'
 import { currentAccountSelector } from 'src/web3/selectors'
 import {
   CICOFlow,
-  CloudFunctionDigitalAsset,
   FiatExchangeFlow,
   LegacyMobileMoneyProvider,
   PaymentMethod,
@@ -150,7 +149,7 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
           userLocation,
           walletAddress: account,
           fiatCurrency: localCurrency,
-          digitalAsset: CloudFunctionDigitalAsset[tokenInfo?.symbol],
+          digitalAsset: tokenInfo?.symbol,
           fiatAmount,
           digitalAssetAmount: cryptoAmount,
           txType: flow === CICOFlow.CashIn ? 'buy' : 'sell',
@@ -181,7 +180,7 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
     flow,
     fiatConnectQuotes,
     asyncProviders.result?.externalProviders,
-    digitalAsset
+    tokenInfo?.symbol
   )
 
   const exchanges = asyncExchanges.result ?? []
@@ -196,7 +195,7 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
     !coinbaseProvider.restricted &&
     coinbasePayEnabled &&
     appId &&
-    digitalAsset === CiCoCurrency.CELO
+    tokenInfo?.symbol === 'CELO'
 
   const anyProviders =
     normalizedQuotes.length ||
@@ -212,7 +211,7 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
     centralizedExchanges: exchanges,
     coinbasePayAvailable: coinbasePayVisible,
     transferCryptoAmount: cryptoAmount,
-    cryptoType: digitalAsset,
+    cryptoType: tokenInfo?.symbol,
   })
 
   useEffect(() => {
@@ -261,7 +260,7 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
       <View style={styles.noPaymentMethodsContainer}>
         <Text testID="NoPaymentMethods" style={styles.noPaymentMethods}>
           {t('noPaymentMethods', {
-            digitalAsset,
+            digitalAsset: tokenInfo?.symbol,
           })}
         </Text>
         <TextButton
@@ -295,7 +294,7 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
           normalizedQuotes={normalizedQuotes}
           paymentMethod={paymentMethod}
           flow={flow}
-          cryptoType={digitalAsset}
+          tokenId={tokenId}
           analyticsData={analyticsData}
         />
       ))}
@@ -303,7 +302,7 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
         tokenInfo?.networkId === NetworkId['celo-alfajores']) && (
         <LegacyMobileMoneySection
           providers={legacyMobileMoneyProviders || []}
-          digitalAsset={digitalAsset}
+          tokenSymbol={tokenInfo?.symbol}
           flow={flow}
           analyticsData={analyticsData}
         />
@@ -318,7 +317,7 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
       )}
       <ExchangesSection
         exchanges={exchanges}
-        selectedCurrency={digitalAsset}
+        selectedTokenId={tokenId}
         flow={flow}
         analyticsData={analyticsData}
       />
@@ -359,7 +358,7 @@ function AmountSpentInfo({ flow, tokenId, amount }: Props['route']['params']) {
           ) : (
             <CryptoAmount
               amount={amount.crypto}
-              currency={selectedCrypto}
+              tokenId={tokenId}
               testID="AmountSpentInfo/Crypto"
             />
           )}
@@ -415,12 +414,12 @@ function LimitedPaymentMethods({ flow }: { flow: CICOFlow }) {
 function ExchangesSection({
   exchanges = [],
   flow,
-  selectedCurrency,
+  selectedTokenId,
   analyticsData,
 }: {
   exchanges: ExternalExchangeProvider[]
   flow: CICOFlow
-  selectedCurrency: CiCoCurrency
+  selectedTokenId: string
   analyticsData: ProviderSelectionAnalyticsData
 }) {
   const { t } = useTranslation()
@@ -452,7 +451,7 @@ function ExchangesSection({
       navigate(Screens.ExchangeQR, { flow, exchanges })
     } else {
       navigate(Screens.ExternalExchanges, {
-        currency: selectedCurrency,
+        tokenId: selectedTokenId,
         exchanges,
       })
     }
@@ -493,12 +492,12 @@ function ExchangesSection({
 
 function LegacyMobileMoneySection({
   providers,
-  digitalAsset,
+  tokenSymbol,
   flow,
   analyticsData,
 }: {
   providers: LegacyMobileMoneyProvider[]
-  digitalAsset: CiCoCurrency
+  tokenSymbol: string
   flow: CICOFlow
   analyticsData: ProviderSelectionAnalyticsData
 }) {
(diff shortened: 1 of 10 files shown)


src/analytics/Properties.tsx
```diff
@@ -776,21 +776,21 @@ interface FiatExchangeEventsProperties {
   [FiatExchangeEvents.cico_landing_how_to_fund]: undefined
   [FiatExchangeEvents.cico_currency_chosen]: {
     flow: FiatExchangeFlow
-    currency: AnalyticsCurrency
+    currency: string
```
finnian0826, line 779: this is being changed from `AnalyticsCurrency` to string to allow any token symbol to be passed here. In the future, this allows for adding new CICO currencies without wallet changes.

src/fiatExchanges/FiatExchangeAmount.tsx
```diff
@@ -71,7 +67,7 @@ type Props = RouteProps
 
 function FiatExchangeAmount({ route }: Props) {
   const { t } = useTranslation()
-  const { currency, flow, network, tokenId } = route.params
+  const { flow, tokenId, symbol } = route.params
```
finnian0826, line 70: Here, `symbol` could be found from `useTokenInfo(tokenId)`, however this param is needed later in the `FiatExchangeAmount.navOptions` function which cannot call that hook, so it can also be used here.

src/fiatExchanges/quotes/FiatConnectQuote.ts
```diff
@@ -333,4 +333,8 @@ export default class FiatConnectQuote extends NormalizedQuote {
       this.flow === CICOFlow.CashIn ? this.getCryptoAmount() : this.getFiatAmount()
     )
   }
+
+  getTokenId(): string | null {
```
finnian0826, line 337: In this case (as opposed to in `ExternalQuote`) it returns string or null because there is a case where normalizeQuotes is called in a place where `tokenId` is not available, so the normalized fiat connect quote has no `tokenId`.

src/utils/currencies.ts
```diff
@@ -25,6 +25,8 @@ export const currencyForAnalytics: {
   [CiCoCurrency.cREAL]: 'cReal',
   [CiCoCurrency.ETH]: 'ETH',
 }
+export const symbolToAnalyticsCurrency = (symbol: string): string =>
+  symbol === 'CELO' ? 'cGLD' : symbol === 'cREAL' ? 'cReal' : symbol
```
finnian0826, line 29: Needed to make backwards compatible with how analytics are being done currently.

src/fiatconnect/saga.ts
```diff
@@ -731,11 +731,14 @@ export function* handleSelectFiatConnectQuote({
       crypto: parseFloat(quote.getCryptoAmount()),
       fiat: parseFloat(quote.getFiatAmount()),
     }
+    const tokenId = quote.getTokenId()
+    if (!tokenId) {
+      throw new Error(`No tokenId found in FiatConnect quote for token ${quote.getCryptoType()}`)
```
satish-ravi, line 736: is this safe to do if a fiat connect quote doesn't guarantee a tokenId being present in the quote?

src/fiatconnect/saga.test.ts
```diff
@@ -1324,6 +1328,14 @@ describe('Fiatconnect saga', () => {
...
   describe('handleAttemptReturnUserFlow', () => {
     const selectProviderParams = {
+      amount: {
+        crypto: 2,
+        fiat: 2,
+      },
+      flow: CICOFlow.CashOut,
+      tokenId: mockCusdTokenId,
+    }
+    const selectProviderParamsReturn = {
```
satish-ravi, line 1338: not sure I get this variable name, seems like its only used to construct attemptReturnUserFlow action? and is Network required?

src/fiatconnect/kyc/KycExpired.tsx
```diff
@@ -51,12 +56,11 @@ function KycExpired({ route, navigation }: Props) {
     })
     navigate(Screens.SelectProvider, {
       flow,
-      selectedCrypto: quote.getCryptoType(),
+      tokenId: tokenId,
```
satish-ravi, line 59:
```suggestion
      tokenId,
```

src/fiatconnect/kyc/KycDenied.tsx
```diff
@@ -51,12 +56,11 @@ function KycDenied({ route, navigation }: Props) {
     })
     navigate(Screens.SelectProvider, {
       flow,
-      selectedCrypto: quote.getCryptoType(),
+      tokenId: tokenId,
```
satish-ravi, line 59:
```suggestion
      tokenId,
```

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -65,6 +64,11 @@ export default function FiatConnectReviewScreen({ route, navigation }: Props) {
   const feeType = FeeType.SEND
   const tokenList: TokenBalance[] = useSelector(tokensListWithAddressSelector)
   const cryptoType = normalizedQuote.getCryptoTypeString()
+  // TODO: remove deprecated function call when FiatConnect updated to take token IDs
```
satish-ravi, line 67: Let's have a placeholder issue and link them here and other places so this isn't a TODO we forget.

src/fiatconnect/saga.test.ts
```diff
@@ -103,13 +103,21 @@ import {
...
 } from 'test/values'
 import { v4 as uuidv4 } from 'uuid'
 
+const mockUseTokenInfo = jest.fn(() => {
+  return { address: mockCusdAddress, symbol: 'cUSD', tokenId: mockCusdTokenId }
+})
+jest.mock('src/tokens/hooks', () => ({
+  useTokenInfo: () => mockUseTokenInfo(),
+  useTokenInfoWithAddressBySymbol: () => mockUseTokenInfo(),
+}))
```
satish-ravi, line 120: is this mock required?

src/fiatExchanges/quotes/ExternalQuote.test.ts
```diff
@@ -40,7 +41,10 @@ describe('ExternalQuote', () => {
       expect(
         () =>
           new ExternalQuote({
-            quote: (mockProviders[4].quote as RawProviderQuote[])[0],
+            quote: {
+              ...(mockProviders[4].quote as RawProviderQuote[])[0],
+              tokenId: mockCusdTokenId,
```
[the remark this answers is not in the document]
finnian0826: I don't think so, `mockProviders` is used in multiple different places as a return value from `fetchProviders`, which will not include a tokenId.

cajubelt (commented):
only reviewed the first few files but it looks like there's more changes needed. pausing for now.

src/fiatExchanges/FiatExchangeAmount.tsx
```diff
@@ -147,7 +133,7 @@ function FiatExchangeAmount({ route }: Props) {
 
     const previousFiatAccount = cachedFiatAccountUses.find(
       (account) =>
-        account.cryptoType === currency &&
+        account.cryptoType === resolveCICOCurrency(tokenSymbol) &&
```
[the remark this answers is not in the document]
finnian0826: `account.cryptoType` is of type `CiCoCurrency`, which won't work to compare to a tokenSymbol of type string right? I could change it to be of type string but it is part of root state so I think that would require updating the root state to expect string for that field instead.

src/fiatExchanges/FiatExchangeAmount.tsx
```diff
@@ -157,20 +143,20 @@ function FiatExchangeAmount({ route }: Props) {
       dispatch(
         attemptReturnUserFlow({
           flow,
-          selectedCrypto: currency,
+          selectedCrypto: resolveCICOCurrency(tokenSymbol),
```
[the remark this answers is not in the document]
finnian0826: In this case (as well as when a similar thing is done in `SelectProviders.tsx`) I am not quite sure what to do. This is part of the fiat connect flow, which the ticket says is out of scope for this issue. In looking through the code I'm not sure how to avoid this without also reworking all of the fiat connect quote fetching, is that what I should do?

src/fiatExchanges/quotes/normalizeQuotes.ts
```diff
@@ -81,7 +82,7 @@ export function normalizeFiatConnectQuotes(
         ([key, value]: [string, FiatAccountTypeQuoteData | undefined]) => {
           try {
             const normalizedQuote = new FiatConnectQuote({
-              quote,
+              quote: { ...quote, tokenId },
```
satish-ravi, line 85: Instead of adding the tokenId to the `quote` response object we get back from fiat connect response or external quote response, could we just pass it as a separate arg to the FiatConnectQuote or ExternalQuote classes? E.g.,

```
const normalizedQuote = new FiatConnectQuote({
  quote,
  fiatAccountType: key as FiatAccountType,
  flow,
  tokenId
}
```

That way we can avoid all the intermediate `WithTokenId` types

src/utils/currencies.ts
```diff
@@ -25,6 +25,8 @@ export const currencyForAnalytics: {
   [CiCoCurrency.cREAL]: 'cReal',
   [CiCoCurrency.ETH]: 'ETH',
 }
+export const symbolToAnalyticsCurrency = (symbol: string): string =>
```
MuckT, line 28: Might be easier to follow as a switch statement especially if it grows in the future

```TypeScript
export const symbolToAnalyticsCurrency = (symbol: string): string => {
  switch (symbol) {
    case crealTokenSymbol:
      return 'cReal'
    case celoTokenSymbol:
      return 'cGLD'
    default:
      return symbol
  }
}
```

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -334,7 +339,6 @@ function getDisplayAmounts({
   networkFee: BigNumber
   tokenInfo: TokenBalance | undefined
 }) {
-  const cryptoType = normalizedQuote.getCryptoType()
```
MuckT, line 337: Golf `const tokenId = normalizedQuote.getTokenId()` similar to what was done with `cryptoType` before

src/fiatExchanges/FiatExchangeCurrency.tsx
```diff
@@ -16,7 +16,7 @@ import { FiatExchangeEvents } from 'src/analytics/Events'
 import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
 import BackButton from 'src/components/BackButton'
 import Button, { BtnSizes, BtnTypes } from 'src/components/Button'
-import { CiCoCurrencyNetworkMap } from 'src/fiatExchanges/types'
```
cajubelt, line 19: I think this was the last use of cicocurrencynetworkmap. can you delete it?

src/fiatExchanges/quotes/ExternalQuote.ts
```diff
@@ -43,16 +42,20 @@ export default class ExternalQuote extends NormalizedQuote {
...
     flow,
+    tokenId,
   }: {
     quote: RawProviderQuote | SimplexQuote
     provider: FetchProvidersOutput
     flow: CICOFlow
+    tokenId: string
   }) {
     super()
+    this.tokenId = tokenId
```
MuckT, line 58: Should something throw here if tokenId is invalid? If so we should add a test for this.

src/fiatExchanges/FiatExchangeCurrency.tsx
cajubelt: interesting that knip isn't flagging it. cc @jophish

src/fiatExchanges/SelectProvider.tsx
```diff
@@ -129,7 +130,10 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
 
   const asyncExchanges = useAsync(async () => {
     try {
-      const availableExchanges = await fetchExchanges(userLocation.countryCodeAlpha2, digitalAsset)
+      const availableExchanges = await fetchExchanges(
+        userLocation.countryCodeAlpha2,
+        tokenInfo.symbol
```
cajubelt, line 135: does this also need to be upper cased?

cajubelt (commented):
reviewed up to SelectProvider (halfway thru). Will do the rest soon. Have you tested manually with each of the current digital assets? Maybe spoofing your location to make sure the max number of providers appear (or at least not none)? Seems important to test the integrations with cloud functions in particular and make sure providers show up when they should

src/fiatExchanges/SelectProvider.tsx
```diff
@@ -195,7 +200,7 @@ export default function SelectProviderScreen({ route, navigation }: Props) {
     !coinbaseProvider.restricted &&
     coinbasePayEnabled &&
     appId &&
-    digitalAsset === CiCoCurrency.CELO
+    tokenInfo.symbol === celoTokenSymbol
```
cajubelt, line 203: I'm not sure if using `celoTokenSymbol` is better than the string literal 'CELO', since Celo has historically had two different symbols, and it's not clear which one is meant by `celoTokenSymbol`

cajubelt: It's a little more opaque than `CiCoCurrency.CELO` since our convention with enums is to name the key the same as the value, so you know that `SomeEnum.Foo` is going to be the same as `"Foo"` (not that I think we should go back to using CiCoCurrency-- we need to deprecate that enum since it no longer serves a purpose)

src/fiatExchanges/quotes/ExternalQuote.test.ts
```diff
@@ -285,15 +307,20 @@ describe('ExternalQuote', () => {
...
       })
       expect(quote.getReceiveAmount()).toEqual(new BigNumber(20))
     })
 
     it('returns null if returned amount is not set in quote', () => {
       const quote = new ExternalQuote({
-        quote: { paymentMethod: PaymentMethod.Bank, digitalAsset: CiCoCurrency.CELO },
+        quote: {
+          paymentMethod: PaymentMethod.Bank,
+          digitalAsset: CiCoCurrency.CELO,
```
cajubelt, line 319:
```suggestion
          digitalAsset: 'CELO',
```

src/fiatExchanges/quotes/ExternalQuote.ts
```diff
@@ -76,10 +79,10 @@ export default class ExternalQuote extends NormalizedQuote {
     return isSimplexQuote(this.quote) ? this.provider.paymentMethods[0] : this.quote.paymentMethod
   }
 
-  getCryptoType(): CiCoCurrency {
+  getCryptoType(): string {
     return isSimplexQuote(this.quote)
-      ? resolveCICOCurrency(this.quote.digital_money.currency)!
-      : resolveCICOCurrency(this.quote.digitalAsset)!
+      ? this.quote.digital_money.currency!
+      : this.quote.digitalAsset!
```
cajubelt, line 85:
```suggestion
      ? this.quote.digital_money.currency
      : this.quote.digitalAsset
```
these aren't needed anymore. hooray!

cajubelt (commented):
reviewed thru NormalizedQuote. dang this is a lot of files

src/fiatExchanges/quotes/ExternalQuote.ts
jophish: wdym invalid exactly? invalid as in "this token is definitely not supported for quotes"?

also total nit, but can move this assignment down to the bottom of the constructor with the other instance variables.

src/fiatExchanges/quotes/FiatConnectQuote.ts
```diff
@@ -297,12 +301,11 @@ export default class FiatConnectQuote extends NormalizedQuote {
     return this.quote.quote.cryptoAmount
   }
 
-  getCryptoTypeString(): string {
-    // todo eventually rename this to getCryptoType and rename getCryptoType to getCryptoCurrency
+  getCryptoType(): string {
```
jophish, line 304: I think that previous comment was a bit misleading; this should technically return a `CryptoType` enum from `fiatconnect-types`, in alignment with what we do with `getFiatType`.

jophish: although i wonder if TS will complain since the interface calls for `getCryptoType(): string`

src/fiatExchanges/utils.tsx
```diff
@@ -259,12 +260,12 @@ export const filterLegacyMobileMoneyProviders = (
   providers: LegacyMobileMoneyProvider[] | undefined,
   flow: CICOFlow,
   userCountry: string | null,
-  selectedCurrency: CiCoCurrency
+  selectedCurrency: string
```
jophish, line 263: looking at how this is used (only one use, from `SelectProvider`), we can prolly use `tokenId` here instead of symbol, if we use consts like `cUsdTokenId` as suggested in another comment.

src/fiatExchanges/SelectProvider.tsx
finnian0826: I don't think so, before it was just passing in `digitalAsset` which was equivalent to the token symbol. It is uppercased below for fetchProviders because there it was changing it to uppercase before through an enum map.

src/fiatExchanges/quotes/FiatConnectQuote.ts
finnian0826: TS was okay with it in the file, but complained in other files where it was accessed and compared and expected to be a string.

src/fiatconnect/kyc/KycExpired.test.tsx
```diff
@@ -1,21 +1,19 @@
-import React from 'react'
+import { FiatAccountType, KycStatus as FiatConnectKycStatus } from '@fiatconnect/fiatconnect-types'
```
cajubelt, line 1: you may want to update your IDE settings to avoid this noisy reordering of imports. in some cases it is even possible to cause bugs by reordering imports, if imports have side effects (which is a mega anti pattern, but still crops up from time to time)

cajubelt (changes requested):
looks close, but I found one important issue

src/fiatconnect/ReviewScreen.tsx
```diff
@@ -436,17 +435,17 @@ function TransactionDetails({
     })
   const { t } = useTranslation()
   let tokenDisplay: string
-  switch (normalizedQuote.getCryptoType()) {
-    case CiCoCurrency.cUSD:
+  switch (normalizedQuote.getTokenId()) {
```
satish-ravi, line 438: I know refactoring FC isn't in scope, but could we just use tokenInfo.name here? I think the only difference would be for celo which would be Celo instead of CELO which seems fine. It would help avoiding having more references to hardcoded token ids in code, which I think we should try to minimize. I am going to make a ticket to avoid hardcoding token ids / addresses in the wallet separately.

src/tokens/selectors.ts
```diff
@@ -357,7 +359,13 @@ export const swappableTokensByNetworkIdSelector = createSelector(
 export const cashInTokensByNetworkIdSelector = createSelector(
   (state: RootState, networkIds: NetworkId[]) => tokensListSelector(state, networkIds),
   (tokens) =>
-    tokens.filter((tokenInfo) => tokenInfo.isCashInEligible && isCicoToken(tokenInfo.symbol))
+    tokens.filter(
+      (tokenInfo) =>
+        tokenInfo.isCashInEligible &&
+        (getFeatureGate(StatsigFeatureGates.USE_CICO_CURRENCY_BOTTOM_SHEET) //TODO: Remove after CiCo currency bottom sheet is rolled out
+          ? isCicoToken(tokenInfo.symbol)
+          : true)
```
cajubelt, line 367:
```suggestion
        (getFeatureGate(StatsigFeatureGates.USE_CICO_CURRENCY_BOTTOM_SHEET) //TODO: Remove after CiCo currency bottom sheet is rolled out
          || isCicoToken(tokenInfo.symbol))
```
I think your logic is buggy here. Please add a regression test. 

Also, using a boolean expression to gate whether to return a boolean is not necessary-- you can just return the expression itself

src/tokens/selectors.ts
```diff
@@ -372,7 +380,9 @@ export const cashOutTokensByNetworkIdSelector = createSelector(
         ((showZeroBalanceTokens ? tokenInfo.showZeroBalance : false) ||
           tokenInfo.balance.gt(TOKEN_MIN_AMOUNT)) &&
         tokenInfo.isCashOutEligible &&
-        isCicoToken(tokenInfo.symbol)
+        (getFeatureGate(StatsigFeatureGates.USE_CICO_CURRENCY_BOTTOM_SHEET) //TODO: Remove after CiCo currency bottom sheet is rolled out
```
cajubelt, line 383: same here.

src/tokens/selectors.ts
cajubelt: if the bottom sheet is not shown, that's when it matters that `isCicoToken(tokenInfo.symbol)` evaluates to true

finnian0826: @cajubelt I thought it mattered in the case when the bottom sheet is shown? When the bottom sheet is not shown, this helper is not used since the token list is hard coded. I thought the idea was to have this here so that if the bottom sheet is used it still shows the same tokens as when the bottom sheet is not used, in the case that a new CICO token is added.

finnian0826: that makes sense, I think I was a bit confused by Satish's comment because he said to retain the check when the feature flag is true, but looking closer it seems like the check is necessary on the token details screen only when the flag is false, so I changed it to be that case like you suggested.

src/tokens/selectors.ts
```diff
@@ -356,8 +355,7 @@ export const swappableTokensByNetworkIdSelector = createSelector(
 
 export const cashInTokensByNetworkIdSelector = createSelector(
   (state: RootState, networkIds: NetworkId[]) => tokensListSelector(state, networkIds),
-  (tokens) =>
-    tokens.filter((tokenInfo) => tokenInfo.isCashInEligible && isCicoToken(tokenInfo.symbol))
```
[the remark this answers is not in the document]
cajubelt: > Maybe the safest thing to do would be to retain this check if the bottom sheet feature gate is true

wouldn't the check be needed when the bottom sheet feature gate is false? since that's when the old screen is shown which doesn't handle tokens where `isCicoToken` evaluates to false?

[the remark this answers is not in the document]
cajubelt: @satish-ravi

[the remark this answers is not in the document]
satish-ravi: oops yes, I meant when the feature gate is false

src/tokens/TokenDetails.tsx
```diff
@@ -193,15 +192,11 @@ export const useActions = (token: TokenBalance) => {
...
-        // this should always be true given that we only show Add / Withdraw if a
-        // token is CiCoCurrency, but adding it here to ensure type safety
         if (isCicoToken(tokenSymbol)) {
           navigate(Screens.FiatExchangeAmount, {
-            // TODO(ACT-954): only pass token id
-            currency: tokenSymbol,
             tokenId: token.tokenId,
             flow: CICOFlow.CashIn,
-            network: networkIdToNetwork[token.networkId],
+            tokenSymbol: token.symbol,
```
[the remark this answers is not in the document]
satish-ravi: Not a blocker, but if you need to update nav options based on redux state / hooks, you can use the `useLayoutEffect` hook. An example here https://github.com/valora-inc/wallet/blob/93db460b392841164bf0b886c7831ba672e71c9e/src/onboarding/registration/EnableBiometry.tsx#L63-L80

PR #4451, opened 2023-11-09
chore(fees): use pessimistic comment length for fee estimation in send flow

cajubelt:
### Description

Longer comments means higher gas costs for transferWithComment transactions (to pay for higher memory footprint). We encrypt comments before sending, and include metadata like encryption public keys, making the length of the comment a bit hard to predict, even though comments are limited to 70 characters. I created a transaction with a max-length comment and found it to be about 600 characters, so this PR uses a 640-character placeholder comment to estimate fees. 

### Test plan

- jest tests
- tested manually and made sure that transactions succeeded using a 640-character placeholder comment to calculate gas limits

### Related issues

na

### Backwards compatibility

na


satish-ravi (approved):

jeanregisser (commented):
👍

PR #4452, opened 2023-11-09
feat: show bridge info in swap token bottom sheet

kathaypacific:
### Description

As the title. As we think about marketing expanded swap assets, it'll be important to be able to distinguish the bridge of any wrapped tokens.

### Test plan

Simulator Screenshot - iPhone 14 Pro - 2023-11-09 at 10 46 21

### Related issues

Related to  RET-865

### Backwards compatibility

Y


jeanregisser (approved):
🚀

PR #4453, opened 2023-11-09
feat(prepare-transactions): multichain support

jeanregisser:
### Description

`prepareTransactions` is now ready for multichain ⛓️ 

### Test plan

- Updated tests
- Manually tested send (new flow) + swap with viem

### Related issues

- Fixes RET-912

### Backwards compatibility

Yes


src/viem/estimateFeesPerGas.test.ts
```diff
@@ -25,5 +38,30 @@ describe(estimateFeesPerGas, () => {
...
+  })
+
+  it('should return the default fees per gas on other networks when fee currency is specified', async () => {
+    jest
+      .mocked(defaultEstimateFeesPerGas)
+      .mockResolvedValue({ maxFeePerGas: BigInt(110), maxPriorityFeePerGas: BigInt(10) })
+    const client = {
+      chain: { id: 1 },
+    }
+    const fees = await estimateFeesPerGas(client as any, '0x123')
```
dievazqu, line 63: nit: As far as I understand the code, passing the `feeCurrency` here is pointless.
I wonder if this might be confusing, for example, I want to use some specific token as fee currency, but passing it here won't do the job.

I'm not sure how is the best way to solve this, since currently only Celo supports this feature.
I'm wondering if we should just throw if `feeCurrency` is not undefined for non-Celo transactions.
I think the behavior should be similar to what happens when we pass a invalid `feeCurrency` for Celo. Does `eth_gasPrice` fail when the `feeCurrency` is invalid?

dievazqu (approved):
LGTM

src/viem/estimateFeesPerGas.test.ts
jeanregisser: Yes good idea to throw if not Celo.

src/viem/estimateFeesPerGas.ts
```diff
@@ -1,17 +1,26 @@
...
+export async function estimateFeesPerGas(
+  client: Client,
+  feeCurrency?: Address
+): Promise<{ maxFeePerGas: bigint; maxPriorityFeePerGas: bigint }> {
+  // Custom path for Celo that can be removed once it's supported in viem
+  // See https://github.com/wagmi-dev/viem/discussions/914
+  if (client.chain?.id === networkConfig.viemChain.celo.id) {
+    // The gasPrice returned on Celo is already roughly 2x baseFeePerGas
+    // See this interesting thread for more details:
+    // https://valora-app.slack.com/archives/CNJ7KTHQU/p1697717100995909?thread_ts=1697647756.662059&cid=CNJ7KTHQU
```
cajubelt, line 14: I read through this thread and didn't find it very clear.

cajubelt: if you do want to link to it, maybe linking to a comment in the thread that summarizes what was covered there would be helpful

src/viem/prepareTransactions.ts
```diff
@@ -71,17 +79,19 @@ export function getFeeCurrencyAddress(feeCurrency: TokenBalance) {
  * @param maxPriorityFeePerGas
  */
 export async function tryEstimateTransaction({
+  client,
```
cajubelt, line 82: please update the jsdoc

src/viem/prepareTransactions.ts
```diff
@@ -94,8 +104,8 @@ export async function tryEstimateTransaction({
 
   // TODO maybe cache this? and add static padding when using non-native fee currency
   try {
-    tx.gas = await publicClient.celo.estimateGas({
-      ...tx,
+    tx.gas = await estimateGas(client, {
+      ...(tx as TransactionRequestBase),
```
cajubelt, line 108: If you give `tx` the type `TransactionRequestBase` explicitly, there is a TS error. It would be great to avoid type casting to a type that we know doesn't actually work for the data.

cajubelt: I tinkered with this and found something that kinda works:
- use `TransactionRequest` type for `tx` (e.g. `const tx: TransactionRequest = { ... `)
- estimateGas doesn't accept 'cip64' as a "type" on tx. Instead of covering this up with a type cast, explicitly omit type if it's 'cip64'. `type: tx.type === 'cip64' ? undefined : tx.type,`. Should probably report this to viem and link to the issue in a code comment as well.
- update the Logger.info with this: `feeCurrency: 'feeCurrency' in tx ? tx.feeCurrency : undefined,` to avoid a type error

cajubelt: I think this makes our type workarounds more transparent and readable, while keeping the static typing guarantees for all other fields in `tx` besides the troublesome `type` field

src/swap/getFeeCurrency.tsx
```diff
@@ -1,6 +1,20 @@
...
-export function getFeeCurrency(preparedTransactions: TransactionRequestCIP42[]) {
-  // The prepared transactions always use the same fee currency
-  return preparedTransactions[0].feeCurrency
+export function getFeeCurrency(preparedTransactions: TransactionRequest[]): Address | undefined
+export function getFeeCurrency(preparedTransaction: TransactionRequest): Address | undefined
+export function getFeeCurrency(x: TransactionRequest[] | TransactionRequest): Address | undefined {
+  let preparedTransaction: TransactionRequest
+  if (Array.isArray(x)) {
+    // The prepared transactions always use the same fee currency
+    preparedTransaction = x[0]
```
cajubelt, line 10: nit: might be worth throwing an error if they don't

cajubelt (commented):
thanks for doing this, excited to see the framework get extended with relatively little code needed to support the new use cases

src/viem/prepareTransactions.ts
cajubelt: well anyway. looks like this one's actually pretty hard. I'd like to minimize type casting, but we may have to settle for some hacky workaround for the sake of expediency

src/viem/estimateFeesPerGas.ts
jeanregisser: Yes the thread is a bit confusing now, was meant to link to some context at least.

src/viem/prepareTransactions.ts
jeanregisser: Right 😅

src/viem/prepareTransactions.ts
jeanregisser: Yes exactly. I try to avoid casting as much as possible, until it's too "much".
Thanks for trying.

src/swap/getFeeCurrency.tsx
jeanregisser: Yep was thinking about this too. Adding it now. Thanks.

src/viem/prepareTransactions.ts
```diff
@@ -396,3 +397,34 @@ export function getFeeCurrencyAndAmount(
     feeCurrency,
   }
 }
+
+/**
+ * Given prepared transaction(s), get the fee currency set.
+ *
+ * NOTE: throws if the fee currency is not the same for all transactions
+ */
+export function getFeeCurrency(preparedTransactions: TransactionRequest[]): Address | undefined
```
cajubelt, line 406: I didn't know you could do this function overloading in TS. Cool

cajubelt (approved):

PR #4455, opened 2023-11-09
feat(send): show contacts on new recipient screen

satish-ravi:
### Description

On clicking the "Send crypto to your friends", open the contacts page if permissions are granted. Also made a V2 RecipientPicker which uses the V2 recipient item.

Showing valora icon, showing permissions modal and navigation on selecting recipients will be in follow up PRs

### Test plan

Unit tests, manual

https://github.com/valora-inc/wallet/assets/5062591/23fa3d60-355f-4acf-b5e1-2e487f0c0265

Note that the Get started box pins to the bottom, this is the behavior on main and will be fixed in a separate PR

### Related issues

- Part of ACT-952

### Backwards compatibility

N/A

### DIFF
--- a/src/recipients/RecipientPickerV2.test.tsx
+++ b/src/recipients/RecipientPickerV2.test.tsx
@@ -1,0 +1,86 @@
+import { render } from '@testing-library/react-native'
+import * as React from 'react'
+import { ReactTestInstance } from 'react-test-renderer'
+import RecipientPicker from 'src/recipients/RecipientPickerV2'
+import { mockRecipient, mockRecipient2, mockRecipient3 } from 'test/values'
+
+const mockRecipients = [mockRecipient, mockRecipient2, mockRecipient3]
+
+describe('RecipientPickerV2', () => {
+  it('renders all recipients', () => {
+    const { getAllByTestId } = render(
+      <RecipientPicker
+        recipients={mockRecipients}
+        onSelectRecipient={jest.fn()}
+        selectedRecipient={null}
+        isSelectedRecipientLoading={false}
+      />
+    )
+
+    expect(getAllByTestId('RecipientItem')).toHaveLength(3)
+  })
+
+  it('renders all recipients with title', () => {
+    const { getAllByTestId, getByText } = render(
+      <RecipientPicker
+        title="mockRecipientTitle"
+        recipients={mockRecipients}
+        onSelectRecipient={jest.fn()}
+        selectedRecipient={null}
+        isSelectedRecipientLoading={false}
+      />
+    )
+
+    expect(getByText('mockRecipientTitle')).toBeTruthy()
+    expect(getAllByTestId('RecipientItem')).toHaveLength(3)
+  })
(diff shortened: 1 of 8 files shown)


src/send/SendSelectRecipient.tsx
```diff
@@ -138,32 +140,29 @@ function SendSelectRecipient() {
...
-    if (shouldImportContacts) {
-      // TODO (ACT-949): Currently if a user does not have a phone number linked, this silently fails
-      const permissionGranted = await requestContactsPermission()
-      if (permissionGranted) {
-        dispatch(importContacts())
-      }
-      setShouldImportContacts(false)
+  const onPressContacts = async () => {
+    ValoraAnalytics.track(SendEvents.send_select_recipient_contacts)
+    const permissionGranted = await requestContactsPermission()
```
jophish, line 160: Just to check, on iOS, the behavior here is:
* If permissions have already been granted, this does nothing and we just show contacts.
* If permissions have been previously denied, this will not show a native modal, and just silently return false.
* If permissions have been neither granted nor denied previously, this will show a native modal.

Is that understanding correct? How do we differentiate between the second bullet and the case where the user rejects the native modal in the third case?

On Android, we can call `requestContactsPermission` as many times as we like, and the native modal will appear each time, right?

src/recipients/RecipientPickerV2.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+        onSelectRecipient={jest.fn()}
+        selectedRecipient={mockRecipient}
+        isSelectedRecipientLoading={true}
+      />
+    )
+
+    expect(getAllByTestId('RecipientItem')).toHaveLength(3)
+    expect(getAllByTestId('RecipientItem/ActivityIndicator')).toHaveLength(1)
+    expect(
+      (getAllByTestId('RecipientItem')[0].children[0] as ReactTestInstance).props.style[1]
```
jophish, line 74: Interesting, don't think I've seen this pattern before (checking individual styles on components). Is this why we need to cast to `ReactTestInstance` here?

jophish (approved):

src/send/SendSelectRecipient.tsx
satish-ravi: yes your understanding on iOS behavior is correct. I think we have an API that returns "authorized" for 1, "declined" for 2 and "undefined" for 3.
For Android, I am still digging into this but I think if the user denies multiple times, it gets into a "never_ask_again" state, in which case `requestContactsPermission` will silently fail like the iOS denied case.

src/recipients/RecipientPickerV2.test.tsx
satish-ravi: the cast was required because the return type is `string | ReactTestInstance`. I guess I could've just added another testID in the recipient item component for selected case and asserted for it. But both required knowing the internals of the RecipientItem component, so didn't feel strongly either way.

PR #4457, opened 2023-11-10
fix(send): styles on select recipient screen

satish-ravi:
### Description

Fixed Get Started section to match designs [here](https://www.figma.com/file/URF0fqjWMR6ffWDOcFK8ic/Send-Flow?node-id=3095%3A2103&mode=dev)
Changes include:
- Dropping wrapper views for get started and recent recipients and moving the styles to the corresponding components
- Dropping default styles (`flexDirection = 'column'`, `justifyContent = 'flex-start'`, etc)
- Using Spacing constants where possible
- Using `gap` over `padding` to space elements in a flex container
- Adding flex: 1 to components housing Text so they don't overflow

### Test plan

| View | Before | After |
|--------|:--------:|:--------:|
| Get Started | |  |
| Recent recipients |  |  | 
| Large font size |  |  | 
| Android touchable | <video src="https://github.com/valora-inc/wallet/assets/5062591/ff68152c-7fce-4b96-a339-1417c896c217" width="250"> | <video src="https://github.com/valora-inc/wallet/assets/5062591/41ade14b-5c62-47df-85f5-da9700110745" width="250"> | 

(Note: there's no change in the Recent recipients view, just included above to show that the changes don't break anything)

### Related issues

N/A

### Backwards compatibility

Yes


cajubelt (approved):
can you tag whoever did this screen originally so they can take note?

src/send/SendSelectRecipient.tsx
```diff
@@ -81,20 +81,25 @@ function GetStartedSection() {
 
   return (
     <View style={getStartedStyles.container} testID={'SelectRecipient/GetStarted'}>
-      <Text style={getStartedStyles.subtitle}>{t('sendSelectRecipient.getStarted.subtitle')}</Text>
```
MuckT, line 84: This order and naming threw me through a loop for a bit, `headline` and `title` are probably more appropriate.

satish-ravi:
> can you tag whoever did this screen originally so they can take note?

@jophish

MuckT (commented):
It might be outside of the focus of this PR but we should probably use a ScrollView instead of View on `SendSelectRecipient.tsx`. There might be some issues with the text input, but that can be address in #4404

I also see some text overflow with `SelectRecipientButton`.

src/send/SendSelectRecipient.tsx
satish-ravi: I think this can be fixed in a separate PR

satish-ravi:
@MuckT I've fixed the clipping issue for the SelectRecipientButton. The scrolling issue on large fonts can be fixed in a separate PR, we also have issues when the keyboard shows up. And the search input fix is being done by @jophish in a separate PR

jophish (approved):

MuckT (approved):
LGTM!

PR #4459, opened 2023-11-10
test(e2e): properly wait for WC requests + add logs

jeanregisser:
### Description

We have a few occasional WC tests failing:
```
    TypeError: Cannot read properties of null (reading 'status')

      232 |
      233 |         // Wait for transaction and get receipt
    > 234 |         const *** status, from, to *** = await kit.connection.getTransactionReceipt(txHash)
          |                 ^
      235 |
      236 |         jestExpect(status).toStrictEqual(***)
      237 |         jestExpect(from).toStrictEqual(walletAddress)
```

I'm suspecting it's because we don't properly wait for the request results.
This tries to fix it and add logs.

### Test plan

- e2e tests pass

### Related issues

- See Slack [thread](https://valora-app.slack.com/archives/C02E2FE98P2/p1699569663993199)

### Backwards compatibility

Yes


kathaypacific (approved):
💅🏼 TY!!

PR #4460, opened 2023-11-10
chore: remove apollo and related graphql libs

jeanregisser:
### Description

This fully removes Apollo from our code base.

There were 2 requests left using Apollo client:
- fetching the local exchange rate
- fetching token balances

We don't need Apollo for these simple needs. I switched them to simple requests using `fetch` like we did for our TX feed.
Also we may switch these to REST in the future anyway.

The Apollo libs hadn't been updated in 3+ years (😅). Now they are gone.
This should help with our goal to update old deps.

### Test plan

- Tests pass
- Manually tested related requests still work

### Related issues

- Tech debt

### Backwards compatibility

Yes


src/tokens/saga.ts
```diff
@@ -133,26 +132,32 @@ export async function fetchTokenBalancesForAddress(
...
-        fetchPolicy: 'network-only',
-        errorPolicy: 'all',
+          `,
+          variables: {
+            address,
+            networkId: networkId.replaceAll('-', '_'), // GraphQL does not support hyphens in enum values
+          },
+        }),
       })
+      return (await response.json()) as { data: UserBalancesResponse }
```
kathaypacific, line 160: do we want to check for the status code and throw if not 200? i think it doesn't make too much of a difference since the consumers are wrapping this with try/catch, just wanted to check if we've thought about it (since apollo always returns a 200)

src/utils/moneyAmount.ts
```diff
@@ -0,0 +1,16 @@
+import BigNumber from 'bignumber.js'
+
+export interface LocalMoneyAmount {
+  value: BigNumber.Value
+  currencyCode: string
+  exchangeRate: BigNumber.Value
+}
+
+/**
+ * @deprecated Use TokenAmount instead
```
kathaypacific, line 10: do we maybe want to put these types with TokenAmount then?

kathaypacific (approved):
YAY ✂️🧹💅🏼

src/tokens/saga.ts
jeanregisser: Good catch! I'll add this 👍

src/utils/moneyAmount.ts
jeanregisser: I'm not too sure yet, TokenAmount is currently within `src/transactions/types`, this type is more generic.

PR #4462, opened 2023-11-10
fix(send): avoid showing invite modal for contacts that have an address

cajubelt:
### Description

When a contact is selected and a CPV lookup is done for an address with that contact's phone number, the recipient object given to the SendAmount screen as a prop is not filled in with an address, even if an address is found for them. Instead, their verification status is just [updated](https://github.com/valora-inc/wallet/blob/9393d6a4bfd2e5425394e60710e275b0ce30f149/src/send/useFetchRecipientVerificationStatus.ts#L34) to "VERIFIED" and the user is ushered along in the flow, and it is expected that redux state is used later on to fill in the recipient address. This breaks an assumption I made in #4415 that verified recipients will always have addresses.

Note: contact address lookups for the "new" send flow are still broken, but that is behind a feature flag and incomplete anyway. Since this bug is on the release branch, we need to get a stopgap out sooner, and we can follow up to fix the new flow later.

### Test plan

- jest
- tested manually to verify that invite modal does not erroneously appear when an address exists for that PN

### Related issues

na

### Backwards compatibility

na


satish-ravi (approved):

src/send/Send.tsx
```diff
@@ -126,22 +128,27 @@ function Send({ route }: Props) {
...
 
     if (getFeatureGate(StatsigFeatureGates.USE_NEW_SEND_FLOW)) {
-      navigate(Screens.SendEnterAmount, {
-        isFromScan: false,
-        defaultTokenIdOverride,
-        forceTokenId,
-        recipient,
-        origin: SendOrigin.AppSendFlow,
-      })
+      if (hasAddressField(recipient)) {
```
jeanregisser, line 137: nit: looks like we have duplicated helpers for this
https://github.com/valora-inc/wallet/blob/7ff040fa585fa3beb70e2a2b2ee0abe9d1558ad3/src/recipients/recipient.ts#L52-L54


```suggestion
      if (recipientHasAddress(recipient)) {
```

jeanregisser, line 137: Does it still handle secure send?

jeanregisser (approved):
Thanks!
It's the 2nd time we break send to phone number.
Good that it was caughter early this time.

We have a task to add an e2e test but didn't yet get to it.
https://linear.app/valora/issue/RET-854/add-send-to-phone-number-tests

src/send/Send.tsx
cajubelt: no-- we hadn't ported that over to the new flow yet, so this PR doesn't change anything. we're handling it this upcoming sprint

PR #4463, opened 2023-11-13
test(e2e): wait for TX receipt

jeanregisser:
### Description

Follow up for https://github.com/valora-inc/wallet/pull/4459 since the failure still [happened](https://github.com/valora-inc/wallet/actions/runs/6848121582/job/18617673592) after merging it, but this time it was clear it wasn't because of an `undefined` TX hash.

Note: this shows there's a slight difference in our implementation when using ContractKit vs viem.
i.e. with ContractKit we are waiting for the receipt in the wallet, but not with viem. I think this is fine though.

### Test plan

- Updated e2e test

### Related issues

- N/A

### Backwards compatibility

Yes


e2e/src/usecases/WalletConnectV2.js
```diff
@@ -22,6 +23,11 @@ const dappName = 'WalletConnectV2 E2E'
 const kitUrl = process.env.FORNO_URL || 'https://alfajores-forno.celo-testnet.org'
 const kit = newKit(kitUrl)
 
+const celoClient = createPublicClient({
+  chain: celo,
```
bakoushin, line 27: I wonder if chain here should indeed be `celo`? Because the testcases are using Alfajores chain id `44787`.

bakoushin (approved):
LGTM!

Thanks for fixing! Left one small not blocking question.

e2e/src/usecases/WalletConnectV2.js
jeanregisser: Yes you're right technically it should be `celoAlfajores`. Though for the current usage it doesn't matter, I'm changing it to avoid surprises.

PR #4464, opened 2023-11-13
chore: add inviterAddress to wallet created event

kathaypacific:
### Description

The `opened_via_invite_url` is fired on app open and does not have the invitee address, which makes connecting the inviter with invitee difficult. This PR adds the inviter address if it exists to the initialize wallet event, so that we can do the attribution at the time when a wallet is created. 

### Test plan

Tested manually that the event property shows up as expected

### Related issues

n/a

### Backwards compatibility

Y


bakoushin (approved):
LTMG

PR #4465, opened 2023-11-13
chore: use node 20

jeanregisser:
### Description

Node 16 has reached [End-of-Life](https://endoflife.date/nodejs).

Should also unblock https://github.com/valora-inc/wallet/pull/4463 (see this other [thread](https://valora-app.slack.com/archives/C02898GN22V/p1699893928107209?thread_ts=1699874627.172549&cid=C02898GN22V)).

### Test plan

- CI passes

### Related issues

- Fixes ENG-96
- See also this Slack [thread](https://valora-app.slack.com/archives/C025V1D6F3J/p1699530192936749?thread_ts=1699458183.451969&cid=C025V1D6F3J).

### Backwards compatibility

Yes


.github/workflows/check.yml
```diff
@@ -26,15 +26,16 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
+        id: setup-node
         with:
-          node-version: '16'
+          node-version-file: 'package.json'
```
jeanregisser, line 31: This uses `package.json`'s `engines` to determine which node version to use.

This way we'll have less places to touch next time we upgrade.

package.json
```diff
@@ -4,6 +4,9 @@
   "author": "Valora Inc",
   "license": "Apache-2.0",
   "private": true,
+  "engines": {
+    "node": "^20"
+  },
```
jeanregisser, line 9: This enforces the use of node 20.x (not <20, but not >20).

.github/workflows/test.yml
```diff
@@ -21,15 +21,16 @@ jobs:
...
-          node-version: '16'
+          node-version-file: 'package.json'
           check-latest: true
       - uses: actions/cache@v3
         with:
           path: |
             node_modules
             */*/node_modules
-          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock', 'patches/*.patch') }}
+          key: ${{ runner.os }}-node-${{ steps.setup-node.outputs.node-version }}-${{ hashFiles('**/yarn.lock', 'patches/*.patch') }}
```
jeanregisser, line 33: I initially got issues on the PR, because it was using cached `node_modules` produced with a different version of node. This ensures the cache depends on the node version used to avoid unexpected issues.

jeanregisser:
Note: our workflows could be more dry with node setup + cache. We can address this separately.

WALLET.md
```diff
@@ -62,8 +62,8 @@ Install Node version listed in [.nvmrc](.nvmrc) and make it default (example for
 
 ```bash
 # Check actual version number in .nvmrc
-nvm install 16.20.2
-nvm alias default 16.20.2
+nvm install 20
+nvm alias default 20
 ```
```
denvalora, line 67: The version number can be omitted. Also the alias can be set via `--default` option.
```md
nvm install [<version>]                     Download and install a <version>. Uses .nvmrc if available and version is omitted.
  --default                                 After installing, set default alias to the version specified. (same as: nvm alias default <version>)
```

```suggestion
nvm install --default
```

src/nfts/saga.test.tsx
```diff
@@ -60,8 +60,7 @@ describe('Given Nfts saga', () => {
         ])
         .put(
           fetchNftsFailed({
-            error:
-              'invalid json response body at  reason: Unexpected token i in JSON at position 0',
+            error: `invalid json response body at  reason: Unexpected token 'i', "invalid json" is not valid JSON`,
```
denvalora, line 63: Why did this change?

package.json
denvalora: Should we enforce this in `.npmrc` file via `engine-strict=true`? https://docs.npmjs.com/cli/v10/using-npm/config#engine-strict

src/nfts/saga.test.tsx
jeanregisser: Apparently different node versions use different error messages for `JSON.parse`.

package.json
jeanregisser: `yarn` works with only engines. Is there an advantage to using `.npmrc`?

.github/workflows/check.yml
denvalora: This would mean Node with `^20` version, which will change over time. There may be minor changes, maybe bugs. While in the development environment we've pinned `20.9.0` in  `.nvmrc`.

I'm leaning towards using `node-version-file: '.nvmrc'` to keep the versions consistent across the environments. An alternative would be Volta which we don't use.

jeanregisser: Good points, but the downside of `.nvmrc` is it's not really enforced. 
What do you think of pinning node version in the `engines` field then?

package.json
denvalora: Pardon for the mistake. I think I was wrong as `.npmrc` do not seem to be used in `yarn` projects.

I've seen projects also use Volta instead of nvm, and they define the engine version and Volta's node version separately in the package.json (https://runme.dev/blog/nodejs-version-managers-nvm-volta-asdf). It automatically sets the correct node version and environment without using `nvm use`. It may be nice to look into in the future.

cajubelt (approved):
🥳 🥳 🥳

PR #4466, opened 2023-11-13
feat(send): show valora icon for phone number recipients

satish-ravi:
### Description

Show valora icon for phone number contacts if we can map it to an address. Showing the Send / Invite button and showing icon for address contacts will be in a follow up PR

### Test plan

Unit tests, manual

Video recording: https://drive.google.com/file/d/1DaMreLTRTGjxRtWUnzaLBdwuzRNPg79j/view?usp=sharing

Includes clicking 3 types of contacts in order:

1. Contact which has never been lookup before and mapped to an address
2. Contact with no phone number mapping 
3. Contact which already has a mapping - momentarily removes the icon and then brings it back up once the lookup is complete

### Related issues

- Part of ACT-952

### Backwards compatibility

N/A

### DIFF
--- a/src/recipients/RecipientItemV2.tsx
+++ b/src/recipients/RecipientItemV2.tsx
@@ -1,11 +1,18 @@
-import React, { memo } from 'react'
+import React, { memo, useMemo } from 'react'
 import { useTranslation } from 'react-i18next'
 import { ActivityIndicator, StyleSheet, Text, View } from 'react-native'
 import ContactCircle from 'src/components/ContactCircle'
 import Touchable from 'src/components/Touchable'
 import Logo, { LogoTypes } from 'src/icons/Logo'
 import QuestionIcon from 'src/icons/QuestionIcon'
-import { Recipient, getDisplayDetail, getDisplayName } from 'src/recipients/recipient'
+import { e164NumberToAddressSelector } from 'src/identity/selectors'
+import {
+  Recipient,
+  RecipientType,
+  getDisplayDetail,
+  getDisplayName,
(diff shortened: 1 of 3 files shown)


src/recipients/RecipientItemV2.tsx
```diff
@@ -14,19 +21,26 @@ interface Props {
...
 
-function RecipientItem({ recipient, onSelectRecipient, loading, showValoraIcon, selected }: Props) {
+function RecipientItem({ recipient, onSelectRecipient, loading, selected }: Props) {
   const { t } = useTranslation()
 
   const onPress = () => {
     onSelectRecipient(recipient)
   }
 
+  const e164NumberToAddress = useSelector(e164NumberToAddressSelector)
```
satish-ravi, line 36: One downside of using this state is that whenever we lookup a PN mapping for a contact that has already been looked up and has an address mapping, the mapping is cleared before being populated again once we get back a response. This causes a small flash of the icon (see 3rd contact click on the screen recording attached). One option to avoid this is to have a separate list of known valora contacts in redux state which is populated along with the `e164NumberToAddress` state. We can't change the behavior of `e164NumberToAddress` state because the value being `undefined` acts as a proxy for "loading" state for a PN being looked up. Thoughts?

satish-ravi: yeah we could track that differently but that would mean changes to the existing recipient screen, since we use this state to show the loading indicator next to a recipient. I'll write up an issue for this with the couple of ideas we have.

PR #4468, opened 2023-11-13
feat(send): Implement new Send Search UI

jophish:
### Description

Part of ACT-950. Implements the UI components of the new search flow, including the paste button, search result logic, and send/invite button.

Currently, this does _not_ account for the new address verification scheme we have planned. (e.g., we still consider all addresses to be "verified" and should in theory show the Valora V).

### Test plan

#### Search for Verified Phone Number 

https://github.com/valora-inc/wallet/assets/569401/f65c32c0-a382-498d-ae38-9671f4b90b82

#### Search for Unverified Phone Number

https://github.com/valora-inc/wallet/assets/569401/f843343e-921b-48c2-8a9b-7a436af6f679

#### Search for Address (includes paste button behavior & empty state)

https://github.com/valora-inc/wallet/assets/569401/4f0e6b51-402c-43d6-85a8-d3b1a4371b55

#### Search for Nomspace Address

https://github.com/valora-inc/wallet/assets/569401/ffccb9b6-19f1-4678-a76d-9e4916bf60ca

### Related issues

- Fixes ACT-952

### Backwards compatibility

Yes.


src/send/SendSelectRecipient.tsx
```diff
@@ -135,27 +152,167 @@ const getStartedStyles = StyleSheet.create({
...
+      case ResolutionKind.Address:
+        return getRecipientFromAddress(lowerCaseAddress, recipientInfo)
+      case ResolutionKind.Nom:
+        return {
+          address: lowerCaseAddress,
+          name: t('nomSpaceRecipient', { name: resolution.name ?? searchQuery }),
+          recipientType: RecipientType.Nomspace,
+        }
+      default:
+        return getRecipientFromAddress(lowerCaseAddress, recipientInfo)
```
cajubelt, line 262:
```suggestion
        const otherResolutionKind: never = resolution.kind
        throw new Error(
          `should never happen-- unrecotnized resolution kind ${otherResolutionKind} should be caught at compile time`
        )
```
an idea. this will help us notice we need to add a new case if we ever add a resolution kind that isn't handled here

src/send/SendSelectRecipient.tsx
```diff
@@ -135,27 +152,167 @@ const getStartedStyles = StyleSheet.create({
...
+        recipientType: RecipientType.PhoneNumber,
+        name: t('sendToMobileNumber'),
+      }
+    }
+    if (isValidAddress(searchQuery)) {
+      return getRecipientFromAddress(searchQuery.toLowerCase(), recipientInfo)
+    }
+  }
+
+  const mapResolutionToRecipient = (resolution: NameResolution): Recipient => {
```
cajubelt, line 250: same here re taking functions out to reduce complexity

src/send/SendSelectRecipient.tsx
```diff
@@ -135,27 +152,167 @@ const getStartedStyles = StyleSheet.create({
...
+          recipientType: RecipientType.Nomspace,
+        }
+      default:
+        return getRecipientFromAddress(lowerCaseAddress, recipientInfo)
+    }
+  }
+
+  const mergedFilteredRecipients = useMemo(() => {
+    const allRecipients: Recipient[] = []
+    if (resolveAddressResult && resolveAddressResult.resolutions.length > 0) {
```
cajubelt, line 268:
```suggestion
    if (resolveAddressResult?.resolutions) {
```
code golf

src/send/SendSelectRecipient.tsx
```diff
@@ -135,27 +152,167 @@ const getStartedStyles = StyleSheet.create({
...
+      }
+    }
+
+    const uniqueRecipient = getUniqueSearchRecipient()
+    if (!mergedRecipients.length && uniqueRecipient) {
+      mergedRecipients.push(uniqueRecipient)
+    }
+
+    return mergedRecipients
+  }, [recentFiltered, contactsFiltered, resolveAddressResult])
```
cajubelt, line 294: uniqueRecipient could also change whenever searchQuery changes

src/send/SendSelectRecipient.tsx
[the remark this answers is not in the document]
cajubelt: or maybe just decomposing parts of it into multiple hooks could be helpful.

src/send/SendSelectRecipient.tsx
```diff
@@ -135,27 +152,167 @@ const getStartedStyles = StyleSheet.create({
...
   const contactRecipients = useMemo(
     () => sortRecipients(Object.values(contactsCache)),
     [contactsCache]
   )
+  const [contactsFiltered, setContactsFiltered] = useState(() =>
+    sortRecipients(Object.values(contactRecipients))
+  )
+  const [recentFiltered, setRecentFiltered] = useState(() => recentRecipients)
+
+  const recentRecipientsFilter = useMemo(
```
cajubelt, line 223: these are defined pretty far away from where they are used, which makes it harder to understand what they are supposed to do, and harder to understand the later code that relies on them

cajubelt: by these I mean this and contactRecipientsFilter, though there's probably more examples

src/send/SendSelectRecipient.tsx
```diff
@@ -135,27 +152,167 @@ const getStartedStyles = StyleSheet.create({
...
+
+  const throttledSearch = throttle((searchInput: string) => {
+    setSearchQueryWrapper(searchInput)
+    setRecentFiltered(recentRecipientsFilter(searchInput))
+    setContactsFiltered(contactRecipientsFilter(searchInput))
+  }, SEARCH_THROTTLE_TIME)
+
+  useEffect(() => {
+    // Clear search when recipients change to avoid tricky states
+    throttledSearch('')
```
cajubelt, line 304: if contacts change, do we really want the clearing out to be throttled?

src/send/SendSelectRecipient.tsx
```diff
@@ -135,27 +152,167 @@ const getStartedStyles = StyleSheet.create({
...
+  }, [recentRecipientsFilter, contactRecipientsFilter])
```
cajubelt, line 305: Seems like it should depend on recentRecipients and contactRecipients, e.g. the data itself, rather than these functions that depend on the data

src/send/SendOrInviteButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+import variables from 'src/styles/variables'
+
+interface Props {
+  hidden: boolean
+  disabled: boolean
+  text: string
+  onPress: () => void
+}
+
+export default function SendOrInviteButton({ hidden, disabled, text, onPress }: Props) {
```
cajubelt, line 13: how about a separate button for send and invite? Having functions serve a single purpose will make our code easier to reason about

cajubelt: Also, does this need to be its own component..? Seems like it's just an alias for Button with a certain style

src/send/SendSelectRecipient.tsx
```diff
@@ -135,27 +152,167 @@ const getStartedStyles = StyleSheet.create({
...
-  }, SEARCH_THROTTLE_TIME)
+  const [showSearchResults, setShowSearchResults] = useState(false)
+
+  const defaultCountryCode = useSelector(defaultCountryCodeSelector)
+  const [sendOrInviteButtonHidden, setSendOrInviteButtonHidden] = useState(true)
+
+  // We debounce the search query in order to perform network calls to check
+  // if the query resolves to some special sort of identifier.
+  const [debouncedSearchQuery, setDebouncedSearchQuery] = useState(searchQuery)
+  const debounceSearchQuery = useCallback(
```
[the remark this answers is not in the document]
jophish: This is behavior that was ported over from the old `RecipientPicker`; I can't think of a use for it, so I'll remove it.

src/send/SendOrInviteButton.tsx
jophish: Yeah, I originally thought this was going to be "smart" and derive props internally from data (and also that we were going to use this in multiple places) but it ended up being so simple that it doesn't justify its own component.

src/send/SendSelectRecipient.tsx
jophish: Good catch -- more ported behavior here..

src/send/SendSelectRecipient.tsx
jophish: Ah yeah, that's a good point -- I think without this, we could run into a situation where this doesn't update when it actually should.

src/send/SendOrInviteButton.tsx
```diff
@@ -0,0 +1,33 @@
...
+  if (hidden) {
```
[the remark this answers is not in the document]
jophish: Yep -- ditching this component anyways, but that's a good point to keep in mind.

src/send/SendSelectRecipient.tsx
```diff
@@ -135,27 +152,167 @@ const getStartedStyles = StyleSheet.create({
...
   const recentRecipients = useSelector((state) => state.send.recentRecipients)
   const contactsCache = useSelector(phoneRecipientCacheSelector)
+  const recipientInfo = useSelector(recipientInfoSelector)
+
   const contactRecipients = useMemo(
     () => sortRecipients(Object.values(contactsCache)),
     [contactsCache]
   )
+  const [contactsFiltered, setContactsFiltered] = useState(() =>
+    sortRecipients(Object.values(contactRecipients))
```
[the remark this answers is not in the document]
jophish: Yeah, I agree, this is a weird pattern, inherited from the previous implementation. It's actually even more misleading, since the `Object.values(contactsCache)` above _is_ operating on a proper object, but the rest are operating on arrays. I'll update this to do a regular shallow copy.

src/send/SendSelectRecipient.tsx
```diff
@@ -130,34 +126,123 @@ const getStartedStyles = StyleSheet.create({
...
+        origin: SendOrigin.AppSendFlow,
+      })
+    }
+  }
+  const renderSendOrInviteButton = () => {
+    if (sendOrInviteButtonHidden) {
+      return <></>
+    }
+    return (
+      <Button
```
satish-ravi, line 199: Padding seems off here, might need adjusting SafeAreaView edges

src/icons/Clipboard.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react'
+import Svg, { Path } from 'react-native-svg'
+
+const Clipboard = () => (
+  <Svg width={24} height={24} fill="none">
```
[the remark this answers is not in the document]
jophish: Sure, added a viewBox in.

src/send/SendSelectRecipient.tsx
```diff
@@ -130,34 +126,123 @@ const getStartedStyles = StyleSheet.create({
...
-  const throttledSearch = throttle((searchInput: string) => {
-    setSearchQuery(searchInput)
-  }, SEARCH_THROTTLE_TIME)
+  const forceTokenId = route.params?.forceTokenId
+  const defaultTokenIdOverride = route.params?.defaultTokenIdOverride
 
+  const [sendOrInviteButtonHidden, setSendOrInviteButtonHidden] = useState(true)
+  const [showInviteModal, setShowInviteModal] = useState(false)
   const [showContacts, setShowContacts] = useState(false)
+  const [showSearchResults, setShowSearchResults] = useState(false)
```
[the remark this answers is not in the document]
jophish: Invite doesn't easily slot in here since it's an "overlay" over either the search or contacts screen. Get Started/Recents can be merged into a single state, since "get started" turns into "recents" when `recentRecipients` isn't empty, but I don't think we want to do something like call `setActiveView('recent')` inside of a useMemo when `recentRecipients` changes. I've boiled it down to `get started`, `contacts`, and `search` however and it does clean things up a little bit.

src/send/SendSelectRecipient.tsx
```diff
@@ -130,34 +126,123 @@ const getStartedStyles = StyleSheet.create({
...
+  const dispatch = useDispatch()
 
-  const [searchQuery, setSearchQuery] = useState('')
-  const throttledSearch = throttle((searchInput: string) => {
-    setSearchQuery(searchInput)
-  }, SEARCH_THROTTLE_TIME)
+  const forceTokenId = route.params?.forceTokenId
+  const defaultTokenIdOverride = route.params?.defaultTokenIdOverride
 
+  const [sendOrInviteButtonHidden, setSendOrInviteButtonHidden] = useState(true)
```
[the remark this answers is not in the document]
jophish: No reason -- updated for consistency.

src/send/SendSelectRecipient.tsx
```diff
@@ -130,34 +126,123 @@ const getStartedStyles = StyleSheet.create({
...
+      navigate(Screens.SendAmount, {
+        isFromScan: false,
+        defaultTokenIdOverride,
+        forceTokenId,
+        recipient,
+        origin: SendOrigin.AppSendFlow,
+      })
+    }
+  }
+  const renderSendOrInviteButton = () => {
```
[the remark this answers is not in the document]
jophish: Sure; I think it needs to a bit more complicated than that (onPress handler), but this should pull a bit of code out of the main screen.

src/send/SendSelectRecipient.tsx
```diff
@@ -194,7 +287,7 @@ function SendSelectRecipient() {
                 testID={'SelectRecipient/RecentRecipientPicker'}
                 recipients={recentRecipients}
                 title={t('sendSelectRecipient.recents')}
-                onSelectRecipient={setSelectedRecipient}
+                onSelectRecipient={setSelectedRecipientWrapper}
```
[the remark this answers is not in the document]
jophish: Yeah that's correct, thanks for the catch there.

cajubelt (commented):
partial review, need to go to meetings

src/send/hooks.test.tsx
```diff
@@ -0,0 +1,261 @@
...
+      resolutions: [
+        {
+          kind: ResolutionKind.Address,
+          address: mockAccount,
+        },
+      ],
+    })
+  })
+
+  async function renderHook(searchQuery: string) {
```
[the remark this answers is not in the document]
jophish: I actually didn't have any luck getting this to work after tinkering with it for a bit -- had trouble getting the built-in `renderHook` to wait for nested hooks/state changes in the hooks under test.

src/send/hooks.ts
```diff
@@ -0,0 +1,219 @@
...
+ *
+ * This hook is tested via the SendSelectRecipient.test.tsx file.
+ */
+export function useMergedSearchRecipients(onSearch: (searchQuery: string) => void) {
+  const [searchQuery, setSearchQuery] = useState('')
+
+  const { contactRecipients, recentRecipients } = useSendRecipients()
+
+  const [contactsFiltered, setContactsFiltered] = useState(() =>
+    sortRecipients([...contactRecipients])
```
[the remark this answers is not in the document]
jophish: Ah yeah, I think this is prolly redundant now.

src/send/hooks.ts
```diff
@@ -0,0 +1,219 @@
...
+  )
+  const [recentFiltered, setRecentFiltered] = useState(() => recentRecipients)
+
+  const recentRecipientsFilter = useMemo(
+    () => filterRecipientFactory(recentRecipients, false),
+    [recentRecipients]
+  )
+  const contactRecipientsFilter = useMemo(
+    () => filterRecipientFactory([...contactRecipients], true),
```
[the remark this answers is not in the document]
jophish: Looking at `filterRecipientFactory`, it doesn't appear that it ever actually mutates the recipients list, so I don't actually think we need to create a new list here in either case?

src/send/PasteAddressButton.tsx
```diff
@@ -0,0 +1,76 @@
...
+import CircledIcon from 'src/icons/CircledIcon'
+import colors from 'src/styles/colors'
+import { typeScale } from 'src/styles/fonts'
+
+type Props = PasteAwareWrappedElementProps
+
+function PasteAddressButton(props: Props) {
+  const { isPasteIconVisible, onPressPaste, clipboardContent } = props
+  const { t } = useTranslation()
+  if (!isPasteIconVisible) {
```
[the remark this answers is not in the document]
jophish: @cajubelt This is actually passed from within the `WithPasteAware` component to its child (which is this `PasteAddressButton` component) so it actually isn't available one level up in `SelectRecipientScreen`. I do agree though that in general we should deal with stuff like this higher up if possible.

src/send/SendSelectRecipient.tsx
```diff
@@ -130,34 +126,123 @@ const getStartedStyles = StyleSheet.create({
...
+  }
 
   const onContactsPermissionGranted = () => {
     dispatch(importContacts())
     setShowContacts(true)
   }
 
+  const shouldShowClipboard = (content: string) => {
+    return content !== searchQuery && isAddressFormat(content)
+  }
```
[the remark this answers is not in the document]
jophish: This was a quirk of how our WithPasteAware wrapper was handling iOS 14+. I've made some adjustments and this shouldn't happen anymore (on iOS, the intended behavior for this screen from Nitya is that if we detect that the clipboard is non-empty, we should always show the native modal, and then show the button if there's an address).

src/send/SendSelectRecipient.tsx
jophish: Ah, does this just need to have `bottom` in the edges list?

src/send/SendSelectRecipient.tsx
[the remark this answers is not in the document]
jophish: yeah, pushed!

test/values.ts
```diff
@@ -293,15 +295,15 @@ export const mockNavigation: NativeStackNavigationProp<StackParamList, any> = {
...
 }
 
 export const mockE164NumberToAddress: E164NumberToAddressType = {
-  [mockE164Number]: [mockAccount],
-  [mockE164NumberInvite]: [mockAccountInvite],
-  [mockE164Number2Invite]: [mockAccount2Invite],
+  [mockE164Number]: [mockAccount.toLowerCase()],
+  [mockE164NumberInvite]: [mockAccountInvite.toLowerCase()],
+  [mockE164Number2Invite]: [mockAccount2Invite.toLowerCase()],
 }
```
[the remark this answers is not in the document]
jophish: I'm _pretty sure_ we do -- I think there's the potential for stuff to not work correctly if we didn't.

src/send/SendSelectRecipient.tsx
jophish: Ah, added some padding to the bottom.

src/send/SendSelectRecipient.tsx
```diff
@@ -130,32 +127,164 @@ const getStartedStyles = StyleSheet.create({
...
+
+  const onSelectRecentRecipient = (recentRecipient: Recipient) => {
+    setSelectedRecipient(recentRecipient)
+    navigateToSendAmount(recentRecipient)
+  }
+
+  const navigateToSendAmount = (selectedRecipient: Recipient) => {
+    // TODO (ACT-973): Ensure we're able to to navigate to the next screen
```
jh2oman, line 212: 👍

src/send/SendEnterAmount.tsx
```diff
@@ -214,6 +214,10 @@ function SendEnterAmount({ route }: Props) {
     if (parsedAmount.isLessThanOrEqualTo(0) || parsedAmount.isGreaterThan(token.balance)) {
       return
     }
+    // TODO (ACT-973?): Address is not always set on the recipient, even for PN recipients.
```
jh2oman, line 217: 👍

jh2oman (approved):
Wow that was some dense code. You did a great job organizing though!

src/send/SendSelectRecipient.tsx
```diff
@@ -130,32 +127,164 @@ const getStartedStyles = StyleSheet.create({
...
+            isSelectedRecipientLoading={
+              !!recipient && recipientVerificationStatus === RecipientVerificationStatus.UNKNOWN
+            }
+          />
+          {renderSendOrInviteButton()}
+        </>
+      )
+    } else {
+      return (
+        <>
```
satish-ravi, line 277: nit: is the wrapper `<> </>`  required?

src/send/SendSelectRecipient.tsx
```diff
@@ -130,32 +127,164 @@ const getStartedStyles = StyleSheet.create({
...
+  }
+  const { contactRecipients, recentRecipients } = useSendRecipients()
+  const { mergedRecipients, searchQuery, setSearchQuery } = useMergedSearchRecipients(onSearch)
+
+  const { recipientVerificationStatus, recipient, setSelectedRecipient, unsetSelectedRecipient } =
     useFetchRecipientVerificationStatus()
 
-  const dispatch = useDispatch()
+  const setSelectedRecipientWrapper = (selectedRecipient: Recipient) => {
+    setSelectedRecipient(selectedRecipient)
```
satish-ravi, line 193: should we include an analytics event here? We don't show the send button for recent recipients right?

src/send/hooks.ts
```diff
@@ -0,0 +1,219 @@
...
+    [contactRecipients]
+  )
+
+  const throttledSearch = throttle((searchInput: string) => {
+    onSearch(searchInput)
+    setSearchQuery(searchInput)
+    setRecentFiltered(recentRecipientsFilter(searchInput))
+    setContactsFiltered(contactRecipientsFilter(searchInput))
```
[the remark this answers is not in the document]
satish-ravi: we could move this initial contacts permission check to the screen and potentially use that. https://github.com/valora-inc/wallet/blob/d1b2e3e0cbd6168d558b55c02e003596b8305883/src/send/SelectRecipientButtons.tsx#L42-L46
Also, another potential edge case is that if there's a new contact, it won't be visible unless the user clicks "send crypto to your friends" button.
Maybe these edge cases aren't too important, but in any case these can go in a separate PR, this one is already big 😄

src/send/SendSelectRecipient.tsx
jophish: Ah yeah, good catch

src/analytics/Properties.tsx
```diff
@@ -640,6 +640,8 @@ interface SendEventsProperties {
   [SendEvents.request_contacts_permission_completed]: {
     permissionStatus: PermissionStatus
   }
+  [SendEvents.send_select_recipient_send_press]: undefined
+  [SendEvents.send_select_recipient_invite_press]: undefined
```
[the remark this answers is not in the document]
jophish: Hmm yeah, recipient type is a good idea -- we currently send that on the existing flow.

src/send/saga.test.ts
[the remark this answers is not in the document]
jophish: Yeah, these are intended from the changes to the mock data.

src/send/hooks.ts
```diff
@@ -0,0 +1,226 @@
...
+/**
+ * Returns a single ordered list of all recipients to show in search results,
+ * as well as the search query state variable itself and its setter.
+ *
+ * onSearch is a callback function which will be called with the search query
+ * just before it's updated.
+ *
+ * This hook is tested via the SendSelectRecipient.test.tsx file.
+ */
+export function useMergedSearchRecipients(onSearch: (searchQuery: string) => void) {
```
MuckT, line 33: I think something strange is happing in this function. Currently I only get one result regardless of my query. For example if I search an area code I only see one result for that area code despite having several contacts with matching area codes.

src/send/hooks.ts
```diff
@@ -0,0 +1,226 @@
...
+  allRecipients.push(...recentRecipients)
+  allRecipients.push(...contactRecipients)
+
+  const mergedRecipients: Recipient[] = []
+  for (const potentialRecipient of allRecipients) {
+    if (
+      !mergedRecipients.find(
+        (mergedRecipient) =>
+          mergedRecipient.e164PhoneNumber === potentialRecipient.e164PhoneNumber ||
+          mergedRecipient.address === potentialRecipient.address
```
MuckT, line 162: Multiple recipient's can have a undefined address, not on Valora, so this causes only one match to be returned in cases where there is actually more than one match.

MuckT (changes requested):
Bug fix for multiple contacts matching.

MuckT (approved):

src/send/hooks.ts
jophish: fixed! thanks a ton for catching this

MuckT:
This wasn't touched in this PR but I noticed that `executeFuzzySearch` in `src/recipients/recipient.ts` has `parsedQuery.length < 2`  so that sorts only happen queries of 2 or more characters. Should this be removed?

MuckT:
@jophish found a small bug with this screen when working on #4490 - [Slack Thread](https://valora-app.slack.com/archives/C02KBT0DAHJ/p1701381547390999).

### Repo Steps
1. Tap 'Send crypto to your friends'
2. Quickly tap a contact right after the screen loads.
3. Observe bottom send / invite button displays and is then removed.

### Notes
- This appears to happen if import contacts has not completed prior to selecting. 
- If the users waits ~3 seconds then taps a contact they proceed to send / invite correctly.

### Video
https://github.com/valora-inc/wallet/assets/26950305/e251f3ad-1bd0-4188-9f55-53fd6325ecb8

src/send/SendSelectRecipientSearchInput.tsx
MuckT, line 1: There are some changes to this file in #4490. Not sure if you want to pull those into this PR or have that PR be a fast follow. Since this is all behind a feature flag either seems fine to me.

MuckT (approved):

PR #4469, opened 2023-11-14
feat(swap): improve loading UI

kathaypacific:
### Description

Summary of changes (designs [here](https://www.figma.com/file/YxVf7jqvALnHfzr1Vgz5uV/Swap-2.0?type=design&node-id=1129-24262&mode=design&t=ZrMfCZuvDSXvpHLA-0)):
1. since we've added swap transaction details to the screen and this could push any warnings behind the keyboard, require the user to scroll to the end of the screen to reach the continue button (the "review" button is no longer pushed up by the keyboard)
2. add token value for the network fee
3. add estimated value of swap (this is currently the fiat value of the sell amount)
4. move the exchange rate out of the "to" input box and into the transaction details section
5. add skeleton loader for the "to" value
6. disable proceed button if there is a quote request in flight
7. update disclaimer copy so that it is not specific to Celo

### Test plan

https://github.com/valora-inc/wallet/assets/20150449/51bfc45d-934c-4af0-99bc-54714af5d843

### Related issues

- Fixes RET-865

### Backwards compatibility

Y


src/swap/SwapScreen.tsx
```diff
@@ -54,7 +53,7 @@ import { divideByWei } from 'src/utils/formatting'
 import { getFeeCurrencyAndAmount } from 'src/viem/prepareTransactions'
 import networkConfig from 'src/web3/networkConfig'
 
-const FETCH_UPDATED_QUOTE_DEBOUNCE_TIME = 500
```
kathaypacific, line 57: it felt laggy to have half a second worth of wait time between the user stops typing + kicking off the quote fetch

jeanregisser (approved):
This is great! 🚀 

One small question, the designs had the rate + fee also use the skeleton placeholder, are we going to do it separately?

kathaypacific:
> This is great! 🚀
> 
> One small question, the designs had the rate + fee also use the skeleton placeholder, are we going to do it separately?

yes i didn't put them in here because i don't think they look good 🫠 the designs were updated to include more items in the transaction details (swap price + estimated fiat value) - previously the skeleton loader would have been only for the fees, but now we'd have 3 of them and they're not even next to each other. i'll double check with Nitya if we still want it, and if so then i'll follow up with another PR

kathaypacific:
> > This is great! 🚀
> > One small question, the designs had the rate + fee also use the skeleton placeholder, are we going to do it separately?
> 
> yes i didn't put them in here because i don't think they look good 🫠 the designs were updated to include more items in the transaction details (swap price + estimated fiat value) - previously the skeleton loader would have been only for the fees, but now we'd have 2 of them. i'll double check with Nitya if we still want it, and if so then i'll follow up with another PR

update on this - Nitya has confirmed that we'll just have the skeleton loader on the "to" field so no follow up changes needed for now.

PR #4474, opened 2023-11-14
feat(token-import): Create Import Token screen

denvalora:
### Description

Adds `Import Token` screen, which uses the recent `InLineNotification` https://github.com/valora-inc/wallet/pull/4436 component. Also uses `PasteButton` and updated `TextInput` from https://github.com/valora-inc/wallet/pull/4475

### Test plan

Visual changes. Analytics tested locally when clicking the top bar button `Import`, pasting using `Paste` button, and then once token symbol is typed in, tapping on `Import` button in the Import Token screen. This also tested the paste button works. For now the address is assumed is valid if the length is 42 characters, then the import button works.

#### Visuals
| Blank screen | +Pasted address | +Entered symbol | Success banner |
| -- | -- | -- | -- |

### Related issues

- Rebased on https://github.com/valora-inc/wallet/pull/4475 
- Completes [RET-893/import-tokens-screen](https://linear.app/valora/issue/RET-893/import-tokens-screen)

Future tasks:
- [RET-892/lookup-contract-info](https://linear.app/valora/issue/RET-892/lookup-contract-info) validate entered values (whether valid address, already supported or imported address), check contract
- [RET-891/save-and-display-custom-tokens](https://linear.app/valora/issue/RET-891/save-and-display-custom-tokens) save custom token

### Backwards compatibility

No breaking changes.


src/tokens/TokenImport.tsx
```diff
@@ -0,0 +1,193 @@
...
+                right={isValidAddress && <GreenLoadingSpinner height={32} />} // TODO RET-892: once loaded, hide the spinner
+              />
+              {errorMessage()}
+            </View>
+
+            {/* Network */}
+            <View style={styles.textInputGroup}>
+              <Text style={styles.label}>{t('tokenImport.network')}</Text>
+              <TextInput
+                onChangeText={setTokenSymbol}
```
kathaypacific, line 118: is this right?

src/tokens/TokenImport.tsx
```diff
@@ -0,0 +1,193 @@
...
+                value={t('celoGold') ?? undefined}
```
kathaypacific, line 119: you can use `NETWORK_NAMES` instead...doesn't seem right to use the name of the celo asset here

src/tokens/TokenImport.tsx
```diff
@@ -0,0 +1,193 @@
...
+  },
+  buttonContainer: {
+    paddingVertical: Spacing.Regular16,
+    paddingHorizontal: Spacing.Thick24,
+    borderTopColor: Colors.gray2,
+  },
+  textInputGroup: {
+    flexDirection: 'column',
+    justifyContent: 'flex-start',
+    alignItems: 'flex-start',
```
kathaypacific, line 182: flex-start is the default value, do you need these?

kathaypacific (commented):
looking good, although this PR also contains changes from https://github.com/valora-inc/wallet/pull/4475 - it'd easier to see the isolated changes if the target branch for this PR was denvalora/textinput-right-element

src/analytics/Properties.tsx
```diff
@@ -1394,6 +1394,9 @@ interface AssetsEventsProperties {
   [AssetsEvents.tap_token_details_bottom_sheet_action]: {
     action: TokenDetailsActionName
   } & TokenProperties
+  [AssetsEvents.tap_import_token_screen_open]: undefined
+  [AssetsEvents.tap_import_token_action]: undefined
```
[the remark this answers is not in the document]
denvalora: I'll also log the symbol and networkId.

src/tokens/TokenImport.tsx
```diff
@@ -0,0 +1,193 @@
...
+  const navigateBackAndToast = () => {
+    ValoraAnalytics.track(AssetsEvents.tap_import_token_action)
+    navigateBack()
+    // TODO RET-891: do this only when actually imported
+    dispatch(showMessage(t('tokenImport.importSuccess', { tokenSymbol })))
+  }
+
+  const isValidAddress = useMemo(() => {
+    // TODO RET-892: later will be changed to validate via viem
+    return tokenAddress.length == 42 // 0x prefix and 20 bytes address
```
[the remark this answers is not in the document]
denvalora: It's just a placeholder. I'll update this and come back later when we will use viem.

We'll want to validate the input as the user types it in and thus disallow non-hex. `@celo/utils/lib/address` returns `false` for the above case. When I was working with the error states it kept showing the error message until the address was fully typed in.

src/tokens/TokenImport.tsx
denvalora: I believe you meant for `justifyContent`. Correct, I'll remove.

`alignItems`'s default is `stretch` ([docs](https://reactnative.dev/docs/flexbox#align-items))

src/tokens/TokenImport.tsx
denvalora: No. I've added `setNetworkName` and changed to that.

Although as it's not editable and likely won't be, I could probably change to `Text` component instead. I would have to copy a lot of the styling from the actual `TextInput` component into this one, so to keep consistent I'll leave it as TextInput.

src/tokens/TokenImport.tsx
```diff
@@ -0,0 +1,195 @@
...
+        title={<Text style={styles.headerTitle}>{t('tokenImport.title')}</Text>}
+      />
+      <ScrollView>
+        <View style={styles.titleContainer}>
+          <InLineNotification
+            severity={Severity.Informational}
+            description={t('tokenImport.notification')}
+          />
+
+          <View style={styles.inputContainer}>
```
kathaypacific, line 74: if you apply the gap style to the ScrollView instead, do you still need this View?

src/tokens/TokenImport.tsx
```diff
@@ -0,0 +1,195 @@
...
+            {/* Token Address */}
+            <View style={styles.textInputGroup}>
+              <Text style={styles.label}>{t('tokenImport.input.tokenAddress')}</Text>
+              <TextInput
```
kathaypacific, line 78: one thing i thought about while reading this, is the number of props that are repeated with each TextInput. it seems like we'd want to apply the same props consistently throughout this screen, and if we want to change / add / remove props, we'd need to do it in all 3 places. classic signals that we might want to componentise this - what do you think about an input component that includes the View, label, and text input? the props could be something like label, children, ...textInputProps.

src/tokens/TokenImport.tsx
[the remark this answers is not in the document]
kathaypacific: i think this is a problem that we can avoid altogether. it looks like you have `editable={isValidAddress(tokenAddress)}` on the TextInput for token address, but is this necessary? what about if it is editable always except for when the token symbol is being fetched? (that part isn't implemented yet, so editable can just be true for now). if the form validation is only done on blur, then we don't need to worry about checking for the length of the input (we don't really need to do validation as the user types anyway)

src/tokens/TokenImport.tsx
denvalora: The gaps differ between the notification and the container of the input groups. Otherwise we could remove this `<View>`. I'll check with Nitya if the gaps should be equal instead. For now I'll keep this `<View>`.

image

src/tokens/TokenImport.tsx
denvalora: Agree, I think the code refactoring rules of thumb are good to follow:
- WET: Write Everything Twice
- Rule of three: once duplication repeats three times, do abstraction

Slight concerns that the network component is not editable and that it will likely be a dropdown list in the future.

I'll add a render function to abstract this.

kathaypacific (approved):
nice work, excited to see this in the app! ✨

PR #4475, opened 2023-11-15
feat(token-import): Extend TextInput, create PasteButton

denvalora:
### Description

- Extended TextInput with additional property `rightElement` passed by the consumer. This component is shown on the right-most side of the text input. Clear button (when visible) is on the left side of this `rightElement`.
- Created `PasteButton` component for tokens, which is able to capture clipboard content. Example use:
   ```jsx
   <PasteButton onPress={(content) => {
    setName(content)
    ValoraAnalytics.track(DappExplorerEvents.dapp_rankings_open)
  }} />
   ```

Currently TextInput's clear icon has a left margin of 8. This is changed to have a gap of 8 instead. End result looks the same, but now can put a component on the left side of the clear icon neatly.

### Test plan

Used the following input:
>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi at.

#### Before and After
Only the `Name` text input was changed. Ignore the cursor in the `Message`.

| Before (focused) | After (focused) | Before | After |
| -- | -- | -- | -- |

#### Example
| Blank | Tapped on "Paste" | Example (focused) |
| -- | -- | -- |
| image | image | image |
| image | image | image |

When the user has not tapped into the text input, the cursor nor the clear icon appear. If the user needs to make changes, they need to focus on the text input by tapping on it.

### Related issues

- To be used in [RET-893/import-tokens-screen](https://linear.app/valora/issue/RET-893/import-tokens-screen)

### Backwards compatibility

Backwards compatible - `rightElement` in TextInput is optional, and `PasteButton` will be used later.


src/components/PasteButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+    const content = await getFreshClipboardContent()
+    setValue(content)
+    onPress?.(event)
+  }
+
+  // when a value is already set, remove the button
+  if (value) return null
+
+  return (
+    <Touchable onPress={pasteClipboard}>
```
kathaypacific, line 29:
```suggestion
    <Touchable onPress={handlePasteClipboard}>
```
or `onPasteClipboard`. we usually prefix any handlers with `on` or `handle`. `on` is usually used for passed through props, and `handle` for any handlers that the component declares by itself (although we've not been so consistent about the latter, sometimes the `on` prefix is used)

src/components/PasteButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { typeScale } from 'src/styles/fonts'
+import { useClipboard } from 'src/utils/useClipboard'
+
+interface Props {
+  value: string
+  setValue: React.Dispatch<React.SetStateAction<string>>
+  onPress?: (event: GestureResponderEvent) => void
+}
+
+export const PasteButton = ({ value, setValue, onPress }: Props) => {
```
kathaypacific, line 15: i'm consumed about these props, it sounds like a consumer would do something like `const [value, setValue] = useState(something)` and pass the `value` and `setValue` in to here? why? passing state setters directly into a component is usually a sign that the abstraction doesn't make sense. i'd expect a paste button to simply paste what's in the clipboard and it should not worry about any values.

i'd recommend that this takes 1 prop (onPress) and the consumer should do the `setValue` inside the onPress.

src/components/PasteButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const [, , getFreshClipboardContent] = useClipboard()
+
+  const pasteClipboard = async (event: GestureResponderEvent) => {
+    const content = await getFreshClipboardContent()
+    setValue(content)
+    onPress?.(event)
+  }
+
+  // when a value is already set, remove the button
+  if (value) return null
```
kathaypacific, line 26: this also feels like this component needs to know too much about the consumer. i think it makes more sense if 
1. the component takes a prop for `isVisible`
2. the consumer should show or hide the paste button

as a general rule of thumb, generic components are more easily reused if the consumer has control of them. avoiding as many conditions as possible, and making props as instructive as possible usually helps with this. (e.g. having a `value` prop that actually sometimes hides the component is not obvious, and may not work for other scenarios. but having something like `isVisible` allows the consumer to decide the logic.)

having said this, i actually prefer option 2. because there's less props to manage.

src/components/PasteButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+  return (
+    <Touchable onPress={pasteClipboard}>
+      <Text style={styles.text}>{t('paste')}</Text>
+    </Touchable>
+  )
+}
+
+const styles = StyleSheet.create({
+  text: {
+    textAlign: 'center',
```
kathaypacific, line 37: do we need this?

src/components/TextInput.tsx
```diff
@@ -26,6 +27,7 @@ type Props = Omit<RNTextInputProps, 'style'> & {
   onChangeText: (value: string) => void
   testID?: string
   showClearButton?: boolean
+  right?: React.ReactNode
```
kathaypacific, line 30: this variable name is hard to understand. 
```suggestion
  rightElement?: React.ReactNode
```

src/components/TextInput.tsx
```diff
@@ -84,14 +87,15 @@ export class CTextInput extends React.Component<Props, State> {
           onFocus={this.handleInputFocus}
           onBlur={this.handleInputBlur}
         />
+        {right}
```
kathaypacific, line 90: i expected the right element to be on the right? shouldn't this come after the circle button then?

kathaypacific (commented):
i'm kind of wondering about the value of the Paste button. we have a few places that do pasting (e.g. the import wallet screen) and the button is not standardised. i guess i don't feel strongly that we need to have a standardised Paste button - i can see when an icon or bigger button would make sense, and the paste functionality is already abstracted into a hook. i actually think it'd be more straightforward if this component lives with the import tokens stuff, either directly inside the import tokens screen or in the same folder. wdyt?

src/components/TextInput.tsx
denvalora: I was not sure about whether to prioritise the clear button or this custom button. I'll check with Nitya. I'll make the change for now.

In our examples we'll have only one showing at the same time.

| {right} in the middle | {right} on the right-most side |
| -- | -- |
| image | image |

src/components/TextInput.tsx
denvalora: I named it initially this way to follow the same naming as in CustomHeader:
https://github.com/valora-inc/wallet/blob/ac248834c6b5dfed860330c51bd9a8c3ac6dbaab/src/components/header/CustomHeader.tsx#L7

Renamed for clarity

src/components/PasteButton.tsx
denvalora: No, I forgot to remove that.

src/components/PasteButton.tsx
denvalora: That makes sense. I was considering removing those properties as well. I've made the change.

I've manually tested in SupportContact with the below code. It works.
```ts
            <TextInput
              onChangeText={setName}
              multiline={false}
              value={name}
              style={styles.singleLineTextInput}
              showClearButton={true}
              testID={'NameEntry'}
              rightElement={ !name &&
                <PasteButton onPress={(content, event) => {
                  setName(content)
                  ValoraAnalytics.track(SettingsEvents.settings_revoke_phone_number_confirm)
                }} />
              }
            />
```

src/components/PasteButton.tsx
denvalora: I chose Option 2.

src/tokens/PasteButton.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+import { useTranslation } from 'react-i18next'
+import { StyleSheet, Text } from 'react-native'
+import Touchable from 'src/components/Touchable'
+import { Colors } from 'src/styles/colors'
+import { typeScale } from 'src/styles/fonts'
+import { useClipboard } from 'src/utils/useClipboard'
+
+interface Props {
+  onPress?: (clipboardContent: string) => void
```
kathaypacific, line 10: why is this optional? (what is the purpose of a paste button that takes no callback?)

kathaypacific (approved):
looks good ✨ would be nice to make the paste button callback required though

src/tokens/PasteButton.tsx
denvalora: It was from before when I had setValue (which was later removed) and onPress.

I've changed it to a required property

PR #4477, opened 2023-11-15
feat: enable multichain WalletConnect transactions

kathaypacific:
### Description

This PR enables WalletConnect transactions on Ethereum.
- add dynamic config for allowed chains
- accept sessions for allowed chains
- use the chainId from the session request to choose the viem wallet to use

### Test plan

I only did one test transaction on Ethereum:

https://github.com/valora-inc/wallet/assets/20150449/5f25a84c-9fac-4e1d-9abf-3770ea9e08a5

Unfortunately the wallet connect example app does not support Sepolia :( 

### Related issues

- Fixes RET-904


src/walletConnect/request.ts
```diff
@@ -22,13 +23,34 @@ import Web3 from 'web3'
 
 const TAG = 'WalletConnect/handle-request'
 
-export function* handleRequest({ method, params }: { method: string; params: any[] }) {
-  const wallet: ViemWallet = yield* call(getViemWallet, networkConfig.viemChain.celo)
+export const networkIdToWalletConnectChainId: Record<NetworkId, string> = {
```
kathaypacific, line 26: i opted for these mappings because the chainId for walletconnect contains the namespace so we cannot use the chainId directly. we would need to make assumptions about the format of the chainId (i.e. split by `:` and take the second element) which seems just as brittle as having a custom mapping...

bakoushin (approved):
LGTM!  ⛓️⛓️⛓️

bakoushin:
What do you think about adding an e2e test for signing and sending an Ethereum transaction?

kathaypacific:
> What do you think about adding an e2e test for signing and sending an Ethereum transaction?

yes i'll do this in a separate PR so that it's easier to review :)

src/walletConnect/request.ts
kathaypacific: yes good idea. i moved it to networkConfig because of the second point...although i was tempted to put it in the wc/types.ts file

bakoushin: my naive guess is if we add in the future another EVM compatible network like Polygon, probably it would be simpler to copy-paste and fill some _template_ rather than sniffing all around the codebase to find all the places to insert it

PR #4479, opened 2023-11-16
feat: enable display of eth on providers cash in

MuckT:
### Description

Enables the display of eth providers by moving from `address` to `tokenId`.

### Test plan

Tested locally by deploying a modified fetch providers cloud function on Alfajores.

| On this Branch | On #4450 |
| ----- | ----- |
|  |  |

### Related issues

- ACT-965

### Backwards compatibility

Yes

### DIFF
--- a/src/fiatExchanges/amount.test.tsx
+++ b/src/fiatExchanges/amount.test.tsx
@@ -2,13 +2,13 @@ import { render } from '@testing-library/react-native'
 import React from 'react'
 import { Provider } from 'react-redux'
 import CurrencyDisplay from 'src/components/CurrencyDisplay'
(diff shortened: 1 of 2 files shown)


cajubelt (commented):
this is already being handled in Finn's PR https://github.com/valora-inc/wallet/pull/4450 . Also, I don't think this is strictly necessary to start displaying eth providers, since `useTokenInfoWithAddressBySymbol` works as a stopgap. I think the only situation we'd start to get problems is if there was a collision on token symbols

MuckT:
Closing in favor of #4450 as that will handle the wallet display.

PR #4481, opened 2023-11-16
feat(swap): delete SwapReviewScreen

bakoushin:
### Description

* `SwapReviewScreen` deleted
* pressing 'Confirm Swap' on the `SwapScreen` executes the swap for the user and takes the user to the swap execution screen

### Test plan

* Tested manually on mainnet swapping CELO for cUSD and vice versa
* Updated unit tests

### Related issues

- Fixes RET-909

### Backwards compatibility

Y


src/analytics/Properties.tsx
```diff
@@ -1290,9 +1290,7 @@ interface SwapEventsProperties {
     tokenSymbol?: string
   }
   [SwapEvents.swap_gas_fees_learn_more]: undefined
-  [SwapEvents.swap_screen_review_swap]: undefined
-  [SwapEvents.swap_review_screen_open]: SwapEvent & Web3LibraryProps & Partial<SwapTxsProperties>
-  [SwapEvents.swap_review_submit]: SwapQuoteEvent &
+  [SwapEvents.swap_screen_confirm_swap]: SwapQuoteEvent &
```
kathaypacific, line 1293: maybe we can keep `swap_review_submit`? then we wouldn't need to update any data dashboards (since they do the same thing?)

src/swap/SwapExecuteScreen.tsx
```diff
@@ -46,7 +34,7 @@ export function SwapExecuteScreen() {
           <View style={styles.actionBar}>
             <Button
               text={t('SwapExecuteScreen.swapActionBar.tryAgain')}
-              onPress={navigateToReviewScreen}
+              onPress={navigateBack}
```
kathaypacific, line 37: sorry, this is an existing issue but i didn't notice until now...do you think we should add analytics for these buttons? :/ i'm not sure, just noting it because i noticed it

src/swap/SwapScreen.tsx
```diff
@@ -397,7 +461,7 @@ export function SwapScreen({ route }: Props) {
     }
   }
 
-  const allowReview = useMemo(
+  const allowSwap = useMemo(
     () => Object.values(parsedSwapAmount).every((amount) => amount.gt(0)) && !fetchingSwapQuote,
```
kathaypacific, line 465: should we add a check for if the quote is "possible" if using viem?

src/swap/SwapScreen.tsx
```diff
@@ -300,14 +344,34 @@ export function SwapScreen({ route }: Props) {
...
+        .toNumber(),
+      allowanceTarget,
+      estimatedPriceImpact,
+      price,
+      provider: swapResponse.details.swapProvider,
+      web3Library: 'contract-kit',
+    })
+
+    dispatch(setSwapUserInput(userInput))
+    dispatch(
```
kathaypacific, line 368: ahh, the dreaded double dispatch. i think both `swapStart*` actions already set the swapInfo, so we can just remove this action altogether?

kathaypacific (approved):
YAY excited to see this change! i had a few small suggestions

src/swap/SwapScreen.tsx
bakoushin: great catch, thanks! looks like `userInput` passed here is not used anywhere, and can be removed.

although, it seems logical to me to set `swapState` to `USER_INPUT` just after the swap screen is open, so I renamed and moved this reducer to the `useEffect` fired at the first render

since the code changes are spread all over the codebase, I created a separate PR https://github.com/valora-inc/wallet/pull/4509 to merge into this one

src/swap/SwapScreen.tsx
bakoushin: I think not: because if the quote is not "possible" it should allow to press the button, but instead of starting swap it will display a bottom sheet

src/swap/SwapExecuteScreen.tsx
bakoushin: I agree, let's at least record if any of those buttons are tapped.
Added analytics events for "Swap again" and "Try again" taps.

PR #4482, opened 2023-11-16
fix(token-details): show asset value as title on token details screen

cajubelt:
### Description

fix to match designs on what should be shown on token details screen. and use fallback from designs when price is unavailable

designs [here](https://www.figma.com/file/OE6RFzqXAMqU9ygHtNYyof/Assets-Page-(Multichain)?type=design&node-id=2099-15029&mode=design&t=oidiTlDQ3N5HX2bm-0)

current asset details screen (on main branch)
Screenshot 2023-11-16 at 11 00 16 AM

after updates in this PR:
- with usd price:
simulator_screenshot_D0761EB2-7CD0-4CA6-A890-0369E5D1A203

- without usd price:

simulator_screenshot_F53E7C2D-49B6-405E-BF42-1511822B53D4

### Test plan

- manual test
- jest

### Related issues

na

### Backwards compatibility

na


src/tokens/TokenDetails.tsx
```diff
@@ -76,12 +78,19 @@ export default function TokenDetailsScreen({ route }: Props) {
...
-          tokenId={tokenId}
-          style={styles.balance}
-          testID="TokenDetails/Balance"
-        />
+        {token?.priceUsd !== null ? (
+          <TokenDisplay
+            amount={1}
+            tokenId={tokenId}
+            style={styles.assetValue}
+            testID="TokenDetails/AssetValue"
```
satish-ravi, line 86: Looks like e2e test is failing as it looks for `TokenDetails/Balance`

src/tokens/TokenDetails.tsx
```diff
@@ -76,12 +78,19 @@ export default function TokenDetailsScreen({ route }: Props) {
...
+          />
+        ) : (
+          <Text style={styles.assetValue} testID="TokenDetails/AssetValue">
+            {localCurrencySymbol ?? '$'}
+            {' --'}
+          </Text>
```
satish-ravi, line 92: Looking at token details, looks like we end up showing `-` if either `tokenInfo.priceUsd` is null or if `localCurrencyExchangeRate` is also null. Wonder if we should use the same condition here and for showing "Price unavailable" text.

src/tokens/TokenDetails.tsx
cajubelt: thanks, fixed

src/tokens/TokenDetails.tsx
cajubelt: I went a different direction and added an errorFallback prop to TokenDisplay

cajubelt: to avoid duplicating logic from TokenDisplay and having to somehow keep it in sync

cajubelt: sounds like you might be concerned about a separate corner case, where TokenDisplay is in an error state but PriceInfo is not. I think the main purpose of this PR is to show the correct data, and not to deal with the other bugs in this component related to error states, so I suggest we deal with that separately, or maybe skip it, since a solution is not obvious to me and I'm not sure it will ever come up

cajubelt: I guess we'd need to just ditch TokenDisplay and use something else

satish-ravi (approved):

src/tokens/TokenDetails.tsx
cajubelt: ok now I'm interested in trying that lol

PR #4484, opened 2023-11-17
chore(deps): update fiatconnect sdk, ethers

cajubelt:
### Description

needed a babel plugin to get the build going again

groups fiatconnect sdk together since that depends on ethers v6

surprisingly, we actually did not have a declared dependency on ethers yet, even though we use it directly

### Test plan

ci

### Related issues

na

### Backwards compatibility

na


package.json
satish-ravi, line 1: Should we remove the ethers entry under resolution? It's the one that renovate was trying to upgrade and failing.

cajubelt: thanks, I didn't see that

satish-ravi (approved):

PR #4490, opened 2023-11-17
feat: send select recipient display, search and selection improvements

MuckT:
### Description

Fixes a few things from the new search-ui
- Large font displays 
- Send / Invite button placement
- ~~Passthrough clicks to the recipients behind keyboard~~ Creates a bug documented in ACT-997.

| Large Font Android | Large Font iOS |
| ----- | ----- |
| <video src="https://github.com/valora-inc/wallet/assets/26950305/123c46c7-a767-46ce-987a-3780f3fc5f35" /> | <video src="https://github.com/valora-inc/wallet/assets/26950305/979bb825-623d-47e5-898d-a1c5942cea39" /> | 

#### Notes

Removed the floating keyboard spacer in de193a70b928f203782ee0864feadc8e740bf93b as it wasn't needed.

| With Spacer | Without Spacer |
| ----- | ----- |
| <video src="https://github.com/valora-inc/wallet/assets/26950305/7032515d-38e2-40a8-8df2-1516f9984b26" /> | <video src="https://github.com/valora-inc/wallet/assets/26950305/8efee65e-4095-4690-a9e1-1500cdd1f88b" /> |

### Test plan

- [x] Tested locally on iOS

### Related issues

- ACT-977

### Backwards compatibility

Yes


MuckT:
If the misalignment of the placeholder and paste button is an issue we could add `allowFontScaling={false}` [here](https://github.com/valora-inc/wallet/blob/main/src/send/SendSelectRecipientSearchInput.tsx#L24-L33) which is similar to what we do on the dapps list screen.

src/send/SendSelectRecipient.test.tsx
```diff
@@ -5,6 +5,12 @@ import SendSelectRecipient from 'src/send/SendSelectRecipient'
 import { createMockStore } from 'test/utils'
 import { mockPhoneRecipientCache, mockRecipient, mockRecipient2 } from 'test/values'
 
+jest.mock('src/utils/permissions')
```
satish-ravi, line 8: is this needed? there's no longer a dependency on `src/utils/permissions`

src/send/SendSelectRecipient.tsx
```diff
@@ -40,8 +41,14 @@ function GetStartedSection() {
...
-        <CircledIcon radius={24} style={getStartedStyles.optionNum} backgroundColor={colors.white}>
-          <Text style={getStartedStyles.optionNumText}>{optionNum}</Text>
+        <CircledIcon
+          radius={Math.min(24 * getFontScaleSync(), 50)}
+          style={getStartedStyles.optionNum}
+          backgroundColor={colors.white}
+        >
+          <Text adjustsFontSizeToFit={true} style={getStartedStyles.optionNumText}>
+            {optionNum}
+          </Text>
```
satish-ravi, line 51: I am seeing this number disappear on very large font sizes

src/send/SendSelectRecipientSearchInput.tsx
```diff
@@ -45,13 +45,11 @@ const styles = StyleSheet.create({
     flex: 1,
   },
   search: {
-    height: 44,
-    alignItems: 'flex-start',
+    alignItems: 'center',
```
satish-ravi, line 48: Looking at the screenshots, seems like this is still not center aligned on iOS? Also cc @jophish who I think is fixing the alignment on a different PR

src/send/SendSelectRecipient.test.tsx
```diff
@@ -5,6 +5,12 @@ import SendSelectRecipient from 'src/send/SendSelectRecipient'
...
+jest.mock('react-native-device-info', () => {
+  return {
+    getFontScaleSync: () => 1,
+  }
+})
```
satish-ravi, line 13:
```suggestion
jest.mock('react-native-device-info', () => ({ getFontScaleSync: () => 1 }))
```

satish-ravi:
> Note we do get an error using a FlatList within a ScrollView as seen below, but we also get the same error for the Celo news, currency selection and language selection screens.

I don't think we should add more of this just because there are other screens that also have this warning. Could we just turn off scrolling on the FlatList with the [onScrollEnabled](https://reactnative.dev/docs/scrollview#scrollenabled) option?

src/send/SendSelectRecipient.tsx
MuckT: This is a bit tricky, if you change your font size while on the screen the getFontScaleSync() will not refresh; however, if you go back and navigate to the screen then it should display correctly. This is a bit of an edge case where a user adjusts their font size while this screen displayed.

cajubelt (commented):
I'm confused by the third bullet in the description-- are you saying this PR introduces a bug?

cajubelt:
also the semantic commits standard allows "fix" as a prefix, which might be more appropriate than "feat" here given that no new features appear to be added

MuckT:
The bug isn't present in the PR, but if you pass the clicks through to the child flatlist a bug is surfaced - adding a695ecd063339b42e871c790f10becff19678286 to this PR or main will surface the bug.

cajubelt:
@MuckT can you explain why we would need to pass clicks through?

cajubelt (approved):

src/send/SendSelectRecipient.tsx
```diff
@@ -231,11 +240,14 @@ function SendSelectRecipient({ route }: Props) {
   const onPressSendOrInvite = (shouldInviteRecipient: boolean) => {
     if (!recipient) {
       return
+    } else {
+      Keyboard.dismiss()
```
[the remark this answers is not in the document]
MuckT: Removed in fd7d3d4bf3d31d26410bd9ea0882c04e6a9cb316!

src/send/SendSelectRecipient.tsx
```diff
@@ -190,6 +198,7 @@ function SendSelectRecipient({ route }: Props) {
     useFetchRecipientVerificationStatus()
 
   const setSelectedRecipientWrapper = (selectedRecipient: Recipient) => {
+    Keyboard.dismiss()
```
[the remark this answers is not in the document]
MuckT: Removed in fd7d3d4bf3d31d26410bd9ea0882c04e6a9cb316!

MuckT:
@cajubelt some notes on taps.
- I've removed persisting taps from this PR as passing that tap down to the children introduces more impactful render bug where the invite / send button is cleared after selecting a recipient; That issue is documented in ACT-997.
- On why we'd want pass the tap to the children is for a more intuitive UI - currently after searching a user must tap the screen once to dismiss the keyboard and then again to select the contact. Having `keyboardShouldPersistTaps={'always'}` on the child component makes the selection take one tap instead of two.

PR #4491, opened 2023-11-17
chore(eth-fees): calculate eth send fees on SendEnterAmount

cajubelt:
### Description

uses our prepareTransactions helper to get the cost of sending eth, and displays it on SendEnterAmount screen

### Test plan

- jest
- manual test on dev build

https://github.com/valora-inc/wallet/assets/7979215/cf90ab42-2256-4922-a582-9041d9d44cdd

### Related issues

fixes https://linear.app/valora/issue/ACT-956/ethereum-send-fees

### Backwards compatibility

na


src/send/usePrepareSendTransactions.ts
```diff
@@ -32,22 +33,26 @@ export async function _prepareSendTransactionsCallback({
   if (amount.isLessThanOrEqualTo(0)) {
     return
   }
+  const baseTransactionParams = {
+    // not including sendToken yet because of typing. need to check whether token has address field or not first, required for erc-20 transfers
```
satish-ravi, line 37: this is because erc-20 transfer methods require TokenBalanceWithAddress right? Looks like it's deprecated, should it no longer be deprecated?

src/viem/prepareTransactions.ts
```diff
@@ -372,7 +372,47 @@ export async function prepareTransferWithCommentTransaction(
...
+    toWalletAddress,
+    amount,
+    feeCurrencies,
+    sendToken,
+  }: {
+    fromWalletAddress: string
+    toWalletAddress: string
+    amount: bigint
+    feeCurrencies: TokenBalance[]
+    sendToken: TokenBalance
```
satish-ravi, line 398:
```suggestion
    sendToken: TokenBalance & {isNative: true}
```
should we enforce type here? Although I guess this opens up for handling a case in usePrepareSendTransactions for tokens that don't have an address and is also non native, which I believe should never happen?

test/values.ts
```diff
@@ -509,6 +509,20 @@ export const mockTokenBalances: Record<string, StoredTokenBalance> = {
     priceFetchedAt: Date.now(),
     isCashInEligible: true,
   },
+  [mockEthTokenId]: {
```
satish-ravi, line 512: I've been hesitant to add this in many prior multichain PRs thinking it may cause some unexpected test failures and instead just creating an eth token balance entry in multiple test files, but looks like things are fine 😄

src/viem/prepareTransactions.ts
```diff
@@ -372,7 +372,47 @@ export async function prepareTransferWithCommentTransaction(
...
+  },
+  prepareTxs = prepareTransactions
+): Promise<PreparedTransactionsResult> {
+  const baseSendTx: TransactionRequest = {
+    from: fromWalletAddress as Address,
+    to: toWalletAddress as Address,
+    value: amount,
+  }
+  Logger.info(TAG, 'prepareSendNativeAssetTransaction')
```
satish-ravi, line 407: is this log required?

satish-ravi (approved):

src/send/usePrepareSendTransactions.ts
cajubelt: strangely, the same file exports two interfaces with the same name, only one of which is deprecated. I'm surprised that's possible. Cleaning it up...

src/viem/prepareTransactions.ts
cajubelt: added static typing and an error in the logs if we get a token that isn't native and doesn't have an address

cajubelt: > I guess this opens up for handling a case in usePrepareSendTransactions for tokens that don't have an address and is also non native, which I believe should never happen

we have tests in address-metadata to prevent this, so it should be ok

src/viem/prepareTransactions.ts
cajubelt: removed

test/values.ts
cajubelt: lol yeah I was wondering if it would break stuff. It did at first, but I was able to fix just by making sure the balance was not the highest of the other balances here

jeanregisser (approved):
Great! 🚀

PR #4512, opened 2023-11-21
refactor(max-button): use viem for max button

cajubelt:
### Description

Use viem for new send flow's max button. For now, just uses the balance of that token. If it is a gas-paying token, and the user has no other balance that can be used to pay for gas, this will result in a warning for now which instructs the user to lower their send amount to continue. I confirmed with @lvsuttie that this is ok for now.

I also gave some effort to making a "smart" max button that proactively subtracts the amount needed for gas if the currency being sent is a gas-paying token. Specifically, I tried using our usePreparedTransactions helper and seeing if I could get a `need-decrease-spend-amount-for-gas` result type that would include a decreased amount that would work for the send. However, I found that for sends, that result type doesn't occur in the wild. I think `need-decrease-spend-amount-for-gas` is just happening for swaps, where the second transaction ("swap", which comes after "approve") has a gas estimate given to us by the swap provider, not by simulating the transaction with Viem. (@jeanregisser, please correct me on this if I'm wrong). 

### Test plan

- existing unit test coverage for the max button resulting in max amount being shown, and on the warning message when there is not enough for gas

- manual tests on CELO, cUSD, ETH

https://github.com/valora-inc/wallet/assets/7979215/d3793032-00a2-40c3-966c-f4b85ef7a45d

### Related issues

fixes https://linear.app/valora/issue/ACT-946/update-usemaxamount-to-handle-non-celo-tokens

### Backwards compatibility

na


src/fees/hooks.ts
```diff
@@ -45,6 +51,58 @@ export function usePaidFees(fees: Fee[]) {
...
+      return output
+    } catch (error) {
+      Logger.error(TAG, 'Error calculating max send amount', error)
+      return
+    }
+  }, [token])
+  if (loading) {
+    return { loading: true, maxAmount: null }
+  }
+  if (prepareTransactionsResult?.type === 'need-decrease-spend-amount-for-gas') {
```
cajubelt, line 95: I noticed that this doesn't seem to ever happen for sends-- it might only work right now for transactions where we include a gas estimate rather than querying for it from Viem

jeanregisser:
> I also gave some effort to making a "smart" max button that proactively subtracts the amount needed for gas if the currency being sent is a gas-paying token. Specifically, I tried using our usePreparedTransactions helper and seeing if I could get a `need-decrease-spend-amount-for-gas` result type that would include a decreased amount that would work for the send. However, I found that for sends, that result type doesn't occur in the wild. I think `need-decrease-spend-amount-for-gas` is just happening for swaps, where the second transaction ("swap", which comes after "approve") has a gas estimate given to us by the swap provider, not by simulating the transaction with Viem. (@jeanregisser, please correct me on this if I'm wrong).

Ah yes that's right, `need-decrease-spend-amount-for-gas` can only be returned if the `gas` is already set.

We can probably come back to this later. 

One way to have a "smart" max would be to call `prepareTransactions` twice, first trying to send 1 smallest unit of the token. And then reusing the returned `gas` value.
Then the 2nd call would be able to return `need-decrease-spend-amount-for-gas`.

I'm not entirely sure this is correct in all cases. i.e. it would assume the amount doesn't change the gas estimate. Which is probably right for most ERC token implementations, but I'm sure there could be exceptions 👀

jh2oman (approved):

PR #4516, opened 2023-11-21
chore(swaps): Removing swap native tokens experiment

dievazqu:
### Description

Removing swap native tokens experiment.

### Related issues
- Fixes RET-918


src/swap/SwapScreen.test.tsx
```diff
@@ -278,7 +278,7 @@ describe('SwapScreen', () => {
     })
 
     mockExperimentParams.mockReturnValue({
-      swappingNonNativeTokensEnabled: false,
+      swapBuyAmountEnabled: true,
     })
```
jeanregisser, line 282: nit: could we even remove this?

jeanregisser (approved):
🚀

src/swap/SwapScreen.test.tsx
dievazqu: Yes, there is another [task](https://linear.app/valora/issue/RET-920/remove-swap-buy-amount-experiment) to remove this. So we do it in separate PRs

bakoushin (approved):
Looks great! I was confused by the same thing as @jeanregisser

src/swap/SwapScreen.test.tsx
jeanregisser: Ok I got confused as I didn't see that value in the existing test.

dievazqu: There was an error in a test if I removed the `mockExperimentParams.mockReturnValue` completely.
I'll handle that in the other task.

PR #4518, opened 2023-11-21
refactor(swap): use token id

jeanregisser:
### Description

First step before using the updated swap API.

### Test plan

- Tests pass

### Related issues

- Fixes RET-911

### Backwards compatibility

Yes


src/redux/migrations.ts
```diff
@@ -1428,4 +1428,5 @@ export const migrations = {
     ...state,
     swap: _.omit(state.swap, 'swapUserInput'),
   }),
+  169: (state: any) => state,
```
jeanregisser, line 1431: Nothing to do, the updated state is set to `null` during rehydration:

https://github.com/valora-inc/wallet/blob/a8bc7c1cfbf7619e942396e87490d68255504b88/src/swap/slice.ts#L86

dievazqu (approved):
LGTM!

denvalora (approved):

PR #4522, opened 2023-11-22
chore: show fiat value for both swap tokens and tweak warnings

kathaypacific:
### Description

This PR:
- adds the fiat value of the swap tokens in the token text input fields
- adds a placeholder that we can change through translations (e.g. if we want to have a more descriptive text)
- adds skeleton loader for the fiat value
- change priority of the warnings (warnings should only be displayed after the quote is received, show price impact warning before missing price impact warning)

### Test plan

- swap when a token has no usd price + high price impact -> price impact warning
- swap when both tokens have usd price + low price impact -> no warnings
- swap when a token has no usd price + low price impact -> "We can't estimate the value of this swap" warning

https://github.com/valora-inc/wallet/assets/20150449/7b18c9be-03ba-4cee-aca1-796af69ad7e0

### Related issues

- Fixes RET-928

### Backwards compatibility

Y


locales/base/translation.json
```diff
@@ -1669,6 +1669,7 @@
     "disclaimer": "Best rate determined by network partners. <0>Learn more</0>",
     "swapFromTokenSelection": "SWAP FROM",
     "swapToTokenSelection": "SWAP TO",
+    "tokenUSDValueUnknown": "-",
```
jeanregisser, line 1672: nit:
```suggestion
    "tokenUsdValueUnknown": "-",
```

src/swap/SwapAmountInput.tsx
```diff
@@ -147,6 +153,28 @@ const SwapAmountInput = ({
           )}
         </Touchable>
       </View>
+      <View testID="SwapAmountInput/FiatValue">
+        <Text style={[styles.fiatValue, { opacity: loading || !swapAmount || !token ? 0 : 1 }]}>
+          {token?.priceUsd && swapAmount?.gt(0) && localCurrencyExchangeRate
+            ? `≈${localCurrencySymbol}${swapAmount
+                .multipliedBy(token.priceUsd)
+                .multipliedBy(localCurrencyExchangeRate)
+                .toFormat(2, BigNumber.ROUND_DOWN)}`
```
jeanregisser, line 162: I think this format should be translatable. Also maybe a space after the approximate symbol?
Why don't we use our `TokenAmount` component here too?
Otherwise we risk deviating from our "standard" way of showing amounts.

src/swap/SwapScreen.tsx
```diff
@@ -536,21 +536,27 @@ export function SwapScreen({ route }: Props) {
...
+                showMissingPriceImpactWarning && (
+                  <InLineNotification
+                    severity={Severity.Warning}
+                    title={t('swapScreen.missingSwapImpactWarning.title')}
+                    description={t('swapScreen.missingSwapImpactWarning.body')}
+                    style={styles.warning}
+                  />
+                )
+              )}
+            </>
```
jeanregisser, line 559: nit: then when `showPriceImpactWarning` or `showMissingPriceImpactWarning` is true it doesn't necessarily mean they should really be shown. Maybe this logic should be included directly when setting those variables?

jeanregisser (commented):
Great! 👍

locales/base/translation.json
kathaypacific: updated!

src/swap/SwapAmountInput.tsx
kathaypacific: oh yeah....i was inconsistent about this, i used TokenDisplay in some places in the details section and not here. i've updated to use TokenDisplay, thanks for the catch!

src/swap/SwapScreen.tsx
kathaypacific: yeah you're right...i found the conditions were getting long and confusing to read, so thought it might be clearer to have the container control what happens during loading / the overall hierarchy. but indeed the tradeoff is that the logic is split :/ i moved the logic back to the variables themselves

src/swap/SwapAmountInput.tsx
```diff
@@ -147,6 +148,29 @@ const SwapAmountInput = ({
           )}
         </Touchable>
       </View>
+      <View testID="SwapAmountInput/FiatValue">
+        <Text style={[styles.fiatValue, { opacity: loading || !inputValue || !token ? 0 : 1 }]}>
+          <TokenDisplay
+            amount={inputValue ?? 0}
+            showLocalAmount
+            showApprox
```
jeanregisser, line 156: oh nice!! I forgot about this 👍

jeanregisser (approved):
🚀

PR #4524, opened 2023-11-22
chore: reduce the number of colors in the colors enum

kathaypacific:
### Description

- greenBrand -> primary - greenBrand was used mostly for the loading spinner. seems like a safe change.
  Screenshot 2023-11-23 at 10 56 06

- onboardingLightBlue -> infoLight. the new color is much lighter but the only places using is are the dapp filter chips, and the "confirming" chip on the dapp shortcut claim card.
  Screenshot 2023-11-23 at 10 57 07

- goldDark -> warningDark (these have the same hex values, so no visual change. only the celo newsfeed author text color is now "warningDark" which is odd, but i think this is not so important for now)
- move modalBackdrop, dark_0_2 into the components that use them
- remove goldUI and goldFaint which were only used in the tertiary button. the tertiary button is only used in the last screen of the celo gold education flow, so i removed the tertiary button altogether and used a custom color on the education button. there is nothing in the design system files to suggest we want to have a tertiary button. no visual changes here.
- gray6 -> gray2 (affects only the bottom sheet handle bar + the disabled "resend sms" text on the phone verification screen)
Screenshot 2023-11-23 at 12 32 52
- lightGreen -> greenBackground (affects only the token select pill in the send amount screen)
Screenshot 2023-11-23 at 12 34 03
- greenLight -> greenBackground (affects only the dapps carousel "view all" circle background)
Screenshot 2023-11-23 at 12 34 27
- change the hex value of greenBackground to use what the home actions carousel uses since this is the most recent shade of light green. this should be safe to apply, since it is a background shade already.
Screenshot 2023-11-23 at 12 35 05

### Test plan

Example of onboardingLightBlue -> infoLight in the dapp filter:
Simulator Screenshot - iPhone 14 Pro - 2023-11-23 at 10 59 48

### Related issues

Relates to RET-913

### Backwards compatibility

Y


jeanregisser (approved):
👩‍🎨

PR #4527, opened 2023-11-22
feat(send): show fees on confirmation screen for all tokens

satish-ravi:
### Description

For the new send flow, pass down prepared tx and fee amount to send confirmation and display fees using those instead of the estimates from redux. Replaces FeeDrawer with LineItemRow since dek fee doesn't need to be displayed. 

Using preparedTx for sendPayment will be in a follow up.

### Test plan

Unit tests, manual

https://github.com/valora-inc/wallet/assets/5062591/dbba8fe2-f173-43c3-a256-bf10981b3936

### Related issues

- Part of ACT-968

### Backwards compatibility

N/A


cajubelt (approved):
nice!

PR #4533, opened 2023-11-23
chore: reduce the number of colors in the palette

kathaypacific:
### Description

Taking some shortcuts to reduce the number of similar colors we have.
- gray6 -> gray2 (there is hardly any noticeable difference between these 2 hex values, and affects only the bottom sheet handle bar + the disabled "resend sms" text on the phone verification screen)
- lightGreen -> greenBackground (affects only the token select pill in the send amount screen)
- greenLight -> greenBackground (affects only the dapps carousel "view all" circle background)
- change the hex value of greenBackground to use what the home actions carousel uses since this is the most recent shade of light green.

### Test plan

Example of the greenBackground on a pill in the send amount screen. It should be safe to use because it is already being used as a background shade in the home screen.
Simulator Screenshot - iPhone 14 Pro - 2023-11-23 at 12 11 30

### Related issues

Relates to RET-913

### Backwards compatibility

Y

### DIFF
--- a/src/components/BottomSheet.tsx
+++ b/src/components/BottomSheet.tsx
@@ -105,7 +105,7 @@ const BottomSheet = ({
 
 const styles = StyleSheet.create({
   handle: {
-    backgroundColor: Colors.gray6,
+    backgroundColor: Colors.gray2,
     width: 40,
   },
   headerContentSpacing: {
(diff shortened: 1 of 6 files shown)


kathaypacific:
this ended up being so small that i'll just add the changes to #4524

PR #4534, opened 2023-11-23
feat: add approval transactions to homefeed

kathaypacific:
### Description

This PR:
- add approval transactions to the transaction feed query
- add approval home feed item
- add approval details item

Approval transactions are only for Ethereum right now, so this change is effectively gated by the multichain transactions dynamic config.

I'll make a separate PR to update font styles in all the tx feed items so that the "Approval" doesn't look so bold.

### Test plan

The details view looks a little bare, I'm working with Nitya to refine what information to put here / add more stuff.

https://github.com/valora-inc/wallet/assets/20150449/18cc76f5-abde-4383-9534-dd9058d1fedb

### Related issues

- Fixes RET-903

### Backwards compatibility

Y


locales/base/translation.json
```diff
@@ -1965,5 +1965,12 @@
       "title": "You need more {{feeTokenSymbol}} for gas",
       "description": "Add {{feeTokenSymbol}} to cover your gas fees or adjust the amount you are sending"
     }
+  },
+  "transactionFeed": {
```
bakoushin, line 1969: This name invites to put everything related to txs feed here, not only what is related to approvals. Is that the intent?

src/icons/Activity.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react'
+import colors from 'src/styles/colors'
+import Svg, { Path } from 'svgs'
+
+interface Props {
+  size?: number
```
bakoushin, line 6: Do we plan to use `size`?

src/transactions/feed/detailContent/NetworkFeeRowItem.tsx
```diff
@@ -0,0 +1,73 @@
...
+import Colors from 'src/styles/colors'
+import { typeScale } from 'src/styles/fonts'
+import { Fee, TransactionStatus } from 'src/transactions/types'
+
+export interface Props {
+  fees: Fee[]
+  transactionStatus: TransactionStatus
+}
+
+export default function NetworkFeeRowItem({ fees, transactionStatus }: Props) {
```
bakoushin, line 14: Is that correct that we call this `Item` and are rendering only a single row, while the props are accepting the whole list (`Fee[]`)? It appears not particularly consistent, but maybe there is a smart plan behind the scenes?

src/transactions/feed/detailContent/NetworkFeeRowItem.tsx
```diff
@@ -0,0 +1,73 @@
...
+                : t('transactionFeed.networkFee')}
+            </Text>
+            <TokenDisplay
+              style={styles.currencyAmountPrimaryText}
+              amount={fees[0].amount.value}
+              tokenId={fees[0].amount.tokenId}
+              showLocalAmount={false}
+              showSymbol={true}
+              hideSign={true}
+              testID="TransactionDetails/EstimatedFee"
```
bakoushin, line 34: Should we keep the word `Estimated` here?

src/transactions/feed/TransactionDetailsScreen.test.tsx
```diff
@@ -313,6 +324,58 @@ describe('TransactionDetailsScreen', () => {
...
+
+  it(`renders the correct details for ${TokenTransactionTypeV2.Approval} transacton`, () => {
+    const { getByText, getByTestId } = renderScreen({
+      transaction: approvalTransaction({
+        status: TransactionStatus.Complete,
+      }),
+      storeOverrides: {
+        tokens: {
+          tokenBalances: {
+            'ethereum-sepolia:0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48': {
```
bakoushin, line 345: Do we need this token balance?

bakoushin (approved):
✅ Approval Approved

Left few nits here and there.

Note for your discussion with @nityarao: `approvedAmount === 0` means "approval revoked". Maybe we need a separate representation for that.

locales/base/translation.json
kathaypacific: yes that was my intention! transaction feed is really old and i notice that all the keys are on the top level and not very descriptive...so decided not to contribute to this pattern

src/icons/Activity.tsx
kathaypacific: oops, no lol classic copy and paste and didn't check 😅 i'll update

src/transactions/feed/detailContent/NetworkFeeRowItem.tsx
kathaypacific: i don't really have a smart plan but my idea is to make this row reusable at least for all the transaction details. maybe then we can move it back into the main screen component. for now, this is only shared between approvals and swaps - i'll work with Nitya to make the various transaction details more consistent

src/transactions/feed/TransactionDetailsScreen.test.tsx
kathaypacific: yes, although technically i only need this object in order to get the token symbol for the message (i.e. you've approved USDC). i can remove some unnecessary properties

kathaypacific:
> ✅ Approval Approved
> 
> Left few nits here and there.
> 
> Note for your discussion with @nityarao: `approvedAmount === 0` means "approval revoked". Maybe we need a separate representation for that.

thanks for flagging this! i hadn't considered this. i've updated the details component to have separate strings for finite / infinite / revoke approval now so that it is more clear. i'll work with Nitya to see if we want any other UI changes for the revoke scenario

src/icons/Activity.tsx
```diff
@@ -0,0 +1,21 @@
...
+  color?: colors
+}
+
+function Activity({ color = colors.successDark, size = 24 }: Props) {
+  return (
+    <Svg width={size} height={size} fill="none">
```
bakoushin, line 12: I believe we also need a `viewBox` property for [correct scaling](https://graphicdesign.stackexchange.com/a/71574).

```suggestion
    <Svg width={size} height={size} fill="none" viewBox="0 0 24 24">
```

src/transactions/feed/detailContent/NetworkFeeRowItem.tsx
bakoushin: I was confused that it renders only a first item of the array. What if instead of passing the whole fees array in the props, we pass just a single fee item (the very same it will render)?

src/transactions/feed/TransactionDetailsScreen.test.tsx
bakoushin: does this `tokenBalances` object always have all supported tokens, even if the balance is 0?

because what if we had an approval for a token sometimes earlier, but then drained its balance? can we rely that it will be there?

kathaypacific: good question, yes the token info will be there even if the balance is zero. we fetch the token balances from the getTokenInfo CF and combine the last known balance with it before storing it all in redux https://github.com/valora-inc/wallet/blob/2d551d94a99955a93564fc9792d7a0e033dda9d9/src/tokens/saga.ts#L195-L207

src/transactions/feed/detailContent/NetworkFeeRowItem.tsx
kathaypacific: i passed the whole fees array here because then consumers don't need to know that the network fee is the first element of the fees array. otherwise every transaction feed item needs to have the logic `fees[0] ? <NetworkFee /> : null`. it also didn't seem unreasonable to assume that we might one day change the network fees to be a different element index or a combination of the elements in the fees array...

PR #4537, opened 2023-11-24
Fees in pending tx

dievazqu:
### Related issues

- Fixes #[issue number here]


src/fees/hooks.ts
```diff
@@ -27,7 +27,9 @@ export function usePaidFees(fees: Fee[]) {
   const securityFeeAmount = fees.find((fee) => fee.type === TransactionFeeType.SecurityFee)
   const dekFeeAmount = fees.find((fee) => fee.type === TransactionFeeType.EncryptionFee)
   const feeCurrencyInfo = Object.entries(tokensByCurrency).find(
-    ([_, tokenInfo]) => tokenInfo?.address === securityFeeAmount?.amount.tokenAddress
+    ([_, tokenInfo]) =>
+      tokenInfo?.address === securityFeeAmount?.amount.tokenAddress ||
+      tokenInfo?.tokenId === securityFeeAmount?.amount.tokenId
```
dievazqu, line 32: Adding this to be compatible with new versions that uses `tokenId`.
Another option would be to pass also the token address to the transaction confirmed event. Let me know if you prefer that solution.

src/transactions/utils.ts
```diff
@@ -47,3 +54,43 @@ export function groupFeedItemsInSections(
       data: value.data,
     }))
 }
+
+export function* buildBaseTransactionReceipt(
```
dievazqu, line 58: I wasn't sure where to put this function, any suggestion?

src/transactions/utils.ts
```diff
@@ -47,3 +54,42 @@ export function groupFeedItemsInSections(
...
+  receipt: TransactionReceipt | CeloTxReceipt,
+  networkId: NetworkId
+) {
+  const tokensById = yield* select((state) => tokensByIdSelector(state, [networkId]))
+
+  // Receipt gas information is based on the native token (there is no feeCurrency)
```
dievazqu, line 64: I have tested the flow with contract kit and viem. And the receipt doesn't have any `feeCurrency` information, and checking the numbers with celo scan, the returned values represents the cost in Celo (I assume it's gonna be the native token for each chain). But we might want to revisit this logic in the future.

dievazqu: Actually, I think this is a bug in celo scan. The `feeCurrency` is the one that we configure when we send the transaction. If you check how celo scan shows the transaction details:

Screenshot 2023-11-24 at 16 36 14
Screenshot 2023-11-24 at 16 36 45

You can see that the cUSD transfers add up exactly to the transaction fee value.
The same with the gas price, it should be in cUSD.

So, I'm gonna re do this logic to make it right in a follow up PR.

src/transactions/utils.ts
kathaypacific: i think the scope of abstraction is a little odd, because it seems like it tries to turn the transaction receipt into the arguments that the reducer accepts except the reducer then needs to take the output of this function and do more logic (buildGasFees) for the fees. 

i have a couple of suggestions, maybe there are others that make sense too:
1. have a handleTransactionConfirmed saga that is responsible for dispatching transactionConfirmed given the receipt (and formats the fees in the format that can be directly saved into redux).
2. have a toNetworkFees saga that is just responsible for getting the network fees format from the transaction receipt

src/transactions/utils.ts
kathaypacific: ooooh yes i remember this now, this is probably why we didn't implement the fees before.

is the proposal that we add the feeCurrency to the standby transaction, and then rely on the gas value in the receipt being for that feeCurrency? 

i think that we shouldn't merge any logic that we know is incorrect - it seems like assuming that the gas token is always the native token is only correct for Ethereum, so maybe we should only add fees for Ethereum right now and add a TODO for Celo (which i assume is the follow up PR that you mention). wdyt?

src/fees/hooks.ts
kathaypacific: this seems okay, it is only used in the transfer item details screen which. might get a makeover soon anyway

kathaypacific (commented):
looking good!!

src/transactions/utils.ts
dievazqu: Yes, I thought about this, I think the easiest solution would be to add the `feeCurrency` in the standby transaction. I don't like the fact that we will be storing a duplicated feeCurrency also in the `fees` to display it. But I think it's ok...

Yes, I'll merge this only for Eth, and I'll fix this flow for Celo in a next PR.

src/transactions/utils.ts
dievazqu: Hmm... Creating a new saga for this? It sounds a lot for just a mapper.
What would be the difference between a saga and this? Maybe I can move the `buildGasFees` function to here.

What do you think if I put this logic inside the `transactionConfirmed` action?
I think that's the place to put this logic.
But since all our actions are pure, I didn't want to add code there, also I don't know if I'm able to access the redux state from there, since I need the token list to get some info.

I think we can think this code as a "parser" from the blockchain response into our schema.

kathaypacific: hum, i'm not sure i understand the concern. what you've got here in buildBaseTransactionReceipt is already a saga? (maybe the concern is that i am using "saga" and "generator function" interchangably?) i'm only proposing that we keep logic in one place rather than having it broken up. most of the logic here i see is trying to massage the tx receipt to be what we want to store in redux. so i thought having a generator function that takes a receipt and calls the `transactionConfirmed` action with the expected params seemed reasonable. 

i think redux actions are meant to be simple and you won't have access to the redux state from there. i also thought maybe this logic could live in the reducer, but we shouldn't pass unserialisable data to the reducers (i believe the viem tx receipts have bigInt's). so i think having a generator function / saga is probably the easiest solution

src/transactions/saga.ts
```diff
@@ -307,3 +313,58 @@ export function* transactionSaga() {
...
+
+export function* handleTransactionReceiptReceived(
+  txId: string,
+  receipt: TransactionReceipt | CeloTxReceipt,
+  networkId: NetworkId
+) {
+  const tokensById = yield* select((state) => tokensByIdSelector(state, [networkId]))
+
+  // TODO: Receive feeCurrencyId from transaction request.
+  const feeCurrencyId = getTokenId(networkId)
```
jeanregisser, line 325: Yes we discussed the fact the `feeCurrency` is missing from the TX receipt with cLabs and it's by design. Things that are already in the TX itself aren't present in the receipt.

https://valora-app.slack.com/archives/CNJ7KTHQU/p1698843275835969?thread_ts=1698688855.441649&cid=CNJ7KTHQU

Do you want to address the TODO in a separate PR? My only concern is it should be done relatively close to merging this one otherwise we risk merging this and display incorrect fee to the user if it's paid with another currency.

src/transactions/saga.ts
```diff
@@ -307,3 +313,58 @@ export function* transactionSaga() {
...
+  const feeTokenInfo = tokensById[feeCurrencyId]
+
+  if (!feeTokenInfo) {
+    Logger.warn(`No information found for token ${feeCurrencyId} in network ${networkId}`)
+  }
+
+  const gasFeeInWei = new BigNumber(receipt.gasUsed.toString()).times(
```
jeanregisser, line 332: nit: it's not in wei, it's in the smallest unit of the feeCurrency.
wei is really [10^-18 ETH](https://ethereum.org/gl/developers/docs/intro-to-ether/#denominations). We should avoid using this naming to avoid bugs. Unless it's really wei we want.

```suggestion
  const gasFeeInSmallestUnit = new BigNumber(receipt.gasUsed.toString()).times(
```

src/transactions/saga.ts
```diff
@@ -307,3 +313,58 @@ export function* transactionSaga() {
...
+const buildGasFees = (feeCurrencyInfo: BaseToken, gasFeeInWei: BigNumber): Fee[] => [
+  //
+  {
+    type: 'SECURITY_FEE',
+    amount: {
+      value: gasFeeInWei.shiftedBy(-feeCurrencyInfo.decimals).toFixed(),
+      tokenId: feeCurrencyInfo.tokenId,
+    },
+  },
+]
```
jeanregisser, line 370: nit: using the `function` syntax is easier to spot functions while reading the code IMHO.
```suggestion
function buildGasFees(feeCurrencyInfo: BaseToken, gasFeeInWei: BigNumber): Fee[] {
  return [
    //
    {
      type: 'SECURITY_FEE',
      amount: {
        value: gasFeeInWei.shiftedBy(-feeCurrencyInfo.decimals).toFixed(),
        tokenId: feeCurrencyInfo.tokenId,
      },
    },
  ]
}
```

jeanregisser (commented):
Great! 🚀

src/transactions/saga.test.ts
```diff
@@ -180,6 +182,8 @@ describe('watchPendingTransactions', () => {
       status: 'success',
       blockNumber: BigInt(123),
       transactionHash,
+      gasUsed: 2_000_000,
+      effectiveGasPrice: 1e9,
```
kathaypacific, line 186: should we add a test where these are used?

src/transactions/saga.ts
```diff
@@ -307,3 +313,58 @@ export function* transactionSaga() {
...
+        ? TransactionStatus.Failed
+        : TransactionStatus.Complete,
+  }
+
+  // Only add fee data in non-celo transactions
+  const feeDetails =
+    !feeTokenInfo || networkId === networkConfig.defaultNetworkId
+      ? {}
+      : {
+          fees: buildGasFees(feeTokenInfo, gasFeeInWei),
```
kathaypacific, line 350: nit: any reason why you decided to create this object and method? why not have the condition directly for the `fees` property of the action payload, where the value would be an empty array or an array with the security fee? e.g. 
```
transactionConfirmed(txId, {
      ...baseDetails,
      fees: !feeTokenInfo || networkId === networkConfig.defaultNetworkId ? [] : [...]
    })
```
?

src/transactions/saga.ts
kathaypacific: i agree, would be nice to address this TODO asap. i think there's no risk that the wrong fee is displayed though because further in this function, we omit the fees calculation if the network is Celo?

src/transactions/saga.ts
```diff
@@ -307,3 +313,58 @@ export function* transactionSaga() {
...
+  networkId: NetworkId
+) {
+  const tokensById = yield* select((state) => tokensByIdSelector(state, [networkId]))
+
+  // TODO: Receive feeCurrencyId from transaction request.
+  const feeCurrencyId = getTokenId(networkId)
+  const feeTokenInfo = tokensById[feeCurrencyId]
+
+  if (!feeTokenInfo) {
+    Logger.warn(`No information found for token ${feeCurrencyId} in network ${networkId}`)
```
kathaypacific, line 329: this should never happen right? then should this be Logger.error?

kathaypacific (approved):
looks good from my side, i left a few small comments that would be nice to think about / address (also i see a few comments from Jean too)

src/transactions/saga.ts
dievazqu: Yes, I prefer to address this TODO separate, because that change will require some refactor since it's not straight forward to get the `feeCurrency`, there is some abstraction in the middle.
With the current code, the CELO fees are ignored.

src/transactions/saga.ts
dievazqu: Nop, let me clean this. I think I'm gonna leave the function because I think it's easier to read.

dievazqu:
@kathaypacific @jeanregisser Thanks for the suggestions!
The code is ready for another review.

jeanregisser (approved):
🚀

PR #4539, opened 2023-11-24
fix(swap): ensure fiat token value works with comma decimal separator

kathaypacific:
### Description

As the title

### Test plan

Simulator Screenshot - iPhone 14 Pro - 2023-11-24 at 14 39 09

### Related issues

Related to RET-928

### Backwards compatibility

Y


jeanregisser (approved):
👍

PR #4542, opened 2023-11-24
fix(swaps): ensure clear button works on swap input

kathaypacific:
### Description

Context https://valora-app.slack.com/archives/C04B61SJ6DS/p1700841917661059

A change was made to require users to dismiss the keyboard before proceeding with the swap, to ensure all warnings are visible. As a side effect, dismissing the keyboard => the textinput is not focused => cannot clear the input using the clear button. This PR adds the prop to allow actions to be handled on tap.

### Test plan

https://github.com/valora-inc/wallet/assets/20150449/8fd10e6f-0a46-49e8-8dd2-020b76a5cc32

### Related issues

n/a

### Backwards compatibility

Y

### DIFF
--- a/src/swap/SwapScreen.tsx
+++ b/src/swap/SwapScreen.tsx
@@ -529,7 +529,10 @@ export function SwapScreen({ route }: Props) {
         left={<BackButton />}
         title={t('swapScreen.title')}
       />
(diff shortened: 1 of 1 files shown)


jeanregisser (approved):
🙌

PR #4550, opened 2023-11-27
chore: ensure homefeed items have the same font

kathaypacific:
### Description

The new design system fonts are ever so slightly different to the deprecated ones. This PR prevents subtle differences in the feed items, when the Approvals items is merged.

I also put on my cowboy hat and made some adjustments to the font scaling behaviour for the transfer and swap feed items. i opted to ellipsize the title and subtitle to make room for long token values on the right hand side, also made more of an allowance for the width of the transaction values. my goal was to stop items from becoming 3-4 lines when either the token value was large or when the feed item description is long. there is a detail view on tap, so i think some ellipsized information is okay (the token values tell pretty much the whole story anyway)

### Test plan
Simulator Screenshot - iPhone 14 Pro - 2023-11-24 at 16 54 01

With font scaling, to show the ellipsis behaviour
Simulator Screenshot - iPhone 14 Pro - 2023-11-24 at 16 53 37

### Related issues

Related to RET-903

### Backwards compatibility

Y


src/transactions/feed/TransactionFeed.test.tsx
```diff
@@ -168,7 +168,12 @@ describe('TransactionFeed', () => {
     expect(tree.queryByTestId('NoActivity/loading')).toBeNull()
     expect(tree.queryByTestId('NoActivity/error')).toBeNull()
     expect(mockFetch).toHaveBeenCalledTimes(1)
-    expect(tree).toMatchSnapshot()
+    expect(tree.getAllByTestId('TransferFeedItem').length).toBe(1)
+    expect(
+      within(tree.getByTestId('TransferFeedItem')).getByTestId('TransferFeedItem/title')
+    ).toHaveTextContent(
+      'feedItemReceivedTitle, {"displayName":"feedItemAddress, {\\"address\\":\\"0xd683...ea33\\"}"}'
+    )
```
kathaypacific, line 176: there is a separate unit test specifically for TransferFeedItem so I did not add too many assertions to this TransactionFeed test case

bakoushin (approved):
🎨

jeanregisser (commented):
Thanks! What's the impact of the changes when the user uses a really large font?

kathaypacific:
> Thanks! What's the impact of the changes when the user uses a really large font?

this is what it looks like with very large font:
Simulator Screenshot - iPhone 14 Pro - 2023-11-28 at 09 29 38

my changes only affect the homefeed item content, and increased the width allowance for the token display. but it does not affect the token display actually, i thought about ellipsizing everything so that we keep 2 lines only but i think the feed items do not make sense without the token symbol. so i've left it to scale to multiple lines, since it's existing behaviour.

note that i have a screenshot for slightly increased font in the PR description, where i think it looks good 😅

jeanregisser (approved):

PR #4559, opened 2023-11-29
fix(swap): incorrect network clients used for multichain

jeanregisser:
### Description

No wonder the demo didn't work 🤦 😅 

### Test plan

- Updated unit tests
- Successfully tested manually using a local forked Ethereum node and swap provider fixed to 0x (there's a bug with our squid integration when swapping native Eth, I'll make another PR for it in our API)

https://github.com/valora-inc/wallet/assets/57791/9e9d5569-cd87-4d1b-bda9-2f2e51c8176c

### Related issues

- Fixes RET-912

### Backwards compatibility

Yes


src/swap/saga.test.ts
```diff
@@ -608,29 +722,159 @@ describe(swapSubmitPreparedSaga, () => {
...
+      swapTxEffectiveGasPrice: 5_000_000_000,
+      swapTxFeeCurrency: undefined,
+      swapTxFeeCurrencySymbol: 'ETH',
+      swapTxGas: 1_325_000,
+      swapTxMaxGasFee: 0.0159,
+      swapTxMaxGasFeeUsd: 0.00795,
+      swapTxGasUsed: 371_674,
+      swapTxGasFee: 0.00185837,
+      swapTxGasFeeUsd: 0.000929185,
+      swapTxHash: '0x2',
```
kathaypacific, line 822: holy heck, this is a long list 😅 do we need both this list and the assertions below?

jeanregisser: Yes just to be sure the math checks out and we don't base our analytics on wrong numbers.

This test is really a duplicate of the Celo one above, so I went ahead and used `it.each` to make it more succint.

src/swap/saga.test.ts
```diff
@@ -596,257 +596,168 @@ describe(swapSubmitPreparedSaga, () => {
...
+        .call([publicClient[network], 'waitForTransactionReceipt'], { hash: '0x2' })
+        .run()
+
+      expect(mockViemWallet.signTransaction).toHaveBeenCalledTimes(2)
+      expect(mockViemWallet.sendRawTransaction).toHaveBeenCalledTimes(2)
+      expect(loggerErrorSpy).not.toHaveBeenCalled()
+
+      expect(ValoraAnalytics.track).toHaveBeenCalledTimes(1)
+      expect(ValoraAnalytics.track).toHaveBeenCalledWith(SwapEvents.swap_execute_success, {
+        toToken: sellTokenAddress,
```
kathaypacific, line 681: are these correct? i thought the sellToken was the fromToken?

kathaypacific (approved):
nice!! mostly good, i think the buy/sell token declarations need to be switched around though?

src/swap/saga.test.ts
jeanregisser: Good catch, I'll let Copilot refactor this and I'm guilty of not checking this important detail 🤦

PR #4560, opened 2023-11-29
fix(swap): user stuck after cancelling PIN input in the viem flow

jeanregisser:
### Description

With viem enabled for swap, I noticed cancelling PIN input left the user on the loader screen.
This is because the exception happens in the `catch` handler itself.

Maybe we could have a helper to ensure the required action is always dispatched in the error handler.
For now, I've fixed the bug and moved the dispatch toward the top of the catch handler.

### Test plan

**before**

https://github.com/valora-inc/wallet/assets/57791/9cbc1113-1d7c-484e-b73b-5caa7f7202fd

**after**

https://github.com/valora-inc/wallet/assets/57791/f5b51ce4-6c62-49b6-9917-0ce98fce7772

### Related issues

- N/A

### Backwards compatibility

Yes


src/swap/saga.ts
```diff
@@ -349,14 +349,14 @@ function getSwapTxsReceiptAnalyticsProperties(
   )
 
   const approveTx = trackedTxs.length > 1 ? txs[0] : undefined
-  const swapTx = txs[txs.length - 1]
+  const swapTx = trackedTxs.length > 0 ? txs[txs.length - 1] : undefined
```
kathaypacific, line 352: do we need these changes?

kathaypacific (approved):
thanks!!

src/swap/saga.ts
jeanregisser: yes this is what caused the initial crash, `swapTx` could be undefined, but TS didn't know it.

PR #4562, opened 2023-11-29
chore: make spendToken optional in prepareTransactions

kathaypacific:
### Description

For dapp transactions, we most likely will not know what the transaction is doing and cannot provide spendAmount and spendToken.

This PR updates prepareTransactions to function when spendToken and spendAmount are not available.
- if the user has enough gas tokens to pay for gas, then return a `possible` result
- if the user does not have enough balance of any gas token, then return a `not-enough-balance-for-gas` result.

This is brittle if the transaction is for the chosen gas currency (e.g. if the transaction is to swap all Celo to another token, and the gas logic chooses Celo to pay gas, then the transaction would fail) however we can't do more precise logic when we do not decode the transaction.

### Test plan

Unit tests

### Related issues

Related to RET-905

### Backwards compatibility

Y


src/viem/prepareTransactions.ts
```diff
@@ -256,7 +267,15 @@ export async function prepareTransactions({
...
+  // if no spend amount is provided, we conclude that the user does not have enough balance to pay for gas
+  if (!spendToken) {
+    return {
+      type: 'not-enough-balance-for-gas',
+      feeCurrencies,
+    } satisfies PreparedTransactionsNotEnoughBalanceForGas
+  }
+
+  // let's see if we can decrease the spend amount, if provided
   const result = maxGasFees.find(({ feeCurrency }) => feeCurrency.tokenId === spendToken.tokenId)
```
jeanregisser, line 279: nit: maybe?
```suggestion
  // let's see if we can decrease the spend amount, if provided
  const result = maxGasFees.find(({ feeCurrency }) => feeCurrency.tokenId === spendToken?.tokenId)
```

src/viem/prepareTransactions.ts
```diff
@@ -237,8 +245,11 @@ export async function prepareTransactions({
       // Not enough balance to pay for gas, try next fee currency
       continue
     }
-    const spendAmountDecimal = spendTokenAmount.shiftedBy(-spendToken.decimals)
+    const spendAmountDecimal = spendToken
+      ? spendTokenAmount.shiftedBy(-spendToken.decimals)
+      : spendTokenAmount
```
jeanregisser, line 250: nit: maybe?
```suggestion
    const spendAmountDecimal = spendTokenAmount.shiftedBy(-spendToken?.decimals ?? 0)
```

jeanregisser (approved):
🚀

jeanregisser:
> This is brittle if the transaction is for the chosen gas currency (e.g. if the transaction is to swap all Celo to another token, and the gas logic chooses Celo to pay gas, then the transaction would fail) however we can't do more precise logic when we do not decode the transaction.

Yes, if the dapp TX specifies `gas` we won't try to call `estimateGas` ourselves and won't notice it fails.
Because of this, I wonder if we should always strip out the provided `gas` param?
unless both `gas` and `feeCurrency` are set (meaning the dapp knows what they are doing).

kathaypacific:
> > This is brittle if the transaction is for the chosen gas currency (e.g. if the transaction is to swap all Celo to another token, and the gas logic chooses Celo to pay gas, then the transaction would fail) however we can't do more precise logic when we do not decode the transaction.
> 
> Yes, if the dapp TX specifies `gas` we won't try to call `estimateGas` ourselves and won't notice it fails. Because of this, I wonder if we should always strip out the provided `gas` param? unless both `gas` and `feeCurrency` are set (meaning the dapp knows what they are doing).

yes i think that's a good suggestion, i'll do that from the dapps flow in a follow up

PR #4563, opened 2023-11-29
chore: add static gas padding when no gas is provided

kathaypacific:
### Description

I came across this when writing a unit test and it surprised me that the static padding for non native gas currency was not applied if the tx does not have a `gas` parameter. i then saw a TODO to add this, that is what this PR is doing:
- ensure static gas padding added for non native tokens when using `tryEstimateTransaction`
- update unit tests

### Test plan

Unit tests

### Related issues

Related to RET-905

### Backwards compatibility

Y

### DIFF
--- a/src/viem/prepareTransactions.ts
+++ b/src/viem/prepareTransactions.ts
@@ -85,15 +85,13 @@ export async function tryEstimateTransaction({
   baseTransaction,
   maxFeePerGas,
   maxPriorityFeePerGas,
-  feeCurrencySymbol,
-  feeCurrencyAddress,
+  feeCurrency,
 }: {
   client: Client
   baseTransaction: TransactionRequest
   maxFeePerGas: bigint
   maxPriorityFeePerGas?: bigint
-  feeCurrencySymbol: string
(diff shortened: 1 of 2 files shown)


src/viem/prepareTransactions.ts
```diff
@@ -85,33 +85,33 @@ export async function tryEstimateTransaction({
...
 
-  // TODO maybe cache this? and add static padding when using non-native fee currency
+  // TODO maybe cache this?
   try {
-    tx.gas = await estimateGas(client, {
+    const estimatedGas = await estimateGas(client, {
       ...(tx as TransactionRequestBase),
       account: tx.from,
     })
+    tx.gas = estimatedGas + BigInt(feeCurrency.isNative ? 0 : STATIC_GAS_PADDING)
```
jeanregisser, line 112: Sorry the comment may have been a bit misleading.
The padding would be necessary only if caching this value and the initial value was without a feeCurrency set.

We don't need the padding here if we call `estimateGas` with the `feeCurrency` set.

jeanregisser (changes requested):

PR #4564, opened 2023-11-29
chore(swaps): Remove buy amount experiment

dievazqu:
### Description

As title says

### Related issues

- Fixes RET-920

### DIFF
--- a/src/statsig/constants.ts
+++ b/src/statsig/constants.ts
@@ -68,12 +68,6 @@ export const ExperimentConfigs = {
       discoverCopyEnabled: false,
     },
   },
-  [StatsigExperiments.SWAP_BUY_AMOUNT]: {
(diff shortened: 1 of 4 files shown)


jeanregisser (commented):
Great! 🙌 

Could we also remove this field and related code?
https://github.com/valora-inc/wallet/blob/1b07f2114d78f3fee7ed149115bfa0f182c68686/src/swap/types.ts#L23

dievazqu:
> Great! 🙌
> 
> Could/should we also remove this field and related code?
> 
> https://github.com/valora-inc/wallet/blob/1b07f2114d78f3fee7ed149115bfa0f182c68686/src/swap/types.ts#L23

Yes, I thought about removing that too, but I didn't want to mess it up. But yes, I'll take a deeper look and remove that too.

jeanregisser:
Cool, as you want, feel free to use another PR if you prefer.

PR #4566, opened 2023-11-30
fix(cico): Add sorting to CICO currency bottom sheet

finnian0826:
### Description
No ordering before, now tokens are sorted before being displayed in the bottom sheet.

Dynamic config: https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/dynamic_configs/cico_token_info

Dynamic config to review: https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/dynamic_configs/cico_token_info/preview/6mIWqOsAuVnZiwwU7D4EL3

### Test plan

Updated unit tests to check for ordering.

Manual test shows correct order, logged ordering to confirm that sorting was done correctly.
cicoOrder

### Related issues

N/A

### Backwards compatibility

Yes


src/fiatExchanges/FiatExchangeCurrencyBottomSheet.tsx
```diff
@@ -26,13 +27,15 @@ function FiatExchangeCurrencyBottomSheet({ route }: Props) {
...
   const spendTokens = useSpendTokens()
-  const tokenList =
+  const unsortedTokenList =
     flow === FiatExchangeFlow.CashIn
       ? cashInTokens
       : flow === FiatExchangeFlow.CashOut
       ? cashOutTokens
       : spendTokens
 
+  const tokenList = unsortedTokenList.sort(sortFirstStableThenCeloThenOthersByUsdBalance)
```
satish-ravi, line 37: can we wrap this in a useMemo? Also this needs unit tests

finnian0826: would the was to unit test this be to do a snapshot comparison for what the component renders? I'm not quite sure how else to unit test this.

satish-ravi: you could select all token balance items (I think they should have a common testID) and assert the ordering with mocked input.

src/tokens/utils.ts
```diff
@@ -87,6 +87,22 @@ export function sortFirstStableThenCeloThenOthersByUsdBalance(
...
+  if (!cicoTokenInfo.cicoOrder[token1.tokenId] && !cicoTokenInfo.cicoOrder[token2.tokenId]) {
+    return sortFirstStableThenCeloThenOthersByUsdBalance(token1, token2)
+  }
+  if (!cicoTokenInfo.cicoOrder[token1.tokenId]) {
+    return 1
+  }
+  if (!cicoTokenInfo.cicoOrder[token2.tokenId]) {
+    return -1
+  }
+  return cicoTokenInfo.cicoOrder[token1.tokenId] < cicoTokenInfo.cicoOrder[token2.tokenId] ? -1 : 1
```
jophish, line 103: It's possible that multiple tokens map to the same value, right? It seems like the dynamic config just specifies token "priority", rather than actual position in the list, so it's possible that two tokens have the same priority? Can we handle the case where the values are the same here?

finnian0826: good catch, I think hopefully the dynamic config will have unique numbers for the different tokens so it can sort correctly, but I added a safeguard in case they are the same.

src/statsig/constants.ts
```diff
@@ -116,4 +116,10 @@ export const DynamicConfigs = {
       maxSlippagePercentage: '0.3',
     },
   },
+  [StatsigDynamicConfigs.CICO_TOKEN_INFO]: {
+    configName: StatsigDynamicConfigs.CICO_TOKEN_INFO,
+    defaultValues: {
+      cicoOrder: {} as { [key: string]: number },
```
satish-ravi, line 122: Wonder if we should do `{ <tokenId>: { <propName>: <propValue> }}` instead? This will make the current dynamic config value more verbose (`cicoOrder` will be repeated on every token), but the benefit is that this will match the format of `/tokensInfo` endpoint and we can just deep merge with it if we want to have more dynamic overrides for any token metadata. We could also make dynamic config name more generic to allow any metadata, but maybe it's best to make them more single purpose (like cico specific stuff in one config) so if at all we have a bunch of overrides, the number of rules doesn't explode. Thoughts? cc @cajubelt @jophish @jeanregisser

src/tokens/utils.ts
```diff
@@ -87,6 +87,25 @@ export function sortFirstStableThenCeloThenOthersByUsdBalance(
...
+  if (
+    (!cicoTokenInfo.cicoOrder[token1.tokenId] && !cicoTokenInfo.cicoOrder[token2.tokenId]) ||
+    cicoTokenInfo.cicoOrder[token1.tokenId] === cicoTokenInfo.cicoOrder[token2.tokenId]
+  ) {
+    return sortFirstStableThenCeloThenOthersByUsdBalance(token1, token2)
+  }
+  if (!cicoTokenInfo.cicoOrder[token1.tokenId]) {
+    return 1
+  }
+  if (!cicoTokenInfo.cicoOrder[token2.tokenId]) {
```
satish-ravi, line 103: can we add test cases for missing / equal balances?

satish-ravi (approved):

PR #4567, opened 2023-12-01
fix(send): use correct fee for total in send confirmation screen

satish-ravi:
### Description

Total was still being computed using redux fees for the new send flow. This updates it to use the feeAmount arg.

### Test plan

Unit tests, manually by going through new send flow

| Fee currency same as send currency | Fee currency different from send currency |
|--------|--------|
|  |  |

Note that the fees in different currency add up correctly on the screenshot. This is because we use the same exchange rate for all tokens on testnet

### Related issues

- Fixes ACT-968

### Backwards compatibility

Yes


src/send/SendConfirmation.test.tsx
```diff
@@ -207,10 +207,10 @@ describe('SendConfirmation', () => {
     const { getByTestId } = renderScreen({}, mockScreenPropsWithPreparedTx)
 
     const feeComponent = getByTestId('LineItemRow/SendConfirmation/fee')
-    expect(getElementText(feeComponent)).toEqual('0.004 CELO')
+    expect(getElementText(feeComponent)).toEqual('0.01 CELO')
 
     const totalComponent = getByTestId('TotalLineItem/Total')
-    expect(getElementText(totalComponent)).toEqual('~1.02 cUSD')
+    expect(getElementText(totalComponent)).toEqual('~1.05 cUSD')
```
satish-ravi, line 213: this only shows 2 decimals currently if the amount is >= 0.01, so if fees is <0.01, they won't show up here. Checking with product on how many decimals we want to show here and will make a follow up issue if this needs to be changed. (Needs some refactor to the TokenDisplay component)

cajubelt: is that the case even if the fee is in ETH? like we don't show ETH fees if they're less than 0.01 ETH?

PR #4573, opened 2023-12-01
chore(qrcode): Clean up QR code experiment

jh2oman:
### Description

Cleaning up exp that has been rolled out since april https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/experiments/qr_part_2/summary

### Test plan

Manually checked out all the QR instances in the wallet and everything looks good.


src/qrcode/NewQRCode.tsx
MuckT, line 1: Since we are removing the old `QRCode.tsx` can we rename this to `QRCode.tsx`?

src/statsig/constants.ts
```diff
@@ -1,26 +1,7 @@
...
-  StatsigDynamicConfigs,
-  StatsigExperiments,
-  StatsigFeatureGates,
-  StatsigLayers,
-} from 'src/statsig/types'
+import { StatsigDynamicConfigs, StatsigExperiments, StatsigFeatureGates } from 'src/statsig/types'
 import networkConfig from 'src/web3/networkConfig'
 
-export const LayerParams = {
-  // TODO(ACT-659): refactor to imitate defaultExperimentParamValues (more type safe, less boilerplate)
```
cajubelt, line 12: seems like we didn't want to keep this enum around in the long term anyway, we wanted to do something else. can we just delete it now?

PR #4574, opened 2023-12-01
chore(analytics): add networkId to provider selection analytics event data

cajubelt:
### Description

see title

### Test plan

update jest tests

### Related issues

fixes https://linear.app/valora/issue/ACT-992/ability-to-group-analytics-events-by-network-id

### Backwards compatibility

na


src/fiatExchanges/utils.tsx
```diff
@@ -389,5 +389,6 @@ export function getProviderSelectionAnalyticsData({
       (coinbasePayAvailable ? 1 : 0) +
       (legacyMobileMoneyProviders?.length ?? 0) +
       normalizedQuotes.length,
+    networkId: tokenInfo?.networkId,
```
cajubelt, line 392: this is why networkId can be undefined

jh2oman (approved):

src/fiatExchanges/utils.tsx
satish-ravi: could we make the tokenInfo required here and make networkId also required? Seems like the only usage of this function ensures tokenInfo is always defined
https://github.com/valora-inc/wallet/blob/0176dc6103d3c595ee136a57fd795afeae1faff4/src/fiatExchanges/SelectProvider.tsx#L105-L107

PR #4576, opened 2023-12-01
feat(token-import): Validate token address

denvalora:
### Description

- Once the user has entered an address for a token **and pressed** "search" on the keyboard (or pressed "Paste" button), address validation is performed. Note: when the address field is input, it won't automatically start validation process.
- When the import button is pressed, the address is converted to a tokenId. In [RET-891/save-and-display-custom-tokens](https://linear.app/valora/issue/RET-891/save-and-display-custom-tokens) it will be saved to the Redux state.

### Test plan

User prevented from importing (Import button is disabled), when token address
- Already supported token, e.g. `0x874069fa1eb16d44d622f2e0ca25eea172369bc1` (cUSD in Alfajores)
- ~Already imported token~
- Invalid EVM address

~Manually run the app to import a token address, and then try to reimport it to see error message. This will verify that the state has been updated.~

### Related issues

- validates and saves to the state the address for [RET-892/lookup-contract-info](https://linear.app/valora/issue/RET-892/lookup-contract-info). In a following PR the contract will be fetched and validated as well.

### Backwards compatibility

Yes.


src/tokens/slice.ts
```diff
@@ -100,6 +100,7 @@ export interface TokenBalancesWithAddress {
 
 export interface State {
   tokenBalances: StoredTokenBalances
+  importedTokenIds: string[]
```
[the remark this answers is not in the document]
kathaypacific: ah, i see where the confusion comes from. since the main focus for this task is the token lookup, i think it's probably sensible to defer the token already imported error until the next task. you can add a TODO in the code. once the tokens are saved, it is only a few more lines of code to display the error (compared to if you do it now, to get the error you basically need to implement the whole token saving part).

we should avoid storing derived state in redux (same as any other data source) - we'll need to eventually be able to store the token info and display it in some screens, so we'll need more than just the token id. we can create selectors to get a list of token id's that are imported, but we should not have a specific state that tracks the token id's in this case

src/tokens/TokenImport.tsx
```diff
@@ -95,9 +141,11 @@ export default function TokenImportScreen(_: Props) {
             label={t('tokenImport.input.tokenSymbol')}
             value={tokenSymbol}
             onChangeText={setTokenSymbol}
-            editable={isValidAddress(tokenAddress)}
+            editable={addressState === AddressState.Valid}
```
dievazqu, line 144: I think this shouldn't be editable at all, right? We should use the `tokenSymbol` that comes from the blockchain data, so the user shouldn't fill the field.

dievazqu (commented):
Maybe I'm missing something, but I think the `tokenSymbol` field shouldn't be editable.

denvalora:
1. The address still needs to be set with a `0x` prefix when the address text input focus is lost (`onBlur`). Also the app shouldn't rerender with the `setState` hook when if the value hasn't changed ([StackOverflow](https://stackoverflow.com/questions/59489959/set-state-with-same-value-using-hooks-will-cause-a-rerender)).
2. ✅ Perhaps it may be clearer to refer to an error in the code, I've removed the `AddressState`.
   1. If the question also refers to why the enums have no initializers: As the enum is not serialized nor deserialized, string enums are not necessary and automatically incrementing numeric enums are fine.
3. ✅ Fixed. Caused when using `isHex` as it not only checked `0x` prefix but also the rest of the characters. So if a non-hex character is used, the code in this PR then would prepend `0x`.
4. ✅ Fixed. TextInput component's "clear" button doesn't work with the keyboard open. I changed my `ScrollView` to `KeyboardAwareScrollView` which fixed it.

Some more changes:
- `isCompleteAddress` may not be needed later, but helps visualize the spinning icon in this PR, and is a reminder in the next PR when fetching the data from the contract. I'll keep it in this PR.
- Added `autoCorrect` and `spellCheck` to prevent entering typo errors.

src/tokens/TokenImport.tsx
```diff
@@ -11,60 +10,98 @@ import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
...
+    !address || address.startsWith('0x') ? address : `0x${address}`
+
+  const handleAddressBlur = () => {
+    const address = ensure0xPrefixOrEmpty(tokenAddress)
+    setTokenAddress(address)
+    validateAddress(address)
+  }
+
+  const handlePaste = (address: string) => {
+    address = ensure0xPrefixOrEmpty(address)
```
kathaypacific, line 78:
```suggestion
    const addressWith0xPrefix = ensure0xPrefixOrEmpty(address)
```
avoid mutations :)

src/tokens/TokenImport.test.tsx
```diff
@@ -47,24 +33,41 @@ describe('TokenImport', () => {
...
-        <TokenImportScreen {...mockScreenProps} />
-      </Provider>
-    )
-    const symbolInput = getByTestId('tokenSymbol')
-    expect(symbolInput.props.editable).toBeFalsy()
-    fireEvent.changeText(
-      getByPlaceholderText('tokenImport.input.tokenAddressPlaceholder'),
-      '0xef4229c8c3250C675F21BCefa42f58EfbfF6002a'
-    )
+  describe('renderErrorMessage when token address', () => {
```
kathaypacific, line 42:
```suggestion
  describe('error messages for token address', () => {
```

src/tokens/TokenImport.test.tsx
```diff
@@ -47,24 +33,41 @@ describe('TokenImport', () => {
...
+    it('is already supported', () => {
```
kathaypacific, line 43:
```suggestion
    it('should display the correct error when the token is already supported', () => {
```

src/tokens/TokenImport.test.tsx
```diff
@@ -47,24 +33,41 @@ describe('TokenImport', () => {
...
+      )
+
+      fireEvent.changeText(getByText('tokenImport.input.tokenAddress'), mockCusdAddress)
+      fireEvent(getByText('tokenImport.input.tokenAddress'), 'blur')
+
+      expect(getByText('tokenImport.error.alreadySupported')).toBeTruthy()
+      expect(getByText('tokenImport.importButton')).toBeDisabled()
+    })
+
+    it('is invalid', () => {
```
kathaypacific, line 58:
```suggestion
    it('should display the correct error when the token address is invalid', () => {
```

kathaypacific (approved):
nice, thanks!

PR #4577, opened 2023-12-01
refactor(qr): make QR Scanner logic more modular

jh2oman:
### Description

* renamed things from Barcode => QRCode
* removed all secure send specific logic from the QR components and instead created a callback `onQRCodeDetected` function on QRScanner to separate the business logic of secure send from QR code logic

### Test plan

Went through all of the QR flows manually and checked that they still worked


src/navigator/types.tsx
```diff
@@ -347,8 +348,7 @@ export type QRTabParamList = {
   [Screens.QRScanner]:
     | {
         scanIsForSecureSend?: true
```
jh2oman, line 350: this prop is still needed for *visual* differences with QR scanner for the secure send flow

src/qrcode/utils.ts
```diff
@@ -114,63 +120,78 @@ function* handleSecureSend(
...
+// Catch all handler for QR Codes
+// includes support for WalletConnect, hooks, and send flow (non-secure send)
+export function* handleQRCodeDefault({ qrCode }: HandleQRCodeDetectedAction) {
+  const walletConnectEnabled: boolean = yield* call(isWalletConnectEnabled, qrCode.data)
+
   // Regex matches any 40 hexadecimal characters prefixed with "0x" (case insensitive)
-  if (/^0x[a-f0-9]{40}$/gi.test(barcode.data)) {
-    barcode.data = `celo://wallet/pay?address=${barcode.data}`
+  if (/^0x[a-f0-9]{40}$/gi.test(qrCode.data)) {
+    qrCode.data = `celo://wallet/pay?address=${qrCode.data}`
```
cajubelt, line 130: if you moved this part down to just before `let qrData: UriData` (which is safe because the next if statements are mutually exclusive with this one) then it becomes the same as 11 lines of handleQRCodeSecureSend and you can create a shared helper for both

src/send/saga.ts
```diff
@@ -325,7 +291,16 @@ export function* watchSendPayment() {
   yield* takeLeading(Actions.SEND_PAYMENT, safely(sendPaymentSaga))
 }
 
+function* watchQrCodeDetections() {
```
cajubelt, line 294: nice, looks much simpler

src/navigator/types.tsx
cajubelt: could be nice to rename it like `showSecureSendStyling` (or something) to make that more explicit

cajubelt (approved):
had a few relatively minor ideas but nothing major. looks like a helpful refactor.

PR #4578, opened 2023-12-02
chore(deps): update resolve-kit, remove minimatch resolution

cajubelt:
### Description

minimatch vulnerability was 3.0.4 and below

no longer needed the resolution once resolve-kit was updated

### Test plan

ci

### Related issues

na

### Backwards compatibility

na


cajubelt:
this works with the latest code from resolve-kit but that needs to get published before it can be updated here

jeanregisser (changes requested):
Oops sorry, I saw it's duplicating many @celo libs, we should resolve this first.

PR #4583, opened 2023-12-04
feat: invite rewards card

MuckT:
### Description

Adds an invite rewards card.

| iOS | Android | iOS Large Font | 
| ----- | ----- | ----- |
|  |  |  |

### Test plan

- [x] Tested locally on iOS
- [x] Unit Tests added

### Related issues

- ACT-972

### Backwards compatibility

Yes


src/recipients/recipient.ts
```diff
@@ -252,7 +252,7 @@ function executeFuzzySearch(
   shouldSort?: boolean
 ): FuzzyRecipient[] {
   const parsedQuery = query.replace(/[()-\s/\\]/g, '')
-  if (parsedQuery === '' || parsedQuery.length < 2) {
+  if (parsedQuery === '') {
```
cajubelt: this looks like it could be valuable, but isn't related to the main purpose of the PR. can you split it out to keep the PR single-purpose?

src/recipients/RecipientPickerV2.tsx
```diff
@@ -29,6 +29,7 @@ function RecipientPicker({
     <View style={[styles.body, style]} testID={testID}>
       {title && <Text style={styles.title}>{title}</Text>}
       <FlatList
+        scrollEnabled={false}
```
cajubelt: is this related?

cajubelt: I saw it in your other PR too... which one does it actually belong with?

src/send/InviteRewardsCard.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { inviteRewardsTypeSelector } from 'src/send/selectors'
+import { InviteRewardsType } from 'src/send/types'
+import Colors from 'src/styles/colors'
+import { Spacing } from 'src/styles/styles'
+
+export default function InviteRewardsCard() {
+  const { t } = useTranslation()
+  const inviteRewardsType = useSelector(inviteRewardsTypeSelector)
+
+  // Default to NFT invite rewards banner
```
cajubelt, line 15: this doesn't seem safe. we add invite rewards features relatively often and it would be easy to forget to update this. how about logging an error and returning null if the type of invite rewards isn't nft or cusd? you can also add static validation with something like this:
```
const inviteRewardsToTranslationPrefix: Record<InviteRewardsType, string> = {
[InviteRewardsType.CUSD]: 'inviteRewardsBannerCUSD',
[InviteRewardsType.NFT]: 'inviteRewardsBanner',
}
const inviteRewardsTranslationPrefix = inviteRewardsToTranslationPrefix[inviteRewardsType]
const titleKey = `${inviteRewardsPrefix}.title`
const bodyKey = ...
```

cajubelt (commented):
pausing for now because the diff has exploded, appears to be out of sync with its base branch

src/recipients/RecipientPickerV2.tsx
MuckT: Should be in #4490.

src/recipients/recipient.ts
MuckT: Should be in #4490.

src/send/InviteRewardsCard.tsx
MuckT: Fixed in 2425b09f570a00993c0664900786b1623e61a8d4!

src/send/InviteRewardsCard.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { Spacing } from 'src/styles/styles'
+
+export default function InviteRewardsCard() {
+  const { t } = useTranslation()
+  const inviteRewardsType = useSelector(inviteRewardsTypeSelector)
+
+  const inviteRewardsToTranslationPrefix: Record<InviteRewardsType, string> = {
+    [InviteRewardsType.CUSD]: 'inviteRewardsBannerCUSD',
+    [InviteRewardsType.NFT]: 'inviteRewardsBanner',
+    [InviteRewardsType.NONE]: 'inviteRewardsBanner',
```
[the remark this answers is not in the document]
MuckT: Fixed in 1b354b05bb8828a3a61c92816fe7f04500b65af8. I left some of the display logic in `src/send/SendSelectRecipient.tsx` as it seems like it would be more performant to avoid attempting to render at all vs attempting to render the component and returning null when inviteRewardsActive is false.

Applied your recommend where if the prefix is null we don't display the card.

PR #4585, opened 2023-12-05
feat: add automatic fee currency selection to walletconnect viem flow

kathaypacific:
### Description

This PR:
1. create prepared transactions on receiving action request
2. use the prepared transactions to show a gas fee warning if the user does not have enough gas fees to cover the transaction. we can also use the prepared transaction to show the gas fees later.
3. on accepting the action request, the viem flow should use the prepared transaction directly

Other:
- since we need to get the fee currencies from the walletconnect saga, turn the useFeeCurrencies hook into a selector, replace its usages.

### Related issues

- Fixes RET-905


kathaypacific:
closing in favor of #4621

PR #4584, opened 2023-12-05
feat(send): show warning for unknown addresses

satish-ravi:
### Description

Mainly the UI, the condition for address will be updated in a followup once address lookup is ready (after #4586)

### Test plan

Unit tests, manually by searching for different types of recipients

https://github.com/valora-inc/wallet/assets/5062591/d8ac828b-bdeb-43cb-82a7-814c61c19c27

### Related issues

- Part of ACT-979

### Backwards compatibility

Yes

PR #4586, opened 2023-12-05
feat(send): Show Valora V icon for address-based recipients depending on phone verification

jophish:
### Description

Updates the behavior of the Valora V icon to show only for address-based recipients if the address is associated with a phone number verified with Valora. "Address-based recipients" here means either `RecipientType.Address` or `Recipient.Nomspace`, since previously, the Valora V icon behaviors for both were treated identically.

### Test plan

Unit and manual tested. See videos below:

#### Address has no associated phone number

https://github.com/valora-inc/wallet/assets/569401/6ff32d95-8230-4ba4-aaf0-78cffb0f6612

#### Address does have associated phone number

https://github.com/valora-inc/wallet/assets/569401/88afbedc-6395-40bd-be20-68beabd5cd94

### Related issues

- Fixes #[issue number here]


src/recipients/RecipientPickerV2.tsx
```diff
@@ -25,6 +26,8 @@ function RecipientPicker({
   isSelectedRecipientLoading,
   style,
 }: Props) {
+  const isRecipientSelected = (recipient: Recipient) =>
```
jophish, line 29: Because of re-renders, it's not guaranteed that we'll have object equality here. The default check is still object equality, but we also now check deep equality as a fallback.

src/send/useFetchRecipientVerificationStatus.ts
```diff
@@ -13,6 +17,8 @@ const useFetchRecipientVerificationStatus = () => {
   )
 
   const e164NumberToAddress = useSelector(e164NumberToAddressSelector)
+  const verifiedAddresses = useSelector(verifiedAddressesSelector)
+  const verifiedAddressesLoading = useSelector(verifiedAddressesLoadingSelector)
```
jophish, line 21: ;The reason why we need this extra piece of state is kind of subtle, and tripped me up for a bit. Basically, `getRecipientVerificationStatus` (in the `useEffect` hook below) is designed such that it assumes its parameters (`e164NumberToAddress`, `verifiedAddresses`) are not in-flight, i.e., not currently "loading". As such, it returns `RecipientVerificationStatus.VERIFIED` when the address is missing from `verifiedAddresses`, which _will always be the case_ while we're fetching an address we haven't seen before. Because of the `RecipientVerificationStatus.UNKNOWN` check below, we'll never change the status after the fetch finishes, and as soon as the verification status is set to verified, we enable the "Send" button.

What we actually want is to be able to keep the Send button disabled until the fetch is done, then _always_ enable it, but show a banner if the address is unverified, which is what this loading state allows for (only updating the verification status once the fetch is definitely done). @satish-ravi, I think in #4584 you should just be able to look at a combination of `recipientVerificationStatus` and `selectedRecipient.recipientType` and only show the banner if:

```ts
recipientVerificationStatus === RecipientVerificationStatus.UNVERIFIED && selectedRecipient.recipientType === RecipientType.Address
```

src/recipients/recipient.ts
```diff
@@ -174,26 +174,29 @@ export function getRecipientFromAddress(
 
 export function getRecipientVerificationStatus(
   recipient: Recipient,
-  e164NumberToAddress: E164NumberToAddressType
+  e164NumberToAddress: E164NumberToAddressType,
+  verifiedAddresses: string[]
 ): RecipientVerificationStatus {
-  if (recipientHasAddress(recipient)) {
```
jophish, line 179: this logic needed some rejiggering - if we just replaced this top check with e.g.

```
if (recipientHasAddress(recipient)) {
      return verifiedAddresses.includes(recipient.address)
        ? RecipientVerificationStatus.VERIFIED
        : RecipientVerificationStatus.UNVERIFIED
}
```

then we'd potentially return `UNVERIFIED` for phone recipients who happen to have an address. this new layout gives preference to phone numbers; e.g., if a recipient has both an address and a phone number, we do the phone number checks; we only do the address checks if the recipient _only_ has an address.

src/identity/contactMapping.ts
```diff
@@ -207,6 +211,34 @@ export function* fetchAddressesAndValidateSaga({
...
+      ValoraAnalytics.track(IdentityEvents.address_lookup_start)
+      const addressVerified = yield* call(fetchAddressVerification, address)
+      if (addressVerified) {
+        yield* put(updateVerifiedAddresses([...verifiedAddresses, address]))
+      }
+      ValoraAnalytics.track(IdentityEvents.address_lookup_complete)
+    }
+  } catch (err) {
+    const error = ensureError(err)
+    Logger.debug(
```
satish-ravi, line 229: should this be a higher log level? or could it be noisy w/ rate limiting stuff?

src/identity/contactMapping.ts
```diff
@@ -243,6 +275,43 @@ function* fetchWalletAddresses(e164Number: string) {
...
+    const walletAddress = yield* select(walletAddressSelector)
+    const signedMessage = yield* call(retrieveSignedMessage)
+
+    const addressVerificationQueryParams = new URLSearchParams({
+      address,
+      clientPlatform: Platform.OS,
+      clientVersion: DeviceInfo.getVersion(),
+    }).toString()
+
+    const response: Response = yield* call(
```
satish-ravi, line 289: not sure if we do this for other identity service requests, but should these have a timeout?

src/identity/contactMapping.ts
```diff
@@ -243,6 +275,43 @@ function* fetchWalletAddresses(e164Number: string) {
...
+    if (!response.ok) {
+      throw new Error(
+        `Failed to look up address verification: ${response.status} ${response.statusText}`
+      )
+    }
+
+    const { data }: { data: { addressVerified: boolean } } = yield* call([response, 'json'])
+    return data.addressVerified
+  } catch (error) {
+    Logger.debug(`${TAG}/fetchAddressVerification`, 'Unable to look up address', error)
```
satish-ravi, line 310: same log question as above, probably this one is more useful than the above one since this catch swallows the error and throws a generic message

src/recipients/RecipientItemV2.tsx
```diff
@@ -34,12 +34,17 @@ function RecipientItem({ recipient, onSelectRecipient, loading, selected }: Prop
...
+  const verifiedAddresses = useSelector(verifiedAddressesSelector)
 
   // TODO(ACT-980): avoid icon flash when a known valora contact is clicked
-  // TODO(ACT-950): show icon for address recipients
   const showValoraIcon = useMemo(() => {
     if (recipient.recipientType === RecipientType.PhoneNumber) {
       return recipient.e164PhoneNumber && !!e164NumberToAddress[recipient.e164PhoneNumber]
+    } else if (
+      recipient.recipientType === RecipientType.Address ||
+      recipient.recipientType === RecipientType.Nomspace
```
satish-ravi, line 45: we can probably drop these conditions? The only other recipient type is Merchant, which I am not even sure if its used anymore. I see 0 references to it.

src/identity/contactMapping.ts
```diff
@@ -207,6 +211,34 @@ export function* fetchAddressesAndValidateSaga({
   }
 }
 
+export function* fetchAddressVerificationSaga({ address }: FetchAddressVerificationAction) {
+  try {
+    const verifiedAddresses = yield* select(verifiedAddressesSelector)
+    if (verifiedAddresses.includes(address)) {
+      ValoraAnalytics.track(IdentityEvents.address_lookup_skip)
```
satish-ravi, line 218: would this ever be true? seems like the action is only dispatched if address isn't in the verifiedAddresses list

src/identity/contactMapping.test.ts
```diff
@@ -146,3 +154,58 @@ describe('Fetch Addresses Saga', () => {
...
+    it('handles errors gracefully', async () => {
+      mockFetch.mockReject()
+      await expectSaga(fetchAddressVerificationSaga, fetchAddressVerification(mockAccount))
+        .provide([
+          [select(verifiedAddressesSelector), []],
+          [select(walletAddressSelector), '0xxyz'],
+          [call(retrieveSignedMessage), 'some signed message'],
+        ])
+        .put(endFetchAddressVerification())
+        .run()
```
satish-ravi, line 208: should there be some assertion for failure? For example, if someone accidentally removes the mockFetch.mockReject() above the test would still pass?

src/identity/contactMapping.test.ts
```diff
@@ -146,3 +154,58 @@ describe('Fetch Addresses Saga', () => {
     })
   })
 })
+
+describe('Fetch Address Verification Saga', () => {
+  describe('central lookup', () => {
```
satish-ravi, line 159: nit: is this nesting required?

src/identity/reducer.ts
```diff
@@ -83,6 +83,10 @@ export interface State {
   importContactsProgress: ImportContactProgress
   // Contacts found during the matchmaking process
   secureSendPhoneNumberMapping: SecureSendPhoneNumberMapping
+  // List of addresses known to be associated with a Valora wallet
+  verifiedAddresses: string[]
+  // Whether or not verified addresses are currently being fetched
+  verifiedAddressesLoading: boolean
```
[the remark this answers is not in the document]
jophish: I _think_ this is probably something we can punt on for now. I'll make a note of this for the testing ticket I'm currently working on; we can either spin this into a ticket or try to get a fix out before the release, but I think the risk is very low as-is. Additionally, I think it's rather hard (if not impossible? not sure) to get two unverified addresses to show up next to eachother in the search results.

src/identity/contactMapping.ts
```diff
@@ -207,6 +211,34 @@ export function* fetchAddressesAndValidateSaga({
...
+    } else {
+      ValoraAnalytics.track(IdentityEvents.address_lookup_start)
+      const addressVerified = yield* call(fetchAddressVerification, address)
+      if (addressVerified) {
+        yield* put(updateVerifiedAddresses([...verifiedAddresses, address]))
+      }
```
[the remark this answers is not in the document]
jophish: Yeah, that's correct. I can leave a note here.

src/identity/contactMapping.ts
jophish: I'll turn this one into a warn, and leave the other as debug.

src/identity/contactMapping.ts
jophish: Not sure; we don't appear to do that for other calls in this file.

src/identity/contactMapping.ts
jophish: addresses in other comment

src/recipients/RecipientItemV2.tsx
jophish: Yeah, agreed -- dropped this if statement.

src/identity/contactMapping.ts
jophish: No, at least, not with how this is being called now. Not strictly necessary to have this, but it's a simple check and a bit more defensive.

src/identity/contactMapping.ts
satish-ravi: I think it's good practice to have timeouts, otherwise the request can hang for a long time. Maybe we can create a follow up ticket / PR to add timeouts?

src/identity/contactMapping.ts
satish-ravi: Should the analytics event here then if it's not ever going to be called in the current usage of this saga? Can we drop it or move it to the SelectRecipient screen?

jophish: I think we're probably fine to drop it, I'll get rid of it.

satish-ravi (approved):

src/identity/reducer.ts
```diff
@@ -257,6 +264,22 @@ export const reducer = (
         e164NumberToSalt: state.e164NumberToSalt,
         secureSendPhoneNumberMapping: state.secureSendPhoneNumberMapping,
       }
+    case Actions.FETCH_ADDRESS_VERIFICATION:
+      return {
+        ...state,
+        addressToVerificationStatus: {
+          ...state.addressToVerificationStatus,
+          [action.address]: !state.addressToVerificationStatus[action.address] ? undefined : true,
```
satish-ravi, line 272: nit: 
```suggestion
          [action.address]: state.addressToVerificationStatus[action.address] || undefined
```
Also maybe leave a comment on why its set to undefined if `false`?

src/recipients/recipient.ts
```diff
@@ -174,26 +175,34 @@ export function getRecipientFromAddress(
...
-
   if (!recipientHasNumber(recipient)) {
-    return RecipientVerificationStatus.UNKNOWN
-  }
+    if (recipientHasAddress(recipient) && recipient.address in addressToVerificationStatus) {
+      switch (addressToVerificationStatus[recipient.address]) {
+        case true:
+          return RecipientVerificationStatus.VERIFIED
+        case false:
+          return RecipientVerificationStatus.UNVERIFIED
```
satish-ravi, line 187: is this backwards compatible? seems like we previously returned VERIFIED for all addresses.

jophish: yeah, if we're in the old flow, we set the status to verified in the verification hook one level up and don't call this function.

PR #4587, opened 2023-12-05
feat(swap): skip swap confirmation screen

jeanregisser:
### Description

This implements the following changes in the swap flow:
- remain on the SwapScreen after tapping "Confirm Swap", showing a loader on the button
- if the transaction(s) are successfully signed and submitted to the network, it navigates to the home screen without waiting for them to be confirmed. The status (pending, success, failed) can be seen from the home screen.
- if there's an error while signing/submitting, an inline error message is shown
  - the error is dismissed if the user changes the input, or submits again

The SwapScreen and state were updated to be able to handle sending different swaps in parallel. Each swap gets a swapId which is a `UUID`, so the screen only displays relevant information and doesn't mix it with other swaps possibly happening in parallel.

Note: I'll delete the `SwapExecuteScreen` in a separate PR.

### Test plan

- Updated tests
- Manually tested:

https://github.com/valora-inc/wallet/assets/57791/532fb9e2-3b43-4972-8460-5b3ab1c1715f

- Inline error (simulated by throwing while submitting):

https://github.com/valora-inc/wallet/assets/57791/c861a87a-97ca-4c58-a9b6-02d11fa4e6d2

### Related issues

- Fixes RET-910

### Backwards compatibility

Yes


src/swap/SwapScreen.test.tsx
```diff
@@ -1026,6 +1028,118 @@ describe('SwapScreen', () => {
...
+    expect(getByText('swapScreen.confirmSwapFailedWarning.body')).toBeTruthy()
+    // NOT disabled, so users can retry
+    expect(getByText('swapScreen.confirmSwap')).not.toBeDisabled()
+
+    // Now change some input, and the warning should disappear
+    fireEvent.changeText(within(swapFromContainer).getByTestId('SwapAmountInput/Input'), '2')
+
+    expect(queryByText('swapScreen.confirmSwapFailedWarning.title')).toBeFalsy()
+    expect(queryByText('swapScreen.confirmSwapFailedWarning.body')).toBeFalsy()
+  })
```
kathaypacific, line 1141: great tests!

kathaypacific (approved):
love this!! ✨

kathaypacific:
one thought i had (not related to these changes) is now we rely on the pending transaction as visual feedback that something has happened following the swap. if the network is too fast, and the transaction is already confirmed by the time the navigation finishes, it may not be obvious that something happened if the user's homefeed is full of swaps. kind of edge casey but made me wonder if we should add an animation for new homefeed items (e.g. new items have a faint highlighted background that fades for X seconds). maybe something we can think about in the future

jeanregisser:
> one thought i had (not related to these changes) is now we rely on the pending transaction as visual feedback that something has happened following the swap. if the network is too fast, and the transaction is already confirmed by the time the navigation finishes, it may not be obvious that something happened if the user's homefeed is full of swaps. kind of edge casey but made me wonder if we should add an animation for new homefeed items (e.g. new items have a faint highlighted background that fades for X seconds). maybe something we can think about in the future

Oh yeah that would be great!

PR #4588, opened 2023-12-05
fix: persist taps in recipient flatlist

MuckT:
### Related issues

- Fixes #[issue number here]

### DIFF
--- a/src/recipients/RecipientPickerV2.tsx
+++ b/src/recipients/RecipientPickerV2.tsx
@@ -29,6 +29,7 @@ function RecipientPicker({
     <View style={[styles.body, style]} testID={testID}>
       {title && <Text style={styles.title}>{title}</Text>}
       <FlatList
+        keyboardShouldPersistTaps={'always'}
         scrollEnabled={false}
         data={recipients}
         renderItem={({ item }) => (


src/send/hooks.ts
```diff
@@ -49,6 +50,7 @@ export function useMergedSearchRecipients(onSearch: (searchQuery: string) => voi
 
   const throttledSearch = throttle((searchInput: string) => {
     onSearch(searchInput)
+    searchQueryRef.current = searchInput
```
cajubelt, line 53: seems like searchQueryRef and searchQuery state are very similar. did you try combining them?

cajubelt (commented):
would be great to get some kind of regression test coverage, maybe by mocking the keyboard dismiss function

cajubelt:
also the description seems to only describe one of the bugs fixed here-- doesn't it also allow you to dismiss the keyboard by tapping a recipient?

src/recipients/RecipientItemV2.tsx
```diff
@@ -30,6 +30,7 @@ function RecipientItem({ recipient, onSelectRecipient, loading, selected }: Prop
   const { t } = useTranslation()
 
   const onPress = () => {
+    Keyboard.dismiss()
```
satish-ravi, line 33: this seems a bit out of place in the component. Is this something can be done in the onSelectRecipient callback itself?

src/send/hooks.ts
satish-ravi: We can drop the ref and do something like `if (searchQuery === searchInput) return` at the start of this function right?

MuckT: Fixed in 091f1fae818fdf92b744fbd0e16aa4effba424ca using @satish-ravi suggestion; seems a lot cleaner and more contained.

src/recipients/RecipientItemV2.tsx
MuckT: Open to moving it, but I also can't think of a case we wouldn't want to dismiss the keyboard on a recipient press. This way we don't have a the dismiss in both `setSelectedRecipientWrapper` and `onSelectRecentRecipient`.

cajubelt (approved):

PR #4595, opened 2023-12-06
feat(token-import): Validate ERC-20 token contract

denvalora:
### Related issues

- Fixes #[issue number here]


src/tokens/TokenImport.tsx
```diff
@@ -63,13 +68,41 @@ export default function TokenImportScreen(_: Props) {
...
+  const validateCheck = async (address: Address) => {
+    setAddressState(AddressState.FetchingContract)
+
+    const contract = getContract({
+      abi: erc20.abi,
+      address,
+      publicClient: publicClient[networkIdToNetwork[networkId]],
+    })
+
+    const state = await Promise.race<AddressState>([
```
kathaypacific, line 87: what do you think about splitting up the form validation and address fetch? it seems like they are different topics - one is a static check based only on user input, and one is asynchronous which could fail for reasons unrelated to the user.

i would propose that onBlur / on paste -> address validation happens, if there are no errors, then proceed to call the async function to fetch the details.

for async actions, we've generally used the `react-async-hook` lib which helps us to handle edge cases like the screen unmounting before all async actions have finished. this helps to avoid unresolved promises. i think we should use that here too

src/tokens/TokenImport.tsx
```diff
@@ -63,13 +68,41 @@ export default function TokenImportScreen(_: Props) {
...
+      new Promise((resolve) => {
+        Promise.all([
+          contract.read.symbol(),
+          contract.read.decimals(),
+          contract.read.name(),
+          contract.read.totalSupply(),
```
kathaypacific, line 93: do we need total supply? we'll eventually need to also query the user's balance, so could we add a balance check here?

src/tokens/TokenImport.tsx
```diff
@@ -63,13 +68,41 @@ export default function TokenImportScreen(_: Props) {
...
+    const contract = getContract({
+      abi: erc20.abi,
+      address,
+      publicClient: publicClient[networkIdToNetwork[networkId]],
+    })
+
+    const state = await Promise.race<AddressState>([
+      new Promise((resolve) => {
+        Promise.all([
+          contract.read.symbol(),
```
kathaypacific, line 90: in the viem example docs i see the syntax
```
publicClient.readContract({
    abi: erc20.abi,
      address,
      functionName: 'symbol',
 }),
```
is there any advantage in your approach over theirs? it seems like what is here we fetch the contract and then read it, but `readContract` does this in one step? i'm not sure, just curious

denvalora: Contract instance (`getContract(..)` returned value) abstracts calls to `readContract` as the instance knows the ABI and address. If we need to read multiple public variables (or functions with view or pure state mutability), the code will be more concise with the contract instance. Otherwise they're the same.

[Documentation](https://viem.sh/docs/contract/getContract.html#:~:text=Contract%20Instances%20can%20make%20it%20easier%20to%20work%20with%20contracts%20if%20you%20don%27t%20want%20to%20pass%20the%20abi%20and%20address%20properties)

PR #4601, opened 2023-12-07
feat(swap): show network switch warning

jeanregisser:
### Description

Title says it all.

### Test plan

- Added new test
- Manually tested:

https://github.com/valora-inc/wallet/assets/57791/9debd736-18bc-4df7-b96e-3d2c962d8536

### Related issues

- Fixes RET-934

### Backwards compatibility

Yes


src/analytics/Properties.tsx
```diff
@@ -1326,6 +1326,13 @@ interface SwapEventsProperties {
     tokenSymbol: string
     tokenId: string
     tokenNetworkId: string
+    fromTokenSymbol: string | undefined
+    fromTokenId: string | undefined
+    fromTokenNetworkId: string | undefined
+    toTokenSymbol: string | undefined
+    toTokenId: string | undefined
+    toTokenNetworkId: string | undefined
+    switchedNetworkId: boolean
```
jeanregisser, line 1335: Adding from and to here though it will duplicate some of the existing info, but will give easier access to what cross chain pairs users are interested in.

src/swap/SwapScreen.tsx
```diff
@@ -519,7 +550,10 @@ export function SwapScreen({ route }: Props) {
     navigate(Screens.WebViewScreen, { uri: TRANSACTION_FEES_LEARN_MORE })
   }
 
-  const showMaxSwapAmountWarning = !confirmSwapFailed && shouldShowMaxSwapAmountWarning
+  const showSwitchedToNetworkWarning = !!switchedToNetworkId
+  const switchedToNetworkName = switchedToNetworkId && NETWORK_NAMES[switchedToNetworkId]
+  const showMaxSwapAmountWarning =
+    !confirmSwapFailed && !showSwitchedToNetworkWarning && shouldShowMaxSwapAmountWarning
```
jeanregisser, line 556: It's now becoming quite hard to understand what warning has precedence over the others.
I'll want to refactor this in another PR.

src/swap/SwapScreen.tsx
```diff
@@ -429,28 +430,58 @@ export function SwapScreen({ route }: Props) {
...
-      setToToken(selectedToken)
+      newToToken = selectedToken
+      setToToken(newToToken)
+      switchedToNetworkId =
+        fromToken && fromToken.networkId !== newToToken.networkId ? newToToken.networkId : null
+      setSwitchedToNetworkId(switchedToNetworkId)
+      if (switchedToNetworkId) {
+        // reset the fromToken if the user is switching networks
+        setFromToken(undefined)
+      }
```
jeanregisser, line 463: It may look like we have duplicated similar logic, I tried to make it more concise but then I thought it was making it complicated to understand.

But let me know if you have a better idea.

src/swap/SwapScreen.tsx
```diff
@@ -429,28 +430,61 @@ export function SwapScreen({ route }: Props) {
...
     ) {
-      setFromToken(toToken)
-      setToToken(fromToken)
+      newFromToken = toToken
+      newToToken = fromToken
     } else if (selectingToken === Field.FROM) {
-      setFromToken(selectedToken)
+      newFromToken = selectedToken
+      newSwitchedToNetworkId =
+        toToken && toToken.networkId !== newFromToken.networkId ? newFromToken.networkId : null
```
kathaypacific, line 452: nit: i thought networkId is a required type, any reason why we need to have this ternary?

kathaypacific (approved):
great!! love this!!

src/swap/SwapScreen.tsx
jeanregisser: Yes it's because `newSwitchedToNetworkId` is used for the state to display the warning and we won't always want to display it.

PR #4602, opened 2023-12-07
chore: use block time for confirmed transaction time

kathaypacific:
### Description

This PR updates the timestamp of a confirmed transaction (confirmed via our watcher) to use the block time rather than Date.now().

This is particularly important on Celo right now where approvals do not get overwritten by a backend data source. Without this change, approvals always are displayed as more recent than the corresponding transaction (since our transaction feed sources return the block timestamp of transactions, and the pending transaction "watcher" will always lag behind the block time of when the transaction was confirmed).

### Test plan

Before - notice how the approval transactions are more recent than the corresponding swap:
Simulator Screenshot - iPhone 14 Pro - 2023-12-14 at 12 33 25

After:

https://github.com/valora-inc/wallet/assets/20150449/e5e8ec1b-2015-4651-a497-c39ce077627f

### Related issues

Related to RET-942

### Backwards compatibility

Y


bakoushin (approved):
LGTM! ⏱️

dievazqu (approved):

PR #4606, opened 2023-12-08
chore: `isCoreToken` -> `isFeeCurrency`

bakoushin:
### Description

In the future we may have tokens which are accepted as a fee currency, but may not support [transferWithComment](https://celoscan.io/address/0x765de816845861e75a25fca122bb6898b8b1282a#writeProxyContract#F15) method in their smart contract (e.g. [GoodDollar](https://celoscan.io/token/0x62b8b11039fcfe5ab0c56e502b1c372a3d2a9c7a?a=0xb6d4ab615915f2570f879f7aeb986e204e1e694a#writeProxyContract)). 

Although currently it's not the case: all accepted fee currency tokens can transfer with comment, we want to prepare for it making the property set more granular.

To achieve that we decouple `isCoreToken` property into two separate ones:
* `isFeeCurrency`
* `canTransferWithComment`

### Test plan

* Tested manually on mainnet and alfajores (home screen, send, swap, assets screen dapps screen)
* Updated unit tests

##### Note about mock objects:

If the mocked `tokenBalance` object contains extra properties probably not strictly need for the test, like `imageUrl`, both new properties are added in place of `isCoreToken` for sake of completeness.

If the mocked object apparently contains only limited set of properties, only `isFeeCurrency` was added in place `isCoreToken` to not clog the object.

### Related issues

Related to RET-899

`getTokensInfo` endpoint update: https://github.com/valora-inc/address-metadata/pull/306

### Backwards compatibility

Y


src/tokens/selectors.ts
```diff
@@ -187,7 +187,7 @@ export const tokensByUsdBalanceSelector = createSelector(
  * @deprecated
  */
 export const coreTokensSelector = createSelector(tokensByUsdBalanceSelector, (tokens) => {
```
bakoushin, line 189: I didn't bother to rename this selector to `feeCurrencyTokensSelector` according to the change, since it's deprecated

src/viem/saga.test.ts
```diff
@@ -129,7 +129,7 @@ describe('sendPayment', () => {
     imageUrl: '',
     balance: '10',
     priceUsd: '1',
-    isCoreToken: true,
+    isNative: true,
```
bakoushin, line 132: this is ETH, it seems right to have `isNative` here.

kathaypacific (approved):
🚀✨

PR #4607, opened 2023-12-08
chore: update hooks to call the v2/getShortcuts endpoint

silasbw:
### Description

https://github.com/valora-inc/hooks/commit/4b6d61f76f101a6d7504bc2c1d99d885638f4016 facilitates more types of shortcuts. We expect at least one to be contributed soon.  The change in this PR will cause that shortcut to work in future versions of Valora.


jeanregisser (approved):
👍 

We'll need to update the join logic too. Right now if there's no positions referencing it, the shortcut won't be displayed.

PR #4609, opened 2023-12-08
feat(qr-code): Update copy on qr-code screen

jh2oman:
### Description

This change in copy actually has much larger implications, mainly that we no longer need exchanges on the qr screen so the ExchangeQR screen can be removed (this probably could have been done beforehand but definitely should be done now)

https://www.figma.com/file/RsO1zYTuQZbClu9zjxOKe5/Celo-Audit%2FQR?type=design&node-id=123-51760&mode=design&t=miQTWy5n5ZRoR74g-0

Screenshot 2023-12-08 at 12 24 58 PM

### Test plan

Updated unit tests, compared designs

### Related issues

https://linear.app/valora/issue/ACT-999/update-my-code-qr-screen-to-show-ethereum-as-supported-network


src/components/InLineNotification.tsx
```diff
@@ -63,7 +64,7 @@ export function InLineNotification({
         </View>
         <View style={styles.contentContainer}>
           {title && <Text style={styles.titleText}>{title}</Text>}
-          <Text style={[styles.bodyText]}>{description}</Text>
+          {<Text style={[styles.bodyText]}>{description}</Text>}
```
satish-ravi, line 67: Should this even be rendered if it's null/undefined now that it's optional?

src/components/InLineNotification.tsx
```diff
@@ -127,6 +128,10 @@ const severityColors: Record<Severity, CustomColors> = {
     primary: Colors.errorDark,
     secondary: Colors.errorLight,
   },
+  [Severity.SoftInformational]: {
+    primary: Colors.black,
+    secondary: Colors.gray1,
+  },
```
satish-ravi, line 134: Is this going to be part of the design system and used in multiple places? If so, should we have design update the style guide? I see this component [here](https://www.figma.com/file/erFfzHvSTm5g1sjK6jWyEH/Working-Design-System-Components?node-id=493%3A11&mode=dev) and [here](https://www.figma.com/file/l6cnlCx2Ht2YSgnbBdzZph/App-Style-Guide?node-id=0%3A1&mode=dev). Not sure which one is the style guide we should be using

src/qrcode/QRCode.tsx
```diff
@@ -1,60 +1,65 @@
...
+    ).showBalances
+    const networks = supportedNetworkIds.map((id: NetworkId) => {
+      const network = networkIdToNetwork[id]
+      return network.charAt(0).toUpperCase() + network.slice(1)
+    })
+    return networks.join(', ')
   }
 
+  const description = () => (
+    <Text style={styles.description}>
```
satish-ravi, line 54: is the text wrapper needed here? seems like the text style is duplicated here and in the InLineNotification component

src/qrcode/QRCode.tsx
```diff
@@ -1,60 +1,65 @@
...
-
-  const onPressExchange = (exchange: ExternalExchangeProvider) => {
-    props.onPressExchange?.(exchange)
+  const getSupportedNetworks = () => {
+    const supportedNetworkIds = getDynamicConfigParams(
+      DynamicConfigs[StatsigDynamicConfigs.MULTI_CHAIN_FEATURES]
+    ).showBalances
+    const networks = supportedNetworkIds.map((id: NetworkId) => {
+      const network = networkIdToNetwork[id]
+      return network.charAt(0).toUpperCase() + network.slice(1)
```
satish-ravi, line 48: this seems brittle, should we just use this instead? https://github.com/valora-inc/wallet/blob/a2c6ed00a8de3deadeeb9c145f4531b15a7e9a73/src/shared/conts.ts#L3-L7

src/components/InLineNotification.tsx
```diff
@@ -9,12 +9,13 @@ export enum Severity {
...
   Warning,
   Error,
+  SoftInformational,
 }
 
 interface Props {
   severity: Severity
   title?: string | null
-  description: string
+  description?: string | JSX.Element | null
```
satish-ravi, line 18: Any reason to make this optional? Seems like the current usage you're adding should always pass a JSX element. Also is there a way to limit the type of JSX.Element? The typing seems too broad. What if a <View> for example is passed to this?

src/fiatExchanges/SelectProvider.tsx
```diff
@@ -453,7 +453,9 @@ function ExchangesSection({
       ...analyticsData,
     })
     if (exchangesLink === SelectProviderExchangesLink.ExchangeQRScreen) {
-      navigate(Screens.ExchangeQR, { flow, exchanges })
+      navigate(Screens.QRNavigator, {
+        screen: Screens.QRCode,
+      })
```
satish-ravi, line 458: Is this change ok'd by product? I didn't see it in the linear issue, but seems like the title of the screen changes from Deposit Crypto to the My Code / Scan tab switcher.

src/qrcode/QRCode.tsx
jh2oman: ah, i missed this!

src/fiatExchanges/SelectProvider.tsx
jh2oman: I'll clarify

src/components/InLineNotification.tsx
jh2oman: I'll clarify, thanks i was unaware of those style guides

MuckT:
@jh2oman ended up not being too bad handling comma's & and based on the number of items as seen [here](https://github.com/valora-inc/wallet/pull/4626/files#diff-f6bcbafa1793bbf55e8709e023e4a217a0e0c5ebb275952e4776714a798cdb4eR195-R205).

PR #4618, opened 2023-12-11
feat(token-import): Validate ERC-20 token contract

denvalora:
### Description

Fetches details for an ERC-20 token, logs those, shows error on network timeout or contract not returning values.

#### Happy path
Try to import `0x7d91E51C8F218f7140188A155f5C75388630B6a8` token (DAI) in Alfajores.

#### Edge cases
These are covered in the tests, but can manually be checked for outgoing analytics events:

- Paste button: ensure only one `import_token_error` event is sent
  - when screen is open, tapping paste button
     - `0x874069fa1eb16d44d622f2e0ca25eea172369bc1`
     - `874069fa1eb16d44d622f2e0ca25eea172369bc1`
  - tapping first on text input and then paste button
     - `0x874069fa1eb16d44d622f2e0ca25eea172369bc1`
     - `874069fa1eb16d44d622f2e0ca25eea172369bc1`

- Address validation:
   - `0xa` - invalid (too short), no analytics event is logged
   - `0x874069fa1eb16d44d622f2e0ca25eea172369bc1` - already supported (cUSD)
   - … - currently not testing for already imported tokens

- Contract validation:
   - ~`0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee` - address which is not a contract~
   - `0x9563a59C15842a6f322B10f69d1dD88b41f2E97B` - contract which is not ERC-20
   - For timeout, temporarily change HTTP transport in `src/viem/index.ts` and try to import the token `0x7d91E51C8F218f7140188A155f5C75388630B6a8`. It should show network timeout error.
      ```ts
        [Network.Celo]: http(undefined, {
          timeout: 1,
        }),
      ```

### Related issues

- closed draft with resolved comments: https://github.com/valora-inc/wallet/pull/4595
- finishes https://linear.app/valora/issue/RET-892/lookup-contract-info

### Backwards compatibility

Compatible.


src/tokens/TokenImport.tsx
```diff
@@ -27,57 +30,149 @@ import variables from 'src/styles/variables'
...
+import { publicClient } from 'src/viem'
+import networkConfig, { networkIdToNetwork } from 'src/web3/networkConfig'
+import { walletAddressSelector } from 'src/web3/selectors'
+import { Address, formatUnits, getContract, isAddress } from 'viem'
+
+const TAG = 'tokens/TokenImport'
 
 type Props = NativeStackScreenProps<StackParamList, Screens.TokenImport>
 
+const FETCH_TOKEN_DETAILS_TIMEOUT_MS = 5_000
```
kathaypacific, line 43: i've heard that RPC / general network requests are significantly slower in some regions (some folks mentioned this after a research trip to Nigeria) - i think 5 seconds may be too short, could we increase this? maybe can use the existing `FETCH_TIMEOUT_DURATION` instead

src/tokens/TokenImport.tsx
```diff
@@ -27,57 +30,149 @@ import variables from 'src/styles/variables'
...
+        // should never happen
+        throw new Error('No wallet address found when fetching token details')
+      }
+
+      const details = fetchTokenDetails(address, walletAddress)
+      const timeout = new Promise<never>((_, reject) =>
+        setTimeout(reject, FETCH_TOKEN_DETAILS_TIMEOUT_MS, new Error('Timeout'))
+      )
+
+      return Promise.race([timeout, details])
```
kathaypacific, line 121: promise.race doesn't cancel the in flight promises...i couldn't find an easy way to specify a timeout in the viem requests to ensure no unresolved promises when the timeout is hit :/ so i'm just noting this thought here without having a good suggestion of how to handle this here. one strawperson proposal i have is to implement no timeout (since i think there is a 60s default timeout with viem for RPC calls) and we can always measure the performance of this action with Sentry as a follow up if we are concerned. since this is not a routine user action, it seems acceptable if the request takes a long time (in comparison to a user with a very slow / bad connection that always hits the timeout can never import a token)

src/tokens/TokenImport.tsx
```diff
@@ -112,10 +207,14 @@ export default function TokenImportScreen(_: Props) {
           <TextInputGroup
             label={t('tokenImport.input.tokenAddress')}
             value={tokenAddress}
-            onChangeText={setTokenAddress}
+            onChangeText={(address) => {
+              setTokenAddress(address)
+              setError(null)
+              setTokenSymbol('')
+              validateContract.reset()
```
kathaypacific, line 214: maybe instead of calling .reset on every character change, we could make the text input disabled if the async request is loading?

kathaypacific (commented):
this is looking great! ✨

src/tokens/TokenImport.tsx
bakoushin: `+1` to just rely on Viem API doing it's thing. they have a sophisticated [timeout handling](https://github.com/wevm/viem/blob/e74c9ab92a8b0635b7f99512d5276527421a2bf4/src/utils/rpc.ts#L92).

src/tokens/TokenImport.tsx
```diff
@@ -27,57 +30,149 @@ import variables from 'src/styles/variables'
...
+      abi: erc20.abi,
+      address: tokenAddress,
+      publicClient: publicClient[networkIdToNetwork[networkId]],
+    })
+
+    const [symbol, decimals, name, balance] = await Promise.all([
+      contract.read.symbol(),
+      contract.read.decimals(),
+      contract.read.name(),
+      contract.read.balanceOf([walletAddress]),
```
dievazqu, line 103: Do we need to fetch the balance?

src/tokens/TokenImport.tsx
```diff
@@ -27,57 +30,149 @@ import variables from 'src/styles/variables'
...
-  const handleAddressBlur = () => {
+  const handleAddressBlur = async () => {
+    if (validateContract.status !== 'not-requested') {
+      return
+    }
     const address = ensure0xPrefixOrEmpty(tokenAddress)
     setTokenAddress(address)
-    validateAddress(address)
+    if (validateAddress(address)) {
+      await validateContract.execute(address).catch(() => undefined)
```
dievazqu, line 165: Hmm, I'm not sure but the `.catch` looks weird to me. I guess we are handling the error directly in the `onError`. But if we don't care about handling the error here, can we remove the `.catch` block?

dievazqu (commented):
Looking good!

src/tokens/TokenImport.tsx
kathaypacific: ah, i had the same thought. i came across something similar when i was working on revoke phone number https://github.com/valora-inc/wallet/blob/263144a9cf2f126bd448aa3867f3d5ec203461d1/src/RevokePhoneNumber.tsx#L61-L67 - i am guessing that this catch block is serving the same purpose. perhaps it is better to be explicit and add a comment for the catch block

src/tokens/TokenImport.tsx
denvalora: Not for this screen, however, once imported we'll need to show the balance.

If it fails to read balance here, then we know it's not an ERC-20 token and avoid importing it.

src/tokens/TokenImport.tsx
denvalora: Thanks, initially I could not find Viem's timeout settings. The default HTTP transport uses a timeout of 10s.

I'll remove the Promise.race and the custom timeout function.

src/tokens/TokenImport.tsx
denvalora: I've changed to using Viem's timeout. As we use the default HTTP transport, it means 10 second timeout. We can change that later in the future in `src/viem/index.ts`.

src/tokens/TokenImport.tsx
```diff
@@ -27,57 +30,149 @@ import variables from 'src/styles/variables'
...
 
-  const handleAddressFocus = () => {
-    setIsCompleteAddress(false)
     setError(null)
+    return true
+  }
+
+  const fetchTokenDetails = async (
+    tokenAddress: Address,
+    walletAddress: Address
```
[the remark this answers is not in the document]
denvalora: Ideally I would like to move it to some other place so it could be hopefully reused.

There's `fetchTokenBalancesForAddress` in `src/tokens/saga.ts` which currently fetches details via GraphQL. I was thinking of leaving `fetchTokenDetails` in that file, and just using it in `TokenImport`. What do you think?

src/tokens/TokenImport.tsx
```diff
@@ -27,57 +30,149 @@ import variables from 'src/styles/variables'
...
+    if (!isAddress(address)) {
       setError(t('tokenImport.error.invalidToken'))
+      return false
+    }
+
+    // TODO(RET-891): if already imported, set state as AddressState.AlreadyImported
+    const tokenId = getTokenId(networkId, address.toLowerCase())
+    if (supportedTokens[tokenId]) {
+      setError(t('tokenImport.error.alreadySupported'))
+      ValoraAnalytics.track(AssetsEvents.import_token_already_supported, {
```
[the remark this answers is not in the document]
denvalora: Added `NotContractError`, which is thrown if the bytecode is empty (meaning the address is not a smart contract).

denvalora:
As this is for the MVP, I have removed custom error messages - https://github.com/valora-inc/wallet/pull/4618/commits/146348c86ea6cf517c59b3e0aa7ffbf8bd071faa. I've kept the details in the analytics, however.

src/tokens/TokenImport.test.tsx
```diff
@@ -1,14 +1,50 @@
...
 
 const mockScreenProps = getMockStackScreenProps(Screens.TokenImport)
 
 describe('TokenImport', () => {
+  const mockBytecode = jest.fn()
+  const mockSymbol = jest.fn()
+
+  beforeEach(() => {
+    jest.resetAllMocks()
+    const client = publicClient[networkIdToNetwork[networkConfig.defaultNetworkId]]
```
kathaypacific, line 35:
```suggestion
    const client = publicClient[Network.Celo]
```
maybe more straightforward?

src/tokens/TokenImport.tsx
```diff
@@ -27,58 +30,175 @@ import variables from 'src/styles/variables'
...
+      }
+
+      setError(t('tokenImport.error.invalidToken'))
+      ValoraAnalytics.track(AssetsEvents.import_token_error, {
+        networkId,
+        tokenId,
+        tokenAddress,
+        error: Errors.NotERC20,
+      })
+    },
```
kathaypacific, line 175: nit: this error block could be simplified by having something (function / ternary / code block) that decides on the type of error, and then set the error + dispatch the analytics with the error type? it's easier to quickly read the intention of the function that way, currently it seems like depending on the type of error we may be doing custom actions

src/tokens/TokenImport.tsx
kathaypacific: i think i understand, seems reasonable to leave it like this. are we allowing for the scenario that a user successfully imports a token and we allow them to stay on this screen to import more? i thought that we'd navigate away from the screen on successful import (i see this in https://linear.app/valora/issue/RET-893/import-tokens-screen) ...with that assumption, i was thinking that we might not need the .reset at all

src/tokens/TokenImport.tsx
```diff
@@ -27,58 +30,175 @@ import variables from 'src/styles/variables'
...
+
+    const client = publicClient[networkIdToNetwork[networkId]]
+    const contract = getContract({
+      abi: erc20.abi,
+      address: tokenAddress,
+      publicClient: client,
+    })
+
+    const contractCode = await client.getBytecode({ address: tokenAddress })
+    if (!contractCode) throw new NotContractError('Contract code is empty')
```
kathaypacific, line 117: this is interesting, but do we need to capture this specific error? if the contract is empty, the fetches for symbol and decimals etc would also fail right? i guess i'm wondering what we would learn from seeing that users are importing tokens that have empty contracts

src/tokens/TokenImport.test.tsx
```diff
@@ -24,6 +60,28 @@ describe('TokenImport', () => {
...
+    const tokenSymbolInput = getByTestId('tokenSymbol')
+    const importButton = getByText('tokenImport.importButton')
+
+    fireEvent.changeText(tokenAddressInput, mockPoofAddress)
+    expect(tokenSymbolInput).toBeDisabled()
+    expect(importButton).toBeDisabled()
+    fireEvent(tokenAddressInput, 'blur')
+
+    await waitFor(() => {
+      expect(tokenSymbolInput.props.value).toBe('ABC')
```
kathaypacific, line 80: nit: it'd be nice to have only 1 thing inside the waitFor https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#having-multiple-assertions-in-a-single-waitfor-callback

src/tokens/TokenImport.tsx
denvalora: Hmm, perhaps I misunderstood the earlier comment - https://github.com/valora-inc/wallet/pull/4618#discussion_r1422094981 - logging error if not a valid contract - I thought it was about checking if it's a contract and not an ERC-20.

Regarding question: Yes, exactly. We don't really need it. When we check for `symbol`, if that fails, then that's enough for us to check that it implements some of the ERC-20 interface.

I'll remove this bytecode check then and leave only the `symbol()` and the other functions fetching details from ERC-20 contract.

src/tokens/TokenImport.tsx
kathaypacific: > It won't allow rerunning the async function otherwise.

i think that requirement is imposed by https://github.com/valora-inc/wallet/pull/4618/files#diff-c7b8e61e180cc110ced65b99b00a4bb32ccc463744e802ddef49b9d033ed792fR183, i'm not sure what breaks if we remove it. i haven't thought it completely through so if you say we need the .reset then we can keep it

src/tokens/TokenImport.tsx
```diff
@@ -27,58 +30,175 @@ import variables from 'src/styles/variables'
...
+    onError: (error) => {
+      Logger.error(TAG, `Could not fetch token details`, error)
+      const tokenId = getTokenId(networkId, tokenAddress.toLowerCase())
+
+      // it doesn't directly show if it's a timeout error, need to check cause recursively
+      const hasTimeout = (error: unknown): boolean =>
+        error instanceof BaseError &&
+        (error instanceof TimeoutError || (error.cause !== undefined && hasTimeout(error.cause)))
+      if (hasTimeout(error)) {
+        setError(t('tokenImport.error.invalidToken'))
```
bakoushin, line 147: looks like we can set this once on top, leaving `if` clauses just for analytics

src/tokens/TokenImport.tsx
denvalora: Similar feedback to https://github.com/valora-inc/wallet/pull/4618#discussion_r1424216258

Updated now! It was a remnant of those other error messages and I didn't refactor this code.

PR #4620, opened 2023-12-11
chore: refactor getting fee currencies from selector instead of hook

kathaypacific:
### Description

Pulling this part out of #4585. Currently we have a hook for getting the fee currencies given a network, but for the wallet connect work, we will need to prepare the transactions / get the fee currency from sagas so we need a selector instead. I don't think we need both a selector and a hook, so in this PR I've replaced the existing hook with a selector.

I got a bit of analysis paralysis doing this work so I think there might be a way simpler solution that I've overlooked. I wanted to keep the interface the same - consumers should be able to fetch a list of fee currencies for one networkId, however selectors have only 1 memoized value. The simplest option I could think of was to have 1 (unexported) selector calculate fee currencies for all networks upfront, and then have a selector that returns the fee currencies that consumers actually need. Open to alternative ideas...

### Test plan

Unit tests

### Related issues

- Related to RET-905

### Backwards compatibility

Y

### DIFF
--- a/src/fees/hooks.test.tsx
+++ b/src/fees/hooks.test.tsx
@@ -2,10 +2,9 @@ import { render } from '@testing-library/react-native'
 import React from 'react'
 import { Text, View } from 'react-native'
 import { Provider } from 'react-redux'
-import { useFeeCurrencies, useMaxSendAmountByAddress } from 'src/fees/hooks'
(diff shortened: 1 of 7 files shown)


jeanregisser (approved):
👍

PR #4621, opened 2023-12-11
feat: add automatic fee currency selection to walletconnect viem flow

kathaypacific:
### Description

This PR:
1. always create prepared transactions on receiving an action request from walletconnect, pass the prepared transactions to the ActionRequest display component. always calculate the gas and feeCurrency for celo network transactions.
2. if using viem, use the prepared transaction to show a warning if the user does not have enough balance to pay for gas. we can also use the prepared transaction to show the gas fees later.
3. on accepting the action request, both the walletconnect payload _and_ prepared transactions are passed to the saga that handles submitting the transactions to the network. if using viem, it is expected that the prepared transactions are there and can be directly sent to the network. if not using viem, the prepared transactions are ignored.

### Test plan

I am able to make a dapp transaction on Celo and the fee currency is paid in a non native currency.

https://github.com/valora-inc/wallet/assets/20150449/41086562-6ba3-49ad-829d-d9d7a184b634

### Related issues

- Fixes RET-905

### Backwards compatibility

Y

### DIFF
--- a/e2e/src/usecases/WalletConnectV2.js
+++ b/e2e/src/usecases/WalletConnectV2.js
@@ -53,6 +53,7 @@ async function formatTestTransaction(address, web3Library) {
         maxFeePerGas: '0x1DCD65000',
         maxPriorityFeePerGas: '0x77359400',
         value: '0x01',
+        feeCurrency: undefined,
       }
     } else {
       return {
--- a/e2e/src/usecases/WalletConnectV2.js
+++ b/e2e/src/usecases/WalletConnectV2.js
@@ -96,10 +96,6 @@ const verifySuccessfulTransaction = async (title = 'Confirm transaction', tx) =>
     .toBeVisible()
     .withTimeout(15 * 1000)
 
-  await expect(element(by.id('WalletConnectRequest/ActionRequestPayload/Value'))).toHaveText(
-    `[${JSON.stringify(tx)}]`
-  )
-
   await element(by.id('WalletConnectActionRequest/Allow')).tap()
   await enterPinUiIfNecessary()
   await verifySuccessfulConnection()
--- a/e2e/src/usecases/WalletConnectV2.js
+++ b/e2e/src/usecases/WalletConnectV2.js
@@ -261,12 +261,7 @@ export default WalletConnect = () => {
           chainId: 'eip155:44787',
           request: {
             method: 'eth_signTransaction',
-            params: [
-              {
-                ...tx,
-                feeCurrency: '0xf194afdf50b03e69bd7d057c1aa9e10c9954e4c9', // CELO native currency
-              },
-            ],
+            params: [tx],
           },
         })
 
@@ -284,7 +279,6 @@ export default WalletConnect = () => {
           const recoveredTx = parseTransactionCelo(signedTx)
           jestExpect(recoveredTx.nonce).toEqual(hexToNumber(tx.nonce))
           jestExpect(recoveredTx.to).toEqual(tx.to)
-          jestExpect(recoveredTx.gas).toEqual(BigInt(tx.gas))
           jestExpect(recoveredTx.value).toEqual(BigInt(tx.value))
         } else {
           const [recoveredTx, recoveredSigner] = recoverTransaction(signedTx)
--- a/e2e/src/usecases/WalletConnectV2.js
+++ b/e2e/src/usecases/WalletConnectV2.js
@@ -53,7 +53,6 @@ async function formatTestTransaction(address, web3Library) {
         maxFeePerGas: '0x1DCD65000',
         maxPriorityFeePerGas: '0x77359400',
         value: '0x01',
-        feeCurrency: undefined,
       }
     } else {
       return {
@@ -262,7 +261,12 @@ export default WalletConnect = () => {
           chainId: 'eip155:44787',
           request: {
             method: 'eth_signTransaction',
-            params: [tx],
+            params: [
+              {
+                ...tx,
+                feeCurrency: '0xf194afdf50b03e69bd7d057c1aa9e10c9954e4c9', // CELO native currency
+              },
+            ],
           },
         })
 
--- a/src/viem/preparedTransactionSerialization.ts
+++ b/src/viem/preparedTransactionSerialization.ts
@@ -1,6 +1,10 @@
 // Helper functions making prepared transactions serializable
 // so they can be used in redux actions (or even stores).
-import { TransactionRequest } from 'src/viem/prepareTransactions'
+import {
+  PreparedTransactionsPossible,
+  PreparedTransactionsResult,
+  TransactionRequest,
+} from 'src/viem/prepareTransactions'
 
 const bigIntProps = ['value', 'gas', 'maxFeePerGas', 'maxPriorityFeePerGas'] as const
 
@@ -27,6 +31,15 @@ function mapBigIntsToStrings<T extends object, K extends keyof T>(
 }
 
 export type SerializableTransactionRequest = MapValuesToString<TransactionRequest, BigIntProps>
+type SerializablePreparedTransactionsPossible = Omit<
+  PreparedTransactionsPossible,
+  'transactions'
+> & {
+  transactions: SerializableTransactionRequest[]
+}
+export type SerializablePreparedTransactionsResult =
+  | Exclude<PreparedTransactionsResult, PreparedTransactionsPossible>
+  | SerializablePreparedTransactionsPossible
 
 export function getSerializablePreparedTransaction(
   preparedTransaction: TransactionRequest
--- a/src/viem/preparedTransactionSerialization.ts
+++ b/src/viem/preparedTransactionSerialization.ts
@@ -1,10 +1,6 @@
 // Helper functions making prepared transactions serializable
 // so they can be used in redux actions (or even stores).
-import {
-  PreparedTransactionsPossible,
-  PreparedTransactionsResult,
-  TransactionRequest,
-} from 'src/viem/prepareTransactions'
+import { TransactionRequest } from 'src/viem/prepareTransactions'
 
 const bigIntProps = ['value', 'gas', 'maxFeePerGas', 'maxPriorityFeePerGas'] as const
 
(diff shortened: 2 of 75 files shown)


src/walletConnect/saga.ts
```diff
@@ -331,12 +343,46 @@ function getSupportedChains() {
...
+      tx.gas = hexToBigInt(tx.gas)
+    }
+  }
+
+  // Force upgrade legacy tx to EIP-1559/CIP-42/CIP-64
+  if (tx.gasPrice !== undefined) {
+    delete tx.gasPrice
+  }
+
+  // TODO: check if we need to add the nonce
```
jeanregisser, line 374: Yes, viem's `sendTransaction` adds it if not there, but if we're only signing, we should ensure it's set.

jeanregisser (commented):
Awesome! 🙌 

Mostly some details to check, to ensure it works as expected.

src/viem/preparedTransactionSerialization.ts
```diff
@@ -27,6 +31,15 @@ function mapBigIntsToStrings<T extends object, K extends keyof T>(
 }
 
 export type SerializableTransactionRequest = MapValuesToString<TransactionRequest, BigIntProps>
+type SerializablePreparedTransactionsPossible = Omit<
+  PreparedTransactionsPossible,
```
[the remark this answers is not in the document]
kathaypacific: yes you're right...i did a half hearted job here. i removed these types and started passing unserializable navigation params around, that resulted only in warnings but the app works. this didn't make me feel good, so i ended up making more explicit navigation parameters rather than passing the whole preparedTransactionsResult. i think i like where it ended up, what do you think?

src/viem/preparedTransactionSerialization.ts
```diff
@@ -27,6 +31,15 @@ function mapBigIntsToStrings<T extends object, K extends keyof T>(
...
+  'transactions'
+> & {
+  transactions: SerializableTransactionRequest[]
+}
+export type SerializablePreparedTransactionsResult =
+  | Exclude<PreparedTransactionsResult, PreparedTransactionsPossible>
+  | SerializablePreparedTransactionsPossible
```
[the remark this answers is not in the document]
kathaypacific: i think that is part of the `Exclude<PreparedTransactionsResult, PreparedTransactionsPossible>`? i've removed this anyway

src/walletConnect/request.ts
```diff
@@ -44,20 +49,27 @@ export function* handleRequest({
...
-        const txRequest: any = yield* call(prepareTransactionRequest, wallet, normalizedTx)
-        return (yield* call([wallet, 'sendTransaction'], txRequest)) as string
+        if (!serializableTransactionRequest) {
+          throw new Error('preparedTransaction is required when using viem')
+        }
+        return (yield* call(
+          // @ts-ignore TODO: fix types
+          [wallet, 'sendTransaction'],
+          getPreparedTransaction(serializableTransactionRequest)
+        )) as string
```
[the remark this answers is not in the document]
kathaypacific: i need to dig this further but i made a transaction with Uniswap and the transaction did use cUSD as the fee currency, i can see that the method that was called was eth_sendTransaction. so i think it is working correctly, i'll spend some more time digging into why / adding some e2e tests to be sure (but would like to unblock this PR)

src/walletConnect/saga.ts
kathaypacific: thanks for clarifying, i added the nonce logic from the old normalize function back

src/walletConnect/saga.ts
```diff
@@ -331,12 +347,58 @@ function getSupportedChains() {
   })
 }
 
+export function* normalizeTransaction(rawTx: any, network: Network) {
+  const tx: TransactionRequest = { ...rawTx }
+
+  // Handle `gasLimit` as a misnomer for `gas`, it usually comes through in hex format
+  if ('gasLimit' in tx && tx.gas === undefined) {
+    tx.gas = isHex(tx.gasLimit) ? hexToBigInt(tx.gasLimit as Hex) : (tx.gasLimit as bigint)
```
bakoushin, line 355: why do we need conversion to a `BigInt`?

src/walletConnect/saga.ts
```diff
@@ -331,12 +347,58 @@ function getSupportedChains() {
...
+    delete tx.gasLimit
+  }
+
+  // re-calculate the feeCurrency and gas since it is not possible to tell from
+  // the request payload if the feeCurrency is set to undefined (native
+  // currency) explicitly or due to lack of feeCurrency support
+  if (network === Network.Celo) {
+    delete tx.gas
+    if ('feeCurrency' in tx) {
```
bakoushin, line 364: apparently we can omit this check, because delete would't be upset if the property is not present

src/viem/preparedTransactionSerialization.ts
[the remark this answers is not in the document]
jeanregisser: oh right, thanks 👍

src/walletConnect/saga.ts
```diff
@@ -331,12 +347,58 @@ function getSupportedChains() {
...
+      delete tx.feeCurrency
+    }
+  } else {
+    if (isHex(tx.gas)) {
+      tx.gas = hexToBigInt(tx.gas)
```
bakoushin, line 369: this looks not consistent with casting the `gasLimit` property to `bigint` earlier in this function

src/walletConnect/saga.ts
```diff
@@ -361,11 +423,37 @@ function* showActionRequest(request: Web3WalletTypes.EventArguments['session_req
...
   navigate(Screens.WalletConnectRequest, {
     type: WalletConnectRequestType.Action,
     pendingAction: request,
     supportedChains,
     version: 2,
+    hasInsufficientGasFunds: preparedTransactionsResult?.type === 'not-enough-balance-for-gas',
+    feeCurrencies,
+    preparedTransaction:
+      preparedTransactionsResult?.type === 'possible'
+        ? getSerializablePreparedTransaction(preparedTransactionsResult.transactions[0])
```
bakoushin, line 455: do we need to serialize `TransactionReqest` here only to [deserialize back to it later](https://github.com/valora-inc/wallet/blob/c60c69682f9e1b8dc26e99b9285517a587980f42/src/walletConnect/request.ts#L61)?

jeanregisser (approved):
Already approved, and agree with what @bakoushin raised.

src/walletConnect/screens/ActionRequest.tsx
```diff
@@ -15,13 +19,22 @@ import RequestContent, { useDappMetadata } from 'src/walletConnect/screens/Reque
...
+  preparedTransaction?: SerializableTransactionRequest
 }
 
-function ActionRequest({ pendingAction, supportedChains }: Props) {
+function ActionRequest({
+  pendingAction,
+  supportedChains,
+  hasInsufficientGasFunds,
+  feeCurrencies,
+  preparedTransaction,
```
bakoushin, line 36: we still display the initial unaltered transaction in the bottom sheet. will we address it in another task?

bakoushin (approved):
Epic work! 🎀

src/walletConnect/saga.ts
kathaypacific: good question, i moved this function from the request.ts file and didn't look at the contents too closely. since gasLimit shouldn't really exist, it's hard to anticipate what type it will be which is why i am assuming that the type cast existed in the first place. i have added a proper transform to bigint if the gasLimit is an acceptable type

src/walletConnect/saga.ts
kathaypacific: yes....but then typescript complains 🙈 so i've left it like this

src/walletConnect/saga.ts
kathaypacific: it is mainly to prevent non serializable values being passed through navigation parameters. it generates a warning and the side effects are not so serious for us ([unserializable nav params mess with state persistence / deep linking](https://reactnavigation.org/docs/troubleshooting/#i-get-the-warning-non-serializable-values-were-found-in-the-navigation-state)) but i tried to avoid the bad practice here anyway. the prepared transaction then gets sent through a redux action, reducers also expect serializable values. non serializable action parameters stop Flipper from working ([context](https://valora-app.slack.com/archives/C025V1D6F3J/p1698416818574259))

src/walletConnect/saga.ts
kathaypacific: yes good spot...i've created a function to use for this

src/walletConnect/saga.ts
```diff
@@ -361,11 +423,37 @@ function* showActionRequest(request: Web3WalletTypes.EventArguments['session_req
...
+    })
+  }
+
   navigate(Screens.WalletConnectRequest, {
     type: WalletConnectRequestType.Action,
     pendingAction: request,
     supportedChains,
     version: 2,
+    hasInsufficientGasFunds: preparedTransactionsResult?.type === 'not-enough-balance-for-gas',
+    feeCurrencies,
```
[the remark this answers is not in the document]
kathaypacific: yes.....i kind of accepted this because feeCurrencies is coming directly from redux where we should have no unserializable values anyway (if we fix the redux problem, this would be solved too) but i suppose there's no reason to keep contributing to the problem for this component. we only need to display the symbols so i've updated this to be a string array

src/walletConnect/request.ts
```diff
@@ -44,20 +49,27 @@ export function* handleRequest({
...
-        const txRequest: any = yield* call(prepareTransactionRequest, wallet, normalizedTx)
-        return (yield* call([wallet, 'signTransaction'], txRequest)) as string
+        if (!serializableTransactionRequest) {
+          throw new Error('preparedTransaction is required when using viem')
+        }
+        return (yield* call(
+          [wallet, 'signTransaction'],
+          // TODO: fix types
+          getPreparedTransaction(serializableTransactionRequest) as any
+        )) as string
```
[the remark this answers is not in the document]
kathaypacific: yes, this was existing before so i didn't question it. but it seems to work without so i've removed it

src/walletConnect/screens/ActionRequest.tsx
kathaypacific: thanks for the catch! i wasn't thinking about this, seems to make sense to add it here so i've updated the corresponding component

kathaypacific:
TY for the reviews @bakoushin @jeanregisser !! i've addressed comments and i think the changes are minor enough where i'll just merge. i verified again this morning that we're able to use a non native fee currency for sending tx's using the current solution, but i still intend to dig into the viem sendTransaction feeCurrency topic. so if there are any lingering comments i can also address as follow ups. i'll merge to unblock the other WC tasks

e2e/src/usecases/WalletConnectV2.js
```diff
@@ -96,10 +96,6 @@ const verifySuccessfulTransaction = async (title = 'Confirm transaction', tx) =>
     .toBeVisible()
     .withTimeout(15 * 1000)
 
-  await expect(element(by.id('WalletConnectRequest/ActionRequestPayload/Value'))).toHaveText(
```
kathaypacific, line 99: i hate to fix a test by removing an assertion, but this part is already covered by unit tests and with the changes in this PR, the tx will use viem to estimate the gas / fee currency. there will no longer be a static tx object to compare against, so i think it makes sense that this assertion is reserved for unit tests rather than e2e

PR #4622, opened 2023-12-11
feat(swap): update max fee currency warning

jeanregisser:
### Description

This shows the MAX amount warning for ETH too.

Also it tweaks the logic on Celo, so it only displays the warning if there's only 1 fee currency with a balance.
For instance if the user only has cUSD and tries to swap the MAX amount of cUSD, the warning is also shown.

### Test plan

- Updated tests
- Manual tests:

|  **before**  | **after**   |
|---|---|
| image   |  image |

Warning when only cUSD is left to pay for gas and MAX amount is chosen:

### Related issues

- Fixes RET-946

### Backwards compatibility

Yes

### DIFF
--- a/src/swap/SwapScreen.tsx
+++ b/src/swap/SwapScreen.tsx
@@ -46,7 +46,7 @@ import { swapStart, swapStartPrepared } from 'src/swap/slice'
 import { Field, SwapAmount } from 'src/swap/types'
 import useSwapQuote, { QuoteResult } from 'src/swap/useSwapQuote'
 import { useSwappableTokens, useTokenInfo } from 'src/tokens/hooks'
-import { feeCurrenciesSelector, tokensByIdSelector } from 'src/tokens/selectors'
+import { feeCurrenciesWithPositiveBalancesSelector, tokensByIdSelector } from 'src/tokens/selectors'
 import { TokenBalance } from 'src/tokens/slice'
 import { getSupportedNetworkIdsForSwap, getTokenId } from 'src/tokens/utils'
 import { NetworkId } from 'src/transactions/types'
@@ -88,18 +88,6 @@ function getNetworkFee(quote: QuoteResult | null, networkId?: NetworkId) {
     networkFee: feeAmount,
     feeTokenId: feeCurrency?.tokenId ?? getTokenId(networkId ?? networkConfig.defaultNetworkId), // native token
   }
-}
-
-function useFeeCurrenciesWithBalances(networkId: NetworkId) {
-  const feeCurrencies = useSelector((state) => feeCurrenciesSelector(state, networkId))
-
-  const feeCurrenciesWithBalances = useMemo(() => {
-    return feeCurrencies.filter((feeCurrency) => {
-      return feeCurrency.balance.isGreaterThan(0)
-    })
-  }, [feeCurrencies])
-
-  return feeCurrenciesWithBalances
 }
 
 export function SwapScreen({ route }: Props) {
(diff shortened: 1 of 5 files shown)


src/swap/SwapScreen.tsx
```diff
@@ -90,6 +90,18 @@ function getNetworkFee(quote: QuoteResult | null, networkId?: NetworkId) {
...
+function useFeeCurrenciesWithBalances(networkId: NetworkId) {
+  const feeCurrencies = useSelector((state) => feeCurrenciesSelector(state, networkId))
+
+  const feeCurrenciesWithBalances = useMemo(() => {
+    return feeCurrencies.filter((feeCurrency) => {
+      return feeCurrency.balance.isGreaterThan(0)
+    })
+  }, [feeCurrencies])
+
+  return feeCurrenciesWithBalances
```
kathaypacific, line 102: what do you think about making this a selector instead? i can't think of an example now but this seems like it could be useful outside of swaps too. i know we've been creating selectors _and_ hook pairs and i'm not sure that i understand why, it seems like selectors are superior because they can be more easily reused in sagas and we don't have to worry about `useMemo`?

src/swap/SwapScreen.tsx
```diff
@@ -640,8 +652,10 @@ export function SwapScreen({ route }: Props) {
           {showMaxSwapAmountWarning && (
             <InLineNotification
               severity={Severity.Warning}
-              title={t('swapScreen.maxSwapAmountWarning.title')}
-              description={t('swapScreen.maxSwapAmountWarning.body')}
+              title={t('swapScreen.maxSwapAmountWarning.title', { tokenSymbol: fromToken?.symbol })}
```
kathaypacific, line 655: did you add the tokenSymbol here in anticipation that we might use it in the future? its not currently needed so i wondered

src/swap/SwapScreen.tsx
```diff
@@ -525,7 +536,7 @@ export function SwapScreen({ route }: Props) {
         decimalSeparator,
       }),
     }))
-    showMaxCeloSwapWarning()
+    hideOrShowMaxFeeCurrencySwapWarning(fromToken)
```
kathaypacific, line 539: just wanted to check if you considered triggering this from only a useEffect after the fromToken / fromAmount has changed, rather than explicitly triggering it within the function handlers? idk if the useEffect works for all the scenarios, i didn't think it through entirely but it'd be less likely to break if we didn't manually coordinate the warning so it'd be nice if it worked 😀

src/swap/SwapScreen.tsx
```diff
@@ -640,8 +652,10 @@ export function SwapScreen({ route }: Props) {
           {showMaxSwapAmountWarning && (
             <InLineNotification
               severity={Severity.Warning}
-              title={t('swapScreen.maxSwapAmountWarning.title')}
-              description={t('swapScreen.maxSwapAmountWarning.body')}
+              title={t('swapScreen.maxSwapAmountWarning.title', { tokenSymbol: fromToken?.symbol })}
+              description={t('swapScreen.maxSwapAmountWarning.body', {
```
kathaypacific, line 656: should we use new translation keys so that this is backwards compatible?

kathaypacific (commented):
love this, just a few small questions!

src/swap/SwapScreen.tsx
jeanregisser: Yes I'll turn it into a selector 👍

src/swap/SwapScreen.tsx
jeanregisser: yes.

src/swap/SwapScreen.tsx
jeanregisser: Yes I thought about this, but it seems we only want to show it if the MAX button has been pushed, or when hitting confirm and the amount is greater than the balance.
So here's there's not much alternative, than to leave it in the max button handler.

But I could have kept the `useEffect` for when `fromToken` changes, but decided to move it to the token selection handler given all  the logic is there when they change (along with the network switch), so I felt it belonged better there with the rest.

It's true this could break if we later update the code and change `fromToken` elsewhere. But other things will break anyway.

I agree it's not ideal, I think the whole state logic in this screen could be better handled with `useReducer`.

src/swap/SwapScreen.tsx
jeanregisser: oh right, good catch!

src/swap/SwapScreen.tsx
jeanregisser: The hook with `useMemo` is simpler for situations where it's not needed for sagas.
Also allows keeping the memoized values only for when the component is mounted.

Whereas memoized `createSelector` are global and never cleared.

kathaypacific (approved):
🚀

PR #4623, opened 2023-12-11
fix(viem): Fix Viem gas estimation bug

jophish:
### Description

Previously, in cases where we were attempting to estimate gas for a TX where `gas + amount > balance`, we would _throw_ rather than returning a null estimation. This had the downstream effect of causing the gas estimation on the SendEnterAmount screen to spin forever, _without_ a warning to the user telling them to enter a lower amount. This change catches that case, and sets the gas estimate to null, causing the appropriate warning to appear to the user.

### Test plan

Manual and unit tested. See video below:

https://github.com/valora-inc/wallet/assets/569401/e012637f-45fd-4ad6-87a8-03cf8319d0b5

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/src/viem/prepareTransactions.test.ts
+++ b/src/viem/prepareTransactions.test.ts
@@ -25,6 +25,7 @@ import {
   EstimateGasExecutionError,
   ExecutionRevertedError,
   InsufficientFundsError,
+  InvalidInputRpcError,
   encodeFunctionData,
 } from 'viem'
 import { estimateGas } from 'viem/actions'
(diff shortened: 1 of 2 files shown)


MuckT (approved):
Small issue with the flashing loading state while fees are fetched, but that seems outside of the scope of the PR, but noting that here.

https://github.com/valora-inc/wallet/assets/26950305/ae357c2c-9785-4239-8d2a-0128ffe61993

PR #4625, opened 2023-12-12
fix(send): Assorted Send recipient bugfixes

jophish:
### Description

This PR adjusts the behavior of recipients on the Send flow in order to fix a handful of bugs. Really, it's trading in very undesirable behavior for somewhat less undesirable behavior.

Functionally, this fixes the following issues:

* Fixes an issue where you revoke your phone number and can no longer send to addresses associated with phone numbers that were previously cached on your device. Searching for such an address now will allow you to send to that address.
* Fixes V icon behavior with recipients from above.
* If your phone number becomes unlinked, we do _not_ show contact recipients anywhere. Since contact recipients are always considered `RecipientType.PhoneNumber`, we would always attempt to perform a CPV lookup, which we _cannot do_ since the user's wallet is unlinked. If you happened to send to a contact before revoking your phone number such that that contact has become a recent recipient, you can still send to that recent recipient via the fix from the first bullet. 

Some potential issues/buggy behavior that still exist after these changes:

 - You have a PN cached that maps to some address A
   - Address A is revoked from PN mapping OR your revoke your phone number
 - You search address A
 - RecipientType.Address recipient appears, as a phone number line item
 - Pressing on the recipient will kick off an address verification lookup
   - This will either fail because address A has been revoked OR fail because /you/ revoked your phone number
 - Valora V will be removed from phone number item if present
 - "unknown address" warning will appear at bottom

also:

- Send to a phone number, so it becomes a recent recipient and stores PN -> address A in cache
- That phone number is revoked
- Searching for a prefix of address A will show the recent (PN) recipient from above
  - Pressing this will show an "invite" button, despite there being an address associated with that recent recipient
  - (Searching for the full address A will yield a resolved address recipient, which we /can/ send to - same case as bug above)
- However, pressing the recent recipient from the initial `SendSelectRecipient` screen will take you straight to `SendEnterAmount` since we skip the extra button press for the list of recent recipients

### Test plan

Unit tests, manually

Searching for addresses connected to phone number, first address is still connected, second is no longer connected

| Before | After |
|--------|--------|
| <video src="https://github.com/valora-inc/wallet/assets/5062591/8ca4b979-c130-439f-a5b0-4d2ad9e9c62e" /> | <video src="https://github.com/valora-inc/wallet/assets/5062591/6fe671ea-eb7e-4800-bf3a-7690b1949eb7" /> | 

Searching for contact with phone unverified, but contacts previously shared

| Before | After |
|--------|--------|
| <video src="https://github.com/valora-inc/wallet/assets/5062591/1fd610bb-1191-4783-9a3f-4ce612a6d39c" /> | <video src="https://github.com/valora-inc/wallet/assets/5062591/b8f4d3c7-a3c3-465d-87f0-eb91a18b5207" /> | 

### Related issues

- Part of ACT-971

### Backwards compatibility

Yes

### DIFF
--- a/src/recipients/recipient.ts
+++ b/src/recipients/recipient.ts
@@ -6,8 +6,8 @@ import { formatShortenedAddress } from 'src/components/ShortenedAddress'
 import {
   AddressToDisplayNameType,
   AddressToE164NumberType,
-  E164NumberToAddressType,
   AddressToVerificationStatus,
+  E164NumberToAddressType,
 } from 'src/identity/reducer'
 import { RecipientVerificationStatus } from 'src/identity/types'
 import Logger from 'src/utils/Logger'
@@ -178,20 +178,7 @@ export function getRecipientVerificationStatus(
   e164NumberToAddress: E164NumberToAddressType,
   addressToVerificationStatus: AddressToVerificationStatus
 ): RecipientVerificationStatus {
-  if (!recipientHasNumber(recipient)) {
-    if (recipientHasAddress(recipient) && recipient.address in addressToVerificationStatus) {
-      switch (addressToVerificationStatus[recipient.address]) {
-        case true:
-          return RecipientVerificationStatus.VERIFIED
-        case false:
-          return RecipientVerificationStatus.UNVERIFIED
-        case undefined:
-          return RecipientVerificationStatus.UNKNOWN
-      }
-    } else {
-      return RecipientVerificationStatus.UNKNOWN
-    }
-  } else {
+  if (recipient.recipientType === RecipientType.PhoneNumber && recipientHasNumber(recipient)) {
     const addresses = e164NumberToAddress[recipient.e164PhoneNumber]
     if (addresses === undefined) {
       return RecipientVerificationStatus.UNKNOWN
@@ -202,6 +189,18 @@ export function getRecipientVerificationStatus(
     }
 
     return RecipientVerificationStatus.VERIFIED
+  }
+  if (recipientHasAddress(recipient) && recipient.address in addressToVerificationStatus) {
+    switch (addressToVerificationStatus[recipient.address]) {
+      case true:
+        return RecipientVerificationStatus.VERIFIED
+      case false:
+        return RecipientVerificationStatus.UNVERIFIED
+      case undefined:
+        return RecipientVerificationStatus.UNKNOWN
+    }
+  } else {
+    return RecipientVerificationStatus.UNKNOWN
   }
 }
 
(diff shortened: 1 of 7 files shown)


src/recipients/recipient.ts
```diff
@@ -178,20 +178,7 @@ export function getRecipientVerificationStatus(
...
-        case false:
-          return RecipientVerificationStatus.UNVERIFIED
-        case undefined:
-          return RecipientVerificationStatus.UNKNOWN
-      }
-    } else {
-      return RecipientVerificationStatus.UNKNOWN
-    }
-  } else {
+  if (recipient.recipientType === RecipientType.PhoneNumber && recipientHasNumber(recipient)) {
```
jh2oman, line 181: In an ideal world we would just use `recipient.recipientType === RecipientType.PhoneNumber ` to do a check like this, right? Like ideally `PhoneNumber` recipients *should* always have a phone number? If I'm understanding that right, maybe a TODO will be helpful here for future us to help explain why we have this double phone number check.

src/recipients/recipient.ts
```diff
@@ -203,6 +190,18 @@ export function getRecipientVerificationStatus(
 
     return RecipientVerificationStatus.VERIFIED
   }
+  if (recipientHasAddress(recipient) && recipient.address in addressToVerificationStatus) {
```
jh2oman, line 193: same question / suggestion for a todo

src/send/useFetchRecipientVerificationStatus.ts
```diff
@@ -36,7 +36,10 @@ const useFetchRecipientVerificationStatus = () => {
     setRecipient(selectedRecipient)
     setRecipientVerificationStatus(RecipientVerificationStatus.UNKNOWN)
 
-    if (selectedRecipient?.e164PhoneNumber) {
+    if (
+      selectedRecipient.recipientType === RecipientType.PhoneNumber &&
```
jh2oman, line 40: same question / suggestion for a todo

jh2oman (approved):
just a question & suggestion for comments

src/recipients/recipient.ts
satish-ravi: this was mainly to make TS happy, I can add a comment.

src/recipients/recipient.ts
satish-ravi: this is checking two different things

jh2oman: oh, you're right. my bad on this one

PR #4627, opened 2023-12-12
chore(l10n): update translations from Crowdin

valora-bot-crowdin:


locales/base/translation.json
```diff
@@ -1676,7 +1676,9 @@
     },
     "maxSwapAmountWarning": {
       "title": "Swapping the MAX amount",
+      "titleV1_74": "Swapping the MAX amount",
```
MuckT, line 1679: What's going on with this key?

PR #4631, opened 2023-12-13
chore(analytics): add analytics events for deeplinks and qr code scans

jh2oman:
### Description

Adding in analytics events for deeplinks and qr code scans so we can get a better idea of how these features are used.

### Test plan

Unit tests


src/app/saga.ts
```diff
@@ -339,6 +339,14 @@ export function* handleDeepLink(action: OpenDeepLink) {
   const rawParams = parse(deepLink)
   if (rawParams.path) {
     const pathParts = rawParams.path.split('/')
+    ValoraAnalytics.track(AppEvents.handle_deeplink, {
+      pathStartsWith: pathParts[1].split('?')[0],
+      fullPath:
+        pathParts[1] === 'jumpstart'
+          ? 'redacted-due-to-presense-of-private-key'
```
jh2oman, line 346: don't want to cause a privacy issue

satish-ravi: nit: maybe enclose these in `[]` so it doesn't look similar to other full paths

src/navigator/QRNavigator.tsx
```diff
@@ -151,7 +151,6 @@ function AnimatedScannerScene({ route, position }: AnimatedScannerSceneProps) {
     if (lastScannedQR.current === qrCode.data) {
       return
     }
-
```
satish-ravi, line 154: nit: this seems like an intended change for this PR

src/analytics/Properties.tsx
```diff
@@ -67,6 +67,7 @@ import { NftOrigin } from 'src/nfts/types'
 import { NotificationReceiveState } from 'src/notifications/types'
 import { AdventureCardName } from 'src/onboarding/types'
 import { RecipientType } from 'src/recipients/recipient'
+import { QrCode } from 'src/send/actions'
```
satish-ravi, line 70: this is probably fine as is but would be nice if this is on send/types so there isn't any cyclic imports.

satish-ravi (approved):

src/app/saga.ts
jophish: Wow, didn't realize we had a deeplink that passes in a private key! :exploding_head: 

This does feel a bit brittle though wrt avoiding privacy issues. If a deeplink is added that contains sensitive information in the path, this will need to be manually updated, which we could easily forget to do. Probably not a big deal, but thought I'd call it out - maybe worth a comment in code?

cajubelt: great catch @jh2oman . maybe instead of including query params by default, we could allowlist them, to err on the side of caution here

src/qrcode/utils.ts
```diff
@@ -144,6 +144,8 @@ function* extractQRAddressData(qrCode: QrCode) {
 // Catch all handler for QR Codes
 // includes support for WalletConnect, hooks, and send flow (non-secure send)
 export function* handleQRCodeDefault({ qrCode }: HandleQRCodeDetectedAction) {
+  ValoraAnalytics.track(QrScreenEvents.qr_scanner_scanned, qrCode)
```
cajubelt, line 147: naming nit: how about qr_scanned? instead of qr_scanner_scanned

cajubelt (approved):
+1 to Satish's comments, also left a couple ideas, but nothing major so approving

PR #4633, opened 2023-12-13
fix(logs): logs sometimes cannot be attached

jeanregisser:
### Description

This should fix logs failing to be attached when sent to Zendesk.

Instead of using `react-native-fs` to read the logs in chunk, we now create a `Blob` which is kept on the native side,
then it's uploaded. 
The advantage is the data doesn't need to be copied in JS memory and we should avoid the utf8 read error.
So more performant and will work better with large files.

Note: I noticed my log files are rather large (30-40MB!), we should address this too. Either by compressing or reducing the unnecessary noise they contain.

### Test plan

- Updated unit tests
- Manually tested

### Related issues

- Fixes RET-948

### Backwards compatibility

Yes

### DIFF
--- a/src/account/zendesk.test.ts
+++ b/src/account/zendesk.test.ts
@@ -4,9 +4,33 @@ import { ZENDESK_API_KEY } from 'src/config'
 
 const mockFetch = fetch as FetchMock
 
-jest.mock('src/utils/readFile', () => ({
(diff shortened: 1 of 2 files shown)


MuckT (approved):
Tested Locally on Android 13 and it looks good to me. Once we remove the native email path and update the debugger screen it looks like we'll be able to remove all of the file chunking logic.

PR #4634, opened 2023-12-13
chore(swap): remove swap with ContractKit

jeanregisser:
### Description

As the title says.

### Test plan

- Updated tests

### Related issues

- Fixes RET-939

### Backwards compatibility

Yes


src/swap/slice.ts
```diff
@@ -34,20 +34,12 @@ export const slice = createSlice({
...
   reducers: {
-    // Legacy
     swapStart: (state, action: PayloadAction<SwapInfo>) => {
       state.currentSwap = {
         id: action.payload.swapId,
         status: 'started',
       }
     },
-    // New flow with prepared transactions
-    swapStartPrepared: (state, action: PayloadAction<SwapInfoPrepared>) => {
```
jeanregisser, line 45: The `swapStartPrepared` is now just `swapStart`

src/swap/SwapScreen.test.tsx
jeanregisser, line 1: It's a bit hard to read this diff, because of the changed indentation for the test cases that were only for viem. Sorry.

kathaypacific: it's not so bad, there's a review setting to hide white spaces!

kathaypacific (approved):
🧹✨ yay so great to see the contractkit stuff gone!!

are you also planning to remove the rawResponse property from the quote result as part of this work? we can also create a separate task for that

jeanregisser:
> are you also planning to remove the rawResponse property from the quote result as part of this work? we can also create a separate task for that

yes I'll do that too.

PR #4635, opened 2023-12-13
test(e2e): e2e tests for new send flow

finnian0826:
### Description

[ACT-970](https://linear.app/valora/issue/ACT-970/e2e-tests-for-updated-send-ui-celo), adds e2e tests for the new send flow.

Adds tests for:
- Send to an address
- Send to a recent recipient
- Send to a PN with one address mapping to it
- Send to a PN with multiple addresses (secure send flow)

Also, removed the contract-kit testing for the old flow, now just does viem (that's why the old tests look changed, but really they are just un-indented, not other change).

### Test plan

e2e tests pass, tests run locally and emulator watched to make sure all is as expected

### Related issues

- N/A

### Backwards compatibility

Yes


e2e/src/usecases/SecureSend.js
```diff
@@ -16,77 +16,147 @@ const faker = require('@faker-js/faker')
...
-        await waitFor(element(by.text(`${randomContent}`)))
-          .toBeVisible()
-          .withTimeout(60000)
+    it('Send cUSD to phone number with multiple mappings', async () => {
+      let randomContent = faker.lorem.words()
+      await waitFor(element(by.id('HomeAction-Send')))
+        .toBeVisible()
+        .withTimeout(30000)
+      await element(by.id('HomeAction-Send')).tap()
+      await waitFor(element(by.id('SendSearchInput'))).toBeVisible()
```
MuckT, line 40: These waitFor(s) need a `.withTimeout(<time in ms>)`. There is also a helper function `waitForElementId` in the `e2e/src/utils/utils.js` that could also be used instead.

MuckT (commented):
It would be nice to move to some of shared utility functions such as `waitForElementId` & `waitForElementByIdAndTap`.

e2e/src/usecases/SecureSend.js
```diff
@@ -10,83 +10,152 @@ import {
...
+          statsigGateOverrides: `use_new_send_flow=true,use_viem_for_send=true`,
+        },
       })
-    }
-  )
+      await quickOnboarding(SAMPLE_BACKUP_KEY_VERIFIED)
+    })
+
+    it('Send cUSD to phone number with multiple mappings', async () => {
+      let randomContent = faker.lorem.words()
```
cajubelt, line 112: is this actually random content? if so we shouldn't be testing with it-- you could imagine a case where it triggers some bug and fails, but then when the test is rerun, a new comment is chosen that doesn't trigger the bug and passes, so we can't repro the bug

e2e/src/usecases/SecureSend.js
```diff
@@ -10,83 +10,152 @@ import {
...
+      await waitForElementByIdAndTap('HomeAction-Send', 30000)
+      await waitForElementByIdAndTap('SendSelectRecipientSearchInput', 3000)
+      await element(by.id('SendSelectRecipientSearchInput')).replaceText(VERIFIED_PHONE_NUMBER)
+      await element(by.id('RecipientItem')).tap()
+
+      await waitForElementByIdAndTap('SendOrInviteButton', 30000)
+
+      // Use the last digits of the account to confirm the sender.
+      await waitForElementByIdAndTap('confirmAccountButton', 30000)
+      // TODO: test case for AddressValidationType.PARTIAL but relies on mapping phone number to another address with unique last 4 digits
```
cajubelt, line 122: is the problem here that there are two secure send flows, one for when the recipient has 2+ addresses verified with the same PN that share the last 4 digits, and one where the 2+ addresses have different last 4 digits?

cajubelt: seems like the case where they have the same last 4 digits would be more rare.. is that what's tested for here?

cajubelt: I think it's fine to choose only one of those cases for an e2e test and leave the other one, but the one chosen should be the more common case, ideally

cajubelt: if it's much harder to test the other case for some reason, punting might be ok, but it should be put into a linear issue. this todo is not going to get noticed otherwise for a long time probly.. this file doesn't get a lot of foot traffic

e2e/src/usecases/Send.js
```diff
@@ -154,4 +159,226 @@ export default Send = () => {
...
+      })
+    })
+
+    it('Then should navigate to send search input from home action', async () => {
+      await waitForElementByIdAndTap('HomeAction-Send', 30 * 1000)
+      await waitForElementId('SendSelectRecipientSearchInput', 30 * 1000)
+    })
+
+    it('Then should be able to enter an address', async () => {
+      await waitForElementByIdAndTap('SendSelectRecipientSearchInput', 30 * 1000)
```
cajubelt, line 179:
```suggestion
      await waitForElementByIdAndTap('SendSelectRecipientSearchInput', 30_000)
```
code golf

e2e/src/usecases/Send.js
```diff
@@ -154,4 +159,226 @@ export default Send = () => {
...
+      await expect(element(by.text('CELO')).atIndex(0)).toBeVisible()
+      await element(by.id('SendEnterAmount/TokenSelect')).tap()
+      await element(by.id('cUSDSymbol')).tap()
+      await expect(element(by.text('cUSD')).atIndex(0)).toBeVisible()
+      await element(by.id('SendEnterAmount/TokenSelect')).tap()
+      await element(by.id('cEURSymbol')).tap()
+      await expect(element(by.text('cEUR')).atIndex(0)).toBeVisible()
+    })
+
+    it.todo('Then should be able to use max button')
```
cajubelt, line 207: did you mean to do this? if not when is it going to get done?

cajubelt: same thing here re: putting things into linear issues

cajubelt: fwiw I think it's probably ok to just not do this though

e2e/src/usecases/Send.js
```diff
@@ -154,4 +159,226 @@ export default Send = () => {
...
+      await expect(element(by.text('0xe5f5...8846'))).toBeVisible()
+    })
+
+    it('Then should be able to add a comment', async () => {
+      await addComment('Starting Comment ❤️')
+      let comment = await element(by.id('commentInput/send')).getAttributes()
+      jestExpect(comment.text).toEqual('Starting Comment ❤️')
+    })
+
+    it.todo('Then should be able to display fee modal')
```
cajubelt, line 227: no new todos!

e2e/src/usecases/Send.js
```diff
@@ -154,4 +159,226 @@ export default Send = () => {
...
+      await enterPinUiIfNecessary()
+      await expect(element(by.id('errorBanner'))).not.toBeVisible()
+      await waitForElementId('HomeAction-Send', 30 * 1000)
+    })
+
+    it.todo('Then should display transaction as pending')
+
+    it.todo('Then should display transaction as confirmed')
+
+    it.todo('Then should display correct transaction details')
```
cajubelt, line 257: same re todos

e2e/src/usecases/SecureSend.js
cajubelt: it's ok to leave a todo if you plan to deal with it in the same issue you are currently working on, but if that's the case, it's good to leave some indication of that, like `// TODO(ACT-XYZ): foo bar`

e2e/src/usecases/Send.js
```diff
@@ -154,4 +159,226 @@ export default Send = () => {
...
+    })
+
+    it('Then should display correct recipient', async () => {
+      await expect(element(by.text('0xe5f5...8846'))).toBeVisible()
+    })
+
+    it('Then should be able to add a comment', async () => {
+      await addComment(randomComment)
+      let comment = await element(by.id('commentInput/send')).getAttributes()
+      jestExpect(comment.text).toEqual(randomComment)
```
cajubelt, line 301: no random tests!

e2e/src/usecases/Send.js
```diff
@@ -154,4 +159,226 @@ export default Send = () => {
...
+      await enterPinUiIfNecessary()
+      await expect(element(by.id('errorBanner'))).not.toBeVisible()
+      await waitForElementId('HomeAction-Send', 30 * 1000)
+    })
+
+    it.todo('Then should display transaction as pending')
+
+    it.todo('Then should display transaction as confirmed')
+
+    it.todo('Then should display correct transaction details')
```
cajubelt, line 315: no new todos!@

e2e/src/usecases/Send.js
```diff
@@ -154,4 +159,226 @@ export default Send = () => {
...
```
cajubelt, line 382: todos

e2e/src/usecases/Send.js
MuckT: This was originally used to verify that the comment showed in the transaction feed; however, there were issues with the most recent transaction not guaranteed to be from this test run and that check was removed. With more recent changes to [getAttributes()](https://wix.github.io/Detox/docs/api/actions/#getattributes) it might be possible to add this check back in.

e2e/src/usecases/SecureSend.js
MuckT: This waitFor needs a [`.withTimeout()`](https://github.com/nubank/detox/blob/master/docs/APIRef.waitFor.md#tobevisible).

e2e/src/usecases/Send.js
MuckT: I agree, if there's no active issue for adding these e.g. todo(ACT-xyz) we should just remove them.

e2e/src/usecases/SecureSend.js
```diff
@@ -1,92 +1,175 @@
...
+      await enterPinUiIfNecessary()
 
-        // Return to home screen.
-        await waitFor(element(by.id('HomeAction-Send')))
-          .toBeVisible()
-          .withTimeout(30 * 1000)
+      // Return to home screen.
+      await waitFor(element(by.id('HomeAction-Send')))
+        .toBeVisible()
+        .withTimeout(30_000)
```
MuckT, line 98: nit: consistent number styles. Personally I prefer `30 * 1000`, but `30_000` or `30000` is fine too as long as it's consistent.

e2e/src/usecases/SecureSend.js
```diff
@@ -1,92 +1,175 @@
...
   addComment,
   enterPinUiIfNecessary,
+  fundWallet,
   inputNumberKeypad,
   scrollIntoView,
   quickOnboarding,
+  waitForElementId,
+  waitForElementByIdAndTap,
 } from '../utils/utils'
 const faker = require('@faker-js/faker')
```
MuckT, line 19: Faker can be removed here if it's no longer being used. Personally, I think it's worth having to verify the transaction in the home screen; however, that's not currently tested so it's fine to remove. We can add it back later if we want to assert on the comment used in the send.

e2e/src/usecases/SecureSend.js
```diff
@@ -1,92 +1,175 @@
...
+    beforeAll(async () => {
+      // uninstall the app to remove secure send mapping
+      await device.uninstallApp()
+      await device.installApp()
+      // fund wallet for send
+      await fundWallet(
+        SAMPLE_PRIVATE_KEY,
+        SAMPLE_WALLET_ADDRESS_SINGLE_ADDRESS_VERIFIED,
+        'cUSD',
+        `${AMOUNT_TO_SEND * 2}`
```
MuckT, line 34: Is `AMOUNT_TO_SEND * 2` enough to cover the send amount and fee's? Funding the wallet each test run might be error prone and instead we should just add this wallet to`./e2e/scripts/fund-e2e-accounts.ts`; however, that script is currently failing so `AMOUNT_TO_SEND * 2.2` might be enough for this PR.

e2e/src/usecases/SecureSend.js
```diff
@@ -1,92 +1,175 @@
...
-        await waitFor(element(by.text(`${randomContent}`)))
-          .toBeVisible()
-          .withTimeout(60000)
+      await waitFor(element(by.text(`${commentText}`)))
+        .toBeVisible()
+        .withTimeout(60000)
+    })
+  })
+
+  describe('Secure send flow with phone number lookup (new flow)', () => {
```
MuckT, line 106: We might need to skip this test to avoid blocking CI for other PR's, but we can add a todo linked to ACT-1044.

jh2oman (approved):
Great job on all of these iterations! This PR added a lot of coverage

MuckT (approved):

PR #4640, opened 2023-12-14
chore: only add a swap approve transaction if the allowance is not enough

kathaypacific:
### Description

Since gas fees are expensive in some networks, it would be beneficial for users if we could minimise any redundant transactions sent to the network. Currently for the swap flow, we will always create an approval transaction for an ERC-20 sell token. There could possibly be a scenario where a user has approved a larger amount in the past, and the swap amount in question still fits into the allowance.

This PR adds a check for the spender allowance, before adding the approval transaction to the list.

### Test plan

Verified that approval transactions are omitted if there is enough allowance - this is pretty edge case for a user who is swapping on Valora only since we only approve the necessary amount. I tested by hacking an approval that is very large, and checking the subsequent swap does not contain an approval transaction.

I still think there's value here though, e.g.
1. an approval succeeds but the swap transaction fails. the user can then try again without having to pay for the approval again.
2. a user is swapping a very large amount, the swap transaction succeeds but due to slippage, some allowance is left over (enough to cover a subsequent swap that is small)

### Related issues

- Fixes RET-942

### Backwards compatibility

Y


src/swap/useSwapQuote.ts
```diff
@@ -90,17 +103,18 @@ function createBaseSwapTransactions(
   }
 }
 
-export async function prepareSwapTransactions(
+async function prepareSwapTransactions(
   fromToken: TokenBalance,
   updatedField: Field,
   unvalidatedSwapTransaction: SwapTransaction,
-  price: string,
```
kathaypacific, line 97: price is unused so i removed it

src/swap/useSwapQuote.ts
```diff
@@ -54,21 +56,32 @@ function createBaseSwapTransactions(
       ? BigInt(new BigNumber(buyAmount).times(guaranteedPrice).toFixed(0, 0))
       : BigInt(sellAmount)
 
-  // Approve transaction if the sell token is ERC-20
-  // TODO skip this if the allowance is already enough
-  if (allowanceTarget !== zeroAddress && fromToken.address) {
-    const data = encodeFunctionData({
+  // If the sell token is ERC-20, we need to check the allowance and add an
+  // approval transaction if necessary
+  if (allowanceTarget !== zeroAddress && fromToken.address && isAddress(fromToken.address)) {
```
jeanregisser, line 61: nit: Why checking `isAddress(fromToken.address)`? seems a bit too defensive, as if we were not trusting our own data.

src/swap/useSwapQuote.ts
```diff
@@ -126,6 +140,12 @@ function useSwapQuote(networkId: NetworkId, slippagePercentage: string) {
       swapAmount: ParsedSwapAmount,
       updatedField: Field
     ) => {
+      if (!walletAddress || !isAddress(walletAddress)) {
```
jeanregisser, line 143: nit: we usually aren't that defensive about walletAddress.
```suggestion
      if (!walletAddress) {
```

jeanregisser (approved):
This is great!! 🙌

src/swap/SwapScreen.tsx
```diff
@@ -108,8 +90,6 @@ export function SwapScreen({ route }: Props) {
   ).maxSlippagePercentage
   const parsedSlippagePercentage = new BigNumber(slippagePercentage).toFormat()
 
-  const useViemForSwap = getFeatureGate(StatsigFeatureGates.USE_VIEM_FOR_SWAP)
```
dievazqu, line 111: So this PR also forces the new client to use `viem`.
I think this is fine, but I didn't expect this by reading the PR title or description.

dievazqu (approved):
LGTM

src/swap/SwapScreen.tsx
jeanregisser: It's because it was initially based on my branch where I removed ContractKit, but then this PR got switched to the main branch when it was merged:

Screenshot 2023-12-15 at 10 15 39

src/swap/useSwapQuote.ts
kathaypacific: the viem functions require the address to be `Address` type, so typescript complains about the address being the `string` type (less specific). i guess broadly we've been using `as Address`, but we discussed recently to avoid type casting so i went with the typeguard here. i think it should make no difference at runtime. i'm not sure what is best, going for consistency + type casting or going with the type guard (which IMO is more correct)?

src/swap/useSwapQuote.ts
kathaypacific: same reason as above...walletAddress has type string which means later I need to type cast it to `Address` to use with viem...perhaps i should just change the type of walletAddress to be `Address` in redux, but that'll be a bigger change...

src/swap/SwapScreen.tsx
kathaypacific: yes sorry i should have mentioned this 😅 i decided not to bother with the contractkit flow because Jean's PR was about to be merged

src/swap/useSwapQuote.ts
kathaypacific: i think consistency is top priority and makes it easier to do a project wide refactor if we want in the future...so i've removed the type guards and did `as Address` in the viem function directly.

jeanregisser: I think we should turn out tokensInfo address field (and others) to be `Address` this would remain compatible with things not using viem and will avoid the need to casting when using viem.

What do you think?

kathaypacific: yes i agree - we should update the redux type to be Address, i can create a task so that we don't forget

PR #4641, opened 2023-12-14
chore(swaps): remove unused analytics events + incorrect usdTotal prop

jeanregisser:
### Description

As the title says.

### Test plan

- Tests pass

### Related issues

- Fixes RET-950

### Backwards compatibility

Yes


src/swap/SwapScreen.tsx
```diff
@@ -309,7 +309,6 @@ export function SwapScreen({ route }: Props) {
           fromTokenNetworkId: fromToken.networkId,
           amount: swapAmount[updatedField],
           amountType: swapAmountParam,
-          usdTotal: exchangeRate.swapAmount.multipliedBy(exchangeRate.price).toNumber(),
```
dievazqu, line 312: I feel we have discussed this somewhere else, but should we calculate correctly this field instead of removing it?
I'm not sure if there is any value in keeping it though. So, I think it's fine to just remove it.

dievazqu (approved):
LGTM

src/swap/SwapScreen.tsx
jeanregisser: Yes this is covered in `swap_execute_success` or `swap_execute_error`

https://github.com/valora-inc/wallet/blob/8e62e78216c2ce6bb5fac1b8db2c062474c06c19/src/analytics/Properties.tsx#L1358-L1359

which seems enough for now.

PR #4643, opened 2023-12-14
refactor(qr): Refactor the different QR screens

jh2oman:
### Description

Refactors the QR code screens to be:
* `QRNavigator` - The QR Tab bar screen that can switch between Scan and Code
* `ExchangeQRCode` - No Tab bar, just shows QR code and a bottom sheet with the list of exchanges. Only used in the cash-in flow
* `ScanQRScreen`- No Tab bar, just shows the scanner. Only used by secure send flow

### ScanQRScreen

Screenshot_20231214_114424

### ExchangeQRCode

Screenshot_20231214_114508

### QRNavigator

Screenshot_20231214_114544
Screenshot_20231214_114554

### Test plan

Went through all the flows and made sure the screens & functionality was the same


src/navigator/types.tsx
```diff
@@ -26,8 +26,8 @@ import { WalletConnectRequestType } from 'src/walletConnect/types'
 // Typed nested navigator params
 type NestedNavigatorParams<ParamList> = {
   [K in keyof ParamList]: undefined extends ParamList[K]
-    ? { screen: K; params?: ParamList[K] }
-    : { screen: K; params: ParamList[K] }
+    ? { tab: K; params?: ParamList[K] }
+    : { tab: K; params: ParamList[K] }
```
jh2oman, line 30: `screen` is really a proper noun in the wallet and these are not screens these are tabs. Using this opportunity to fix this misname

src/qrcode/components/QRTabBar.tsx
```diff
@@ -91,33 +77,14 @@ export default function QRTabBar({
   return (
     <SafeAreaView style={styles.container} edges={['top']}>
       <View style={styles.leftContainer}>
-        <TopBarIconButton
-          icon={leftIcon === 'times' ? <Times color={color} /> : <BackChevron color={color} />}
-          onPress={onPressClose}
-        />
+        <TopBarIconButton icon={<Times color={color} />} onPress={onPressClose} />
       </View>
-      {canSwitch ? (
```
jh2oman, line 99: If we can't switch then we should just not use the QRNavigator or the QRTabBar. Opted to instead create the `ScanQRScreen` to handle the case when the user shouldn't be able to switch

src/navigator/types.tsx
satish-ravi: I agree that these look more like tabs, but whenever I switch between the two, a screen change redux action is fired, so these should still be screens? Otherwise there will be a type mismatch here:

https://github.com/valora-inc/wallet/blob/765c5594d69e906ca054b9f5237e548ee4c8ee65/src/app/reducers.ts#L23

```
 DEBUG  [20:49:29.170Z] iPhone 13 redux/saga@logger {"activeScreen": "QRScanner", "type": "APP/ACTIVE_SCREEN_CHANGED"}
 DEBUG  [20:49:30.052Z] iPhone 13 ValoraAnalytics Analytics is disabled, not tracking screen QRCode
 DEBUG  [20:49:30.063Z] iPhone 13 redux/saga@logger {"activeScreen": "QRCode", "type": "APP/ACTIVE_SCREEN_CHANGED"}
```

jh2oman: ah yea missed that. I'll keep the prop name for QRNavigator as `tab` but change the type back to Screens

src/navigator/Screens.tsx
```diff
@@ -101,3 +102,5 @@ export enum Screens {
   Welcome = 'Welcome',
   WithdrawSpend = 'WithdrawSpend',
 }
+
+export enum QRTabs {}
```
satish-ravi, line 106: remove

src/qrcode/QRNavigator.tsx
```diff
@@ -153,7 +152,7 @@ function AnimatedScannerScene({ route, position }: AnimatedScannerSceneProps) {
     }
 
     Logger.debug('QRScanner', 'Bar code detected')
-    onQRCodeDetectedParam(qrCode)
+    defaultOnQRCodeDetected(qrCode)
```
satish-ravi, line 155: nit: drop the `default` naming? Seems like it's no longer overridden

src/qrcode/ScanQRScreen.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { StackParamList } from 'src/navigator/types'
+import QRScanner from 'src/qrcode/components/QRScanner'
+import { QrCode } from 'src/send/actions'
+import Colors from 'src/styles/colors'
+import fontStyles from 'src/styles/fonts'
+import Logger from 'src/utils/Logger'
+
+type Props = NativeStackScreenProps<StackParamList, Screens.ScanQRScreen>
+
+export default function ScanQRScreen({ route, navigation }: Props) {
```
satish-ravi, line 17: this screen looks broken when camera permissions are denied

src/qrcode/ExchangeQRCode.tsx
```diff
@@ -12,9 +12,9 @@ import { emptyHeader } from 'src/navigator/Headers'
 import { Screens } from 'src/navigator/Screens'
 import { TopBarIconButton } from 'src/navigator/TopBarButton'
 import { StackParamList } from 'src/navigator/types'
-import QRCode from 'src/qrcode/QRCode'
+import QRCode from 'src/qrcode/components/QRCode'
 import { SVG, shareQRCode } from 'src/send/actions'
-import { CICOFlow } from './utils'
+import { CICOFlow } from '../fiatExchanges/utils'
```
satish-ravi, line 17: nit. switch to absolute imports while you're at it

PR #4644, opened 2023-12-14
ci: fix knip regression

satish-ravi:
### Description

The knip regression script has a bug where it was actually running `knip` against the PR head twice, instead of once against PR head and once against main. So the results were always the same and no new regressions were caught ([example](https://github.com/valora-inc/wallet/pull/4643/files/b51531f174ef59a276651fcd1e03ec14e7e04daa#diff-29452a8762fc07bcefd5348c7e764862b8c13c0dfe79e8432f59095a8a90d3b1R106)). This was because the script was attempting to use branch names but the checkout action only fetches the HEAD as a merge commit and doesn't have the full history. The `git checkout main` cmd was failing silently ([example](https://github.com/valora-inc/wallet/actions/runs/7214600919/job/19657053544#step:4:720)). This PR updates it to correctly checkout the PR base and run against it.

### Test plan

CI by adding a couple of regressions (8929b18 and 073ce3b) and ensuring the job caught it: https://github.com/valora-inc/wallet/actions/runs/7215985957/job/19661309569

### Related issues

N/A

### Backwards compatibility

N/A


silasbw (approved):

jeanregisser:
oh nice! Thanks for spotting/fixing this, I was wondering too why it never caught anything.

PR #4646, opened 2023-12-15
feat: setup saga, slice and feature gate for token historical prices

MuckT:
### Description

Adds the ability to fetch token price histories from blockchain-api and store them in redux. Also adds a Statsig feature gate `use_new_price_history`. Does not include displays of the price graph and should not impact the current `src/tokens/TokenDetails.tsx` screen.

### Test plan

- [x] Tested locally on iOS with a local instance of blockchain-api.
- [x] Unit tests updated for selectors.
- [x] Unit tests added for `watchFetchTokenPriceHistory`  in `src/tokens/saga.test.ts`.

### Related issues

- Part of ACT-990

### Backwards compatibility

Yes


src/tokens/TokenDetails.tsx
```diff
@@ -56,12 +59,42 @@ const MAX_ACTION_BUTTONS = 3
...
   const { tokenId } = route.params
   const { t } = useTranslation()
+  const dispatch = useDispatch()
   const token = useTokenInfo(tokenId)
   if (!token) throw new Error(`token with id ${tokenId} not found`)
   const actions = useActions(token)
   const tokenDetailsMoreActionsBottomSheetRef = useRef<BottomSheetRefType>(null)
   const localCurrencySymbol = useSelector(getLocalCurrencySymbol)
 
+  // Price History Things - Maybe this should all move to the chart component and we just pass in the token id
```
MuckT, line 69: Was helpful for testing, but will be removed prior to taking this PR out of draft including the useEffect hook.

src/tokens/slice.ts
```diff
@@ -129,6 +143,44 @@ const slice = createSlice({
...
+    fetchPriceHistoryStart: (
+      state,
+      action: PayloadAction<{
+        tokenId: string
+        startTimestamp: number
+        endTimestamp: number
+      }>
+    ) => {
+      const { tokenId } = action.payload
+      // @ts-expect-error 'string' can't be used to index type 'WritableDraft<{}>'
```
MuckT, line 155: Not so sure about this, but since we don't have the list of tokens when initializing state I didn't know a better way. Open to any feedback or suggestions.

cajubelt: you may need to make lastDay optional, but I haven't really looked into the consequences of doing that

src/statsig/types.ts
```diff
@@ -30,6 +30,7 @@ export enum StatsigFeatureGates {
   USE_CICO_CURRENCY_BOTTOM_SHEET = 'use_cico_currency_bottom_sheet',
   SHOW_MULTICHAIN_BETA_SCREEN = 'show_multichain_beta_screen',
   SHOW_BETA_TAG = 'show_beta_tag',
+  USE_NEW_PRICE_HISTORY = 'use_new_price_history',
```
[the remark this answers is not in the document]
MuckT: Fixed in 7627e2c870cb36053b38845056e1e7e6fb4dcdee!

src/tokens/saga.ts
```diff
@@ -217,6 +223,47 @@ export function* fetchTokenBalancesSaga() {
...
+    throw new Error(
+      `Failed to fetch price history for ${tokenId}: ${response.status} ${response.statusText}`
+    )
+  }
+  return await response.json()
+}
+
+export function* fetchTokenPriceHistorySaga({
+  payload: { tokenId, startTimestamp, endTimestamp },
+}: any) {
```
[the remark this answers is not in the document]
MuckT: Fixed in 59381bf80307bd841adb0c8fe844217a639a05a8!

src/tokens/slice.ts
```diff
@@ -129,6 +138,41 @@ const slice = createSlice({
...
+        }
+      }
+    },
+    fetchPriceHistorySuccess: (
+      state: Draft<State>,
+      action: PayloadAction<{ tokenId: string; priceHistory: TokenPriceHistoryEntry[] }>
+    ) => {
+      const { tokenId, priceHistory } = action.payload
+      const token = state.tokenBalances[tokenId]
+      if (priceHistory && token?.historicalPricesUsd) {
```
cajubelt, line 164: and if there are no existing historical prices in state, the new ones get ignored? that doesn't seem right

src/tokens/slice.ts
```diff
@@ -129,6 +138,41 @@ const slice = createSlice({
...
+        token.historicalPricesUsd.priceHistory = priceHistory
+        token.historicalPricesUsd.priceHistoryStatus = 'success'
+      }
+    },
+    fetchPriceHistoryFailure: (state: Draft<State>, action: PayloadAction<{ tokenId: string }>) => {
+      const { tokenId } = action.payload
+      const token = state.tokenBalances[tokenId]
+      if (token?.historicalPricesUsd) {
```
cajubelt, line 172: same here, seems like we might want to store the error even if there isn't existing state for that token

src/tokens/saga.ts
```diff
@@ -217,6 +223,50 @@ export function* fetchTokenBalancesSaga() {
...
+    )
+  }
+  return await response.json()
+}
+
+export function* fetchTokenPriceHistorySaga({
+  payload: { tokenId, startTimestamp, endTimestamp },
+}: ReturnType<typeof fetchPriceHistoryStart>) {
+  try {
+    if (!getFeatureGate(StatsigFeatureGates.USE_PRICE_HISTORY_FROM_BLOCKCHAIN_API)) {
```
cajubelt, line 255: should we also just avoid dispatching that action if this feature gate is false?

cajubelt: where is this action dispatched anyway?

src/tokens/slice.ts
MuckT: At this point if there are no existing historical prices in the state it'll still have the priceHistoryStatus in historicalPricesUsd so it'll load prices.
```TypeScript
'celo-mainnet:native': {
  ...StoredTokenBalance
  historicalPricesUsd: {
    priceHistoryStatus: 'loading',
  }
}
```

src/tokens/saga.ts
MuckT: It is not dispatched yet. I planned to do it in the reworked component on mount.

src/tokens/saga.ts
```diff
@@ -273,7 +323,12 @@ export function* watchAccountFundedOrLiquidated() {
   }
 }
 
+export function* watchFetchTokenPriceHistory() {
+  yield* takeEvery(fetchPriceHistoryStart.type, safely(fetchTokenPriceHistorySaga))
```
jh2oman, line 327: What do you think about `takeLatest` instead? If someone switches to a different timescale for instance before the data has loaded i think we should cancel the old request and start the new one right away.

MuckT: Was thinking we'd probably fetch more data than needed then filter down in the frontend to avoid a fetching and for each timespan change, but that's a bit outside of the scope of this PR and using `takeLatest` makes sense in either case.

src/tokens/selectors.ts
```diff
@@ -490,3 +490,23 @@ export const feeCurrenciesWithPositiveBalancesSelector = createSelector(
...
+  }
+)
+
+export const tokenPriceHistoryStatusSelector = createSelector(
+  [
+    (state: RootState) => state.tokens.tokenBalances,
+    (_state: RootState, tokenId: string) => tokenId,
+  ],
+  (tokenBalances, tokenId) => {
+    return tokenBalances[tokenId]?.historicalPricesUsd?.priceHistoryStatus ?? 'idle'
```
jh2oman, line 510: If `undefined` is the exact same as `idle` then there isn't much use for the idle state. I think either make priceHistoryStatus required and set the status to `idle` on the initial data OR remove the `idle` state. I have a slight preference for making it required so its absolutely clear.

jh2oman (approved):
Approved once my priceHistoryStatus comment is addressed

src/tokens/saga.ts
MuckT: Fixed in d0265a568d2bd337a1ec85acfca25b2cd39db72b!

MuckT:
Closing to move to it's own slice and saga.

PR #4649, opened 2023-12-15
docs: account_funded and account_liquidated

cajubelt:
### Description

docs on account_funded, account_liquidated

I've been asked about these a couple times lately so I wanted to add docs for people to self-service

for those curious, the code for account funded is [here](https://github.com/valora-inc/wallet/blob/3f7b428456ea00f55e2d78c94b6c0923ea4ffab2/src/tokens/saga.ts#L245). 

I added a disclaimer for a situation where the event may not fire because it the code keeps the user's previous balance in memory, which would get wiped if the user restarts the app

### Test plan

ci

### Related issues

na

### Backwards compatibility

na


src/analytics/docs.ts
```diff
@@ -64,8 +64,8 @@ export const eventDocs: Record<AnalyticsEventType, string> = {
   [AppEvents.request_tracking_permission_started]: ``,
   [AppEvents.request_tracking_permission_declined]: ``,
   [AppEvents.request_tracking_permission_accepted]: ``,
-  [AppEvents.account_funded]: ``,
-  [AppEvents.account_liquidated]: ``,
+  [AppEvents.account_funded]: `When a user's balance has changed from below 1 cent (USD) to above. May not fire if user leaves app before funds land and restarts before re-opening. User must enter app for this to get published (like all events in this file).`,
```
kathaypacific, line 67: > May not fire if user leaves app before funds land and restarts before re-opening

i'm not sure that this is true...if a user has < $0.1 USD worth of token balance, funds their wallet, leaves the app before the funds land, then the last known token state in redux is the < $0.1 USD state. if the user restarts the app / reopens the app, the token state will start from the last known state before being refreshed so the account_funded event should be dispatched.

in writing this, i guess if a user funds their account -> leaves the app or does not open the app to see the funds land -> liquidates their wallet using a wallet other than valora -> opens valora....then the account_funded event will not fire even though technically it happened...

cajubelt: > if the user restarts the app / reopens the app, the token state will start from the last known state before being refreshed so the account_funded event should be dispatched.

I was concerned that since `prevTokenBalance` is just held in memory, if the app is restarted, it would be wiped and then initialized to whatever the current balance is *after the balance update*, making us unable to detect changes. 

Looking at the ordering of the sagas, I'm seeing `watchFetchBalance` actually comes before `watchAccountFundedOrLiquidated`. So we are relying on `watchFetchBalance` actually suspending execution (I guess at some `yield*`, to wait for an outgoing request) and not completing before `watchAccountFundedOrLiquidated`-- which also has a `yield*`-- manages to update `prevTokenBalance`. Seems like a shaky assumption. We could test it, but it seems like there might be a race condition here that would make the result inconsistent

kathaypacific: you're right in that the root saga spawns a bunch of async functions and on app start, there is a bunch of race conditions. however, it's my understanding that the redux store is always created / rehydrated before sagas are kicked off. it seems like that [here](https://github.com/valora-inc/wallet/blob/2a467d1413164d24b930e29e3704c4a68996ee25/src/redux/store.ts#L152-L162), where the store is created with the persisted reducers first before the root saga is run.

`watchFetchBalance` and `watchAccountFundedOrLiquidated` are both spawned from the root saga, but in order for the `prevTokenBalance ` variable in `watchAccountFundedOrLiquidated` saga to use the user's most up to date balance, it'd require that `fetchTokenBalancesSaga` successfully completes (and updates the redux store values) before `tokenBalance` is assigned for the first time in `watchAccountFundedOrLiquidated`. i think that it's not impossible for this to happen, but it would be highly unlikely since `fetchTokenBalancesSaga` requires a yield to a network request...

PR #4626, opened 2023-12-12
feat: add supported networks to navigation drawer

MuckT:
### Description

Adds the supported networks in the drawer navigator also some minor spacing adjustments to match the mocks.

| iOS One Network | iOS Two Networks | iOS Three Networks | iOS Three Networks Large Font |
| ----- | ----- | ----- | ----- |
|  |  |  |  |

### Test plan

- Tested locally on iOS
- Unit tests added

### Related issues

- Fixes ACT-1001

### Backwards compatibility

Yes

PR #4650, opened 2023-12-15
feat(qr-code): new copy for QR code

jh2oman:
### Description

QR code screen now has the `exchange` info when navigated to from the Cash-In flow and an InlineNotification talking about supported networks otherwise.

[Figma](https://www.figma.com/file/RsO1zYTuQZbClu9zjxOKe5/Celo-Audit%2FQR?type=design&node-id=123-51760&mode=design&t=ojzXvl6smnB1ZgRJ-0)

Screenshot_20231216_100909

### Related issues

https://linear.app/valora/issue/ACT-999/update-my-code-qr-screen-to-show-ethereum-as-supported-network


src/components/InLineNotification.tsx
```diff
@@ -116,8 +116,8 @@ const styles = StyleSheet.create({
 
 const severityColors: Record<Severity, CustomColors> = {
   [Severity.Informational]: {
-    primary: Colors.infoDark,
-    secondary: Colors.infoLight,
+    primary: Colors.black,
```
jh2oman, line 119: Checked with Kayla and Nitya, they want this style updated to all InlineNotifications

src/qrcode/QRCode.tsx
```diff
@@ -55,6 +61,28 @@ export default function QRCodeDisplay(props: Props) {
...
 
+  const getSupportedNetworks = () => {
+    const supportedNetworkIds = getDynamicConfigParams(
+      DynamicConfigs[StatsigDynamicConfigs.MULTI_CHAIN_FEATURES]
+    ).showBalances
+    const networks = supportedNetworkIds.map((id: NetworkId) => {
+      const network = networkIdToNetwork[id]
+      return network.charAt(0).toUpperCase() + network.slice(1)
+    })
+    return networks.join(', ')
```
satish-ravi, line 72: use networkNames. Maybe we could use the same logic as added in #4626

src/qrcode/QRCode.tsx
```diff
@@ -107,6 +148,19 @@ const styles = StyleSheet.create({
...
   },
+  notificationWrapper: {
+    position: 'absolute',
+    bottom: 20,
+    justifyContent: 'flex-end',
+    width: '100%',
+    paddingHorizontal: 16,
+  },
+  bold: {
+    fontWeight: 'bold',
```
satish-ravi, line 159: could we use the font style `labelSemiBoldXSmall` instead of this?

satish-ravi (commented):
Looks like there's a different icon for copy in the figma. Should that be in this PR ?

jh2oman:
Switched Copy Icon. Its not *exactly* the same, but very close. see screenshot

cajubelt (approved):

PR #4653, opened 2023-12-15
fix(send): Fix when secure send flow happens

finnian0826:
### Description

In the new send flow, any time a PN or contact was being searched and it wasn't a recent recipient, it would automatically go to the secure send flow. This fixes it so secure send flow is only entered when it is supposed to be.

### Test plan
Unit tests added.

Ran new e2e tests from [here](https://github.com/valora-inc/wallet/pull/4635) and they passed. 

Ran locally, works for PN with only one mapping:

https://github.com/valora-inc/wallet/assets/140328381/85a0fc77-2a70-40d1-b464-5fde9ab4548d

### Related issues

- N/A

### Backwards compatibility

Yes


src/send/SendSelectRecipient.tsx
```diff
@@ -232,22 +241,39 @@ function SendSelectRecipient({ route }: Props) {
...
 
   const nextScreen = (selectedRecipient: Recipient) => {
-    if (selectedRecipient.address) {
-      navigate(Screens.SendEnterAmount, {
-        isFromScan: false,
+    const addressValidationType: AddressValidationType = getAddressValidationType(
+      selectedRecipient,
+      secureSendPhoneNumberMapping
+    )
+    if (!selectedRecipient.address && addressValidationType !== AddressValidationType.NONE) {
```
satish-ravi, line 248: can we add a unit test that covers this?

satish-ravi (approved):
small update needed on description, this bug exists for both searching PN and contacts

PR #4655, opened 2023-12-18
refactor(swap): data structures simplifications

jeanregisser:
### Description

This refactors swaps to simplify a few things around the data structures used.

It contains the following changes:
- stop passing and using `rawSwapResponse`
  - instead we pass the fields that are needed
- made `estimatedPriceImpact` consistent by being always between 0 and 100
  - previously the unit was either between 0 and 1 or 0 and 100 depending on the swap event
- made the swap `amount` analytics property always decimal
  - previously it was either in token smallest unit or in decimal depending on the event
- the standby in/out values are now the same as what we display in the swap screen
  - it doesn't use the `guaranteedPrice` anymore, we should use `guaranteedSwapPriceEnabled` if we want to avoid surprises, or make it the default as @kathaypacific is proposing in [this thread](https://valora-app.slack.com/archives/C029Z1QMD7B/p1697017573119649?thread_ts=1697017142.969239&cid=C029Z1QMD7B).
- sync'ed the `SwapTransaction` type with [our API](https://github.com/valora-inc/valora-rest-api/blob/14e7ab9a1514aaf724be2b8964e05bdcc09ca434/src/swaps/providers/types.ts#L32-L53)
  - there were a bunch of fields we didn't use or were only for 0x

### Test plan

- Updated tests

### Related issues

- Fixes RET-950

### Backwards compatibility

Yes. Except some analytics properties unit were made consistent:
- `estimatedPriceImpact` now always between 0 and 100
- `amount` now always in decimal

### DIFF
--- a/src/swap/saga.test.ts
+++ b/src/swap/saga.test.ts
@@ -1,5 +1,4 @@
 import { PayloadAction } from '@reduxjs/toolkit'
-import BigNumber from 'bignumber.js'
 import { expectSaga } from 'redux-saga-test-plan'
 import * as matchers from 'redux-saga-test-plan/matchers'
 import { EffectProviders, StaticProvider } from 'redux-saga-test-plan/providers'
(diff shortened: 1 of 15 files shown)


src/swap/saga.test.ts
```diff
@@ -518,8 +453,8 @@ describe(swapSubmitSaga, () => {
         swapExecuteTxId: 'id-swap/saga-Swap/Execute',
         quoteToUserConfirmsSwapElapsedTimeInMs: 2_500,
         quoteToTransactionElapsedTimeInMs: 10_000,
-        estimatedBuyTokenUsdValue: 0.005,
-        estimatedSellTokenUsdValue: 0.01,
+        estimatedBuyTokenUsdValue: 100,
```
kathaypacific, line 456: just checking that it's expected that these values changed? i

kathaypacific (approved):
✨🚀

src/swap/saga.test.ts
jeanregisser: Yes, previously it was using the values attached to the received `SwapTransaction`, but now it uses the swap input (which depends on the received quote price).

PR #4656, opened 2023-12-18
chore(supercharge): skip fetching rewards for empty wallets

bakoushin:
### Description

To potentially reduce pressure on the backend from empty wallets this PR proposes not make a backend request for fetching available rewards, if the wallet doesn't have any token with balance greater than [TOKEN_MIN_AMOUNT](https://github.com/valora-inc/wallet/blob/d5149c7e0b8d3e948c45ddf606f657aa50c23a15/src/config.ts#L59).

Potentially it could save several thousands requests per day in peak periods like reward campaigns.

Data: [unique users with 0 tokens](https://mixpanel.com/project/2786968/view/3321744/app/insights#n4zbVy2TMznC)

Context: https://valora-app.slack.com/archives/C029Z1QMD7B/p1702630218806749

### Test plan

* Updated unit tests

### Related issues

- Fixes RET-951

### Backwards compatibility

Y

### DIFF
--- a/src/consumerIncentives/saga.test.ts
+++ b/src/consumerIncentives/saga.test.ts
@@ -28,12 +28,20 @@ import { navigateHome } from 'src/navigator/NavigationService'
 import { tokensByAddressSelector } from 'src/tokens/selectors'
 import { Actions as TransactionActions } from 'src/transactions/actions'
 import { sendTransaction } from 'src/transactions/send'
+import { NetworkId } from 'src/transactions/types'
 import { fetchWithTimeout } from 'src/utils/fetchWithTimeout'
 import { getContractKit } from 'src/web3/contracts'
 import config from 'src/web3/networkConfig'
 import { getConnectedUnlockedAccount } from 'src/web3/saga'
 import { walletAddressSelector } from 'src/web3/selectors'
-import { mockAccount, mockCeurAddress, mockCusdAddress } from 'test/values'
+import { createMockStore } from 'test/utils'
+import {
+  mockAccount,
+  mockCeurAddress,
+  mockCeurTokenId,
+  mockCusdAddress,
+  mockCusdTokenId,
+} from 'test/values'
 
 const mockBaseNonce = 10
 
@@ -105,6 +113,7 @@ describe('fetchAvailableRewardsSaga', () => {
     const uri = `${config.fetchAvailableSuperchargeRewards}?address=${userAddress}`
 
     await expectSaga(fetchAvailableRewardsSaga, fetchAvailableRewards())
+      .withState(createMockStore({}).getState())
       .provide([
         [select(phoneNumberVerifiedSelector), true],
         [select(walletAddressSelector), userAddress],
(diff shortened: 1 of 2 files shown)


src/consumerIncentives/saga.ts
```diff
@@ -152,6 +152,16 @@ export function* fetchAvailableRewardsSaga({ payload }: ReturnType<typeof fetchA
     return
   }
 
+  const supportedNetworkIds = yield* call(getSupportedNetworkIdsForTokenBalances)
+  const tokensWithTokenBalance = yield* select(tokensWithTokenBalanceSelector, supportedNetworkIds)
```
kathaypacific, line 156: supercharge is only available on Celo. afaik, we don't plan to make it available on Ethereum. or at least, if we do, it will be a good incentive for people to update their apps to participate. so i think it's more correct to only use the celo network id here? (although i think the code here does the bulk of what we need, which is preventing new wallets from fetching supercharge rewards unnecessarily)

src/consumerIncentives/saga.test.ts
```diff
@@ -161,16 +172,46 @@ describe('fetchAvailableRewardsSaga', () => {
...
         [select(walletAddressSelector), userAddress],
       ])
-      .not.call(fetchWithTimeout)
+      .not.call.fn(fetchWithTimeout)
+      .run()
+  })
+
+  it('skips fetching rewards in absence of tokens with sufficient balance', async () => {
+    await expectSaga(fetchAvailableRewardsSaga, fetchAvailableRewards())
+      .withState(
```
kathaypacific, line 186: any reason why you prefer to mock the state rather than mock the outcome of the tokens selector in the `provide` block? i guess there's not so much difference but it seems like mocking the selector outcome is easier / easier to read in this case?

kathaypacific (approved):
nice!!

src/consumerIncentives/saga.ts
bakoushin: thanks for highlighting this! sure, it should be Celo only.

src/consumerIncentives/saga.test.ts
bakoushin: after introducing this new check, I had to alter other tests to pass, given the new check.

I found it simpler to provide a default state to other tests in this suite instead of explicitly mocking the selector, because that way I don't need to partially recreate slice of state for the selector mock, just can reuse what is already there.

then for this particular case I just altered a part of the state (being consistent in relying on mocked state).

if it is harder to read and reason about, I'm happy to mock a selector instead of state for this particular case. 

@kathaypacific what do you think about other tests in the suite? would it make sense to alter them to mock the selector instead of state as well?

kathaypacific: i don't have strong feelings on this - i just noted it because i thought about it. so up to your preference, since we haven't been terribly consistent with this.

i guess there are tradeoffs with each approach. mocking the state is more cumbersome and it's harder to see which values are important for the tests (especially relying on the default state by passing in no arguments to the mock store). i prefer using providers in saga tests because it is easier to see which values are changing. you could argue that you get more robust test coverage by mocking the state, because it also exercises the selectors, so that might be a reason why you want to leave it like this.

in terms of verboseness, i think both are similar (you could add a `[select(tokensWithTokenBalanceSelector, networkConfig.defaultNetworkId]),  [mockCeloTokenBalance]],` instead of the mock store)

bakoushin: thanks for the suggestion! indeed, it's not that verbose.

I will refactor it back to mock selectors instead of store to keep the existing pattern of using selectors in this particular suite. it will also keep the suite on the "atomic" side of the atomicity-broadness tradeoff. otherwise, we actually had to get rid of other selector mocks in favor of the mocked state to be consistent at least across this one suite 😅 

thanks for highlighting this!

PR #4663, opened 2023-12-19
refactor(swap): remove unneeded state + rename variable

jeanregisser:
### Description

`react-async-hook` already keeps the resolved value for us.

Additionally, this fixes an edge case where if `refreshQuote` was called multiple times, the returned value could be incorrect (out-of-order).
This could happen because `onSuccess` is always called, even if it's not the `current` promise. 

See https://github.com/slorber/react-async-hook/blob/9351dab27b4e42623b9acdacbf3524ef220035b1/src/index.ts#L263-L275

Also renames `exchangeRate` -> `quote`.

Note: I noticed another edge case, if we get a quote and then get an error, we'll try to fetch the quote again. Not a big problem but not entirely expected. I'll fix that separately. It was already present before this PR.

### Test plan

- Tests pass
- Also tested manually

### Related issues

- Fixes RET-950

### Backwards compatibility

Yes


src/swap/useSwapQuote.ts
```diff
@@ -128,15 +127,14 @@ function useSwapQuote(networkId: NetworkId, slippagePercentage: string) {
   const walletAddress = useSelector(walletAddressSelector)
   const useGuaranteedPrice = useSelector(guaranteedSwapPriceEnabledSelector)
   const feeCurrencies = useSelector((state) => feeCurrenciesSelector(state, networkId))
-  const [exchangeRate, setExchangeRate] = useState<QuoteResult | null>(null)
```
jeanregisser, line 131: This is the main change.

The rest is mostly the rename.

src/swap/useSwapQuote.ts
```diff
@@ -204,22 +202,20 @@ function useSwapQuote(networkId: NetworkId, slippagePercentage: string) {
       return quoteResult
     },
     {
-      onSuccess: (updatedExchangeRate: QuoteResult | null) => {
-        setExchangeRate(updatedExchangeRate)
-      },
+      // Keep last result when refreshing
+      setLoading: (state) => ({ ...state, loading: true }),
       onError: (error: Error) => {
-        setExchangeRate(null)
```
kathaypacific, line 211: in the event that a user has a quote -> updates the swap pair -> the quote fetch produces an error, in this old flow the quote would be cleared but now we would continue to show the previous quote - do I read this correctly / is this intentional?

jeanregisser: The quote is cleared if there's an error.
It's only kept while loading, not if there's an error.

kathaypacific: oh i see! thanks!

kathaypacific (approved):

PR #4665, opened 2023-12-20
fix: use new send screen when enabled

MuckT:
### Description

The old send flow was being used despite the status of use_new_send_flow Statsig feature gate. This PR fixes that by checking the status in: Token Details Send, External Exchanges Send and Retry Transaction Send.

### Test plan

- [x] Tested locally on iOS.
- [x] Unit tests updated.

### Related issues

- Fixes ACT-1022

### Backwards compatibility

Yes


satish-ravi (approved):
we could wrap the feature gate fetching and choosing the appropriate screen in a function similar to this https://github.com/valora-inc/wallet/blob/c076efe29ab569210a92877649ea72c6bf7613a5/src/navigator/NavigationService.ts#L254
to avoid duplicating it in multiple places, but don't feel too strongly since we hopefully can roll this out soon and clean up the old flow

PR #4666, opened 2023-12-21
fix: prevent approval for CELO native token

kathaypacific:
### Description

I saw an approval that was stuck in pending on my tx feed. On closer inspection, it was an approval tx for CELO. 2 problems that this PR fixes:
1. don't create approval transactions for CELO
2. don't get stuck on confirming the transaction if effectiveGasPrice is null

### Test plan

My tx is no longer stuck after these changes

### Related issues

n/a

### Backwards compatibility

Y

### DIFF
--- a/src/swap/useSwapQuote.ts
+++ b/src/swap/useSwapQuote.ts
@@ -55,7 +55,7 @@ async function createBaseSwapTransactions(
 
   // If the sell token is ERC-20, we need to check the allowance and add an
   // approval transaction if necessary
-  if (allowanceTarget !== zeroAddress && fromToken.address) {
+  if (allowanceTarget !== zeroAddress && !fromToken.isNative) {
     const approvedAllowanceForSpender = await publicClient[
       networkIdToNetwork[fromToken.networkId]
     ].readContract({
--- a/src/transactions/saga.ts
+++ b/src/transactions/saga.ts
@@ -328,8 +328,13 @@ export function* handleTransactionReceiptReceived(
     Logger.error(TAG, `No information found for token ${feeCurrencyId} in network ${networkId}`)
   }
 
+  // even though effectiveGasPrice is a number in the type definition, it has
+  // come through as null before in a Celo transaction (tx hash
+  // 0x48ab4e3aeef306cc2866429017bca067367c4f905eccb2dcf8ce770c346f7b00).
+  // Setting a default value here so that the function does not fail at this
+  // point
   const gasFeeInSmallestUnit = new BigNumber(receipt.gasUsed.toString()).times(
-    new BigNumber(receipt.effectiveGasPrice.toString())
+    new BigNumber((receipt.effectiveGasPrice ?? 0).toString())
   )
 
   const baseDetails = {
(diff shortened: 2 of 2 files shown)


src/swap/useSwapQuote.ts
```diff
@@ -55,7 +55,7 @@ async function createBaseSwapTransactions(
 
   // If the sell token is ERC-20, we need to check the allowance and add an
   // approval transaction if necessary
-  if (allowanceTarget !== zeroAddress && fromToken.address) {
+  if (allowanceTarget !== zeroAddress && !fromToken.isNative) {
```
jeanregisser, line 58: The approval can be needed for CELO tokens.
It depends what the router does.

The approval is not needed for CELO when the `value` field is not 0.
But we can/should control that in our swap API instead.

jeanregisser (commented):
I don't get why the approval for CELO would be stuck.
What was the quote response?
Do you have a TX hash?

kathaypacific:
> I don't get why the approval for CELO would be stuck. What was the quote response? Do you have a TX hash?

in this case, the approval was only stuck in Valora (it was approved on chain correctly) - was stuck before the effectiveGasPrice was null in the receipt and we couldn't convert this to a big number and the function failed before "transactionConfirmed" action could be dispatched. this doesn't happen for all Celo approvals, i had some go through correctly but i thought it was maybe related. 0x48ab4e3aeef306cc2866429017bca067367c4f905eccb2dcf8ce770c346f7b00 is the tx hash where i saw null  effectiveGasPrice, i put this in the code as a comment

jeanregisser:
I see, `effectiveGasPrice` is pruned after some time (not sure exactly how many blocks though), when not using a full archive node.

https://valora-app.slack.com/archives/CNJ7KTHQU/p1698841813303359?thread_ts=1698688855.441649&cid=CNJ7KTHQU

src/swap/useSwapQuote.ts
dievazqu: Yes! I think we should return the `zeroAddress` in the allowance target when the value field is different than 0 in our backend.

src/transactions/saga.ts
```diff
@@ -328,8 +328,13 @@ export function* handleTransactionReceiptReceived(
...
   }
 
+  // even though effectiveGasPrice is a number in the type definition, it has
+  // come through as null before in a Celo transaction (tx hash
+  // 0x48ab4e3aeef306cc2866429017bca067367c4f905eccb2dcf8ce770c346f7b00).
+  // Setting a default value here so that the function does not fail at this
+  // point
   const gasFeeInSmallestUnit = new BigNumber(receipt.gasUsed.toString()).times(
-    new BigNumber(receipt.effectiveGasPrice.toString())
+    new BigNumber((receipt.effectiveGasPrice ?? 0).toString())
```
dievazqu, line 337: Hmm, this is weird.
I can see a `effectiveGasPrice` for this transaction (when querying for the receipt). I'm wondering if this value gets populated after some event.
I'm concern than this missing field indicates a bigger problem (for example, that we can't rely on the receipt data, or we should wait for some field to be present before using the receipt information).

src/swap/useSwapQuote.ts
dievazqu: Just to be clear, this change will make all non-native Celo swaps fail, since it will be missing the approval for it.

dievazqu:
> I see, `effectiveGasPrice` is pruned after some time (not sure exactly how many blocks though), when not using a full archive node.
> 
> https://valora-app.slack.com/archives/CNJ7KTHQU/p1698841813303359?thread_ts=1698688855.441649&cid=CNJ7KTHQU

Oh, that's bad... I guess it's fine for now to skip that info.
So weird that data is already pruned since the code is checking for recently minted transactions.
Should we consider using an archive node? Maybe we can "hide" it behind an API or something.

PR #4667, opened 2023-12-21
chore: modify some translations to account for supported chains

kathaypacific:
### Description

As the title. Ideally it'd be nice to say "Chain A or Chain B" but I couldn't find an easy way to do this in a predictable way for our translators on Crowdin. I wanted to avoid manually joining chains using `t('or')`, there's also the option to add a formatter in code but that seems unpredictable / opaque for our translators. So i opted for a comma separated string, which makes translations clunky but predictable :/ open for suggestions on how else to do this.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y


jeanregisser (approved):
Thanks! 👍

PR #4671, opened 2023-12-22
fix(wc): PIN / Biometrics asked too early

jeanregisser:
### Description

`getViemWallet` imports the private key and hence requires the PIN / Biometrics unlock.
Since it was called before showing the WalletConnect request, it caused an unexpected experience for users.

I think we should refactor this as it's a bit unexpected compared to what was happening with ContractKit.
In the meantime, here's an alternative.

Note: Maybe I missed something but I didn't see the loading state of the WC request, and since we now make network requests, there's no visual feedback while this happens, until the signing bottom sheet is displayed. We can fix/investigate this separately.

### Test plan

- Updated unit tests
- Manual tests:

**before**

https://github.com/valora-inc/wallet/assets/57791/7e9f618b-3e0f-4a2b-8dcf-962ee5172ab1

**after**

https://github.com/valora-inc/wallet/assets/57791/3e6870a2-fdba-48a8-abd3-7bdcd002a6a9

### Related issues

- Fixes RET-966

### Backwards compatibility

Yes

### DIFF
--- a/src/walletConnect/saga.test.ts
+++ b/src/walletConnect/saga.test.ts
@@ -2,7 +2,6 @@ import { CoreTypes, SessionTypes } from '@walletconnect/types'
 import { buildApprovedNamespaces } from '@walletconnect/utils'
 import { Web3WalletTypes } from '@walletconnect/web3wallet'
 import { expectSaga } from 'redux-saga-test-plan'
-import * as matchers from 'redux-saga-test-plan/matchers'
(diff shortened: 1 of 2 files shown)


kathaypacific (approved):
Thanks for fixing this!! i'll create a follow up task for the loading state

bakoushin (approved):
🎯

PR #4673, opened 2023-12-22
refactor(swap): use useReducer to make SwapScreen state changes easier to reason about

jeanregisser:
### Description

Switching to `useReducer` instead of many individual `useState` to clarify possible/allowed state changes.

A couple of notes:
- This should make it harder to do invalid state changes. And facilitate further changes.
  - They all need to be triggered via defined actions (modelled as events)
- I was able to remove the state for `shouldShowMaxSwapAmountWarning` and instead derive it from the value entered.
  - The only difference is the warning will now show before hitting "Confirm Swap". But I think this is good.
- `fromToken` and `toToken` are now always up-to-date (for instance in case of balance changes), as we now only store the token IDs in the state
- We could argue the reducer could be in the global redux reducer for swap, but I think this is really component state. For instance, it doesn't really make sense to trigger a global redux action every time a character is entered.
  - Also, with further changes, we could even stop going through redux for executing the swap.

See also: [Comparing useState and useReducer](https://react.dev/learn/extracting-state-logic-into-a-reducer#comparing-usestate-and-usereducer)

Next up: making it easier to understand which warning is displayed.

### Test plan

- Tests pass
- Manual tests

### Related issues

- Fixes RET-950

### Backwards compatibility

Yes


src/swap/SwapScreen.tsx
```diff
@@ -62,7 +63,126 @@ const DEFAULT_SWAP_AMOUNT: SwapAmount = {
   [Field.TO]: '',
 }
 
-type Props = NativeStackScreenProps<StackParamList, Screens.SwapScreenWithBack>
+interface SwapState {
+  fromTokenId: string | undefined
+  toTokenId: string | undefined
+  // Raw input values (can contain region specific decimal separators)
+  swapAmount: SwapAmount
+  updatedField: Field
```
bakoushin, line 71: what about `lastUpdatedField`?

src/swap/SwapScreen.tsx
```diff
@@ -62,7 +63,126 @@ const DEFAULT_SWAP_AMOUNT: SwapAmount = {
...
+      const { field, value } = action.payload
+      state.confirmingSwap = false
+      state.startedSwapId = null
+      state.updatedField = field
+      if (!value) {
+        state.swapAmount = DEFAULT_SWAP_AMOUNT
+        return
+      }
+      // Regex to match only numbers and one decimal separator
+      const sanitizedValue = value.match(/^(?:\d+[.,]?\d*|[.,]\d*|[.,])$/)?.join('')
```
bakoushin, line 107: do we really need to declare all these alternatives in regex?
probably we can simplify it a bit?
```suggestion
      const sanitizedValue = value.match(/^\d*[.,]?\d*$/)?.join('')
```

src/swap/SwapScreen.tsx
```diff
@@ -62,7 +63,126 @@ const DEFAULT_SWAP_AMOUNT: SwapAmount = {
...
+      if (!sanitizedValue) {
+        return
+      }
+      state.swapAmount[field] = sanitizedValue
+    },
+    chooseMaxFromAmount: (state, action: PayloadAction<{ fromTokenBalance: BigNumber }>) => {
```
bakoushin, line 113: do we need `From` in the name? are there benefits of being so verbose?

src/swap/SwapScreen.tsx
```diff
@@ -62,7 +63,126 @@ const DEFAULT_SWAP_AMOUNT: SwapAmount = {
...
+      if (!quote) {
+        state.swapAmount[otherField] = ''
+        return
+      }
+
+      const { decimalSeparator } = getNumberFormatSettings()
+      const parsedAmount = parseInputAmount(state.swapAmount[updatedField], decimalSeparator)
+
+      const newAmount = parsedAmount.multipliedBy(
+        new BigNumber(quote.price).pow(updatedField === Field.FROM ? 1 : -1)
```
bakoushin, line 158: why we need the trick `.pow(-1)`?

bakoushin (approved):
LGTM! 🎼

Looks easier to reason about state changes when they are grouped together.

I agree that we probably can get rid of the current swap data in the global state in the future, so it makes sense to manage the local state within this screen. 

I left some nits about naming, because I believe meaningful naming also might contribute better understanding of what's going on.

A side note: why do we show the insufficient balance warning only when user confirms the swap? Probably we can perform that check as user types and display some warning immediately once it enters an invalid value?

src/swap/SwapScreen.tsx
```diff
@@ -62,7 +63,126 @@ const DEFAULT_SWAP_AMOUNT: SwapAmount = {
   [Field.TO]: '',
 }
 
-type Props = NativeStackScreenProps<StackParamList, Screens.SwapScreenWithBack>
+interface SwapState {
+  fromTokenId: string | undefined
+  toTokenId: string | undefined
+  // Raw input values (can contain region specific decimal separators)
+  swapAmount: SwapAmount
```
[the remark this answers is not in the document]
jeanregisser: Good idea!

src/swap/SwapScreen.tsx
jeanregisser: Good idea!

src/swap/SwapScreen.tsx
jeanregisser: Yes we may be able to simplify, since this isn't directly related to the focus of this PR I'll leave it as is.

src/swap/SwapScreen.tsx
jeanregisser: I guess it doesn't hurt? mostly copied the existing naming.

src/swap/SwapScreen.tsx
jeanregisser: It's because the price doesn't mean the same thing if it's buy vs sell flow.
I'm not sure if we could have done it differently / easier to understand.
But It'll be removed soon though.

src/swap/SwapScreen.tsx
jeanregisser: Backtracking a bit on this, I started making the change but then noticed some [other structures use `updatedField`](https://github.com/valora-inc/wallet/blob/c55a5c78335565f75193a06fb42f567667cf194a/src/swap/types.ts#L23), where `lastUpdatedField` wouldn't really apply I think.
So leaving as is, to keep consistency.

PR #4674, opened 2023-12-22
fix(walletconnect): ensure there is a loading state when preparing transaction

kathaypacific:
### Description

As the title

### Test plan

https://github.com/valora-inc/wallet/assets/20150449/39f0ed1f-dcf0-48f6-a563-bd788fdc5a70

### Related issues

- Fixes RET-971

### Backwards compatibility

Y


src/walletConnect/saga.ts
```diff
@@ -428,6 +428,13 @@ function* showActionRequest(request: Web3WalletTypes.EventArguments['session_req
     return
   }
 
+  // since there are some network requests needed to prepare the transaction,
+  // add a loading state
+  navigate(Screens.WalletConnectRequest, {
```
dievazqu, line 433: Will this create a bottom sheet in top of the other?
For example, if the user is dismissing the "Action" bottom sheet, they will see the loading state?

jeanregisser (approved):
Awesome! 🚀

src/walletConnect/saga.ts
bakoushin: apparently no.
we are already doing [similar trick](https://github.com/valora-inc/wallet/blob/02ba92e639ea3169910896afcc3fd3c61b3d4e2e/src/walletConnect/walletConnect.ts#L66-L87) in session request

bakoushin (approved):
🎀

PR #4675, opened 2023-12-22
chore: cleanup use_zendesk_api_for_support experiment

jeanregisser:
### Description

As the title says.

- Allowed removing 2 dependencies:
  - `react-native-mail`
  - `react-native-email-link`
- On Android, a nice side effect is we don't need anymore to copy the logs to the external files directory.

### Test plan

- Updated tests
- Tested manually

### Related issues

- Fixes RET-972

### Backwards compatibility

Yes


ios/celo/Info.plist
```diff
@@ -53,17 +53,6 @@
...
-	<array>
-		<string>message</string>
-		<string>readdle-spark</string>
-		<string>airmail</string>
-		<string>ms-outlook</string>
-		<string>googlegmail</string>
-		<string>inbox-gmail</string>
-		<string>ymail</string>
-		<string>superhuman</string>
-	</array>
```
jeanregisser, line 66: These were added in https://github.com/celo-org/celo-monorepo/commit/69005a0144dfa52964c74b806d167ed4c7951716 for `react-native-email-link`.

src/utils/Logger.ts
```diff
@@ -163,36 +168,13 @@ class Logger {
   }
 
   getLogsToAttach = async () => {
-    const toAttach: Email['attachments'] = []
+    const toAttach: Attachments = []
     try {
       const logDir = this.getReactNativeLogsDir()
       const logFiles = await RNFS.readDir(logDir)
-      // On Android we need to move the files to a directory we have access to
```
cajubelt, line 170: is this change related?

cajubelt: like did we only do this to enable emailing logs..? I can't really tell

cajubelt (approved):
left a comment on something that might be split out, though it's hard to tell for me

src/utils/Logger.ts
jeanregisser: Yes this was only done to enable emailing logs.

PR #4676, opened 2023-12-22
chore: remove chunked file read + debug log view

jeanregisser:
### Description

Follow up from https://github.com/valora-inc/wallet/pull/4633#pullrequestreview-1780234436, the (buggy) chunked file read can now be removed. Along with the debug log view which was using it.

### Test plan

- Tests pass

### Related issues

- Fixes RET-948

### Backwards compatibility

Yes


src/app/Debug.tsx
```diff
@@ -33,9 +33,9 @@ export class Debug extends React.Component<RootState, State> {
   }
 
   updateLogs = async () => {
-    const reactNativeLogs = await Logger.getMonthLogs()
     this.setState({
-      reactNativeLogs: reactNativeLogs || 'Not Found',
+      reactNativeLogs:
+        "This feature has been removed. It didn't work with log files above a certain size.",
```
jeanregisser, line 38: I don't think anyone on the team is really using that debug screen to read logs. But let me know if you do.

cajubelt: can you just remove the feature? this seems like it leaves code hanging around that doesn't actually do anything anymore

cajubelt (changes requested):

src/app/Debug.tsx
jeanregisser: Yes, I first wanted to do this separately, but the whole log view is now removed.

kathaypacific (approved):
🧹

PR #4678, opened 2023-12-23
feat: price history in redux

MuckT:
### Description

Adds historical prices into redux with the `priceHistory` saga and slice.

### Test plan

- Unit tests added
- Tested locally with graph display with #4679
- Tested locally with Flipper by enabling priceHistory debugging.

### Related issues

- Fixes ACT-990

### Backwards compatibility

Yes


src/priceHistory/saga.ts
```diff
@@ -0,0 +1,65 @@
...
+import Logger from 'src/utils/Logger'
+import { ensureError } from 'src/utils/ensureError'
+import { fetchWithTimeout } from 'src/utils/fetchWithTimeout'
+import { safely } from 'src/utils/safely'
+import networkConfig from 'src/web3/networkConfig'
+import { call, put, spawn, takeLatest } from 'typed-redux-saga'
+
+const TAG = 'priceHistory/saga'
+
+/** @beta - Exclude from Knip dep check */
```
cajubelt, line 16: this is great! seems perfect for our use case

cajubelt, line 16:
```suggestion
/** @beta - Exclude from Knip unused code check */
```

cajubelt: might be worth noting in the docstring what the defaults are if start/end timestamps are excluded

cajubelt: but is it actually needed? seems like this function is used by the saga below it

src/priceHistory/saga.ts
```diff
@@ -0,0 +1,65 @@
...
+export async function fetchTokenPriceHistory(
```
cajubelt, line 17: probly worth adding a unit test for this

src/priceHistory/saga.ts
```diff
@@ -0,0 +1,65 @@
...
+    const prices = yield* call(fetchTokenPriceHistory, tokenId, startTimestamp, endTimestamp)
+    yield* put(fetchPriceHistorySuccess({ tokenId, prices }))
+  } catch (err) {
+    const error = ensureError(err)
+    Logger.error(TAG, 'error fetching token price history', error.message)
+    yield* put(fetchPriceHistoryFailure({ tokenId }))
+  }
+}
+
+/** @beta - Exclude from Knip dep check */
```
cajubelt, line 58: again, not clear to me why this beta tag is needed

src/priceHistory/saga.ts
```diff
@@ -0,0 +1,65 @@
...
+  return await response.json()
+}
+
+export function* fetchTokenPriceHistorySaga({
+  payload: { tokenId, startTimestamp, endTimestamp },
+}: ReturnType<typeof fetchPriceHistoryStart>) {
+  try {
+    if (!tokenId) {
+      throw new Error('TokenId is required')
+    }
```
cajubelt, line 48: do we need this? tokenId is typed as string, not string | undefined

src/priceHistory/slice.ts
```diff
@@ -0,0 +1,68 @@
...
+    fetchPriceHistorySuccess: (
+      state: Draft<State>,
+      action: PayloadAction<{
+        tokenId: string
+        prices: Price[]
+      }>
+    ) => {
+      const { tokenId, prices } = action.payload
+      const token = state[tokenId]
+      if (token) {
```
cajubelt, line 45: and what if the token isn't there? seems like it would be good to still include the updated state. I think I left a similar comment on your last PR, though I've lost the link for that so idk what the outcome of the discussion was

src/priceHistory/slice.ts
```diff
@@ -0,0 +1,68 @@
...
+    },
+    fetchPriceHistoryFailure: (
+      state: Draft<State>,
+      action: PayloadAction<{
+        tokenId: string
+      }>
+    ) => {
+      const { tokenId } = action.payload
+      const token = state[tokenId]
+      if (token) {
```
cajubelt, line 58: same here

src/priceHistory/slice.ts
```diff
@@ -0,0 +1,68 @@
...
+        token.status = 'error'
+      }
+    },
+  },
+})
```
cajubelt, line 63: agree with codecov here-- these^ should have tests

src/redux/store.ts
```diff
@@ -142,6 +142,7 @@ export const setupStore = (initialState = {}, config = persistConfig) => {
           'keylessBackup',
           'nfts',
           'swap',
+          // 'priceHistory'
```
cajubelt, line 145: can you leave some indication of when this should be added? like //TODO(ACT-XYZ): uncomment this

cajubelt: or if you have no immediate plans for uncommenting it, you can just delete this line..

src/priceHistory/slice.ts
```diff
@@ -0,0 +1,68 @@
+import { Draft, PayloadAction, createSlice } from '@reduxjs/toolkit'
+import BigNumber from 'bignumber.js'
+
+type PriceHistoryStatus = 'idle' | 'loading' | 'success' | 'error'
```
cajubelt, line 4: 'idle' isn't clear to me. What does it mean, and when does it occur? Should it be 'notSet' (assuming that's what it means)?

cajubelt: seems like it is only used in the selector to indicate that there is no price history status found for the token. we could potentially use undefined/null for that, but if we do use a string for it, I'd advocate for something a bit clearer, like 'notSet' or 'notFound'

src/priceHistory/slice.ts
MuckT: The token will be there because of the previous fetchPriceHistoryStart, but maybe it's better the handle similar to that?

```
state[tokenId] = {
  ...state[tokenId],
  status: 'success',
  prices,
}
```

src/redux/store.ts
MuckT: This won't be added it's just for local debugging similar to `tokens` and `exchange`.

src/priceHistory/saga.ts
MuckT: It got flagged here for an unused export: https://github.com/valora-inc/wallet/actions/runs/7306296211/job/19910864603.

src/priceHistory/slice.ts
MuckT: removed in b6428a8ce1aa3b4bed5bd6948d4caf8aeb6f4f9c!

src/priceHistory/slice.ts
MuckT: Fixed in b1a3b4499954a6839f866a02e1782668d06c34c5!

src/priceHistory/saga.ts
MuckT: Removed this check in 4b2c2c81fa9ac4d36af3849c947f86b4138da9cf!

src/priceHistory/saga.ts
MuckT: Knip flagged it 🤷. Will try removing it after the tests are updated to see if it's still needed.

src/priceHistory/saga.ts
MuckT: Added in ddff343278ceb7d786ffb4ba5ca36344bedf9019!

src/priceHistory/slice.ts
MuckT: added in b7454fdfb4264dabe34f01afc8a6c35168569821!

src/priceHistory/saga.ts
MuckT: added in 47f933ccf89a32ede28d88f650e2e80743e876ad!

src/redux/store.ts
cajubelt: then please delete it

src/priceHistory/saga.ts
cajubelt: is it still flagged now that there is a test?

cajubelt: in general, if an export is unused and you don't actually plan to use it outside the file, seems like you should just not export it, instead of adding a beta tag to get knip to ignore it. seems like we should use beta tags for beta features, like things that we plan to use but want to build out in a separate PR

src/priceHistory/saga.ts
```diff
@@ -55,7 +57,7 @@ export function* fetchTokenPriceHistorySaga({
   }
 }
 
-/** @beta - Exclude from Knip dep check */
+/** @beta - Exclude from Knip unused exports */
```
cajubelt, line 60: if you never plan to actually use this outside the file, why not just remove the "export"?

src/priceHistory/saga.ts
```diff
@@ -13,7 +13,12 @@ import { call, put, spawn, takeLatest } from 'typed-redux-saga'
 
 const TAG = 'priceHistory/saga'
 
-/** @beta - Exclude from Knip dep check */
+/** @beta - Exclude from Knip unused exports
```
cajubelt, line 16: is this still needed?

src/priceHistory/selectors.ts
```diff
@@ -0,0 +1,16 @@
...
+  [(state: RootState) => state.priceHistory, (_state: RootState, tokenId: string) => tokenId],
+  (priceHistory, tokenId) => {
+    return priceHistory[tokenId]?.prices ?? []
+  }
+)
+
+export const priceHistoryStatusSelector = createSelector(
+  [(state: RootState) => state.priceHistory, (_state: RootState, tokenId: string) => tokenId],
+  (priceHistory, tokenId) => {
+    return priceHistory[tokenId]?.status ?? 'idle'
```
cajubelt, line 14: seems like this should return a valid status, not 'idle', now that you've removed that status

cajubelt: probly worth adding an explicit return type for this selector to avoid issues like this moving forward

src/redux/store.ts
cajubelt: if you feel strongly about keeping it please at least add a code comment explaining why it's there, like 
```
// the following can be uncommented for debugging:
// 'priceHistory',
// 'tokens',
// 'exchange',
```

cajubelt: otherwise it's not clear why it's there (was it used previously but commented out to *solve* a bug? is it something you meant to build on? or still plan to do something with?)

src/priceHistory/selectors.ts
MuckT: Fixed in 87377019f32ae067b8cbe445ae83677a4614236e & 9d59b1c94036c018749ced024cf12f5cfcf91213!

src/priceHistory/saga.ts
MuckT: Removed in e988643a3f991108c7f0af6dc87381c2840a667a!

src/priceHistory/saga.ts
MuckT: Removed in e988643a3f991108c7f0af6dc87381c2840a667a!

src/priceHistory/saga.ts
MuckT: The function is no longer flagged by knip after being used in a test.

src/redux/store.ts
MuckT: Left more detailed comments in f83c36a6029bbbef1b4f3de05bd54772cf26e2fd!

src/priceHistory/saga.ts
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Fetches the price history for a token from blockchain-api
+ * @param {string} tokenId - The tokenId to fetch the price history for
+ * @param {number} startTimestamp - The timestamp to start fetching prices from defaults to 30 days ago
+ * @param {number} endTimestamp - The timestamp to stop fetching prices at defaults to now
+ */
+export async function fetchTokenPriceHistory(
+  tokenId: string,
+  startTimestamp?: number,
+  endTimestamp?: number
```
jh2oman, line 25: Why make these optional? It doesn't appear that this function is ever called without these fields and opens the opportunity for a nasty bug with the function not differentiating between 0 and undefined since they are both falsy

jh2oman (approved):

src/priceHistory/saga.ts
MuckT: Fixed in 17d1cbcb9fc7779f7a75612e15f37227467df39c!

PR #4679, opened 2023-12-26
feat: price history chart display

MuckT:
### Description

Uses the new slice and saga created in #4678 to populate a price chart that is displayed on the token details. If the `use_price_history_from_blockchain_api` feature gate is enabled. A sizable amount of chart display logic was copied from `src/exchange/CeloGoldHistoryChart.tsx` and that existing component was marked as deprecated and will be replaced by `src/priceHistory/PriceHistoryChart.tsx` in the future.

### Test plan

- Tested locally on iOS
- Tested locally on Android
- Unit tests added

### Screenshots

| Celo iOS | Eth iOS | Celo Android | Eth Android |
|  |  |  |  

### Load Videos

- There's a can bit a bit of strange on layout rendering that moves the prices on the x axis if they would be rendered in the margins, but otherwise the load is WAI.

#### Load on Celo News Page
https://github.com/valora-inc/wallet/assets/26950305/9cad3a2e-286e-4e1c-a554-6c022bc91d93

#### Eth Load and Celo Price Token Details
https://github.com/valora-inc/wallet/assets/26950305/f928a249-ce5d-4ea0-b869-006d88e5214b

### Related issues

- Fixes ACT-99

### Backwards compatibility

Yes


src/priceHistory/priceHistoryChart.tsx
```diff
@@ -0,0 +1,321 @@
...
+      lastTimestampAdded = priceHistoryPrices[i].priceFetchedAt
+      const { priceUsd } = priceHistoryPrices[i]
+      const localAmount = dollarsToLocal(priceUsd)
+      chartData.push({
+        amount: localAmount ? localAmount.toNumber() : 0,
+        displayValue: localAmount ? displayLocalCurrency(localAmount) : '',
+        priceFetchedAt: priceHistoryPrices[i].priceFetchedAt,
+      })
+    }
+  }
```
MuckT, line 250: This is needed as the biggest slow down on page renders was loading all the data points into the chart and this cut down render times significantly. We could move this logic to blockchain-api if we wanted, but handling it on the frontend didn't appear to be a bottleneck.

src/priceHistory/__snapshots__/priceHistoryChart.test.tsx.snap
MuckT, line 1: Used snapshots like the existing test, but open to trying `data-testid` as specified in https://github.com/FormidableLabs/victory/pull/2192.

src/tokens/TokenDetails.tsx
```diff
@@ -68,6 +69,9 @@ export default function TokenDetailsScreen({ route }: Props) {
   const actions = useActions(token)
   const tokenDetailsMoreActionsBottomSheetRef = useRef<BottomSheetRefType>(null)
   const localCurrencySymbol = useSelector(getLocalCurrencySymbol)
+  const usePriceHistoryFromBlockchainApi = getFeatureGate(
```
MuckT, line 72: Add your wallet to `use_price_history_from_blockchain_api` in statsig to test locally.

src/priceHistory/priceHistoryChart.tsx
```diff
@@ -0,0 +1,356 @@
...
+  }, [])
+
+  if (priceHistoryStatus === 'loading' && priceHistoryPrices.length === 0) {
+    return <Loader />
+  } else if (priceHistoryPrices.length === 0) {
+    return null
+  }
+
+  // Create chart data from price history
+  const chartData = []
```
MuckT, line 231: This section creating the chartData is largely new instead of relying on the `aggregateExchangeRates` present in `src/exchange/reducer.ts`. This is preferable as if we ever have data issue we can populate the fixed the data from the API and if the user is unable to fetch new prices we will use the prices already present in redux.

src/tokens/TokenDetails.tsx
```diff
@@ -92,14 +96,25 @@ export default function TokenDetailsScreen({ route }: Props) {
...
-        {token.isNative && token.symbol === 'CELO' && (
-          <CeloGoldHistoryChart
-            color={Colors.black}
-            containerStyle={styles.chartContainer}
-            chartPadding={Spacing.Thick24}
-            testID="TokenDetails/Chart"
-          />
-        )}
+        {usePriceHistoryFromBlockchainApi
+          ? !token.isStableCoin && (
```
satish-ravi, line 100: do we want to show the chart for all non stable coin tokens? I think the original [design](https://www.figma.com/file/OE6RFzqXAMqU9ygHtNYyof/Assets-Page-(Multichain)?node-id=2099%3A17366&mode=dev) only included charts for native tokens. I don't think it hurts showing the chart for everything except stables, but would be good to get product confirmation.

src/priceHistory/priceHistoryChart.tsx
```diff
@@ -0,0 +1,357 @@
...
+}
+
+export default function PriceHistoryChart({
+  tokenId,
+  containerStyle,
+  testID,
+  chartPadding,
+  color = colors.black,
+}: PriceHistoryChartProps) {
+  const priceHistoryPrices = useSelector((state: RootState) =>
```
satish-ravi, line 187: nit: `historicalPrices` over `priceHistoryPrices`? same for the selector

src/priceHistory/priceHistoryChart.tsx
```diff
@@ -0,0 +1,357 @@
...
+      highPrice = priceHistoryPrices[i]
+    }
+    if (!lowPrice || priceHistoryPrices[i].priceUsd < lowPrice.priceUsd) {
+      lowPrice = priceHistoryPrices[i]
+    }
+
+    // Only grab one price per chart step & the most recent price
+    if (
+      lastTimestampAdded &&
+      priceHistoryPrices[i].priceFetchedAt - lastTimestampAdded <
```
satish-ravi, line 245: this assumes that the data is sorted by priceFetchedAt right?

src/priceHistory/priceHistoryChart.tsx
```diff
@@ -0,0 +1,357 @@
...
+      priceFetchedAt: highPrice.priceFetchedAt,
+    })
+  }
+  if (lowPrice) {
+    chartData.push({
+      amount: dollarsToLocal(lowPrice.priceUsd)?.toNumber() || 0,
+      displayValue: displayLocalCurrency(dollarsToLocal(lowPrice.priceUsd) || 0),
+      priceFetchedAt: lowPrice.priceFetchedAt,
+    })
+  }
```
satish-ravi, line 276: is it strictly necessary to include these? I guess these could cause uneven data points and I assume the sorting and unique steps below are only required because of adding these..

src/priceHistory/priceHistoryChart.tsx
```diff
@@ -0,0 +1,357 @@
...
+      <VictoryGroup
+        domainPadding={CHART_DOMAIN_PADDING}
+        singleQuadrantDomainPadding={false}
+        padding={{ left: chartPadding, right: chartPadding }}
+        width={CHART_WIDTH}
+        height={CHART_HEIGHT}
+        data={uniqueChartData.map((el) => el.amount)}
+        domain={domain}
+      >
+        {/* @ts-ignore */}
```
satish-ravi, line 312: could the `@ts-ignore`'s be avoided? If not, could we document why this is required and maybe use `@ts-expect-error` instead?

src/priceHistory/priceHistoryChart.tsx
satish-ravi: this seems a bit dense, should this be in its own function with unit tests?

satish-ravi (commented):
Should the Celo News page also be updated to the blockchain-api data based on the feature gate so it's consistent with the token details page?

src/priceHistory/priceHistoryChart.tsx
MuckT: With the filtering based on step CHART_STEP_IN_HOURS it's very possible if we have a lot of data points we will exclude the highest and lowest prices. The sorting is needed to get this into the right place, but I didn't notice much of a performance hit sorting the filtered list. Displaying the unfiltered list was a much bigger bottleneck.

src/priceHistory/priceHistoryChart.tsx
MuckT: Yes which it should be from blockchain API. We could sort before to be sure, but I haven't tried to see how long this would take on the larger datasets.

src/tokens/TokenDetails.test.tsx
```diff
@@ -177,7 +177,8 @@ describe('TokenDetails', () => {
     expect(queryByText('tokenDetails.priceUnavailable')).toBeFalsy()
   })
 
-  it('renders chart if token is native (celo)', () => {
+  it('renders chart if token is native (celo) from', () => {
```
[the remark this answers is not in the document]
MuckT: Fixed this and added some more tests in 231f0a4d9c77ae7b5fe82333612fc6e5097c4ef3 and 825cf6562a390ea4c86f91ffdb1b38d61793ba30

src/tokens/TokenDetails.tsx
MuckT: Handling tokens with a price < $0.01 does at a bit of complexity, but it looks like we could adjust `getLocalCurrencyDisplayValue` to take an optional param to resolve this also. We just need to derive what number of significant digits we need based on the high and low prices.

Simulator Screenshot - iPhone 15 Pro Max - 2024-01-03 at 18 17 56

kathaypacific:
_a wild kathy appeared_ sorry for the out of the blue question - i'm currently working on adding 100 more ethereum tokens and it made me think of how much extra token pricing data we will have. i just wanted to make sure that we're fetching price history on demand, rather than storing everything in redux by default?  i'm worried about the volume of data for lower end devices. i had a quick look through the code but didn't find what i was looking for so just wanted to call it out in case :)

also super nit, can we please uppercase the price history chart file name? 🙏🏻

MuckT:
@kathaypacific we are storing the price history in redux that was merged in this PR: https://github.com/valora-inc/wallet/pull/4678 there's also a requirement in ACT-990 that partially calls for this. If we want to store less in redux we could do one of a few things: move the step filtering to the api, resulting in less data to store in redux; revert #4678 and just fetch the data for each token, or limit the number of tokens we want to support with the graph.

kathaypacific:
> @kathaypacific we are storing the price history in redux that was merged in this PR: #4678 there's also a requirement in [ACT-990](https://linear.app/valora/issue/ACT-990) that partially calls for this. If we want to store less in redux we could do one of a few things: move the step filtering to the api, resulting in less data to store in redux; revert #4678 and just fetch the data for each token; or limit the number of tokens we want to support with the graph.
> 
> Made is so we are only using this for Native tokens in [50f67ab](https://github.com/valora-inc/wallet/commit/50f67abdd713c1351d8a12c8e463268a782da1e2). Later if we want to use this for more tokens removing the redux state and simply fetching should be relatively straight forward as the `PriceHistoryChart.tsx` just takes a `tokenId` and some styling props.

Thanks for looking into this :) I defer to your / the squad's judgement. My personal preference would be to avoid storing more data than we need because it seems unsustainable to store historical data for every token that we support for every time range that we support - we could support many chains in the future, hundreds / thousands of tokens, and maybe allow price history up to a year. Storing too much data could result in a slower app performance and this is the type of slow degraded experience that users are not likely to reach out to support for but just stop using the app. I don't really see a huge need to have the price chart load immediately, that benefit seems too small for the potential risks...

src/priceHistory/PriceHistoryChart.tsx
```diff
@@ -0,0 +1,368 @@
...
+        return null
+      case 1:
+        return result[0]
+      default:
+        return <>{result}</>
+    }
+  }
+}
+
+export function createChartData(
```
MuckT, line 172: Exported for testing - wip.

src/priceHistory/priceHistoryChart.tsx
MuckT: Ended up going with `prices` and `status` in b52551b3dd85ab7fd661e85bd76ecf3eba8c5b0e.

src/priceHistory/priceHistoryChart.tsx
MuckT: It looks like it's needed as we dynamically create the datum. Switched to `@ts-expect-error` and added a note in 5f2ba7b5069f4e4da2fb07bf176179e24f1fe956.

src/exchange/ExchangeHomeScreen.test.tsx
```diff
@@ -40,6 +41,18 @@ describe('ExchangeHomeScreen', () => {
     expect(getByTestId('PriceChart')).toBeTruthy()
   })
 
+  it('renders the price chart using PriceHistoryChart and blockchain api when enabled', async () => {
+    jest.mocked(getFeatureGate).mockReturnValue(true)
```
satish-ravi, line 45: nit: can we reset the mock in a beforeEach or explicitly set in other tests so it doesn't affect them and only applies to this test?

src/exchange/ExchangeHomeScreen.tsx
```diff
@@ -150,7 +157,17 @@ function ExchangeHomeScreen() {
             </View>
           </View>
 
-          <CeloGoldHistoryChart testID="PriceChart" />
+          {usePriceHistoryFromBlockchainApi && tokensBySymbol['CELO']?.tokenId ? (
```
satish-ravi, line 160: nit: can we just use `networkConfig.celoTokenId` instead of extracting this from tokensBySymbol?

src/tokens/TokenDetails.tsx
```diff
@@ -92,13 +96,23 @@ export default function TokenDetailsScreen({ route }: Props) {
...
+          <PriceHistoryChart
+            tokenId={tokenId}
             containerStyle={styles.chartContainer}
             chartPadding={Spacing.Thick24}
-            testID="TokenDetails/Chart"
+            testID={`TokenDetails/Chart/${tokenId}`}
+            color={Colors.black}
           />
+        ) : (
+          token.symbol === 'CELO' && (
```
satish-ravi, line 108: we can also switch this to use `networkConfig.celoTokenId`

src/priceHistory/PriceHistoryChart.tsx
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    dispatch(
+      fetchPriceHistoryStart({
+        tokenId,
+        startTimestamp: Date.now() - ONE_DAY_IN_MILLIS * 30,
+        endTimestamp: Date.now(),
+      })
+    )
+  }, [])
```
satish-ravi, line 279: should we include tokenId and prices in the dependency array?

src/priceHistory/PriceHistoryChart.tsx
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    switch (result.length) {
+      case 0:
+        return null
+      case 1:
+        return result[0]
+      default:
+        return <>{result}</>
+    }
```
satish-ravi, line 168: can these be condensed into: 
```suggestion
    return (
      <>
        <G key={idx + 'dot'}>
          {idx === 0 && (
            <Line x1={0} y1={y} x2={chartWidth} y2={y} stroke={colors.gray2} strokeWidth="1" />
          )}
          <Circle cx={x} cy={y} r="4" fill={color} />
        </G>
        {[highestRateIdx, lowestRateIdx].includes(idx) && (
          <ChartAwareSvgText
            x={x}
            y={y}
            key={idx}
            value={chartData[highestRateIdx].displayValue}
            position={idx === highestRateIdx ? 'top' : 'bottom'}
            chartWidth={chartWidth}
          />
        )}
      </>
    )
```

src/tokens/TokenDetails.tsx
MuckT: Fixed in c138a17891cf98987f197d691a48ea1e5890a04e!

src/exchange/ExchangeHomeScreen.tsx
MuckT: Fixed in c138a17891cf98987f197d691a48ea1e5890a04e!

src/exchange/ExchangeHomeScreen.test.tsx
MuckT: Fixed in 73561642b875cb292b7d136e1f82b5acae66c9b1!

src/priceHistory/PriceHistoryChart.tsx
MuckT: Fixed in bd1989815b43713eb2834cad17ec59de6d2abacf!

src/priceHistory/PriceHistoryChart.tsx
MuckT: Went with a slightly modified version in f3829090717396e4bf8865f8fb223976f2362d75, that appears to be working well.

satish-ravi (approved):

src/priceHistory/PriceHistoryChart.tsx
MuckT: Actually looking at the snapshots I see empty  `<g />` elements added with this method.

PR #4683, opened 2024-01-03
chore(analytics): add super properties for token balance

satish-ravi:
### Description

Add super properties for whether a user has balance for any supported token per chain and overall.

### Test plan

Unit tests

### Related issues

- Fixes ACT-1011

### Backwards compatibility

Yes


src/analytics/selectors.ts
```diff
@@ -116,6 +127,16 @@ export const getCurrentUserTraits = createSelector(
...
     }
 
+    const hasTokenBalanceFields: Record<string, boolean> = {
+      hasTokenBalance: tokensWithBalance.length > 0,
+      ...Object.fromEntries(
+        networkIds.map((networkId) => [
+          `has${toPascalCase(networkId)}TokenBalance`,
+          tokensWithBalance.some((token) => token.networkId === networkId),
+        ])
+      ),
```
satish-ravi, line 137: I was going back and forth with one property per network vs a single property listing all network ids with balances (e.g., `networkIdsWithBalance: celo-mainnet, ethereum-mainnet`) but went with the former as it seems easier to handle booleans over csv strings in analytics tools, and also aligns with the chain segmentation for `totalBalanceUsd` where we will need one prop per chain. But could be convinced otherwise.

src/analytics/selectors.ts
```diff
@@ -16,13 +17,23 @@ import {
...
+import { NetworkId } from 'src/transactions/types'
 import { mtwAddressSelector, rawWalletAddressSelector } from 'src/web3/selectors'
 
+function toPascalCase(str: string) {
+  const camelCaseStr = camelCase(str)
+  return `${camelCaseStr.charAt(0).toUpperCase()}${camelCaseStr.slice(1)}`
+}
+
 const tokensSelector = createSelector(
-  [tokensWithTokenBalanceAndAddressSelector, coreTokensSelector],
```
satish-ravi, line 24: these selectors are deprecated, replaced with selectors that support multichain

src/analytics/selectors.ts
jh2oman: I think i prefer this way too for ease of working with in analytics tools

jh2oman (approved):

PR #4684, opened 2024-01-03
chore(knip): make it easier to spot regressions with colors

jeanregisser:
### Description

It was a bit hard to spot the exact knip regression(s).
This helps with colors and a diff.

### Test plan

| Before | After |
|--------|-------|
| Screenshot 2024-01-03 at 18 31 35 | Screenshot 2024-01-03 at 18 30 40 |

### Related issues

Fixes ACT-1038

### Backwards compatibility

Yes


silasbw (approved):

.github/scripts/checkKnipRegression.ts
```diff
@@ -25,31 +29,43 @@ function compareKnipResults(
...
-  console.table(rows)
+
+  // Use easy-table instead of console.table, because console.table doesn't support colors (it escapes them)
+  console.log(easyTable.print(rows))
   return shouldFail
 }
 
 if (process.env.GITHUB_EVENT_NAME === 'pull_request') {
-  const branchKnipOutput = $.exec('yarn knip --no-gitignore').stdout.trim()
+  const branchKnipOutput = stripAnsi($.exec('yarn knip --no-gitignore').stdout.trim())
```
satish-ravi, line 49: Wonder if we should silence the entire knip output (with a silent: true option) and only print the diff. I've found this noisy and there's a ton of duplicates with both the PR branch and base outputs.

.github/scripts/checkKnipRegression.ts
```diff
@@ -25,31 +29,43 @@ function compareKnipResults(
...
   if (compareKnipResults(baseKnipResults, branchKnipResults)) {
     console.log('Knip check failed. PR branch reported more problems than base branch.')
+
+    // Print diff of knip output, for easily seeing what changed
+    console.log(
+      diff(baseKnipOutput, branchKnipOutput, {
+        // limit the number of common lines to print
+        expand: false,
+      })
+    )
```
satish-ravi, line 67: I think this a good enough for our purposes for now but wanted to flag a couple of things:

1. For some categories, the output includes whitespaces based on the length of the filename and unused exported item, so if the item with max length changes between the base and PR branches, we would have all lines within the category reported as a diff. One option could be to strip contiguous whitespaces for those categories.
2. If a PR removes some unused code in one category, but adds in another category, the diff would include the former as well. Ideally we'd want only the latter reported in the diff. 

I attempted using the JSON format for knip output and get a set difference, which should address the above, but for some reason the output was truncated by shelljs.

package.json
```diff
@@ -283,6 +286,7 @@
     "redux-saga-test-plan": "^4.0.6",
     "remote-redux-devtools": "^0.5.12",
     "shelljs": "^0.8.5",
+    "strip-ansi": "^6.0.0",
```
satish-ravi, line 289: should this go in a renovate ignore since we can't use v7 yet? I guess we could also let renovate open a PR for v7 and then close it and a new one won't be opened until there's a v8.

satish-ravi (approved):
Nice! 🎉

.github/scripts/checkKnipRegression.ts
jeanregisser: Oh that’s super helpful!
Is this the shelljs bug you ran into?
https://github.com/shelljs/shelljs/issues/1005

maybe we could try the suggestion? or redirect the output to a temp file we would read afterward?

satish-ravi: It looks similar yes, but I was seeing different truncations in my local and CI, here's a CI run with truncation https://github.com/valora-inc/wallet/actions/runs/7295837665/job/19882780114. I recall trying both maxBuffer and redirecting the output to a file and saw the same result, but haven't tried on CI

.github/scripts/checkKnipRegression.ts
jeanregisser: Yes it is a noisy, but maybe we can do that separately if end up using the json reporter?
Now that we have a summary of the issues having the full output further above can maybe help debug/ensure things are working as expected?

package.json
jeanregisser: Yes letting renovate try it is fine probably.

PR #4685, opened 2024-01-03
chore(analytics): Chain segmentation for totalBalanceUsd

finnian0826:
### Description

`totalBalanceUsd` super property was updated to include ETH in #4683 as a result of updating a selector. This PR adds in super properties for totalBalanceUsd for each network ID.

### Test plan

Unit tests updated.

Example events with super properties: https://mixpanel.com/project/2786968/view/3321744/app/events#WRxJVxJPyHqb

### Related issues

- Fixes #[ACT-1010]

### Backwards compatibility

Yes, just adding new properties.


satish-ravi (approved):
I'd just move the test plan to #4687 since that applies more closely to the changes there than this PR and this can just be tested on any event. Also, can you link to example events including these super props?

finnian0826:
> I'd just move the test plan to #4687 since that applies more closely to the changes there than this PR and this can just be tested on any event. Also, can you link to example events including these super props?

yep just added a link to events that have the super props in the test plan.

PR #4686, opened 2024-01-03
fix: ensure transfer details have fee field populated

kathaypacific:
### Description

This PR updates the LegacyFeeDrawer to use tokenId and TokenDisplay so that the payment sent transaction details screen has the fees field populated correctly. Previously this field did not work for transactions on Ethereum because the gas fee token does not have an address.

Note that this PR just makes the screen work - there will probably be some necessary follow up changes following this discussion https://valora-app.slack.com/archives/C0684TXDR3K/p1704303258842509

### Test plan

Simulator Screenshot - iPhone 14 Pro - 2024-01-03 at 18 29 18

### Related issues

- Fixes RET-960

### Backwards compatibility

Y


src/send/SendConfirmation.test.tsx
```diff
@@ -198,8 +198,7 @@ describe('SendConfirmation', () => {
       jest.runAllTimers()
     })
 
-    const feeComponent = getByTestId('feeDrawer/SendConfirmation/totalFee/value')
-    expect(getElementText(feeComponent)).toEqual('₱0.0266')
+    expect(getByTestId('feeDrawer/SendConfirmation/totalFee')).toHaveTextContent('₱0.027')
```
jeanregisser, line 201: Do you know why the value seems rounded now?

src/components/LegacyFeeDrawer.tsx
jeanregisser, line 1: Should we avoid updating this legacy component and instead update the now broken screens to use `FeeDrawer`?

kathaypacific: i thought about this, but FeeDrawer is currently not being used anywhere. i wasn't sure if it worked well and i'm also not sure that this UI is going to stick around so i went with the easy approach of tweaking the existing component. this makes it easier to delete the FeeDrawer, what do you think?

jeanregisser (approved):
🚀

src/components/LegacyFeeDrawer.tsx
kathaypacific: Looking through the git history, I suspect that it was used at some point but its consumers were removed and now this is a redundant component. (It was not created to replace legacy stuff as part of the multichain work). I think I'll create a separate PR to remove this!

PR #4687, opened 2024-01-03
chore(analytics): Include multichain assets for account_funded/liquidated event

finnian0826:
### Description

Include multichain assets when deciding when to trigger account funded/liquidated analytics events.

### Test plan

Updated / added unit tests

Manual testing:

Be on the account funded/liquidated build on main
Send ETH to that wallet address
Check to make sure account funded event happens
Send CELO to that wallet address
Switch over to totalBalanceUsd build on main
Refresh transaction feed to trigger fetch_balances event
Check that totalCeloMainnetBalanceUsd and totalEthereumMainnetBalanceUsd are both > 0 on those events
Switch back to account funded/liquidated build.
Liquidate all CELO
Check that account liquidated event didn't happen (in the old case it would have because ETH not tracked, but here it should not since still have ETH)
Liquidate all ETH
Check that account liquidated event did happen
To check that events happened and super properties were included I was able to check the logs and also created a mixpanel dashboard to confirm.

This testing confirms that the super property shows up, and also checks that an account can be funded with ETH and the right event will be triggered, and liquidating all CELO assets does not trigger liquidated event, but liquidating ALL assets does trigger.

### Related issues

- Fixes #[ACT-1010]

### Backwards compatibility

Yes


src/tokens/saga.ts
```diff
@@ -248,8 +248,10 @@ export function* watchAccountFundedOrLiquidated() {
     // we reset the usd value of all token balances to 0 if the exchange rate is
     // stale, so it is okay to use stale token prices to monitor the account
     // funded / liquidated status in this case
+    const supportedNetworkIds = getSupportedNetworkIdsForTokenBalances()
```
satish-ravi, line 251: is there a chance that this can be invoked before statsig is initialized? If so, one scenario I can think of is, let's say a user has only ETH assets. And if this is invoked before statsig is initialized, we'd end up checking only celo assets (the default value for the dynamic config) and potentially fire an account_liquidated event.

finnian0826: @satish-ravi hmm yeah I'm not sure, do you know when statsig is initialized vs. when saga stuff starts tracking? I'm still learning about how all of the saga/event stuff works.

satish-ravi: I'd start at the rootSaga in src/redux/sagas.ts. It looks like all other sagas are invoked after appInit, which is what initializes statsig, but would be good to confirm by adding some logs on the network Ids used and making sure the scenario doesn't happen

finnian0826: sounds good I can test out locally and make sure that both network IDs are always included (for my case where I am in the multichain beta)

finnian0826: > is there a chance that this can be invoked before statsig is initialized? If so, one scenario I can think of is, let's say a user has only ETH assets. And if this is invoked before statsig is initialized, we'd end up checking only celo assets (the default value for the dynamic config) and potentially fire an account_liquidated event.

@satish-ravi I tested locally and it seems like statsig was always initialized before the funded/liquidated functions were invoked, so should be okay.

src/tokens/saga.ts
```diff
@@ -259,14 +262,23 @@ export function* watchAccountFundedOrLiquidated() {
         const isAccountFundedBefore = prevTokenBalance?.gt(DOLLAR_MIN_AMOUNT_ACCOUNT_FUNDED)
         const isAccountFundedAfter = tokenBalance?.gt(DOLLAR_MIN_AMOUNT_ACCOUNT_FUNDED)
 
-        if (isAccountFundedBefore && !isAccountFundedAfter) {
+        if (
+          isAccountFundedBefore &&
+          !isAccountFundedAfter &&
+          prevNetworkIds.every((value) => supportedNetworkIds.includes(value))
```
satish-ravi, line 268: can you add a comment explaining this (not super obvious) and add unit tests for the different scenarios? Also consider using sets over arrays.

src/tokens/saga.test.ts
```diff
@@ -252,9 +256,65 @@ describe('watchAccountFundedOrLiquidated', () => {
...
+  it('does not dispatch the account funded event when network ID added', async () => {
+    jest
+      .mocked(getDynamicConfigParams)
+      .mockReturnValueOnce({ showBalances: ['celo-alfajores'] })
+      .mockReturnValueOnce({ showBalances: ['celo-alfajores', 'ethereum-sepolia'] })
+    await expectSaga(watchAccountFundedOrLiquidated)
+      .provide([
+        [
+          select(lastKnownTokenBalancesSelector, [NetworkId['celo-alfajores']]),
+          dynamic(balances(new BigNumber(0), new BigNumber(0))),
```
satish-ravi, line 283: one static return value should be sufficient here right?

src/tokens/saga.test.ts
```diff
@@ -252,9 +256,65 @@ describe('watchAccountFundedOrLiquidated', () => {
...
+        ],
+        [
+          select(lastKnownTokenBalancesSelector, [
+            NetworkId['celo-alfajores'],
+            NetworkId['ethereum-sepolia'],
+          ]),
+          dynamic(balances(new BigNumber(10), new BigNumber(10))),
```
satish-ravi, line 290: same question as above

src/tokens/saga.test.ts
```diff
@@ -252,9 +256,65 @@ describe('watchAccountFundedOrLiquidated', () => {
...
```
satish-ravi, line 316: same question as above

src/tokens/saga.test.ts
```diff
@@ -252,9 +256,65 @@ describe('watchAccountFundedOrLiquidated', () => {
...
+        ],
+      ])
+      .dispatch({ type: 'TEST_ACTION_TYPE' })
+      .dispatch({ type: 'TEST_ACTION_TYPE' })
+      .run()
+
+    expect(ValoraAnalytics.track).toHaveBeenCalledTimes(0)
+  })
+
+  it('does not dispatch the account funded event when network ID added', async () => {
```
satish-ravi, line 274: can we also add tests for the opposite cases and assert an event is fired?

src/tokens/saga.test.ts
finnian0826: yep, I was just following the pattern from the other tests cause I wasn't sure if it needed to be in this format but that makes sense. Just updated this and all other places where I did this.

satish-ravi (approved):

PR #4689, opened 2024-01-04
fix(send): add migration to clean recent recipients with no address

satish-ravi:
### Description

Recent recipients can include items with no address field from the old escrow payments feature. Selecting these cause a crash in the new send flow because it assumes address is always present on recent recipients (this is true for all sends now).  This adds a one time clean up of recent recipients that don't include an address

### Test plan

Unit tests

Manual:

1. On main, add a dummy recent recipient with no address, ensure it shows up in the Send screen
2. Switch to PR branch and restart app
3. Check dummy recipient no longer displays

### Related issues

- Fixes ACT-1024

### Backwards compatibility

Yes


finnian0826 (approved):
looks good to me!

PR #4696, opened 2024-01-05
feat: CleverTap Inbox Notifications

bakoushin:
### Description

[CleverTap Inbox](https://docs.clevertap.com/docs/app-inbox) Messages will be rendered along with other notifications in Notification Center.

[List of currently supported CleverTap properties](https://www.notion.so/valora-inc/CleverTap-Inbox-Notifications-0a1e839cdb304023932034e163c8031a?pvs=4)

#### Notes

1. New messages DO NOT appear within current user session. Unfortunately, CleverTap Inbox is updated only once: when the app starts. Quoting their support:
>  You will have to close the application and re-open it in order to get the new app-inbox messages they will not get updated if the application is already open. 

2. CleverTap Inbox Messages are stored in global state:
* similar approach is used for remote notifications from Admin Dashboard
* ensures the messages are already fetched when user opens Notification Center

3. Any message in Notification Center can now have an optional header:

Screenshot 2024-01-05 at 12 56 16

### Test plan

* Tested manually on Android Emulator
* Updated unit tests

### Related issues

- Fixes RET-943

### Backwards compatibility

Y


src/home/NotificationCenter.tsx
```diff
@@ -19,9 +24,114 @@ import useSelector from 'src/redux/useSelector'
...
+  const { t } = useTranslation()
+  const dispatch = useDispatch()
+
+  const messages = useSelector(cleverTapInboxMessagesSelector)
+
+  const notifications: Notification[] = []
+
+  if (messages) {
+    for (const message of messages) {
+      const messageId = message.id ?? message._id
```
kathaypacific, line 43: could we do all this data massaging in the selector with `createSelector` instead? we'd get memoization for free then :)

src/home/NotificationCenter.tsx
```diff
@@ -19,9 +24,114 @@ import useSelector from 'src/redux/useSelector'
...
 
+function useCleverTapNotifications() {
+  const { t } = useTranslation()
+  const dispatch = useDispatch()
+
+  const messages = useSelector(cleverTapInboxMessagesSelector)
+
+  const notifications: Notification[] = []
+
+  if (messages) {
```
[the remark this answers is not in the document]
bakoushin: my intention was to have some guard rails in case CleverTap sends us something invalid.
i moved that logic to the selector, and now it checks that it's indeed an array (`Array.isArray()`)

src/home/NotificationCenter.test.tsx
```diff
@@ -940,4 +949,109 @@ describe('NotificationCenter', () => {
...
+      })
+      const { getByText } = render(
+        <Provider store={store}>
+          <NotificationCenter {...getMockStackScreenProps(Screens.NotificationCenter)} />
+        </Provider>
+      )
+
+      fireEvent.press(getByText('CleverTap Message CTA'))
+      expect(store.getActions()).toEqual([
+        fetchAvailableRewards(),
```
kathaypacific, line 985: nit: this feels unrelated, you could assert instead that getActions() was last called with openUrl or you could do store.clearActions() after the render instead, to keep the tests specific to notifications?

src/home/NotificationCenter.test.tsx
```diff
@@ -940,4 +949,109 @@ describe('NotificationCenter', () => {
...
+      })
+      const screen = render(
+        <Provider store={store}>
+          <NotificationCenter {...getMockStackScreenProps(Screens.NotificationCenter)} />
+        </Provider>
+      )
+
+      layoutNotificationList(screen)
+
+      await waitFor(() => expect(ValoraAnalytics.track).toHaveBeenCalledTimes(2))
```
kathaypacific, line 1041: why 2 times? perhaps we can do the `waitFor` on the assertion for notification_impression instead?

src/home/NotificationCenter.tsx
```diff
@@ -22,6 +27,73 @@ import { Spacing } from 'src/styles/styles'
 
 type NotificationsProps = NativeStackScreenProps<StackParamList, Screens.NotificationCenter>
 
+function useCleverTapNotifications() {
+  const { t } = useTranslation()
+  const dispatch = useDispatch()
+
+  const messages = useSelector(cleverTapInboxMessagesSelector)
+
+  const notifications: Notification[] = []
```
kathaypacific, line 36: i wondered whether we should `useMemo` here, but there's not much need because the list of clevertap messages should be small right?

src/home/NotificationCenter.tsx
```diff
@@ -22,6 +27,73 @@ import { Spacing } from 'src/styles/styles'
...
+        onPress: (params) => {
+          ValoraAnalytics.track(HomeEvents.notification_select, {
+            notificationType: NotificationType.clevertap_notification,
+            selectedAction: NotificationBannerCTATypes.accept,
+            notificationId,
+            notificationPositionInList: params?.index,
+          })
+          CleverTap.pushInboxNotificationClickedEventForId(messageId)
+
+          const openInExternalBrowser = false
```
kathaypacific, line 52: sorry i didn't notice this before, i wonder if we should set this value from clevertap (and let it default to false if it's not there)? i'm wondering if there will be scenarios where it makes more sense to take the user out of the app, maybe for links related to NFTs it makes sense for iOS users to go out of the app to comply with app store guidelines?

src/home/actions.ts
```diff
@@ -67,3 +78,10 @@ export const startBalanceAutorefresh = () => ({
 export const stopBalanceAutorefresh = () => ({
   type: Actions.STOP_BALANCE_AUTOREFRESH,
 })
+
+export const updateCleverTapInboxMessages = (
```
kathaypacific, line 82: nit: sorry i missed this before too - we've not been so consistent in the past but it's nice to try to model actions as events rather than setters https://redux.js.org/style-guide/#model-actions-as-events-not-setters. maybe something like `cleverTapInboxMessagesReceived`?

kathaypacific (commented):
looks great!! sorry i asked a couple of new questions which i didn't notice last time

src/home/NotificationCenter.tsx
bakoushin: good catch, thanks! yes, we can add another optional tag for this case. i would appreciate naming advice for the tag. i'm thinking of simply `browser` (as opposed to "app") because tags should be typed manually and probably the simpler it is the less error-prone it is.

kathaypacific: ah, naming is always the hardest part 🙈 fwiw, i think we (maybe not us, maybe marketing team) will have some documentation on how to use homecards and in it we could define the tags available. so they can be copy + pasted, and we remove the need to have easily type-able names. `open:browser` seems fine, `openInExternalBrowser` was my first thought

src/home/NotificationCenter.tsx
bakoushin: good point! technically messages in CleveTap inbox can be configured to have infinite TTL. which means they will be kept forever until the user dismisses them. if the user chooses not to dismiss, they will grow up in quantity. i think it's a good idea to `useMemo` as a precaution, thanks!

src/home/actions.ts
```diff
@@ -33,7 +35,16 @@ export interface RefreshBalancesBalancesAction {
   type: Actions.REFRESH_BALANCES
 }
 
-export type ActionTypes = SetLoadingAction | UpdateNotificationsAction | DismissNotificationAction
+interface UpdateCleverTapInboxMessagesAction {
+  type: Actions.UPDATE_CLEVERTAP_INBOX_MESSAGES
```
kathaypacific, line 39:
```suggestion
  type: Actions.CLEVERTAP_INBOX_MESSAGES_RECIEVED
```

(and also the interface name)

kathaypacific (approved):
woohoo looks great!! excited for this!!

sorry i've done a poor review job and giving new feedback every time i come back 🫠 so feel free to address this (if it needs addressing) in a follow up. i saw in the PR description that Clevertap will effectively only give the app notifications at one time - i interpret this to mean that notifications will not be received in real time, if a new notification is published in Clevertap during the user's session...is that right? if so, i wonder if we should just fetch the clevertap notifications on app start rather than using this event channel pattern which is used for data sources that we expect real time data from? the implementation would be simpler but also would be more explicit about how the notifications work

bakoushin:
>  notifications will not be received in real time, if a new notification is published in Clevertap during the user's session

@kathaypacific

generally yes, but there are 2 situations I'm aware of when it is updated in the real time:

1) when user dismisses the notification. [we call](https://github.com/valora-inc/wallet/blob/f411ae68409e33a25cf1aca8241ccfd0841f0864/src/home/NotificationCenter.tsx#L77) `CleverTap.deleteInboxMessageForId()` which in turn fires `CleverTap.CleverTapInboxMessagesDidUpdate` providing us with the messages without deleted one (but sadly without new ones). and yes, that could be mirrored with a single fetch.

2) when we use ["Preview & Test" functionality in CleverTap](https://www.notion.so/valora-inc/CleverTap-Inbox-Notifications-0a1e839cdb304023932034e163c8031a?pvs=4#878d92b328664862906ce7df095f1d07). in this case the message is delivered within user session after user taps on push notification on their phone.

PR #4699, opened 2024-01-05
fix(send): use correct address when searching for an address with PN mapping

satish-ravi:
### Description

If an address is searched and it contains multiple phone numbers mapped, we pick the secure send mapped address which can be different from the entered address, or if secure send is not done, we throw an error causing a crash. This was because if a phone number is present, we always favor the secure send mapping. This PR updates it so that if a recipient object has an address, that is always preferred. 

### Test plan

Unit tests, manually by

1. Have a phone number P1 mapped to two addresses A1, A2
2. On a third wallet A3, connected to a different phone number P2, search for P1, and tap on the Recipient but don't complete secure send
3. Search for A1, select recipient and complete send and ensure A1 received the funds (ensures ACT-1032 is fixed)
4. Search for P1 and complete secure send with A1
5. Search for A2, and complete send and ensure A2 received funds (display will still be PN)
6. Revoke A2 from P1 and repeat 5 (ensures ACT-1035 is fixed)

### Related issues

- Fixes ACT-1032, ACT-1035

### Backwards compatibility

Yes


MuckT (approved):
Followed your testing steps and it appears to be WAI for me!

jophish (approved):

PR #4723, opened 2024-01-08
feat(cash-in): add multichain support to coinbasepay

jh2oman:
### Description

Makes the coinbase pay module more flexible for adding ETH specifically and other assets in the future.

https://github.com/valora-inc/wallet/assets/8432644/94ef56b7-d3e1-4354-9c4d-ef8f26739054

### Test plan

Walked through manually. see screenshots

### Related issues

https://linear.app/valora/issue/ACT-899/fetch-quotes-from-remaining-external-providers

### DIFF
--- a/src/fiatExchanges/CoinbasePaymentSection.test.tsx
+++ b/src/fiatExchanges/CoinbasePaymentSection.test.tsx
@@ -64,7 +64,7 @@ describe('CoinbasePaymentSection', () => {
       assets: ['ETH'],
     },
   ])(
-    'calls generateOnRampURL with correct params',
+    'calls generateOnRampURL with correct params for $tokenId',
     async ({ tokenId, supportedNetworks, assets }) => {
       jest.mocked(generateOnRampURL).mockReturnValue(FAKE_URL)
       render(
--- a/src/fiatExchanges/CoinbasePaymentSection.test.tsx
+++ b/src/fiatExchanges/CoinbasePaymentSection.test.tsx
@@ -12,7 +12,14 @@ import {
 import { PaymentMethod } from 'src/fiatExchanges/utils'
 import { navigate } from 'src/navigator/NavigationService'
 import { createMockStore } from 'test/utils'
-import { mockProviders, mockProviderSelectionAnalyticsData } from 'test/values'
+import {
+  mockAccount,
+  mockCeloTokenId,
+  mockEthTokenId,
+  mockProviderSelectionAnalyticsData,
+  mockProviders,
+  mockTokenBalances,
+} from 'test/values'
 
 const FAKE_APP_ID = 'fake app id'
 const FAKE_URL = 'www.coinbasepay.test'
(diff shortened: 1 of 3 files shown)


src/fiatExchanges/CoinbasePaymentSection.test.tsx
```diff
@@ -36,17 +43,67 @@ describe('CoinbasePaymentSection', () => {
...
+      supportedNetworks: ['celo'],
+      assets: ['CGLD'],
+    },
+    {
+      tokenId: mockEthTokenId,
+      supportedNetworks: ['ethereum'],
+      assets: ['ETH'],
+    },
+  ])(
+    'calls generateOnRampURL with correct params',
```
MuckT, line 67: nit: maybe adding a `testName` to the above array and using that so that we see which network is used from the test output.

src/fiatExchanges/CoinbasePaymentSection.tsx
```diff
@@ -13,28 +13,43 @@ import { navigate } from 'src/navigator/NavigationService'
...
   appId,
   analyticsData,
+  tokenId,
 }: CoinbasePaymentSectionProps) {
   const { t } = useTranslation()
   const walletAddress = useSelector(walletAddressSelector)!
+  const tokenInfo = useTokenInfo(tokenId)
 
+  if (!tokenInfo) {
+    // should never happen
```
MuckT, line 40: A logger error here would be nice so that we known that something that should never happen has happened.

MuckT (approved):
Approved with a nit and one suggestion!

src/fiatExchanges/CoinbasePaymentSection.tsx
```diff
@@ -13,28 +13,43 @@ import { navigate } from 'src/navigator/NavigationService'
...
+    return null
+  }
   // Using 'CGLD' as temp replacement for CiCoCurrency.CELO – digitalAsset
+
+  const network = getNetworkFromNetworkId(tokenInfo.networkId)!
```
satish-ravi, line 45: can we use a static map instead and avoid the null assertion operator? We also have a static map of network id to network, but I think it's safest just define a specific map here.
https://github.com/valora-inc/wallet/blob/5af717e768cef266ee09c190726cd626e48c20fb/src/web3/networkConfig.ts#L386-L391

src/fiatExchanges/CoinbasePaymentSection.test.tsx
satish-ravi: +1, could also just add the token Id. E.g., 
```suggestion
    'calls generateOnRampURL with correct params for $tokenId',
```

src/fiatExchanges/CoinbasePaymentSection.tsx
```diff
@@ -13,28 +13,43 @@ import { navigate } from 'src/navigator/NavigationService'
...
+  const symbol = tokenInfo.symbol === 'CELO' ? 'CGLD' : tokenInfo.symbol
+
   const coinbasePayURL = generateOnRampURL({
     appId,
-    destinationWallets: [{ address: walletAddress, blockchains: ['celo'], assets: ['CGLD'] }],
+    destinationWallets: [
+      { address: walletAddress, supportedNetworks: [network], assets: [symbol] },
```
satish-ravi, line 51: Is either one of blockchains or supportedNetworks enough? And is there a difference between the two?

jh2oman: `blockchain` should be used when we want to display all available tokens to the user per network. `supportedNetworks` is for specifying specific assets on a network https://docs.cloud.coinbase.com/pay-sdk/docs/layer2

MuckT (approved):

PR #4729, opened 2024-01-09
chore: remove unused FeeDrawer component

kathaypacific:
### Description

As the title.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y


src/components/LegacyFeeDrawer.tsx
```diff
@@ -26,7 +26,7 @@ interface Props {
 }
 
 /**
- * @deprecated use FeeDrawer instead
+ * @deprecated
  */
 export default function LegacyFeeDrawer({
```
jeanregisser, line 31: What do we want to do with this one? 
Feels odd to have a deprecated component without an explanation as to what should be done instead.

satish-ravi: We added the FeeDrawer (using tokenIds) to replace the LegacyFeeDrawer (using CurrencyDisplay enum) as part of the new send confirmation screen in the send flow refactor (#4310), but it looks like LegacyFeeDrawer was updated in place to use tokenIds in #4686? Maybe we can just remove the deprecated tag and rename it back to FeeDrawer? It turned out that the new send flow didn't need a fee drawer (#4527)

jeanregisser: Sounds good :+1: 
And we could move the `@deprecated` tag to the currency prop?

PR #4730, opened 2024-01-09
fix(wc): incorrect normalization of transaction fields

jeanregisser:
### Description

Fixes unable to correctly send/sign transactions with some dapps when using viem.

Although Valora was able to successfully sign/send the TX from its point of view, it was actually producing a TX with a really high `nonce` value, when the dapp was providing the `nonce` value as a hex string.

This was the problematic case:
1. Dapp asks to send a TX with `nonce` set to `"0x19"` which is a valid format
2. Valora receives it and keeps the `nonce` as is while [normalizing](https://github.com/valora-inc/wallet/blob/ed4d4266e2e5bfb87297fabab8e497b9a11eb8f1/src/walletConnect/saga.ts#L360)
3. However the output of the normalization should be a fully valid viem TX, which expects some fields as `bigint` and `nonce` as a `number`.
4. Then later when we call the viem client to send or sign, it uses the `"0x19"` nonce thinking it's a `number` and ends up producing an incorrect serialized value for it, which is different and way higher.

To fix this, we're now ensuring the result of the normalization converts all needed fields in the types expected by viem (usually `bigint` or `number`).

Also added more logs to help debug this type of issue.

See also these Slack threads:
- [thread 1](https://valora-app.slack.com/archives/C05BHD5AGSW/p1704311206430959?thread_ts=1704306964.565909&cid=C05BHD5AGSW)
- [thread 2](https://valora-app.slack.com/archives/C05BHD5AGSW/p1704787988979889?thread_ts=1704748140.888959&cid=C05BHD5AGSW)

Note: many dapps on Celo don't pass `nonce`, so it wasn't directly apparent there. But Celo dapps are also affected.

### Test plan

- Updated unit tests
- Manually tested interactions with https://app.safe.global work
- Manually tested with https://react-app.walletconnect.com/
  - Note that this one doesn't surface the nonce issue and happily reports that the received hash is valid

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/src/walletConnect/request.ts
+++ b/src/walletConnect/request.ts
@@ -9,6 +9,7 @@ import { getFeatureGate } from 'src/statsig'
 import { StatsigFeatureGates } from 'src/statsig/types'
 import { chooseTxFeeDetails, sendTransaction } from 'src/transactions/send'
 import { Network, newTransactionContext } from 'src/transactions/types'
+import Logger from 'src/utils/Logger'
 import { ViemWallet } from 'src/viem/getLockableWallet'
 import {
   SerializableTransactionRequest,
@@ -23,7 +24,7 @@ import { call } from 'typed-redux-saga'
 import { SignMessageParameters } from 'viem'
 import Web3 from 'web3'
 
-const TAG = 'WalletConnect/handle-request'
+const TAG = 'WalletConnect/request'
 
 export function* handleRequest(
   {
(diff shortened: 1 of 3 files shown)


dievazqu (approved):

kathaypacific (approved):
✨🙏🏻 awesome, thanks for all the extra logging too!!

PR #4733, opened 2024-01-10
feat(swaps): group tokens without priceUsd at bottom of token selection sheet

kathaypacific:
### Description

As the title.

### Test plan

https://github.com/valora-inc/wallet/assets/20150449/34f590bc-942d-4e13-8d5a-6ad1181e44ee

### Related issues

- Fixes RET-977

### Backwards compatibility

Y


src/tokens/TokenBalanceItem.tsx
```diff
@@ -12,14 +12,20 @@ import { TokenBalance } from 'src/tokens/slice'
...
+export const TokenBalanceItem = ({
+  token,
+  onPress,
+  containerStyle,
+  balanceUsdErrorFallback,
+}: Props) => {
   const { t } = useTranslation()
 
-  const Content = ({ token, containerStyle }: Props) => {
+  const Content = () => {
```
jeanregisser, line 28: Wait, we shoudn't create new components during render. This is bad for perf.
I know it's not directly related to the focus of this PR, we can address separately.

src/tokens/selectors.ts
```diff
@@ -356,7 +343,26 @@ export const swappableTokensByNetworkIdSelector = createSelector(
...
+            return token1.name.localeCompare(token2.name)
+          }
+          // Otherwise, sort such that the token with a non-null price comes first
+          return token1.priceUsd === null ? 1 : -1
+        }
+
+        // Sort by balance (higher balances first)
+        const token1UsdBalance = token1.balance.multipliedBy(token1.priceUsd)
+        const token2UsdBalance = token2.balance.multipliedBy(token2.priceUsd)
+        if (token1UsdBalance > token2UsdBalance) return -1
```
jeanregisser, line 360: I think this converts the values to `number` loosing precision and should be avoided.

src/tokens/selectors.ts
```diff
@@ -356,7 +343,26 @@ export const swappableTokensByNetworkIdSelector = createSelector(
           (tokenInfo.minimumAppVersionToSwap &&
             !isVersionBelowMinimum(appVersion, tokenInfo.minimumAppVersionToSwap))
       )
-      .sort(tokenCompareByUsdBalanceThenByName)
+      .sort((token1, token2) => {
```
jeanregisser, line 346:
```suggestion
      // sort by balance USD (DESC) then name, tokens without a priceUsd are pushed last
      .sort((token1, token2) => {
```

jeanregisser (commented):
Great!

src/tokens/selectors.ts
jeanregisser: Here's a problematic case:

Screenshot 2024-01-11 at 13 52 15

kathaypacific: yikes, i was treating this as a number for some reason. updated to use the big number comparisons, thanks for the catch!

src/tokens/TokenBalanceItem.tsx
jeanregisser: no worries, we could just store it in a variable instead of adding another component.

src/tokens/TokenBalanceItem.tsx
```diff
@@ -12,76 +12,80 @@ import { TokenBalance } from 'src/tokens/slice'
 
 interface Props {
   token: TokenBalance
+  balanceUsdErrorFallback?: string
   onPress?: () => void
   containerStyle?: ViewStyle
 }
 
-export const TokenBalanceItem = ({ token, onPress, containerStyle }: Props) => {
+const TokenBalanceItemContent = ({ token, containerStyle, balanceUsdErrorFallback }: Props) => {
```
jeanregisser, line 20: nit: this extra component isn't really needed, we can store the JSX in a variable in the existing component.

src/tokens/TokenBalanceItem.tsx
kathaypacific: why is it preferable to use a variable rather than a normal component?

jeanregisser (approved):
🚀

src/tokens/TokenBalanceItem.tsx
jeanregisser: Just for simplicity, I don't see a strong case for having 2 of them.

PR #4734, opened 2024-01-10
feat(import-token): Show imported tokens in asset list and send flow.

dievazqu:
### Description

This PR finalizes the work to import and show new tokens.

Note: This PR is missing the correct token icons.

### Test plan

Locally in a emulator + unit test

### Related issues

- Fixes RET-891

### Backwards compatibility

Yes

### Videos

https://github.com/valora-inc/wallet/assets/12504748/3ca6c156-92d4-4365-8762-08357a330c34

https://github.com/valora-inc/wallet/assets/12504748/ded9ec4c-993e-4667-8a2c-e6bdc00933c7


jeanregisser (approved):
Awesome! 🚀 

I really like how you implemented this!

A couple of questions:
- given the approach, we need to fetch the balance once a token is imported, is it already being triggered on import?
- should we also plan to update our analytics to be able to know whether an interaction (for instane send) was with an imported token?

src/tokens/TokenImport.tsx
```diff
@@ -109,24 +109,21 @@ export default function TokenImportScreen(_: Props) {
       publicClient: client,
     })
 
-    const [symbol, decimals, name, balance] = await Promise.all([
```
kathaypacific, line 112: i think this balance check was done here not to store the balance, but to check that the balance can be correctly fetched (to avoid storing the token and getting errors later on). that way we can fail more quickly and show an error / not save this token. is there harm in keeping this?

src/tokens/TokenImport.tsx
```diff
@@ -175,16 +172,26 @@ export default function TokenImportScreen(_: Props) {
   }
 
   const handleImportToken = () => {
+    if (!tokenDetails) {
+      return
```
kathaypacific, line 176: this shouldn't happen, should we log an error or warning?

src/tokens/TokenImport.tsx
```diff
@@ -221,8 +228,8 @@ export default function TokenImportScreen(_: Props) {
           {/* Token Symbol */}
           <TextInputGroup
             label={t('tokenImport.input.tokenSymbol')}
-            value={tokenSymbol}
-            onChangeText={setTokenSymbol}
+            value={tokenDetails?.symbol ?? ''}
+            onChangeText={() => undefined}
```
kathaypacific, line 232: do we need this?

src/tokens/selectors.ts
```diff
@@ -88,6 +89,22 @@ export const tokensByIdSelector = createSelector(
...
+export const importedTokensInfoSelector = createSelector(
+  (state: RootState) => state.tokens.importedTokens,
+  (manualTokens) => {
+    const tokenBalances: StoredTokenBalances = {}
+    for (const manualToken of Object.values(manualTokens)) {
+      tokenBalances[manualToken.tokenId] = {
+        ...manualToken,
+        // Force imported tokens to be visible even with zero balance.
+        showZeroBalance: true,
+        balance: null,
```
kathaypacific, line 101: does this set the balance to be 0 always, so we never show the balance? do we need this?

src/tokens/slice.ts
```diff
@@ -97,10 +97,20 @@ export interface TokenBalancesWithAddress {
   [tokenAddress: string]: TokenBalanceWithAddress | undefined
 }
 
+export interface ImportedToken {
```
kathaypacific, line 100: instead of declaring a new interface that could diverge with the tokens interface, could we use `BaseToken`? or `Omit<BaseToken, ...>`?

src/tokens/selectors.ts
```diff
@@ -88,6 +89,22 @@ export const tokensByIdSelector = createSelector(
   }
 )
 
+export const importedTokensInfoSelector = createSelector(
+  (state: RootState) => state.tokens.importedTokens,
+  (manualTokens) => {
```
kathaypacific, line 94:
```suggestion
  (importedTokens) => {
```

kathaypacific (commented):
looks great! i had the same questions as Jean and also a few more small ones 👨🏼‍🎨

src/tokens/TokenImport.tsx
dievazqu: Hmm, I guess it's fine to fetch the balance here, I removed it because I feel it was really needed, but I can rollback this.

src/tokens/selectors.ts
dievazqu: I'm explicitly setting the balance here to use the same interface `StoredTokenBalances`.
The balances are updated in the saga.
This selector returns only the "static" information for the token.

If we return the balance here, we might return a wrong value (since this would be the balance value when the token was imported).

dievazqu:
> Awesome! 🚀
> 
> I really like how you implemented this!
> 
> A couple of questions:
> 
> * given the approach, we need to fetch the balance once a token is imported, is it already being triggered on import?
> * should we also plan to update our analytics to be able to know whether an interaction (for instance send) was with an imported token?

* I think this is already cover [here](https://github.com/valora-inc/wallet/blob/df7952938cba3b1c304351710dd20a9d97faf8f3/src/tokens/TokenImport.tsx#L190), right?
* That's a good point, I didn't consider it, I was trying to handle the imported tokens as normal token but maybe we should add a boolean to the `StoredTokenBalances` to indicate if a token is imported?

kathaypacific:
the import token feature now is only scoped to show the token balance after import. i think we can probably punt on enabling tracking for interactions with imported tokens when we work on the next iteration :)

PR #4736, opened 2024-01-10
fix: render svg wallet connect dapp images

kathaypacific:
### Description

As the title - we're seeing more of these so worth to fix.

### Test plan

Before:
Simulator Screenshot - iPhone 14 Pro - 2024-01-10 at 17 48 51

After:
Simulator Screenshot - iPhone 14 Pro - 2024-01-10 at 17 02 30

### Related issues

- Fixes RET-957

### Backwards compatibility

Y


src/walletConnect/screens/Logos.tsx
```diff
@@ -29,7 +34,11 @@ export default function Logos({ dappImageUrl, dappName }: Props) {
         </View>
       </View>
       <View style={styles.logoShadow}>
-        {dappImageUrl ? (
+        {isDappImageSvg ? (
+          <View style={styles.logoBackground}>
+            <SvgUri width={24} height={24} uri={dappImageUrl} />
```
jeanregisser, line 39: This works well with the Curve example, 
but if the SVG is a square image with a solid background, it won't be pretty, right?

src/walletConnect/screens/Logos.tsx
```diff
@@ -16,11 +14,18 @@ interface Props {
...
-  const dappConnectInfo = useSelector(dappConnectInfoSelector)
-  if (!dappImageUrl && dappConnectInfo !== DappConnectInfo.Basic) {
+  if (!dappImageUrl) {
     return null
   }
 
+  let isDappImageSvg = false
+  try {
+    const parsedUrl = new URL(dappImageUrl ?? '')
+    isDappImageSvg = parsedUrl.pathname.endsWith('.svg')
```
jeanregisser, line 24:
```suggestion
    isDappImageSvg = parsedUrl.pathname.toLowerCase().endsWith('.svg')
```

I wonder if we should useMemo to avoid parsing it on each render 🤔

jeanregisser (commented):
Oh great!

I always wanted this, kinda pushed it because I didn't like that `react-native-svg` doesn't render like browsers as it only implements a subset of the spec.

But I guess this is good enough until we can switch to an image library that supports SVG out of the box like https://docs.expo.dev/versions/latest/sdk/image/

src/walletConnect/screens/Logos.tsx
kathaypacific: the svg is smaller in size (24) than the containing circle, so even with a colored background the svg should be contained in the circle. i checked with an svg like https://www.svgrepo.com/show/521955/arrow-down-square.svg and it seems okay to me?

Simulator Screenshot - iPhone 14 Pro - 2024-01-11 at 10 16 39

src/walletConnect/screens/Logos.tsx
kathaypacific: i didn't bother with memo-ing here because it is such a small component that exists for a short time? the computation also seems quite inexpensive?

src/walletConnect/screens/Logos.tsx
jeanregisser: It's ok, best we can do with the current design I think.

jeanregisser (approved):

PR #4738, opened 2024-01-10
fix(sends): Fix error checking when estimating gas

jophish:
### Description

Currently, the error strings we check for when estimating gas are not sufficient to catch all types of errors we care about. Basically, what was happening was when a user tries to send MAX balance of some gas-payable asset A, we try to prepare the transaction using asset A (and other gas-payable assets) as gas, and estimation fails when using asset A as gas since gas + transfer amount exceeds balance. This error during estimation wasn't being caught by the `cause` strings we were looking at. This caused gas estimation to hang forever, preventing the user from continuing with the transfer. This PR adds in a new cause string to fix that.

In general, it appears as if this cause string is determined _by the blockchain_ itself, i.e., not by `viem`, and therefore may be liable to change in the wild. This is a temporary fix, but it may be worth revisiting this if it crops up again, and making the errors we check for configurable remotely.

### Test plan

Manual tested.

### Related issues

- Fixes ACT-1045


MuckT (approved):
Tested locally and I am now being correctly served the max warning 🚀

PR #4739, opened 2024-01-10
chore: update e2e test to use ENS (instead of Nom)

silasbw:


jeanregisser:
Maybe we could have have both checks?
Or do we want to drop nom?

silasbw:
dropping nom

PR #4745, opened 2024-01-11
chore: consolidate alert icons

kathaypacific:
### Description

I came across 3x alert icons (they all look very similar....)

This PR: 
- removes Alert.tsx, ExclamationCircleIcon.tsx (use Attention.tsx instead)
- removes unused icons (Help, StarIllustration)
- updates consumers

### Test plan

I didn't do thorough testing because the icons look so similar. 
Alert => Attention with color errorDark
ExclamationCircleIcon => Attention with color primary

From top to bottom, the icons are Attention, InfoIcon (is there because I thought it looks the same...), Alert, ExclamationCircleIcon:
Screenshot 2024-01-11 at 16 23 55

### Related issues

- Related to RET-978

### Backwards compatibility

Y


satish-ravi (commented):
+1 to removing duplicate icons, but are the i and the exclamation icons different? The former seems for extra info / details whereas the latter seems for warning / urgent message. Not sure how much it matters though.

kathaypacific:
> +1 to removing duplicate icons, but are the i and the exclamation icons different? The former seems for extra info / details whereas the latter seems for warning / urgent message. Not sure how much it matters though.

oooohhhh you're right!!!! i didn't even notice there is a difference...but indeed there is an info circle and an exclamation circle 🤦🏻‍♀️ i guess i'll update to keep both

dievazqu (approved):

src/icons/Attention.tsx
```diff
@@ -2,8 +2,16 @@ import * as React from 'react'
 import Svg, { Path } from 'react-native-svg'
 import Colors from 'src/styles/colors'
 
-const AttentionIcon = ({ color = Colors.warningDark }) => (
-  <Svg width={16} height={16} fill="none">
+const AttentionIcon = ({
+  color = Colors.warningDark,
```
jeanregisser, line 6: I wonder if we should use a limited set of colors to ensure consistency?

src/navigator/DrawerNavigator.tsx
```diff
@@ -304,7 +304,7 @@ export default function DrawerNavigator({ route }: Props) {
               <View style={styles.itemStyle}>
                 <Text style={styles.itemTitle}>{t('walletSecurity')}</Text>
                 <View style={styles.drawerItemIcon}>
-                  <ExclamationCircleIcon />
+                  <AttentionIcon color={Colors.primary} />
```
jeanregisser, line 307: Looks like we should pass `size={20}` since it was the size used in `ExclamationCircleIcon`.

src/navigator/DrawerNavigator.tsx
```diff
@@ -324,7 +324,7 @@ export default function DrawerNavigator({ route }: Props) {
               <View style={styles.itemStyle}>
                 <Text style={styles.itemTitle}>{t('accountKey')}</Text>
                 <View style={styles.drawerItemIcon}>
-                  <ExclamationCircleIcon />
+                  <AttentionIcon color={Colors.primary} />
```
jeanregisser, line 327: Same here about `size={20}`

jeanregisser (commented):
👍

src/navigator/DrawerNavigator.tsx
kathaypacific: fixed, thanks!

src/icons/Attention.tsx
kathaypacific: my vote is no unless this request came from design (i.e. we've committed to using a set in future designs) - i think deriving what the limited set should be based on previous usage will not be reliable for predicting future designs and we'll end up augmenting the set, defeating the purpose. the app `colors` enum has shrunk recently and ensures some level of consistency already, so i'm not sure that i see the need to define a subset of the enum

PR #4749, opened 2024-01-12
feat: add no usd price warning to swap token select

kathaypacific:
### Description

This PR:
- add full screen inline notification component
- update swap screen to show full screen inline notification on select token with no usd price
- update swap screen warnings - separate warnings for no usd price and missing price impact
- add unit tests

Note: most of the changes in this PR are for the swap unit tests - i needed to add handling for the no usd warning centrally somehow, which i couldn't avoid updating every test with.

### Test plan

https://github.com/valora-inc/wallet/assets/20150449/21ba41f2-34fe-4bac-bb76-95e9e98cd268

### Related issues

- Fixes RET-978

### Backwards compatibility

Y


src/components/BottomSheetInLineNotification.tsx
```diff
@@ -0,0 +1,64 @@
+import React, { useState } from 'react'
```
kathaypacific, line 1: it's not really clear to me how this component will evolve, so i've made it quite specific to inline notifications rather than very generic to save time

src/components/BottomSheetInLineNotification.tsx
```diff
@@ -0,0 +1,64 @@
...
+import Modal from 'src/components/Modal'
+import { useShowOrHideAnimation } from 'src/components/useShowOrHideAnimation'
+import { Spacing } from 'src/styles/styles'
+
+interface Props extends InLineNotificationProps {
+  showNotification: boolean
+}
+
+// this value is used to ensure the notification is offset by its own height when transitioning in and out of view
+const NOTIFICATION_HEIGHT = 200
```
jeanregisser, line 14: This seems could cause issues if the notification height is bigger than this hardcoded value.
For instance what happens if the text is longer in German?

src/icons/Warning.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react'
+import Svg, { Path } from 'react-native-svg'
+import Colors from 'src/styles/colors'
+
+const Warning = ({ color = Colors.warningDark, size = 16 }: { color?: Colors; size?: number }) => (
```
jeanregisser, line 5: nit: using `function` is easier to read IMHO
```suggestion
export default function Warning({ color = Colors.warningDark, size = 16 }: { color?: Colors; size?: number }) (
```

jeanregisser, line 5: Wait, didn't we have a warning icon already?

src/components/TokenBottomSheet.tsx
```diff
@@ -122,20 +131,38 @@ function TokenBottomSheet<T extends TokenBalance>({
...
-}: Props<T>) {
+  showPriceUsdUnavailableWarning,
+}: TokenBottomSheetProps<T>) {
   const [searchTerm, setSearchTerm] = useState('')
+  const [selectedTokenNoUsdPrice, setSelectedTokenNoUsdPrice] = useState<T | null>(null)
 
   const { t } = useTranslation()
 
-  const onTokenPressed = (token: T) => () => {
+  const handleTokenPressed = (token: T) => () => {
```
jeanregisser, line 141: nit: maybe we should agree to use `onXXX` vs `handleXXX` and aim for consistency across components 😄

src/components/BottomSheetInLineNotification.tsx
```diff
@@ -0,0 +1,64 @@
...
+      setIsVisible(false)
+    }
+  )
+
+  if (!isVisible) {
+    return null
+  }
+
+  return (
+    <Modal style={styles.modal} isVisible={true} backdropOpacity={0.5}>
```
jeanregisser, line 43: Oh no 😢 would be nice to void `Modal` going forward
https://docs.google.com/presentation/d/1QQIRY7r_q3n4gS_y0RD-Z9tMNUMpvy_gnGlkZUbECeg/edit#slide=id.g24597cb3da6_0_10

We don't have an easy alternative yet though.

jeanregisser:
> I am checking if we still need the price impact warning on the swap screen, if we already have the user confirm the action...

The price impact warning is independent from the fact we have a `priceUsd`, right?

src/components/InLineNotification.tsx
```diff
@@ -69,8 +64,21 @@ export function InLineNotification({
...
-          {renderCtaLabel(ctaLabel, onPressCta, severityColor.primary)}
-          {renderCtaLabel(ctaLabel2, onPressCta2, severityColor.primary)}
+          {ctas.map(({ label, onPress }, index) => (
+            <Text
+              key={label}
+              style={[
+                styles.ctaLabel,
+                {
+                  color: severityColor.primary,
+                  marginRight: index === 0 ? Spacing.Small12 : 0,
```
bakoushin, line 74: we can probably leverage the [gap](https://reactnative.dev/docs/layout-props#gap) property in container component instead

src/components/BottomSheetInLineNotification.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+// this value is used to ensure the notification is offset by its own height when transitioning in and out of view
+const NOTIFICATION_HEIGHT = 200
+
+// for now, this notification is launched from the bottom of the screen only
+const BottomSheetInLineNotification = ({ showNotification, ...inLineNotificationProps }: Props) => {
+  const [isVisible, setIsVisible] = useState(showNotification)
+
+  const progress = useSharedValue(0)
+  const animatedStyle = useAnimatedStyle(() => {
```
bakoushin, line 21: looks like if we get rid of `animatedStyle` nothing will change. 
my guess it's because animation is in fact handled by `Modal`.

kathaypacific: i'm not sure that's true - the animatedStyle here is doing the up / down animation for the notification. without this, the notification will fade in / out only

src/components/BottomSheetInLineNotification.tsx
kathaypacific: i increased this value to account for a buffer and made the comment more detailed. if this value is greater than the toast height, it doesn't really matter (the animation happens a bit faster, it's not so noticable). if the value is less than the toast height, the animation doesn't finish in time but it's still not a terrible experience...

src/icons/Warning.tsx
kathaypacific: sadly i could not find one, i went through the whole icons folder

src/components/InLineNotification.tsx
kathaypacific: TY!! i keep forgetting that we can use `gap` now (we weren't on the right version of react native for a long time). i tidied up the styles

src/components/BottomSheetInLineNotification.tsx
```diff
@@ -0,0 +1,81 @@
...
+}
+
+const styles = StyleSheet.create({
+  modal: {
+    height: variables.height,
+    width: variables.width,
+    position: 'absolute',
+    top: 0,
+    left: 0,
+  },
```
jeanregisser, line 66: nit:
```suggestion
  modal: {
    ...StyleSheet.absoluteFillObject,
  },
```

src/components/BottomSheetInLineNotification.tsx
```diff
@@ -0,0 +1,81 @@
...
+import variables from 'src/styles/variables'
+
+interface Props extends InLineNotificationProps {
+  showNotification: boolean
+}
+
+// this value is used to animate the notification on and off the screen. the
+// value should be equal or greater than the height of the notification,
+// otherwise the notication will not transition smoothly.
+const NOTIFICATION_HEIGHT = 500
```
jeanregisser, line 18: I see, I guess we could retrieve the real height instead of hardcoding, but probably ok for now.

src/swap/SwapScreen.tsx
```diff
@@ -202,6 +203,10 @@ export function SwapScreen({ route }: Props) {
   const networkFeeInfoBottomSheetRef = useRef<BottomSheetRefType>(null)
   const slippageInfoBottomSheetRef = useRef<BottomSheetRefType>(null)
 
+  const [selectingNoUsdPriceToken, setSelectingNoUsdPriceToken] = useState<TokenBalance | null>(
+    null
+  )
```
jeanregisser, line 208: Could we move this to the reducer? as a new event?

This way all the component state remains there and we can reset based on the events dispatched.

Otherwise it will be tempting for the next person to touch this to add another `useState` again.

src/swap/SwapScreen.tsx
```diff
@@ -759,6 +801,16 @@ export function SwapScreen({ route }: Props) {
           text={t('swapScreen.transactionDetails.networkFeeInfoDismissButton')}
         />
       </BottomSheet>
+      <BottomSheetInLineNotification
+        showNotification={!!selectingNoUsdPriceToken}
+        severity={Severity.Warning}
+        title={t('swapScreen.noUsdPriceWarning.title')}
+        description={t('swapScreen.noUsdPriceWarning.description')}
+        ctaLabel={t('swapScreen.noUsdPriceWarning.ctaConfirm')}
```
bakoushin, line 809: should we reverse the CTAs to match the design?

src/components/BottomSheetInLineNotification.tsx
bakoushin: i've done [a similar trick](https://github.com/valora-inc/wallet/blob/01a6549621a99097c82997aaf44236d6f624732d/src/home/NotificationCenter.tsx#L220), seems work fine

kathaypacific: 🙈 i am not convinced that we need that level of complexity here (it'd require a state variable and layout measure, sounds like not much but still something) - 500 pixels is over 50% of the screen height of an iphone 13 pro, even more on a smaller device. it's really unlikely that this toast would exceed that height to make the animation janky

src/components/BottomSheetInLineNotification.tsx
kathaypacific: YAY i was looking for something like this, thanks!

src/swap/SwapScreen.tsx
kathaypacific: oops, yes i should have added this to the reducer. i was fixated on refactoring to move the toast into the swap screen and didn't really think further

src/swap/SwapScreen.tsx
kathaypacific: yes done! i think the design was changed, or i was crazy

src/swap/SwapScreen.tsx
```diff
@@ -69,7 +71,8 @@ interface SwapState {
   // Raw input values (can contain region specific decimal separators)
   inputSwapAmount: SwapAmount
   updatedField: Field
-  selectingToken: Field | null
+  selectingField: Field | null
+  selectingToken: TokenBalance | null
```
jeanregisser, line 75:
```suggestion
  selectingNoUsdPriceToken: TokenBalance | null
```

jeanregisser (approved):
🚀

PR #4750, opened 2024-01-12
fix(swap): tokens with balance but no usd price should be near the top of list

kathaypacific:
### Description

As the title. G$ today had no usd price and I noticed it was incorrectly grouped with the "price unavailable" tokens when I had some balance.

### Test plan

Added unit test

Simulator Screenshot - iPhone 14 Pro - 2024-01-12 at 13 24 58

### Related issues

- Fixes RET-977

### Backwards compatibility

Y


jeanregisser (approved):
👍

kathaypacific:
i confirmed with Kayla https://valora-app.slack.com/archives/C0684TXDR3K/p1705312510472529 so will go ahead and merge!

PR #4753, opened 2024-01-12
feat(send): remove picker icon when forceTokenId set

finnian0826:
### Description

When forceTokenId is set, don't show the user the picker icon to change token.

### Test plan

Picker icon shown and can tap with normal send, picker icon not shown and can't tap when forceTokenId set:

https://github.com/valora-inc/wallet/assets/140328381/0d899dad-96e3-4c43-83b5-f66b7da65ba9

### Related issues

- Fixes #[ACT-963]

### Backwards compatibility

Yes


src/send/SendEnterAmount.tsx
```diff
@@ -322,7 +322,7 @@ function SendEnterAmount({ route }: Props) {
                 <>
                   <FastImage source={{ uri: token.imageUrl }} style={styles.tokenImage} />
                   <Text style={styles.tokenName}>{token.symbol}</Text>
-                  <DownArrowIcon color={Colors.gray5} />
+                  {!forceTokenId && <DownArrowIcon color={Colors.gray5} />}
```
satish-ravi, line 325: Seems like this is still touchable, which allows changing the token. I don't think we should allow changing tokens if forceTokenId is set.
Also, can we add unit tests

satish-ravi (changes requested):

src/send/SendEnterAmount.tsx
finnian0826: okay I can update to remove the ability change, the issue just said to remove the icon so wasn't sure if we should also remove the ability to change, but should be easy to update.

jh2oman: @finnian0826 Nitya also said in a comment that the user should not have the ability to change tokens
https://linear.app/valora/issue/ACT-963/handle-forcetokenid-on-enter-amount-screen#comment-4ab5f930

src/send/SendEnterAmount.tsx
```diff
@@ -315,14 +315,18 @@ function SendEnterAmount({ route }: Props) {
               />
               <Touchable
                 borderRadius={TOKEN_SELECTOR_BORDER_RADIUS}
-                onPress={onTokenPickerSelect}
+                onPress={() => {
+                  if (!forceTokenId) {
+                    onTokenPickerSelect()
+                  }
```
satish-ravi, line 321: Instead of this, we should just not render this component as a touchable. Otherwise the component is still tappable.

src/send/SendEnterAmount.tsx
```diff
@@ -313,18 +330,7 @@ function SendEnterAmount({ route }: Props) {
...
-                style={styles.tokenSelectButton}
-                testID="SendEnterAmount/TokenSelect"
-              >
-                <>
-                  <FastImage source={{ uri: token.imageUrl }} style={styles.tokenImage} />
-                  <Text style={styles.tokenName}>{token.symbol}</Text>
-                  <DownArrowIcon color={Colors.gray5} />
-                </>
-              </Touchable>
+              {tokenSelectorComponent}
```
satish-ravi, line 333: nit: you can inline this instead of using the extra `tokenSelectorComponent` variable

satish-ravi (approved):

PR #4754, opened 2024-01-12
feat: add get started section

MuckT:
### Description

Adds the get started section, dummy feature flag and associated tests. Hooks into the existing redux app `superchargeApy` for displaying the % rewards.

### Screenshot

| iOS | Android |
| --- | --- |
|  |  |

### Test plan

- Tested locally on iOS
- Tested locally on Android
- Unit tests added

### Related issues

- Fixes ACT-1004

### Backwards compatibility

Yes


locales/base/translation.json
```diff
@@ -2014,7 +2014,11 @@
     "finiteApprovalDescription": "Approved {{approvedAmount}} {{tokenSymbol}} for trade",
     "infiniteApprovalDescription": "Approved infinite {{tokenSymbol}} for trade",
     "revokeApprovalDescription": "Revoked {{tokenSymbol}} for trade",
-    "descriptionLabel": "Details"
+    "descriptionLabel": "Details",
+    "getStarted": {
```
cajubelt, line 2018: nit: not sure I would add this under transactionFeed, seems like this content is shown as an alternative to the feed, rather than as a subset of it

src/icons/Bird.tsx
```diff
@@ -0,0 +1,23 @@
...
+import Svg, { Path } from 'react-native-svg'
+
+interface Props {
+  color?: string
+  width?: number
+  height?: number
+  viewStyle?: ViewStyle
+}
+
+export default function Bird({ color = '#888FFE', width = 85, height = 83, viewStyle }: Props) {
```
cajubelt, line 12: had no idea this was a bird. I thought it was a flower.

cajubelt: 🐦

src/transactions/feed/TransactionFeed.tsx
```diff
@@ -63,8 +64,10 @@ function TransactionFeed() {
     return groupFeedItemsInSections(pendingTransactions, confirmedFeedTransactions)
   }, [pendingTransactions, confirmedFeedTransactions])
 
+  // TODO(ACT-1006): Replace with real feature gate
+  const dummyGetStartedFlag = false
```
cajubelt, line 68: if you instead imported a function returning the feature flag value (can still be hardcoded to false), you could more easily mock the feature flag in a test and verify that the correct variation of the UI is shown.

cajubelt (approved):
looks fine, but see comments

src/transactions/feed/TransactionFeed.tsx
MuckT: I'll keep this in mind for the future, but since it's tested in the follow-up PR #4755 it seem worth adding to this PR.

locales/base/translation.json
MuckT: Moved in 041a9dd6e624d6e2d1807ca3959f41328fa6f8b0!

PR #4758, opened 2024-01-13
fix: prevent home action button resize

MuckT:
### Description

On larger fonts or longer languages the buttons would wrap text and change size. This prevents font scaling while also using ellipsis if long words are present in various translations.

#### Large Fonts
| iOS Before | iOS After | Android Before | Android After | 
|  |  |  |  |

#### Default Fonts (No Change)
| iOS | Android |
| ---- | ---- |
|  |  |

### Test plan

- Tested locally on iOS
- Tested locally on Android

### Related issues

N/A

### Backwards compatibility

Yes


cajubelt:
is this a request from Design? Agree with @bakoushin 's concerns about accessibility

MuckT:
@bakoushin @cajubelt - Updated the Android before image to highlight an issue with the ripple that was not previously shown in this PR description.

cajubelt (approved):
ios still looks kinda small-- is the main idea of this change that the total balance section becomes invisible on android without it? approving if that's the case

MuckT:
@cajubelt should just impact the button sizes.

I checked with @nityarao and got the following response, "I am on board with preventing the buttons from re-sizing!"

bakoushin:
@MuckT 
i'm still worried about trading the accessibility for a nicer look, but could see the point in treating those button texts as a part of icons, thus not sensitive to font scaling at all.

to solve the ripple scaling issue (no matter what font size solution we're up to), probably we could add `flex: 1` [here](https://github.com/valora-inc/wallet/blob/7d7121cdf119fa1f26de97b7569bbf55f02e7778/src/home/ActionsCarousel.tsx#L129-L132) and [there](https://github.com/valora-inc/wallet/blob/7d7121cdf119fa1f26de97b7569bbf55f02e7778/src/components/Touchable.tsx#L23)?

MuckT:
@cajubelt @bakoushin - update implementation disallow font scaling based on the discussion.

bakoushin (approved):
🎉

PR #4767, opened 2024-01-15
chore: update keychain lib and add more logging

kathaypacific:
### Description

As the title. 

I added a keychainError prop to the `redux_no_matching_keychain_account` event to understand if there are errors when reading from the keychain or if the keychain does not have a matching entry.

### Test plan

n/a

### Related issues

- Related to RET-937

### Backwards compatibility

Y


src/storage/keychain.tsx
```diff
@@ -26,7 +26,7 @@ export async function storeItem({ key, value, options = {} }: SecureStorage) {
   try {
     const result = await Keychain.setGenericPassword('CELO', value, {
       service: key,
-      accessible: Keychain.ACCESSIBLE.ALWAYS_THIS_DEVICE_ONLY,
```
kathaypacific, line 29: needed to change this because `ALWAYS_THIS_DEVICE_ONLY` was removed

dievazqu (approved):

src/storage/keychain.tsx
jeanregisser: This is the sensitive change. Were you able to confirm existing items stored with `ALWAYS_THIS_DEVICE_ONLY` are still readable? I guess they still are.

This was detailed in https://linear.app/valora/issue/RET-849

I'm just wondering if it could cause issues down the line once Apple actually removes [kSecAttrAccessibleAlwaysThisDeviceOnly](https://developer.apple.com/documentation/security/ksecattraccessiblealwaysthisdeviceonly), what will really happen to items stored with this attribute?

jeanregisser (commented):
👍

src/storage/keychain.tsx
bakoushin: From [kSecAttrAccessibleAlwaysThisDeviceOnly docs](https://developer.apple.com/documentation/security/ksecattraccessiblealwaysthisdeviceonly):
> Items with this attribute do not migrate to a new device. Thus, after restoring from a backup of a different device, these items will not be present.

does that mean once user migrates on the new iPhone, they will lost they wallet?!

kathaypacific: @bakoushin yes that's right - this decision to not migrate keychain items on device migration predates me and is maybe something we want to revisit. iirc the reasoning is that we protect users against any vulnerabilities in iCloud account storage by doing this. so if a user restores their iPhone, they will need their recovery phrase.

@jeanregisser on a simulator i can confirm that the existing stored items are still readable. it's my understanding that this change would affect how we store new keys, but existing keys are unaffected. we'll need to address the kSecAttrAccessibleAlwaysThisDeviceOnly deprecation separately - when we read from the keychain, we don't recieve information about what the `accessible` value is so we'd need to probably refresh the keychain items for all iOS users

bakoushin (approved):
🔒

PR #4781, opened 2024-01-18
feat(swaps): fund your wallet bottom sheet

bakoushin:
### Description

If the user proceeds to swap having zero balance of any token, instead of showing an insufficient balance error, we suggest they add funds to their wallet, taking them to cash-in flow.

| Fund your wallet bottom sheet |
|--------|
| fund-your-wallet | 

### Test plan

* Tested manually on iOS simulator
* Updated unit tests

### Related issues

- Fixes RET-979

### Backwards compatibility

Y


src/swap/SwapScreen.test.tsx
```diff
@@ -1488,4 +1489,99 @@ describe('SwapScreen', () => {
...
+
+    expect(getByText('swapScreen.confirmSwap')).not.toBeDisabled()
+    fireEvent.press(getByText('swapScreen.confirmSwap'))
+
+    expect(getByTestId('FundYourWalletBottomSheet')).toBeDefined()
+
+    expect(ValoraAnalytics.track).toHaveBeenCalledWith(SwapEvents.swap_show_fund_your_wallet)
+  })
+
+  it("should navigate to buy token flow when 'Add funds' button is pressed", async () => {
```
kathaypacific, line 1539: nit: perhaps you can combine these test cases into a single one? much of the setup code is the same :)

src/swap/SwapScreen.test.tsx
```diff
@@ -1488,4 +1489,99 @@ describe('SwapScreen', () => {
...
+    fireEvent.changeText(within(swapFromContainer).getByTestId('SwapAmountInput/Input'), '1')
+
+    await act(() => {
+      jest.runOnlyPendingTimers()
+    })
+
+    expect(getByText('swapScreen.confirmSwap')).not.toBeDisabled()
+    fireEvent.press(getByText('swapScreen.confirmSwap'))
+
+    expect(getByTestId('FundYourWalletBottomSheet')).toBeDefined()
```
kathaypacific, line 1534: if i remember correctly, the bottom sheet is always defined in the tests (not only when it is snapped to the 0 index). perhaps it'd be worthwhile to test that swapStart is not dispatched?

kathaypacific (approved):
LGTM!! 🚀

src/swap/SwapScreen.test.tsx
bakoushin: yeah, I was thinking about the same while copypasting. merged all into one.

src/swap/SwapScreen.test.tsx
bakoushin: oh, good catch! i totally overlooked this. thanks for the suggestion!

PR #4784, opened 2024-01-19
chore: precautions for CleverTap Inbox integration

bakoushin:
### Description

To mitigate some possible issues in the future caused by CleverTap Inbox integration the following precautions are implemented:
* Receiving and processing messages from CleverTap is put behind a feature flag
* Instead of storing raw CleverTap data in the redux store, only parsed (and thus expected to be valid) messages are stored. Message parsing is done in the channel before emitting them to a saga.

### Test plan

* Updated unit tests
* Tested manually on iOS and Android emulators

### Related issues

- Related to RET-943

### Backwards compatibility

Y


kathaypacific (approved):
💅🏼

src/home/selectors.ts
```diff
@@ -31,69 +28,5 @@ export const getExtraNotifications = createSelector(
...
-        ctaUrl,
-        priority: !Number.isNaN(priority) ? priority : undefined,
-        openInExternalBrowser,
-      })
-    }
-
-    return messages
-  }
-)
+export const cleverTapInboxMessagesSelector = (state: RootState) =>
```
kathaypacific, line 31: should we add the feature flag here too, and return empty if not enabled?

bakoushin: totally! thanks for the suggestion. i also added a gate inside the event emitter.

PR #4792, opened 2024-01-23
fix: ensure that deep links are consumed after onboarding

kathaypacific:
### Description

This PR adds a "holding pattern" to the deep links handler. If the wallet address is not yet available, and if the user has not finished onboarding, store the deep link to be consumed after these conditions. Specifically:
- if the user has finished onboarding, consume the deep link right away (no change to this flow, does not go through redux)
- if the user has not finished onboarding, store the deep link in redux to be consumed after onboarding
- the deep link is removed after the `OPEN_DEEP_LINK` action is dispatched
- if a user does not finish onboarding in their app session: on app restart, remove all deep links that should not be persisted. most deep links should not be persisted but there are some that should be - particularly, we still want to execute the invites and jumpstart links.
- i used an array for storing the pending deep links. i did this because it removes the need to do extra logic to handle new deep links when there is a stored deep link.
- the selector will only select the first (oldest) deep link and the `useDeepLinks` hook will dispatch the action to process them async (essentially execute the deep links one after another until there are no more stored links)

Side note:
I noticed that the dynamic link for the invitations are processed twice. the short link (e.g. `https://vlra.app/xxxxxx`) is used to invoke the deep link handler through the `Linking` api, and the resolved link (e.g. `https://valoraap.com/share/xxxxxx`) is also used to invoke the deep link handler through the `dynamicLinks` listeners. So i updated the deep link handler to ignore the short links.

This change defers all deep links until the user has created a wallet. I _think_ that this is a reasonable requirement, as I can't think of any use cases for a user without a wallet.

### Test plan

Tested manually that the deep link handler waits until after onboarding to execute. Also relying on CI for regression.

### Related issues

- Fixes RET-898

### Backwards compatibility

Yes


src/app/reducers.ts
```diff
@@ -287,6 +303,21 @@ export const appReducer = (
...
       }
+    case Actions.DEEP_LINK_DEFERRED:
+      return {
+        ...state,
+        pendingDeepLinks: [
+          ...state.pendingDeepLinks,
+          { url: action.deepLink, isSecureOrigin: action.isSecureOrigin },
+        ],
+      }
+    case Actions.OPEN_DEEP_LINK:
```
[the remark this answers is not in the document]
bakoushin: ah, i see, it was already there. please ignore.

bakoushin (approved):
📫 LGTM!

src/app/saga.test.ts
```diff
@@ -245,7 +265,10 @@ describe('handleDeepLink', () => {
   it('Handles jumpstart links', async () => {
     const deepLink = 'celo://wallet/jumpstart/0xPrivateKey'
     await expectSaga(handleDeepLink, openDeepLink(deepLink))
-      .provide([[select(walletAddressSelector), '0xwallet']])
+      .provide([
+        [select(walletAddressSelector), '0xwallet'],
+        [select(walletAddressSelector), mockAccount],
```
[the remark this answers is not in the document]
kathaypacific: oops no, i removed this change. thanks!!

src/app/reducers.ts
```diff
@@ -7,6 +7,8 @@ import { REMOTE_CONFIG_VALUES_DEFAULTS } from 'src/firebase/remoteConfigValuesDe
 import { Screens } from 'src/navigator/Screens'
 import { getRehydratePayload, REHYDRATE, RehydrateAction } from 'src/redux/persist-helper'
 
+const PERSISTED_DEEP_LINKS = ['https://valoraapp.com/share', 'celo://wallet/jumpstart']
+
```
jeanregisser, line 11: I think we don't care so much about the scheme and domain, it's more the `/path` part

src/app/saga.test.ts
```diff
@@ -225,9 +245,9 @@ describe('handleDeepLink', () => {
   })
 
   it('Handles short share deep link', async () => {
-    const deepLink = 'https://vlra.app/someShortLink'
+    const deepLink = 'https://celo.org/share/abc123'
```
jeanregisser, line 248: Why are we changing this?

src/app/saga.ts
```diff
@@ -318,21 +309,19 @@ function convertQueryToScreenParams(query: string) {
...
-  // Try resolve dynamic links
-  if (deepLink.startsWith(DYNAMIC_LINK_DOMAIN_URI_PREFIX)) {
-    const resolvedDynamicLink = yield* call(resolveDynamicLink, deepLink)
-    if (resolvedDynamicLink) {
-      deepLink = resolvedDynamicLink
-    }
+  if (isWalletConnectDeepLink(deepLink)) {
+    yield* call(handleWalletConnectDeepLink, deepLink)
+    return
   }
```
jeanregisser, line 325: Are we sure this isn't needed anymore?
Maybe this was needed for links scanned via the in-app QR scanner?
or when passed from our own notifications?

src/app/saga.test.ts
```diff
@@ -225,9 +245,9 @@ describe('handleDeepLink', () => {
   })
 
   it('Handles short share deep link', async () => {
```
jeanregisser, line 247: If we're sure we don't need to resolve them ourselves,
```suggestion
  it('Handles share deep link', async () => {
```

jeanregisser (commented):
Great! 🚀

src/app/saga.ts
kathaypacific: i *think* that it's not needed anymore...

the QR scanner [saga](https://github.com/valora-inc/wallet/blob/0c4adb8166258c184f171011e94025971ab681ac/src/qrcode/utils.ts#L146-L174) does not trigger the deep link flow currently. if we want to be able to handle dynamic links through the scanner, we'll need to make a code change here anyway so i thought it was a safe change.

the "passed from our own notifications" scenario is interesting, but i didn't think it was a real scenario? the notifications (both homecards and push notifications) actually don't work well with dynamic links (i just tested) because[ it will try to open it in a browser](https://github.com/valora-inc/wallet/blob/main/src/app/saga.ts#L407). both flows end up there. even if it does work, i guess i'm wondering why someone internal would choose to add a dynamic link instead of a deep link to target an existing user. it seems like the end result is the same, but the dynamic link option is more work for the creator of the link. for all other cases, the dynamic link would come through the normal deep links handler.

i want to minimise the amount of dead code, and i'm fairly sure this block will not be used. but if we're unsure then there's no harm to put it back (this is the less risky option for sure)?

src/app/reducers.ts
kathaypacific: is it harmful to have the whole url check? i wanted to avoid complexity here, the `handleDeepLink` saga has some custom logic for parsing the urls which i wanted to avoid here (it uses a deprecated `parse` method to parse the deep link)...

src/app/saga.ts
jeanregisser: Sounds good, thanks for checking everything, we should remove then 👍

src/app/reducers.ts
bakoushin: would that enable any link with `/share` pathname to be persisted? e.g. `https://evil-resource.net/share`

jeanregisser: Valora is only registered to handle `celo://` or specific `https://valoraapp.com` links.
So it won't ever handle `https://evil-resource.net/share`.

I guess we can always add a dynamic config, but would have been nice without it.

kathaypacific: do you feel strongly that we should avoid checking the domain @jeanregisser ? if so, without doing the dynamic config, i think it makes sense to do the same check as in the deep link handler - in which case we'd need to extract the logic.

what this gives us is the ability to handle `celo://wallet/share` and `https://valoraapp.com/jumpstart`, both of which i'm not sure we have immediate plans to support. it's not a lot of work to do so i don't mind doing it, but i guess i'm not totally convinced of the value

kathaypacific: although writing it now, i think we should use a dynamic link for the jumpstart links rather than the `celo://` link :|

kathaypacific: i'll go ahead and merge this, and update the links once we know more

PR #4793, opened 2024-01-23
feat(swap): show estimated network fee

jeanregisser:
### Description

As the title says.

Next up: update analytics to include the estimated fee.

### Test plan

- Updated tests

| Before | After |
|--------|-------|
| image | image |
| image | image |

### Related issues

- Fixes RET-991

### Backwards compatibility

Yes


src/viem/prepareTransactions.ts
```diff
@@ -471,27 +471,37 @@ export function prepareSendNativeAssetTransaction(
...
   } else if (prepareTransactionsResult?.type === 'need-decrease-spend-amount-for-gas') {
     feeCurrency = prepareTransactionsResult.feeCurrency
-    feeAmountInDecimal = prepareTransactionsResult.maxGasFeeInDecimal
+    maxFeeAmount = prepareTransactionsResult.maxGasFeeInDecimal
+    estimatedFeeAmount = prepareTransactionsResult.estimatedGasFeeInDecimal
   }
   return {
-    feeAmount: feeAmountInDecimal,
     feeCurrency,
+    maxFeeAmount,
```
dievazqu, line 503: nit: Wondering if we should include `InDecimal` as part of these names to avoid confusion.

dievazqu (approved):
LGTM

src/viem/prepareTransactions.ts
jeanregisser: ah, I think most (all?) variables which have `amount` in their name currently are in decimal.
but maybe we should make this an explicit rule?

PR #4796, opened 2024-01-23
feat(fees): get proper baseFeePerGas on Celo for the estimated gas fee

jeanregisser:
### Description

As the title says.

This gets us another ~40% decrease in estimated gas fee on Celo.

Typical `baseFeePerGas` with CELO as a `feeCurrency` is `5 gwei`, when the recommended `maxFeePerGas` is
typically `12 gwei` (10 (`gasPrice`) + 2 (`maxPriorityFeePerGas`)).

So `1 - 7 / 12 = ~0.41`

Of course it's already less than a cent, but now it's right 🤓 

### Test plan

- Updated tests

| Before | After |
|--------|-------|
| image | image |

### Related issues

- Fixes RET-991

### Backwards compatibility

Yes

### DIFF
--- a/src/viem/estimateFeesPerGas.ts
+++ b/src/viem/estimateFeesPerGas.ts
@@ -47,7 +47,7 @@ export async function estimateFeesPerGas(
 
 // Get gas price with optional fee currency, this is Celo specific
 // See https://docs.celo.org/developer/viem#gas-price
-export async function getGasPrice(client: Client, feeCurrency: Address | undefined) {
+async function getGasPrice(client: Client, feeCurrency: Address | undefined) {
   const priceHex = await client.request({
     method: 'eth_gasPrice',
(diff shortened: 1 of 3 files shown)


src/viem/estimateFeesPerGas.ts
```diff
@@ -58,3 +63,28 @@ export async function getMaxPriorityFeePerGas(client: Client, feeCurrency?: Addr
   })
   return hexToBigInt(maxPriorityFeePerGasHex)
 }
+
+// Get gas price minimum with optional fee currency, this is Celo specific
+export async function getGasPriceMinimum(client: Client, feeCurrency: Address | undefined) {
```
dievazqu, line 68: nit: Since this is `Celo` specific, how do feel about adding `Celo` to the method name?
Something like: `getCeloGasPriceMinimum`

dievazqu (approved):
LGTM!
Ignore the comment if you like!

PR #4798, opened 2024-01-24
fix(imported-tokens): Fetching imported token balances in the wallet

dievazqu:
### Description

The backend sometimes doesn't return the balance of the imported token (for example, in Celo we only fetch the balance for the tokens we have in the address-metadata repo).
So, we have to manually fetch the balances in the wallet.

### Related issues

- Fixes RET-891

### Backwards compatibility

Y


src/tokens/saga.ts
```diff
@@ -309,6 +312,62 @@ export function* watchAccountFundedOrLiquidated() {
...
+        address: importedToken!.address as Address,
+        client: {
+          public: publicClient[networkIdToNetwork[importedToken.networkId]],
+        },
+      })
+
+      balanceRequests[importedToken.tokenId] = contract.read.balanceOf([address])
+    })
+
+    const balances = (yield* all(balanceRequests)) as Record<string, bigint>
```
dievazqu, line 345: I have tried to execute all these requests using the multicall method, but I can't find why it's not working.
I got some `undefined is not a function` error that I can't understand where is coming from (it seems like inside the multicall call).
I think this is fine if the user only have a few imported tokens.

src/tokens/saga.ts
```diff
@@ -309,6 +333,50 @@ export function* watchAccountFundedOrLiquidated() {
...
+          .toFixed()
+      }
+
+      importedTokensWithBalance[importedToken.tokenId] = {
+        ...importedToken,
+        balance,
+      }
+    })
+
+    await Promise.all(balanceRequests)
```
dievazqu, line 372: I tried to execute these requests with a `multicall`, but I don't know what I'm doing wrong, I can't make it work.
I got a `undefined is not a function` when I tried, I can't even get the balance using the `multicall` function.

I think the current implementation works so we can merge it, but I think we should implement the multicall, so maybe I can make a another branch/PR with the code that throws the error and someone can help me find out the error?

src/import/saga.ts
```diff
@@ -241,7 +241,7 @@ function* attemptBackupPhraseCorrection(mnemonic: string) {
 function* walletHasBalance(address: string) {
   Logger.debug(TAG + '@walletHasBalance', 'Checking account balance')
   const tokenBalances: FetchedTokenBalance[] = yield* call(fetchTokenBalancesForAddress, address)
-  return tokenBalances.length > 0
+  return tokenBalances.filter((token) => token.balance !== '0').length > 0
```
dievazqu, line 244: The backend returns some tokens even with zero balance.

src/tokens/saga.ts
```diff
@@ -309,6 +333,50 @@ export function* watchAccountFundedOrLiquidated() {
...
+    const importedTokensWithBalance: StoredTokenBalances = {}
+
+    const balanceRequests = importedTokensList.map(async (importedToken) => {
+      if (!importedToken) {
+        return
+      }
+
+      let balance
+      if (knownTokenBalances[importedToken.tokenId]) {
+        balance = knownTokenBalances[importedToken.tokenId].balance
```
kathaypacific, line 352: do we need to divide this by the token decimals?

src/tokens/saga.ts
kathaypacific: @jeanregisser maybe you have an idea? 

i also wonder if we should have a try / catch inside each `balanceRequest` promise - it seems like right now, if something fails inside any promise then none of the balances will be updated? (i assume `getContract` and / or `contract.read.balance` can throw, maybe i'm wrong about this?)

kathaypacific (commented):
looks good! do we still need the `importToken` action to trigger a balance refresh? i was thinking that in the import token screen, we already fetch the balance. so why not pass the balance as part of the `importToken` action, and then we don't need to fetch the balances for everything again? there's no harm in fetching balances, but it is background work / consumes mobile bandwidth. since the imported tokens will not have a fiat balance, triggering a full balance refresh should not produce any visible results for the user?

src/tokens/saga.ts
dievazqu: Branch with an example code: https://github.com/valora-inc/wallet/tree/multicall-error-imported-tokens

src/tokens/saga.ts
dievazqu: Yes! You are right!

dievazqu:
> looks good! do we still need the `importToken` action to trigger a balance refresh? i was thinking that in the import token screen, we already fetch the balance. so why not pass the balance as part of the `importToken` action, and then we don't need to fetch the balances for everything again? there's no harm in fetching balances, but it is background work / consumes mobile bandwidth. since the imported tokens will not have a fiat balance, triggering a full balance refresh should not produce any visible results for the user?

Yeah, I guess we could avoid refetching by passing the balance we fetched in the import token screen. I feel that logic is a bit unnecessary if at the end we also need the logic to fetch the balances.
But I guess with this we can show the balance faster the first time, so the user doesn't have to wait until the fetching is done.

src/import/saga.ts
jeanregisser: Good catch! Maybe this could have been done separately to give higher visibility into this fix?

src/tokens/saga.ts
```diff
@@ -177,6 +178,18 @@ export async function fetchTokenBalancesForAddress(
...
 }
 
+export async function fetchTokenBalancesForAddressByTokenId(address: string) {
+  const fetchedTokenBalances: FetchedTokenBalance[] = await fetchTokenBalancesForAddress(address)
+  const fetchedBalancesByTokenId = fetchedTokenBalances.reduce(
+    (acc: Record<string, FetchedTokenBalance>, token) => {
+      acc[token.tokenId] = token
+      return acc
+    },
+    {}
```
jeanregisser, line 188: nit: pet peeve of mine, reduce makes my eyes bleed. I think it's nice that's it succint, but it's usually harder to read. And code is read more than written.

Please simple loops 😢

src/tokens/saga.ts
jeanregisser: oh weird, the multicall works on [blockchain-api](https://github.com/valora-inc/blockchain-api/blob/0db7ebaab4cfdf04fc0c1c726926d6c186dd6287/src/blockchain.ts#L25),
maybe and env specific issue with React Native? what's the `undefined` function it complains about?

src/tokens/saga.ts
```diff
@@ -221,7 +229,22 @@ export function* fetchTokenBalancesSaga() {
...
     }
-    yield* put(setTokenBalances(tokens))
+
+    const importedTokensWithBalance = showImportedTokens
+      ? yield* call(
+          fetchImportedTokenBalances,
+          address as Address,
+          importedTokens,
+          fetchedBalancesByTokenId
+        )
```
jeanregisser, line 239: I'm slightly confused, we're fetching balances for imported tokens also passing a variable saying balances are already fetched since it's called `fetchBalancesByTokenId`?

oh ok it seems it's in case we already have the balance in the known set.

jeanregisser (commented):
Great! 👍

src/tokens/saga.ts
dievazqu: 🤣 
Yeah, I don't like `reduce` neither. I'll change it.

src/import/saga.ts
dievazqu: Ok, I'll put this in another PR :)

src/tokens/saga.ts
dievazqu: Yeah, I copied the code from there, and it's not working 🙁 
I don't know what's `undefined`, `client.multicall` exists, but when I call it, it throws. So it must be happening inside the `multicall`.

src/tokens/saga.ts
dievazqu: Yes, sorry, the logic is a bit weird. But the backend can return the balance for some tokens, so to avoid fetching the balance again I decided this approach.

Maybe I overthought this, and fetching the balance again was just fine and the code would be clearer.

dievazqu:
@jeanregisser @kathaypacific Thanks for the review!
The PR is ready for another look!

src/tokens/saga.ts
kathaypacific: i had a quick look out of curiousity. we're on different versions of viem (blockchain-api is still on v1, wallet is on v2). the multicall in the wallet throws at this block https://github.com/wevm/viem/blob/e2ed0690dd6112e3b46779044d85f9c147847d6e/src/actions/public/multicall.ts#L208-L223 complaining that `"aggregate3Results" is read-only`. i don't really know how to interpret this, but this block is a little different to the one in v1 of viem (https://github.com/wevm/viem/blob/d10b8015104169de550f50b4e562e6279e70596e/src/actions/public/multicall.ts#L175-L186)

anyway, i just wanted to note this in case it makes sense to either of you. you can debug the multicall with console statements in `node_modules/viem/_cjs/actions/public/multicall.js` @dievazqu

src/tokens/saga.ts
```diff
@@ -197,22 +207,17 @@ export function* fetchTokenBalancesSaga() {
...
-    const importedTokens = (yield* call(
+    const showImportedTokens = yield* call(
       getFeatureGate,
       StatsigFeatureGates.SHOW_IMPORT_TOKENS_FLOW
-    ))
-      ? yield* select(importedTokensInfoSelector)
-      : {}
-    const supportedTokens = yield* call(getTokensInfo)
+    )
+    const importedTokens = yield* select(importedTokensInfoSelector)
```
kathaypacific, line 214: i wonder if this selector should just return `[]` if the feature flag is off. that would remove the need for every place that consumes imported tokens to also check the flag, and would remove the need to check this flag later in this function (calling `fetchImportedTokenBalances` with an empty array seems okay?)

src/tokens/saga.ts
kathaypacific: yeah it took me a while to understand this too (essentially this happens because blockscout supports some tokens that we do not support, right?) - can you please add a comment to avoid future head scratching?

src/tokens/saga.ts
```diff
@@ -309,6 +329,54 @@ export function* watchAccountFundedOrLiquidated() {
...
+      } catch (error) {
+        Logger.error(TAG, 'Error fetching imported token balance', error)
+      }
+    })
+
+    await Promise.all(balanceRequests)
+    return importedTokensWithBalance
+  } catch (error) {
+    Logger.error(TAG, 'Error fetching imported tokens balances', error)
+    return {}
```
kathaypacific, line 376: now that you've added a try/catch inside each balance promise, does this overall try/catch still need to be here? i don't see anything else that can throw?

src/tokens/slice.ts
```diff
@@ -149,19 +149,30 @@ const slice = createSlice({
...
-          // Force imported tokens to be visible even with zero balance.
-          showZeroBalance: true,
-          balance: null,
-          isManuallyImported: true,
+    importToken: (state, action: PayloadAction<StoredTokenBalance>) => {
+      const importedTokenDetails = {
+        ...action.payload,
+        // Force imported tokens to be visible even with zero balance.
+        showZeroBalance: true,
+        isManuallyImported: true,
```
kathaypacific, line 157: do you have plans to add this to analytics somehow separately? (or maybe it is already included by default?)

src/tokens/slice.ts
```diff
@@ -149,19 +149,30 @@ const slice = createSlice({
...
+        tokenBalances: {
+          ...state.tokenBalances,
+          [action.payload.tokenId]: importedTokenDetails,
         },
-      },
-    }),
+        importedTokens: {
+          ...state.importedTokens,
+          [action.payload.tokenId]: {
+            ...importedTokenDetails,
```
kathaypacific, line 169: why are we storing the imported token details in both `importedTokens` and `tokenBalances`? we should really only have one source of truth in redux, perhaps the importedTokens state only needs to have a list of id's if you're relying on `tokenBalances` for the token info?

kathaypacific (commented):
looks good, i had a few small comments but the blocking one is the duplicated state in redux which i feel strongly that we should update (or discuss) :)

src/tokens/saga.ts
dievazqu: Yes, I'll add a comment

src/tokens/saga.ts
dievazqu: Oops, sorry, yes I can remove this try/catch

src/tokens/slice.ts
dievazqu: I interpret these fields as:
`tokenBalances` the source of truth for showing tokens with balances.
`importedTokens` the source of truth for the static data of the imported tokens.

But I understand we could derive `importedTokens` from `tokenBalances`, but somehow I feel it changes the meaning of `tokenBalances`, because now it's not only showing tokens with balance but also stores information about the imported tokens. But maybe we are fine we that change.

src/tokens/saga.ts
dievazqu: Is it ok to fetch the ff inside a selector??
The ff doesn't change without an app restart, right?
So I guess yes, it's fine.

src/tokens/slice.ts
dievazqu: I can add this to analytics.
In which events you think we should include this?
I'm thinking in the send flow, but maybe you had another flows in mind?

src/tokens/slice.ts
dievazqu: I made this change [here](https://github.com/valora-inc/wallet/pull/4798/commits/84cb2708bfe70debe00c50a8e949ea65a2b3e032).

I think you were right, we can live without the `importedTokens` field.

src/tokens/slice.ts
dievazqu: I created a [PR](https://github.com/valora-inc/wallet/pull/4809) for this.

src/tokens/slice.ts
```diff
@@ -149,19 +145,22 @@ const slice = createSlice({
...
+        // Force imported tokens to be visible even with zero balance.
+        showZeroBalance: true,
+        isManuallyImported: true,
+      }
+
+      return {
+        ...state,
+        tokenBalances: {
+          ...state.tokenBalances,
+          [action.payload.tokenId]: importedTokenDetails,
```
kathaypacific, line 160: yay awesome!! so now the redux property `importedTokens` is no longer needed right? can we also remove that with a redux migration in this PR?

src/tokens/selectors.ts
```diff
@@ -495,3 +493,11 @@ export const feeCurrenciesWithPositiveBalancesSelector = createSelector(
 
 export const visualizeNFTsEnabledInHomeAssetsPageSelector = (state: RootState) =>
   state.app.visualizeNFTsEnabledInHomeAssetsPage
+
+export const importedTokensSelector = createSelector([tokensListSelector], (tokenList) => {
+  if (!getFeatureGate(StatsigFeatureGates.SHOW_IMPORT_TOKENS_FLOW)) {
```
kathaypacific, line 498: great :D i think this is fine, since the value should be updated once per app session. another way we can do it without this assumption is to create a dummy selector for the feature flag to handle if the value does update (e.g. https://github.com/valora-inc/wallet/blob/fix-imported-tokens-balances/src/positions/selectors.ts#L17)

src/tokens/selectors.test.ts
```diff
@@ -733,4 +734,61 @@ describe('swappableTokensByNetworkIdSelector', () => {
...
+    it('returns all the imported tokens', () => {
+      const importedTokens = importedTokensSelector(mockState, [
+        NetworkId['celo-alfajores'],
+        NetworkId['ethereum-sepolia'],
+      ])
+
+      expect(importedTokens).toMatchObject([
+        {
+          ...mockState.tokens.tokenBalances['celo-alfajores:importedToken'],
+          balance: new BigNumber(0),
```
kathaypacific, line 774: i wondered for a moment if this is a special assertion that the balance should always be 0 when returned from the imported tokens selector. then i realised the assertion is on the big number type. perhaps we can add a non zero balance for one of these tokens?

src/tokens/saga.ts
```diff
@@ -309,6 +325,49 @@ export function* watchAccountFundedOrLiquidated() {
...
+
+      const balance = new BigNumber(fetchedBalance).shiftedBy(-importedToken.decimals).toFixed()
+
+      importedTokensWithBalance[importedToken.tokenId] = {
+        ...importedToken,
+        balance,
+        priceUsd: undefined,
+      }
+    } catch (error) {
+      Logger.error(TAG, 'Error fetching imported token balance', error)
```
kathaypacific, line 363:
```suggestion
      Logger.error(TAG, `Error fetching imported token balance with address ${importedToken?.address}` error)
```
(in case the error does not mention this?)

src/tokens/saga.ts
```diff
@@ -221,7 +221,23 @@ export function* fetchTokenBalancesSaga() {
           .toFixed()
       }
     }
-    yield* put(setTokenBalances(tokens))
+
+    /* We are including the fetchedBalancesByTokenId since some balances might be already fetched
+     * so we avoid fetching them again. This could happen if we turn the FETCH_BALANCES_VIA_BLOCKSCOUT flag on.
```
kathaypacific, line 226:
```suggestion
     * so we avoid fetching them again. This could happen if the data source includes more tokens than we support (e.g. Blockscout).
```

src/tokens/saga.test.ts
```diff
@@ -184,75 +216,184 @@ describe(fetchTokenBalancesSaga, () => {
...
-    })
+    const importedTokens = Object.values(mockImportedTokensInfo)
 
     await expectSaga(fetchTokenBalancesSaga)
       .provide([
         [call(getTokensInfo), mockBlockchainApiTokenInfo],
-        [select(importedTokensInfoSelector), mockImportedTokensInfo],
+        [select(importedTokensSelector, supportedNetworks), importedTokens],
         [select(walletAddressSelector), mockAccount],
+        [call(getFeatureGate, StatsigFeatureGates.SHOW_IMPORT_TOKENS_FLOW), true],
```
kathaypacific, line 226: is this mock still needed?

kathaypacific (approved):
looks great!! approving but i think we should remove the now unused `importTokens` property with a redux migration before merging

src/tokens/saga.test.ts
dievazqu: Nope, thanks!

PR #4799, opened 2024-01-24
fix: no usd price warning for swaps

kathaypacific:
### Related issues

- Fixes #[issue number here]

### DIFF
--- a/locales/base/translation.json
+++ b/locales/base/translation.json
@@ -1711,7 +1711,8 @@
     },
     "noUsdPriceWarning": {
       "title": "{{localCurrency}} price unavailable",
-      "description": "We do not have a {{localCurrency}} price for this token, which might mean that it's not widely traded. {{appName}} is not responsible for any losses incurred if you choose to proceed.",
+      "descriptionTokenSelection": "We do not have a {{localCurrency}} price for this token, which might mean that it's not widely traded. {{appName}} is not responsible for any losses incurred if you choose to proceed.",
+      "descriptionSwap": "We do not have a {{localCurrency}} price for {{toTokenSymbol}}, which might mean that it's not widely traded. {{appName}} is not responsible for any losses incurred if you choose to proceed.",
       "ctaConfirm": "I understand",
       "ctaDismiss": "Cancel"
     },
--- a/src/swap/SwapScreen.tsx
+++ b/src/swap/SwapScreen.tsx
@@ -536,7 +536,7 @@ export function SwapScreen({ route }: Props) {
   }
 
   const handleSelectToken = (selectedToken: TokenBalance) => {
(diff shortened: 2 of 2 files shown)


locales/base/translation.json
```diff
@@ -1711,7 +1711,7 @@
     },
     "noUsdPriceWarning": {
       "title": "{{localCurrency}} price unavailable",
-      "description": "We do not have a {{localCurrency}} price for this token, which might mean that it's not widely traded. {{appName}} is not responsible for any losses incurred if you choose to proceed.",
+      "description": "We do not have a {{localCurrency}} price for {{tokenSymbol}}, which might mean that it's not widely traded. {{appName}} is not responsible for any losses incurred if you choose to proceed.",
```
jeanregisser, line 1714: No live version uses it yet, right?
so no need for backward compatibility?

jeanregisser (approved):
LGTM 👍 

> i found it misleading to see this warning when selecting a token without usd price as a "from" token.

I'm not sure about this, maybe I need to experience this first hand.

src/swap/SwapScreen.tsx
```diff
@@ -725,7 +721,10 @@ export function SwapScreen({ route }: Props) {
             <InLineNotification
               severity={Severity.Warning}
               title={t('swapScreen.noUsdPriceWarning.title', { localCurrency })}
-              description={t('swapScreen.noUsdPriceWarning.description', { localCurrency })}
+              description={t('swapScreen.noUsdPriceWarning.description', {
+                localCurrency,
+                tokenSymbol: toToken?.symbol,
+              })}
```
bakoushin, line 727: super nit: do we need `?`?

kathaypacific: yes, toToken can be undefined. in reality, it won't be if this warning is displayed. i either need optional chaining here or to use `!` (i think optional chaining is less bad)

kathaypacific:
> LGTM 👍
> 
> > i found it misleading to see this warning when selecting a token without usd price as a "from" token.
> 
> I'm not sure about this, maybe I need to experience this first hand.

yes maybe it's not so bad, i'm not sure that we need it for the guardrail that we are trying to build here though. i attached a video, i guess my concern is that this warning makes it seem like users should not swap out of these tokens but ideally you don't really want to hold a lot of tokens that cannot be priced.

if a user does proceed to swap with a "from" token without usd price, i think most times we will show the high price impact warning (which makes sense if we cannot price due to low liquidity). of course, not guaranteed


https://github.com/valora-inc/wallet/assets/20150449/7c31c582-8663-4b01-a654-49261db7fa9b

bakoushin (approved):
⛱️

PR #4801, opened 2024-01-25
fix(ci): ensure secure send transaction has unique comment

kathaypacific:
### Description

Context https://valora-app.slack.com/archives/C02E2FE98P2/p1706173505572049

This PR updates the secure send flow transaction comments to be unique. I actually think that this change makes our tests more robust anyway (it ensures that you are finding the correct transaction, whereas with the current implementation, the test comment could be missing but the tests could pass by finding a previous transaction).

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y

### DIFF
--- a/.github/workflows/e2e-ios.yml
+++ b/.github/workflows/e2e-ios.yml
@@ -63,7 +63,7 @@ jobs:
           DETOX_CONFIG: ios.release
       - name: Run Detox
         run: >
-          yarn detox test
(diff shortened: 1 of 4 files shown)


jeanregisser (approved):
🚀

MuckT:
@kathaypacific this can be closed since #4804 was merged.

PR #4802, opened 2024-01-25
chore(swap): separate to and from token lists

kathaypacific:
### Description

Context https://valora-app.slack.com/archives/C02UAE4MGUF/p1705095677062069

This PR decouples the to and from tokens list. Both will read from address-metadata for whether the token should be swappable, but we should always allow users to swap out of positions (whilst preventing users from swapping into them).

The logic for the "to" tokens has not changed, but the "from" tokens now include all (supported) tokens that the user has a balance of.

We can mark some tokens as not swappable in address-metadata after this is released.

### Test plan

1. acquire some balance of a token that is not swappable
2. see that you can swap out of it, but not into it

### Related issues

n/a

### Backwards compatibility

Y


src/tokens/selectors.ts
```diff
@@ -321,47 +321,64 @@ export const tokensWithTokenBalanceSelector = createSelector(
   }
 )
 
-export const swappableTokensByNetworkIdSelector = createSelector(
+// sort by balance USD (DESC) then name (ASC), tokens without a priceUsd
+// are pushed last, sorted by name (ASC)
+const sortSwapTokens = (token1: TokenBalance, token2: TokenBalance) => {
```
kathaypacific, line 326: there are no logic changes here, just copy and paste

src/swap/SwapScreen.test.tsx
```diff
@@ -99,7 +99,10 @@ const mockStoreTokenBalances = {
     priceUsd: '13.05584965485329753569',
   },
   [mockTestTokenTokenId]: {
-    ...mockTokenBalances[mockTestTokenTokenId],
```
kathaypacific, line 102: test token is not part of the mock token balances object actually, and other tests rely on this object so i didn't want to change it here

src/tokens/selectors.ts
```diff
@@ -321,47 +321,64 @@ export const tokensWithTokenBalanceSelector = createSelector(
...
-            tokenInfo.isSwappable ||
-            (tokenInfo.minimumAppVersionToSwap &&
-              !isVersionBelowMinimum(appVersion, tokenInfo.minimumAppVersionToSwap))
+    return tokens
+      .filter((tokenInfo) => {
+        return (
+          tokenInfo.isSwappable ||
+          (tokenInfo.minimumAppVersionToSwap &&
+            !isVersionBelowMinimum(appVersion, tokenInfo.minimumAppVersionToSwap)) ||
+          tokenInfo.balance.gt(0)
```
bakoushin, line 363: this appear to be the only difference between the two selectors. what do you think about extracting the shared logic into a separate selector?

bakoushin:
i remember discussing with Kayla that we want to show in the FROM list all supported tokens in case the user has a zero balance.
if that still holds, probably the better naming would be not tied to the filed name they are used (to/from) but the specifics of returned value (all/only with balance)?

src/tokens/selectors.ts
bakoushin: i noticed that in [some places](https://github.com/valora-inc/wallet/blob/845274bbf8d662bd4e0b039bed76c578d3641892/src/tokens/selectors.ts#L320) within this file we consider non-zero balance to be greater than `TOKEN_MIN_AMOUNT`. should we follow this pattern here as well?

kathaypacific: thank you for the suggestions! makes sense, i applied both your suggestions :)

kathaypacific:
> i remember discussing with Kayla that we want to show in the FROM list all supported tokens in case the user has a zero balance to avoid the awkward situation when just onboarded user is poking around the app and gets an empty list. if that still holds, probably the better naming would be not tied to the filed name they are used (to/from) but the specifics of returned value (all/only with balance)?

i'm not sure that we should show all tokens in the swap flow. i agree that we should show more than just what the user has balance for, for exploration purposes but address-metadata holds many tokens that we don't want to support any functionality for besides being able to see your balance (and liquidate) if you already have it. an example is [SavingsCelo](https://docs.savingscelo.com/) which has a known vulnerability, it seems like we shouldn't show that in the app at all (swap, send, etc.) unless the user already has it?

bakoushin:
yeah, the idea is always show in FROM only those tokens user owns, except when user has 0 balance. instead of showing an empty list in that case, we can show unfiltered list of all available tokens for exploration purposes.

as for failed tokens, should we filter them out from TO list in the first place? if so, should they have any property indicating that they are not welcome? (or should we maintain a separate list of such tokens?)

kathaypacific:
> yeah, the idea is always show in FROM only those tokens user owns, except when user has 0 balance. instead of showing an empty list in that case, we can show unfiltered list of all available tokens for exploration purposes.
> 
> as for failed tokens, should we filter them out from TO list in the first place? if so, should they have any property indicating that they are not welcome? (or should we maintain a separate list of such tokens?)

hmm i'm not understanding - what i have implemented in this PR is the "from" tokens will be all swappable tokens + any unswappable tokens that the user already has a balance of. what is the behaviour that you are suggesting instead? are you saying that for a user with some token balance => the "from" token selection should _only_ show the tokens that the user has a balance of? (e.g. i only have cUSD, so the token selection sheet shows only cUSD?)

address-metadata gives us all the tokens that we "support". for each token, there are also some properties that we can use to exclude it from swaps (these are the ones we are using to exclude the tokens from the "to" field)

jeanregisser (approved):
🚀

jeanregisser:
Just double checking, this means users won't be able to get quotes for tokens they don't have a balance of?

jeanregisser:
Thanks, in that case we shouldn't do the `TOKEN_MIN_AMOUNT` change then.

kathaypacific:
> Thanks, in that case we shouldn't do the `TOKEN_MIN_AMOUNT` change then.

why is that? i suppose it's up to us to define what "has balance" means, the TOKEN_MIN_AMOUNT is a really small value (that i guess represents anything we see as dust)? the fiat value of a token would have to be over $1 million for the  TOKEN_MIN_AMOUNT to be $0.01 in fiat value, which doesn't seem like worth swapping. we don't show the token in the assets screen with a balance below this, so i thought it was more consistent to not surface the tokens on the swap page too

jeanregisser:
But didn't you said we still want users to be able to get quotes for tokens they don't have?
i.e. I don't have token XXX but I still want to see the quote from XXX to YYY?
this was my original question.

jeanregisser:
oh ok the `MIN_TOKEN_AMOUNT` was just for the users own token, it won't remove tokens that the user doesn't have.
All good sorry 👍

src/tokens/selectors.ts
```diff
@@ -321,7 +321,7 @@ export const tokensWithTokenBalanceSelector = createSelector(
   }
 )
 
-export const swappableTokensByNetworkIdSelector = createSelector(
+export const swappableFromTokensByNetworkIdSelector = createSelector(
   (state: RootState, networkIds: NetworkId[]) => tokensListSelector(state, networkIds),
```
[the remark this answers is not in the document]
jeanregisser: Sorry this was a wrong suggestion.

PR #4807, opened 2024-01-26
fix(balances): Fixing walletHasBalance

dievazqu:
### Description

`fetchTokenBalancesForAddress` can return tokens with zero balance (in the ethereum flow for example), so `walletHasBalance` returns `true` where it should return `false`.


jeanregisser (approved):

bakoushin (approved):
LGTM!

PR #4811, opened 2024-01-26
feat(keyless-backup): Use decrypted mnemonic to restore wallet

jh2oman:
### Description

Use the decrypted mnemonic to restore wallet. Needs some UI pieces hooked up before e2e testing can be done. 

Screenshot_20240126_134521
Screenshot_20240126_134556

### Test plan

unit tests updated

### Related issues

https://linear.app/valora/issue/ACT-778/wallet-code-for-completing-recovery-with-decrypted-mnemonic


src/keylessBackup/saga.ts
```diff
@@ -110,8 +119,49 @@ function* handleKeylessBackupRestore({
...
+    }
+  }
+
+  const account: string | null = yield* call(
+    assignAccountFromPrivateKey,
+    privateKey,
+    decryptedMnemonic
+  )
+  if (!account) {
+    throw new Error('Failed to assign account from private key')
```
satish-ravi, line 155: when can this happen? could this result in a crash? Wonder if this should be handled more gracefully

src/keylessBackup/saga.ts
```diff
@@ -110,8 +119,49 @@ function* handleKeylessBackupRestore({
...
+  }
+  // Set key in phone's secure store
+  yield* call(storeMnemonic, decryptedMnemonic, account)
+  // Set backup complete so user isn't prompted to do backup flow
+  yield* put(setBackupCompleted())
```
satish-ravi, line 160: I guess we have two "backup completed" states now, one for the quiz and one for the cloud backup. We check both states before showing the backup flow. So is this required since we set cloud backup as completed below? If we do want to set this, then is this something that should be set when cloud backup is setup as well?

src/keylessBackup/saga.test.ts
```diff
@@ -235,15 +243,61 @@ describe('keylessBackup saga', () => {
...
+              mockMnemonic,
+            ],
+            [call(generateKeysFromMnemonic, mockMnemonic), { privateKey: mockPrivateKey }],
+            [call(walletHasBalance, privateKeyToAddress(mockPrivateKey)), false],
+          ])
+          .dispatch(keylessBackupBail())
+          .run()
+        expect(ValoraAnalytics.track).toBeCalledWith('cab_handle_keyless_backup_success', {
+          keylessBackupFlow: KeylessBackupFlow.Restore,
+        })
```
satish-ravi, line 299: this event shouldn't be fired if the user bails right?

src/keylessBackup/saga.ts
jh2oman: I don't know when this can happen. I just copied the code from the standard import mnemonic

src/keylessBackup/saga.test.ts
jh2oman: I debated this. We can either throw when the user bails and then we will log a failure or we can return when the user bails and log a success. I thought it was closer to a success since we successfully restored the wallet, the user just declined to use it.

src/keylessBackup/saga.ts
jh2oman: Ah, yea we check both. I don't think we should to / need to set the regular backup completed here. thanks!

src/keylessBackup/KeylessBackupProgress.tsx
```diff
@@ -35,9 +49,44 @@ function KeylessBackupProgress({
     return () => BackHandler.removeEventListener('hardwareBackPress', backPressListener)
   }, [])
 
-  // TODO(ACT-781): Implement Restore flow designs
   if (route.params.keylessBackupFlow === KeylessBackupFlow.Restore) {
-    return <></>
+    switch (keylessBackupStatus) {
+      case KeylessBackupStatus.NotStarted:
```
jophish, line 54: Is it possible for `keylessBackupStatus === KeylessBackupStatus.NotStarted` here? Not obvious to me why an in-progress section would show when the status is `NotStarted`, or maybe I'm missing something on how the status is tracked.

src/keylessBackup/KeylessBackupProgress.tsx
```diff
@@ -35,9 +42,49 @@ function KeylessBackupProgress({
     return () => BackHandler.removeEventListener('hardwareBackPress', backPressListener)
   }, [])
 
-  // TODO(ACT-781): Implement Restore flow designs
   if (route.params.keylessBackupFlow === KeylessBackupFlow.Restore) {
-    return <></>
+    switch (keylessBackupStatus) {
```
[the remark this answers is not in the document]
jophish: +1, I think it's a little hard to follow with the two separate switch statements spaced out from eachother.

src/keylessBackup/saga.ts
```diff
@@ -110,8 +118,48 @@ function* handleKeylessBackupRestore({
...
+  if (!privateKey) {
+    throw new Error('Failed to convert mnemonic to hex')
+  }
+  const backupAccount = privateKeyToAddress(privateKey)
+  if (!(yield* call(walletHasBalance, backupAccount))) {
+    // show zero balance modal
+    yield* put(keylessBackupShowZeroBalance())
+
+    // wait for user to click continue or bail from keyless backup
+    const [_continueAction, bailAction] = yield* race([
```
jophish, line 138: There's no way for a user to exist from the dialog _without_ dispatching one of these actions, right? From the look of it, it seems like not, but wanted to check.

src/keylessBackup/saga.test.ts
jophish: We could alternatively return some sort of status from `handleKeylessBackupRestore` and log analytics in `handleValoraKeyshareIssued` based off the state, if we want more granular information on what the user chose. As it is now, we have no analytics indicating when a user bails, right? (we could log something when they dispatch the bail action from the dialog as well.)

src/keylessBackup/saga.ts
jh2oman: nope, no other way to move forward

src/keylessBackup/saga.test.ts
jh2oman: Decided to add analytics events at the component level when an action is dispatched.

src/keylessBackup/KeylessBackupProgress.tsx
```diff
@@ -24,21 +29,77 @@ const TAG = 'keylessBackup/KeylessBackupProgress'
...
+                  showLocalAmount={true}
+                  amount={new BigNumber(0)}
+                />
+              </Trans>
+            }
+            isVisible={true}
+            actionText={t('importExistingKey.emptyWalletDialog.action')}
+            actionPress={() => {
+              ValoraAnalytics.track(KeylessBackupEvents.cab_restore_zero_balance_accept)
+              dispatch(keylessBackupAcceptZeroBalance())
```
satish-ravi, line 78: This doesn't navigate away from the screen immediately right? We've had issues in the past where the screen hangs because the modal isn't dismissed completely before navigating

src/keylessBackup/KeylessBackupProgress.tsx
```diff
@@ -24,21 +29,77 @@ const TAG = 'keylessBackup/KeylessBackupProgress'
...
+function Restore() {
+  const keylessBackupStatus = useSelector(keylessBackupStatusSelector)
+  const localCurrencyCode = useLocalCurrencyCode()
+
+  const { t } = useTranslation()
+  const dispatch = useDispatch()
+
+  switch (keylessBackupStatus) {
+    case KeylessBackupStatus.InProgress: {
+      return InProgress(t('keylessBackupStatus.restore.inProgress.title'))
```
satish-ravi, line 55: nit: maybe make this a react component instead of a function? or use a lower case fn name? e.g., renderInProgress

src/keylessBackup/KeylessBackupProgress.tsx
```diff
@@ -24,21 +29,77 @@ const TAG = 'keylessBackup/KeylessBackupProgress'
...
+          >
+            {t('importExistingKey.emptyWalletDialog.description')}
+          </Dialog>
+        </SafeAreaView>
+      )
+    }
+    // TODO(ACT-781): Implement Success screen
+    // TODO(ACT-780): Implement Failure screens
+    default:
+      Logger.error(TAG, `Got unexpected keyless backup status: ${keylessBackupStatus}`)
```
satish-ravi, line 95: NotStarted will fall here right? Any reason why this is handled differently for setup and restore?

src/keylessBackup/saga.ts
satish-ravi: will this resume properly if the app is backgrounded when the modal is visible? What about background press?

src/keylessBackup/saga.ts
```diff
@@ -110,8 +118,48 @@ function* handleKeylessBackupRestore({
...
+    privateKey,
+    decryptedMnemonic
+  )
+  if (!account) {
+    throw new Error('Failed to assign account from private key')
+  }
+  // Set key in phone's secure store
+  yield* call(storeMnemonic, decryptedMnemonic, account)
+
+  yield* put(refreshAllBalances())
```
satish-ravi, line 159: is this required? seems like initializeAccountSaga also puts it

src/keylessBackup/KeylessBackupProgress.tsx
jh2oman: no it doesn't, there are a few async methods that complete before navigating. I will check this though.

src/keylessBackup/KeylessBackupProgress.tsx
jh2oman: I believe that the NotStarted state cannot be reached on this screen with the restore flow OR the setup flow. Could be a bug if it is, so I think we should throw an error.

src/keylessBackup/slice.ts
```diff
@@ -43,6 +43,15 @@ export const slice = createSlice({
...
       state.backupStatus = KeylessBackupStatus.Completed
     },
+    keylessBackupShowZeroBalance: (state) => {
+      state.backupStatus = KeylessBackupStatus.RestoreZeroBalance
+    },
+    keylessBackupAcceptZeroBalance: (state) => {
+      state.backupStatus = KeylessBackupStatus.InProgress
+    },
+    keylessBackupBail: (state) => {
+      state = initialState
```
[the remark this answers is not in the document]
jh2oman: I don't think so, but this does raise some red flags now that I see it. I'll be explicit with setting the fields.

src/keylessBackup/saga.ts
jh2oman: Just tested, backgroud press does nothing. App being backgrounded also has no effect.

src/keylessBackup/saga.ts
jh2oman: I don't think so. Its in the normal import flow so I kept it, but I think you're right its not necessary

src/keylessBackup/SignInWithEmail.tsx
```diff
@@ -33,6 +33,11 @@ function SignInWithEmail({ route }: Props) {
 
   const onPressGoogle = async () => {
     setLoading(true)
+    dispatch(
```
jh2oman, line 36: I moved  keylessBackupStarted here because it is the first screen that is part of both the Restore and Setup flow, so I thought it made sense for simplicity and consistency.

satish-ravi (approved):

PR #4813, opened 2024-01-26
chore: remove fiat exchange currency screen

MuckT:
### Description

- Removes the fiat exchange currency screen and uses the bottom sheet instead.
- Removes the `use_cico_currency_bottom_sheet` feature gate and related checks.

### Test plan

- [x] Tested on CI.
- [x] Tested locally on iOS
  -  Navigate to cash in `+ Buy` from home actions.
  -  Navigate to cash in from deeplink `celo://wallet/cashIn`

### Related issues

- Related ACT-1060

### Backwards compatibility

Yes


src/navigator/NavigationService.ts
```diff
@@ -250,10 +247,3 @@ export function navigateToError(errorMessage: string, error?: Error) {
   Logger.debug(`${TAG}@navigateToError`, `Navigating to error screen: ${errorMessage}`, error)
   navigate(Screens.ErrorScreen, { errorMessage })
 }
-
-export function navigateToFiatCurrencySelection(flow: FiatExchangeFlow) {
```
satish-ravi, line 254: Wonder if we should retain this wrapper method similar to `navigateHome`. We've had instances in the past where we forgot to update all entry points for a flow (new send flow).

src/fiatExchanges/FiatExchangeCurrency.tsx
```diff
@@ -1,270 +0,0 @@
...
-    DynamicConfigs[StatsigDynamicConfigs.MULTI_CHAIN_FEATURES]
-  ).showCico.includes(networkConfig.networkToNetworkId[Network.Ethereum])
-
-  // Fetch FiatConnect providers silently in the background early in the CICO funnel
-  useEffect(() => {
-    dispatch(fetchFiatConnectProviders())
-  }, [])
-
-  const goToProvider = () => {
-    ValoraAnalytics.track(FiatExchangeEvents.cico_currency_chosen, {
```
satish-ravi, line 120: can we also clean up this analytics event? We unfortunately don't get knip check because it's used in analytics/properties and analytics/docs

src/navigator/NavigationService.ts
MuckT: Thought about keeping it, but as there's only navigation to the bottom sheet which handles cash in cash out it seemed superfluous: `navigate(Screens.FiatExchangeCurrencyBottomSheet, { flow: FiatExchangeFlow.<CashIn|CashOut> })`.

src/fiatExchanges/FiatExchangeCurrency.tsx
MuckT: Fixed c41425ad26304b4429a68183e1d55bffa54c040d & a48ab6a2ae069bf946d45835e2fef6fcf124c78c!

src/analytics/docs.ts
```diff
@@ -316,10 +316,6 @@ export const eventDocs: Record<AnalyticsEventType, string> = {
   [FiatExchangeEvents.cico_landing_select_flow]: `User taps which CICO flow they want (Cash in, cash out, spend)`,
   [FiatExchangeEvents.cico_landing_how_to_fund]: `User taps help article link at the bottom`,
 
-  // The CICO Currency screen where the user selects their digital currency
-  [FiatExchangeEvents.cico_currency_chosen]: `User selects a currency and taps next`,
-  [FiatExchangeEvents.cico_currency_back]: `User presses the back button from the currency screen`,
```
cajubelt, line 321: there is a section at the bottom that is commented out with deleted events. kinda helps for when someone asks where an event comes from since we can tell it's from a past version of the app. can you add these to that section?

MuckT: Added in 99d0df6989e080ace8b547907abf82da6413d5e4!

PR #4828, opened 2024-01-30
feat(swap): add token bottom sheet filters

kathaypacific:
### Description

This PR:
- adds a reusable filter chip carousel (the component is largely taken from the dapps screen, i will separately update the dapps screen to use this component)
- updates the TokenBottomSheet component to render the filters if passed in as a prop
- updates the TokenBottomSheet to handle filter selection and default filters (alongside search terms)
- updates the SwapScreen with a hook that returns the filters. i did it like this because it seems reasonable to assume that we might want different filters or different rules for the filters depending on where the token selection is launched from (e.g. recently swapped would not be there in the send screen, but a "recently received" filter might be there instead)
- add a feature gate to turn on / off the filters

Not included in this PR: 
- the popular tokens. Still figuring out what those are, i'll add them in a follow up PR.
- analytics i'll add separately because this PR is big already

### Test plan

https://github.com/valora-inc/wallet/assets/20150449/0a495c31-4aba-4d05-b9b4-393f98f411ab

### Related issues

- Fixes RET-984

### Backwards compatibility

Y


src/components/TokenBottomSheet.tsx
```diff
@@ -101,17 +105,26 @@ export function TokenBalanceItemOption({
...
   searchTerm,
+  selectedFilters,
 }: {
   testID?: string
   searchTerm: string
+  selectedFilters: FilterChip<TokenBalance>[]
 }) {
+  const activeSearchTerms = [searchTerm, ...selectedFilters.map((filter) => filter.name)]
+    .filter((term) => !!term)
+    .join(', ')
```
bakoushin, line 116: matter of taste: i personally would prefer filters being listed separately. for example:

> No matching results found for "poof" within applied filters: "My tokens", "Popular".

src/components/FilterChipsCarousel.tsx
```diff
@@ -0,0 +1,99 @@
...
+          </View>
+        )
+      })}
+    </ScrollView>
+  )
+}
+
+const styles = StyleSheet.create({
+  container: {
+    marginHorizontal: -Spacing.Thick24,
```
jeanregisser, line 77: Side note: I guess here it's a legit use of including the margin within the component but in most cases it should be avoided: https://www.joshwcomeau.com/css/styled-components/#isolated-css-3

src/components/FilterChipsCarousel.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+export interface FilterChip<T> {
+  id: string
+  name: string
+  filterFn: (t: T) => boolean
+}
+
+interface Props<T> {
+  chips: FilterChip<T>[]
+  selectedChips: FilterChip<T>[]
```
jeanregisser, line 16: Could we use `selectedChipIds` instead? ensures there's a single source of truth for these.

src/components/TokenBottomSheet.tsx
```diff
@@ -36,9 +38,11 @@ interface Props<T extends TokenBalance> {
   tokens: T[]
   TokenOptionComponent?: React.ComponentType<TokenOptionProps>
   showPriceUsdUnavailableWarning?: boolean
+  filterChips?: FilterChip<TokenBalance>[]
+  preSelectedFilterChips?: FilterChip<TokenBalance>[]
```
jeanregisser, line 42: nit: this naming seems more familiar in a React component:
```suggestion
  initialSelectedChipIds?: string[]
```

src/swap/useFilterChips.ts
```diff
@@ -0,0 +1,76 @@
...
+import { networkIdToNetwork } from 'src/web3/networkConfig'
+
+export default function useFilterChip(selectingField: Field | null) {
+  const { t } = useTranslation()
+  const showSwapTokenFilters = getFeatureGate(StatsigFeatureGates.SHOW_SWAP_TOKEN_FILTERS)
+  const recentlySwappedTokens = useSelector(lastSwappedSelector)
+  const popularTokens: string[] = [] // TODO
+  const supportedNetworkIds = getSupportedNetworkIdsForSwap()
+
+  const preSelectedFilterChipIds =
```
jeanregisser, line 22:
```suggestion
  const initialSelectedChipIds =
```

src/swap/useFilterChips.ts
```diff
@@ -0,0 +1,76 @@
...
+          ]
+        : []),
+    ]
+  }, [t, recentlySwappedTokens, popularTokens, supportedNetworkIds, showSwapTokenFilters])
+
+  return {
+    filterChips,
+    preSelectedFilterChips: filterChips.filter((chip) =>
+      preSelectedFilterChipIds.includes(chip.id)
+    ),
```
jeanregisser, line 74:
```suggestion
    initialSelectedChipIds,
```

jeanregisser (commented):
Great work! 🚀 

About the possible perf issue, I guess we can profile and address separately?

kathaypacific:
TY for the reviews @bakoushin @jeanregisser 

The token bottom sheet is currently using a ScrollView instead of a list view that is optimised for long lists. That is the cause of the perf issues, and I'll address this separately to this PR to avoid this PR becoming very large.

I updated the no results message to separate search and filters, following the suggestion from Alex.

I've also modified the filters to have a single source of truth but keeping everything within one variable rather than having separately the filters and selected filters. i think this is less complex and addresses the concerns.

jeanregisser (approved):

src/swap/useFilterChips.ts
```diff
@@ -0,0 +1,67 @@
...
+      ? [
+          {
+            id: 'celo-network',
+            name: t('tokenBottomSheet.filters.celo'),
+            filterFn: (token: TokenBalance) => networkIdToNetwork[token.networkId] === Network.Celo,
+            isSelected: false,
+          },
+        ]
+      : []),
+    ...(supportedNetworkIds.includes(NetworkId['ethereum-mainnet']) ||
```
cajubelt, line 54: across all our repos, we should strive to make adding chains really easy. It should feel totally idiot-proof and mechanical to do. Should be as easy as:
1. add the network id to a NetworkId enum
2. fix all the compilation errors

The way this code is written won't support the above^. To make this feature work for a new chain, you would need to either search for every place in the app where NetworkId is used (already over a hundred-- easy to miss one even if you do this), or test everything in the app very carefully (also highly error-prone)

cajubelt: separately, this hook should have unit tests

cajubelt: another q-- it looks like even if supportedNetworkIds only had ethereum-mainnet, if there was some token with networkId ethereum-sepolia, the filter function would return 'true'. is that desireable?

dominiquejane:
I was really confused when I swiped over and tapped Celo Network and didn't see anything change. I would expect to only be able to select one chip at a time, instead of having multiple chips selected. I think it needs an icon, like a checkmark if we have multi-select.

In the Swap From list I also see assets that I have a zero value for and I don't think those should show up.

PR #4830, opened 2024-01-31
fix: display issues due to non unique keys for assets

jeanregisser:
### Description

As the title says.

### Test plan

- No more mixed items when switching between tabs in the assets screen

### Related issues

Slack [thread](https://valora-app.slack.com/archives/C04B61SJ6DS/p1706656376321169)

### Backwards compatibility

Yes

### DIFF
--- a/src/tokens/Assets.tsx
+++ b/src/tokens/Assets.tsx
@@ -332,11 +332,11 @@ function AssetsScreen({ navigation, route }: Props) {
 
   const keyExtractor = (item: TokenBalance | Position | NftWithNetworkId[], index: number) => {
     if (assetIsPosition(item)) {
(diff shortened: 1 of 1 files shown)


kathaypacific (approved):

src/tokens/Assets.tsx
```diff
@@ -330,13 +330,16 @@ function AssetsScreen({ navigation, route }: Props) {
     return null
   }
 
-  const keyExtractor = (item: TokenBalance | Position | NftWithNetworkId[]) => {
```
cajubelt, line 333: what about a regression test?

jeanregisser: Yes would be nice, was afraid it would be a time sink though given the complexity of RN’s list view rendering.

PR #4833, opened 2024-01-31
refactor(dapps): use filter chip component and simplify logic

kathaypacific:
### Description

This PR should have no user facing features. The refactor:
1. use the new filter carousel component. since this component uses filters in a different / more generic way than the screen currently does, the set + apply filter logic was slightly tweaked.
2. i noticed some unnecessary complexity in how the screen structured - it uses a SectionList but the only section is "All" (and the "favourites" section is a whole other component). this means that a lot of the filter set / reset / apply logic + handling of empty states has an extra step. in this PR, i remove the "favourites" component and use a `useMemo` hook to create the sections.

I added a small fix:
`id` is a special mixpanel event property, so i have renamed the filter id to `filterId` (the previously set value for `id` was overwritten and never captured in mixpanel)

### Test plan

The scenarios I verified:
1. apply a filter
a) if results in "all" and "favourites" - show results in both
b) if results in "all" and no results in favourites, whole favourites section disappears.
c) if results in "favourites" and no results in "all", display "no matching results" in the "all" section + favourites results.
d) if no results in either section, the section title is "my favourites & all" and show "no matching results"
1. clear a filter 
1. search is applied to the filtered results
1. tapping on a dapp from each section should dispatch analytics results from the correct "openedFrom" value

https://github.com/valora-inc/wallet/assets/20150449/8a7b7ce0-bc88-4ac1-be92-b3bdb517def1

### Related issues

- Related to RET-984

### Backwards compatibility

Y


src/dappsExplorer/DAppsExplorerScreenSearchFilter.tsx
```diff
@@ -69,45 +70,48 @@ export function DAppsExplorerScreenSearchFilter() {
...
-  const [favoriteResultsEmpty, setFavoriteResultsEmpty] = useState(false)
-  const [allResultEmpty, setAllResultEmpty] = useState(false)
+  const [filterChips, setFilterChips] = useState<FilterChip<DappWithCategoryNames>[]>(
+    categories.map((category) => ({
+      id: category.id,
+      name: category.name,
+      filterFn: (dapp: DappWithCategoryNames) => dapp.categories.includes(category.id),
+      isSelected: false,
+    }))
+  )
```
jeanregisser, line 81: nit: this way it only does the mapping once when it needs the initial value
```suggestion
  const [filterChips, setFilterChips] = useState<FilterChip<DappWithCategoryNames>[]>(
    () => 
      categories.map((category) => ({
        id: category.id,
        name: category.name,
        filterFn: (dapp: DappWithCategoryNames) => dapp.categories.includes(category.id),
        isSelected: false,
      }))
  )
```

jeanregisser (approved):
Awesome! 🚀 

> (and the "favourites" section is a whole other component)

oh this is surprising, probably ok with given the small number of dapps we have, but it should have been part of the section list normal rendering.

src/dapps/saga.ts
```diff
@@ -117,7 +118,10 @@ export function* watchDappSelected() {
 }
 
 export function* watchFetchDappsList() {
-  yield* takeLeading([fetchDappsList.type, Actions.SET_ACCOUNT], safely(handleFetchDappsList))
+  yield* takeLeading(
+    [fetchDappsList.type, Actions.SET_ACCOUNT, setLanguage.type],
```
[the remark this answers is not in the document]
kathaypacific: true, i can take this out to a separate PR

src/dappsExplorer/DAppsExplorerScreenSearchFilter.tsx
kathaypacific: oops, thanks for the catch!

PR #4834, opened 2024-01-31
feat(onboarding): remove steps on phone verification screen when importing

satish-ravi:
### Description

Removes steps on phone verification screen when importing. Refactored the screen to accept a property `isOnboarding` instead of `hideOnboardingStep` which was used to identify whether the user is in the onboarding flow and whether to show steps or not.
Also fixed a bug where iOS gestures where disabled when not in onboarding instead of onboarding

### Test plan

Unit tests, manual

| Phone Input (onboarding) | Verification Code input (onboarding) | Phone Input (Settings) 
|--------|--------|--------|
|  |  |  |

### Related issues

- Part of ACT-863

### Backwards compatibility

Yes


src/verify/VerificationStartScreen.tsx
```diff
@@ -105,35 +111,33 @@ function VerificationStartScreen({
...
             testID="PhoneVerificationSkipHeader"
             onPress={onPressSkip}
             titleStyle={{ color: colors.onboardingBrownLight }}
           />
         ),
-      headerLeft: () => route.params?.hideOnboardingStep && <BackButton />,
+      headerLeft: () => !route.params?.isOnboarding && <BackButton />,
       // Disable iOS back during onboarding
-      gestureEnabled: route.params?.hideOnboardingStep ? false : true,
+      gestureEnabled: !route.params?.isOnboarding,
```
satish-ravi, line 134: This condition is flipped now. I believe this was buggy originally as we wanted to disable back gesture during onboarding but this was actually disabling them when not in onboarding

src/onboarding/steps.ts
```diff
@@ -23,12 +23,10 @@ import { StatsigExperiments } from 'src/statsig/types'
 export const END_OF_ONBOARDING_SCREENS = [Screens.WalletHome, Screens.ChooseYourAdventure]
 
 interface NavigatorFunctions {
-  navigate: typeof NavigationService.navigate | ((screen: Screens) => void)
```
satish-ravi, line 26: this dummy type (for calculating total steps) wasn't allowing setting screen, updated the fn below to match navigate's type

MuckT (approved):
LGTM. Thanks for cleaning up `hideOnboardingStep` logic! `isOnboarding` is a lot clearer.

PR #4842, opened 2024-02-03
chore(multichain): static typing for multichain

cajubelt:
### Description

we should be able to add chains by just updating enums and cleaning up the compilation errors-- no steps that result in silent errors that require thorough app testing to uncover.

this gets us closer to that

### Test plan

- compiles
- spot checked celoscan and etherscan translations on a dev build, they look fine

### Related issues

na

### Backwards compatibility

na


src/nfts/NftsInfoCarousel.tsx
```diff
@@ -100,17 +101,6 @@ function NftImageCarousel({ nfts, handleOnPress, activeNft }: NftImageCarouselPr
   )
 }
 
-const EXPLORER_LINK_TRANSLATION_STRINGS: Record<NetworkId, string> = {
```
cajubelt, line 103: just moved this

src/web3/networkConfig.ts
```diff
@@ -72,7 +72,7 @@ interface NetworkConfig {
   }
   celoTokenAddress: Address
   celoGasPriceMinimumAddress: Address
-  alchemyRpcUrl: Partial<Record<Network, string>>
+  alchemyRpcUrl: Record<Exclude<Network, Network.Celo>, string>
```
cajubelt, line 75: tested this by deleting OP from alchemyRpcUrl and verifying that it now gives a compilation error where it didn't before

silasbw (approved):

src/transactions/feed/TransactionDetailsScreen.tsx
```diff
@@ -189,8 +187,7 @@ function TransactionDetailsScreen({ navigation, route }: Props) {
             >
               <View style={styles.rowContainer}>
                 <Text style={styles.blockExplorerLink}>
-                  {transactionNetwork === Network.Celo && t('viewOnCeloScan')}
-                  {transactionNetwork === Network.Ethereum && t('viewOnEthereumBlockExplorer')}
+                  {t(EXPLORER_LINK_TRANSLATION_STRINGS[transaction.networkId])}
```
satish-ravi, line 190: This changes the text from celo scan to celo explorer. Is this intentional?

cajubelt: thanks, it was not. fortunately a test covered that as well

src/transactions/feed/TransactionDetailsScreen.tsx
```diff
@@ -189,8 +187,7 @@ function TransactionDetailsScreen({ navigation, route }: Props) {
             >
               <View style={styles.rowContainer}>
                 <Text style={styles.blockExplorerLink}>
-                  {transactionNetwork === Network.Celo && t('viewOnCeloScan')}
-                  {transactionNetwork === Network.Ethereum && t('viewOnEthereumBlockExplorer')}
+                  {t(TX_EXPLORER_LINK_TRANSLATION_STRINGS[transaction.networkId])}
```
jeanregisser, line 190: The problem with this pattern is it will prevent static extraction of translatable strings.

I admit we haven't been using it yet, but would be nice not to break this now.
Or have dicuss it explicitely if we do not want to pursue this goal.

cajubelt: thanks, addressed this in https://github.com/valora-inc/wallet/pull/4887

PR #4845, opened 2024-02-05
test(e2e): test full send flow on deeplinks send links

MuckT:
### Description

Adds a random comment and completes a send to deeplink to better test this flow end to end. Previously we simply verified that the review screen displayed; however, this allowed issues such as ACT-1089 to make it to production environments. 

### Test plan

- [ ] Tested locally on iOS
- [ ] Should be merged after the fix for ACT-1089,

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/e2e/src/usecases/HandleDeepLinkSend.js
+++ b/e2e/src/usecases/HandleDeepLinkSend.js
@@ -1,11 +1,8 @@
+import { E2E_TEST_FAUCET } from '../../scripts/consts'
 import { launchApp, reloadReactNative } from '../utils/retries'
 import { inputNumberKeypad, quote } from '../utils/utils'
 
 const deepLinks = {
-  withAll:
-    'celo://wallet/pay?address=0xC0509a7dcc69a0B28c7Ca73feD2FF06b9d59E5b9&amount=0.1&currencyCode=USD&token=cUSD&displayName=TestFaucet&comment=sending+usd:+0.1+to+my+wallet',
-  withoutAmount:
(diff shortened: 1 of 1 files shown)


e2e/src/usecases/HandleDeepLinkSend.js
```diff
@@ -23,25 +20,53 @@ const openDeepLink = async (payUrl) => {
 
 export default HandleDeepLinkSend = () => {
   describe('When Launching Deeplink - App Closed', () => {
+    let commentText
     it('Then should handle deeplink with all attributes', async () => {
-      const PAY_URL = quote(deepLinks.withAll)
+      commentText = `${new Date().getTime()}-${parseInt(Math.random() * 100000)}`
```
cajubelt: @MuckT can you use a fixed comment instead? we often just rerun e2e tests when they fail instead of looking closely at what went wrong. so if this failed only some of the time, we could entirely miss it and just think "what a flaky test". in general, I am not a fan of tests that rely on randomness for this reason (also because it can be hard to repro the failure)

satish-ravi:
@MuckT the tests were actually running against the old send flow. Only one of the deep link tests were failing, which I think was because fees were not available on the SendConfirmation screen. I pushed the changes to use the new send flow and now 3 tests fail as expected (all the ones named `Then should handle deeplink with all attributes`). Also confirmed locally that after merging with #4851 all tests pass.

satish-ravi (approved):

PR #4847, opened 2024-02-05
feat(recovery): Add error screen for restore

finnian0826:
### Related issues

- Fixes #[issue number here]


jh2oman (approved):

src/keylessBackup/KeylessBackupProgress.test.tsx
```diff
@@ -141,5 +141,39 @@ describe('KeylessBackupProgress', () => {
...
+      expect(getByTestId('KeylessBackupProgress/RestoreCreateNewWallet')).toBeTruthy()
+      fireEvent.press(getByTestId('KeylessBackupProgress/RestoreCreateNewWallet'))
+
+      expect(navigate).toHaveBeenCalledTimes(1)
+      expect(navigate).toHaveBeenCalledWith(Screens.Welcome)
+      expect(ValoraAnalytics.track).toHaveBeenCalledTimes(1)
+      expect(ValoraAnalytics.track).toHaveBeenCalledWith(
+        KeylessBackupEvents.cab_restore_failed_create_new_wallet
+      )
+    })
```
jophish, line 177: nit: Add a test for when "Help" is pressed/ensure that the appropriate analytics event is fired?

jophish (approved):
Left a comment about a possible test to add, but overall LGTM

src/keylessBackup/KeylessBackupProgress.test.tsx
finnian0826: I tried that but for some reason it seems like the help button wasn't visible in the render? Should the header be visible like that in the test? I looked at other places a help button was done in the header and none of those places have unit tests for it so wasn't sure how to do it.

finnian0826: Screenshot 2024-02-07 at 9 19 28 AM
@jophish here's the error I get, the test is set up the same as the others, just checking for the "Help" button test ID (`KeylessBackupRestoreHelp`)

jophish: Ah yeah, I've also had difficulty testing header behavior that relies on components added from within a `useEffect` hook. You may have luck with a `waitFor` (see [here](https://stackoverflow.com/questions/65635054/how-to-test-async-useeffect)) to ensure the hook has run before attempting to press the Help button. If it's too much trouble, feel free to merge though; I don't think it's critical.

finnian0826: hmm still seems to be failing even with the `waitFor` added in

PR #4848, opened 2024-02-06
feat: keyless backup intro

MuckT:
### Description

Sets up a `KeylessBackupIntro.tsx` screen which handles both setup and restore.

| iOS Setup | iOS |
| ---- | ----- | 
|  |  | 

### Test plan

- Tested locally on iOS
- Unit tests added 

### Related issues

- Fixes ACT-838

### Backwards compatibility

Yes


src/keylessBackup/KeylessBackupIntro.tsx
```diff
@@ -1,69 +1,97 @@
...
-        text={t('continue')}
+        testID="keylessBackupIntro/Continue"
+        onPress={() => {
+          if (isSetup) {
+            ValoraAnalytics.track(KeylessBackupEvents.cab_setup_continue)
+            navigate(Screens.SignInWithEmail, { keylessBackupFlow: KeylessBackupFlow.Setup })
+          } else {
+            ValoraAnalytics.track(KeylessBackupEvents.cab_restore_continue)
+            navigate(Screens.SignInWithEmail, { keylessBackupFlow: KeylessBackupFlow.Restore })
+          }
```
satish-ravi, line 92: nit: we could use the same event for both and avoid the if else here
```suggestion
          ValoraAnalytics.track(KeylessBackupEvents.cab_intro_continue, { keylessBackupFlow })
          navigate(Screens.SignInWithEmail, { keylessBackupFlow })
```

src/keylessBackup/KeylessBackupIntro.tsx
```diff
@@ -1,69 +1,97 @@
...
             <View style={styles.authFactorLine}>
               <SmartphoneIcon style={styles.smartphoneIcon} />
-              <Text style={styles.authFactorText}>{t('setUpKeylessBackup.phoneNumber')}</Text>
+              <Text style={styles.authFactorText}>{t('keylessBackupIntro.phoneNumber')}</Text>
             </View>
           </View>
 
           <Text style={styles.reminderText}>
-            <Trans i18nKey="setupKeylessBackupReminder">
-              {/* namespaced keys did not work for this */}
```
satish-ravi, line 50: interesting that this works now

src/keylessBackup/WalletSecurityPrimer.tsx
```diff
@@ -34,7 +38,14 @@ function WalletSecurityPrimer({ route }: Props) {
         testID="WalletSecurityPrimer/GetStarted"
         onPress={function () {
           ValoraAnalytics.track(KeylessBackupEvents.wallet_security_primer_get_started)
-          navigate(Screens.SetUpKeylessBackup)
+          dispatch(
+            keylessBackupStarted({
+              keylessBackupFlow: KeylessBackupFlow.Setup,
+            })
+          )
```
satish-ravi, line 45: I believe this was moved to SignInWithEmail as part of #4811

src/keylessBackup/KeylessBackupIntro.tsx
```diff
@@ -1,69 +1,97 @@
...
+type Props = NativeStackScreenProps<StackParamList, Screens.KeylessBackupIntro>
+
+function KeylessBackupIntro({ route }: Props) {
+  const { keylessBackupFlow } = route.params
+  const isSetup = keylessBackupFlow === KeylessBackupFlow.Setup
   const { t } = useTranslation()
+
   return (
     <SafeAreaView style={styles.container}>
+      <CustomHeader left={isSetup ? <BackButton /> : <CancelButton />} style={styles.header} />
```
satish-ravi, line 36: I believe [other cab screens](https://github.com/valora-inc/wallet/blob/main/src/keylessBackup/KeylessBackupPhoneInput.tsx#L129) use navigationOptions that renders based on props. Is there a rule on when we should prefer CustomHeader over the navigationOptions?
Also, where does the CancelButton navigate to? We also have a KeylessBackupCancelButton that's shared among all subsequent screens which navigate appropriately based on the flow.

src/keylessBackup/KeylessBackupIntro.tsx
MuckT: I think what made the difference was using curlies e.g. `i18nKey={'keylessBackupIntro.setup.reminder'}`.

src/keylessBackup/KeylessBackupIntro.tsx
MuckT: The reason I used a CustomHeader here is because i needed to use either cancel or back depending if in setup or restore flow, otherwise I think header options are preferred.

src/keylessBackup/WalletSecurityPrimer.tsx
MuckT: Fixed in 2573f3e705e03eb70ad51204714fda2b1ba589d0!

src/keylessBackup/KeylessBackupIntro.tsx
MuckT: Changed to use one analytic event in ee39204f0a845d6f9420bfc5e935c901ce76bab1!

src/keylessBackup/KeylessBackupIntro.tsx
satish-ravi: The other screens also need access to whether it's setup / restore, it's accessed here https://github.com/valora-inc/wallet/blob/62ceef4e0205afb365d41c6256e58572261dc0fd/src/keylessBackup/KeylessBackupPhoneInput.tsx#L133. Also not sure if its related, the alignment of the header was off on android in @jh2oman 's demo today. Maybe related to using CustomHeader?

jh2oman (approved):

src/keylessBackup/KeylessBackupIntro.tsx
MuckT: I was able to switch the navigation to use `SafeAreaView` from `react-native` and the above comment to resolve this issue. There is the bigger question of if we should be using `SafeAreaView` from `react-native-safe-area-context` instead, but for consistency on these screens I think that should be handled all at once rather then screen by screen.

PR #4849, opened 2024-02-06
chore(nft): update `nftsWithMetadataSelector` typings

bakoushin:
### Description

Ensure that `nftsWithMetadataSelector` return type has non-null `metadata` property.

### Test plan

CI

### Related issues

- Related to RET-1000

### Backwards compatibility

Y


kathaypacific (approved):
🚀

PR #4850, opened 2024-02-06
chore: visual tweaks for `BottomSheetInLineNotification`

bakoushin:
### Description

I want to reuse `BottomSheetInLineNotification` to implement [this design](https://www.figma.com/file/rXBDplfMHHqYmuu6EkgMEo/Gamification-experiments?type=design&node-id=96-3560&mode=design&t=WxCtC0CqGJMX4xV5-0).

To achieve that, I've added two optional properties:
* `showIcon: boolean` (default: `true`)
* `position: 'top' | 'bottom'` (default: `'bottom'`)

### Test plan

* Added unit tests

### Related issues

- Related to RET-1000

### Backwards compatibility

Y

### DIFF
--- a/src/components/BottomSheetInLineNotification.test.tsx
+++ b/src/components/BottomSheetInLineNotification.test.tsx
@@ -1,0 +1,54 @@
+import { render } from '@testing-library/react-native'
+import React from 'react'
+import { StyleSheet } from 'react-native'
+import { Severity } from 'src/components/InLineNotification'
+import BottomSheetInLineNotification from './BottomSheetInLineNotification'
+
+describe('BottomSheetInLineNotification', () => {
+  it('renders correctly when showNotification is true', () => {
+    const { getByText } = render(
+      <BottomSheetInLineNotification
+        severity={Severity.Warning}
+        description="Test description"
+        showNotification={true}
+      />
+    )
+    expect(getByText('Test description')).toBeTruthy()
+  })
+
+  it('does not render when showNotification is false', () => {
+    const { queryByText } = render(
+      <BottomSheetInLineNotification
+        severity={Severity.Warning}
+        description="Test description"
+        showNotification={false}
+      />
+    )
+    expect(queryByText('Test description')).toBeNull()
+  })
+
+  it.each(['top', 'bottom'])('renders with a %s position when it is defined', (position: any) => {
+    const { getByTestId } = render(
+      <BottomSheetInLineNotification
+        severity={Severity.Warning}
+        description="Test"
+        showNotification={true}
+        position={position}
+      />
+    )
+    const style = StyleSheet.flatten(getByTestId('notificationContainer').props.style)
+    expect(style).toHaveProperty(position)
+  })
+
+  it('renders with a bottom position when position property is absent', () => {
+    const { getByTestId } = render(
+      <BottomSheetInLineNotification
+        severity={Severity.Warning}
+        description="Test"
+        showNotification={true}
+      />
+    )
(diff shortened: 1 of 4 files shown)


kathaypacific (commented):
I think this implementation is very elegant! so i don't have a problem with the code, but i wonder if you can get what you want with BottomSheetInLineNotification - the design has a white toast + no icon, but the inline notifications have a coloured background and an icon. i see you added a prop to hide the icon, but are we going to add a new color scheme for white? what does the white flavour mean?

i wonder if it makes more sense to try to use the ToastWithCTA component instead? we could rename it to just Toast and have an optional CTA, and change it to white (or have an `inverse` property. maybe you've thought about this already?

kathaypacific:
> I didn’t consider `ToastWIthCTA` initially, because it is far away from the designs:
> 
> * it looks far different in terms of fonts, spacing, and border-radius (but `BottomSheetInLineNotification` looks exactly as per design)
> * it doesn’t have a black backdrop seen in the design (but `BottomSheetInLineNotification` has)
> * it has a CTA to deal with _in addition_ to different screen positions to support
> 
> Adopting `BottomSheetInlineNotification` appeared to me to be just quicker. As for the color scheme, it seems that the "Informational" flavor [has already been switched](https://github.com/valora-inc/wallet/blob/7587c420537df4d72ced2057f14607d6850f94e7/src/components/InLineNotification.tsx#L123) to `gray1`. Our designs are evolving faster than we can get used to them 😅
> 
> Today I've spent some time looking for ways to extend `ToastWithCTA` and I’m still dubious that the design in question fits well into the idea of a "toast". And it creates some uncomfortable questions to deal with:
> 
> * can we consistently reuse a white toast without CTA in any other future cases, besides the NFT celebration?
> * if so: white toast without borders or shadows easily melts into a white overall wallet background, should we add a backdrop (as in design)?
> * if so: should we add a backdrop for a black variant of a toast as well? (it wasn't there before)
> * should _a toast_ have a backdrop at all?
> * since it may not have a CTA, should it be dismissible by user swipe?
> * should it be dismissible by a tap on the backdrop?
> 
> Looking at the problem more broadly, probably a good enough solution could be to abstain from touching any existing components if they are not explicitly considered to be a part of a design system and aren't thoroughly designed to be one. Probably, as with fonts and colors, we should ask the design team if any component in the designs is present in some reusable common components library, they do maintain and reuse consistently. If it isn't, just implement everything entirely inside the functionality in question (~copy-pasting~ borrowing rather than composing).
> 
> It could be perfectly aligned with the exploratory stage of the designs we currently are in. And, probably, that would preserve some time otherwise spent on refactoring and alongside discussions.
> 
> If any time in the future we notice that some consistency in the designs starts to emerge, that would be a sign to abstract some common components. Also, that may coincide with some kind of "major redesign", where we would have to refactor everything one way or the other.

oh boy, i didn't realise the inline notification colors had already changed!!

yes i agree with everything you said, my suggestion to look at ToastWithCTA was probably not the best one. perhaps applying [aha programing](https://kentcdodds.com/blog/aha-programming#aha-) is the path of least resistance here. i definitely agree that we should adopt a general rule of thumb in eng to not build anything into the design system unless we are sure. sometimes i do see us doing this based on (reasonable) assumptions but when we are wrong, we've just done extra work and created extra complexity. 

i think my personal preference here is either to keep this PR and require the design to look like one of the existing flavours of the inline notification, or build a custom component that maybe becomes adopted for other things and eventually replaces the ToastWithCTA. what do you think?

bakoushin:
Closing for now. It was decided to refrain from modifying an existing component to implement a celebration experience.

PR #4851, opened 2024-02-06
fix(payments): create and pass a preparedTx to confirmation screen

satish-ravi:
### Description

Bidali and deeplink payments were broken with the new send flow because a prepared transaction object wasn't passed to the SendConfirmation screen. This prepares a transaction and passes it on to the SendConfirmation screen. If a transaction is not possible, it still navigates to the screen, but the Send button would be disabled (In the old send flow, if a tx is not possible, Send was still clickable, but the tx will fail)

### Test plan

Unit tests, manual. Also E2E tests currently failing in #4845 passes with this fix

Payment deeplink:

https://github.com/valora-inc/wallet/assets/5062591/4774c91c-557a-40f3-916c-4f8c33c87952

Bidali payment:

https://github.com/valora-inc/wallet/assets/5062591/aaeb8b80-471c-4e12-a83c-5aa17710d284

Exceeds balance:

https://github.com/valora-inc/wallet/assets/5062591/4373dae8-5672-4bb8-830d-0b0c22e77fd1

### Related issues

- Fixes ACT-1089

### Backwards compatibility

Yes

### DIFF
--- a/src/fiatExchanges/saga.test.ts
+++ b/src/fiatExchanges/saga.test.ts
@@ -1,5 +1,6 @@
 import BigNumber from 'bignumber.js'
 import { expectSaga } from 'redux-saga-test-plan'
+import * as matchers from 'redux-saga-test-plan/matchers'
 import { call, select } from 'redux-saga/effects'
 import { SendOrigin } from 'src/analytics/types'
 import { activeScreenChanged } from 'src/app/actions'
@@ -15,6 +16,7 @@ import { navigate } from 'src/navigator/NavigationService'
 import { Screens } from 'src/navigator/Screens'
 import { AddressRecipient, RecipientType } from 'src/recipients/recipient'
 import { sendPayment, sendPaymentFailure, sendPaymentSuccess } from 'src/send/actions'
+import { preparePaymentRequestTransaction } from 'src/send/utils'
 import { tokensByCurrencySelector } from 'src/tokens/selectors'
 import { updateTransactions } from 'src/transactions/actions'
 import {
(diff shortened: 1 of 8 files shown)


MuckT (approved):
👏 LGTM! Tested end to end by purchasing a Bidali gift card without issue.

jophish (commented):
This setup sort of makes an assumption about how we want the parameters of `SendConfirmationScreen` to work going forwards -- right now, the (proposed) plan for the `SendConfirmationScreen` once we remove the old send flow is to make `preparedTransaction` a required field, since we should always have one available. In the case of payment deeplinks/bidali, we might _not_ however.

I think what we do here hinges on what kind of behavior we want when a TX isn't possible for a deeplinked/bidali request. This implementation chooses to go with just disabling the send button on the confirmation screen if a TX isn't possible, though there are other solutions that could involve not even navigating to the send confirmation screen at all.

FWIW I think this solution could be fine (it's no worse than what we currently do, which is allow the user to press the "Send" button, but then fail during the send). There is a comment on the `SendConfirmationScreen` [here](https://github.com/valora-inc/wallet/blob/main/src/send/SendConfirmation.tsx#L138) which discusses the plan post-cleanup and says that we want to make `preparedTx` required. If we go with this solution, we should prolly update that comment in this PR to express the new plan, which is to keep `preparedTx` optional, and just hard disable the send button if it isn't present.

satish-ravi:
Thanks @jophish , updated the comment

jophish (approved):

PR #4852, opened 2024-02-06
refactor: use FlatList for token bottom sheet

kathaypacific:
### Description

This change should improve performance of the list in general, especially when using with filters.

This PR also has a change to make the list filtering slightly more performant.

### Test plan

The filters should respond faster and there should be no other visual changes (except for changing some font from deprecated styles to new styles)

https://github.com/valora-inc/wallet/assets/20150449/2b7476c1-f8ba-4ae2-9078-52cc1747a62a

I had a go at profiling the change in the flipper profiler and the time for first render after filter press using FlatList is around under 100ms, whereas with the ScrollView implementation it was ~300ms.

### Related issues

- Related to RET-984

### Backwards compatibility

Y


src/components/TokenBottomSheet.tsx
```diff
@@ -218,68 +229,67 @@ function TokenBottomSheet<T extends TokenBalance>({
...
+        renderItem={({ item, index }) => (
+          <TokenOptionComponent
+            tokenInfo={item}
+            onPress={onTokenPressed(item)}
+            index={index}
+            showPriceUsdUnavailableWarning={showPriceUsdUnavailableWarning}
+          />
+        )}
+        initialNumToRender={8}
+        maxToRenderPerBatch={8}
```
jeanregisser, line 284: Looks like this should be based on the screen size and item height to avoid having some initial blanks, on bigger screens.

jeanregisser (approved):
Great!! 🚀

src/components/TokenBottomSheet.tsx
jeanregisser: Sounds good, I didn't mean we should really measure, but we could have used a hardcoded item height as a quick shortcut too.

PR #4857, opened 2024-02-07
chore(networkConfig): remove 1 place where manual testing and code searching would be needed to add chains

cajubelt:
### Description

we should try to maintain a status quo where all we have to do in every repo to add a chain is update a "NetworkId" enum and fix all compilation errors that result from it

### Test plan

added spot checks for the two affected configs

### Related issues

cc https://linear.app/valora/issue/ACT-1060/add-optimism-arbitrum-to-wallet

### Backwards compatibility

na


src/web3/networkConfig.ts
```diff
@@ -482,26 +483,13 @@ export const networkIdToWalletConnectChainId: Record<NetworkId, string> = {
...
-export const walletConnectChainIdToNetwork: Record<string, Network> = {
-  'eip155:44787': Network.Celo,
-  'eip155:42220': Network.Celo,
-  'eip155:1': Network.Ethereum,
-  'eip155:11155111': Network.Ethereum,
-  'eip155:42161': Network.Arbitrum,
-  'eip155:421614': Network.Arbitrum,
-  'eip155:10': Network.Optimism,
-  'eip155:11155420': Network.Optimism,
+export const walletConnectChainIdToNetwork: Record<string, Network> = {}
```
jophish, line 490: I spent a sec trying to code golf this into a single line using a `Object.entries(walletConnectChainIdToNetworkId).reduce(...)`, but then realized that it's prolly less readable than the current approach :laughing:

jophish (approved):

src/web3/networkConfig.ts
cajubelt: I think you could do `Object.fromEntries(Object.entries(walletConnectChainIdToNetworkId).map(([chainId, networkId]) => [networkId, chainId])))` . But i went with the for loop because it seemed more readable

cajubelt: i may have gotten network id and chain id backwards but whatever

kathaypacific (approved):
TY for the catch, looks good. recently Jean was floating the idea of moving away from lodash, we didn't end up discussing at eng sync like we intended but seeing it used here made me remember [this](https://github.com/valora-inc/wallet/pull/4374#discussion_r1371322127).

PR #4862, opened 2024-02-07
refactor(send): clean some todos

satish-ravi:
### Description

Cleaning some todos linked to closed / cancelled tickets.

### Test plan

CI

### Related issues

- Part of ACT-1075

### Backwards compatibility

Yes


src/fees/selectors.ts
```diff
@@ -15,7 +15,7 @@ export function getFeeEstimateDollars(
   tokenInfo: TokenBalance | undefined
 ) {
   return (state: RootState) => {
-    // TODO(ACT-922): Handle cases where address is null ex: Ethereum
+    // TODO(ACT-1075): Clean up
```
satish-ravi, line 18: only used in send flow, will clean in a followup PR

src/send/SendEnterAmount.test.tsx
```diff
@@ -315,7 +315,6 @@ describe('SendEnterAmount', () => {
       usedSearchTerm: false,
       selectedFilters: [],
     })
-    // TODO(ACT-955): assert fees
```
satish-ravi, line 318: there are separate tests asserting fees

src/send/SendEnterAmount.tsx
```diff
@@ -216,10 +216,6 @@ function SendEnterAmount({ route }: Props) {
     if (parsedAmount.isLessThanOrEqualTo(0) || parsedAmount.isGreaterThan(token.balance)) {
       return
     }
-    // TODO (ACT-973?): Address is not always set on the recipient, even for PN recipients.
```
satish-ravi, line 219: this is no longer true, we have static typing to ensure address is always present at this point

src/viem/saga.test.ts
```diff
@@ -778,8 +778,6 @@ describe('getSendTxFeeDetails', () => {
       .run()
   })
 
-  // TODO(ACT-925): remove this test once we've ensured gas and gasPrice are
-  // consistently BigNumbers or strings
```
satish-ravi, line 782: linked to cancelled issue, will be cleaned up in a future issue (see below)

e2e/src/utils/utils.js
```diff
@@ -427,3 +427,7 @@ export async function fundWallet(senderPrivateKey, recipientAddress, stableToken
     .sendAndWaitForReceipt({ from: senderAddress })
   console.log('Funding TX receipt', receipt)
 }
+
+export const createCommentText = () => {
+  return `${new Date().getTime()}-${parseInt(Math.random() * 100_000)}`
```
jh2oman, line 432: Is there any benefit to this being a dynamic string? Seems unlikely, but it could make for flakier tests.

jh2oman: I'd consider making it just a static comment.

jh2oman (approved):

e2e/src/utils/utils.js
satish-ravi: static comment would mean parallel e2e runs could step on each other since they all use the same wallet. I don't think the randomness here should case any test flake. More context here: https://github.com/valora-inc/wallet/pull/4845#discussion_r1478848911

jh2oman: ah got it, thanks!

PR #4864, opened 2024-02-07
fix(wc): handle errors while preparing transactions

jeanregisser:
### Description

This PR handles errors happening while preparing transactions coming from WC.
- `chainId` is now stripped (we don't need it and it was causing an RPC error on Celo)
- `prepareTransactions` is now wrapped in a try/catch and we now surface the error in the `ActionRequest` bottom sheet

### Test plan

- Updated tests
- Manually tested by transacting with the following dapps: Staked Celo, Moola, ChurritoFi
- New error message if transaction preparation fails, with the underlying error message, to help users/builders understand the problem

Note: the underlying error message won't be translated, but I think it's better than hiding it.

### Related issues

- Fixes RET-1013

### Backwards compatibility

Yes

### DIFF
--- a/src/walletConnect/saga.ts
+++ b/src/walletConnect/saga.ts
@@ -86,7 +86,7 @@ import {
   takeEvery,
   takeLeading,
 } from 'typed-redux-saga'
-import { Address, GetTransactionCountParameters, hexToBigInt, isHex } from 'viem'
+import { Address, BaseError, GetTransactionCountParameters, hexToBigInt, isHex } from 'viem'
 import { getTransactionCount } from 'viem/actions'
 
 let client: IWeb3Wallet | null = null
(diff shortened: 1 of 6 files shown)


src/walletConnect/screens/ActionRequest.test.tsx
```diff
@@ -242,12 +268,7 @@ describe('ActionRequest with WalletConnect V2', () => {
         )
       ).toBeTruthy()
       expect(getByText('dappsDisclaimerUnlistedDapp')).toBeTruthy()
-      expect(
-        getByText('walletConnectRequest.estimatedNetworkFee, {"networkName":"Celo Alfajores"}')
-      ).toBeTruthy()
-      const fee = within(getByTestId('EstimatedNetworkFee'))
-      expect(fee.getByText('0.0012 CELO')).toBeTruthy()
-      expect(fee.getByText('₱0.008')).toBeTruthy()
+      expect(queryByTestId('EstimatedNetworkFee')).toBeFalsy()
```
jeanregisser, line 271: This was accidentally taking the viem enabled path.
I made sure to reset the feature flag mock before each test to avoid this and moved this check to where it belonged.
This test case was testing `personal_sign` which shouldn't have a TX fee anyway.

src/walletConnect/saga.ts
```diff
@@ -410,6 +410,14 @@ export function* normalizeTransaction(rawTx: any, network: Network) {
     tx.nonce = yield* call(getTransactionCount, publicClient[network], txCountParams)
   }
 
+  if ('chainId' in tx) {
```
kathaypacific, line 413: should we do this only for Celo, since the comment below says that this is only a problem there? i also wondered if this can become a whack-a-mole situation where there are more properties on the tx that could cause problems (should we keep an inclusion list of properties we actually want and delete all other properties?)

kathaypacific (approved):
LGTM, i like the idea of surfacing the error on the bottom sheet! that will be helpful for debugging any user reports too

src/walletConnect/saga.ts
jeanregisser: Stripping it is ok on all chains.

Yes I've thought about this whack-a-mole situation, I was looking at the internals of viem as my initial thinking was that it should probably be handled there instead. I'll chat with the authors to see what they think.
Otherwise we can do an inclusion list ourselves, but I'd prefer if we could avoid it.

PR #4863, opened 2024-02-07
refactor(send): remove old send flow

satish-ravi:
### Description

Changes include:
- Removing use_new_send_flow gate and all false code associated with it
- Removing all of the old send flow code and its dependencies (~1k lines flagged by knip)
- Updating secure send code to not take transactionData as parameter, which was only set in the old send flow
- Moving some helper functions in the old send flow that were also required in the new send flow to a more common location (E.g., resolveId, shared types, etc.)

Requires #4867, #4845 for e2e tests to pass

### Test plan

CI

### Related issues

- Fixes ACT-1075

### Backwards compatibility

Yes

PR #4867, opened 2024-02-07
test(e2e): update tests to run against new send flow

satish-ravi:
### Description

Updating tests that touch the old send flow to use the new send flow to unblock #4863. This updates all failing tests [here](https://github.com/valora-inc/wallet/pull/4863/checks?check_run_id=21334412937) to use the new send flow (except DeepLink tests which is updated in #4845. 
Also cleaned some redundant tests in OffRamps.js

### Test plan

CI

### Related issues

- Part of ACT-1075

### Backwards compatibility

Yes


e2e/src/usecases/OffRamps.js
```diff
@@ -60,28 +63,6 @@ export default offRamps = () => {
       await element(by.id('cashOut')).tap()
     })
 
-    it.each`
```
satish-ravi, line 63: this test is same as the test below except for using different amounts and asserting for SendBar.

cajubelt (approved):

PR #4868, opened 2024-02-07
feat(CAB): Add help bottom sheet for CAB phone verification screen

finnianjs:
### Description

Create bottom sheet for when user hits the `Help` button on the phone code input screen during the CAB flow. Analytics events added for hitting the help button and both of the buttons on the bottom sheet.

### Test plan

Setup flow:

https://github.com/valora-inc/wallet/assets/140328381/be46d8f7-ecfc-4f42-bef5-a032eb79c865

Restore flow:

https://github.com/valora-inc/wallet/assets/140328381/090a306f-b722-4d91-a4a6-f3d8971f6646

### Related issues

- Fixes #[ACT-815]

### Backwards compatibility

Yes

### DIFF
--- a/src/keylessBackup/KeylessBackupPhoneCodeInput.test.tsx
+++ b/src/keylessBackup/KeylessBackupPhoneCodeInput.test.tsx
@@ -138,7 +138,8 @@ describe('KeylessBackupPhoneCodeInput', () => {
     fireEvent.press(getByTestId('KeylessBackupPhoneCodeInputHelp'))
     expect(getByTestId('KeylessBackupPhoneCodeInput/HelpInfoBottomSheet')).toBeTruthy()
     expect(ValoraAnalytics.track).toHaveBeenCalledWith(
-      KeylessBackupEvents.cab_phone_verification_help
+      KeylessBackupEvents.cab_phone_verification_help,
+      { keylessBackupFlow: KeylessBackupFlow.Setup }
     )
   })
   it('tracks analytics event when pressing go back button', async () => {
@@ -149,7 +150,8 @@ describe('KeylessBackupPhoneCodeInput', () => {
     expect(getByTestId('KeylessBackupPhoneCodeInput/HelpInfoBottomSheet/GoBack')).toBeTruthy()
     fireEvent.press(getByTestId('KeylessBackupPhoneCodeInput/HelpInfoBottomSheet/GoBack'))
     expect(ValoraAnalytics.track).toHaveBeenCalledWith(
-      KeylessBackupEvents.cab_phone_verification_help_go_back
+      KeylessBackupEvents.cab_phone_verification_help_go_back,
+      { keylessBackupFlow: KeylessBackupFlow.Setup }
     )
   })
   it('goes to WalletHome screen and track analytics event when pressing skip button in setup flow', async () => {
@@ -161,7 +163,8 @@ describe('KeylessBackupPhoneCodeInput', () => {
     fireEvent.press(getByTestId('KeylessBackupPhoneCodeInput/HelpInfoBottomSheet/Skip'))
     expect(navigate).toHaveBeenCalledWith(Screens.WalletHome)
     expect(ValoraAnalytics.track).toHaveBeenCalledWith(
-      KeylessBackupEvents.cab_phone_verification_help_skip
+      KeylessBackupEvents.cab_phone_verification_help_skip,
+      { keylessBackupFlow: KeylessBackupFlow.Setup }
     )
   })
   it('goes to ImportSelect screen and track analytics event when pressing skip button in restore flow', async () => {
@@ -173,7 +176,8 @@ describe('KeylessBackupPhoneCodeInput', () => {
     fireEvent.press(getByTestId('KeylessBackupPhoneCodeInput/HelpInfoBottomSheet/Skip'))
     expect(navigate).toHaveBeenCalledWith(Screens.ImportSelect)
     expect(ValoraAnalytics.track).toHaveBeenCalledWith(
-      KeylessBackupEvents.cab_phone_verification_help_skip
+      KeylessBackupEvents.cab_phone_verification_help_skip,
+      { keylessBackupFlow: KeylessBackupFlow.Restore }
     )
   })
 })


src/keylessBackup/KeylessBackupPhoneCodeInput.tsx
```diff
@@ -77,6 +126,33 @@ const styles = StyleSheet.create({
...
+    ...typeScale.bodyMedium,
+    textAlign: 'center',
+  },
+  buttonContainer: {
+    flexDirection: 'row',
+    marginTop: 37,
+    marginBottom: 9,
+    alignItems: 'center',
+    justifyContent: 'center',
+    gap: 96,
```
finnian0826, line 148: Needed this to correctly space the two buttons, not sure if there is a better way so open to suggestions!

src/keylessBackup/KeylessBackupPhoneCodeInput.tsx
```diff
@@ -77,6 +126,33 @@ const styles = StyleSheet.create({
...
+    flexDirection: 'column',
+  },
+  bottomSheetBody: {
+    ...typeScale.bodyMedium,
+    textAlign: 'center',
+  },
+  buttonContainer: {
+    flexDirection: 'row',
+    marginTop: 37,
+    marginBottom: 9,
```
finnian0826, line 145: These two magic numbers (37 and 9) are copied over from styling from an old deprecated component, it didn't seem like there was a constant with these values (saw them used like this in other places) so not sure if this is okay or something else should be used?

src/keylessBackup/KeylessBackupPhoneCodeInput.tsx
```diff
@@ -77,6 +126,33 @@ const styles = StyleSheet.create({
...
+    alignItems: 'center',
+    justifyContent: 'center',
+    gap: 96,
+  },
+  goBackButton: {
+    color: colors.primary,
+  },
+  skipButton: {
+    color: '#9CA4A9',
```
finnian0826, line 154: Color pulled from the figma doc, doesn't seem to be any of the `colors` in the style type files.

src/keylessBackup/KeylessBackupPhoneCodeInput.test.tsx
```diff
@@ -129,4 +131,50 @@ describe('KeylessBackupPhoneCodeInput', () => {
...
+    )
+  })
+  it('dismisses help bottom sheet and track analytics event when pressing go back button', async () => {
+    const { getByTestId, queryByTestId } = renderComponent()
+
+    expect(getByTestId('KeylessBackupPhoneCodeInputHelp')).toBeTruthy()
+    fireEvent.press(getByTestId('KeylessBackupPhoneCodeInputHelp'))
+    expect(getByTestId('KeylessBackupPhoneCodeInput/HelpInfoBottomSheet/GoBack')).toBeTruthy()
+    fireEvent.press(getByTestId('KeylessBackupPhoneCodeInput/HelpInfoBottomSheet/GoBack'))
+    expect(queryByTestId('KeylessBackupPhoneCodeInput/HelpInfoBottomSheet')).toBeFalsy()
```
finnian0826, line 151: This is failing in a unit test, not sure how else to test that the `Go back` button dismisses the bottom sheet. In manual testing it correctly dismisses it.

jophish: I'm pretty sure this is because the bottom sheet component is technically always rendered, just not necessarily visible? e.g., if you add `expect(getByTestId('KeylessBackupPhoneCodeInput/HelpInfoBottomSheet')).toBeTruthy()` to the beginning of this test, it won't fail early -- trying to see if there's a way to detect whether or not the bottom sheet is actually _visible_ or not.

src/keylessBackup/KeylessBackupPhoneCodeInput.tsx
cajubelt: in the figma it says gray4 which is in the wallet, could we just use that? https://www.figma.com/file/Qj4kphntYzlYwex79XTcEj/Wallet-Recovery-1.5?type=design&node-id=4309-36620&mode=design&t=y05CTe2XxambYJtg-0

cajubelt: in general if there is a font color in the designs that isn't in the design system I recommend following up with design about it instead of hardcoding in a new color

finnian0826: I saw that, however gray4 in the wallet is #666666 so doesn't match the figma one

finnian0826: updated to the correct color now

src/analytics/Properties.tsx
```diff
@@ -258,6 +258,9 @@ interface KeylessBackupEventsProperties {
   [KeylessBackupEvents.cab_restore_failed_try_again]: undefined
   [KeylessBackupEvents.cab_restore_failed_create_new_wallet]: undefined
   [KeylessBackupEvents.cab_restore_failed_help]: undefined
+  [KeylessBackupEvents.cab_phone_verification_help]: undefined
+  [KeylessBackupEvents.cab_phone_verification_help_skip]: undefined
+  [KeylessBackupEvents.cab_phone_verification_help_go_back]: undefined
```
jh2oman, line 263: All of these should probably use CommonKeylessBackupProps since they are events during both setup and restore

src/keylessBackup/KeylessBackupPhoneCodeInput.test.tsx
```diff
@@ -129,4 +131,49 @@ describe('KeylessBackupPhoneCodeInput', () => {
...
+    const { getByTestId } = renderComponent()
+
+    expect(getByTestId('KeylessBackupPhoneCodeInputHelp')).toBeTruthy()
+    fireEvent.press(getByTestId('KeylessBackupPhoneCodeInputHelp'))
+    expect(getByTestId('KeylessBackupPhoneCodeInput/HelpInfoBottomSheet')).toBeTruthy()
+    expect(ValoraAnalytics.track).toHaveBeenCalledWith(
+      KeylessBackupEvents.cab_phone_verification_help
+    )
+  })
+  it('dismisses help bottom sheet and track analytics event when pressing go back button', async () => {
```
jh2oman, line 144: technicaly doesn't test that the bottom sheet is dismissed, just that the analytics event is fired

jh2oman (approved):
Approved pending those two small comments being addressed

jophish (approved):

PR #4874, opened 2024-02-08
refactor(keyless-backup): use viem instead ethers for signing

satish-ravi:
### Description

Replace ethers usages with viem in cloud backup. More context here: https://valora-app.slack.com/archives/C02KBT0DAHJ/p1707385000759899

### Test plan

CI, tested manually that restore still works

### Related issues

N/A

### Backwards compatibility

Yes


jeanregisser (approved):
Amazing, thanks! 🚀

PR #4875, opened 2024-02-08
fix(dapps): ensure dapps list language is correct

kathaypacific:
### Description

As the title. The localised dapp descriptions and category names are returned from the dapp list response, so we should refresh them after changing language.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y


jeanregisser (approved):
🌍

PR #4876, opened 2024-02-08
fix(onboarding): revert default of CPV screen to be in onboarding

satish-ravi:
### Description

Fixes a bug introduced in #4834 , where VerificationStartScreen defaulted to being not in onboarding. This causes an issue if a user bails on the screen during onboarding. On restart, the screen is loaded in the already onboarded mode where the back button does nothing and there's no option to skip (see before video below). This PR reverts the default to be in onboarding as it doesn't seem trivial to pass props on the initialRoute.

### Test plan

Unit tests and manual

Before:

https://github.com/valora-inc/wallet/assets/5062591/496876ec-4adf-48b8-8a12-72d9030f3e37

After:

https://github.com/valora-inc/wallet/assets/5062591/8d345432-4fcb-439f-b63a-03806d5959e0

### Related issues

N/A

### Backwards compatibility

Yes

### DIFF
--- a/src/verify/VerificationStartScreen.tsx
+++ b/src/verify/VerificationStartScreen.tsx
@@ -71,7 +71,7 @@ function VerificationStartScreen({
     onboardingProps
   )
   const choseToRestoreAccount = useSelector(choseToRestoreAccountSelector)
-  const showSteps = route.params?.isOnboarding && !choseToRestoreAccount
+  const showSteps = !route.params?.hasOnboarded && !choseToRestoreAccount
 
   const countries = useMemo(() => new Countries(i18n.language), [i18n.language])
   const country = phoneNumberInfo.countryCodeAlpha2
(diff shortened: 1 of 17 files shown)


src/verify/VerificationStartScreen.tsx
```diff
@@ -121,23 +121,23 @@ function VerificationStartScreen({
...
             testID="PhoneVerificationSkipHeader"
             onPress={onPressSkip}
             titleStyle={{ color: colors.onboardingBrownLight }}
           />
         ),
-      headerLeft: () => !route.params?.isOnboarding && <BackButton />,
+      headerLeft: () => route.params?.hasOnboarded && <BackButton />,
       // Disable iOS back during onboarding
-      gestureEnabled: !route.params?.isOnboarding,
+      gestureEnabled: !!route.params?.hasOnboarded,
```
MuckT, line 134: nit: Does this need to be forced to a boolean?

```suggestion
      gestureEnabled: route.params?.hasOnboarded,
```

MuckT (approved):

src/verify/VerificationStartScreen.tsx
satish-ravi: it has to be forced to a boolean. The default value is `true`

PR #4878, opened 2024-02-08
chore(imported-tokens): Refactoring network icon logic to be in the saga instead of the selector

dievazqu:
### Description

As title says.
This is a follow up of some unaddressed change of https://github.com/valora-inc/wallet/pull/4806

### Test plan

Unit tests

### Backwards compatibility

Y


src/tokens/TokenImport.tsx
```diff
@@ -63,7 +63,6 @@ export default function TokenImportScreen(_: Props) {
   const { t } = useTranslation()
   const dispatch = useDispatch()
 
-  // Should I create a new config for this?
   const supportedNetworkIds = getSupportedNetworkIdsForTokenBalances()
```
dievazqu, line 66: This should be good for now. We can enable/disable this feature by ff in any case.

src/tokens/TokenImport.tsx
```diff
@@ -204,8 +203,6 @@ export default function TokenImportScreen(_: Props) {
     const networkId = tokenDetails.networkId
     const tokenAddress = tokenDetails.address.toLowerCase()
 
-    // This should always be present, unless we support a network without tokens,
-    // or if we can't fetch the token list (e.g. network issues?), should we fail instead to be safe?
```
dievazqu, line 208: This should be good for now, we are updating the network icon the fetching balance step anyways.

src/tokens/selectors.ts
```diff
@@ -518,18 +518,10 @@ export const feeCurrenciesWithPositiveBalancesSelector = createSelector(
...
-      return []
-    }
-
-    return tokenList
-      .filter((token) => token?.isManuallyImported)
-      .map((token) => ({
-        ...token,
-        networkIconUrl: networksIcon[token.networkId] ?? token.networkIconUrl,
-      })) as TokenBalance[]
+export const importedTokensSelector = createSelector([tokensListSelector], (tokenList) => {
```
kathaypacific, line 521:
```suggestion
export const importedTokensSelector = createSelector([tokensListSelector], (tokenList): TokenBalance[] => {
```

i suggest putting the expected return type in the function definition and avoid casting with `as` in the return statement ( if the return statement is not returning `TokenBalance[]`, the `as` type cast will silence any warnings)

kathaypacific (approved):
💅🏼

PR #4879, opened 2024-02-08
fix: import select layout on large fonts and move files

MuckT:
### Description

Large font displays of this screen were broken causing some strange behavior on iOS and and text wrapping issues on Android 🤦

| Android Before | Android After |
| ----- | ----- |
|  |  |

#### Large Font iOS Video Before
https://github.com/valora-inc/wallet/assets/26950305/055b3aba-0080-4853-8ef8-465bdd4840f8

#### Large Font iOS Video After
https://github.com/valora-inc/wallet/assets/26950305/065876a9-b621-4b48-ab43-14d8b894db17

### Test plan

### Related issues

Fixes: ACT-874

### Backwards compatibility

Yes


cajubelt (approved):
can you follow up with whoever introduced these issues so they can learn how to prevent them with their own dev testing and solve them with the styling changes you added?

src/onboarding/registration/ImportSelect.tsx
```diff
@@ -132,40 +127,35 @@ const styles = StyleSheet.create({
   cardTitle: {
     ...typeScale.labelMedium,
     color: colors.primary,
+    flex: 1,
   },
   safeArea: {
-    alignItems: 'center',
```
MuckT, line 137: Removing this made the screen stop being jumpy on iOS.

src/onboarding/registration/ImportSelect.tsx
```diff
@@ -132,40 +127,35 @@ const styles = StyleSheet.create({
   cardTitle: {
     ...typeScale.labelMedium,
     color: colors.primary,
+    flex: 1,
```
MuckT, line 130: This flex is important so the text wraps instead of going into the card padding.

MuckT:
@cajubelt I introduced these changes and I'll leave the other changes as it decreases the number of CSS lines. 

The biggest contributors are called out in the comments, but to reiterate they are: make sure to pass flex down to children you want to inherent the flex property. Don't use `alignItems: 'center'` in a `SafeAreaView` styles.

PR #4884, opened 2024-02-09
chore: remove unused dapps remote configs

kathaypacific:
### Related issues

- Fixes #[issue number here]

### DIFF
--- a/src/app/saga.ts
+++ b/src/app/saga.ts
@@ -246,7 +246,6 @@ export interface RemoteConfigValues {
   showSwapMenuInDrawerMenu: boolean
   maxSwapSlippagePercentage: number
   networkTimeoutSeconds: number
-  dappFavoritesEnabled: boolean
   celoNews: CeloNewsConfig
   twelveWordMnemonicEnabled: boolean
   dappsMinimalDisclaimerEnabled: boolean
--- a/src/app/saga.ts
+++ b/src/app/saga.ts
@@ -248,7 +248,6 @@ export interface RemoteConfigValues {
   networkTimeoutSeconds: number
   celoNews: CeloNewsConfig
   twelveWordMnemonicEnabled: boolean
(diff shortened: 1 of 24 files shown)


e2e/src/usecases/DappListRecent.js
```diff
@@ -32,51 +32,19 @@ export default DappListRecent = () => {
...
     await navigateToHome()
 
     // Check that recently used dapp is now first in list
     const endRecentlyUsedDapps = await getElementTextList('RecentlyUsedDapps/Name')
-    jestExpect(dappPressedName).toEqual(endRecentlyUsedDapps[0])
+    jestExpect(dappToTest).toEqual(endRecentlyUsedDapps[0])
     jestExpect(endRecentlyUsedDapps.length).toEqual(startRecentDappCount + 1)
   })
 
-  it(':android: should show most recently used dapp leftmost on home screen', async () => {
```
kathaypacific, line 50: this test seems to do pretty much the same thing as above, i just combined them

e2e/src/usecases/DappListDisplay.js
```diff
@@ -17,18 +17,9 @@ export default DappListDisplay = () => {
     }
   })
 
-  it('should show dapp bottom sheet when dapp is selected', async () => {
-    await sleep(2000)
```
bakoushin, line 21: i wonder why this delay was there 🤔
could it be some kind of hack to ensure this test suite runs smoothly after the previous one?

bakoushin (approved):
🚀 Awesome!

e2e/src/usecases/DappListDisplay.js
kathaypacific: maybe - i've actually been trying to remove them since `sleep` seems like a brittle solution. most of the time we can fix the need for `sleep` by awaiting a specific UI change (but in cases like this one, the tests seem to function find without these sleeps)

PR #4886, opened 2024-02-09
feat: NFT celebration

bakoushin:
### Description

If a user has an NFT defined in `nft_celebration_config` remote config, on app open they will see a bottom sheet celebrating that NFT. Once bottom sheet is dismissed, users will see a celebration notification with a confetti animation with haptic feedback.

Celebration is displayed only once per installation/NFT.

Under feature flag `show_nft_celebration` (enabled for testers).

Analytics:
* celebration displayed (sent on celebration bottom sheet dismiss)
* animation displayed (send on animation end or dismiss by the user)

https://github.com/valora-inc/wallet/assets/2737872/110dd73d-8a25-4ae5-814b-2f9a7da9c073

### Test plan

* Tested manually on iOS and Android emulators
* Add unit test for celebration bottom sheet rendering

### Related issues

- Fixes RET-1000

### Backwards compatibility

Y


src/home/celebration/NftCelebration.tsx
```diff
@@ -0,0 +1,252 @@
...
+export default function NftCelebration() {
+  const { t } = useTranslation()
+  const dispatch = useDispatch()
+
+  const insets = useSafeAreaInsets()
+  const insetsStyle = { paddingBottom: Math.max(insets.bottom, Spacing.Regular16) }
+
+  const bottomSheetRef = useRef<BottomSheetRefType>(null)
+
+  const [showAnimation, setShowAnimation] = useState(false)
```
kathaypacific, line 36: since the bottom sheet and confetti animation are 2 separate things being coordinated in this file, it'd be helpful if the variable names were more specific to the thing it relates to?

kathaypacific (commented):
this looks great!! i recall you saying that it took some time for you to choose the "best" available lottie animation, and i saw you asked kayla to review the video. perhaps it'd helpful to provide kayla with some confetti animation options (your shortlist) so that she can choose?

src/home/WalletHome.tsx
```diff
@@ -207,6 +211,9 @@ function WalletHome() {
     </View>
   )
 
+  const showCashInBottomSheet = shouldShowCashInBottomSheet()
```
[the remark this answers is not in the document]
bakoushin: sure, that totally makes sense given the celebration is displayed only once.

src/home/celebration/NftCelebration.tsx
```diff
@@ -0,0 +1,252 @@
...
+    if (isVisible) {
+      const timeoutId = setTimeout(() => {
+        bottomSheetRef.current?.expand()
+      }, 1000 /* delay to ensure the bottom sheet is layed out */)
+      return () => clearTimeout(timeoutId)
+    }
+  }, [isVisible])
+
+  if (!isVisible) {
+    return null
```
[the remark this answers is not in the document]
bakoushin: sure!
I was caught with similar situations multiple times. I wonder if we can give a try to [ESLint hooks plugin](https://legacy.reactjs.org/docs/hooks-rules.html#eslint-plugin) which alerts if something deviates from the hooks's feng shui.

src/home/celebration/NftCelebration.tsx
```diff
@@ -0,0 +1,252 @@
...
+
+  const insets = useSafeAreaInsets()
+  const insetsStyle = { paddingBottom: Math.max(insets.bottom, Spacing.Regular16) }
+
+  const bottomSheetRef = useRef<BottomSheetRefType>(null)
+
+  const [showAnimation, setShowAnimation] = useState(false)
+  const animationStartTime = useRef(0)
+
+  const [nftsLoaded, setNftsLoaded] = useState(false)
```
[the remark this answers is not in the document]
bakoushin: yes, you are right: that trick wouldn't help if the fetch hadn't yet started. even worse, having misinterpreted the `nftsLoading` false value as _fetch is completed_ it will prevent the celebration from displaying.
 
my initial aim was to ensure that this bottom sheet doesn't reopen accidentally if the nfts fetch is restarted for some reason before we have finished the celebration, e.g. during the confetti animation. although it's not the case now, I wanted to ensure some future proof. uh, looks like this approach is a suboptimal way of doing it.

I think allowing a null value may not help, since the `nfts` list persists between launches. perhaps we can have a dedicated flag for displaying celebration right in the redux state, and perform all necessary checks within the saga. that way we can ensure that the fetch actually happened and succeeded. the celebration flag would be turned on after a successful fetch and turned off after a successful display. what do you think?

in case of initial fetch failure, I would prefer to leave the responsibility of initiating retries to the nfts fetching saga itself, rather than related components. would be happy to address it in a separate PR.

src/home/celebration/NftCelebration.tsx
```diff
@@ -0,0 +1,252 @@
...
+        onDismiss={handleDismissAnimation}
+      />
+    </>
+  )
+}
+
+const HandleWithImage = ({ nft }: { nft?: NftWithMetadata }) => {
+  const [showError, setShowError] = useState(false)
+  return (
+    <View style={styles.handleWithImage}>
```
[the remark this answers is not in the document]
bakoushin: thanks, sure. I should've use `NftMedia`

src/home/celebration/NftCelebration.tsx
```diff
@@ -0,0 +1,233 @@
...
+    dispatch(nftCelebrationDisplayed())
+  }
+
+  if (!isVisible) {
+    return null
+  }
+
+  return (
+    <>
+      <GorhomBottomSheet
```
kathaypacific, line 143: nit: any chance we can use `BottomSheetBase` here? it could save some code (the background would come for free) - i think it'd work if you add an optional `handleComponent` prop to the `BottomSheetBase`? (the `onChange` handler would probably work with the `onClose` prop in `BottomSheetBase`)

src/home/celebration/NftCelebration.tsx
```diff
@@ -0,0 +1,233 @@
...
+        backdropComponent={renderBackdrop}
+        handleComponent={renderHandleWithImage}
+        backgroundStyle={styles.bottomSheetBackground}
+        onChange={handleBottomSheetPositionChange}
+      >
+        <BottomSheetView style={[styles.container, insetsStyle]}>
+          <View style={styles.content}>
+            <Text style={styles.title}>{t('nftCelebration.bottomSheet.title')}</Text>
+            <Text style={styles.description}>
+              {t('nftCelebration.bottomSheet.description', { appName: APP_NAME })}
```
kathaypacific, line 157: nit: appName comes for free [here](https://github.com/valora-inc/wallet/blob/alex/gift-celebration/src/i18n/index.ts#L57) so you don't need to pass it in here :)

src/nfts/NftMedia.tsx
```diff
@@ -17,7 +17,7 @@ import networkConfig from 'src/web3/networkConfig'
 const DEFAULT_HEIGHT = 360
 
 interface PlaceHolderProps {
-  testID: string
+  testID?: string
```
kathaypacific, line 20: why not just assign a testId for the celebration nft image, rather than make the prop optional? i see some assumptions in the component that relies on this (.e.g `testID={`${testID}/VideoPlaceholder`}`)

kathaypacific (commented):
this is looking great!! sorry, a few more nit comments.

just checking if you looped in with Kayla about which confetti animation to use? i don't want to inject my personal preference here too much so ultimately the decision is yours + Kayla's. i subjectively feel like the current animation is not so fancy and has some colors that are contrasting to our brand colors (e.g. lavender and baby blue) but maybe this is the intention, in which case all good :) 

one other thing i noticed is the animation for the bottom sheet appearing is sometimes staggered, and pressing on the confetti background doesn't dismiss the animation (it takes a couple of tries, in the below video you can see the toast move up a little bit - that is when i first tapped the screen). maybe this is my simulator being slow, is it worth delaying the celebration by 2000ms instead?

https://github.com/valora-inc/wallet/assets/20150449/26b5e730-0283-4edf-a384-38f1fdad2e36

kathaypacific (approved):
TY for checking the animation with Kayla! my nit comments are not blocking so address the ones that you see value in :) let's see how the animation performs in a real (nightly) build on our physical devices and decide on optimisations then

src/home/celebration/NftCelebration.tsx
bakoushin: yes, that makes sense. in addition to `handleComponent` I've also forwarded `backgroundStyle` and `onChange` props:
* `backgroundStyle` because we have to tweak the default background to not interfere with the rounded corners of the image (default background doesn't hide overflow)
* `onChange` because the `onClose` event is dispatched not only when the bottom sheet index changes to `-1`, but also when the bottom sheet component unmounts, which caused analytics to be sent twice.
also, it caused a strange behavior on the Android emulator when sometimes `onClose` was fired even before the bottom sheet appeared. i guess the bottom sheet was mounted and unmounted at least once (this never happened on iOS)
having noticed this, relying on `onChange` seems more safe

PR #4887, opened 2024-02-09
refactor(i18n): use static translation keys

cajubelt:
### Description

...instead of values known only at runtime. This would make it easier to run static extraction of translatable strings later if we want. 

Meant to address pr comment from @jeanregisser [here](https://github.com/valora-inc/wallet/pull/4842/files#r1484091599)

### Test plan

ci

### Related issues

na

### Backwards compatibility

na


jh2oman (approved):

jeanregisser (commented):
Awesome, thanks! 👍

PR #4891, opened 2024-02-10
feat(keylessBackup): restore not found screen

jh2oman:
### Description

The screen for when a user is restoring with CAB and we do not have a record for the requested encrypted mnemonic.

Screenshot_20240209_185708

### Test plan

Added unit tests

### Related issues

- Fixes https://linear.app/valora/issue/ACT-779/wallet-not-found-screen

### Backwards compatibility

n/a


src/keylessBackup/KeylessBackupProgress.tsx
```diff
@@ -87,12 +88,16 @@ function Restore() {
   const dispatch = useDispatch()
 
   const onPressTryAgain = () => {
-    ValoraAnalytics.track(KeylessBackupEvents.cab_restore_failed_try_again)
+    dispatch(keylessBackupBail())
```
jh2oman, line 91: added these to clear the keylessBackup state so that if a user tries again we ask them for new info

src/keylessBackup/index.test.ts
```diff
@@ -96,7 +96,7 @@ describe(getEncryptedMnemonic, () => {
 
   it('throws if non 200 response', async () => {
```
jophish, line 97: nit: change test name here to be a bit more descriptive now that we discriminate between different types of non-200s?

src/keylessBackup/types.ts
```diff
@@ -6,7 +6,8 @@ export enum KeylessBackupFlow {
 export enum KeylessBackupStatus {
   NotStarted = 'NotStarted',
   InProgress = 'InProgress',
-  RestoreZeroBalance = 'RestoreZeroBalance',
+  RestoreZeroBalance = 'RestoreZeroBalance', // only in restore flow
   Completed = 'Completed',
   Failed = 'Failed',
+  NotFound = 'NotFound', // only in restore flow
```
jophish, line 12: Maybe this is just me, but I find mixing the states between recovery and backup into a single enum (and piece of redux state) a little confusing. I understand why we do this, but at first glance it is a bit confusing to me exactly what each state represents between the different flows - the semantics of each state are different depending on the context. It's also not immediately clear if _all_ states are possible in recovery, and only a subset in setup (that is indeed the case, but again, not obvious on first glance). This is all pretty easy to grok though once you go through the code so nothing to really do here I think, just wanted to bring it up.

src/keylessBackup/index.test.ts
```diff
@@ -96,7 +96,7 @@ describe(getEncryptedMnemonic, () => {
...
     mockSiweFetch.mockResolvedValueOnce({
-      status: 404,
+      status: 500,
       ok: false,
       json: () => Promise.resolve({ message: 'not found' }),
```
jophish, line 101: also tiny nit: change 'not found' to 'server error' or something similar

src/keylessBackup/KeylessBackupProgress.tsx
jophish: where is the `X` button (in the top left corner)  behavior defined? it should be identical to this case, right?

jophish: oop, after looking at the image in the PR description again, I'm actually not seeing an X button as appears in [the designs](https://linear.app/valora/issue/ACT-779/wallet-not-found-screen)

src/keylessBackup/KeylessBackupProgress.tsx
```diff
@@ -206,6 +211,38 @@ function Restore() {
           />
         </SafeAreaView>
       )
+    case KeylessBackupStatus.NotFound:
+      return (
+        <SafeAreaView style={styles.container}>
+          <View style={styles.finishedContainer}>
```
MuckT, line 217: Would make this a scroll view just to be safe on smaller phones.

MuckT (approved):

src/keylessBackup/types.ts
jh2oman: I'm with you that I think this is a stretch to have them combined jointly. I cringed a little adding this `NotFound` but didn't think it was worth refactoring everything to be separate between restore and setup at this point. Thanks for brining it up for discussion!

src/keylessBackup/KeylessBackupProgress.tsx
jh2oman: That's so weird. It looks like the designs were changed in Figma to remove the X but the preview in linear didn't update.

MuckT (approved):

PR #4895, opened 2024-02-12
ci: switch to namespace runners for Android e2e tests

jeanregisser:
### Description

This switches the runners for our Android e2e tests to use https://namespace.so
- faster builds, best run was under 10 mins total! 🚀 
- no more maintenance of our dedicated machines!
- more secure, each run gets an isolated instance

### Test plan

- Android e2e test pass

### Related issues

- Fixes RET-751

### Backwards compatibility

Yes


kathaypacific (approved):
🚀

MuckT (approved):

PR #4900, opened 2024-02-13
fix: token bottom sheet interactions with filters

kathaypacific:
### Description

This PR tweaks the token bottom sheet interactions and fixes the [scroll bug](https://valora-app.slack.com/archives/C04B61SJ6DS/p1707757919681089). As discussed [here](https://valora-app.slack.com/archives/C0684TXDR3K/p1707758094751239):
1. the filters now use an "and" condition, so the results will always change when a filter is selected.
2. the default filter for the FROM field is "my tokens", and default filter for TO is "popular"
3. to remind users which filters are selected, the filter carousel now has an option not to scroll and does not scroll for the token bottom sheet (this was in a DM with Kayla, but we explored various options and landed on this one)
4. Kayla suggested the persistence of the user's selections should be per list, so the SwapScreen now has 2 token bottom sheets that do not have any filter / search / scroll reset between launches anymore. 
5. there were a lot of changes in the swap screen test file - it was mainly to account for multiple token bottom sheets and being more specific with actions / assertions (using `within`)

### Test plan

https://github.com/valora-inc/wallet/assets/20150449/7436e51f-0ac4-47d3-b051-f6961d8e4f32

### Related issues

- Related to RET-984

### Backwards compatibility

Y


src/components/TokenBottomSheet.tsx
```diff
@@ -259,48 +272,45 @@ function TokenBottomSheet<T extends TokenBalance>({
...
-        stickyHeaderIndices={[0]}
+        ListHeaderComponent={<View style={{ height: headerHeight }} />}
         ListEmptyComponent={() => {
           if (searchEnabled || filterChips.length > 0) {
             return <NoResults searchTerm={searchTerm} activeFilters={activeFilters} />
           }
           return null
         }}
       />
+      <View style={styles.headerContainer} onLayout={handleMeasureHeader}>
```
kathaypacific, line 283: since the bottom sheet is only mounted once, the header is only measured once. if reusing the same bottom sheet with different headers, this isn't going to work (but we don't have a use case for that so i didn't worry much about it)

kathaypacific: JK, that's not how onLayout works. this will be fine because onLayout is called when the layout changes too

dievazqu:
Do you mind sharing some videos showing the new interactions?

src/swap/SwapScreen.tsx
```diff
@@ -762,23 +785,22 @@ export function SwapScreen({ route }: Props) {
...
+          filterChips={filterChips}
+          origin={origin}
+          snapPoints={['90%']}
+          onTokenSelected={handleSelectToken}
+          searchEnabled={true}
+          TokenOptionComponent={TokenBalanceItemOption}
+          showPriceUsdUnavailableWarning={true}
+          areSwapTokensShuffled={areSwapTokensShuffled}
+        />
+      ))}
```
jeanregisser, line 803: From what I see we need to mount 2 bottom sheets because we want to keep the selected filter chips state separate whether it's "from" vs "to" , right?

What happens when different networks are selected? and the user switches between the from and to? does it feel natural?
I guess I should play with it myself :D

src/swap/SwapScreen.tsx
```diff
@@ -815,7 +837,7 @@ export function SwapScreen({ route }: Props) {
       <BottomSheet
         forwardedRef={slippageInfoBottomSheetRef}
         description={t('swapScreen.transactionDetails.slippageToleranceInfo')}
-        testId="NetworkFeeInfoBottomSheet"
+        testId="SlippageInfoBottomSheet"
```
jeanregisser, line 840: 👍

jeanregisser (approved):
🚀

src/swap/SwapScreen.tsx
kathaypacific: yes - we want the "from" and "to" sheets to behave independently, and for the user to pick up where they left off if they go back to either sheet.

interesting question about whether switching networks feels natural. i'm not sure what my personal preference is, but i think it does not feel so clunky  / unintuitive now that the filters are so easily accessible and you can see at a glance what the current selection is. i'm also happy to merge as is and update it depending on feedback once we've been able to test it on a physical device in the nightly build

PR #4901, opened 2024-02-13
chore(token-import): Adding analytics for the send flow

dievazqu:
### Description

Adding `isTokenManuallyImported` field to the analytics in the send flow.


src/tokens/saga.test.ts
```diff
@@ -381,14 +387,103 @@ describe(fetchImportedTokenBalances, () => {
...
       [mockUSDCTokenId]: {
         ...mockImportedTokens[mockUSDCTokenId],
         balance: new BigNumber(10).toFixed(),
+        priceUsd: undefined,
+      },
+    })
+  })
+})
+
+describe(fetchImportedTokenBalances, () => {
```
kathaypacific, line 406: is this a bad merge? it seems like this test group already exists?

kathaypacific (commented):
i think there's a bad merge here for the tests. will there be a separate PR for the swap flow analytics? could we also make this PR title more specific to the changes (e.g. adding analytics for send flow) so that the change log is clearer? 🙏🏻

dievazqu:
> i think there's a bad merge here for the tests. will there be a separate PR for the swap flow analytics? could we also make this PR title more specific to the changes (e.g. adding analytics for send flow) so that the change log is clearer? 🙏🏻

Yes, the analytics for the swaps are coming with the PR to improve the swap experience for the imported tokens.
And sorry for the PR title, updated.

src/tokens/saga.test.ts
dievazqu: Oh, yes, I messed it up in the merge 🤯

kathaypacific (approved):

PR #4902, opened 2024-02-13
fix(send): Handling prepare transaction errors in send flow

dievazqu:
### Description

Showing an error message in cases where the transaction can't be prepared. For example, when trying to send weird tokens like MNEC (0xf9d922c055a3f1759299467dafafdf43be844f7a)

### Test plan

Locally in an emulator

### Related issues

Fixes RET-1008

### Backwards compatibility

Y

### Demo

https://github.com/valora-inc/wallet/assets/12504748/ea8ecf7c-ea11-41e7-9072-27874860289e


locales/base/translation.json
```diff
@@ -2073,6 +2073,10 @@
     "notEnoughBalanceForGasWarning": {
       "title": "You need more {{feeTokenSymbol}} for gas",
       "description": "Add {{feeTokenSymbol}} to cover your gas fees or adjust the amount you are sending"
+    },
+    "feeEstimationError": {
+      "title": "Error estimating network fee",
+      "description": "There was an error estimating the network fee. Please try again later."
```
dievazqu, line 2079: This error message is a bit generic, but this could appear in case of any error while calculating the fees.
In the MNEC case, probably we want to show a more precise message indicating the problem won't be fixed?
If we want that, I'll address it in another PR.

src/send/usePrepareSendTransactions.ts
```diff
@@ -68,10 +68,12 @@ export function usePrepareSendTransactions() {
   const [prepareTransactionsResult, setPrepareTransactionsResult] = useState<
     PreparedTransactionsResult | undefined
   >()
+  const [prepareTransactionError, setPrepareTransactionError] = useState<Error | undefined>()
```
jeanregisser, line 71: We don't need to store the error, it's already stored in the async callback hook.

We can return it like this: `prepareTransactions.error`

(same for the result by the way).

src/send/SendEnterAmount.tsx
```diff
@@ -383,6 +392,15 @@ function SendEnterAmount({ route }: Props) {
...
         )}
+        {prepareTransactionError && (
+          <InLineNotification
+            severity={Severity.Error}
+            title={t('sendEnterAmountScreen.feeEstimationError.title')}
+            description={t('sendEnterAmountScreen.feeEstimationError.description')}
+            style={styles.warning}
+            testID="SendEnterAmount/PrepareTransactionError"
+          />
+        )}
```
jeanregisser, line 403: Nothing to do now, but it could be a good idea to surface some details about the error (like we recently did for WC prepare transactions failures in https://github.com/valora-inc/wallet/pull/4864).

jeanregisser (commented):
Thanks!

src/send/SendEnterAmount.tsx
jeanregisser: oh, interesting attack vector 🤔

jeanregisser (approved):
🚀

PR #4904, opened 2024-02-13
feat: add use of one time jwt to cloud account storage

MuckT:
### Description

Follow up to ACT-718 that uses the one time jwt in the wallet when storing the encrypted mnemonic.

### Test plan

- Tested locally on iOS
- Unit tests updated

### Related issues

- Fixes ACT-1096

### Backwards compatibility

Yes


src/keylessBackup/slice.ts
```diff
@@ -4,13 +4,15 @@ import { KeylessBackupFlow, KeylessBackupStatus } from 'src/keylessBackup/types'
 export interface State {
   googleIdToken: string | null
   valoraKeyshare: string | null
+  jwt: string | null
```
MuckT, line 7: Thinking we might not actually need this in redux.

MuckT: Removed jwt from redux in 3f3ff34b3bd6ae7b6378951fb586eb6b3fed9aa6.

src/keylessBackup/index.ts
```diff
@@ -28,6 +30,7 @@ export async function storeEncryptedMnemonic({
     body: JSON.stringify({
       encryptedMnemonic,
       encryptionAddress,
+      token: jwt,
```
cajubelt, line 33: I think JWTs are normally given in an auth header

MuckT: Usually yes, @satish-ravi commented something similar [here](https://github.com/valora-inc/cloud-account-backup/pull/158#discussion_r1462537029), but since we're not using as an auth token and instead as a one time token and I opted to keep it in the body. This avoids having to setup middleware for the backend for one endpoint; Also IIRC there was also some default handling of jwt's in as auth header where it was passed to subsequent requests that was also undesirable.

jh2oman (approved):

PR #4908, opened 2024-02-14
feat: add delete account settings item

kathaypacific:
### Description

As the title.

### Test plan

Example of success:
https://github.com/valora-inc/wallet/assets/20150449/7a3303b4-0778-402c-9277-18d6fd2ec3cf

Example of failure to revoke:

https://github.com/valora-inc/wallet/assets/20150449/6aab8db4-0e9d-46a5-9c06-ac512c76b2b2

### Related issues

n/a

### Backwards compatibility

Y


src/account/Settings.tsx
```diff
@@ -445,17 +472,25 @@ export const Account = ({ navigation, route }: Props) => {
...
-              testID="RevokePhoneNumber"
-            />
+            <>
+              <SettingsExpandedItem
+                title={t('revokePhoneNumber.title')}
+                details={t('revokePhoneNumber.description')}
+                onPress={handleShowConfirmRevoke}
+                testID="RevokePhoneNumber"
+              />
+              <SettingsExpandedItem
```
kathaypacific, line 482: unverified users don't need this "reset account" button, since the "delete account" will do the same thing for them (and "delete account" is more compliant with app store guidelines)

locales/base/translation.json
```diff
@@ -6,6 +6,14 @@
   "licenses": "Licenses",
   "removeAccountTitle": "Reset Valora",
   "removeAccountDetails": "Resetting will remove your account from this device. Your funds will remain in the account, but will only be accessible with your Recovery Phrase",
+  "deleteAccountTitle": "Delete Account",
+  "deleteAccountDetails": "Deleting your account will remove it from {{appName}}. Your funds will remain in the account, but will only be accessible with your Recovery Phrase",
```
MuckT, line 10: Should this include copy about also unlinking the phone number?

MuckT:
Two small things.
First the copy for the descriptions of reset and delete is very similar, should the delete copy include a mention of unlinking the phone number.

Second the modal served after completing the account delete references reset which could be confusing to users.

| Duplicate descriptions | Reset Modal (After Delete Account) |
| ---- | ---- |
|  |  |

src/account/Settings.tsx
jeanregisser: Ah, but aren't users used to this reset account?
Maybe it's fine if these 2 buttons currently do the same thing if the user is not verified.

jeanregisser (approved):
Thanks 👍 

Not entirely sure about the wording we should use, but we can tweak this later too.

kathaypacific:
> Two small things. First the copy for the descriptions of reset and delete is very similar, should the delete copy include a mention of unlinking the phone number.
> 
> Second the modal served after completing the account delete references reset which could be confusing to users.
> 
> Duplicate descriptions	Reset Modal (After Delete Account)
> 	

hey @MuckT TY for the feedback! 

the copy on the buttons is what was last suggested in [slack](https://valora-app.slack.com/archives/C025X4WJT09/p1707881309542459?thread_ts=1707844056.413359&cid=C025X4WJT09) - this is a new button label that we can also ship over the air updates for if there is a better suggestion. i do mention the unlinking of phone number in the bottom sheet that is launched on button press. i think that we shouldn't block this feature on this button copy.

the modal text is an interesting point, we can either create a new modal and separate the copy or continue to share the copy across the reset and delete flows. i want to limit the scope of this feature because of the time sensitivity (i had a look into making a new warning and there's navigation state that is shared in many places since the backup flow is involved, and those changes make this PR much harder to review and test). so i had a go at making the copy more sharable, not sure if i nailed it. i suggest shipping over the air copy that can be shared across the 2 flows, unless someone has a better suggestion now? i've been asking chatgpt all morning but i'm not really happy with any of the suggestions 😅

src/account/Settings.tsx
kathaypacific: yeah true, okay i can leave both of them always visible

PR #4910, opened 2024-02-14
build(ios): set os minimum deployment to 15.0

MuckT:
### Description

- Sets the min deployment target for iOS to 15.0: [Slack thread](https://valora-app.slack.com/archives/C025Q9QG8JG/p1707945581586849).
- Sets E2E tests on merges to main only run on iOS 15.0.

### Test plan

- CI E2E tests

### Related issues

- Fixes ACT-1099

### Backwards compatibility

Yes, except for iOS versions < 15.0


.github/workflows/e2e-main.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       max-parallel: 2
       fail-fast: false
       matrix:
-        ios-version: ['14.5', '15.0']
```
jeanregisser, line 46: Can we add iOS 17 then? Since we want to test the lowest and highest supported.

jeanregisser (approved):
Awesome! 🚀 

Before we merge, did you get a confirmation from product?

.github/workflows/e2e-main.yml
MuckT: This needs to be done in a follow up PR. Once we have product approval we can setup the new runners, migrate off the old ones and then enable testing of iOS 17.

PR #4917, opened 2024-02-15
chore: add extra logging for currency conversion, remove logging for rewards senders

kathaypacific:
### Description

Hopefully the currency conversion log will help us debug issues with reports of incorrect fiat values. I am removing the rewards senders log because it is long and doesn't seem to be helpful, and makes me feel better about adding more logs.

### Test plan

n/a

### Related issues

- Related to SUP-487

### Backwards compatibility

Y

### Network scalability

Y


bakoushin (approved):
🔭

PR #4919, opened 2024-02-15
fix: token bottom sheet sticky header position

kathaypacific:
### Description

[Context](https://valora-app.slack.com/archives/C04B61SJ6DS/p1707757919681089)

### Test plan

The sticky header is no longer stuck to the wrong place sometimes when launching the bottom sheet.

### Related issues

- Relates to RET-984

### Backwards compatibility

Y

### Network scalability

Y

### DIFF
--- a/src/components/TokenBottomSheet.tsx
+++ b/src/components/TokenBottomSheet.tsx
@@ -152,6 +152,7 @@ function TokenBottomSheet<T extends TokenBalance>({
 
   const filterChipsCarouselRef = useRef<ScrollView>(null)
   const tokenListRef = useRef<BottomSheetFlatListMethods>(null)
+  const [headerHeight, setHeaderHeight] = useState(0)
   const [searchTerm, setSearchTerm] = useState('')
   const [filters, setFilters] = useState(filterChips)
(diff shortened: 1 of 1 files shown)


src/components/TokenBottomSheet.tsx
```diff
@@ -239,78 +242,91 @@ function TokenBottomSheet<T extends TokenBalance>({
     filterChipsCarouselRef.current?.scrollTo({ x: 0 })
   }
 
+  const handleMeasureHeader = (event: { nativeEvent: { layout: { height: number } } }) => {
+    setHeaderHeight(event.nativeEvent.layout.height)
```
jeanregisser, line 246: Should we reset the height to 0 after closing?

jeanregisser (approved):
🚀

src/components/TokenBottomSheet.tsx
kathaypacific: why do we want to set it to 0? `onLayout` will be called when the component changes, so if the header components do change then the token bottom sheet will already update in the background. my understanding is that the token bottom sheet isn't unmounted so setting the height to 0 after close will not guarantee that the height stays at 0 until the next open anyway?

PR #4921, opened 2024-02-16
chore: add analytics for token selection position in list

kathaypacific:
### Description

This PR adds more tracking so that we can learn:
- how helpful is our token guidance (track the position in the list of the token selected - are users needing to scroll a lot?)
- is the holdout group experience doing harm to swap executions (are there less swaps being executed by this group)

### Test plan

unit tests

### Related issues

- Fixes RET-984

### Backwards compatibility

Y

### Network scalability

Y


jeanregisser (approved):
LGTM 👍

PR #4922, opened 2024-02-16
fix(nfts): NFT celebration dynamic config `defaultValues`

bakoushin:
### Description

All possible dynamic config properties must be listed in `defaultValues` when declaring that config. 
Otherwise, they [won't be forwarded](https://github.com/valora-inc/wallet/blob/a0b4675774579822a6c2ac55e7de38f6658b45c2/src/statsig/index.ts#L29-L34). 

### Test plan

* Updated unit tests

### Related issues

Related to RET-1000

### Backwards compatibility
 
NFT celebration is not released yet

### Network scalability


src/home/celebration/NftCelebration.test.tsx
```diff
@@ -66,10 +66,6 @@ const mockStoreWithDifferentNft = {
 describe('NftCelebration', () => {
   beforeEach(() => {
     jest.mocked(getFeatureGate).mockReturnValue(true)
-    jest.mocked(getDynamicConfigParams).mockReturnValue({
```
bakoushin, line 69: this mock was redundant for this particular test case

dievazqu (approved):

PR #4926, opened 2024-02-16
chore(wc): remove `use_viem_for_walletconnect_transactions` feature flag

jeanregisser:
### Description

As the title says.

Should also speed up WC e2e tests since it won't need to go through the ContractKit variant anymore.

### Test plan

- Updated tests
- Tests pass

### Related issues

- Fixes RET-1018

### Backwards compatibility

Yes (beside it will not read the feature flag anymore).

### Network scalability


e2e/src/usecases/WalletConnectV2.js
jeanregisser, line 1: The diff is hard to read because of the changed indentation, sorry.

src/walletConnect/saga.ts
```diff
@@ -671,10 +664,7 @@ function* handleAcceptRequest({ request, preparedTransaction }: AcceptRequest) {
     }
 
     const result = yield* call(handleRequest, params, preparedTransaction)
-    const response: JsonRpcResult<string> = formatJsonRpcResult(
-      id,
-      (params.request.method = typeof result === 'string' ? result : result.raw)
```
jeanregisser, line 676: Well, looks like this was a bug, why were we mutation the method here?! 😅

kathaypacific: i thought this might have been related to the v2 upgrade (i recall there was one method like eth_sign or something that was causing issues with celo transactions) but it looks like it was unrelated, this was introduced in https://github.com/valora-inc/wallet/pull/3970/files#diff-3560064301c178aea299929cc30db71c6464c5c176261d928f31a4d8d4b8a56bR449 and i'm not sure why. looks fine to remove

kathaypacific (approved):
🙏🏻

bakoushin (approved):

PR #4930, opened 2024-02-16
fix(token-details): Updating logic to show "Swap" for zero balance tokens and imported tokens

dievazqu:
### Description

As title says.
This PR makes the logic consistent between the token details and the options shown in the send and swap flow.

Also this PR adds the imported tokens to the swap flow.

### Test plan

Unit tests and locally in a emulator

### Backwards compatibility

Y

### Network scalability

### Demo

https://github.com/valora-inc/wallet/assets/12504748/c72f889a-4fed-4443-ab56-8fdb7d203c2d


src/tokens/hooks.test.tsx
```diff
@@ -240,10 +241,11 @@ describe('useTokensForSend', () => {
     )
 
     expect(getByTestId('tokenIDs').props.children).toEqual([
-      mockPoofTokenId,
```
dievazqu, line 243: The order or this has changed because now it uses the order from `tokensWithNonZeroBalanceAndShowZeroBalanceSelector`

src/tokens/TokenDetails.tsx
```diff
@@ -210,8 +209,7 @@ export const useActions = (token: TokenBalance) => {
       },
```
dievazqu, line 209: I'll change this logic, if the user doesn't have any balance then I'll send them to the swap screen but with this token in the "to token"

src/tokens/TokenDetails.test.tsx
```diff
@@ -449,4 +451,39 @@ describe('TokenDetails', () => {
...
+      <Provider store={store}>
+        <MockedNavigator
+          component={TokenDetailsScreen}
+          params={{ tokenId: mockTestTokenTokenId }}
+        />
+      </Provider>
+    )
+
+    expect(queryByTestId('TokenDetails/Action/Send')).toBeFalsy()
+    expect(queryByTestId('TokenDetails/Action/Swap')).toBeTruthy()
```
kathaypacific, line 484:
```suggestion
    expect(getByTestId('TokenDetails/Action/Swap')).toBeTruthy()
```

src/tokens/selectors.ts
```diff
@@ -366,6 +368,7 @@ export const swappableToTokensByNetworkIdSelector = createSelector(
     return tokens.filter(
       (tokenInfo) =>
         tokenInfo.isSwappable ||
+        tokenInfo.isManuallyImported ||
```
kathaypacific, line 371: do you need this when the condition is already there in `swappableFromTokensByNetworkIdSelector`?

src/tokens/hooks.test.tsx
```diff
@@ -223,8 +223,8 @@ describe('useTokensForSend', () => {
...
       mockCeloTokenId,
+      mockPoofTokenId,
```
kathaypacific, line 227: just checking that we want to have a different sort order for the send tokens now?

src/send/SendEnterAmount.test.tsx
```diff
@@ -47,7 +47,7 @@ const mockStore = {
       }, // fee currency, matches mockCeloTokenBalance
       [mockEthTokenId]: {
         tokenId: mockEthTokenId,
-        balance: '0',
+        balance: '0.1',
```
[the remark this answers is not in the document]
dievazqu: So, now we don't want to show tokens without balance in the SendEnterAmount, and there was a test failing because Eth wasn't present, so this was the quick fix.
But, maybe I can put this in 0 again and change the failing test a bit.

src/tokens/hooks.test.tsx
dievazqu: This is using the latest order logic with decided. The previous version didn't have any kind of sorting.

src/tokens/selectors.ts
dievazqu: Hmm, I guess we have to, right?
Because if not, those filters will filter it out?

That being said, I don't like that the `toToken` are built from the `fromToken`. I imagine a case where we want a token to be in the `toToken` list but not in the `fromToken` (for example, tokens without balance).
I think both selector should use the whole token list.

src/tokens/TokenDetails.tsx
dievazqu: No, I will do this in another PR. Because the swap screen doesn't support a `toTokenId` as initial param. So this will need some changes there.
I think it's fine to merge this PR without this.

src/tokens/Assets.tsx
```diff
@@ -117,9 +117,7 @@ function AssetsScreen({ navigation, route }: Props) {
   const activeTab = route.params?.activeTab ?? AssetTabType.Tokens
 
   const supportedNetworkIds = getSupportedNetworkIdsForTokenBalances()
-  const tokens = useSelector((state) =>
-    tokensWithNonZeroBalanceAndShowZeroBalanceSelector(state, supportedNetworkIds)
-  )
+  const tokens = useSelector((state) => tokensWithNonZeroBalance(state, supportedNetworkIds, true))
```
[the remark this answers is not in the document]
kathaypacific: haha, good question. is it possible to have a selector that returns tokens with > 0 balance + zero state tokens, and have a second selector that uses that first selector and filters out the zero state tokens? i know that's inefficient from a looping perspective, but this operation is cached so it seems like the readability vs performance tradeoff is worthwhile?

src/tokens/selectors.ts
kathaypacific: yes you're right, my brain confused the intention of the .filter here yesterday.

and i think i understand the concern with the shared selectors - i originally built these selectors totally separate and updated these to share code to make it easier to see the difference between the 2. i agree in the future if we find the selectors are diverging then we should totally separate them completely

src/tokens/Assets.tsx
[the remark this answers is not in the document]
kathaypacific: oh this is what you already did!! awesome!!

kathaypacific (approved):
✨🚀

PR #4936, opened 2024-02-19
fix(wc): allow personal_sign for unsupported networks

kathaypacific:
### Description

Context https://valora-app.slack.com/archives/C04B61SJ6DS/p1708336430158639?thread_ts=1708021233.998389&cid=C04B61SJ6DS

### Test plan

Unit tests + the tofu nfts scenario in the linked slack thread works now.

### Related issues

n/a

### Backwards compatibility

Y

### Network scalability


src/walletConnect/request.ts
```diff
@@ -23,12 +24,28 @@ export function* handleRequest(
...
+    SupportedActions.personal_sign,
+  ]
+  if (!network && !chainAgnosticActions.includes(method as SupportedActions)) {
     throw new Error('unsupported network')
   }
 
-  const wallet: ViemWallet = yield* call(getViemWallet, networkConfig.viemChain[network])
+  const wallet: ViemWallet = yield* call(
+    getViemWallet,
+    networkConfig.viemChain[network ?? Network.Celo]
```
bakoushin, line 47: what if we resolve network from `networkConfig.defaultNetworkId`?

bakoushin (approved):
🚀 Awesome! Left one small nit about network hardcoding.

src/walletConnect/request.test.ts
```diff
@@ -120,24 +120,8 @@ describe(handleRequest, () => {
...
-  it('unlocks the wallet address when a MTW address is set', async () => {
-    const state = createMockStore({ web3: { account: '0xWALLET', mtwAddress: '0xMTW' } }).getState()
-    await expectSaga(handleRequest, signTransactionRequest, serializableTransactionRequest)
-      .provide([[call(getWallet), mockWallet]])
-      .withState(state)
-      .call(unlockAccount, '0xwallet')
-      .run()
-  })
-
-  it('unlocks the wallet address when a MTW address is NOT set', async () => {
```
kathaypacific, line 132: mtw stuff isn't needed anymore

src/walletConnect/request.test.ts
```diff
@@ -154,28 +138,33 @@ describe(handleRequest, () => {
...
       .run()
+
+    await expectSaga(handleRequest, { ...personalSignRequest, chainId: 'eip155:unsupported' })
+      .withState(state)
+      .call([viemWallet, 'signMessage'], { message: { raw: 'Some message' } })
+      .run()
   })
 
-  it('supports eth_signTypedData', async () => {
+  it('supports eth_signTypedData for supported chain', async () => {
```
bakoushin, line 153: nit: i'm a bit confused with teh logic of adding the "supported chain" mention.

i'm doubtful that it's worth to mentioning "supported chain" here, since we don't have the opposite test (throws for unsupported chain).

however, for `eth_signTransaction` and `eth_sendTransaction` we have both. i would expect to have "supported"/"unsupported" in their naming, rather than here.

PR #4939, opened 2024-02-20
chore: add ESLint plugin to VS Code recommendations list

bakoushin:
### Description

We have ESLint rules set up, but don't have ESLint plugin among recommendations. This PR aims to fix this.

### Test plan

NA

### Related issues

NA

### Backwards compatibility

NA

### Network scalability

NA


kathaypacific (approved):
✨

PR #4942, opened 2024-02-20
feat(cab): Use new verifiers with custom domains, update torus libs

jophish:
### Description

Kind of a two-birds-one-stone PR, both for [ACT-875](https://linear.app/valora/issue/ACT-875/use-a-custom-domain-with-auth0) and [ACT-876](https://linear.app/valora/issue/ACT-876/refactor-gettorusprivatekey-to-use-updated-torusweb3auth-dependencies). Basically the work that led to this PR progressed as follows:
* Updated Auth0 to use custom domains for prod/dev
* Tried to update our custom verifiers in web3auth dashboard to use new Auth0 domains
  * Our existing verifiers were nowhere to be found in the dashboard
  * Created new verifiers using the newest Sapphire main/devnets (as opposed to deprecated Cyan network, which our previous verifiers were on)
* Had trouble adapting our existing wallet code to integrate with the new verifiers
* Updated `@torus/*` libraries and referred to their docs for how to migrate, was able to connect to new verifiers
* Was able to go through backup + restore using custom Auth0 domains _and_ new verifiers on Alfajores.

This fulfills the requirements for both ACT-875 and ACT-876 AFAICT. I haven't yet tried to directly use the web3auth libs _instead_ of the torus libs; I'll give it a shot, but if it doesn't seem to work, the changes in this ticket should also suffice.

### Test plan

Manual tested.

### Related issues

- Fixes [ACT-875](https://linear.app/valora/issue/ACT-875/use-a-custom-domain-with-auth0) and [ACT-876](https://linear.app/valora/issue/ACT-876/refactor-gettorusprivatekey-to-use-updated-torusweb3auth-dependencies).

### Backwards compatibility

This is not backwards compatible, in the sense that someone who's backed up will not be able to restore, but this shouldn't matter since we haven't gone live with CAB yet.

### Network scalability


package.json
```diff
@@ -113,8 +113,9 @@
     "@sentry/react-native": "^4.15.2",
     "@sentry/types": "^7.101.0",
     "@th3rdwave/react-navigation-bottom-sheet": "^0.2.7",
-    "@toruslabs/fetch-node-details": "^8.1.1",
-    "@toruslabs/torus.js": "^6.4.1",
+    "@toruslabs/constants": "^13.1.0",
+    "@toruslabs/fetch-node-details": "^13.1.1",
+    "@toruslabs/torus.js": "^12.1.1",
```
jh2oman, line 118: 6 major versions! 🤯

jh2oman (commented):
Awesome! @jophish can you add test coverage for `src/keylessBackup/web3auth.ts` ? Then we could close out this TODO https://github.com/valora-inc/wallet/pull/4942/files#diff-85d748cd60f247c80e4683d73906b699e600482591a17a57dc39a382055ee038R9-R10

PR #4950, opened 2024-02-21
fix(imported-tokens): Improving the swap experience for imported tokens

dievazqu:
### Description

* Adding the network icon to the token 
* Showing a specific message when quote is not available (imported tokens might get to this case).
* Adding imported tokens analytics

### Test plan

Unit tests + locally in emulator

### Related issues

- Fixes RET-1008

### Backwards compatibility

Y

### Network scalability

### Screen shot

Before
Simulator Screen Shot - iPhone 14 Pro - 2024-02-16 at 12 53 18

After
Simulator Screen Shot - iPhone 14 Pro - 2024-02-16 at 12 53 51


src/swap/SwapScreen.test.tsx
```diff
@@ -594,9 +595,11 @@ describe('SwapScreen', () => {
         toToken: mockCusdAddress,
         toTokenId: mockCusdTokenId,
         toTokenNetworkId: NetworkId['celo-alfajores'],
+        toTokenIsImported: false,
```
kathaypacific, line 598: maybe we can update one of these tokens so that one of these values is true?

src/swap/SwapScreen.tsx
```diff
@@ -334,7 +334,12 @@ export function SwapScreen({ route }: Props) {
 
   useEffect(() => {
     if (fetchSwapQuoteError) {
-      dispatch(showError(ErrorMessages.FETCH_SWAP_QUOTE_FAILED))
+      Logger.error(TAG, 'Error fetching swap quote', fetchSwapQuoteError.message)
```
kathaypacific, line 337: nit: do we need to send these errors to sentry? we can see them from the gcp logs already. if we want to log them then Logger.info or .warn is good?

src/swap/useSwapQuote.ts
```diff
@@ -182,6 +182,11 @@ function useSwapQuote(networkId: NetworkId, slippagePercentage: string) {
       }
 
       const quote: FetchQuoteResponse = await response.json()
+
+      if (!quote.unvalidatedSwapTransaction) {
+        throw new Error('No quote available')
```
kathaypacific, line 187: nit: can we hard code this as a const and share it in the swap screen so that we don't accidentally update this and break the functionality?

src/tokens/selectors.ts
```diff
@@ -366,6 +368,7 @@ export const swappableToTokensByNetworkIdSelector = createSelector(
     return tokens.filter(
       (tokenInfo) =>
         tokenInfo.isSwappable ||
+        tokenInfo.isManuallyImported ||
```
kathaypacific, line 371: i think this change is also in your other open PR, but do we need this line when it is already in the `swappableFromTokensByNetworkIdSelector`?

kathaypacific (commented):
looking great!!

src/swap/useSwapQuote.ts
dievazqu: Yes, I also thought about creating a new type of Error for this, but it felt like an overkill.
I'll create the constant 👌

src/tokens/selectors.ts
dievazqu: Yep, this code is twice since I need it to test.
I [answered in the other PR](https://github.com/valora-inc/wallet/pull/4930#discussion_r1499496578)

kathaypacific: oooohhhh no sorry you're totally right, i flipped the intention of the /filter yesterday. this is fine 👌🏼

kathaypacific (approved):
✨

PR #4952, opened 2024-02-21
fix: nav help icon

MuckT:
### Description

The `src/icons/navigator/Help.tsx` icon was changed to have props recently; however, icons used within the drawer navigator cannot support props and can cause issues with rendering: alignment and color. Added a new Help icon in `src/icons/Help.tsx` and updated usages.

| iOS Before | iOS After |
| ---- | ---- |
|  |  |

### Test plan

- Tested locally on iOS

### Related issues

N/A

### Backwards compatibility

Yes

### Network scalability

N/A


jh2oman (approved):

src/icons/navigator/Help.tsx
```diff
@@ -1,15 +1,15 @@
 import * as React from 'react'
 import Svg, { Circle, Path } from 'react-native-svg'
 
-export function Help({ size = 32, color = '#B4B9BD' }) {
+export function Help() {
   return (
-    <Svg testID="Help" width={size} height={size} viewBox="0 0 32 32" fill="none">
-      <Circle cx={16} cy={16} r={11} stroke={color} strokeWidth={2} />
+    <Svg testID="Help" width={32} height={32} viewBox="0 0 32 32" fill="none">
```
satish-ravi, line 6: could we just import the other help icon and use it here instead of duplicating the svg?
E.g.,
```
import { Help as HelpIcon } from 'src/icons/Help'

export function Help() {
  return <HelpIcon size={32} color = '#B4B9BD' />
}
```

PR #4953, opened 2024-02-21
chore(tests): Fixing weird test case in SwapScreen

dievazqu:
### Description

The test in SwapScreen `"should show and hide the switched network warning"` doesn't fail when it runs with the full batch of test cases.
But if it runs alone or first, it was failing.

This PR fixes the test case, however it doesn't address the fact that there's some mismanagement of the state between tests.  

### Test plan

Unit tests

### Backwards compatibility

Y

### Network scalability

### DIFF
--- a/test/values.ts
+++ b/test/values.ts
@@ -551,6 +551,7 @@ export const mockTokenBalances: Record<string, StoredTokenBalance> = {
     imageUrl: '',
     balance: '0',
     priceUsd: '1',
+    priceFetchedAt: Date.now(),
   },
 }
 


kathaypacific (commented):
i think the issue comes from [this mock](https://github.com/valora-inc/wallet/blob/021032f57832caf7bfaff8d1e1cad146f1829184/src/swap/SwapScreen.test.tsx#L1045) which is not cleared and persists for all the tests that come after. 

the wallet requires both both the priceUsd and a recent fetched timestamp to show the fiat price, otherwise it will show no usd price warning and the confirmation overlay. the problem is we haven't replicated it well in our [test conditions](https://github.com/valora-inc/wallet/blob/021032f57832caf7bfaff8d1e1cad146f1829184/src/swap/SwapScreen.test.tsx#L243) - here, we should have tested also for the fetched at timestamp, or enable some way to assert that the price warning is expected. 

this test shouldn't work because the price for USDC doesn't have a fetched at timestamp but has a priceUsd. your fix works because you're now using Date.now() to set the last fetched price time for USDC. i think a more holistic way to fix this issue is to:
1. restore the date.now mock in the beforeEach step
2. update selectSingleSwapToken to account for stale token prices

i suppose your solution does a variant of 2. but it doesn't prevent head scratching in the future when other people bump up against this...

kathaypacific:
@dievazqu is it still on your radar to make the changes in my comment above? lmk if you want me to do it instead

PR #4955, opened 2024-02-22
ci(e2e): test changes for xcode 15 support

MuckT:
### Description

Adds some custom logic to set the default currency and language on e2e test runners. This allows consistency regardless of the device e2e tests are run on e.g. personal dev devices, remote ci machines, etc. While the desirable way to achieve this would be setting the simulator location as in 60a90d7bae690f94e131c3a4907b30b2f141ba4a or with Detox's [`device.setLocation(...)`](https://wix.github.io/Detox/docs/next/api/device/#devicesetlocationlat-lon) it appears that when setting a simulator location in Xcode 15.2 this is not respected and the machine's physical location is used instead.

Alternatively we could do something like this https://stackoverflow.com/a/75914924, but that seems more opaque for when e2e tests behavior deviates.

### Test plan

- Tested in CI on existing Android Runners & new Mac Runner.

### Related issues

- Related ACT-1099

### Backwards compatibility

- Yes - only impacts e2e test runs.

### Network scalability

- N/A


.github/actions/yarn-install/action.yml
```diff
@@ -13,7 +13,7 @@ runs:
         path: |
           node_modules
           */*/node_modules
-        key: ${{ runner.os }}-node-${{ steps.setup-node.outputs.node-version }}-${{ hashFiles('**/yarn.lock', 'patches/*.patch') }}
+        key: ${{ runner.os }}-temp-node-${{ steps.setup-node.outputs.node-version }}-${{ hashFiles('**/yarn.lock', 'patches/*.patch') }}
```
MuckT, line 16: Temporary to prevent cache issues with other runners.

.github/workflows/e2e-ios.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       SECRETS_AVAILABLE: ${{ secrets.SECRETS_AVAILABLE }}
       BASH_ENV: ~/.profile
     name: iOS (${{ inputs.ios-version  }})
-    runs-on: ios-e2e-group
+    runs-on: temp-ios-e2e-group
```
MuckT, line 17: Temporary to target iOS runners with Xcode 15.2.

package.json
```diff
@@ -393,6 +393,13 @@
           "os": "iOS 15.0"
         }
       },
+      "simulator.17.0": {
```
MuckT, line 396: Additional e2e test changes will be needed to run iOS 17.0 in CI as the keyboard is displayed on this emulator by default. This will be addressed in a separate PR.

src/navigator/Navigator.tsx
```diff
@@ -574,7 +575,7 @@ export function MainStackScreen() {
 
     const initialRoute: InitialRouteName = getInitialRoute({
       choseToRestoreAccount,
-      language,
+      language: language ?? isE2EEnv ? 'en-US' : language,
```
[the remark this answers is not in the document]
MuckT: There should be no change for non-e2e behavior. This was needed as the language screen was being served on some tests specifically the new account test in CI.

package.json
MuckT: Removed in 4820351b8c175323807ad646f5206f22ecf06917!

satish-ravi (approved):

.github/actions/yarn-install/action.yml
MuckT: removed in b7efe47eb9a524328190d58a9eb90bd53c67ad03.

.github/workflows/e2e-ios.yml
MuckT: Removed in b7efe47eb9a524328190d58a9eb90bd53c67ad03.

PR #4960, opened 2024-02-22
feat(CAB): Log hashed keyshares

finnian0826:
### Description

Log hashed keyshares for CAB so can help a user if they are failing to restore.

Track analytics event during setup, log in app during restore.

### Related issues

- Fixes #[ACT-1097]

### Backwards compatibility

Yes

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -139,6 +139,8 @@ export const AUTH0_CLIENT_ID =
   // dev app client id as fallback for e2e tests
   'YgsHPq93Egfap5Wc4iEQlGyQMqjLeBf2'
 
+export const PEPPER = keyOrUndefined(secretsFile, DEFAULT_TESTNET, 'PEPPER')
+
 export const AUTH0_DOMAIN = configOrThrow('AUTH0_DOMAIN')
 
 export const SPEND_MERCHANT_LINKS: SpendMerchant[] = [
--- a/src/keylessBackup/saga.ts
+++ b/src/keylessBackup/saga.ts
@@ -3,6 +3,7 @@ import { initializeAccountSaga } from 'src/account/saga'
 import { KeylessBackupEvents } from 'src/analytics/Events'
 import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
 import { generateKeysFromMnemonic, getStoredMnemonic, storeMnemonic } from 'src/backup/utils'
+import { PEPPER } from 'src/config'
 import { walletHasBalance } from 'src/import/saga'
 import {
   decryptPassphrase,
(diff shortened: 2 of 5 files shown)


silasbw (approved):
A'd but someone more versed than me should take a look (@jh2oman ?)

src/config.ts
```diff
@@ -139,6 +139,12 @@ export const AUTH0_CLIENT_ID =
   // dev app client id as fallback for e2e tests
   'YgsHPq93Egfap5Wc4iEQlGyQMqjLeBf2'
 
+export const HASHED_KEYSHARE_PEPPER = keyOrUndefined(
```
jophish, line 142: Spoke with Satish a bit about this (using what is essentially public information as a pepper)... I'm actually not sure this technically counts as _either_ a pepper _or_ a salt in this case, since the only criterion for a pepper is:
* they should be a secret
while the only criterion for a salt is:
* they should be unique among all salts used in the system

This doesn't meet either of those (if we consider it public info), since its lack of secrecy disqualifies it from being a pepper, and its lack of uniqueness disqualifies it from being a salt. I'm not really sure if this matters, but I'm also not enough of an expert in security to know whether or not prefixing/suffixing all secrets with a fixed, public string is in anyway a detriment to the security of the system.

It's prolly fine, but I think I'd err on the side of caution and just drop the idea of using a suffix/prefix at all - it keeps the code simpler, anyhow. May want to check with @nategraf on the security implications of this.

finnian0826: that makes sense, I think I'm a bit confused about how the encrypted secrets.json file works in the wallet then, are those secrets public information if you don't have google access? If so, why are they stored in the secrets file?

jophish: My understanding is that the encryption process is a sort of best-effort at obfuscating the secrets (i.e., not _literally_ checking the secrets in plaintext into version control) but typically I consider anything that's _used_ by a front-end, even if only in-memory (and not checked into/hardcoded into the source), to be public information.

src/keylessBackup/saga.ts
```diff
@@ -43,6 +44,11 @@ export function* handleValoraKeyshareIssued({
 }: ReturnType<typeof valoraKeyshareIssued>) {
   try {
     const torusKeyshare = yield* waitForTorusKeyshare()
+    const hashedKeyshare = calculateSha256Hash(`CAB_KEYSHARE_HASH_${keyshare}`)
+    const hashedTorusKeyshare = calculateSha256Hash(`CAB_KEYSHARE_HASH_${torusKeyshare}`)
```
satish-ravi, line 48: Don't think it makes much difference, but we could use different prefixes for these right?

satish-ravi (approved):

PR #4963, opened 2024-02-23
feat(jumpstart): handle jumpstart state

bakoushin:
### Description

1. Introducing new slice to the redux state for jumpstart link handling:
```
jumpstart: {
  showLoading: boolean
  showError: boolean
}
```
This state is expected to be used for rendering UI components displaying the jumpstart link handling state to the user. The actual components will follow in the subsequent PR once the design discussions converge.

2. After successful jumpstart claim two types of pending transactions are dispatched (getting tx data trough blockchain RPC calls):
* Payment received (for ERC-20 claim)
* NFT received (for ERC-721 claim)

Those transactions are expected to be replaced with blockchain-api data once it's fetched.

3. Added analytics events for jumpstart handler lifecycle:
* claim succeeded
* claim failed
* ERC20 token claimed
* ERC721 nft claimed

### Test plan

* Updated unit tests

### Related issues

- Related to RET-1004

### Backwards compatibility

Y

### Network scalability


kathaypacific (commented):
looking great!! perhaps you want to add the jumpstart state to the redux blacklist or restore the initial state via the rehydrate action, but it seems like we wouldn't want to persist this store across app sessions?

src/analytics/Properties.tsx
```diff
@@ -1485,6 +1486,12 @@ interface TransactionDetailsProperties {
   }
 }
 
+interface WalletJumpstartProperties {
+  [JumpstartEvents.jumpstart_started]: undefined
+  [JumpstartEvents.jumpstart_succeeded]: undefined
+  [JumpstartEvents.jumpstart_failed]: undefined
```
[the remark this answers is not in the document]
bakoushin: yes, `jumpstart_started` is redundant.

added for token/amount tracking separate events:
* `jumpstart_claimed_token` with ERC20 claim details
* `jumpstart_claimed_nft` with ERC721 claim details 

I think it's worth keeping a bare `jumpstart_succeeded` event as well, because getting claim details may fail (but it doesn't mean claim didn't succeeded)

src/jumpstart/saga.ts
```diff
@@ -0,0 +1,152 @@
...
+    Logger.error(TAG, 'No wallet address found in store. This should never happen.')
+    return
+  }
+
+  try {
+    ValoraAnalytics.track(JumpstartEvents.jumpstart_started)
+    yield* put(jumpstartClaimStarted())
+
+    const transactionHashes = yield* call(jumpstartLinkHandler, privateKey, walletAddress)
+    yield* fork(dispatchPendingTransactions, transactionHashes)
```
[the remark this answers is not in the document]
bakoushin: I agree that we should handle networks in a generic way and avoid network hardcoding. Thanks for pointing this out!

I ended up looping through all remote config networkIds, but I think it's suboptimal. If the amount of jumpstart-supported networks grows, we would have to loop trough them all, because we have no idea for which specific network that link is. To avoid looping, perhaps we can encode network id within the jumpstart link?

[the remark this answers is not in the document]
bakoushin: Also, we probably want to refactor jumpstartLinkHandler from ContractKit to Viem to make it multichain? 

For know, I just made it throw an error if network id isn't Celo.

bakoushin:
> perhaps you want to add the jumpstart state to the redux blacklist or restore the initial state via the rehydrate action, but it seems like we wouldn't want to persist this store across app sessions?

yes, this state shouldn't persist, because it's only for presentation purposes during current user session. added it to persist's `blacklist`. thanks for pointing this out!

src/jumpstart/jumpstartLinkHandler.ts
```diff
@@ -1,25 +1,24 @@
...
-    Logger.error(TAG, 'Contract address is not provided in dynamic config')
-    return
+export async function jumpstartLinkHandler(
+  networkId: NetworkId,
+  contractAddress: string,
+  privateKey: string,
+  userAddress: string
+): Promise<Hash[]> {
+  if (networkId !== networkConfig.defaultNetworkId) {
+    throw new Error(`Unsupported network id: ${networkId}`)
```
kathaypacific, line 21: ah, i was wondering why we need to do this, and then i saw that this function is using contract kit 😞 let's leave it for now and make a task to remove contract kit for later

src/jumpstart/saga.ts
```diff
@@ -0,0 +1,219 @@
...
+          contractAddress,
+          privateKey,
+          walletAddress
+        )
+
+        successfulClaimsCount += transactionHashes.length
+
+        yield* fork(dispatchPendingTransactions, networkId, transactionHashes)
+      } catch (error) {
+        Logger.error(TAG, `Error handling jumpstart link for ${networkId}`, error)
```
kathaypacific, line 68: this claim saga is kicked off every time a jumpstart deep link is consumed, but presumably each deep link is only for one network. if we had multichain jumpstart, this logger.error would always be fired because some chains will always have no transactions found right? and we'd always be doing unnecessary work, to try to claim across chains when we should only do it for one chain? perhaps what would be more efficient is to have the jumpstart link contain the network id...what do you think? sorry, i know this is different to what i suggested last time

kathaypacific (commented):
looking great!! sorry i had a different thought while reviewing how we make the claim here, lmk what you think

src/jumpstart/jumpstartLinkHandler.ts
bakoushin: ok, here is the task: https://linear.app/valora/issue/RET-1019/[wallet]-refactor-jumstartlinkhandler-to-viem

src/jumpstart/saga.ts
bakoushin: yes, I agree that we have to store the chain id somewhere in the jumpstart link to avoid unnecessary looping over all supported networks.
should we have a separate task for that?

in the meanwhile we can assume that we get network id from the link, right?

kathaypacific: yes let's do that! since we create the link when the user makes the transaction, we'll already have this information and no reason why we can't add it to the link to make our lives easier

kathaypacific: let's make the network id `pathParts[3]` of the jumpstart link

src/jumpstart/slice.ts
```diff
@@ -0,0 +1,55 @@
+import { createSlice } from '@reduxjs/toolkit'
+
+export interface State {
+  showLoading: boolean
+  showError: boolean
```
jeanregisser, line 5: Just checking, do we need both loading and error state at the same time?
If not, I'd advise to model it differently to avoid impossible states.

For instance: https://github.com/valora-inc/wallet/blob/2b8972311cd1da141b1bb09f5992bb99c783648e/src/positions/slice.ts#L6

bakoushin: good point, thanks! refactored it to have just one consistent property:
```
export interface State {
  claimStatus: 'idle' | 'loading' | 'error'
}
```

src/jumpstart/jumpstartLinkHandler.test.ts
```diff
@@ -43,27 +39,23 @@ describe('jumpstartLinkHandler', () => {
   it('calls executeClaims with correct parameters', async () => {
     ;(fetchWithTimeout as jest.Mock).mockImplementation(() => ({
       ok: true,
+      json: async () => ({ transactionHash: '0xTEST' }),
```
kathaypacific, line 42: nit: maybe this can be different from the `contractAddress`?

src/app/saga.ts
```diff
@@ -298,12 +299,6 @@ export function* handleDeepLink(action: OpenDeepLink) {
   const { isSecureOrigin } = action
   Logger.debug(TAG, 'Handling deep link', deepLink)
 
-  const walletAddress = yield* select(walletAddressSelector)
```
kathaypacific, line 301: nit: i wonder if we should leave this here and continue to expect the wallet address as a param in the jumpstartClaim saga. only the jumpstartClaim saga is technically needing the wallet address right now, so i understand why you made the change. 

it's been annoying in the past for deep links to be processed when wallet address is not available. for example, the opened_via_invite_url event ended up being unhelpful because the walletAddress was undefined and we couldn't match inviters and invitees. the logic in the wallet has changed now where this saga shouldn't be triggered unless the walletAddress is available, but that is done in a hook and perhaps nice to fail early here if something goes wrong?

src/jumpstart/saga.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    for (const {
+      address,
+      args: { token: tokenAddress, amount },
+    } of parsedLogs) {
+      const tokenId = getTokenId(networkId, tokenAddress)
+
+      const token = tokensById[tokenId]
+      if (!token) {
+        Logger.error(TAG, 'Claimed unknown tokenId', tokenId)
```
kathaypacific, line 103:
```suggestion
        Logger.warn(TAG, 'Claimed unknown tokenId', tokenId)
```
seems that we wouldn't want to investigate this right?

src/jumpstart/saga.ts
```diff
@@ -0,0 +1,199 @@
...
+          })
+        )
+
+        ValoraAnalytics.track(JumpstartEvents.jumpstart_claimed_nft, {
+          networkId,
+          contractAddress,
+          tokenId: tokenId.toString(),
+        })
+      } catch (error) {
+        Logger.error(TAG, 'Error adding pending NFT transaction', error)
```
kathaypacific, line 195: same here, i wonder if we should downgrade these to .warn since the pending transactions is a nice to have visual feedback rather than core functionality?

kathaypacific (approved):
looks great!! a few non blocking annoying comments, do with them what you want :)

src/app/saga.ts
bakoushin: but failing early will mean we don't show the user any feedback about the error, right?

we could add a generic alert, but failing inside the specific handler means we could communicate what specific situation we are processing, and what error we've encountered -- depending on the handler features

kathaypacific: hmm yes, i guess i can see it both ways. 

failing early here = we assert all deep links should be handled after the wallet is initialized and it does not matter whether the deep link strictly needs the wallet address in the logic. even if the logic succeeds, our instrumentation will be broken becuase the wallet address is missing.

failing contextually = we have more information about where it failed, but we will allow silent instrumentation failures. 

i have a preference for the first scenario, because i cannot think of a scenario where we'd want to process a deep link when a wallet is not initialised?

bakoushin: I agree. There may be a logic requiring a wallet address not _directly_ in a deep link handler, down the user journey, like when link handler fires a navigate event.

PR #4965, opened 2024-02-23
feat: add jumpstart button to select recipients screen

kathaypacific:
### Description

This PR adds the remote config and button for the start of the escrow jumpstart flow. I think that the most straightforward way to implement the escrow flow is to use the existing send flow rather than separate the flows. I've made a POC that is working e2e in #4964, in case it helps to understand this decision.

### Test plan

Simulator Screenshot - iPhone 14 Pro - 2024-02-23 at 12 22 01

### Related issues

- Related to RET-993

### Backwards compatibility

Y

### Network scalability


src/send/SendEnterAmount.tsx
```diff
@@ -158,6 +159,13 @@ function SendEnterAmount({ route }: Props) {
       // should never happen because button is disabled if send is not possible
       throw new Error('Send is not possible')
     }
+
+    if (origin === SendOrigin.Jumpstart) {
+      // TODO handle send transaction and navigation
+      return
+    }
+
+    const recipient = route.params.recipient
```
bakoushin, line 168: what's the reason behind moving the `recepeint` declaration here?

bakoushin: ah, type guard 😅

bakoushin (approved):
🪄

src/send/SendEnterAmount.tsx
kathaypacific: yes 🙈 typescript doesn't let me destructure this from the route.params when it may not exist

PR #4966, opened 2024-02-23
chore: update wallet jumpstart abi format

kathaypacific:
### Description

As the title. Using a .ts file allows static type checking with typescript when using the abi with viem. This will be useful for upcoming PRs relating to escrow (see #4964)

### Test plan

n/a

### Related issues

- Related to RET-993

### Backwards compatibility

Y

### Network scalability

Y


src/abis/IWalletJumpstart.ts
```diff
@@ -0,0 +1,310 @@
+const jumpstart = {
+  abi: [
```
kathaypacific, line 2: this is exactly the same file contents as WalletJumpStart.json, except wrapped into an object with an `abi` property

bakoushin (approved):
I think I've encountered a similar problem in #4963 and ended up with ad hoc inlining function/event signature using [parseAbi()](https://viem.sh/docs/abi/parseAbi) because that way I can see the expected function signature right next to the argument list and have not to deal with unreadable JSON blob.

Example:
```
const tokenUri = (yield* call([publicClient[network], 'readContract'], {
  address: contractAddress,
  abi: parseAbi(['function tokenURI(uint256 tokenId) returns (string memory)']),
  functionName: 'tokenURI',
  args: [tokenId],
})) as string
```

kathaypacific:
> I think I've encountered a similar problem in #4963 and ended up with ad hoc inlining function/event signature using [parseAbi()](https://viem.sh/docs/abi/parseAbi) because that way I can see the expected function signature right next to the argument list and have not to deal with unreadable JSON blob.
> 
> Example:
> 
> ```
> const tokenUri = (yield* call([publicClient[network], 'readContract'], {
>   address: contractAddress,
>   abi: parseAbi(['function tokenURI(uint256 tokenId) returns (string memory)']),
>   functionName: 'tokenURI',
>   args: [tokenId],
> })) as string
> ```

oh this is good timing then!! :D yes i noticed that i was passing in incorrect types to viem and not getting static warnings about it. i'll merge this now, hope it helps to clean up your PR too

PR #4967, opened 2024-02-23
feat: prepare jumpstart transaction in send amount screen

kathaypacific:
### Description

This PR updates the send amount screen to prepare the jumpstart transactions. 
- if launching from the jumpstart flow, we should restrict the selectable tokens to only those working with jumpstart (i.e. check which networks we have a jumpstart contract for, and restrict tokens to ERC-20 tokens with balance on those networks). add a selector for this.
- the send amount screen creates the public and private key pair used for the jumpstart link. in this way, this key pair only needs to be generated once per flow
- the fee estimation should work and the transactions should be prepared

Not included in this PR:
- anything related to sending the prepared transactions or navigating to the next screen

### Test plan

https://github.com/valora-inc/wallet/assets/20150449/0608dda8-f422-469f-bd20-791ac63654cd

### Related issues

- Fixes RET-993

### Backwards compatibility

Y

### Network scalability

Y


src/send/usePrepareSendTransactions.ts
```diff
@@ -1,35 +1,138 @@
...
+        args: [jumpstartContractAddress as Address, spendAmount],
+      }),
+    }
+    baseTransactions.push(approveTx)
+  }
+
+  const transferTx: TransactionRequest = {
+    from: walletAddress as Address,
+    to: jumpstartContractAddress as Address,
+    value: BigInt(0),
```
bakoushin, line 62: i wonder why we are specifying `value` for this tx, but not doing the same for th e `approveTx`?

bakoushin:
Looking great!

Just wanted to discuss another possible approach: what if instead of branching in `SendEnterAmount` and the helper hook we extract the logic specific to send and jumpstart into dedicated components and hooks respectively? 

For instance, we could extract the shared parts of `SendEnterAmount` to another component, say `EnterAmount`, and provide it with the moving parts as props. Those moving parts we could define in our old `SendEnterAmount` and a new one like `JumpstartEnterAmount`. That way we could probably avoid much of these `if`s and input overload.

What do you think?

kathaypacific:
> Looking great!
> 
> Just wanted to discuss another possible approach: what if instead of branching in `SendEnterAmount` and the helper hook we extract the logic specific to send and jumpstart into dedicated components and hooks respectively?
> 
> For instance, we could extract the shared parts of `SendEnterAmount` to another component, say `EnterAmount`, and provide it with the moving parts as props. Those moving parts we could define in our old `SendEnterAmount` and a new one like `JumpstartEnterAmount`. That way we could probably avoid much of these `if`s and input overload.
> 
> What do you think?

yes i think that's a reasonable suggestion. there are pros and cons of course, but it seems reasonable to assume that the send flow and jumpstart will use the same amount input UI for the foreseeable future. i think this implementation is not too messy and it is somewhat the easier option to review, but i hear the feedback that it introduces more complexity. i'll have a go at seeing what your suggestion looks like

src/send/SendEnterAmount.tsx
```diff
@@ -100,18 +105,31 @@ function SendEnterAmount({ route }: Props) {
...
+    return origin === SendOrigin.Jumpstart ? supportedJumpstartTokens : supportedSendTokens
+  }, [supportedSendTokens, supportedJumpstartTokens, origin])
   const defaultToken = useMemo(() => {
     const defaultToken = tokens.find((token) => token.tokenId === defaultTokenIdOverride)
     const lastUsedToken = tokens.find((token) => token.tokenId === lastUsedTokenId)
 
     return defaultToken ?? lastUsedToken ?? tokens[0]
   }, [tokens, defaultTokenIdOverride, lastUsedTokenId])
 
+  const jumpstartLink = useMemo(() => {
```
dievazqu, line 124: This got me thinking. We need to create a new private key for each invitation (all the rewards sent with the same private key can be claimed with one link, so we shouldn't create more than one invitation per key).
I'm not sure if with this code it could be reused for different invitations. I think we are fine, because we are creating a new `SendEnterAmount` component in every attempt, right? But just double checking I'm not missing some case.

I would put inside the `onPressReview` the code to get a new private key in each attempt and avoid confusions.

src/tokens/selectors.ts
```diff
@@ -505,3 +506,20 @@ export const importedTokensSelector = createSelector(
...
+      .jumpstartContracts
+  ) as NetworkId[]
+
+export const jumpstartSendTokensSelector = createSelector(
+  [(state) => tokensWithTokenBalanceSelector(state, getJumpstartEnabledNetworkIds())],
+  (tokensWithBalance) => {
+    return tokensWithBalance.filter((token) => {
+      // the jumpstart contract currently requires a token address for the
+      // depositERC20 method
+      return !!token.address
```
dievazqu, line 522: Not related to this PR, but I realized this whole flow won't work for depositing native ETH (it will work for WETH), unless we update the contract to support deposits in the native token.

dievazqu (commented):
Looking good!
Left some minor questions.

src/send/SelectRecipientButtons.test.tsx
```diff
@@ -33,20 +33,28 @@ describe('SelectRecipientButtons', () => {
...
+      showBalances: ['celo-alfajores'],
+      jumpstartContracts: {
+        'celo-alfajores': {
+          contractAddress: '0x123',
+        },
+      },
+    })
+    const { getByText, findByTestId } = renderComponent()
 
+    expect(await findByTestId('SelectRecipient/QR')).toBeTruthy()
```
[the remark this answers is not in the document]
kathaypacific: this fixes an "act" warning - there is an asynchronous action to generate the QR code on component mount, and this expect statement will ensure that the async action finishes running before the subsequent assertions are made. it's not strictly necessary for this test but it is good practice to not be making assertions on components that haven't reached their "steady state". we are checking for the SelectRecipientJumpstartButton existence in the `fireEvent` line below (the test will fail if the `getByText('sendSelectRecipient.jumpstart.title')` inside `fireEvent` cannot be found)

src/send/SendEnterAmount.tsx
kathaypacific: we need the public key to prepare the transactions, which is done before executing `onPressReview` so we cannot generate the private key in that handler. it's safe to generate the key pair on mount of this screen, since the user will leave this screen once the transaction is completed

src/tokens/selectors.ts
kathaypacific: yes that is the whole reason for this selector :) otherwise i think we could reuse the send tokens selector

src/tokens/selectors.ts
```diff
@@ -505,3 +506,20 @@ export const importedTokensSelector = createSelector(
...
   }
 )
+
+const getJumpstartEnabledNetworkIds = () =>
+  Object.keys(
+    getDynamicConfigParams(DynamicConfigs[StatsigDynamicConfigs.WALLET_JUMPSTART_CONFIG])
+      .jumpstartContracts
+  ) as NetworkId[]
+
+export const jumpstartSendTokensSelector = createSelector(
```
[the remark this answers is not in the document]
kathaypacific: yes that's fair, we can sort in the same way as the send tokens

src/send/SelectRecipientButtons.test.tsx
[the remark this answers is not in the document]
dievazqu: Yes, I guess this is fine.
But it seems a bit odd to add an unrelated expect to this test.
Let's say tomorrow we change how the QR works, then probably we will need to tweak this test, which doesn't sound nice.

src/send/SendEnterAmount.tsx
dievazqu: Oh, you are right, we need to prepare the transaction.
I think this is fine, since we don't allow the user to go back to this screen after they execute the deposit.

PR #4969, opened 2024-02-23
feat(keylessBackup): add delete happy path

jh2oman:
### Description

Adding in the happy path for deleting your cloud backup, assuming nothing goes wrong. https://linear.app/valora/issue/ACT-766/delete-backup-button-behavior-happy-path

[Screen_recording_20240227_114012.webm](https://github.com/valora-inc/wallet/assets/8432644/f2d3c8d7-e345-4477-8c24-948c9041c519)

### Test plan

Unit tests updated. large manual testing regimine coming soon

### Related issues

https://linear.app/valora/issue/ACT-766/delete-backup-button-behavior-happy-path

### Network scalability

n/a


src/viem/getLockableWallet.ts
```diff
@@ -69,7 +69,7 @@ export default function getLockableViemWallet(
     chain,
     transport: getTransport(chain),
     account,
-  }).extend((client: Client): Actions => {
+  }).extend((client: Client<Transport, Chain, Account>): Actions => {
```
jh2oman, line 72: with our setup we always have a Transport, Chain, and Account. This added specificity makes TS play nice when calling `signMessage` b/c it now assumes (correctly) that there is a valid account attached to the wallet

src/account/Settings.tsx
```diff
@@ -568,6 +594,11 @@ const styles = StyleSheet.create({
   bottomSheetButton: {
     marginTop: Spacing.Regular16,
   },
+  value: {
+    ...fontStyles.regular,
```
MuckT, line 598: typeScale instead of fontStyles.

jh2oman: I wouldn't want to change just this one without updating all of the settings menu, which is all still on `fontStyles`. Also the figma designs themselves haven't been updated to the new typeScale so i think it is best to wait

src/keylessBackup/index.ts
```diff
@@ -81,3 +82,35 @@ export async function getEncryptedMnemonic({
...
 }
+
+export async function deleteEncryptedMnemonic(wallet: ViemWallet) {
+  if (!wallet.account) {
+    // This should never happen
+    throw new Error('no account found in the wallet')
+  }
+  const siweClient = new SiweClient(
+    {
+      accountAddress: wallet.account.address,
```
satish-ravi, line 93: shouldn't this use the encryption address and not the wallet address? (same as what is done in GET). The cab backend has no knowledge of the wallet address

src/account/Settings.tsx
```diff
@@ -411,12 +416,33 @@ export const Account = ({ navigation, route }: Props) => {
...
+                    {deleteKeylessBackupStatus === KeylessBackupDeleteStatus.InProgress
+                      ? t('pleaseWait')
+                      : cloudBackupCompleted
+                        ? t('delete')
+                        : t('setup')}
+                  </Text>
+                  {!cloudBackupCompleted &&
+                    deleteKeylessBackupStatus !== KeylessBackupDeleteStatus.InProgress && (
+                      <ForwardChevron />
+                    )}
```
satish-ravi, line 443: this is quite hard to read with a nested ternaries. maybe we can render in progress and setup / delete as two subcomponents?

satish-ravi (changes requested):

src/account/Settings.tsx
```diff
@@ -411,12 +416,33 @@ export const Account = ({ navigation, route }: Props) => {
...
+                    : onPressSetUpKeylessBackup
               }
               testID="KeylessBackup"
-              ctaText={cloudBackupCompleted ? t('delete') : t('setup')}
-              ctaColor={cloudBackupCompleted ? colors.error : colors.primary}
-              showChevron={!cloudBackupCompleted}
+              cta={
+                <>
+                  {deleteKeylessBackupStatus === KeylessBackupDeleteStatus.InProgress && (
+                    <LoadingSpinner width={32} />
```
MuckT, line 431: In the loading state the setting item resizes a bit not sure if this is expected.

src/keylessBackup/index.ts
MuckT: I am unable to delete my cloud backup despite a successful restore and receive a 404 response.

src/account/Settings.tsx
jophish: +1 to this, seems like we could consolidate some of the  `deleteKeylessBackupStatus === KeylessBackupDeleteStatus.InProgress` checks and clean this up a bit?

src/keylessBackup/index.ts
jophish: I'm also unable to delete, getting a 404 not found as well. @satish-ravi What's required to get the encryption address? Looking at the existing code, it appears that we construct the encryption address by getting both keyshares (torus and valora), constructing the encryption private key, and converting that into the public key (address). We'd also need to sign the SIWE message using the encryption private key, so I suppose getting the public key sets us up for that as well. 

I guess it seems to me like this may change the UX somewhat, since the user will actually need to retrieve both keyshares in order to delete, right?

jophish: Actually, after looking at the ticket (and the code), it appears that the recommendation is to just store their encryption private key in the keychain and lock it behind PIN entry?

jh2oman: yes, I believe that is the recommendation @jophish

src/keylessBackup/encryption.ts
```diff
@@ -51,15 +51,15 @@ export function deriveKeyFromKeyShares(
 export function getSecp256K1KeyPair(
   keyshare1: Buffer,
   keyshare2: Buffer
-): { privateKey: Uint8Array; publicKey: Uint8Array } {
+): { privateKey: Hex; publicKey: Uint8Array } {
   const derivedKey = deriveKeyFromKeyShares(keyshare1, keyshare2, 48) // 40 is the minimum for hashToPrivateKey
   const privateKey = secp.utils.hashToPrivateKey(derivedKey)
   const publicKey = secp.getPublicKey(privateKey, true)
-  return { privateKey, publicKey }
+  return { privateKey: fromBytes(privateKey, 'hex'), publicKey }
```
jh2oman, line 58: we use hex everywhere throughout the code, cleaner to just convert here once

satish-ravi: could we return both in hex to be consistent?

src/keylessBackup/keychain.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+async function encryptPrivateKey(privateKey: string, password: string) {
+  return CryptoJS.AES.encrypt(privateKey, password).toString()
+}
+
+async function decryptPrivateKey(encryptedPrivateKey: string, password: string) {
+  const bytes = CryptoJS.AES.decrypt(encryptedPrivateKey, password)
+  return bytes.toString(CryptoJS.enc.Utf8)
+}
```
satish-ravi, line 24: these seem duplicated from KeychainLock, can we reuse?

src/keylessBackup/saga.test.ts
```diff
@@ -269,9 +305,13 @@ describe('keylessBackup saga', () => {
             [call(generateKeysFromMnemonic, mockMnemonic), { privateKey: mockPrivateKey }],
             [call(walletHasBalance, privateKeyToAddress(mockPrivateKey)), true],
             [call(assignAccountFromPrivateKey, mockPrivateKey, mockMnemonic), mockWalletAddress],
+            [call(initializeAccountSaga), undefined],
+            [call(storeMnemonic, mockMnemonic, mockWalletAddress), undefined],
+            [
+              call(storeSECP256k1PrivateKey, mockEncryptionPrivateKeyHex, mockWalletAddress),
+              undefined,
+            ],
```
satish-ravi, line 313: We should leave the `.call` if we want to use these functions are called. These don't assert that these functions were called, just provides mocks. If for example initializeAccountSaga is removed, this test will still pass.

src/keylessBackup/encryption.ts
jh2oman: sure, that makes sense

satish-ravi (approved):

PR #4973, opened 2024-02-26
fix(build): trailing comma preventing parsing with jq

jeanregisser:
### Description

This fixes the following [error happening during the iOS build](https://github.com/valora-inc/release-automation/actions/runs/8045843098/job/21971906062):
```
[08:38:58]: ▸ + /opt/homebrew/bin/jq -r '.mainnet|to_entries|map("\(.key)=\(.value|tostring)")|.[]' /Users/runner/work/release-automation/release-automation/wallet/ios/../secrets.json
[08:38:58]: ▸ jq: parse error: Expected another key-value pair at line 18, column 3
```

The problem was introduced in https://github.com/valora-inc/wallet/pull/4942

### Test plan

iOS build works again.

### Related issues

N/A

### Backwards compatibility

Yes

### Network scalability


kathaypacific (approved):

bakoushin (approved):
🙏 Thanks for fixing this!

PR #4977, opened 2024-02-26
refactor: abstract UI parts of the send amount screen

kathaypacific:
### Description

This PR decouples the send amount screen logic and UI components, so that the UI components can be shared with the jumpstart flow. This PR should have no changes for the user.

The EnterAmount component looks completely new but there's no logic changes compared to the existing implementation of SendEnterAmount - only some of the functions are now props.

In later PRs, I will add the jumpstart flow on top of this which will use the EnterAmount component.

### Test plan

CI

### Related issues

- Related to RET-993

### Backwards compatibility

Y

### Network scalability

Y


src/send/SendEnterAmount.test.tsx
```diff
@@ -357,17 +357,17 @@ describe('SendEnterAmount', () => {
   })
 
   it('shows warning when prepareTransactionResult type is not-enough-balance-for-gas', () => {
+    mockUsePrepareSendTransactionsOutput.prepareTransactionsResult = {
```
kathaypacific, line 360: these tests were failing after this refactor without this change. i'm not sure why, but it seems the right thing to do to set the mocks first before rendering the component. after the `render`, if we were to simulate the hook's output value changing, we'd also need to get the component to "listen" and react to it - i don't see the value of testing that part. having said that, i'm not sure why these tests were passing like this before...

src/send/EnterAmount.test.tsx
```diff
@@ -5,214 +5,163 @@ import { getNumberFormatSettings } from 'react-native-localize'
...
 import { SendEvents, TokenBottomSheetEvents } from 'src/analytics/Events'
 import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
-import { SendOrigin } from 'src/analytics/types'
-import { navigate } from 'src/navigator/NavigationService'
-import { Screens } from 'src/navigator/Screens'
-import { RecipientType } from 'src/recipients/recipient'
-import SendEnterAmount from 'src/send/SendEnterAmount'
-import { usePrepareSendTransactions } from 'src/send/usePrepareSendTransactions'
-import { getSupportedNetworkIdsForSend } from 'src/tokens/utils'
+import EnterAmount from 'src/send/EnterAmount'
```
kathaypacific, line 8: to minimise the diffs, i've removed the SendEnterAmount tests for now and done a rename here. these tests are mostly the same, but the EnterAmount component takes different props so most of the changes are related to those. the substance of the tests should be unchanged. i'll add a SendEnterAmount test file separately

src/send/EnterAmount.tsx
```diff
@@ -0,0 +1,481 @@
+import { parseInputAmount } from '@celo/utils/lib/parsing'
```
kathaypacific, line 1: this file is a little hard to review - it is basically a copy + paste of SendEnterAmount, with the send specific logic removed and expected as props.

src/send/SendEnterAmount.tsx
```diff
@@ -210,352 +93,39 @@ function SendEnterAmount({ route }: Props) {
...
-  const feeCurrencies = useSelector((state) => feeCurrenciesSelector(state, token.networkId))
-
-  useEffect(() => {
-    if (origin === SendOrigin.Jumpstart) {
-      // TODO: remove this block and handle preparing jumpstart transactions
-      return
-    }
 
-    const recipient = route.params.recipient
+  const handlRefreshPreparedTransactions = (
```
bakoushin, line 97: typo: handle

bakoushin (approved):
🪆Looking good!

PR #4979, opened 2024-02-26
refactor(contacts): Use react-native-permissions for contacts

finnian0826:
### Description

Replace custom contact permissions checking with react-native-permissions

### Test plan

Unit tests updated

Ran the wallet locally and logged the `contactPermissionStatus` for each place it was added. Started with contacts permissions turned off and verified that it logged as something other than `granted`, then turned permissions to allow and verified that it logged as `granted` in each place.

### Related issues

- Fixes ACT-1077

### Backwards compatibility

Yes


src/identity/contactMapping.ts
```diff
@@ -60,6 +64,9 @@ import { call, delay, put, race, select, spawn, take } from 'typed-redux-saga'
 const TAG = 'identity/contactMapping'
 export const IMPORT_CONTACTS_TIMEOUT = 1 * 60 * 1000 // 1 minute
 
+const CONTACTS_PERMISSION =
+  Platform.OS === 'ios' ? PERMISSIONS.IOS.CONTACTS : PERMISSIONS.ANDROID.READ_CONTACTS
```
satish-ravi, line 68: This seems duplicated in many files, can we move this to a shared constant?
Given the number of places where we just check permissions === GRANTED, might be worth having a wrapper that returns a boolean, although I believe we can't use that directly in the Send recipient screen as it needs to access the other statuses.

src/identity/contactMapping.ts
```diff
@@ -92,8 +99,8 @@ export function* doImportContactsWrapper() {
 }
 
 function* doImportContacts() {
-  const hasGivenContactPermission: boolean = yield* call(checkContactsPermission)
-  if (!hasGivenContactPermission) {
+  const contactPermissionStatus = yield* call(checkPermission, CONTACTS_PERMISSION)
+  if (!(contactPermissionStatus === PERMISSION_RESULTS.GRANTED)) {
```
satish-ravi, line 103:
```suggestion
  if (contactPermissionStatus !== PERMISSION_RESULTS.GRANTED) {
```

src/utils/contacts.ts
```diff
@@ -38,8 +45,8 @@ function customGetAll(callback: (error: any, contacts: MinimalContact[]) => void
 }
 
 export async function getAllContacts(): Promise<MinimalContact[] | null> {
-  const contactPermissionsGiven = await checkContactsPermission()
-  if (!contactPermissionsGiven) {
+  const contactPermissionStatus = await checkPermission(CONTACTS_PERMISSION)
+  if (!(contactPermissionStatus === PERMISSION_RESULTS.GRANTED)) {
```
satish-ravi, line 49:
```suggestion
  if (contactPermissionStatus !== PERMISSION_RESULTS.GRANTED) {
```

src/utils/contacts.ts
```diff
@@ -1,10 +1,22 @@
...
+} from 'react-native-permissions'
 import Logger from 'src/utils/Logger'
-import { checkContactsPermission } from 'src/utils/permissions'
 
 const TAG = 'utils/contacts'
 
+export const CONTACTS_PERMISSION =
+  Platform.OS === 'ios' ? PERMISSIONS.IOS.CONTACTS : PERMISSIONS.ANDROID.READ_CONTACTS
+
+export async function checkContactPermissionStatusGranted() {
```
satish-ravi, line 15: nit: this name seems a bit verbose. Maybe something like `hasGrantedContactsPermission`?

PR #4980, opened 2024-02-27
refactor(send): prepare a transaction in send confirmation screen

satish-ravi:
### Description

The prepared tx passed to Send Confirmation cannot be directly executed for tokens that supports comments because it uses a dummy comment to prepare the tx. This PR updates the send confirmation screen to prepare a transaction instead of accepting it as a prop for all tokens. This turned out to be simpler than conditionally preparing a tx (for tokens that support comments) and also avoids preparing a tx for payment requests before navigating to the confirmation screen / modal. 

### Test plan

Manual, CI

https://github.com/valora-inc/wallet/assets/5062591/5f09bccf-b0c5-4202-a5e7-ecaba601bd7e

### Related issues

- Part of ACT-1049

### Network scalability


MuckT (approved):
This looks great! Tested locally sending cUSD via the typical send flow and with deeplinks.

PR #4984, opened 2024-02-27
chore(logs): Removing noisy error log from transaction watcher

dievazqu:
### Description

The error log was triggered in some normal use cases, generating a lot of noise.
With the new condition, if the error log is present it really means there's some unexpected behavior in the app.

- Fixes #[issue number here]

### Backwards compatibility

Y

### Network scalability

Y


jeanregisser (approved):
👍

PR #5001, opened 2024-02-28
refactor(send): execute prepared transaction in saga

satish-ravi:
### Description

Execute prepared tx directly in the send saga instead of using the viem/saga's send payment method, which was simulating a contract call and then executing it. 

### Test plan

Manual, CI

### Related issues

- Fixes ACT-1049

### Backwards compatibility

Yes

### Network scalability


src/viem/saga.ts
satish-ravi, line 1: This was using the [wrapSendTransactionWithRetry](https://github.com/valora-inc/wallet/blob/3c9784523b587725557ed9d498d1165ee0dc1211/src/transactions/send.ts#L269) method, which attempts to retry after a timeout of 90s, but the retry count was set to only 1, all of this is removed in the send saga, which now matches with what the swap saga is doing currently.
Also, this was firing events at each stage of the transaction, but the send saga already covers most of it, we only lose out on some intermediate events,`transaction_hash_received` and `transaction_receipt_received`, which seem more like debugging events.
I don't think either of these are required anymore, but can add these back if there's a strong reason

src/send/saga.ts
```diff
@@ -212,48 +269,11 @@ function* sendPayment(
...
-        preparedTransaction
-      )
-      if (tokenInfo?.symbol === 'CELO') {
-        ValoraAnalytics.track(CeloExchangeEvents.celo_withdraw_completed, {
-          amount: amount.toString(),
-        })
-      }
-    } else {
-      throw new Error('No address found on recipient')
+    if (tokenInfo?.symbol === 'CELO') {
```
jophish, line 273: nit: Maybe worth having a test for this event?

src/viem/saga.ts
jophish: Is there harm in including these in the updated saga? Seems like there are clear places where they would fit? 

If these are truly never used by anyone though, I think it's fine to get rid of them. Are there any other actual uses of these events in code though? If we remove them here, can we clean them up from the app entirely?

src/send/saga.ts
```diff
@@ -163,44 +166,98 @@ export function* buildAndSendPayment(
...
+    yield* call(
+      handleTransactionReceiptReceived,
+      context.id,
+      receipt,
+      networkConfig.networkToNetworkId[network],
+      feeCurrencyId
+    )
+
+    if (receipt.status === 'reverted') {
+      throw new Error('transaction reverted')
```
jeanregisser, line 257: nit: would be nice to include the txHash in the error message for debugging purposes.

src/send/saga.ts
```diff
@@ -264,9 +284,17 @@ export function* sendPaymentSaga({
 
     yield* put(sendPaymentSuccess({ amount, tokenId }))
     SentryTransactionHub.finishTransaction(SentryTransaction.send_payment)
-  } catch (e) {
-    yield* put(showErrorOrFallback(e, ErrorMessages.SEND_PAYMENT_FAILED))
+  } catch (err) {
+    yield* put(showErrorOrFallback(err, ErrorMessages.SEND_PAYMENT_FAILED))
     yield* put(sendPaymentFailure())
```
jeanregisser, line 289: Looks like we wouldn't want the error message if the pin was cancelled.

And maybe not the `sendPaymentFailure()`? Unless it's used for switching the button state back to enabled.

src/send/saga.ts
```diff
@@ -264,9 +284,17 @@ export function* sendPaymentSaga({
...
+    const error = ensureError(err)
+    if (error.message === ErrorMessages.PIN_INPUT_CANCELED) {
+      Logger.info(`${TAG}/sendPaymentSaga`, 'Send cancelled by user')
+      return
+    }
+    Logger.error(`${TAG}/sendPaymentSaga`, 'Send payment failed', error)
+    ValoraAnalytics.track(SendEvents.send_tx_error, { error: error.message })
+    SentryTransactionHub.finishTransaction(SentryTransaction.send_payment)
```
jeanregisser, line 297: Should the sentry transaction be in a `finally` block?

jeanregisser (commented):
Great!! 

Looking forward to when we can combine the logic for send/swap/wc too, but that will be more work :D

src/viem/saga.ts
satish-ravi: It would be weird just to include those tx events and don't fire other tx events. These aren't fired for swaps, so don't think we need to fire them for sends.
These are still used in the old contract-kit sends, we can clean them up once we remove it.

src/send/saga.ts
satish-ravi: The showErrorOrFallback will show a pin required error message, and `sendPaymentFailure` just resets the `isSending` state so should be fine to include.
I was confused by this as well originally 😄. left a couple of comments

src/send/saga.ts
satish-ravi: This was my original plan, but if the pin is cancelled, we don't start the transaction, so I included it here. But looking at finishTransaction, it looks like it does nothing if a tx with given name isn't found, so should be safe and clean to do it in finally.

src/send/saga.ts
jeanregisser: ok, I guess what I'm suggesting is we change the UX to not display an error message if the pin was cancelled, and not proceed. It's slightly annoying to see the error in that case, since the user decided to cancel.
We did that for swap. https://github.com/valora-inc/wallet/blob/fb39d9985490b4cb7cfa09f559710d193539db06/src/swap/saga.ts#L378-L383

jeanregisser (approved):

src/send/saga.ts
satish-ravi: makes sense, I think there are a few differences with how send and swap UX works, like for swap we navigate immediately after submitting the tx to the chain but we block on the confirmation screen for send. Would be good to address all that in a separate PR with all the UX changes.

jophish (approved):

PR #5004, opened 2024-02-29
ci(renovate): add viem to dependency ignore list

MuckT:
### Description

Prevents [viem](https://github.com/wevm/viem) from being targeted by renovate in update PRs. Allowing renovate to focus on other dependencies.

### Test plan

N/A

### Related issues

N/A

### Backwards compatibility

N/A

### Network scalability

N/A


MuckT:
@Jean there's an activation issue for the update, ACT-1111,  it looks like fee's types move from `BigInt | null` to `BigInt` in the latest version which appears to be a pretty involved upgrade.

jeanregisser:
Could we just close those PR when they happen?
I'd prefer not to add more friction with this, even if we have a follow up task.

satish-ravi:
+1 to @jeanregisser's comment, unless its one of the immortal renovate updates, we can just close PRs. If its a noisy dependency that's taking up most of renovate's time, we should just consider using schedules (like do it only once a week).

MuckT:
Closing in favor of closing renovate PR's until ACT-1111 is completed.

PR #5009, opened 2024-02-29
feat: add blocking max amount warning for jumpstart send flow

kathaypacific:
### Description

As the title. This warning could be shown in combination with any other warnings about the prepared transactions that are there in the EnterAmount component, which I think is acceptable. 

### Test plan

With the max value set at $1USD:

https://github.com/valora-inc/wallet/assets/20150449/83634079-1f12-4e7b-8fc4-3770ecf13da7

### Related issues

- Fixes RET-993

### Backwards compatibility

Y

### Network scalability


bakoushin (approved):
🚀 Awesome!

A small thought on the limit amount representation.
The amount in EUR itched my eye because it is not round. Why this weird limit? Maybe we can explicitly show the USD amount? Not sure though, because then the message is more cluttered.

> no more than 0.9 EUR (1 USD).

dievazqu (approved):

kathaypacific:
> 🚀 Awesome!
> 
> A small thought on the limit amount representation. The amount in EUR itched my eye because it is not round. Why this weird limit? Maybe we can explicitly show the USD amount? Not sure though, because then the message is more cluttered.
> 
> > no more than 0.9 EUR (1 USD).

i think that's a fair point, although if i understand correctly, the feedback seems more related to this arbitrary send limit that blocks a user from progressing. i can imagine an un-round figure like 92 EUR may be just as surprising as a 100 USD because it is an "arbitrary" limit?

my main concern with adding the USD value here is that it does not add value for the user - if someone is used to thinking about their portfolio in a local currency, this USD value doesn't add extra information about next steps, except let them understand that our source of truth is a nice round USD value. i'm not sure that maintaining extra logic for separate translations (i.e. one with just "100 USD" and one with "X {{localcurrency}} (Y USD)") is worth the added value for the user here. 

i think we're planning for users not to hit this threshold often - we're starting with 100 now but tracking how often it is displayed, and we can remotely configure this value to usage patterns.

i'd prefer to keep the logic simple as it is right now, but i've asked the question in Figma and happy to make a change to add this later, pending feedback from Nitya (and/or Diego)

PR #5013, opened 2024-02-29
refactor(send): cleanup confirmation screen

satish-ravi:
### Description

Changes include:
- Dropping useInputAmounts - was only applicable for the old send flow where amount could be entered in local amount, also dropped convertToMaxPrecision, which was hardcoded to 18 decimals. The prepared tx hook uses the appropriate decimals based on token
- Dropping useRecipientToSendTo - was only applicable for the old send flow where secure send flow was just before the screen. The new send flow guarantees an address in the recipient object and we have typing that ensures this as well
- Trimming comment on send - If a comment is typed and send is immediately pressed, the send payment action will be dispatched before onBlur updates the comment, this leads to the standby tx having an untrimmed comment

### Test plan

CI, manual

### Related issues

Part of ACT-1049

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/src/send/SendConfirmation.test.tsx
+++ b/src/send/SendConfirmation.test.tsx
@@ -264,4 +264,33 @@ describe('SendConfirmation', () => {
       )
     )
   })
+
+  it('trims comment when encrypting and sending', () => {
+    const { getByTestId, queryAllByDisplayValue, store } = renderScreen()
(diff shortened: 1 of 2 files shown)


MuckT (approved):
LGMT! Tested locally on iOS sending a few transactions with comments.

PR #5016, opened 2024-02-29
chore: remove nft gallery and old token details

MuckT:
### Description

Cleanup of old Nft Gallery feature now handled in Assets and of the old token balances screen also now handled in assets. Additionally the following feature flags which have been rolled out to 100% were removed: `show_in_app_nft_gallery`, `show_in_app_nft_viewer`, `view_nft_home_assets` and `show_asset_details_screen`.

### Test plan

- Tested locally on iOS
- E2E tests passing

### Related issues

- Fixes ACT-1074

### Backwards compatibility

Yes

### Network scalability

Yes


src/components/TokenBalance.tsx
```diff
@@ -301,31 +283,22 @@ export function HomeTokenBalance() {
...
-        : Screens.TokenBalances
-    )
+    navigate(Screens.Assets)
   }
 
   return (
     <View style={styles.container} testID="FiatExchangeTokenBalance">
       <View style={styles.titleExchange}>
         <View style={styles.row}>
-          {tokenBalances.length > 1 ? (
```
MuckT, line 321: This behavior was updated because the tests were coupled together. Now the behavior of the home token balances and the fiat exchange token balances matches.

src/components/TokenBalance.test.tsx
```diff
@@ -57,11 +57,7 @@ jest.mocked(getDynamicConfigParams).mockReturnValue({
 describe('FiatExchangeTokenBalance and HomeTokenBalance', () => {
   beforeEach(() => {
     jest.clearAllMocks()
-    jest
-      .mocked(getFeatureGate)
-      .mockImplementation(
-        (featureGate) => featureGate !== StatsigFeatureGates.SHOW_ASSET_DETAILS_SCREEN
-      )
+    jest.mocked(getFeatureGate).mockImplementation(() => true)
```
MuckT, line 60: This is controlling multiple features flags. Open to other options on how to handle this.

src/navigator/Screens.tsx
```diff
@@ -54,7 +54,6 @@ export enum Screens {
   MultichainBeta = 'MultichainBeta',
   NameAndPicture = 'NameAndPicture',
   NotificationCenter = 'NotificationCenter',
-  NftGallery = 'NftGallery',
   NftsInfoCarousel = 'NftsInfoCarousel',
   OnboardingRecoveryPhrase = 'OnboardingRecoveryPhrase',
   OnboardingSuccessScreen = 'OnboardingSuccessScreen',
```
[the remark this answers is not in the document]
MuckT: Removed in 4f7090546bb3fc2b2dcf0c2a1347ce0d93389fff!

src/components/TokenBalance.tsx
MuckT: The behavior is different when show_asset_details_screen was set to true which is 100%. If we want to keep the behavior of the existing FiatExchangeTokenBalance some tests will need to be updated. Due to the coupled testing it was unclear if we ever expected these two components to diverge in behavior like they have.

satish-ravi: I believe the FiatExchangeTokenBalance behavior has always been different from the HomeTokenBalance behavior. 
FiatExchangeTokenBalance shows the ViewBalances touchable if num tokens > 1 both before and after the asset details change. HomeTokenBalance before asset details only included the ViewBalances touchable if num tokens >= 1 and after asset details, it is always included (since we also show tokens with zero balance in the assets screen).

I think we had coupled tests prior just because the case for 0 tokens overlapped. Can't we just drop the [`it.each`](https://github.com/valora-inc/wallet/blob/f0c74d56180c6f336f9884c2fec3480e92af96fb/src/components/TokenBalance.test.tsx#L199) and just keep this test for `FiatExchangeTokenBalance` and keep [this test](https://github.com/valora-inc/wallet/blob/f0c74d56180c6f336f9884c2fec3480e92af96fb/src/components/TokenBalance.test.tsx#L223) for HomeTokenBalance and just remove the feature gate in the name and mocks?

MuckT: Fixed in 62c8ea5201f4d31e467122df8b60f22f5ceedbcc and tests updated in 99c38dcbe8a4679859fef4da97d01ec008ab911c!

src/components/TokenBalance.test.tsx
MuckT: Fixed in 6fb2c1d1f5fc2f5efbb05fe375e373bfd32718c4!

src/components/TokenBalance.test.tsx
```diff
@@ -148,123 +491,12 @@ describe('FiatExchangeTokenBalance and HomeTokenBalance', () => {
...
-        </Provider>
-      )
-
-      expect(tree.queryByTestId('ViewBalances')).toBeFalsy()
-      expect(getElementText(tree.getByTestId('TotalTokenBalance'))).toEqual('$7.91')
-    }
-  )
-
-  it.each([HomeTokenBalance, FiatExchangeTokenBalance])(
+  it.each([HomeTokenBalance])(
```
satish-ravi, line 499: Since this applies to only home token balance, can we move this to its own
describe section? This would also allow us to change the `it.each` to a
`describe.each` so we don't end up adding tests in this section that
only applies to one of them?

src/components/TokenBalance.test.tsx
```diff
@@ -54,14 +54,357 @@ jest.mocked(getDynamicConfigParams).mockReturnValue({
   showBalances: [NetworkId['celo-alfajores']],
 })
 
-describe('FiatExchangeTokenBalance and HomeTokenBalance', () => {
+describe('FiatExchangeTokenBalance', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+    jest.mocked(getFeatureGate).mockReturnValue(true)
```
satish-ravi, line 60: is this required? this component doesn't use a feature gate right?

src/components/TokenBalance.test.tsx
```diff
@@ -54,14 +54,357 @@ jest.mocked(getDynamicConfigParams).mockReturnValue({
...
+    fireEvent.press(tree.getByTestId('EyeIcon'))
+    expect(ValoraAnalytics.track).toHaveBeenCalledTimes(1)
+    expect(ValoraAnalytics.track).toHaveBeenCalledWith(HomeEvents.hide_balances)
+  })
+})
+
+describe('FiatExchangeTokenBalance and HomeTokenBalance', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+    jest.mocked(getFeatureGate).mockReturnValue(true)
```
satish-ravi, line 407: can drop this too since this doesn't do any tests related to the balance toggle
feature gate

MuckT: Was able to drop this feature gate in fcedcf6c96c47fbb27854528ba7e1a59633efdb3.

src/components/TokenBalance.test.tsx
MuckT: This started failing when I removed the feature gate.

satish-ravi: Ah, looks like this is needed for positions to be included. I think we also need this for the describe('FiatExchangeTokenBalance and HomeTokenBalance') section too. I think they currently pass from a stray mock from another describe. We should add that back, because I think running them standalone would fail

src/components/TokenBalance.test.tsx
```diff
@@ -737,6 +419,262 @@ describe('FiatExchangeTokenBalance and HomeTokenBalance', () => {
   })
 })
 
+describe('FiatExchangeTokenBalance and HomeTokenBalance', () => {
```
satish-ravi, line 422: nit: can we just do a describe.each and drop the it.each everywhere else?

satish-ravi (approved):
A once getFeatureGate mock is added

src/components/TokenBalance.test.tsx
MuckT: Fixed in 0459fa9b179ffa6e5c9b340679a9329bde61ca79!

PR #5018, opened 2024-03-01
chore: remove twelveWordMnemonicEnabled remote config

kathaypacific:
### Description

As the title

### Test plan

n/a

### Related issues

- Fixes RET-879

### Backwards compatibility

y

### Network scalability

y


src/onboarding/registration/OnboardingRecoveryPhrase.tsx
```diff
@@ -30,13 +30,11 @@ import { default as useTypedSelector } from 'src/redux/useSelector'
...
 import fontStyles from 'src/styles/fonts'
 import Logger from 'src/utils/Logger'
-import { twelveWordMnemonicEnabledSelector } from 'src/web3/selectors'
 
 type Props = NativeStackScreenProps<StackParamList, Screens.OnboardingRecoveryPhrase>
 
 function OnboardingRecoveryPhrase({ navigation }: Props) {
-  const twelveWordMnemonicEnabled = useSelector(twelveWordMnemonicEnabledSelector)
-  const mnemonicLength = twelveWordMnemonicEnabled ? '12' : '24'
+  const mnemonicLength = 12
```
bakoushin, line 37: nit: maybe we can have this a sort of a global constant, instead of declaring it in every component?

bakoushin (approved):
🧹

src/onboarding/registration/OnboardingRecoveryPhrase.tsx
kathaypacific: ah, thanks for calling this out! i think we don't need this to be easily configurable (seems unlikely we'd want to do this without also thinking about the derivation path) and since it's only being used for translations, i've just updated the translations to use a hard coded value and removed the need for this constant :)

PR #5022, opened 2024-03-02
chore(deps): upgrade viem to ^2.7.16

satish-ravi:
### Description

Includes some typing changes.
- WalletActions need a more specific account type, it doesn't like the `undefined` part of `Account | undefined`
- `baseFeePerGas` from `getBlock` returns `BigInt | null`, which I believe has always been the case. In previous versions, we were getting back `any` as the type which was assigned to `BigInt`

### Test plan

CI

### Related issues

- Fixes ACT-1111

### Backwards compatibility

Yes


src/viem/estimateFeesPerGas.ts
```diff
@@ -9,7 +9,7 @@ import {
 export async function estimateFeesPerGas(
   client: Client,
   feeCurrency?: Address
-): Promise<{ maxFeePerGas: bigint; maxPriorityFeePerGas: bigint; baseFeePerGas: bigint }> {
+): Promise<{ maxFeePerGas: bigint; maxPriorityFeePerGas: bigint; baseFeePerGas: bigint | null }> {
```
satish-ravi, line 12: `baseFeePerGas` from `getBlock` now has a type of `BigInt | null`, it used to be `any` in prior versions. Looks like this will always be returned for the latest block, it's only null for older blocks before the EIP 1559 upgrade https://www.quicknode.com/docs/ethereum/eth_getBlockByHash. Another option to fix this would be to just throw if this is `null` as it would never happen in practice for the latest block.

jeanregisser: I have a slight preference for throwing here, since we're only dealing with EIP-1559 based TXs in the rest of the code.

jeanregisser (approved):
🚀

PR #5024, opened 2024-03-04
feat: reward bottom sheet

bakoushin:
### Description

Show a bottom sheet with a description of a reward associated with a recently celebrated NFT (on NFT celebration see #4886).

The bottom sheet must show up:
* only once within the app session
* if the reward has not expired
* after the NFT celebration was displayed or if the reward is about to expire

https://github.com/valora-inc/wallet/assets/2737872/7ffb0a71-7f53-4e9f-be8e-1a9690bc61f0

The whole NFT celebration / reward / reminder journey is facilitated through reward status in a redux state:

| Status | Expected behavior |
|--------|--------|
| `celebrationReadyToDisplay` | display NFT celebration |
| `celebrationDisplayed` | check if we can display a reward |
| `rewardReadyToDisplay` | display reward bottom sheet |
| `rewardDisplayed` | check if we can display a reminder |
| `reminderReadyToDisplay` | display reminder |
| `reminderDisplayed` | end of journey | 

Notes: 
1. The expiration pill copy when the reward is about to expire (brown one) deliberately doesn't follow the designs because of localization concerns.
2. The former `nftCelebration` state is discarded during migration, because it was never used yet.

### Test plan

 * Tested manually
 * Updated unit tests

### Related issues

- Fixes RET-1002

### Backwards compatibility

Y

### Network scalability


src/home/celebration/NftReward.tsx
```diff
@@ -0,0 +1,245 @@
...
+
+    bottomSheetRef.current?.close()
+
+    const isSecureOrigin = true
+    dispatch(openDeepLink(deepLink, isSecureOrigin))
+  }
+
+  const pillLabel =
+    expirationStatus === ExpirationStatus.expired
+      ? t('nftCelebration.rewardBottomSheet.expired')
```
kathaypacific, line 112: if the reward is expired, should we hide the sheet rather than show a user that they've got expired rewards?

src/home/celebration/NftReward.tsx
```diff
@@ -0,0 +1,245 @@
...
+  )
+}
+
+enum ExpirationStatus {
+  active = 'active',
+  aboutToExpire = 'aboutToExpire',
+  expired = 'expired',
+}
+
+// recalculates expiration status every second
```
kathaypacific, line 148: if i understand correctly, this is necessary only towards the end of the reward period? i think that we can simplify this by setting an expiry date in statsig that is like 30min-1 hour before the actual expiry date - the worst case is that a user who is seeing this really close to the end of the program does not see the reward bottom sheet, however i think this tradeoff is worth the reduced code complexity? (we can also think of it like, we allow a grace period in the backend from the expiry time)

src/home/celebration/NftReward.tsx
```diff
@@ -0,0 +1,245 @@
...
+
+    setExpirationStatus({
+      expirationLabelText: getExpirationLabelText(),
+      expirationStatus: expired
+        ? ExpirationStatus.expired
+        : aboutToExpire
+          ? ExpirationStatus.aboutToExpire
+          : ExpirationStatus.active,
+    })
+    timeoutID.current = setTimeout(updateExpirationStatus, UPDATE_INTERVAL)
```
kathaypacific, line 179: this recursion is kind of blowing my mind, why not do a setInterval instead?

src/home/celebration/NftReward.tsx
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  return { expirationStatus, expirationLabelText }
+}
+
+type ExpirationPillProps = { status: ExpirationStatus; label: string }
+
+const ExpirationPill = ({ status, label }: ExpirationPillProps) => (
+  <View style={styles.pillContainer}>
```
kathaypacific, line 188: nit: i don't think you need this div?

kathaypacific (commented):
i haven't reviewed the tests because i had some questions about the implementation, but looks good!!

src/statsig/constants.ts
```diff
@@ -111,6 +111,9 @@ export const DynamicConfigs = {
     configName: StatsigDynamicConfigs.NFT_CELEBRATION_CONFIG,
     defaultValues: {
       celebratedNft: {} as { networkId?: NetworkId; contractAddress?: string },
+      deepLink: '',
+      expirationDate: new Date(0).toISOString(),
```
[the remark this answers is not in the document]
bakoushin: because the strings are human readable. let machine do the conversion, not us 🤖

src/home/celebration/NftReward.tsx
bakoushin: yes, hiding the sheet could be an option. with it, we have one string less to translate.

src/home/celebration/NftReward.tsx
bakoushin: yes, it's mostly for handling the situation when the reward expires while the bottom sheet is open.

I agree that we could expect that user wouldn't spend too much time looking at this bottom sheet and it's acceptable to not update expiration time and not auto dismiss the bottom sheet once it expires.

I like the idea of having a grace period, giving the late comers the chance to exercise their rewards.

src/home/actions.ts
```diff
@@ -61,6 +85,9 @@ export type ActionTypes =
   | VisitHomeAction
   | CelebratedNftFoundAction
   | NftCelebrationDisplayedAction
+  | NftRewardReadyToDisplayAction
+  | NftRewardDisplayedAction
+  | NftRewardReminderReadyToDisplayAction
```
kathaypacific, line 90: nit: i wonder if NftRewardReadyToDisplayAction and NftRewardReminderReadyToDisplayAction could be combined, the action + reducer + their usage seem very similar, perhaps it'd be more simple to have a single action that takes a boolean `showReminder` param?

src/home/celebration/NftReward.tsx
```diff
@@ -0,0 +1,166 @@
...
+import Colors from 'src/styles/colors'
+import { typeScale } from 'src/styles/fonts'
+import { Spacing } from 'src/styles/styles'
+import { formatDistanceToNow } from 'src/utils/time'
+
+export default function NftRewardBottomSheet() {
+  const { t } = useTranslation()
+  const dispatch = useDispatch()
+
+  const canShowNftReward = useSelector(showNftRewardSelector)
```
kathaypacific, line 29: nit: maybe you don't need to check this again because it is checked in the WalletHome already?

src/home/celebration/NftReward.tsx
```diff
@@ -0,0 +1,166 @@
...
+    }
+  }, [isVisible])
+
+  const handleBottomSheetPositionChange = (index: number) => {
+    if (!nftCelebration) {
+      return // This should never happen
+    }
+
+    if (index === -1) {
+      ValoraAnalytics.track(HomeEvents.nft_reward_dismiss, {
```
kathaypacific, line 80: i think this is fine, but i think this dismiss event will always be fired (even if the accept event is also fired) - perhaps something we should note in the docs so that it's clear how to use this event. currently the doc says that this event is only triggered when the user dismisses the sheet (which i interpreted as dismiss without accepting the reward)

src/home/celebration/NftReward.tsx
```diff
@@ -0,0 +1,166 @@
...
+  button: {
+    marginTop: Spacing.XLarge48,
+  },
+  pill: {
+    alignSelf: 'flex-start',
+    marginBottom: Spacing.Regular16,
+    ...typeScale.labelSemiBoldXSmall,
+    paddingHorizontal: Spacing.Small12,
+    paddingVertical: Spacing.Smallest8,
+    borderRadius: 1000,
```
kathaypacific, line 161: this big radius made me lol 😂

src/nfts/saga.ts
```diff
@@ -112,17 +116,108 @@ export function* findCelebratedNft({ payload: { nfts } }: PayloadAction<FetchNft
...
+  }
+
+  // sync values in the redux state with remote config
+  const valuesToSync = {
+    expirationDate: expirationDateString,
+    rewardReminderDate: rewardReminderDateString,
+    deepLink,
+  }
+
+  const showReminder = isToday(rewardReminderDate) || isPast(rewardReminderDate)
```
kathaypacific, line 193: is `isPast` supposed to be here?

src/nfts/saga.ts
```diff
@@ -112,17 +116,108 @@ export function* findCelebratedNft({ payload: { nfts } }: PayloadAction<FetchNft
...
+  }
+
+  // TODO: check if the user is in allowed list (see ACT-1100)
+
+  const lastNftCelebration = yield* select(nftCelebrationSelector)
+  if (!lastNftCelebration) {
+    return
+  }
+
+  if (!isSameNftContract(lastNftCelebration, celebratedNft)) {
```
kathaypacific, line 182: i found this function still quite hard to understand because there are so many early return conditions. i think there's room to simplify this, e.g. the checks for whether network id / contract address are already done in `isSameNftContract`, so we can remove some of the early return conditions relating to that? the bottom sheet itself also checks if the user owns the nft, so perhaps that condition can also be removed? it'd be ideal to have a condition that is the negative of all these early returns but idk if that's possible

src/home/celebration/NftReward.test.tsx
```diff
@@ -0,0 +1,280 @@
...
+        <NftReward />
+      </Provider>
+    )
+
+    expect(queryByText('nftCelebration.rewardBottomSheet.title')).toBeNull()
+    expect(queryByText('nftCelebration.rewardBottomSheet.description')).toBeNull()
+    expect(queryByText('nftCelebration.rewardBottomSheet.cta')).toBeNull()
+  })
+
+  it('hanldes the cta correctly', () => {
```
kathaypacific, line 214:
```suggestion
  it('handles the cta correctly', () => {
```

src/home/celebration/NftReward.test.tsx
```diff
@@ -0,0 +1,280 @@
...
+    jest.useFakeTimers().setSystemTime(new Date('3000-11-01T00:00:00.000Z').getTime())
```
kathaypacific, line 215: i think we need to reset this mock in the beforeEach's probably? otherwise this will affect the next tests, not a big deal now but could potentially cost some head scratching time in the future (like https://github.com/valora-inc/wallet/pull/4953#pullrequestreview-1895509126)

kathaypacific (approved):
LGTM!!

src/nfts/saga.ts
bakoushin: I believe, yes.

here we are checking if we are within the reminder period.

if user opens the app within the reminder period, but not exactly on the reminder start date, `rewardReminderDate` would be in the past.

for example:
* user opened app: March, 2 
* `rewardReminderDate`: March, 1 (`isPast === true`)

src/home/celebration/NftReward.tsx
bakoushin: that's a great catch! thanks!
the docs are correct: this event must dispatch only if the reward wasn't accepted. fixed it by adding a state variable.

sadly I can't figure out yet how to simulate bottom sheet dismiss event in tests to cover it with a test.

src/home/celebration/NftReward.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+  const insets = useSafeAreaInsets()
+  const insetsStyle = { paddingBottom: Math.max(insets.bottom, Spacing.Regular16) }
+
+  const bottomSheetRef = useRef<BottomSheetRefType>(null)
+
+  const expirationDate = new Date(nftCelebration?.expirationDate ?? 0)
+  const rewardReminderDate = new Date(nftCelebration?.rewardReminderDate ?? 0)
+
+  const aboutToExpire = isToday(rewardReminderDate) || isPast(rewardReminderDate)
```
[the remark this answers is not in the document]
bakoushin: I'm sorry, but I'm not getting the idea against `isPast`.

we want to show this bottom sheet in a "reminder state" if the current date falls within the reminder period.
it is not necessary that the user will open the app on the first day of this period, that's why we want to check if the reminder date is in the past.

but I agree that instead of looking at the reminder date we can rely on the `NftCelebrationStatus` instead. thanks for the suggestion!

src/nfts/saga.ts
kathaypacific: OH of course, you're right. i was confusing this with the expiry date

src/home/actions.ts
bakoushin: sure, thanks!

PR #5027, opened 2024-03-04
feat: add tab navigator

MuckT:
### Description

Adds the tab navigator and associated icons. Behavior is enabled with the `use_tab_navigator` feature gate. Please disregard the header display as this will be covered in ACT-1108.

| Android | iOS 15 Pro Max - Large Font | iOS SE (3rd Generation) | 
| ----- | ----- | ----- | 
|  |  |  Assets") |

### Test plan

- [x] Tested locally on iOS with feature flag enabled and disabled
- [x] Tested locally on Android with feature flag enabled and disabled
- [x] Unit tests updated

### Related issues

- Fixes ACT-1103

### Backwards compatibility

Yes - behind the `use_tab_navigator` feature gate.

### Network scalability

N/A


src/navigator/NavigationService.ts
```diff
@@ -238,7 +240,9 @@ export function navigateHome(options?: NavigateHomeOptions) {
   setTimeout(() => {
     navigationRef.current?.reset({
       index: 0,
-      routes: [{ name: Screens.DrawerNavigator, params }],
+      routes: getFeatureGate(StatsigFeatureGates.USE_TAB_NAVIGATOR)
```
MuckT, line 243: On Android I am getting the error: `The action 'RESET' with payload {"index":0,"routes":[{"name":"TabNavigator"}]} was not handled by any navigator.` likely form here. When testing the non tab navigator path no error is served. On iOS I have not observed this same issue. Perhaps this should be fixed the follow up ACT-1110.

src/navigator/TabNavigator.tsx
```diff
@@ -0,0 +1,73 @@
...
+        headerShown: false,
+        tabBarActiveTintColor: Colors.black,
+        tabBarInactiveTintColor: Colors.gray3,
+        tabBarLabelStyle: styles.label,
+      }}
+    >
+      <Tab.Screen
+        // TODO(act-1104) new assets screen
+        name={Screens.Assets}
+        // @ts-expect-error Type '{}' is missing the following properties from type 'Props': navigation, route
```
silasbw, line 38: can we avoid using `@ts-expect-error` here? I worry (in general) that will forget about suppressed errors and they eventually surface as errors to the user experience.

src/navigator/NavigationService.ts
silasbw: i'm confused by your PR comment and the comment in the code: the code says there's a TODO to handle undefined/falsey, the code seems to handle it by reverting to previous behavior (DrawerNavigator), and you PR comment seems to imply that addresses act-1110 will solve a related (unrelated?) problem?

src/icons/AccountCircle.tsx
```diff
@@ -0,0 +1,18 @@
...
+import Svg, { Path } from 'react-native-svg'
+import colors, { Colors } from 'src/styles/colors'
+
+interface Props {
+  size?: number
+  color?: colors
+}
+
+const AccountCircle = ({ size = 24, color = Colors.successDark }: Props) => (
+  <Svg width={size} height={size} fill="none">
```
satish-ravi, line 11: can you include a viewBox param in all the icons? Otherwise adjusting with size won't work as expected

locales/base/translation.json
```diff
@@ -2140,5 +2140,10 @@
     }
   },
   "pleaseWait": "Please wait",
-  "error": "Error"
+  "error": "Error",
+  "tabBar": {
```
satish-ravi, line 2144: naming nit: can we make this more explicit? since we also have a tab bar in the assets page / wallet tab

src/navigator/DrawerTopBar.tsx
```diff
@@ -20,6 +25,7 @@ interface Props {
 
 function DrawerTopBar({ leftElement, middleElement, rightElement, scrollPosition, testID }: Props) {
   const navigation = useNavigation()
+  const useTabNavigator = getFeatureGate(StatsigFeatureGates.USE_TAB_NAVIGATOR)
```
satish-ravi, line 28: this works for now, but ideally we should make a separate component for the new header, which we could get to in ACT-1108, maybe leave a TODO?

src/navigator/Navigator.tsx
```diff
@@ -602,7 +605,15 @@ export function MainStackScreen() {
 
   return (
     <Stack.Navigator initialRouteName={initialRouteName} screenOptions={emptyHeader}>
-      <Stack.Screen name={Screens.DrawerNavigator} component={DrawerNavigator} options={noHeader} />
+      {getFeatureGate(StatsigFeatureGates.USE_TAB_NAVIGATOR) ? (
```
satish-ravi, line 608: do we have to conditionally register these screens?

src/navigator/types.tsx
```diff
@@ -267,6 +267,10 @@ export type StackParamList = {
       }
     | undefined
   [Screens.SwapScreenWithBack]: { fromTokenId: string } | undefined
+  [Screens.TabNavigator]: {
+    initialScreen?: Screens
+    fromModal?: boolean
```
satish-ravi, line 272: is this param required?

satish-ravi (commented):
The padding on the tab bar looks off compared to figma

src/navigator/TabNavigator.tsx
MuckT: This is just linking to the existing assets screen, but that screen be redone in act-1104 which should resolve this todo.

src/navigator/NavigationService.ts
MuckT: Act-1110 is the feature gate ticket, as this relates to the feature gate not being defined when the navigation service is initialized, we try to initialize Statsig as early as possible, it seems appropriate to lump it into that work.

locales/base/translation.json
MuckT: Changed in d428981588b06c5901efb0ed709f27b02cee26b7!

src/navigator/DrawerTopBar.tsx
MuckT: Added Todo's in d7ba1693653b9e9699a3e790e12190a5a7ce6efd!

src/navigator/Navigator.tsx
MuckT: Registered both in 5e5a279b44481f32f53dd4b4f9bd411c79d6f71a!

src/navigator/types.tsx
MuckT: It's needed for `navigateHome` in `src/navigator/NavigationService.ts`.

src/icons/AccountCircle.tsx
MuckT: Added in 370ffa7e401ae7c2b9dc33be98054855d04e1c6f!

MuckT:
@satish-ravi added some padding in to better match styles in c5bc02d62f5e9e2a0e8621ed8030c068da90d950. It's a bit of a compromise as getting it pixel perfect on one screen can lead to some strange display on smaller screens; I've updated the screenshots in the description.

src/navigator/NavigationService.ts
```diff
@@ -224,7 +226,7 @@ export async function isBottomSheetVisible(screen: Screens) {
 }
 
 interface NavigateHomeOptions {
-  params?: StackParamList[Screens.DrawerNavigator]
+  params?: StackParamList[Screens.DrawerNavigator] | StackParamList[Screens.TabNavigator]
```
satish-ravi, line 229: This type is kind of a problem, as it masks a potential issue. The `initialScreenName` accepted by the two screens are actually going to be different. What if a caller passes an `initialScreenName` that only applies to the DrawerNavigator. E.g., we have a the CYA screen passing `DAppsExplorerScreen` and `ExchangeHomeScreen`, would it be on the caller to also get the gate and pass the appropriate screen name?

src/navigator/TabNavigator.tsx
```diff
@@ -0,0 +1,81 @@
...
+        tabBarActiveTintColor: Colors.black,
+        tabBarInactiveTintColor: Colors.gray3,
+        tabBarLabelStyle: styles.label,
+        tabBarItemStyle: styles.tabBarItem,
+        tabBarAllowFontScaling: false,
+      }}
+    >
+      <Tab.Screen
+        // TODO(act-1104) new assets screen
+        name={Screens.Assets}
```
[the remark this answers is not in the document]
MuckT: I think it's fine to leave these as navigating as is. I've tested the navigation when `StatsigFeatureGates.USE_TAB_NAVIGATOR` is false and the existing navigation is tested in the e2e tests.

src/navigator/NavigationService.ts
MuckT: It looks like react-navigation handles internally. I've tested navigating from the CYA screen to each option and the correct screen is served for both paths for the `use_tab_navigator` feature gate.

src/navigator/TabNavigator.tsx
```diff
@@ -0,0 +1,85 @@
...
+        // @ts-expect-error Type '{}' is missing the following properties from type 'Props': navigation, route
+        component={AssetsScreen}
+        options={{
+          tabBarLabel: t('bottomTabsNavigator.wallet.tabName') as string,
+          tabBarIcon: Wallet,
+        }}
+      />
+      <Tab.Screen
+        // TODO(act-1105) new home tab screen
+        name={Screens.TabWallet}
```
satish-ravi, line 56:
```suggestion
        name={Screens.TabHome}
```

src/navigator/NavigationService.ts
satish-ravi: guess this might be broken now with changing the screen names, but we can fix it in ACT-1110

satish-ravi (approved):

src/navigator/TabNavigator.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { Spacing } from 'src/styles/styles'
+import variables from 'src/styles/variables'
+import AssetsScreen from 'src/tokens/Assets'
+
+const Tab = createBottomTabNavigator()
+
+type Props = NativeStackScreenProps<StackParamList, Screens.TabNavigator>
+
+export default function TabNavigator({ route }: Props) {
+  const initialScreen = route.params?.initialScreen ?? Screens.WalletHome
```
satish-ravi, line 24:
```suggestion
  const initialScreen = route.params?.initialScreen ?? Screens.TabHome
```

PR #5029, opened 2024-03-04
ci: don't run jobs on main that already run on the merge queue

satish-ravi:
### Description

Now that we have merge queue enabled for main, every PR runs checks at least twice, once on the PR before being added to the queue and once on the merge queue. The commit on the merge queue is the same commit that lands on main, so we don't need to run checks again on main that have already run on the merge queue (see image below on checks running twice on commits on main). This disables `check` and `test` workflows on main, and updates the `e2e-main` workflow to only run tests on OS that have not already run on the merge queue. This should save us some CI time, especially on E2E machines. 

image

### Test plan

CI

### Related issues

N/A

### Backwards compatibility

N/A

### Network scalability

N/A

### DIFF
--- a/.github/workflows/e2e-main.yml
+++ b/.github/workflows/e2e-main.yml
@@ -22,23 +22,24 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  android:
-    name: Android
-    strategy:
-      max-parallel: 2
-      fail-fast: false
-      matrix:
-        # TODO if/when more CI machines: enable Android sdk 21, 23, 29 and 31
-        # 21 is failing 9 specs
-        # 23 is failing ? spec(s)
-        # 31 is failing ? spec(s)
-        # 24 is currently failing all the time, disabling it for now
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e-main.yml
```diff
@@ -33,7 +33,8 @@ jobs:
         # 23 is failing ? spec(s)
         # 31 is failing ? spec(s)
         # 24 is currently failing all the time, disabling it for now
-        android-api-level: [30]
+        # 30 is not included as it runs on the merge queue
+        android-api-level: []
```
MuckT, line 37: Does this work with an empty array?

MuckT (approved):

.github/workflows/e2e-main.yml
satish-ravi: good call. It doesn't. commented the entire job for now

silasbw (approved):

PR #5038, opened 2024-03-05
feat(jumpstart): Adding support for jumpstart transactions

dievazqu:
### Description

As title says, updating the feed logic to support new jumpstart transactions.

Note: The transaction details screen is a mock one, it's gonna be improved in a next PR

### Test plan

Locally in a emulator hitting the blockchain-api locally with the new code.

### Related issues

- Fixes RET-996

### Backwards compatibility

No, we have to deploy the blockchain-api first.

### Network scalability

Continue to work without code changes.

### Screenshot

Simulator Screen Shot - iPhone SE (3rd generation) - 2024-03-05 at 09 43 06


src/transactions/feed/TransferFeedItem.tsx
```diff
@@ -51,6 +50,7 @@ function TransferFeedItem({ transfer }: Props) {
           recipient={recipient}
           status={transfer.status}
           transactionType={transfer.__typename}
+          transactionSubtype={transfer.type}
```
bakoushin, line 53: i wonder if we need this, because `__typename` is already different for `TokenTransver` and `JumpstartDeposit`

src/transactions/feed/TransactionFeed.tsx
```diff
@@ -84,6 +84,8 @@ function TransactionFeed() {
         return <NftFeedItem key={tx.transactionHash} transaction={tx} />
       case 'TokenApproval':
         return <TokenApprovalFeedItem key={tx.transactionHash} transaction={tx} />
+      case 'JumpstartDeposit':
```
bakoushin, line 87: nit: what if we list this case right below `TokenTransferV3`?
```typescript
case 'TokenTransferV3':
case 'JumpstartDeposit':
  return <TransferFeedItem key={tx.transactionHash} transfer={tx} />
```

src/transactions/feed/TransactionFeedItemImage.tsx
```diff
@@ -20,12 +21,17 @@ type Props =
...
   | {
       status: TransactionStatus
       transactionType: 'TokenApproval'
     }
+  | {
+      status: TransactionStatus
+      transactionType: 'JumpstartDeposit'
+    }
 
 function TransactionFeedItemImage(props: Props) {
```
bakoushin, line 36: maybe we can can use `transfer.type` instead of `__typename` as a `transactionType` because looks like it's more detailed? then we can not bother with introducing a subtype.

src/transactions/feed/TransactionFeed.tsx
dievazqu: Yes, sorry! At first I was considering using a different component...
Thanks!

locales/base/translation.json
```diff
@@ -1309,6 +1309,11 @@
   "feedItemGoldWithdrawal": "Withdrawal to {{displayName}}",
   "feedItemGoldReceived": "Received from {{displayName}}",
   "feedItemSwapPath": "{{token1}} to {{token2}}",
+  "feedItemJumpstartDepositTitle": "Claim link",
+  "feedItemJumpstartSubtitle": "Invitation",
+  "feedItemJumpstartReclaimedTitle": "Reclaim",
+  "feedItemJumpstartReceivedTitle": "Funds received",
+  "feedItemJumpstartReceivedSubitle": "- -",
```
bakoushin, line 1316: looks like this one is never used

src/transactions/feed/TransactionFeedItemImage.tsx
bakoushin: thanks for the detailed analysis! 
yes, I agree that this is a less destructive approach to reach the goal.

bakoushin (approved):

locales/base/translation.json
dievazqu: Ups, typo!

src/icons/JumpstartIcon.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react'
+import Svg, { Path, Rect } from 'react-native-svg'
+
+const JumpstartIcon = () => (
```
kathaypacific, line 4: is this the magic wand icon? i already committed this, it is called MagicWand

src/transactions/feed/TransferFeedItem.tsx
kathaypacific, line 53: nit: perhaps it'd be less confusing if we renamed these props to `transactionTypename` and `transactionType`?

src/transactions/feed/TransactionFeedItemImage.tsx
```diff
@@ -51,7 +57,16 @@ function TransactionFeedItemImage(props: Props) {
     )
   }
   if (transactionType === 'TokenTransferV3') {
-    return <ContactCircle recipient={props.recipient} size={AVATAR_SIZE} />
+    switch (props.transactionSubtype) {
```
kathaypacific, line 60: why are we using the "subType" instead of the __typename here? the jumpstart transactions have a different __typename, it's confusing that we implicitly assume that they have the __typename of 'TokenTransferV3' here?

src/transactions/types.ts
```diff
@@ -100,7 +101,12 @@ export enum TransactionStatus {
   Failed = 'Failed',
 }
 
-export type TokenTransaction = TokenTransfer | TokenExchange | NftTransfer | TokenApproval
+export type TokenTransaction =
+  | TokenTransfer
+  | TokenExchange
+  | NftTransfer
+  | TokenApproval
+  | JumpstartDeposit
```
kathaypacific, line 109:
```suggestion
  | JumpstartTransfer
```

nit: naming is hard, but there is a type called `JumpstartDeposit` in `TokenTransactionTypeV2` already, can we avoid duplicates? the "deposit" part in `TokenTransactionTypeV2` refers to a user putting funds in the jumpstart contract right...so then this transfer type should not also be called "deposit" since it also means users reclaiming and receiving funds from the jumpstart contract?

kathaypacific (commented):
looking good!

src/transactions/feed/TransactionFeedItemImage.tsx
dievazqu: Well... Actually the only tx that has `__typename` as `JumpstartDeposit` is the deposit...
The reason for this is because the `reclaim` and the `received` don't need the extra fields, so it was enough to use the `TokenTransferV3` schema.
I guess we can identify those txs also as `JumpstartTransactions`, so the code would be clearer here, but I think I prefer to use the new type to indicate there are new fields in the schema.

src/transactions/types.ts
dievazqu: This is related to the other comment too, I see how this is confusing.

In my proposed solution:
The received and reclaim are a new type of `TokenTransferV3`. At the end, it's just really a transfer that we want some special UX. And the new transaction type is the deposit that includes the relevant fields that are needed to execute the reclaim.

An alternative solution would be to put all the jumpstart transactions under a single `__typename` but the fact that the schema is different was taking me back from this solution.

src/transactions/feed/TransactionFeedItemImage.tsx
kathaypacific: ah, i understood the types incorrectly. i thought that all jumpstart transactions would have the same __typename but i see that was not your intention [here](https://github.com/valora-inc/blockchain-api/pull/380/files#diff-f25aebc206e96407bd5d65c2ec780b48aa1f7f04efb1b9c7ddd41c617f87b51aL178)

kathaypacific (approved):

dievazqu:
Replaced by https://github.com/valora-inc/wallet/pull/5120

PR #5044, opened 2024-03-05
fix(keylessBackup): fix hasSeenVerification for linkPhone number and when homecard is shown

jh2oman:
### Description

* Homecard should not be shown if the user restored with keylessBackup
* If verification is skipped after keylessbackup, set the state correctly so the user isn't shown verification next time they open the app

### Test plan

Unit tests, manual test

### DIFF
--- a/src/home/NotificationBox.tsx
+++ b/src/home/NotificationBox.tsx
@@ -101,8 +101,8 @@ export function useSimpleActions() {
   const cloudBackupCompleted = useSelector(cloudBackupCompletedSelector)
 
   const actions: SimpleAction[] = []
-  if (!backupCompleted) {
-    if (showKeylessBackup && !cloudBackupCompleted) {
+  if (!backupCompleted && !cloudBackupCompleted) {
+    if (showKeylessBackup) {
       actions.push({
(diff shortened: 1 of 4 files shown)


src/home/NotificationBox.tsx
```diff
@@ -98,10 +98,11 @@ export function useSimpleActions() {
   }, [])
 
   const superchargeRewards = useSelector((state) => state.supercharge.availableRewards)
+  const cloudBackupCompleted = useSelector(cloudBackupCompletedSelector)
 
   const actions: SimpleAction[] = []
   if (!backupCompleted) {
-    if (showKeylessBackup) {
+    if (showKeylessBackup && !cloudBackupCompleted) {
```
satish-ravi, line 105: should the cloud backup check be in the if above? We don't want to show any backup related home card if they restored right? Maybe worth doing a migration to rename `backupCompleted` or make a wrapper selector that checks both

jh2oman: Yea you're right, i should move it up

satish-ravi (approved):

PR #5045, opened 2024-03-05
fix(keylessBackup): Use existing component for keyless backup delete error

jophish:
### Description

Updates the keyless backup delete error banner to use an existing component. UI remains unchanged.

### Test plan

Manual and unit tested (existing unit test needed no change).

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

Yes.

### Network scalability


src/components/BottomSheetInLineNotification.tsx
```diff
@@ -17,9 +18,21 @@ interface Props extends InLineNotificationProps {
...
 // for now, this notification is launched from the bottom of the screen only
-const BottomSheetInLineNotification = ({ showNotification, ...inLineNotificationProps }: Props) => {
+const BottomSheetInLineNotification = ({
+  showNotification,
+  onUnmount,
+  ...inLineNotificationProps
+}: Props) => {
   const [isVisible, setIsVisible] = useState(showNotification)
 
+  useEffect(() => {
```
jophish, line 28: This unmount hook is needed for the edge case where the user does not dismiss the error banner, and opens the side nav to navigate away from the page. When the banner is visible, the settings screen is otherwise uninteractable, but the user can still open the side nav. Without this hook, the user would see the error banner immediately the next time they navigate to the settings screen.

satish-ravi (approved):

PR #5050, opened 2024-03-06
feat: add jumpstart confirmation screen

kathaypacific:
### Description

This PR adds:
- a new jumpstart confirmation screen. it is just displaying information at this stage.
- navigation to this new screen from the enter amount screen
- some analytics 
- unit tests for the new screen

Not included: sending the transaction

### Test plan

Simulator Screenshot - iPhone 14 Pro - 2024-03-07 at 10 38 04

### Related issues

- Related to RET-994

### Backwards compatibility

Y

### Network scalability

Y


src/analytics/Events.tsx
```diff
@@ -330,11 +330,6 @@ export enum SendEvents {
   send_select_recipient_recent_press = 'send_select_recipient_recent_press',
 }
 
-export enum JumpstartEvents {
```
jeanregisser, line 333: Interesting that TS lets us export the same name twice 🤔

src/jumpstart/JumpstartSendConfirmation.tsx
```diff
@@ -0,0 +1,122 @@
...
+      tokenSymbol: token.symbol,
+      tokenAmount: parsedAmount.toString(),
+      amountInUsd: parsedAmount.multipliedBy(token.priceUsd ?? 0).toFixed(2),
+      tokenId: token.tokenId,
+      networkId: token.networkId,
+    })
+  }
+
+  return (
+    <ScrollView contentContainerStyle={styles.container}>
```
jeanregisser, line 43: nit: maybe we should pad the bottom with the safe area? for when the content can grow big due to a large font for instance

```
<SafeAreaView edges={['bottom']}>
```

jeanregisser (approved):
🚀

PR #5056, opened 2024-03-07
refactor(tokens): break out AssetTabBar and AssetList into its own components

satish-ravi:
### Description

Break out AssetTabBar and AssetList from the Assets screen into its own components. This makes it easier to reason about and also would allow us to reuse components for the Wallets tab screen

### Test plan

Unit and manual tested

### Related issues

- Relates to ACT-1104

### Backwards compatibility

Yes

### Network scalability

N/A


src/tokens/AssetList.test.tsx
satish-ravi, line 1: This is just a subset of tests from Assets.test.tsx

src/tokens/AssetList.tsx
satish-ravi, line 1: Entirely copied from Assets.tsx

src/tokens/AssetTabBar.tsx
satish-ravi, line 1: Copied from Assets.tsx

src/tokens/PositionItem.tsx
```diff
@@ -77,94 +73,19 @@ export const PositionItem = ({ position }: { position: Position }) => {
   )
 }
 
-export const TokenBalanceItem = ({
```
satish-ravi, line 80: No longer needed used after #5016 , knip didn't catch it because it was referenced in tests

silasbw (approved):

PR #5060, opened 2024-03-07
chore: refactor send payment saga to be more generic

kathaypacific:
### Description

There should be no user facing changes in this PR. This PR moves things around so that it will be easier to reuse the send payment saga for the jumpstart flow:
1. lots of trivial import path changes related to moving the `newTransactionContext` helper from the types.ts to utils.ts file. types files should be reserved only for type definitions, it felt very bad to be doing `jest.mocked('whateverPath/types')`
2. the payment saga was previously responsible for sending the saga + adding the pending transaction + navigating to the next screen. i removed the navigation logic out of the saga - every place that can send a transaction needs a different navigation behaviour. it seems like screens should be monitoring the transaction and adding their own navigation logic, rather than the saga needing to know every navigation expectation. 
3. in order for the screens to be able to monitor the success of "their" transaction, i am now creating the transaction context in the screen itself and passing it to the saga. previously the saga was creating the context. this means that the screen can then monitor the "recentPayments" redux state (previously unused.....) and conclude that the tx was successful when an item the matching context id appears there.
4. the pending tx still needs to be managed in the saga because it's only there that we have access to the tx hash. so i've overloaded the action parameters and now the consumer must specify which type of pending tx is expected.
5. I did not include a redux migration because the "recentPayments" is not used elsewhere and the logic here is safe to execute without the contextId property

### Test plan

Rely on CI to detect regression.

### Related issues

- Related to RET-994

### Backwards compatibility

Y

### Network scalability
Y


src/send/SendConfirmation.tsx
```diff
@@ -87,9 +89,27 @@ function SendConfirmation(props: Props) {
...
+    )
+    if (transactionSuccessful) {
+      if (props.route.name === Screens.SendConfirmationModal) {
+        navigateBack()
+      } else {
+        navigateHome()
+      }
+    }
+  }, [recentSentTransactions])
+
```
kathaypacific, line 112: this part is the key change

src/send/reducers.ts
```diff
@@ -41,30 +42,42 @@ export const sendReducer = (
...
       return {
         ...state,
         ...getRehydratePayload(action, 'send'),
         isSending: false,
         encryptedComment: null,
         isEncryptingComment: false,
+        // Keep only the last 24 hours recent payments
+        recentPayments: state.recentPayments.filter(
+          (p: PaymentInfo) => timeDeltaInHours(now, p.timestamp) < 24
+        ),
```
kathaypacific, line 55: this is just cleanup, seems like we should do this on rehydrate than on new payments succeeded

kathaypacific:
closing in favour of #5067

PR #5062, opened 2024-03-08
fix: ensure screens that are bottom sheets scroll correctly

kathaypacific:
### Description

Context https://valora-app.slack.com/archives/C025V1D6F3J/p1709855852378389

The bottom sheet screens were not scrolling correctly, this was probably a regression but unnoticed because most of these screens do not require scrolling. Root cause is a known issue that is documented in a comment in the Navigator - we need to use the react-native-gesture-handler components to scroll inside bottom sheet screens.

This PR adds a prop to the BottomSheetScrollView component to use the correct scrollview component.

### Test plan

Scrolling content:

https://github.com/valora-inc/wallet/assets/20150449/349c2628-017c-46b1-af88-06d20bcb19ce

Short content:

https://github.com/valora-inc/wallet/assets/20150449/5929f8f9-08f1-495f-8131-211cd055fbdb

Tested also on the WC flow.

### Related issues

- n/a

### Backwards compatibility

Y

### Network scalability

Y

### DIFF
--- a/src/components/BottomSheetScrollView.tsx
+++ b/src/components/BottomSheetScrollView.tsx
@@ -1,27 +1,50 @@
 import { BottomSheetScrollView as GorhomBottomSheetScrollView } from '@gorhom/bottom-sheet'
 import React, { useState } from 'react'
-import { LayoutChangeEvent, ScrollView, StyleProp, StyleSheet, View, ViewStyle } from 'react-native'
+import { LayoutChangeEvent, StyleProp, StyleSheet, View, ViewStyle } from 'react-native'
+import { ScrollView } from 'react-native-gesture-handler'
 import { useSafeAreaInsets } from 'react-native-safe-area-context'
 import { Spacing } from 'src/styles/styles'
+import variables from 'src/styles/variables'
 
 interface Props {
   containerStyle?: StyleProp<ViewStyle>
   testId?: string
   forwardedRef?: React.RefObject<ScrollView>
+  isScreen?: boolean
   children: React.ReactNode
 }
 
-function BottomSheetScrollView({ forwardedRef, containerStyle, testId, children }: Props) {
+function BottomSheetScrollView({
+  forwardedRef,
+  containerStyle,
+  testId,
+  isScreen,
+  children,
+}: Props) {
   const [containerHeight, setContainerHeight] = useState(0)
   const [contentHeight, setContentHeight] = useState(0)
 
   const insets = useSafeAreaInsets()
   const scrollEnabled = contentHeight > containerHeight
 
+  // Note: scrolling views inside bottom sheet screens should use the relevant
+  // components from react-native-gesture-handler instead of directly from
+  // react-native, otherwise they do not scroll correctly. This isScreen prop
+  // should be set to true if the bottom sheet is registered as screen in the
+  // Navigator. It is still handy for screens and components to share this
+  // component for the styling and layout logic.
+  // https://github.com/osdnk/react-native-reanimated-bottom-sheet/issues/264#issuecomment-674757545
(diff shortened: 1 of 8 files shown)


src/components/BottomSheetScrollView.tsx
```diff
@@ -1,27 +1,50 @@
...
+  // react-native, otherwise they do not scroll correctly. This isScreen prop
+  // should be set to true if the bottom sheet is registered as screen in the
+  // Navigator. It is still handy for screens and components to share this
+  // component for the styling and layout logic.
+  // https://github.com/osdnk/react-native-reanimated-bottom-sheet/issues/264#issuecomment-674757545
+  const ScrollViewComponent = isScreen ? ScrollView : GorhomBottomSheetScrollView
+  // use max height simulate max 90% snap point for screens. when bottom sheets
+  // take up the whole screen, it is no longer obvious that they are a bottom
+  // sheet / how to navigate away
+  const maxHeight = isScreen ? variables.height * 0.9 : undefined
```
kathaypacific, line 41: annoyingly i couldn't find a better way to set the snappoints to be dynamic + set an upper bound on the height

src/components/BottomSheetScrollView.tsx
```diff
@@ -1,27 +1,50 @@
...
   const scrollEnabled = contentHeight > containerHeight
 
+  // Note: scrolling views inside bottom sheet screens should use the relevant
+  // components from react-native-gesture-handler instead of directly from
+  // react-native, otherwise they do not scroll correctly. This isScreen prop
+  // should be set to true if the bottom sheet is registered as screen in the
+  // Navigator. It is still handy for screens and components to share this
+  // component for the styling and layout logic.
+  // https://github.com/osdnk/react-native-reanimated-bottom-sheet/issues/264#issuecomment-674757545
+  const ScrollViewComponent = isScreen ? ScrollView : GorhomBottomSheetScrollView
```
jeanregisser, line 37: What happens if we use the RNGH ScrollView everywhere?
Does it break any gesture? Maybe the overscroll to dismiss?

src/components/BottomSheetScrollView.tsx
```diff
@@ -1,27 +1,50 @@
...
+import { ScrollView } from 'react-native-gesture-handler'
 import { useSafeAreaInsets } from 'react-native-safe-area-context'
 import { Spacing } from 'src/styles/styles'
+import variables from 'src/styles/variables'
 
 interface Props {
   containerStyle?: StyleProp<ViewStyle>
   testId?: string
   forwardedRef?: React.RefObject<ScrollView>
+  isScreen?: boolean
```
jeanregisser, line 13: Maybe add a small comment to explain when/why this is needed?

jeanregisser (approved):
🚀

src/components/BottomSheetScrollView.tsx
kathaypacific: yes it is the overscroll to dismiss that is not available with ScrollView. i updated the comment to be more clear about this. it's annoying that this is an issue, but i think the overscroll / bottom sheet gestures are probably the thing preventing the BottomSheetScrollView to work correctly from the navigator

src/components/BottomSheetScrollView.tsx
kathaypacific: yes updated

kathaypacific:
it seems like BottomSheetScrollView and ScrollView manage component identity differently - using ScrollView causes the WalletConnectRequest component to not work well because the inner RequestContent is unmounted (i guess, remounted) at least once. unfortunately the unmount dispatches a deny action here, but for the dappkit and fiatexchange use cases this is not a problem. i'm going to merge as is to unblock the release and work on a better fix for the WC stuff

PR #5064, opened 2024-03-08
fix: new wallets only created with english mnemonics

MuckT:
### Description

Wallets created when the users language was set to Spanish or Portuguese had mnemonics that were in Spanish or Portuguese. This causes interoperability issues for Valora users who wish to try other wallets: [Slack Thread](https://valora-app.slack.com/archives/C02KBT0DAHJ/p1709917662991769).

### Test plan

- Tested locally on iOS
- Unit tests updated

### Related issues

N/A

### Backwards compatibility

Yes - Users with Spanish and Portuguese mnemonics are still able to restore into Valora.

### Network scalability

N/A

### DIFF
--- a/src/web3/saga.ts
+++ b/src/web3/saga.ts
@@ -61,12 +61,7 @@ export function* getOrCreateAccount() {
 
     const mnemonicBitLength = MnemonicStrength.s128_12words
     const mnemonicLanguage = MnemonicLanguages.english
-    let mnemonic: string = yield* call(
-      generateMnemonic,
-      mnemonicBitLength,
-      MnemonicLanguages.english,
-      bip39
-    )
+    let mnemonic: string = yield* call(generateMnemonic, mnemonicBitLength, mnemonicLanguage, bip39)
 
     // Ensure no duplicates in mnemonic
     const checkDuplicate = (someString: string) => {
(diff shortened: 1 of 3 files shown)


src/web3/saga.ts
```diff
@@ -61,8 +60,13 @@ export function* getOrCreateAccount() {
     Logger.debug(TAG + '@getOrCreateAccount', 'Creating a new account')
 
     const mnemonicBitLength = MnemonicStrength.s128_12words
-    const mnemonicLanguage = getMnemonicLanguage(yield* select(currentLanguageSelector))
-    let mnemonic: string = yield* call(generateMnemonic, mnemonicBitLength, mnemonicLanguage, bip39)
+    const mnemonicLanguage = MnemonicLanguages.english
+    let mnemonic: string = yield* call(
+      generateMnemonic,
+      mnemonicBitLength,
+      MnemonicLanguages.english,
```
finnian0826, line 67: Can this just be set to `mnemonicLanguage`? since it is set as english in the lines above.

MuckT: Fixed in b9d9e56c4f7f8219b05fdd5ad152a918146c1835!

finnian0826 (approved):
looks like knip check is failing because `getMnemonicLanguage` is no longer used anywhere so should be removed. Other than that looks good!

PR #5067, opened 2024-03-11
refactor: shared transaction sending saga for swap and send

kathaypacific:
### Related issues

- Fixes #[issue number here]

### Network scalability


src/send/saga.ts
```diff
@@ -303,14 +287,72 @@ export function* sendPaymentSaga({
...
 }
 
+export function* sendTransactionsSaga(
+  serializablePreparedTransactions: SerializableTransactionRequest[],
+  networkId: NetworkId,
+  createBaseStandbyTransactions: ((
+    hash: string,
+    feeCurrencyId?: string
+  ) => BaseStandbyTransaction)[]
+) {
```
jeanregisser, line 297: Great!

Should we maybe document the arguments?
Explaining what `createBaseStandbyTransactions` is for.

Also what do you think of moving this saga to `src/viem/sagas`?
And maybe renaming it to `sendPreparedTransactions`?

src/send/saga.ts
```diff
@@ -303,14 +287,72 @@ export function* sendPaymentSaga({
...
+  let nonce: number = yield* call(getTransactionCount, wallet, {
+    address: wallet.account.address,
+    blockTag: 'pending',
+  })
+
+  const preparedTransactions = getPreparedTransactions(serializablePreparedTransactions)
+  const txHashes: Hash[] = []
+  for (let i = 0; i < preparedTransactions.length; i++) {
+    const preparedTransaction = preparedTransactions[i]
+    const createBaseStandbyTransaction = createBaseStandbyTransactions[i]
```
jeanregisser, line 322: We should check `createBaseStandbyTransaction` is not undefined and throw.
To prevent incorrect use of the saga.

This reminds maybe we should consider enabling [noUncheckedIndexedAccess](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess).
I'll try to bring this up at the next eng meeting.

src/swap/saga.ts
```diff
@@ -62,12 +60,11 @@ function calculateEstimatedUsdValue({
 
 interface TrackedTx {
   tx: TransactionRequest | undefined
-  txHash: Hash | undefined
   txReceipt: TransactionReceipt | undefined
 }
 
 function getTxReceiptAnalyticsProperties(
-  { tx, txHash, txReceipt }: TrackedTx,
+  { tx, txReceipt }: TrackedTx,
```
jeanregisser, line 67: The reason I used `txHash` instead `txReceipt.transactionHash` is in case we fail to get the receipt. Could we keep it this way?

jeanregisser (approved):
Love to see this!! 🙌

src/swap/saga.ts
```diff
@@ -299,74 +268,71 @@ export function* swapSubmitSaga(action: PayloadAction<SwapInfo>) {
...
-        Logger.debug('Got approve transaction receipt', approveTxReceipt)
-        trackedTxs[0].txReceipt = approveTxReceipt
-      }
-    } catch (e) {
-      Logger.warn(TAG, 'Error getting approve transaction receipt', e)
+    // wait for the tx receipts, so that we can track them
+    for (let i = 0; i < txHashes.length; i++) {
+      const txReceipt = yield* call([publicClient[network], 'waitForTransactionReceipt'], {
+        hash: txHashes[i],
+      })
```
jeanregisser, line 328: Should we do this in parallel to avoid unnecessary wait?
Well now that I'm writing this, this probably doesn't matter since there's no user facing wait here.

src/send/saga.ts
kathaypacific: yes good ideas! i renamed and moved to the viem folder. also added some jsdocs for the saga input params!

src/swap/saga.ts
kathaypacific: ah right, makes sense. i've reverted this change

src/swap/saga.ts
kathaypacific: yeah i'd prefer to leave it like this because it's a background task and easier to read this way

PR #5068, opened 2024-03-11
feat(txs): Show Approval TXs only on supported chains

jophish:
### Description

For [ACT-1081](https://linear.app/valora/issue/ACT-1081/use-showapprovaltxsinhomefeed-from-dynamic-config-in-wallet). Adds an additional filter to the cached TXs selector to filter out Approval TXs that aren't supported via Statsig.

### Test plan

Unit tested.

### Related issues

- Fixes [ACT-1081](https://linear.app/valora/issue/ACT-1081/use-showapprovaltxsinhomefeed-from-dynamic-config-in-wallet).

### Backwards compatibility

Yes.

### Network scalability


jh2oman (approved):

jh2oman:
+1 after other test is updated

kathaypacific (commented):
sorry i'm missing some context here but is this change different to the dynamic config that already exists for this? https://github.com/valora-inc/wallet/blob/jumpstart-feed-items/src/statsig/constants.ts#L76

PR #5069, opened 2024-03-11
feat(bottom-tabs): extend wallet home to support home tab

satish-ravi:
### Description

Updating WalletHome component to take a prop denoting whether we're in the tab view, and if so display appropriate components.

### Test plan

Unit tests, manual

https://github.com/valora-inc/wallet/assets/5062591/eb19a647-783c-42ca-91e1-878fe0cd1ac2

Couple of things out of scope for this PR
- Header - will be done in a separate issue
- Horizontal padding - inconsistent between different portions of the home screen, will be in a follow up PR
- Notification bell spotlight - disabled for now for tab view, confirming with product that if we need to carry it forward

### Related issues

- Fixes ACT-1105

### Backwards compatibility

Yes

### Network scalability


jh2oman (approved):
Surprisingly simple!

PR #5071, opened 2024-03-11
feat(home): remove cash in bottom sheet

satish-ravi:
### Description

Confirmed with product that this is no longer required.

### Test plan

Unit tests, CI

### Related issues

- Part of ACT-1105

### Backwards compatibility

Yes

### Network scalability

N/A


MuckT (approved):
I think `cashInBottomSheetEnabled` should also be removed from `src/statsig/constants.ts`, but otherwise LGTM.

satish-ravi:
> I think `cashInBottomSheetEnabled` should also be removed from `src/statsig/constants.ts`, but otherwise LGTM.

I left it as is as it was one of the experiment params, we can clean it up when we remove the experiment.

PR #5074, opened 2024-03-12
feat(wallet-tab): Add Hide Balance ability to wallet tab

jh2oman:
### Description

Generalizes the `hideHomeBalances` state into `hideBalances` and updates the `hideHomeBalancesSelector` and `hideWalletBalancesSelector` to use the statsig `USE_TAB_NAVIGATOR` feature flag.

### Video

In the video you can see me switch between the current experience and the new tab experience showing how the changes affect both home and wallet/assets.
 

https://github.com/valora-inc/wallet/assets/8432644/287fca7b-f2e9-4fac-ade0-12f07067ade1

### Test plan

updated unit tests

### Related issues

https://linear.app/valora/issue/ACT-1109/hide-token-balance


src/components/TokenBalance.tsx
```diff
@@ -223,7 +222,11 @@ export function AssetsTokenBalance({
             </TouchableOpacity>
           )}
         </View>
-        <TokenBalance style={styles.totalBalance} singleTokenViewEnabled={false} />
+        <TokenBalance
+          style={styles.totalBalance}
+          singleTokenViewEnabled={false}
+          showBalanceToggle={true}
```
satish-ravi, line 228:
```suggestion
          showBalanceToggle={isWalletTab}
```
otherwise this will show up on the old Assets page

src/tokens/TokenBalanceItem.tsx
```diff
@@ -28,6 +30,8 @@ export const TokenBalanceItem = ({
 }: Props) => {
   const { t } = useTranslation()
 
+  const hideWalletBalance = useSelector(hideWalletBalancesSelector)
```
satish-ravi, line 33: This would end up hiding the token balance everywhere this used, e.g., send/swap/cico token bottom sheets, old Assets screen, etc. We only want to hide this on the wallet tab.

src/transactions/feed/TransferFeedItem.tsx
```diff
@@ -41,8 +39,7 @@ function TransferFeedItem({ transfer }: Props) {
   const colorStyle = new BigNumber(amount.value).isPositive() ? { color: colors.primary } : {}
 
   const hideHomeBalanceState = useSelector(hideHomeBalancesSelector)
-  const hideBalance =
-    getFeatureGate(StatsigFeatureGates.SHOW_HIDE_HOME_BALANCES_TOGGLE) && hideHomeBalanceState
+  const hideBalance = hideHomeBalanceState
```
satish-ravi, line 42: nit: do you need both the variables?

src/transactions/feed/SwapFeedItem.tsx
```diff
@@ -34,8 +32,7 @@ function SwapFeedItem({ exchange }: Props) {
   }
 
   const hideHomeBalanceState = useSelector(hideHomeBalancesSelector)
-  const hideBalance =
-    getFeatureGate(StatsigFeatureGates.SHOW_HIDE_HOME_BALANCES_TOGGLE) && hideHomeBalanceState
+  const hideBalance = hideHomeBalanceState
```
satish-ravi, line 35: nit: do you need both the variables?

src/components/TokenBalance.tsx
```diff
@@ -78,8 +76,9 @@ function TokenBalance({
   const { decimalSeparator } = getNumberFormatSettings()
 
   const hideHomeBalanceState = useSelector(hideHomeBalancesSelector)
-  const hideBalance = showHideHomeBalancesToggle && hideHomeBalanceState
-  const balanceDisplay = hideBalance ? `XX${decimalSeparator}XX` : totalBalanceLocal?.toFormat(2)
+  const hideWalletBalance = useSelector(hideWalletBalancesSelector)
+  const hideBalance = showBalanceToggle && (hideHomeBalanceState || hideWalletBalance)
+  const balanceDisplay = hideBalance ? `Xx${decimalSeparator}xx` : totalBalanceLocal?.toFormat(2)
```
satish-ravi, line 81: Left a question on figma on whether the Xx.xx change was actually intended. https://www.figma.com/file/EkJ8MVKoc6QVrrdKYFBARw?node-id=2722:1444&mode=design#732563880

satish-ravi (changes requested):
We should hide balances on PositionItem as well since its part of the wallet tab

src/tokens/TokenBalanceItem.tsx
jh2oman: oof thanks for that!

src/tokens/AssetList.tsx
```diff
@@ -100,6 +101,8 @@ export default function AssetList({
     sortedTokensWithBalanceOrShowZeroBalanceSelector(state, supportedNetworkIds)
   )
 
+  const hideWalletBalances = useSelector(hideWalletBalancesSelector)
```
satish-ravi, line 104:
```suggestion
  const hideWalletBalances = useSelector(hideWalletBalancesSelector) && isWalletTab
```
we could do this to be extra safe, but in practice should never make a difference as long as the old assets screen cannot be accessed when the tab navigator is on.

src/transactions/feed/SwapFeedItem.test.tsx
```diff
@@ -124,8 +125,10 @@ describe('SwapFeedItem', () => {
     expect(getElementText(getByTestId('SwapFeedItem/outgoingAmount'))).toEqual('-17.54 cEUR')
   })
 
-  it('still shows balances when feature gate false, hide balances root state true', async () => {
-    jest.mocked(getFeatureGate).mockReturnValue(false)
+  it('still shows balances when feature gate true, hide balances root state true', async () => {
+    jest
+      .mocked(getFeatureGate)
+      .mockImplementation((featureGate) => featureGate === StatsigFeatureGates.USE_TAB_NAVIGATOR)
```
satish-ravi, line 131: nit: are there multiple feature gates accessed in the component under test? If not could we just do mockReturnValue?

src/transactions/feed/SwapFeedItem.test.tsx
```diff
@@ -124,8 +125,10 @@ describe('SwapFeedItem', () => {
     expect(getElementText(getByTestId('SwapFeedItem/outgoingAmount'))).toEqual('-17.54 cEUR')
   })
 
-  it('still shows balances when feature gate false, hide balances root state true', async () => {
-    jest.mocked(getFeatureGate).mockReturnValue(false)
+  it('still shows balances when feature gate true, hide balances root state true', async () => {
```
satish-ravi, line 128: nit: might be cleaner to include the feature gate name on the test, as it seems to depend on a different one now than before

src/transactions/feed/TransferFeedItem.test.tsx
```diff
@@ -627,15 +628,20 @@ describe('TransferFeedItem', () => {
     })
   })
 
-  it('shows balance when feature gate false, root state hide home balances flag is set', async () => {
-    jest.mocked(getFeatureGate).mockReturnValue(false)
-    const { getByTestId } = renderScreen({ storeOverrides: { app: { hideHomeBalances: true } } })
+  it('shows balance when feature gate true, root state hide home balances flag is set', async () => {
```
satish-ravi, line 631: same nits as above

satish-ravi (approved):

src/transactions/feed/SwapFeedItem.test.tsx
jh2oman: I know we could, but I saw this in another test and actually like it because its more explicit. just mocking all getFeatureGate calls could lead to some inaccurate tests.

satish-ravi: I think its not too different, especially the `false` case because it sets the gate in question to `false` and every other gate to true. This pattern is just inverting what other gates return.

PR #5076, opened 2024-03-12
fix: the NFT reward bottom sheet unexpectedly reappears after CTA button is pressed

bakoushin:
### Description

On the iOS production build the NFT reward bottom sheet unexpectedly reappears after CTA button is pressed.
The hypothesis is that `onChange` event isn't fired for some reason when the bottom sheet is closed programmatically.

The fix is to fire `dispatch` on CTA press handler instead of relying on `onChange` event.

Another possible option is to rely on `onClose` event. But unfortunately it is called when component unmounts on Android.

### Test plan

Tested manually on:
* Android emulator
* iOS simulator
* dev iOS build on device

Planing to additionally test on the nightly iOS build.

### Related issues

- Related to  RET-1002

### Backwards compatibility

Y

### Network scalability

NA

### DIFF
--- a/src/home/celebration/NftReward.tsx
+++ b/src/home/celebration/NftReward.tsx
@@ -71,22 +71,20 @@ export default function NftRewardBottomSheet() {
     }
   }, [isVisible])
 
-  const handleBottomSheetPositionChange = (index: number) => {
+  const handleBottomSheetClose = () => {
     if (!nftCelebration) {
       return // This should never happen
     }
 
-    if (index === -1) {
-      if (!rewardAccepted) {
-        ValoraAnalytics.track(HomeEvents.nft_reward_dismiss, {
-          networkId: nftCelebration.networkId,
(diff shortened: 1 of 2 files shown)


jophish (approved):

src/home/celebration/NftReward.tsx
```diff
@@ -83,9 +83,9 @@ export default function NftRewardBottomSheet() {
           contractAddress: nftCelebration.contractAddress,
           remainingDays: differenceInDays(rewardExpirationDate, Date.now()),
         })
-      }
 
-      dispatch(nftRewardDisplayed())
+        dispatch(nftRewardDisplayed())
```
jophish, line 87: can this be easily tested? looks like patch coverage is failing

bakoushin: unfortunately I don't know the way to simulate bottom sheet close.
the only way I see is to trick coverage bot for now.

PR #5078, opened 2024-03-12
feat: discover tab screen

MuckT:
### Description

- Adds the discover tab screen.
- Makes the screen padding 16px to both variants.

| iOS Enabled | iOS Default |
| ----- | ----- | 
|  |  |

### Test plan

- Tested locally on iOS
- Unit tests updated

### Related issues

- Fixes ACT-1106

### Backwards compatibility

Yes

### Network scalability

N/A


src/dappsExplorer/DAppsExplorerScreenSearchFilter.test.tsx
```diff
@@ -39,670 +40,680 @@ const defaultStore = createMockStore({
...
-  })
-
-  it('pens dapps directly', () => {
-    const store = createMockStore({
-      dapps: {
-        dappListApiUrl: 'http://url.com',
-        dappsList,
-        dappsCategories,
-      },
+describe.each([{ isTabNavigator: false }, { isTabNavigator: true }, {}])(
```
MuckT, line 43: Tests were wrapped in this describe for each possible route param.

src/dappsExplorer/DAppsExplorerScreenSearchFilter.test.tsx
```diff
@@ -39,670 +40,680 @@ const defaultStore = createMockStore({
...
+        <Provider store={defaultStore}>
+          <MockedNavigator component={DAppsExplorerScreenSearchFilter} params={routeParams} />
         </Provider>
       )
 
-      expect(queryByText('dappsScreen.noFavorites.title')).toBeNull()
-      expect(queryByText('dappsScreen.noFavorites.description')).toBeNull()
+      // TODO (ACT-1108): should be true in all cases
+      // Shows the QR button
+      if (!routeParams.isTabNavigator) {
```
MuckT, line 60: This is a bit hacky, but will be true in all cases once act-1108 is created and can be removed then.

src/navigator/types.tsx
```diff
@@ -272,7 +272,7 @@ export type StackParamList = {
       }
     | undefined
   [Screens.SwapScreenWithBack]: { fromTokenId: string } | undefined
-  [Screens.TabDiscover]: undefined
+  [Screens.TabDiscover]: { isTabNavigator?: boolean } | undefined
```
finnian0826, line 275: Why does this have to be the union of the param that gets used and undefined? Do screens that use the same component need to have the same params? It seems like `isTabNavigator` is always used for the TabDiscover screen, and there are no props passed in ever for DAppsExplorerScreen, and then the component takes the union of these for `Props`

MuckT: That is a great question. I was following the pattern used in https://github.com/valora-inc/wallet/pull/5069, but I'll check with @satish-ravi as to why this was done.

satish-ravi: In this specific case since the options are `undefined` or an optional boolean parameter, it should be fine to have their own params. But if there are different params, there will be a type error when accessing params in the component. E.g., if props are `{a: string}` for one screen and `{b: string}` for another, the type of route.params will be `{a: string } | {b: string}` and TS won't let you do `route.params.a` or `route.params.b`. In any case this is only temporary while we switch to the tab navigator, we should generally not have screens sharing components

satish-ravi (approved):

PR #5082, opened 2024-03-13
feat: add tab bar header navigation

MuckT:
### Description

Adds the top navigation for tab navigation screens. In the screen shots below, the header is the main point of focus as different wallets were used to take enabled and disabled screenshots. Header icons were adjusted to allow for ripples on tap on Android and a few changes had to be made to keep alignment with the existing side `DrawerTopBar.tsx`.

#### Screenshots

| Android Disabled | Android Enabled | iOS Disabled | iOS Enabled |
|  |  |  |  | 

### Test plan

- [x] Tested locally on iOS
- [x] Unit tests added

### Related issues

- Fixes #ACT-1108

### Backwards compatibility

Yes

### Network scalability

N/A


src/components/AccountCircleButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export default function AccountCircleButton({ style, size, testID }: Props) {
+  const onPress = () => {
+    navigate(Screens.ProfileMenu)
+  }
+
+  return (
+    <View style={styles.container}>
+      <Touchable
```
MuckT, line 22: Touchable used here and in other added icons instead of `TopBarButton.tsx` to better handle ripple on Android.

src/escrow/EscrowedPaymentListScreen.tsx
```diff
@@ -24,9 +23,12 @@ export default function EscrowedPaymentListScreen() {
   return <NotificationList items={sentEscrowedPayments} listItemRenderer={listItemRenderer} />
 }
 
-EscrowedPaymentListScreen.navigationOptions = titleWithBalanceNavigationOptions(
-  i18n.t('escrowedPaymentReminder')
-)
+EscrowedPaymentListScreen.navigationOptions = () => ({
```
MuckT, line 26: Changed due to a failing test case in 2aeef881a4aa3947622bf09e450896055a154b96.

src/home/WalletHome.tsx
```diff
@@ -96,6 +98,27 @@ function WalletHome({ route }: Props) {
     )
   }
 
+  // Scroll Aware Header
```
MuckT, line 101: It might be worth abstracting this as we are now using the same pattern in two places.

src/notifications/NotificationList.tsx
```diff
@@ -33,13 +31,6 @@ export function NotificationList<T>(props: Props<T>) {
   )
 }
 
-export function titleWithBalanceNavigationOptions(title: string) {
```
MuckT, line 36: Unused after 2aeef881a4aa3947622bf09e450896055a154b96.

src/tokens/Assets.tsx
```diff
@@ -205,7 +205,7 @@ function AssetsScreen({ navigation, route }: Props) {
   }
 
   return (
-    <>
+    <Animated.View>
```
MuckT, line 208: A weird edge case case were an animated view wrapped in a fragment can cause transparency issues.

src/tokens/Assets.tsx
```diff
@@ -266,8 +267,8 @@ AssetsScreen.navigationOptions = {
 const styles = StyleSheet.create({
   listHeaderContainer: {
     ...getShadowStyle(Shadow.SoftLight),
-    paddingHorizontal: Spacing.Thick24,
-    paddingTop: Spacing.Smallest8,
+    paddingHorizontal: Spacing.Regular16,
```
satish-ravi, line 270: should we keep these at 24 for now, since the token balance items are spaced at 24?

src/navigator/TabNavigator.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { createMockStore } from 'test/utils'
+
+jest.mock('src/statsig')
+
+describe('TabNavigator', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+    jest
+      .mocked(getFeatureGate)
+      .mockImplementation((featureGate) => featureGate !== StatsigFeatureGates.USE_TAB_NAVIGATOR)
```
satish-ravi, line 19: personally not a fan of this pattern as it sets every other gate to true, but I guess this is good enough for documenting what gate we really care about in these tests.

src/navigator/Headers.tsx
```diff
@@ -244,10 +262,12 @@ export function HeaderTitleWithTokenBalance({
 
 export function HeaderTitleWithSubtitle({
   title,
+  titleStyle,
```
satish-ravi, line 265: wonder if we should just update this universally instead of a prop. Maybe worth checking with design?

src/components/AccountCircleButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+interface Props {
+  style?: StyleProp<ViewStyle>
+  size?: number
+  testID?: string
+}
+
+export default function AccountCircleButton({ style, size, testID }: Props) {
+  const onPress = () => {
+    navigate(Screens.ProfileMenu)
```
satish-ravi, line 17: should we add an analytics event for this?

src/components/AccountCircleButton.tsx
satish-ravi: is this something TopBarIconButton can support? if not, is there a purpose of keeping TopBarIconButton around?

src/navigator/TabNavigator.tsx
```diff
@@ -47,6 +48,8 @@ export default function TabNavigator({ route }: Props) {
         options={{
           tabBarLabel: t('bottomTabsNavigator.wallet.tabName') as string,
           tabBarIcon: Wallet,
+          tabBarTestID: 'Tab/Wallet',
+          ...(tabHeader as NativeStackHeaderProps),
```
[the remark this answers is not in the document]
MuckT: Moved to screenOptions in d5cc11c5732aad5f61c8315e33da00f98195517b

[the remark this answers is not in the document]
MuckT: All of our headers using the `NativeStackNavigationOptions`, but `screenOptions` in `Tab.Navigator` expects a type of `NativeStackHeaderProps`. I believe this is due with our nested navigators, but it works in all the cases I tested. I tried avoiding the type casting and using a ts-expect-error, but that wasn't working.

src/dappsExplorer/DAppsExplorerScreenSearchFilter.tsx
```diff
@@ -136,6 +145,28 @@ export function DAppsExplorerScreenSearchFilter({ route }: Props) {
...
+  // Scroll Aware Header
+  const scrollPosition = useSharedValue(0)
+  const [titleHeight, setTitleHeight] = useState(0)
+
+  const handleMeasureTitleHeight = (event: LayoutChangeEvent) => {
+    setTitleHeight(event.nativeEvent.layout.height)
+  }
+
+  const handleScroll = useAnimatedScrollHandler((event) => {
+    Animated.event([{ nativeEvent: { contentOffset: { y: scrollPositionValue } } }])
```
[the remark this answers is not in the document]
MuckT: Moved to calling onScroll in e2846b7e0f2773d7e96efbf620aae6af3eb48166 and I no longer see the error.

src/dappsExplorer/DAppsExplorerScreenSearchFilter.tsx
```diff
@@ -31,10 +39,11 @@ import { currentLanguageSelector } from 'src/i18n/selectors'
 import DrawerTopBar from 'src/navigator/DrawerTopBar'
 import { styles as headerStyles } from 'src/navigator/Headers'
 import { Screens } from 'src/navigator/Screens'
+import useScrollAwareHeader from 'src/navigator/ScrollAwareHeader'
 import { StackParamList } from 'src/navigator/types'
 import { useDispatch, useSelector } from 'src/redux/hooks'
-import colors from 'src/styles/colors'
-import fontStyles from 'src/styles/fonts'
+import colors, { Colors } from 'src/styles/colors'
```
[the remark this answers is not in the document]
MuckT: Switched to using the named export in 74a55b5b27671f11c04a104b01d0842607486502.

src/home/WalletHome.tsx
```diff
@@ -96,6 +98,28 @@ function WalletHome({ route }: Props) {
...
+  const scrollPosition = useSharedValue(0)
+  const [titleHeight, setTitleHeight] = useState(0)
+
+  const handleMeasureTitleHeight = (event: LayoutChangeEvent) => {
+    setTitleHeight(event.nativeEvent.layout.height)
+  }
+
+  const handleScroll = useAnimatedScrollHandler((event) => {
+    Animated.event([{ nativeEvent: { contentOffset: { y: scrollPositionValue } } }])
```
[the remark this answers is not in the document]
MuckT: Moved to calling onScroll in https://github.com/valora-inc/wallet/commit/e2846b7e0f2773d7e96efbf620aae6af3eb48166 and I no longer see the error.

src/navigator/TabNavigator.test.tsx
MuckT: I am open to other suggestions. It looks like mockImplementation always expects a boolean value returned so it's a bit tricky in that regard.

src/components/AccountCircleButton.tsx
MuckT: Added in c4cf8fe45442b62faaf8bbba250a11a5a5f8d0ad!

src/navigator/Headers.tsx
MuckT: Made it an optional prop to avoid impact on any of the existing screens. Currently this is used by the `useScrollAwareHeader.tsx`.

src/tokens/Assets.tsx
MuckT: I think the header being different from the other tabs is more noticeable the the spacing below. We could do this conditionally for the bottom tab.

MuckT: Made it conditional base on the new tab view in 54e3825fe5a00cc63d3ddd37096625fe6f3e18f7. The difference in header placement was a bit to noticeable in my opinion. Once we have a follow up PR to adjust padding we can remove the non tab navigator path.

MuckT:
@satish-ravi in regards to keeping the `TopBarIconButton.tsx` it was used in too many other places to justify removing it in this PR. The main reason not to use is its lack of handing Android ripples. There are some notes in `src/components/BackButton.tsx` from @jeanregisser about what it took to get the ripple correct and it requires placement specific styling and seems like a general hassle.  In my opinion we should create a better Icon wrapper and deprecate `TopBarIconButton.tsx` in a separate PR / ticket.

src/navigator/Headers.tsx
satish-ravi: I was actually proposing changing every screen to the new style. Seems low risk to me, and avoids another style prop which has to be passed down 2 layers deep, and this would be something that is never cleaned up.

src/home/WalletHome.tsx
```diff
@@ -96,6 +98,28 @@ function WalletHome({ route }: Props) {
...
+  // Scroll Aware Header
+  const scrollPosition = useSharedValue(0)
+  const [titleHeight, setTitleHeight] = useState(0)
+
+  const handleMeasureTitleHeight = (event: LayoutChangeEvent) => {
+    setTitleHeight(event.nativeEvent.layout.height)
+  }
+
+  const handleScroll = useAnimatedScrollHandler((event) => {
+    onScroll
```
satish-ravi, line 110: Is this statement doing anything? `onScroll` I believe is a function, so this seems like just a line that references the function and doesn't invoke it.
Also, I think you should be able to use a single event handler instead of the 2 different handlers based on the gate. I believe both are just extracting `event.contentOffset.y` and setting it into a variable. `onScroll` sets it on the `scrollPositionValue` (of type `Animated.Value`) and `handleScroll` sets it on `scrollPosition` (of type `SharedValue`). We could just make all consumers use the SharedValue or just have a single handler that sets both variables.

satish-ravi:
> @satish-ravi in regards to keeping the `TopBarIconButton.tsx` it was used in too many other places to justify removing it in this PR. The main reason not to use is its lack of handing Android ripples. There are some notes in `src/components/BackButton.tsx` from @jeanregisser about what it took to get the ripple correct and it requires placement specific styling and seems like a general hassle. In my opinion we should create a better Icon wrapper and deprecate `TopBarIconButton.tsx` in a separate PR / ticket.

that sounds good. Can we mark this deprecated now and link to a placeholder issue so no new stuff is added that uses this?

src/home/WalletHome.tsx
MuckT: I tried conditionally setting the scroll in fc03d3e4ceb26b0698b73bb54a63d3c5b7d279e9. It looks like the other scroll is only being used by the drawer header so cleaning this up might be easier once that is removed. Perhaps a ticket to clean-up the drawer header and add todo's linked to that?

src/navigator/Headers.tsx
MuckT: Changed all the title styles in beb06fe87c70646fed7c43813ae32436c6aaf06d. A quick look through the app didn't reveal any issues with this change on both large and default fonts.

satish-ravi (approved):

PR #5089, opened 2024-03-13
chore: inline notification visual tweaks

bakoushin:
### Description

Visual tweaks to accommodate current [design variants](https://www.figma.com/file/erFfzHvSTm5g1sjK6jWyEH/Working-Design-System-Components?type=design&node-id=1144-35&mode=design&t=kUIyzJiXtW1ebQAL-0):
* added `Success` variant (green background)
* added icon override: can provide alternative icon or `null` to render without one

Renames:
* `Severity` -> `NotificationVariant` because from my standpoint it's better aligned with all options it contains
* `Informational` -> `Info` just because it's shorter

### Test plan

* Added unit test for icon override 

### Related issues

- Related to RET-1004

### Backwards compatibility

Y

### Network scalability

NA

### DIFF
--- a/src/account/Profile.tsx
+++ b/src/account/Profile.tsx
@@ -12,7 +12,7 @@ import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
 import { ErrorMessages } from 'src/app/ErrorMessages'
 import Button from 'src/components/Button'
 import CancelButton from 'src/components/CancelButton'
-import InLineNotification, { Variant } from 'src/components/InLineNotification'
+import InLineNotification, { NotificationVariant } from 'src/components/InLineNotification'
 import KeyboardAwareScrollView from 'src/components/KeyboardAwareScrollView'
 import KeyboardSpacer from 'src/components/KeyboardSpacer'
 import TextInput from 'src/components/TextInput'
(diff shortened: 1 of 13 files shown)


src/account/Profile.tsx
```diff
@@ -110,7 +110,7 @@ function Profile({ navigation, route }: Props) {
         </View>
         <View style={styles.disclaimerContainer}>
           <InLineNotification
-            severity={Severity.Informational}
+            variant={Variant.Info}
```
kathaypacific, line 113:
```suggestion
            variant={NotificationVariant.Info}
```
maybe? i'm not sure, i wonder if the generic names are confusing. we haven't been terribly consistent with this but the old "Severity" name bothered me. for the button enums, we prefix with "Btn" so it'd be nice to do something similar here?

src/components/InLineNotification.tsx
```diff
@@ -52,17 +55,21 @@ export function InLineNotification({
...
     <View
-      style={[styles.container, { backgroundColor: severityColor.secondary }, style]}
+      style={[styles.container, { backgroundColor: variantColor.secondary }, style]}
       testID={testID}
     >
       <View style={styles.row}>
-        <View style={styles.attentionIcon}>
-          <Icon color={severityColor.primary} size={20} />
-        </View>
+        {icon !== null && (
```
kathaypacific, line 66: nit: it seems like we're using `icon` to either provide a custom icon _or_ hide the default icon, to reduce surprises and implicit meanings in props, could we split this into 2 props? hideIcon and customIcon maybe?

kathaypacific (approved):
💅🏼

PR #5090, opened 2024-03-13
fix(jumpstart): ensure that jumpstart transactions are prepared correctly

kathaypacific:
### Description

We were using an incorrect address for the jumpstart contract previously so didn't catch these issues. This PR:
- fixes some approval transactions not added to the prepared transactions. the enter amount screen was passing the user input token amount to prepare transactions, however it needs to be the token amount in smallest units. the incorrect units were causing the approval transaction to never be added because the spend amount would get rounded to 0 for entered values under 1.
- add dynamic config for the transaction gas estimate, since we cannot estimate gas for a transaction without approval

### Test plan

No visual changes, manually verified that the transactions created are using the expected values.

### Related issues

- Related to RET-994

### Backwards compatibility

Y

### Network scalability

Y

### DIFF
--- a/src/transactions/saga.ts
+++ b/src/transactions/saga.ts
@@ -256,7 +256,7 @@ export function* getTransactionReceipt(
   const { feeCurrencyId, transactionHash } = transaction
 
   try {
-    const receipt = yield* call([publicClient[network], 'getTransactionReceipt'], {
+    const receipt = yield* call([publicClient[network], 'waitForTransactionReceipt'], {
       hash: transactionHash as Hash,
     })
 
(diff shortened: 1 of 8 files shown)


src/transactions/saga.ts
```diff
@@ -256,7 +256,7 @@ export function* getTransactionReceipt(
   const { feeCurrencyId, transactionHash } = transaction
 
   try {
-    const receipt = yield* call([publicClient[network], 'getTransactionReceipt'], {
+    const receipt = yield* call([publicClient[network], 'waitForTransactionReceipt'], {
```
kathaypacific, line 259: this is kind of unrelated but since we are polling for the tx receipt, it's expected that at the beginning the receipt is not there. this change prevents the catch block from being hit unnecessarily

jophish (approved):
LGTM I think (not super up to speed on the jumpstart work though) -- just so I understand correctly, before this change, if a user entered `n` to send, we would only ever approve `⌊n⌋` wei, which means the user on the other end could only ever claim that much, right?

PR #5093, opened 2024-03-14
feat(positions): multichain positions

cajubelt:
### Description

gets positions and shortcuts for whichever networks are enabled in our dynamic config, rather than just celo

relies on https://github.com/valora-inc/hooks/pull/362

### Test plan

- ci
- manual test with local hooks api on wallet with both ethereum and celo positions (also unearthed a redux migration bug with this! hooray for testing!)

### Related issues

fixes https://linear.app/valora/issue/RET-941/%5Bhooks%5D-multichain-support

### Backwards compatibility

there is only one place where the new API will affect application code for old app versions, which is in the `keyExtractor` function in `AssetList.tsx`. The function returns the following for positions: `${activeTab}-${item.appId}-${item.networkId}-${item.address}-${index}`. I think this is ok, since old app versions only get celo positions right now, so when `item.networkId` evaluates to `undefined` for them, it shouldn't cause collisions with other positions on other networks. 

the other changes are all in analytics events, which should be something we can live with

### Network scalability


MuckT:
Looks like there's a schema update needed and some tests that also need to be updated in:
- `src/tokens/PositionItem.test.tsx`
- `src/account/SupportContact.test.tsx`
- `src/account/zendesk.test.ts`
- `src/analytics/selectors.test.ts`

I also see one case where we use `celoNetwork` in `src/analytics/ValoraAnalytics.test.ts` should this also be updated?

src/analytics/Properties.tsx
cajubelt, line 1: for analytics types I kept the "network" key the same, to avoid adding a column to BQ tables. for application code I migrated to "networkId" for consistency with other features

silasbw (approved):

jeanregisser (approved):
Awesome 🚀

PR #5103, opened 2024-03-14
feat(jumpstart): add error state on failed transaction

kathaypacific:
### Description

As the title

### Test plan

https://github.com/valora-inc/wallet/assets/20150449/53fee34f-d9aa-4e08-a00e-b48d349f1d2c

### Related issues

- Fixes RET-994

### Backwards compatibility

Y

### Network scalability

Y


src/components/BottomSheetInLineNotification.tsx
```diff
@@ -54,35 +57,43 @@ const BottomSheetInLineNotification = ({
...
   background: {
     backgroundColor: Colors.black,
   },
   notificationContainer: {
     position: 'absolute',
-    bottom: Spacing.Large32,
-    width: '100%',
+    bottom: Spacing.XLarge48,
+    left: Spacing.Regular16,
+    width: variables.width - Spacing.Large32,
```
jeanregisser, line 96: nit: would this work?
```suggestion
    left: Spacing.Regular16,
    right: Spacing.Regular16,
```

locales/base/translation.json
```diff
@@ -2115,7 +2115,12 @@
     "sendAmountLabel": "CLAIM LINK AMOUNT",
     "confirmButton": "Send funds and create link",
     "detailsLabel": "HOW IT WORKS",
-    "details": "1. Tap generate link & your funds are sent into the claim link\n2. Send the claim link to a friend\n3. They’ll accept the funds\n4. If funds aren’t claimed, you can reclaim them anytime."
+    "details": "1. Tap generate link & your funds are sent into the claim link\n2. Send the claim link to a friend\n3. They’ll accept the funds\n4. If funds aren’t claimed, you can reclaim them anytime.",
+    "error": {
```
jeanregisser, line 2119: nit: maybe we can be slightly more specific?
```suggestion
    "sendError": {
```

jeanregisser (approved):
Looks great!

src/jumpstart/JumpstartSendConfirmation.tsx
```diff
@@ -25,17 +31,37 @@ type Props = NativeStackScreenProps<StackParamList, Screens.JumpstartSendConfirm
...
   const parsedAmount = new BigNumber(sendAmount)
   const { t } = useTranslation()
+  const dispatch = useDispatch()
 
   const token = useTokenInfo(tokenId)
   const usdToLocalRate = useSelector(usdToLocalCurrencyRateSelector)
   const localCurrencyCode = useSelector(getLocalCurrencyCode)
+  const jumpstartSendStatus = useSelector(jumpstartSendStatusSelector)
+
+  const [showError, setShowError] = useState(false)
```
bakoushin, line 44: what's the reason of having this local state instead of relying on redux one?

kathaypacific: it's so that you can dismiss the notification without re-submitting the transaction. the alternative is we can create another action that changes the redux status back to 'idle'...

src/components/BottomSheetInLineNotification.tsx
kathaypacific: i think so, but i ended up using the new Toast component after it was merged so didn't need changes to this file after all :)

src/jumpstart/JumpstartSendConfirmation.tsx
kathaypacific: no, we only need it for displaying the error indeed. i've added a new redux action, it's indeed cleaner to rely on one place for the source of truth

PR #5104, opened 2024-03-14
chore(jumpstart): update copy and design tweaks

kathaypacific:
### Description

There were some copy and design changes, designs [here](https://www.figma.com/file/6sJ4fCxCptTybHB6ZZ7Pqi/Escrow-Feature?type=design&node-id=2020-8072&mode=design&t=FE8oYsVh1PkdMhnc-0).

The key changes in this PR:
1. update copy to the designs
2. replace "how it works" section with info inline notification on the send confirmation screen

### Test plan

n/a

### Related issues

- Fixes RET-994

### Backwards compatibility

Y

### Network scalability

Y


src/components/InLineNotification.tsx
```diff
@@ -68,7 +68,7 @@ export function InLineNotification({
         {!hideIcon && (
           <View style={styles.iconContainer}>
             {customIcon ?? (
-              <Icon color={variantColor.primary} size={20} testId="InLineNotification/Icon" />
+              <Icon color={variantColor.primary} size={16} testId="InLineNotification/Icon" />
```
kathaypacific, line 71: i think the icon size should always be 16 according to the [design system?](https://www.figma.com/file/erFfzHvSTm5g1sjK6jWyEH/Working-Design-System-Components?type=design&node-id=493-4&mode=design&t=mb5pNPoDPqSiXIKZ-0)

src/jumpstart/JumpstartEnterAmount.tsx
```diff
@@ -170,11 +172,13 @@ function JumpstartEnterAmount() {
...
           variant={NotificationVariant.Warning}
-          title={t('jumpstartEnterAmountScreen.maxAmountWarning.title')}
-          description={t('jumpstartEnterAmountScreen.maxAmountWarning.description', {
-            maxAmount: maxSendAmountLocalCurrency?.toFormat(2),
-            localCurrencyCode,
+          title={t('jumpstartEnterAmountScreen.maxAmountWarning.title', {
+            amountInLocalCurrency: maxSendAmountLocalCurrency,
+            formatParams: {
+              amountInLocalCurrency: { currency: localCurrencyCode, locale },
+            },
```
jeanregisser, line 179: oh interesting, is `formatParams` a built-in i18next feature?

jeanregisser: ah yes https://www.i18next.com/translation-function/formatting
And do we have the necessary `Intl` support working in both iOS and Android RN env?

jeanregisser (approved):
Awesom! 🚀

src/components/InLineNotification.tsx
bakoushin: ah, that's a good question!

some images are indeed 16, but some apparently are not.
in this [figma components library](https://www.figma.com/file/erFfzHvSTm5g1sjK6jWyEH/Working-Design-System-Components?type=design&node-id=311-3&mode=design&t=VzDUgt4lZH6YtD6z-0) every icon with the surrounding padding is 24 no matter of the actual image size.
notice that in notification designs the icons also have this padding around them.

my guess if we consider icons being 24 as a default approach, it should fit the designs nicely

bakoushin (approved):
💅

src/jumpstart/JumpstartEnterAmount.tsx
kathaypacific: i'll check android - on iOS it looks like it's working correctly!

kathaypacific: android is also working correctly - checked with currency set to euros, if the app language is en-US, then the currency format displayed is like "€100". if the language is de-DE, the format is "100 €"

src/jumpstart/SelectJumpstartRecipientButton.tsx
```diff
@@ -63,6 +64,7 @@ const styles = StyleSheet.create({
   body: {
     flexDirection: 'row',
     alignItems: 'center',
+    padding: Spacing.Regular16,
```
kathaypacific, line 67: the changes in this file is just moving containers around so that the android ripple on tap does not exceed the rounded green border

src/send/EnterAmount.tsx
```diff
@@ -477,8 +477,7 @@ const styles = StyleSheet.create({
     ...typeScale.labelSemiBoldMedium,
   },
   warning: {
-    marginTop: Spacing.Regular16,
-    marginBottom: Spacing.Smallest8,
+    marginBottom: Spacing.Regular16,
```
kathaypacific, line 480: also no visual changes here, this ensures that every element is applying their own bottom spaces, so that the jumpstart warning does not need to apply top spacing

src/components/InLineNotification.tsx
kathaypacific: i reverted this change, i figure it's something we can address holistically later

PR #5105, opened 2024-03-14
fix(jumpstart): use long dynamic link

kathaypacific:
### Description

As the title, we should use a long dynamic link to make sure that the links are not stored on a central server / valora should not have access to the links.

### Test plan

Tested that the deep links can be correctly consumed still

### Related issues

- Fixes RET-994

### Backwards compatibility

Y

### Network scalability

Y

### DIFF
--- a/src/firebase/dynamicLinks.ts
+++ b/src/firebase/dynamicLinks.ts
@@ -7,7 +7,7 @@ import {
 } from 'src/config'
 import { NetworkId } from 'src/transactions/types'
 
-const commentDynamicLinkParams: Omit<FirebaseDynamicLinksTypes.DynamicLinkParameters, 'link'> = {
+const commonDynamicLinkParams: Omit<FirebaseDynamicLinksTypes.DynamicLinkParameters, 'link'> = {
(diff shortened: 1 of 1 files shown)


src/firebase/dynamicLinks.ts
```diff
@@ -7,24 +7,27 @@ import {
...
-    domainUriPrefix: baseURI,
-    ios: {
-      appStoreId,
-      bundleId,
-    },
-    android: {
-      packageName: bundleId,
-    },
-  })
+const commentDynamicLinkParams: Omit<FirebaseDynamicLinksTypes.DynamicLinkParameters, 'link'> = {
```
dievazqu, line 10: nit: The name for this shouldn't be `commonDynamicLinkParams`? I don't understand where the `comment` comes from.

dievazqu (approved):

src/firebase/dynamicLinks.ts
kathaypacific: oops thank you! that was a typo 😅

PR #5111, opened 2024-03-15
feat(jumpstart): jumpstart claim status

bakoushin:
### Description

Display toast:
* when jumpstart claim is in progress
* when jumpstart claim failed

| Claim in progress | Claim failed |
|--------|--------|
|  Simulator Screenshot - iPhone 15 - 2024-03-15 at 17 09 44 | Simulator Screenshot - iPhone 15 - 2024-03-15 at 17 09 57 | 

### Test plan

 * Updated unit test

### Related issues

- Fixes RET-1004

### Backwards compatibility

Y

### Network scalability

NA


locales/base/translation.json
```diff
@@ -2185,5 +2185,17 @@
     "discover": {
       "tabName": "Discover"
     }
+  },
+  "jumpstartStatus": {
+    "loading": {
+      "title": "Your funds are on their way",
+      "description": "The funds that your friend sent you are being sent over using crypto magic. Please wait..."
```
kathaypacific, line 2192: lol can we update the copy? i was told that the copy was finalised in figma last week

src/jumpstart/JumpstartClaimStatusError.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const dispatch = useDispatch()
+  const showToast = useSelector(showJumstartError)
+
+  const handleDismiss = () => {
+    dispatch(jumpstartErrorDismissed())
+  }
+
+  const handleContactSupport = () => {
+    navigate(Screens.SupportContact)
+    dispatch(jumpstartErrorDismissed())
```
kathaypacific, line 22: can we add analytics for the button presses? (not sure if you planned to do that in another PR)

src/jumpstart/JumpstartClaimStatusToasts.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import JumpstartClaimStatusError from 'src/jumpstart/JumpstartClaimStatusError'
+import JumpstartClaimStatusLoading from 'src/jumpstart/JumpstartClaimStatusLoading'
+
+export default function JumpstartClaimStatusToasts() {
```
kathaypacific, line 5: nit: sometimes i feel like reading react components is like peeling an onion which can be annoying when there are too many layers...i wonder if we need this layer in the middle? the loading and error toast components are relatively short, perhaps we can have them together in one component?

kathaypacific (commented):
🧙🏻‍♂️

src/jumpstart/JumpstartClaimStatusToasts.tsx
bakoushin: we certainly don't need them to be separated. I was thinking about encapsulating all logic specific to a single toast to its own component. but yeah looks like it's overkill in this case.

kathaypacific (approved):
✨

PR #5112, opened 2024-03-15
chore(jumpstart): add `claim` to the names of claim-related actions and selectors

bakoushin:
### Description

As the title

### Test plan

CI

### Related issues

- Related to RET-1004

### Backwards compatibility

Y

### Network scalability

NA


jeanregisser (approved):
👍

PR #5117, opened 2024-03-18
chore: `BottomSheetInLineNotification` -> `Toast`

bakoushin:
### Description

Rewire all components using [`BottomSheetInLineNotification`](src/components/BottomSheetInLineNotification.tsx) to use [`Toast`](src/components/Toast.tsx).

It has the same look and properties signature.

### Test plan

CI

### Related issues

- Related to RET-1004

### Backwards compatibility

Y

### Network scalability

NA


src/components/Toast.tsx
```diff
@@ -55,10 +56,17 @@ const Toast = ({
   swipeable,
   position = 'bottom',
   onDismiss,
+  onUnmount,
```
bakoushin, line 59: forwarded from `BottomSheetInlineNotification` to handle the edge case when user navigates from the screen without explicitly dismissing the toast.

context: https://github.com/valora-inc/wallet/pull/5045/files#r1513591182

kathaypacific (approved):
💛

PR #5120, opened 2024-03-18
feat(jumpstart): Adding jumpstart transaction detection

dievazqu:
### Description

This PR handles jumpstart transaction home feed and common details screen.

### Test plan

Unit test + emulator

### Related issues

- Fixes RET-996

### Backwards compatibility

Y

### Video recording

https://github.com/valora-inc/wallet/assets/12504748/4686fa6d-0bca-4268-be92-6d60c9b098e8


src/jumpstart/selectors.ts
```diff
@@ -11,3 +15,8 @@ export const showJumstartClaimError = (state: RootState) => {
 export const jumpstartSendStatusSelector = (state: RootState) => {
   return state.jumpstart.depositStatus
 }
+
+export const getJumpstartContractAddress = (networkId: NetworkId) => {
```
kathaypacific, line 19: this feels a little out of place in a selector file since it doesn't relate to redux. i see it is only used in the utils file, can we remove this function here and use the logic directly in the util?

src/jumpstart/utils.ts
```diff
@@ -0,0 +1,7 @@
+import { getJumpstartContractAddress } from 'src/jumpstart/selectors'
```
kathaypacific, line 1: i think we are trying to not create more general "utils" files because they become a dumping ground for random functions (sometimes poorly thought out abstractions). perhaps we can rename this file to something like "transactionClassificationHelpers" or something? or perhaps we can add to the mess in the transactions folder, there is already a utils file there that this function might make sense in :)

src/transactions/feed/JumpstartFeedItem.tsx
```diff
@@ -0,0 +1,127 @@
...
+interface Props {
+  transfer: TokenTransfer
+}
+
+function JumpstartFeedItem({ transfer }: Props) {
+  const { t } = useTranslation()
+  const { amount, type } = transfer
+
+  const openTransferDetails = () => {
+    navigate(Screens.TransactionDetailsScreen, { transaction: transfer })
```
kathaypacific, line 32: since we already know that this transaction is a jumpstart transaction here, can we add this information as a nav parameter? then the TransactionDetailsScreen would already receive this information as a boolean, and wouldn't need to call `isJumpstartTransaction` at all. (and then `isJumpstartTransaction` would only need to be called once in the TransactionFeed, and perhaps can live as an unexported function in that file.

src/transactions/feed/JumpstartFeedItem.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const hideHomeBalanceState = useSelector(hideHomeBalancesSelector)
+  const hideBalance =
+    getFeatureGate(StatsigFeatureGates.SHOW_HIDE_HOME_BALANCES_TOGGLE) && hideHomeBalanceState
+
+  const subtitle =
+    type === TokenTransactionTypeV2.Received
+      ? t('feedItemJumpstartReceivedSubtitle')
+      : t('feedItemJumpstartSentSubtitle')
+
+  return (
```
kathaypacific, line 49: reading this component, it is so similar to the existing TransferFeedItem...did you consider sharing code between the components? in particular, i think the whole render block could be shared? it seems like the only areas of customisation are the image, title, and subtitle?

src/transactions/feed/TransactionFeed.test.tsx
```diff
@@ -16,9 +16,9 @@ import {
 } from 'src/transactions/types'
 import networkConfig from 'src/web3/networkConfig'
 import { createMockStore, RecursivePartial } from 'test/utils'
-import { mockCusdAddress, mockCusdTokenId } from 'test/values'
-import { mockApprovalTransaction } from 'test/values'
+import { mockApprovalTransaction, mockCusdAddress, mockCusdTokenId } from 'test/values'
 
+jest.mock('src/jumpstart/utils')
```
kathaypacific, line 21: perhaps we can mock the return value of statsig rather than mocking this module? there are no unit tests for this module so kind of nice to not exclude this file intentionally

kathaypacific (commented):
looking great!! ✨

src/jumpstart/selectors.ts
dievazqu: Sure, I can move this function.
I saw this example: https://github.com/valora-inc/wallet/blob/409ace74227f31ec46bb0015c3b44bad28c65daa/src/tokens/selectors.ts#L522
So I thought it was ok to leave this method here.

src/transactions/feed/JumpstartFeedItem.tsx
dievazqu: Yes! I considered it but the code was a bit ugly because how the `TransactionFeedItemImage` was used.
And I went with the quick solution, but now I changed `TransactionFeedItemImage` implementation, so I think I can merge this feed item to the transfer one.
Thanks for the call out!

src/transactions/transferFeedUtils.ts
```diff
@@ -123,6 +123,7 @@ export function useTransferFeedDetails(transfer: FeedTokenTransfer) {
   const tokenInfo = useTokenInfoByAddress(transfer.amount.tokenAddress)
   const coinbasePaySenders = useSelector(coinbasePaySendersSelector)
   const fcTransferDisplayInfo = useFiatConnectTransferDisplayInfo(transfer)
+  const isJumpstart = isJumpstartTransaction(transfer)
```
kathaypacific, line 126:
```suggestion
  const isJumpstartTransaction = isJumpstartTransaction(transfer)
```
super nit 😅

kathaypacific (approved):
LGTM!! i didn't really look at the changes in the TransactionDetailsScreen / TransactionDetailsScreen.tests though since it looks like the changes are reverted in the next PRs. it'd be nice to remove the unnecessary changes from this PR before merging? i think it's okay if the tx details screen for jumpstart just shows as payment sent / recieved

src/transactions/transferFeedUtils.ts
dievazqu: I can't use that name, because the function is called like that. 😅 
I hope leaving `isJumpstart` is fine.

kathaypacific (approved):
🚀

PR #5121, opened 2024-03-18
fix(profile): register new screens for settings, help, invite

finnian0826:
### Description

Registers new screens and modifies headers so can navigate back.

### Test plan

Unit tests updated.

**Manual tests:**

Feature gate false:

https://github.com/valora-inc/wallet/assets/140328381/74b676ea-334a-4c49-ae17-f868ac98b776

Feature gate true:

https://github.com/valora-inc/wallet/assets/140328381/91ded2a7-d54f-43a8-8f65-a33ac8a3fbaf

### Related issues

- Fixes ACT-1132

### Network scalability

- N/A


src/account/Settings.tsx
```diff
@@ -435,7 +435,7 @@ export const Account = ({ navigation, route }: Props) => {
   }
 
   return (
-    <SafeAreaView style={styles.container}>
+    <SafeAreaView style={styles.container} edges={['bottom']}>
```
MuckT, line 438: Does the edges need to be conditional also?

src/backup/BackupComplete.tsx
```diff
@@ -29,7 +31,12 @@ function BackupComplete({ route }: Props) {
   useEffect(() => {
     const timer = setTimeout(() => {
       if (navigatedFromSettings) {
-        navigate(Screens.Settings, { promptConfirmRemovalModal: true })
+        navigate(
+          getFeatureGate(StatsigFeatureGates.USE_TAB_NAVIGATOR)
+            ? Screens.Settings
+            : Screens.SettingsDrawer,
```
satish-ravi, line 37: I believe this will lead to a crash if a feature gate changes when the app is backgrounded. Can you test the follow scenario?
1. Have the feature on for your wallet, navigate to this screen (or any of the below screens) 
2. Background the app
3. Change your wallet to be on the false variant and give a few mins for things to be updated
4. Foreground the app and invoke the action to go to settings

src/backup/BackupComplete.tsx
```diff
@@ -29,7 +31,12 @@ function BackupComplete({ route }: Props) {
   useEffect(() => {
     const timer = setTimeout(() => {
       if (navigatedFromSettings) {
-        navigate(Screens.Settings, { promptConfirmRemovalModal: true })
+        navigate(
+          getFeatureGate(StatsigFeatureGates.USE_TAB_NAVIGATOR)
+            ? Screens.Settings
```
satish-ravi, line 36: Wonder if we have to clear the stack before navigating to the new settings screen with back. Previously we landed back on the drawer and there was no way for the user to go back, but wonder if this opens up the possibility on landing on an inconsistent state in the app.

src/pincode/PincodeSet.test.tsx
```diff
@@ -166,7 +166,7 @@ describe('Pincode', () => {
     })
 
     expect(updatePin).toHaveBeenCalledWith(mockAccount.toLowerCase(), oldPin, mockPin)
-    expect(navigate).toBeCalledWith(Screens.Settings)
+    expect(navigate).toBeCalledWith(Screens.SettingsDrawer)
```
satish-ravi, line 169: better to switch any test that checks settings navigation to run against the `true` variant (or test both scenarios). Might also be worth wrapping navigating to settings into a helper function given my comments above. I don't think it will be as simple as navigating one or the other screen

src/account/Support.tsx
```diff
@@ -20,8 +22,15 @@ const Support = () => {
   const { t } = useTranslation()
 
   return (
-    <SafeAreaView style={styles.container}>
-      <DrawerTopBar />
+    <SafeAreaView
+      style={styles.container}
+      edges={
+        getFeatureGate(StatsigFeatureGates.USE_TAB_NAVIGATOR)
```
MuckT, line 28: Instead of getting the gate twice we can assign it to a constant. 
`const isTabNav = getFeatureGate(StatsigFeatureGates.USE_TAB_NAVIGATOR)`

src/backup/BackupComplete.tsx
finnian0826: Simulator Screenshot - iPhone 14 Pro - 2024-03-19 at 10 31 36
@satish-ravi you're right, I get this error when following this procedure to go back to settings. Is there a way to do navigate to make sure the screens are available, or should it just navigate home?

satish-ravi: you can navigate to Screens.DrawerNavigator with initial screen name as Screens.SettingsDrawer, there should be examples of others.

src/navigator/NavigationService.ts
```diff
@@ -231,6 +231,12 @@ export function navigateHome(fromModal?: boolean) {
   }, timeout)
 }
 
+export function navigateToSettings(promptConfirmRemovalModal: boolean) {
+  getFeatureGate(StatsigFeatureGates.USE_TAB_NAVIGATOR)
+    ? navigateClearingStack(Screens.Settings, { promptConfirmRemovalModal })
```
satish-ravi, line 236: actually would the user get stuck on the settings screen here? this would reset settings to the first screen. wonder if we should just navigate home, although not sure if it will break stuff. Maybe good to check in with product as well

finnian0826: yeah I was thinking of just navigating to home, I'll check with product and see

satish-ravi: I synced with Laura on this and we don't want to navigate home instead of settings in most places. Seems like there's only a handful of places where we navigate to settings. So let's do this:
- change keyless backup progress screen to navigate home instead of settings (as the flow can start from either settings or notifications screen)
- to solve the edge case of feature gate changing when the user is on the Settings screen and backgrounded, I think we can get away with using the same screen enum. If we do that, then the version of settings screen that the flow started with will already be on the stack, so any `navigate(Screen.Settings)` would just pop back to that screen, instead of adding the new version of the screen to the stack. This should work as long as there is no flow that starts from somewhere else in the app and ends in settings, which I believe is only the keyless backup progress screen which we've addressed above.

src/navigator/Navigator.tsx
```diff
@@ -521,6 +528,17 @@ const generalScreens = (Navigator: typeof Stack) => (
       component={ProfileMenu}
       options={ProfileMenu.navigationOptions as NativeStackNavigationOptions}
     />
+    <Navigator.Screen
+      name={Screens.Settings}
+      component={SettingsScreen}
+      options={isTabNav ? headerWithBackButton : noHeader}
```
satish-ravi, line 534: could we just register the back button always? I think this version of the screen should only be used in the tab navigator anyway. The screens I believe are registered at startup and won't be refreshed if the gate changes when app is backgrounded.

MuckT, line 534:
```suggestion
      options={headerWithBackButton}
```

src/navigator/Navigator.tsx
```diff
@@ -521,6 +528,17 @@ const generalScreens = (Navigator: typeof Stack) => (
...
+    />
+    <Navigator.Screen name={Screens.Invite} component={Invite} options={noHeader} />
+    <Navigator.Screen
+      name={Screens.Support}
+      component={Support}
+      options={isTabNav ? headerWithBackButton : noHeader}
```
MuckT, line 540:
```suggestion
      options={headerWithBackButton}
```

src/navigator/Navigator.tsx
```diff
@@ -489,6 +494,8 @@ const settingsScreens = (Navigator: typeof Stack) => (
   </>
 )
 
+const isTabNav = getFeatureGate(StatsigFeatureGates.USE_TAB_NAVIGATOR)
```
MuckT, line 497:
```suggestion
```

MuckT (approved):
@finnian0826 Thanks for working to getting this across the finish line! 🚀

src/backup/BackupQuiz.tsx
```diff
@@ -73,15 +73,12 @@ const mapStateToProps = (state: RootState): StateProps => {
...
-    navigate(Screens.Settings)
-  }
+  const settingsScreen = route.params?.settingsScreen ?? undefined
   return {
     ...emptyHeader,
     headerLeft: () => {
-      return navigatedFromSettings ? (
-        <CancelButton onCancel={onCancel} style={styles.cancelButton} />
+      return settingsScreen ? (
+        <CancelButton onCancel={() => navigateHome()} />
```
satish-ravi, line 81: probably fine to go home here but the settingsScreen param now allows us to retain existing behavior right? We can navigate similar to what is done in BackupComplete?

src/backup/BackupQuiz.tsx
```diff
@@ -419,9 +416,6 @@ const styles = StyleSheet.create({
     paddingVertical: 4,
   },
   resetButton: { alignItems: 'center', padding: 24, marginTop: 8 },
-  cancelButton: {
-    color: colors.gray4,
-  },
```
satish-ravi, line 424: same question as above

src/backup/BackupPhrase.tsx
```diff
@@ -97,24 +96,23 @@ class BackupPhrase extends React.Component<Props, State> {
...
     const { t, backupCompleted } = this.props
     const { mnemonic, isConfirmChecked } = this.state
-    const navigatedFromSettings = this.navigatedFromSettings()
     return (
       <SafeAreaView style={styles.container}>
         <CustomHeader
           style={{ paddingHorizontal: variables.contentPadding }}
           left={
-            navigatedFromSettings ? (
-              <CancelButton style={styles.cancelButton} />
```
satish-ravi, line 117: seems like this was just navigating back previously ? could we retain the same behavior here?

src/backup/BackupPhrase.tsx
```diff
@@ -194,9 +192,6 @@ const styles = StyleSheet.create({
     ...fontStyles.regular,
     paddingLeft: 8,
   },
-  cancelButton: {
-    color: colors.gray4,
-  },
```
satish-ravi, line 199: any reason to remove this styling?

src/backup/BackupComplete.tsx
```diff
@@ -22,14 +22,17 @@ import fontStyles from 'src/styles/fonts'
...
-  const navigatedFromSettings = route.params?.navigatedFromSettings ?? false
+  const settingsScreen = route.params?.settingsScreen ?? undefined
   const backupCompleted = useSelector(backupCompletedSelector)
   const { t } = useTranslation()
 
   useEffect(() => {
     const timer = setTimeout(() => {
-      if (navigatedFromSettings) {
-        navigate(Screens.Settings, { promptConfirmRemovalModal: true })
+      if (settingsScreen) {
```
satish-ravi, line 31: might be worth leaving a cleanup comment here. I liked your naming on `isAccountRemoval`, maybe we can do that when we clean the drawer ? Should be also fine to just describe this in the linear issue @MuckT created.

satish-ravi (approved):

PR #5123, opened 2024-03-19
test(e2e): update tests to run against tab navigator

satish-ravi:
### Description

Updates the following tests to run against both drawer and tab navigator. Not all tests run against both versions, some tests in suite run against drawer and some against tabs so we cover both navigation patterns.

- HomeFeed.spec.js - run against both drawer and tabs
- Discover.spec.js - specific to tabs, just runs the DappListDisplay test from existing Dapps test
- QRScanner.spec.js - runs one suite against drawer and one against tabs
- FiatConnectTransferOut.spec.js - KYC runs against tabs, non KYC runs against drawer
- Send.spec.js - 2 suites run against tabs, 1 against drawer (secure send also runs against drawer)
- OnRamps.spec.js - runs against tabs and OffRamps test runs against drawer

### Test plan

CI

### Related issues

- Part of ACT-1113

### Backwards compatibility

N/A

### Network scalability

N/A


e2e/src/usecases/NewAccountCeloScreen.js
satish-ravi, line 1: This is not used in any `.spec.js` and seems outdated

e2e/src/usecases/DappListDisplay.js
```diff
@@ -28,14 +27,11 @@ export default DappListDisplay = () => {
...
       .withTimeout(10 * 1000)
+    await element(by.id('WebViewScreen/CloseButton')).tap()
   })
 
   it(':ios: should correctly filter dapp list based on user agent', async () => {
     const iOSDappList = await fetchDappList('Valora/1.0.0 (iOS 15.0; iPhone)')
-    await reloadReactNative()
-    await navigateToDappList()
-    // Give a few seconds for the dapp list to load
-    await sleep(5 * 1000)
```
satish-ravi, line 38: this made it hard to reuse the test for Discover tab, so cleaned up to be independent of how dapps page is reached. We typically don't need a fresh load here if the above test passes.

finnian0826 (approved):

PR #5124, opened 2024-03-19
fix: prevent phone input number from being hidden behind keyboard

MuckT:
### Description

Uses the keyboard aware scroll view to prevent the keyboard from obscuring the 'continue' button when the phone number input is selected. Also sets SafeAreaView to `height: '100%'` instead of `flexGrow: 1` to prevent continue buttons from render off screen when large fonts are used as this can prevent users from advancing to the next screen.

#### Backup Introduction

| iOS Before | iOS After | Android Before | Android After| 
| ---- | ---- | ---- | ---- |
|  |  |  |  |

#### Wallet Security Primer

| iOS Before | iOS After | Android Before | Android After| 
| ---- | ---- | ---- | ---- |
|  |  |  |  |

#### Phone Number Input

| iOS Before | iOS After | Android Before | Android After| 
| ---- | ---- | ---- | ---- |
|  |  |  |  |
### Test plan

- [x] Tested locally on iOS

### Related issues

- Fixes ACT-1117

### Backwards compatibility

Yes

### Network scalability

N/A

### DIFF
--- a/src/keylessBackup/KeylessBackupIntro.tsx
+++ b/src/keylessBackup/KeylessBackupIntro.tsx
@@ -106,7 +106,7 @@ export default KeylessBackupIntro
 const styles = StyleSheet.create({
   container: {
     justifyContent: 'space-between',
-    flexGrow: 1,
+    height: '100%',
   },
   scrollContainer: {
     paddingHorizontal: Spacing.Thick24,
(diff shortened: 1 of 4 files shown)


finnian0826 (approved):
lgtm!

PR #5125, opened 2024-03-19
feat(jumpstart): add reclaim bottom sheet

dievazqu:
### Description

This PR adds the reclaim button to the escrow deposits.
It checks if the escrow was claimed and open the bottom sheet, but it doesn't execute the transaction.

### Test plan

Emulator

### Related issues

- Fixes RET-1003

### Backwards compatibility

Y

### Network scalability

- [y] Continue to work without code changes, OR trigger a compilation error (guaranteeing we find it when a new network is added)


src/navigator/Navigator.tsx
```diff
@@ -685,6 +686,10 @@ function nativeBottomSheets(BottomSheet: typeof RootStack) {
         name={Screens.FiatExchangeCurrencyBottomSheet}
         component={FiatExchangeCurrencyBottomSheet}
       />
+      <BottomSheet.Screen
```
kathaypacific, line 689: curious why you chose a screen over a bottom sheet component directly inside the transaction details component? it seems like the only place that needs to launch this bottom sheet is the details screen (the components we have registered as screens so far are ones that can launched from many places...)

src/transactions/feed/detailContent/JumpstartReclaimBottomSheet.tsx
```diff
@@ -0,0 +1,68 @@
...
+function JumpstartReclaimBottomSheet({
+  route: {
+    params: { reclaimTx, networkId, tokenAmount },
+  },
+}: Props) {
+  const { t } = useTranslation()
+
+  const transactionString = JSON.stringify(reclaimTx)
+
+  async function onConfirm() {
```
kathaypacific, line 31: we haven't been terribly consistent about this but i think it is a bad pattern to trigger async functions on button presses because then we don't handle rejected promises well (and don't necessarily await the outcome before moving on). i started using the useAsyncCallback hook to handle this, maybe we can do something similar here? it would also enable the loading state out of the box. https://github.com/valora-inc/wallet/blob/0af2960d9bb876cfdd9ade50b1ab210fcd911345/src/jumpstart/JumpstartEnterAmount.tsx#L68

src/transactions/feed/detailContent/JumpstartReclaimBottomSheet.tsx
```diff
@@ -0,0 +1,68 @@
...
+    Logger.debug(TAG, 'Reclaiming', { reclaimTx, networkId, tokenAmount })
+    // TODO: Send transaction
+    navigateBack()
+    navigateHome()
```
kathaypacific, line 35: maybe just navigateHome is okay here?

src/transactions/feed/detailContent/JumpstartContent.tsx
```diff
@@ -1,35 +1,168 @@
...
 
+  const transferTokenInfo = useTokenInfo(transfer.amount.tokenId)
   const parsedAmount = new BigNumber(transfer.amount.value).abs()
   const token = useTokenInfo(transfer.amount.tokenId)
   const isDeposit = transfer.type === TokenTransactionTypeV2.Sent
 
+  const walletAddress = useSelector(walletAddressSelector)
+  const jumpstartContractAddress = getJumpstartContractAddress(transfer.networkId)
+
+  const fetchEscrowData = useAsync(
```
kathaypacific, line 85: nit: can we be consistent with how we call this feature 😅 jumpstart rather than escrow?
```suggestion
  const fetchJumpstartDepositData = useAsync(
```

src/transactions/feed/detailContent/JumpstartContent.tsx
```diff
@@ -1,35 +1,168 @@
...
+      tokenAmount: {
+        value: parsedAmount.toString(),
+        tokenId: token.tokenId,
+        tokenAddress: token.address ?? undefined,
+        localAmount: transfer.amount.localAmount,
+      },
+    })
+  }
+
+  const ReclaimButton = () => (
```
kathaypacific, line 151: i think declaring new components with this syntax is bad for performance because these are recreated on every re-render...
```suggestion
  const ReclaimButton = (
```
alternatively, why not just use this render block directly inside the render function? i'd argue that it's easier to read the render function this way, because you don't need to jump around to understand the components

src/transactions/feed/detailContent/JumpstartContent.tsx
```diff
@@ -1,35 +1,168 @@
...
+    <View style={styles.buttonContainer}>
+      {fetchEscrowData.result?.claimed ? (
+        <ClaimedLabel />
+      ) : (
+        <Button
+          showLoading={fetchEscrowData.loading}
+          disabled={fetchEscrowData.loading}
```
kathaypacific, line 158: we need a separate loading button right? otherwise when loading whether a user can reclaim, the button will show "Claim" + loader?

src/transactions/feed/detailContent/JumpstartContent.tsx
```diff
@@ -1,35 +1,168 @@
...
+  )
+
   if (!token) {
     // should never happen
     Logger.error(TAG, 'Token is undefined')
     return null
   }
 
-  // TODO: Add reclaim button
+  const ClaimedLabel = () => (
```
kathaypacific, line 108: i wonder if you can use the Button component with `disabled` and `touchableStyle` set to the green background? maybe you also need to set the color of the font with `fontStyle` but then it'd ensure that this disabled button won't look different from the enabled / loading buttons?

kathaypacific (commented):
i wasn't sure if this PR is ready for review (it's not in draft) but i left some early comments anyway 🙈 looking good!

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
```diff
@@ -0,0 +1,320 @@
...
+        type={!isClaimed ? BtnTypes.PRIMARY : BtnTypes.LABEL_PRIMARY}
+        text={!isClaimed ? t('reclaim') : t('claimed') + ' ✓'}
+        fontStyle={typeScale.labelSemiBoldMedium}
+        size={BtnSizes.FULL}
+      />
+    </View>
+  )
+
+  return (
+    <SafeAreaView style={styles.safeAreaContainer}>
```
kathaypacific, line 168:
```suggestion
    <SafeAreaView style={styles.safeAreaContainer} edges={['top'])>
```
^ there is already a safe area view inside the scroll view, only doing the top edge here means that we can show scrollable content all the way past the bottom of the screen.

please also make sure that we're using the safe area view from the react-native-safe-area-context lib

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
```diff
@@ -0,0 +1,320 @@
...
+      gas: '130000',
+      maxFeePerGas: '1000000000',
+    })
+
+    bottomSheetRef.current?.snapToIndex(0)
+  }
+
+  const isClaimed = fetchClaimData.result?.claimed
+
+  const reclaimButton = (
```
kathaypacific, line 153: this is personal preference but i think this can go directly into the render rather than be a variable, it would be easier to read all the render logic in one place

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
```diff
@@ -0,0 +1,320 @@
...
+    <View style={styles.buttonContainer}>
+      <Button
+        showLoading={!fetchClaimData.result}
+        disabled={!fetchClaimData.result}
+        onPress={onReclaimPress}
+        type={!isClaimed ? BtnTypes.PRIMARY : BtnTypes.LABEL_PRIMARY}
+        text={!isClaimed ? t('reclaim') : t('claimed') + ' ✓'}
```
kathaypacific, line 160: we should avoid creating strings in the app - why not use the `<Checkmark />` icon instead? the button provides a prop for icons

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
```diff
@@ -0,0 +1,320 @@
...
+
+    bottomSheetRef.current?.snapToIndex(0)
+  }
+
+  const isClaimed = fetchClaimData.result?.claimed
+
+  const reclaimButton = (
+    <View style={styles.buttonContainer}>
+      <Button
+        showLoading={!fetchClaimData.result}
```
kathaypacific, line 156: curious why this is not loading when `fetchClaimData.loading` only?

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
```diff
@@ -0,0 +1,320 @@
...
+        disabled={!fetchClaimData.result}
+        onPress={onReclaimPress}
+        type={!isClaimed ? BtnTypes.PRIMARY : BtnTypes.LABEL_PRIMARY}
+        text={!isClaimed ? t('reclaim') : t('claimed') + ' ✓'}
+        fontStyle={typeScale.labelSemiBoldMedium}
```
kathaypacific, line 161: can we avoid using a custom font here? i think we should leave the button as it is, even if it doesn't match designs. consistency is key

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
```diff
@@ -0,0 +1,320 @@
...
+
+  const onReclaimPress = () => {
+    if (!fetchClaimData.result) {
+      Logger.error(TAG, 'No escrow data found when trying to reclaim')
+      return
+    }
+
+    const { beneficiary, index } = fetchClaimData.result
+
+    setReclaimTx({
```
kathaypacific, line 134: we should move this to the `onSuccess` handler of the `useAsync`. state changes are batched so doing it here means that the reclaimTx might not have been set to this new value when the bottom sheet appears. i don't see the advantage of doing it here

kathaypacific (commented):
is there a trick to getting the reclaim button to stop spinning? i got an error every time i open the transaction details screen and the button keeps spinning, i tried with cEUR and CELO for the deposit token

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
dievazqu: Oh, I didn't know, I'll change it

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
dievazqu: Hmm, sure... But this is to match the same font we are using in the `LineItemRow` components (which are also used in the `TransferSentContext`).
I think is intended to show the same font in the same screen?

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
dievazqu: Yes, I went back and forth on this, when there is an error fetching the data should we show a disable "Reclaim" or show a disable spinner?
I changed it to `loading` in the next PR, but I'll bring that change here.

I will show a disable "Reclaim" and also the error message, and then if the user dismisses the error message, I'll trigger a new fetch. What do you think?
It's that or just leave the "Reclaim" as disable (so the user has to go back and enter again to try to reclaim again).

What I like of leaving the spinner is that if there is an error when fetching the data, we don't know whether the escrow was claimed or not, so showing "Reclaim" as disable gives a hint that it wasn't claim yet (but I think we can live with this). Or should I ask Kayla for some feedback for this case, maybe we can to change the wording of the button in that case? 🙈

src/components/Button.tsx
```diff
@@ -194,6 +195,10 @@ function getColors(type: BtnTypes, disabled: boolean | undefined) {
       backgroundColor = colors.gray1
       borderColor = colors.gray2
       break
+    case BtnTypes.LABEL_PRIMARY:
```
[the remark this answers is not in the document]
dievazqu: No, I didn't... I'll check with her if we are gonna reuse this pattern.

I did this to avoid more complexity... It seems that the [button decides the color of the text](https://github.com/valora-inc/wallet/blob/2f60f03cb3b15ce8f1340adc4c3d408651f462cd/src/components/Button.tsx#L121).
And then the only "pretty" way was [adding a new type here](https://github.com/valora-inc/wallet/blob/2f60f03cb3b15ce8f1340adc4c3d408651f462cd/src/components/Button.tsx#L164).
Even if we don't use that combination again, I think we should have a specific button type for it. Unless [this comment](https://github.com/valora-inc/wallet/blob/2f60f03cb3b15ce8f1340adc4c3d408651f462cd/src/components/Button.tsx#L120) is no longer true (or we explore another workaround for it).

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
dievazqu: Yeah, I think I prefer to keep the render tree as simple as I can. But I lost that fight in this repo, so I should stick to the standard, this is simple enough to be there. 👍

dievazqu:
> is there a trick to getting the reclaim button to stop spinning? i got an error every time i open the transaction details screen and the button keeps spinning, i tried with cEUR and CELO for the deposit token

@kathaypacific Yeah, sorry, this doesn't work because the contract address in the statsig config is linked to the old contract that don't support reclaim. I didn't want to change it to avoid messing the config for you.
Once we deploy the latest contract we should update the config. Should I create a small task for it? We are exploring deployment options with the NFT stuff, so probably we should use the same pattern for this deployment?

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
dievazqu: Sorry, this is also changed in the next PR, think of it as a mock for this PR 😅 

We need the gas estimation before setting the reclaimTx, and that happens when the user taps on the "Reclaim" button. Or... Maybe we can try to estimate the gas when we fetch the escrow data even if the user doesn't tap on "Reclaim"? It might take a bit more of time to load at the beginning but then the experience will be smoother.
I'm gonna do that in the following PR since I implemented the prepared transaction there. Happy to follow this discussion there.

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
kathaypacific: aw man, tricky. i think we should show loaders only when there is a background load happening, when loaders are shown for too long then it looks like the app is buggy. so i think for the error state, we shouldn't show the loader.

so the question is what should we show 😅 i think it's probably find to leave it as a disabled "Reclaim" state. let's track how many times users are hitting this error to see if we want to spend more time on this - fwiw i think the error popup is already enough UI feedback that something has gone wrong, so the disabled button label shouldn't be too consequential.

i also think it's okay to not retry the fetch, and let users come back to the screen. if i understand the fetch function, errors are likely _not_ to be transient so adding a retry doesn't really increase the chances of users being able to proceed? so seems not worth the effort for now

src/transactions/feed/JumpstartTransactionDetailsScreen.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+  })
+
+  function renderScreen({
+    storeOverrides = {},
+    transaction,
+  }: {
+    storeOverrides?: RecursivePartial<RootState>
+    transaction: TokenTransfer
+  }) {
+    const store = createMockStore({
```
kathaypacific, line 55: nit: we don't really need to create a fancy mock store right? the identity and recipients stores are not relevant for this test?

src/transactions/feed/JumpstartTransactionDetailsScreen.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+      }),
+    })
+
+    const amountSendComponent = getByTestId('LineItemRowTitle/JumpstartContent/TokenDetails')
+    expect(getElementText(amountSendComponent)).toBe('amountSent')
+    const amountValue = getByTestId('JumpstartContent/AmountValue')
+    expect(getElementText(amountValue)).toBe('10.00 cUSD')
+  })
+
+  it('handles jumpstart received transactions correctly', async () => {
```
kathaypacific, line 129: can we add some tests for the claim button states?

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
```diff
@@ -0,0 +1,323 @@
...
+  const erc20Claim = await viemClient.readContract({
+    address: jumpstartContractAddress as Address,
+    abi: walletJumpstart.abi,
+    functionName: 'erc20Claims',
+    args: [beneficiary, index],
+  })
+
+  const claimed = erc20Claim[3]
+
+  Logger.debug(`Reward deposited in ${transactionHash} ${claimed ? 'was' : 'was NOT'} claimed`)
```
kathaypacific, line 76: nit: to make the logs easier to read, can we add a `TAG` in all the log statements?

locales/base/translation.json
```diff
@@ -2219,5 +2221,6 @@
       "contactSupport": "Contact Support",
       "dismiss": "Dismiss"
     }
-  }
+  },
+  "jumpstartReclaimDescription": "Reclaiming your link incurs a network gas fee, would you like to proceed?"
```
kathaypacific, line 2225: nit: could we create an object to group the translations under? i think you'll need to add error messages (the generic fetch + the failed to reclaim ones) and it'd be easier to understand where those errors belong to if they're grouped within a `jumpstartReclaim` key?

kathaypacific (commented):
i think this is looking great, just small nit comments (but the blocking one is really the test coverage 😅)

kathaypacific (approved):

PR #5126, opened 2024-03-19
feat(jumpstart): execute reclaim transaction

dievazqu:
### Description

This PR implements the actual call to the blockchain to execute the reclaim. It also implements the error handling.

### Test plan

Successful reclaim:

https://github.com/valora-inc/wallet/assets/20150449/d474908b-e738-4340-ad77-54dcaf7f391c

Reclaim status fetch error:

https://github.com/valora-inc/wallet/assets/20150449/8ddf298f-4d94-44f2-9905-6fd909c48592

Reclaim error:

https://github.com/valora-inc/wallet/assets/20150449/67c1d2b5-b772-4911-99c8-961330e01a73

### Related issues

- Fixes RET-1003

### Backwards compatibility

Y

### Network scalability

- [Y] Continue to work without code changes, OR trigger a compilation error (guaranteeing we find it when a new network is added)

### Demo

Happy path
https://github.com/valora-inc/wallet/assets/12504748/ff782b85-3045-4cba-bedd-c376f181d3ee

Error when fetching the status
https://github.com/valora-inc/wallet/assets/12504748/799c5849-000e-4770-8fe8-d8a6d95ba0bf


src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
```diff
@@ -90,10 +99,34 @@ function JumpstartTransactionDetailsScreen({ route }: Props) {
...
+  const tokenAmount = transaction.amount
 
   const walletAddress = useSelector(walletAddressSelector)
   const bottomSheetRef = useRef<BottomSheetRefType>(null)
 
+  const [error, setError] = React.useState<Error | null>(null)
+  const feeCurrencies = useSelector((state) => feeCurrenciesSelector(state, networkId))
+
+  const reclaimStatus = useSelector(jumpstartReclaimStatusSelector)
+  const [isReclaimIniciated, setReclaimInitiated] = useState(false)
```
kathaypacific, line 112: i don't think you need this state variable at all. when you dispatch the jumpstartReclaimStarted action, reclaimStatus will change to 'loading' so you can use that directly

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
```diff
@@ -90,10 +99,34 @@ function JumpstartTransactionDetailsScreen({ route }: Props) {
   const token = useTokenInfo(transaction.amount.tokenId)
   const isDeposit = transaction.type === TokenTransactionTypeV2.Sent
   const jumpstartContractAddress = transaction.address
+  const networkId = transaction.networkId
+  const tokenAmount = transaction.amount
 
   const walletAddress = useSelector(walletAddressSelector)
   const bottomSheetRef = useRef<BottomSheetRefType>(null)
 
+  const [error, setError] = React.useState<Error | null>(null)
```
kathaypacific, line 108: i don't think you need this state variable either? it seems like you are not using the value of this error so you can replace it with `fetchClaimData.error || reclaimStatus === 'error'`?

kathaypacific: speaking of which, i think the `fetchClaimData` error handler needs to be populated

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
```diff
@@ -81,6 +89,7 @@ async function getClaimDataAndStatus(
 function JumpstartTransactionDetailsScreen({ route }: Props) {
   const { transaction } = route.params
   const { t } = useTranslation()
+  const dispatch = useDispatch()
 
   const [reclaimTx, setReclaimTx] = useState<SerializableTransactionRequest | null>(null)
```
kathaypacific, line 94: can we remove this state variable and use fetchClaimData.result?.preparedTransaction instead?

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
```diff
@@ -104,47 +137,88 @@ function JumpstartTransactionDetailsScreen({ route }: Props) {
...
+        baseTransactions: [reclaimTx],
+      })
+
+      const resultType = preparedTransactions.type
+      switch (resultType) {
+        case 'need-decrease-spend-amount-for-gas': // fallthrough on purpose
+        case 'not-enough-balance-for-gas':
+          return {
+            claimed: false,
+            error: true,
```
kathaypacific, line 173: instead of returning an error in a success state, can we throw an error so that this goes through the generic error handling logic?

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
```diff
@@ -104,47 +137,88 @@ function JumpstartTransactionDetailsScreen({ route }: Props) {
...
 
-  const onReclaimPress = () => {
-    if (!fetchClaimData.result) {
-      Logger.error(TAG, 'No escrow data found when trying to reclaim')
-      return
-    }
+  const resetState = () => {
+    setReclaimTx(null)
+    setError(null)
+    void fetchClaimData.execute()
```
kathaypacific, line 204: nit: i don't think we need to refetch here, because the button CTA doesn't indicate that it will try again?

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
```diff
@@ -238,6 +312,22 @@ function JumpstartTransactionDetailsScreen({ route }: Props) {
           style={styles.tokenFiatValueContainer}
           textStyle={styles.tokenFiatValueText}
         />
+        {error && (
```
kathaypacific, line 315: i wondered if it's worth it to surface a different error message if the user doesn't have enough gas for reclaim....it's probably okay with a generic message for now

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
```diff
@@ -249,13 +339,12 @@ function JumpstartTransactionDetailsScreen({ route }: Props) {
...
-        {reclaimTx && (
-          <EstimatedNetworkFee networkId={transaction.networkId} transaction={reclaimTx} />
-        )}
-        {/* TODO: Implement confirm */}
+        {reclaimTx && <EstimatedNetworkFee networkId={networkId} transaction={reclaimTx} />}
         <Button
           text={t('confirm')}
-          onPress={() => Logger.debug('confirmed')}
+          showLoading={reclaimStatus === 'loading'}
+          disabled={reclaimStatus !== 'idle'}
```
kathaypacific, line 346:
```suggestion
          disabled={reclaimStatus === 'loading' || reclaimStatus === 'success'}
```

src/transactions/feed/TransferFeedItem.tsx
```diff
@@ -95,9 +92,10 @@ function TransferFeedItem({ transfer }: Props) {
 }
 
 function isJumpstartTransaction(tx: TokenTransfer) {
-  const jumpstartAddress = getDynamicConfigParams(
-    DynamicConfigs[StatsigDynamicConfigs.WALLET_JUMPSTART_CONFIG]
-  ).jumpstartContracts[tx.networkId]?.contractAddress
+  const jumpstartAddress = '0x7bf3fefe9881127553d23a8cd225a2c2442c438c'
+  // const jumpstartAddress = getDynamicConfigParams(
+  //   DynamicConfigs[StatsigDynamicConfigs.WALLET_JUMPSTART_CONFIG]
+  // ).jumpstartContracts[tx.networkId]?.contractAddress
```
kathaypacific, line 98: lol

kathaypacific (commented):
looking great!! i had mainly comments about simplifying state. it'd also be nice to add test coverage for the reclaim ✨

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
dievazqu: Hmm, why?
I think we don't want to show the button enabled when there was an error. So it should only be enabled if it's in `idle` state.

src/transactions/feed/TransferFeedItem.tsx
dievazqu: Oh, no, all my secrets haven been pushed 🙈

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
dievazqu: Yeah, I know the CTA is only dismiss. But we say in the copy "Try again", but what do we offer if the fetch failed?
The user has to go back to the feed and enter again to try? I think this refetch is useful in that case.
The other option would be to leave the button disabled.

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
dievazqu: Yeah, I didn't like that I'm missing the `claimed` information. I wanted to differentiate this error case when the error happens before and we don't know the claim status.
But at the end, I'm handling those scenarios as the same. So you are right, I'm gonna throw an error here.

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
dievazqu: Hmm, yeah, that should work... `reclaimStatus` could be on `error` when we first load this view because of the last attempt, so I have to ignore the reclaimStatus if the users didn't request the refetch, that's why I'm using `isReclaimIniciated`.

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
dievazqu: I'm using this variable to avoid executing the error state (or even the success state) when the view is loaded the first time and the `reclaimStatus` is from the last interaction.
Is there a way I can execute `dispatch(jumpstartReclaimFlowStarted())` once before showing the component? That would avoid this variable.

kathaypacific: i had a similar problem with the jumpstart send flow, you can decide when the start or end of the reclaim flow is and reset the state there. you can either reset the state before you first come into this screen, or reset on unmount?

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
kathaypacific: ah, i think this also relates to our discussion about the refetch above. i was thinking that it is a more obvious pattern to press the "reclaim" again to trigger the reclaim, so when there is an error, the button should be enabled again. if we think that is a good pattern, then it also removes the need for the error message to trigger the claim (but we can do it from there too, i feel more strongly that a user should be able to reclaim by pressing the reclaim)

src/transactions/feed/JumpstartTransactionDetailsScreen.tsx
dievazqu: Actually, I need a variable for this to implement the dismiss behavior.

kathaypacific (approved):

PR #5128, opened 2024-03-19
feat(points): Implement Points Home screen

jophish:
### Description

For [ACT-1022](https://linear.app/valora/issue/RET-1022/[wallet]-build-points-landing-screen-add-statsig-config). Adds in the Points home screen and sets up Statsig to store points config information.

See the Statsig Dynamic Config [here](https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/dynamic_configs/points_config). Points visibility is gated by the feature gate [here](https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/gates/show_points).

Note that at the moment, this PR is complete _except_ for:
* The actual Points logo/icon - I'm waiting on a high-quality vector image before adding it to the branding repo.
* Some copy on the Swap bottom sheet.

I added in the functionality to automatically create "sections" on the home screen for different point values - we won't need this immediately, but it was kinda fun to add to figured why not :)

### Test plan

https://github.com/valora-inc/wallet/assets/569401/c854b3d3-31c8-4d54-92f2-e65ed6bc2b76

### Related issues

- Fixes [ACT-1022](https://linear.app/valora/issue/RET-1022/[wallet]-build-points-landing-screen-add-statsig-config).

### Backwards compatibility

Yes.

### Network scalability


src/statsig/constants.ts
```diff
@@ -115,4 +117,10 @@ export const DynamicConfigs = {
       rewardReminderDate: new Date(0).toISOString(),
     },
   },
+  [StatsigDynamicConfigs.POINTS_CONFIG]: {
+    configName: StatsigDynamicConfigs.POINTS_CONFIG,
+    defaultValues: {
+      pointsMetadata: [] as PointsMetadata[],
```
jophish, line 123: The type here may seem a bit wonky at first, but I wanted something that checks the following boxes:
* Can have new pieces of metadata added while maintaining backwards compatibility
* Supports explicit card ordering

I toyed around with a few other ideas before settling on this one, but happy to consider alternatives!

src/points/PointsHome.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+type Props = NativeStackScreenProps<StackParamList, Screens.PointsHome>
+
+export default function PointsHome({ route, navigation }: Props) {
+  const { t } = useTranslation()
+
+  // TODO: Use real points balance
+  const pointsBalance = 50
+
+  const bottomSheetRef = useRef<BottomSheetRefType>(null)
```
jophish, line 28: Getting the bottom sheet to work was a little tricky, so the implementation here might seem a bit wonky... basically, we'll want to support _arbitrarily many_ bottom sheets on this page, one for each pressable card. Rather than dynamically generating refs equal to the number of pressable cards and adding a `BottomSheet` component for each one, I just decided to use a _single_ ref, and when a card is pressed, pipe the data required to render the bottom sheet back up from where it's defined (from within the cards defined in `src/points/activityCards`) to here. The `useEffect` is here to ensure that we only show the bottom sheet once the state has definitely been set.

locales/base/translation.json
```diff
@@ -2185,5 +2185,33 @@
     "discover": {
       "tabName": "Discover"
     }
+  },
+  "points": {
+    "title": "Valora Points",
+    "activity": "Activity",
+    "infoCard": {
+      "title": "More ways to earn points coming soon 🚀",
```
jophish, line 2193: i haven't actually added an emoji icon like this to the app before... doing it this way felt kinda wrong, but also maybe correct? let me know if it should be an icon that we check in instead!

src/components/PointsButton.tsx
```diff
@@ -0,0 +1,30 @@
...
+export default function PointsButton({ style, size, testID }: Props) {
+  const onPress = () => {
+    ValoraAnalytics.track(PointsEvents.points_screen_open)
+    navigate(Screens.PointsHome)
+  }
+
+  return (
+    <TopBarIconButton
+      testID={testID}
+      icon={<AttentionIcon size={size} />}
```
jophish, line 25: (this whole way of accessing the points screen is temporary, just for testers atm)

src/points/activityCards/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+  )
+  if (onCardPress) {
+    return (
+      <Touchable testID={testID} style={styles.container} onPress={onCardPress}>
+        {content}
+      </Touchable>
+    )
+  } else {
+    return (
+      <View testID={testID} style={styles.container}>
```
kathaypacific, line 43: instead of a separate component per activity, could we have a config that contains all the customisable info and render everything together? i think this is in essence what you've done, but i'm surprised by how many layers and new files there are. naively i thought it'd be as straightforward as something like 
```
const pointsConfig = [{ // some props, like image, name }, ...]
...
render(
    <>
    ...some other stuff
    {pointsConfig.map(config => <ActivityCard {...config} />}
    </>
)
```

the activity card would then take the name and decide on the stuff that is customised in a predictable way (e.g. we could have something like the bottomSheetTitle is always ```t(`points.activityCards.${activityName}.bottomSheet.title`)```

src/statsig/constants.ts
kathaypacific: i think this config shape is fine and i like the idea of supporting explicit card ordering, but the designs / wallet logic don't support this which is a shame 😅

kathaypacific (commented):
looks good!! sorry i didn't quite understand the display logic but let's chat about it at our 1:1 later!

locales/base/translation.json
```diff
@@ -2216,5 +2216,33 @@
...
+    },
+    "activityCards": {
+      "createWallet": {
+        "title": "Create a Valora Wallet"
+      },
+      "swap": {
+        "title": "Swap assets",
+        "bottomSheet": {
+          "title": "Swap assets",
+          "body": "This space will be used to give all the details needed for this. Maybe minimum amount to swap. \n\nEarn 50 points whenever you swap an asset using Valora.",
```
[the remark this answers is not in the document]
jophish: 🤦 of course!

src/analytics/Events.tsx
```diff
@@ -664,3 +664,10 @@ export enum JumpstartEvents {
   jumpstart_claim_error_dismissed = 'jumpstart_claim_error_dismissed',
   jumpstart_claim_error_contact_support = 'jumpstart_claim_error_contact_support',
 }
+
+export enum PointsEvents {
+  points_screen_open = 'points_screen_open',
+  points_screen_back = 'points_screen_back',
+  points_swap_card_press = 'points_swap_card_press',
```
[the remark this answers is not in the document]
jophish: yep, i had that idea kicking around as well -- will opt for that instead

src/icons/CheckmarkWithCircleBorder.tsx
```diff
@@ -0,0 +1,34 @@
...
+import colors from 'src/styles/colors'
+import Svg, { Path } from 'svgs'
+
+interface Props {
+  height?: number
+  width?: number
+  color?: string
+}
+
+export default class CheckmarkWithCircleBorder extends React.PureComponent<Props> {
```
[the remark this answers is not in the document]
jophish: IIRC I looked at the existing checkmark and it seemed to be different than the one used in the design on a quick inspection.. might not be a huge deal but it did look different to me

locales/base/translation.json
jeanregisser: I guess both iOS and Android share most emojis, so we're mostly fine.
Otherwise it shows squares :D

src/icons/Celebration.tsx
```diff
@@ -0,0 +1,34 @@
+import * as React from 'react'
+import colors from 'src/styles/colors'
+import Svg, { Path } from 'svgs'
+
+interface Props {
+  height?: number
+  width?: number
```
jeanregisser, line 7: Are we only conna use square images? if so we should probably use a single prop to avoid improper use.
```suggestion
  size?: number
```

src/points/ActivityCardSection.tsx
```diff
@@ -0,0 +1,143 @@
...
+interface Props {
+  onCardPress: (bottomSheetDetails: BottomSheetDetails) => void
+}
+
+export default function ActivityCardSection({ onCardPress }: Props) {
+  const { t } = useTranslation()
+
+  const pointsConfig = getDynamicConfigParams(DynamicConfigs[StatsigDynamicConfigs.POINTS_CONFIG])
+
+  function makeSection(pointsMetadata: PointsMetadata): React.ReactNode {
```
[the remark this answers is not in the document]
jeanregisser: +1 to using trying to use styles to avoid this manual way of doing layout.

src/points/activityCards/index.tsx
jeanregisser, line 1: nit: can we avoid using index files? It's easier to find components when they are in a file named with the component's name.

src/points/types.ts
```diff
@@ -0,0 +1,25 @@
+export enum PointsActivities {
+  CreateWallet = 'CreateWallet',
+  Swap = 'Swap',
+  MoreComing = 'MoreComing',
+}
```
jeanregisser, line 5: nit: I'm on a war against TS enums :D https://stackoverflow.com/questions/40275832/typescript-has-unions-so-are-enums-redundant
```suggestion
export type PointsActivity = 'CreateWallet' | 'Swap' | 'MoreComing',
```

jeanregisser (commented):
Awesome! 🚀

src/points/activityCards/index.tsx
jeanregisser: We should avoid prefer using static translation keys so we can eventually use static extraction.

https://valora-app.slack.com/archives/C025V1D6F3J/p1707472640807039

src/points/types.ts
jophish: Oh interesting... I've implemented this, but this actually introduces a new complexity re: testing whether or not a string value is a member of this union at runtime (which we want to do to ensure we don't try to render an activity that doesn't exist).

src/points/activityCards/index.tsx
jophish: @jeanregisser What are the requirements for this? Does this mean we can't store a translation key in a variable and then pass that to the translation function? I understand that we shouldn't dynamically generate the translation key _itself_ (i.e., they should be considered opaque) but can we safely do something like?

ts```
const translationKey = condition ? 'some.key' : 'some.other.key'
...
t(translationKey)
```

File: src/points/types.ts (hunk 11 above)
Comment (jeanregisser) [reply src/points/types.ts:5]: Ah sorry if you need a runtime check too then an enum is warranted 👍

File: src/points/ActivityCardSection.tsx (hunk 10 above)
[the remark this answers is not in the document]
Comment (jophish) [reply src/points/ActivityCardSection.tsx:24]: yeah, ended up being able to implement it pretty simply! thanks for the advice on this one :)

File: src/points/types.ts (hunk 11 above)
Comment (jophish) [reply src/points/types.ts]: It's actually not too bad, we just need a simple typeguard, which isn't much more difficult than checking e.g. `Object.values(SomeEnum)`, but something to keep in mind!

File: src/points/PointsHome.tsx [hunk 12]
```diff
@@ -0,0 +1,156 @@
...
+  const onCtaPressWrapper = (onPress: () => void, activity: PointsActivities) => {
+    return () => {
+      ValoraAnalytics.track(PointsEvents.points_screen_card_cta_press, {
+        activity,
+      })
+      onPress()
+    }
+  }
+  return (
+    <SafeAreaView style={styles.outerContainer}>
```
jophish, line 52: Fun little bug encountered during development for those knowledgeable/wanting to do some sleuthing - without the `outerContainer` style here, the page visually renders totally correctly, except is completely uninteractable (no scrolling, can't press on anything at all) under _only_ the following circumstances:
* Running on Android
* Page contents exceed screen size (meaning scrolling is required)
Notably, removing _either_ one of `CustomHeader` or `BottomSheet` made this issue go away... Satish and I spent some time trying to fix this, but still don't know why we were seeing this behavior :exploding_head:

src/points/activityCards/index.tsx
jeanregisser: yes almost, best is like this:

```ts
const translation = condition ? t('some.key') : t('some.other.key')
```

src/points/types.ts
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+
+const pointsActivities = ['create-wallet', 'swap', 'more-coming'] as const
+export type PointsActivities = (typeof pointsActivities)[number]
```
jeanregisser, line 4:
```suggestion
export type PointsActivity = (typeof pointsActivities)[number]
```

src/points/PointsHome.tsx
jeanregisser: oh interesting.

Not directly related to that bug, but does the `ScrollView` extend to the bottom of the screen? If so the outer `SafeAreaView` should only apply to the top, and we should either add another `SafeAreaView` just for the bottom within the `ScrollView`, or adjust it's bottom padding using `useSafeAreaInsets()`.
So it doesn't hit an invisible wall on devices with a home bar.

This is how it should behave toward the bottom:
scroll extend to bottom

src/points/ActivityCard.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  const cardContent = (
+    <>
+      {isCompleted && (
+        <View style={styles.checkmarkIcon}>
+          <CheckmarkWithCircleBorder />
+        </View>
+      )}
+      {cardDefinition.icon}
+      <Text style={styles.cardTitle}>{t(cardDefinition.title)}</Text>
```
jeanregisser, line 50: Can we make this static?
i.e. moving it to cardDefinition directly?

```ts
...
title: t('xxx')
...
```

jeanregisser (approved):
Nice work! Already approved with some suggestions.

src/points/PointsHome.tsx
```diff
@@ -0,0 +1,156 @@
...
+        </View>
+        <ActivityCardSection onCardPress={onCardPress} />
+      </ScrollView>
+      <BottomSheet
+        snapPoints={['50%']}
+        forwardedRef={bottomSheetRef}
+        testId={`PointsActivityBottomSheet`}
+      >
+        {bottomSheetParams && (
+          <View>
```
kathaypacific, line 74: nit: maybe you don't need a view here, and can get away with a fragment`<>`?

src/points/cardDefinitions.tsx
```diff
@@ -0,0 +1,37 @@
...
+    title: 'points.activityCards.createWallet.title',
+    icon: <Celebration />,
+    defaultCompletionStatus: true,
+  },
+  swap: {
+    title: 'points.activityCards.swap.title',
+    icon: <SwapArrows />,
+    defaultCompletionStatus: false,
+    bottomSheet: {
+      title: 'points.activityCards.swap.bottomSheet.title',
```
kathaypacific, line 20: idk how serious we are about doing static extraction (we've been talking about it since i first joined and haven't really made a start on it, i'm also not sure it will work with OTA translations because once unused keys are removed, they might also be removed in Crowdin?) i saw @jeanregisser comment on this in another part of the PR, so we're not dynamically constructing the keys to keep the door for this open but i think the requirement is then also that we need to pass the `t` function in here?

src/icons/CheckmarkWithCircleBorder.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react'
```
kathaypacific, line 1: i lost the original thread but i think you mentioned that we want to create a new icon here because the checkmark is not quite the same as the other checkmark 🫠 i have a couple feelings about this, one of my triggers is stuff that looks the same but slightly different because it's just not clear the value of maintaining extra code for the subtle differences + subtle differences = inconsistency. i'd encourage us to be gatekeepers of consistency and question designers when we see something that is same same but different, most of the time we're able to keep using existing components (or refresh existing components) - and if the answer is "we absolutely must use this slightly different thing because of reasons XYZ" then at least we feel better about the extra code :) 

for this could i request that we check with Kayla if we need to use a new checkmark? and if so, perhaps we can replace the existing checkmark with this one. 

i also had a thought that perhaps for these icons they're intended only for the points use case, and maybe we can put the icons in the points folder...i'd be less annoying if this icon wasn't going in the global folder which is already hard to navigate

src/points/PointsHome.tsx
kathaypacific: having this safeareaview outside is fine since we've now got the custom header which shouldn't scroll :) to elaborate on Jean's comment, we should have `edges={['top']}` here and add a paddingBottom equivalent to the safe area bottom (or use another safe area view with the edges set to bottom only) on the scrollview contents to ensure that the end of the content isn't cut off.

weird android bug, i'm not sure i understand but perhaps you were experiencing [this one](https://valora-app.slack.com/archives/C024ZFDRFR9/p1710938221828669)?

kathaypacific (approved):
i have a few more small suggestions but don't feel like i need to look at the PR again! :)

MuckT:
@jophish looks like the touchable area for the header info icon is overlapping with the others. We should wrap the icon similarly what's done in `src/components/QrScanButton.tsx` and include it along with others in ACT-1136. This seems fine as a follow up to me.

Header-Touchables

PR #5129, opened 2024-03-19
build(deps): upgrade wallet connect dependencies

MuckT:
### Description

Upgrades Wallet Connect dependencies to the latest versions: originally pinned in #4406.

### Test plan

- Tested locally on iOS and Android
  - Connect
  - Verify
  - Sign Transaction

### Related issues

- Related: ACT-962

### Backwards compatibility

Yes

### Network scalability

N/A


MuckT:
Not sure why Codecov is complaining. The new type is tested in `src/walletConnect/screens/SessionRequest.test.tsx`.

package.json
```diff
@@ -245,8 +245,8 @@
     "@valora/eslint-config-typescript": "^1.0.2",
     "@valora/resolve-kit": "^1.0.1",
     "@walletconnect/legacy-types": "2.0.0",
-    "@walletconnect/sign-client": "2.10.0",
-    "@walletconnect/types": "2.10.0",
+    "@walletconnect/sign-client": "2.11.3",
+    "@walletconnect/types": "2.11.3",
```
satish-ravi, line 249: do we still need to keep these as exact versions instead of range versions?

src/walletConnect/reducer.ts
```diff
@@ -87,7 +87,7 @@ export const reducer = (
             .includes(pendingAction.topic)
         ),
         pendingSessions: state.pendingSessions.filter(
-          (pendingSession) => pendingSession.params.expiry > action.dateInSeconds
+          (pendingSession) => pendingSession.params.expiryTimestamp > action.dateInSeconds
```
satish-ravi, line 90: @MuckT codecov is complaining about this line not having tests. Maybe worth adding one given this is a more complex reducer?
Also, would this need a migration? could there be a persisted pendingSession with `expiry` instead of `expiryTimestamp`?

MuckT: I thought pending sessions where pretty ephemeral, but at second glance we might need migration otherwise the old ones may never expire.

MuckT: Cleared pending sessions with a migration in 057244a9b0e71725f532bed91d53e953ee230cf7 and tested in 04a4fbcb506947f05238199673b9f969df58e103.

package.json
MuckT: Fixed in c1c69fffdbe247f1eca6a728b4fad1aa527b6d97!

src/walletConnect/reducer.ts
satish-ravi: If the only change in schema is expiry to expiryTimestamp, then an inplace migration is fine, but if other fields are also changing, clearing seems good. Also, just for my understanding, how do pending sessions work when the app restarts?

satish-ravi (approved):

MuckT:
@satish-ravi pending sessions are added whenever there is a SESSION_PROPOSAL action and are removed whenever this proposal is accepted or denied. In the case of an application close while there are pending session it should be served when the application is reopened.

PR #5133, opened 2024-03-20
fix(jumpstart): prevent possibility of sending multiple transactions

kathaypacific:
### Description

As the title - prevent triggering the send action again while there is a transaction inflight by disabling the button while loading.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y

### Network scalability

Y


dievazqu (approved):
Thanks!

PR #5134, opened 2024-03-20
fix: ensure send flow does not crash when user has no balance

kathaypacific:
### Description

Also snuck in some onboarding warning fixes - 
- `onboardingPropsSelector` was returning a new value each time, causing unnecessary rerenders
- the pesky "unique key" warning on the CYA screen

### Test plan

A wallet with no balance should be able to enter the send amount screen

### Related issues

n/a

### Backwards compatibility

Y

### Network scalability

Y

### DIFF
--- a/src/send/SendEnterAmount.tsx
+++ b/src/send/SendEnterAmount.tsx
@@ -12,9 +12,9 @@ import EnterAmount from 'src/send/EnterAmount'
 import { lastUsedTokenIdSelector } from 'src/send/selectors'
 import { usePrepareSendTransactions } from 'src/send/usePrepareSendTransactions'
 import { COMMENT_PLACEHOLDER_FOR_FEE_ESTIMATE } from 'src/send/utils'
-import { useTokensForSend } from 'src/tokens/hooks'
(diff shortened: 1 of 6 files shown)


MuckT (approved):
Tested locally with Zero balance and no crash was observed. Also thanks for fixing that pesky developer experience in CYA.

src/send/SendEnterAmount.tsx
```diff
@@ -24,7 +24,12 @@ const TAG = 'SendEnterAmount'
 
 function SendEnterAmount({ route }: Props) {
   const { defaultTokenIdOverride, origin, recipient, isFromScan, forceTokenId } = route.params
-  const tokens = useTokensForSend()
```
dievazqu, line 27: @kathaypacific Sorry, I missed this PR.
But, I have a little concern with this...
I think we should be consistent, and if we change the logic of what tokens to show in the send flow, that should be inside `useTokensForSend`?
`useTokensForSend` is used in the TokenDetails to decide if we show the "Send" quick action, if the token appears in the send flow, I would expect to see the "Send" button, right?

PR #5136, opened 2024-03-20
fix(jumpstart): fix jumpstart claim

bakoushin:
### Description

Fix bugs causing error in claim status despite the claim itself succeeds:
* wrong assumption about `jumpstartClaim` cloud function result format
* use of non-lowercased addresses

### Test plan

* Tested manually
* Updated unit rests

### Related issues

- Related to [RET-1004](https://linear.app/valora/issue/RET-1004)

### Backwards compatibility

Y

### Network scalability

No. 

Jumpstart works only for Celo. See also RET-1019.


bakoushin:
@dievazqu yes, it was it!

PR #5140, opened 2024-03-21
test(e2e): update tests to run against tab navigator

satish-ravi:
### Description

Updates the following tests

- AccountManagement.spec.js - updates to run against both variants
- Pin.spec.js - updates to run against both variants
- Verify.spec.js - updated to run against tab navigator (no need for drawer here since there's sufficient coverage against drawer settings)
- WalletConnect.spec.js - still runs agains the drawer (hard to override statsig gate since app launched from deeplink in some cases) but reuses the `navigateToSettings` helper so its easier to update when tab is rolled out

### Test plan

CI

### Related issues

- Part of ACT-1113

### Backwards compatibility

N/A

### Network scalability

N/A

### DIFF
--- a/e2e/src/AccountManagement.spec.js
+++ b/e2e/src/AccountManagement.spec.js
@@ -9,6 +9,9 @@ describe('Account', () => {
   })
 
   describe('Support', Support)
-  describe('Settings', Settings)
-  describe('Reset Account', ResetAccount)
+  describe.each([{ navType: 'drawer' }, { navType: 'tab' }])('Settings ($navType)', Settings)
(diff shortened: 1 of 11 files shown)


e2e/src/AccountManagement.spec.js
```diff
@@ -9,6 +9,9 @@ describe('Account', () => {
   })
 
   describe('Support', Support)
```
satish-ravi, line 11: not doing a describe.each here because not all tests need to run against both navs, see Support.js for tests running against both variants

MuckT (approved):
Looks great and also doesn't look like it added that much time to the e2e tests!

PR #5141, opened 2024-03-21
feat: add new close button

MuckT:
### Description

- Adds a new `CloseButton.tsx` to match hitSlop and ripple used in the Tab Navigator buttons. The ripple seen in the Android screenshots is the touchable area.
- Adjusts the existing `headerWithCloseButton` to use the new `CloseButton.tsx`.

| Android Before w/ ripple | Android After w/ripple | iOS Before | iOS After |
| ---- | ---- | ---- | ---- |
|  |  |  |  | 

### Test plan

- Tested locally on iOS
- Tested locally on Android

### Related issues

N/A

### Backwards compatibility

Yes

### Network scalability

N/A


src/icons/Times.tsx
```diff
@@ -11,6 +11,10 @@ export interface Props {
   strokeWidth?: number
 }
 
+/**
+ * @deprecated
+ * use src/icons/X.tsx instead
```
satish-ravi, line 16: Times supports animated colors, which I believe we use at least in one place (qr screen headers, when we switch between the tabs). If X should replace this, is this something X should also support?

src/navigator/Headers.tsx
```diff
@@ -300,6 +295,17 @@ export const tabHeader: NativeStackNavigationOptions = {
   ),
 }
 
+export const headerWithCloseButton: NativeStackNavigationOptions = {
+  ...emptyHeader,
+  headerLeft: () => (
+    // The negative margin is to fix an issue with margin added via the stack navigator
+    // https://github.com/react-navigation/react-navigation/issues/11116
```
satish-ravi, line 302: this issue seems Android specific, but have the same in both iOS and Android?

MuckT: Updated docs in 4bb88204626fa33c4a905591dd76a2dbe79e6229!

src/icons/Times.tsx
MuckT: Updated the deprecated note to include a caveat for animated. I don't think we should bring this into the new X icon as it's unclear what support this has in react-native-reanimated v3 - which we should switch to at some point in the future.

jh2oman (approved):
Played around with it on android.  looks good!

src/icons/Times.tsx
MuckT: Removed in 0c069630a7fadb12517810ac4fab7d4693dfb034!

satish-ravi (approved):

src/components/CloseButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+    navigateBack()
+  }
+
+  return (
+    <View style={styles.container}>
+      <Touchable
+        testID={testID}
+        onPress={onPress}
+        style={[style, styles.button]}
+        borderRadius={Spacing.Thick24}
```
satish-ravi, line 25: this should probably be a function of size ideally? we probably won't pass a higher size for icon, but something to consider

MuckT: I played around a bit with abstracting the buttons to a shared component and ultimately decided against it in this PR and just focused on the touchable fix. The `CloseButton.tsx`, `AccountCircleButton.tsx`, `QrScanButton.tsx` and `NotificationBell.tsx` buttons should probably be refactored to use a shared component and I agree making the borderRadius a function of size would make sense there.

PR #5142, opened 2024-03-21
test(e2e): add new account creation test for new tab and cloud backup

satish-ravi:
### Description

Renames existing NewAccountOnboarding test to NewAccountOnboardingDrawer and updates NewAccountOnboarding to run against tab navigator and cloud backup

### Test plan

CI

### Related issues

- Part of ACT-1113

### Backwards compatibility

N/A

### Network scalability

N/A

### DIFF
--- a/e2e/src/usecases/NewAccountOnboarding.js
+++ b/e2e/src/usecases/NewAccountOnboarding.js
@@ -122,7 +122,7 @@ export default NewAccountOnboarding = () => {
       .withTimeout(10 * 1000)
   })
 
-  it('Account Address shown in drawer menu', async () => {
+  it('Account Address shown in profile / menu', async () => {
     await waitForElementByIdAndTap('WalletHome/AccountCircle')
     await scrollIntoView('Account Address', 'SettingsScrollView')
     const accountAddressElement = await element(by.id('AccountNumber')).getAttributes()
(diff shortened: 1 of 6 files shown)


e2e/src/usecases/NewAccountOnboardingDrawer.js
satish-ravi, line 1: This is exactly NewAccountOnboarding.js before the above changes

package.json
```diff
@@ -262,6 +262,7 @@
     "eslint-plugin-react-hooks": "^4.6.0",
     "eslint-plugin-react-native": "^4.0.0",
     "ethers": "^5.7.2",
+    "expect": "^29.7.0",
```
satish-ravi, line 265: fixes a knip error. We use expect directly in e2e tests. It was previously a transitive dependency from jest which knip doesn't allow https://knip.dev/guides/handling-issues#unlisted-dependencies

e2e/src/usecases/NewAccountOnboarding.js
```diff
@@ -121,20 +123,22 @@ export default NewAccountOnboarding = () => {
   })
 
   it('Account Address shown in drawer menu', async () => {
```
jh2oman, line 125: nit: no longer drawer menu.

jh2oman (approved):

PR #5143, opened 2024-03-22
feat(multi-select): New UI component for selecting multiple networks

jh2oman:
### Description

Re-usable component for doing multi selection based on the designs for selecting multiple Networks.

TODO(next): Use the new filter in Swap & Dapp filters per designs

[Figma designs
](https://www.figma.com/file/gFuRy5HJrV34ehkY0dvWWb/Swap-guidance?type=design&node-id=2632-4282&mode=design&t=xbvdPqAXfbaUiK4f-0)

#### Tapping around

[tapping.webm](https://github.com/valora-inc/wallet/assets/8432644/774a9fc8-02bc-4a91-bf1e-8b67c222c46c)

#### Close / Open

[closing.webm](https://github.com/valora-inc/wallet/assets/8432644/809f9c4d-bbff-4416-ba28-866f4b6f9cad)

#### Lots of networks

[many-networks.webm](https://github.com/valora-inc/wallet/assets/8432644/a0bf9d5a-1b8e-4e4e-a115-1a2dc4103eb6)

#### Long network names

Screenshot_20240321_135129

### Test plan

Unit tests

### Related issues

https://linear.app/valora/issue/ACT-1098/select-token-by-network

### Network scalability

Built to handle more networks, no-code required for more networks added


src/components/multiSelect/MultiSelectBottomSheet.tsx
```diff
@@ -0,0 +1,271 @@
...
+          setSelectedItems({ ...noItemSelected, [itemId]: isSelected })
+        } else {
+          setSelectedItems({ ...selectedItems, [itemId]: !isSelected })
+        }
+      },
+    })
+  )
+
+  return (
+    <GorhomBottomSheet
```
satish-ravi, line 117: can we reuse BottomSheet or BottomSheetBase components? seems like some of the things are duplicated here, like handleAnimate, renderBackdrop, etc

src/components/multiSelect/MultiSelectBottomSheet.tsx
```diff
@@ -0,0 +1,271 @@
...
+  itemContainer: {
+    flexDirection: 'column',
+  },
+  item: {
+    height: ITEM_HEIGHT,
+    flexGrow: 1,
+    alignItems: 'center',
+    justifyContent: 'center',
+    paddingVertical: 5,
+    paddingHorizontal: 24,
```
satish-ravi, line 239: use Spacing constants where possible

src/components/multiSelect/MultiSelectBottomSheet.tsx
```diff
@@ -0,0 +1,271 @@
...
+  )
+}
+
+export interface ItemProps {
+  onPress?: () => void
+  text: string
+  isSelected: boolean
+  iconUrl?: string
+}
+function renderItem({ onPress, text, iconUrl, isSelected }: ItemProps) {
```
satish-ravi, line 163: nit: could you make this a react component. E.g., function Item(...) { }, and then do `<Item ...>` above, think it will make the above a bit more readable.

src/components/multiSelect/MultiSelectBottomSheet.tsx
```diff
@@ -0,0 +1,271 @@
...
+const ITEM_HEIGHT = 60
+const MAX_ITEMS_IN_VIEW = 5
+
+interface MultiSelectBottomSheetProps<T extends string> {
+  forwardedRef: React.RefObject<GorhomBottomSheet>
+  onChange?: BottomSheetProps['onChange']
+  onClose?: () => void
+  onOpen?: () => void
+  handleComponent?: BottomSheetProps['handleComponent']
+  selectedItems: Record<T, boolean>
```
satish-ravi, line 24: my initial thought was this was a list of selected items, and I was searching for a prop containing all items before realizing this includes all items with a true / false. I was looking at some other packages to see what interface they support (e.g., https://www.npmjs.com/package/react-multi-select-component), could we adopt something similar? I guess you can also infer all items from `textAndIconMap` since it should include all keys for rendering the item, which can act analogous to the options param in the package. I think you can also avoid the `as T[]` casting if you change the shape of `textAndIconMap` slightly. Or is there a strong reason to have consumers pass in an object for selectedItems instead of an array?

src/components/multiSelect/NetworkMultiSelectBottomSheet.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+  const textAndIconMap = allNetworkIds.reduce(
+    (acc, networkId) => {
+      acc[networkId] = {
+        text: NETWORK_NAMES[networkId],
+        iconUrl: networkIconByNetworkId[networkId],
+      }
+      return acc
+    },
+    {} as Record<NetworkId, Omit<ItemProps, 'onPress' | 'isSelected'>>
```
satish-ravi, line 40: this same Omit is done in a couple of places, maybe define a type for Item and then have ItemProps extend it?

src/components/multiSelect/MultiSelectBottomSheet.tsx
jh2oman: An object just made more sense to me when I was initially thinking about it. Static options, O(1) for updates to the object since we always know the keys. The difference between that an and o(n) array.filter to update an array of items is really unimportant though at this scale.

I can update to an array similar to the props that you linked!

src/components/multiSelect/MultiSelectBottomSheet.tsx
jh2oman: I'll try and see if any sharing can be done.

src/components/multiSelect/MultiSelectBottomSheet.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+interface MultiSelectBottomSheetProps<T extends string> {
+  forwardedRef: React.RefObject<GorhomBottomSheet>
+  onChange?: BottomSheetProps['onChange']
+  onClose?: () => void
+  onOpen?: () => void
+  handleComponent?: BottomSheetProps['handleComponent']
+  options: Option[]
+  selectedOptions: Option[]
+  setSelectedOptions: (selectedOptions: Option[]) => void
```
satish-ravi, line 26: these can just be `string[]` or `T[]` and just include the id right? Is there a reason to pass around all the fields in Option? Using Sets are also an option if we wanted to avoid duplicates

src/components/multiSelect/MultiSelectBottomSheet.tsx
```diff
@@ -0,0 +1,232 @@
...
+  )
+}
+
+interface OptionProps {
+  onPress?: () => void
+  text: string
+  isSelected: boolean
+  iconUrl?: string
+}
+function Option({ onPress, text, iconUrl, isSelected }: OptionProps) {
```
satish-ravi, line 124: I guess only the interface Option is exported, but there can be potential collisions of this also needs to be exported.

src/components/multiSelect/MultiSelectBottomSheet.tsx
```diff
@@ -0,0 +1,232 @@
...
+          </View>
+        </View>
+      </Touchable>
+    </View>
+  )
+}
+
+const styles = StyleSheet.create({
+  doneButton: {
+    borderRadius: 16,
```
MuckT, line 157:
```suggestion
    borderRadius: Spacing.Regular16,
```

src/components/multiSelect/MultiSelectBottomSheet.tsx
```diff
@@ -0,0 +1,232 @@
...
+            <Option
+              text={selectAllText}
+              isSelected={isEveryOptionSelected}
+              onPress={() => setSelectedOptions(options)}
+            />
+            {options.map((option) => (
+              <Option
+                text={option.text}
+                iconUrl={option.iconUrl}
+                isSelected={
```
MuckT, line 83: This all select behavior differs from the [material design standards](https://m3.material.io/components/checkbox/accessibility) on parent check boxes. Also left a comment in [Figma](https://www.figma.com/file/gFuRy5HJrV34ehkY0dvWWb?type=design&node-id=2632-4282&mode=design#745860228).

jh2oman: waiting on nitya's response. I'd prefer if it didn't block the pr :)

src/components/multiSelect/MultiSelectBottomSheet.tsx
```diff
@@ -0,0 +1,232 @@
...
+              text={selectAllText}
+              isSelected={isEveryOptionSelected}
+              onPress={() => setSelectedOptions(options.map((option) => option.id))}
+            />
+            {options.map((option) => (
+              <OptionLineItem
+                text={option.text}
+                iconUrl={option.iconUrl}
+                isSelected={
+                  !!selectedOptions.find((selectedOption) => selectedOption === option.id) &&
```
satish-ravi, line 84:
```suggestion
                  !!selectedOptions.includes(option.id) &&
```

src/components/multiSelect/MultiSelectBottomSheet.tsx
```diff
@@ -0,0 +1,232 @@
...
+                onPress={() => {
+                  if (isEveryOptionSelected) {
+                    setSelectedOptions([option.id])
+                  } else {
+                    if (selectedOptions.find((selectedOption) => selectedOption === option.id)) {
+                      setSelectedOptions(
+                        selectedOptions.filter((selectedOption) => selectedOption !== option.id)
+                      )
+                    } else {
+                      setSelectedOptions([...selectedOptions, option.id])
```
satish-ravi, line 96: could there be a race if 2 elements are clicked really fast? chatgpt suggests something like this
```ts
const toggleOption = (option) => {
  setSelectedOptions(prevSelectedOptions => {
    const isSelected = prevSelectedOptions.includes(option);

    if (isSelected) {
      return prevSelectedOptions.filter((item) => item !== option);
    } else {
      return [...prevSelectedOptions, option];
    }
  });
};

jh2oman: Ah interesting! I've never used passing a function into the `setState` function as an updater.

jh2oman: This actually forced a much nicer unit tests!

satish-ravi (approved):
A once build is green

PR #5146, opened 2024-03-22
fix(tab nav): fix navigating back from language selection

finnian0826:
### Description

issue found in bugapalooza when changing language. Fixed so now you don't see back arrow + hamburger when in new tab experience. 

### Test plan

Manual test changing language:

https://github.com/valora-inc/wallet/assets/140328381/96c727ce-1b4a-44ff-ab89-68510fc097ad

Testing other paths:

https://github.com/valora-inc/wallet/assets/140328381/8f951b09-f204-4f8c-8f3a-977aa30ccf16

### Related issues

- N/A

### Backwards compatibility

Yes

### Network scalability

- N/A


src/navigator/types.tsx
```diff
@@ -265,15 +265,15 @@ export type StackParamList = {
...
+  [Screens.Settings]: { promptConfirmRemovalModal?: boolean } | undefined
+  [Screens.SettingsDrawer]: { promptConfirmRemovalModal?: boolean } | undefined
   [Screens.SignInWithEmail]: {
     keylessBackupFlow: KeylessBackupFlow
   }
   [Screens.Spend]: undefined
   [Screens.StoreWipeRecoveryScreen]: undefined
-  [Screens.Support]: { isTabNav?: boolean } | undefined
-  [Screens.SupportDrawer]: { isTabNav?: boolean } | undefined
+  [Screens.Support]: {} | undefined
```
finnian0826, line 275: having `{} | undefined` feels janky but I needed it to avoid a build error, is there a nicer way to do this (i.e. to have `undefined` but still be able to use `route` in the component)?

src/account/Support.test.tsx
```diff
@@ -12,7 +12,7 @@ import { createMockStore, getMockStackScreenProps } from 'test/utils'
 const renderSupport = () =>
   render(
     <Provider store={createMockStore()}>
-      <Support {...getMockStackScreenProps(Screens.Support)} />
```
finnian0826, line 15: Since we weren't passing `isTabNav: true`, this was actually using the SupportDrawer screen, so snapshot was looking to match that.

src/account/Settings.test.tsx
```diff
@@ -470,7 +470,7 @@ describe('Account', () => {
       body: '{"phoneNumber":"+14155550000","clientPlatform":"android","clientVersion":"0.0.1"}',
     })
     expect(navigate).toHaveBeenLastCalledWith(Screens.BackupPhrase, {
-      settingsScreen: Screens.SettingsDrawer,
+      settingsScreen: Screens.Settings,
```
satish-ravi, line 473: can we keep one test for SettingsDrawer and another for Settings?

src/backup/BackupComplete.tsx
```diff
@@ -32,7 +32,6 @@ function BackupComplete({ route }: Props) {
         // TODO(ACT-1133): change settingsScreen props to isAccountRemoval boolean to know if we need to go back to settings to show promptConfirmRemovalModal
         navigate(settingsScreen === Screens.Settings ? Screens.Settings : Screens.SettingsDrawer, {
```
satish-ravi, line 33:
```suggestion
        navigate(settingsScreen, {
```

src/backup/BackupQuiz.tsx
```diff
@@ -81,10 +81,7 @@ export const navOptionsForQuiz = ({ route }: OwnProps) => {
         <CancelButton
           onCancel={() =>
             navigate(
-              settingsScreen === Screens.Settings ? Screens.Settings : Screens.SettingsDrawer,
-              {
-                isTabNav: settingsScreen === Screens.Settings,
-              }
+              settingsScreen === Screens.Settings ? Screens.Settings : Screens.SettingsDrawer
```
satish-ravi, line 84:
```suggestion
              settingsScreen
```

src/navigator/types.tsx
satish-ravi: The build error you're referring to is the one on DrawerNavigator right? I think we've had similar issues with other components, see other drawer screens that have a @ts-ignore. I think you can just keep this type in the SupportDrawer or add ts-ignore, as we'll be cleaning up the drawer navigator soon anyway

src/backup/BackupComplete.tsx
finnian0826: doing this I get the following error:
```
Argument of type '[Screens, { promptConfirmRemovalModal: true; }]' is not assignable to parameter of type '[keyof StackParamList] | [keyof StackParamList, {} | { settingsScreen: Screens | undefined; } | { nextScreen: keyof StackParamList; } | ... 68 more ... | undefined]'.
  Type '[Screens, { promptConfirmRemovalModal: true; }]' is not assignable to type '[keyof StackParamList, {} | { settingsScreen: Screens | undefined; } | { nextScreen: keyof StackParamList; } | { activeAssetTab: AssetTabType; isWalletTab?: boolean | undefined; } | ... 67 more ... | undefined]'.
    Type at position 0 in source is not compatible with type at position 0 in target.
      Type 'Screens' is not assignable to type 'keyof StackParamList'.
```

src/backup/BackupQuiz.tsx
finnian0826: with this I get the same error as mentioned above

src/backup/BackupComplete.tsx
satish-ravi: Ah, I think you can avoid this if you narrow the type of settingsScreen to `Screens.Settings | Screens.SettingsDrawer`

src/navigator/types.tsx
satish-ravi: would be good to updated type of `Support` to just be `undefined` and keep the weird `{} |` for SettingsDrawer right now

src/account/Settings.test.tsx
satish-ravi: I actually meant keeping one of the existing delete tests for drawer and one for non drawer like how it was prior to the original change. PN being verified / unverified is independent of this screen. We don't need all combinations, but also doesn't hurt except for a tiny bump in CI time.

satish-ravi (approved):

MuckT (approved):

PR #5147, opened 2024-03-25
refactor(tokens): update networksIconSelector to return all icons

satish-ravi:
### Description

Accepting network ids as an array would mean that the selector recomputes the result even if the same list of network ids are passed and returns a new result, resulting in unnecessary rerenders. This updates it so that the selector returns all network id's icons. An alternative approach would be to set `memoizeOptions` like `tokenIdsSelector`, but there doesn't seem to be any harm in returning all network's icon in the function. The consumer of this selector can pick the required networks' icon from the object

### Test plan

CI, ensured the below warning doesn't show in #5148 when merged with this change

### Related issues

N/A

### Backwards compatibility

Yes

### Network scalability


kathaypacific (approved):
✨

PR #5148, opened 2024-03-25
feat(feed): include network icon in tx feed items

satish-ravi:
### Description

Made a reusable component to create an icon with a badge and used for all icons in the tx feed

### Test plan

Manually

| Pending, send/receive | Nfts and swaps |
|:--------:|:--------:|
|  |  | 

### Related issues

- Fixes ACT-105

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/src/components/IconWithNetworkBadge.tsx
+++ b/src/components/IconWithNetworkBadge.tsx
@@ -11,7 +11,7 @@ interface Props {
 }
 
 export default function IconWithNetworkBadge({ networkId, children }: Props) {
-  const networkIcons = useSelector((state) => networksIconSelector(state, [networkId]))
+  const networkIcons = useSelector(networksIconSelector)
   const networkImageUrl = networkIcons[networkId]
   return (
     <View>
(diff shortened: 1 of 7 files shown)


src/components/IconWithNetworkBadge.tsx
```diff
@@ -0,0 +1,41 @@
...
+        />
+      )}
+    </View>
+  )
+}
+
+const styles = StyleSheet.create({
+  networkIconStyle: {
+    position: 'absolute',
+    width: 16,
```
MuckT, line 35:
```suggestion
    width: Spacing.Regular16,
```

MuckT: Unclear if we want to using Spacing for non margin / padding numbers. I am fine with this being hardcoded so the icon doesn't change if spacing across the app changes.

MuckT (approved):
🎨 Looks great to me - tested on iOS and Android.

PR #5149, opened 2024-03-25
fix(jumpstart): generate link outside of firebase sdk

kathaypacific:
### Description

As the title.

### Test plan

Tested that the generated link is exactly the same as before this change.

### Related issues

n/a

### Backwards compatibility

n/a

### Network scalability

n/a


src/firebase/dynamicLinks.ts
```diff
@@ -26,8 +26,22 @@ export async function createInviteLink(address: string) {
...
-  })
+  // avoid calling firebase sdk with private key during link creation to protect
+  // the private key from being stored
+  const dynamicLink = new URL(
+    await dynamicLinks().buildLink({
+      ...commonDynamicLinkParams,
+      link: WEB_LINK,
+    })
+  )
+  const dynamicUrl = new URL(dynamicLink)
```
jophish, line 37: maybe i'm misunderstanding something here, but do we need to wrap this with another `new URL` constructor? isn't this essentially `new URL(new URL(stuff...))`?

kathaypacific: oops, i meant to remove that. thanks for the catch!

jophish (approved):

src/firebase/dynamicLinks.ts
```diff
@@ -26,8 +26,20 @@ export async function createInviteLink(address: string) {
...
+  dynamicUrl.searchParams.set('link', `${WEB_LINK}jumpstart/${privateKey}/${networkId}`)
+
+  // the firebase dynamic links sdk encodes dots and dashes even though it is
+  // not strictly required for urls. calling searchParams.set seems to transform
+  // __all__ search params to be url encoded, where dots and dashes are no
+  // longer encoded. This is probably okay, but to be extra safe we will put
+  // back the encoding ourselves.
+  const searchParams = dynamicUrl.search.replace(/\./g, '%2E').replace(/-/g, '%2D')
+  return `${dynamicUrl.origin}/${searchParams}`
```
jeanregisser, line 44: Could we "not" be extra safe and remove this?
Feels a bit surprising we're only doing it for dots and dashes.

jeanregisser (approved):

src/firebase/dynamicLinks.ts
kathaypacific: i thought about this and tried consuming a link without the encoding on a simulator and it does work, but since the firebase sdk were producing links that are encoded in this way i thought it was safest to keep it incase somewhere in the sdk when consuming deep links on some specific devices, the encoding becomes important :/ since i don't understand why they did the encoding in the first place, i was thinking it wasn't worth the risk of removing it...at least this mess is well documented and inside this function?

PR #5150, opened 2024-03-25
chore(vulnerability): update vulnerabilities list for the new web3 issue

jh2oman:
### Description

https://github.com/advisories/GHSA-87qp-7cw8-8q9c

~The most susceptible environments for this attack are Web servers, application servers, and web browsers so I think we are relatively safe in the wallet. Nevertheless this gives us another incentive to remove more old wallet code~

~New vulnerability with web3 utils. Our only use of web3 is with contractkit and we are 3 major versions behind. I created this placeholder project for removing all `@celo` dependencies and have linked this PR there.~

It appears that we are on such an old version of the code that these utils don't exist yet 😅 . @satish-ravi and I could find no object manipulation helpers in the version we use of web3. The security vulnerability likely is mislabeled and should have a lower bound on the affected versions.

### Related issues

https://linear.app/valora/project/remove-all-celo-dependencies-in-favor-of-viem-a7e601818b8e/ENG


satish-ravi (commented):
Thanks for doing this. I was also reading about this, and it seems like the version we use (1.10.4), doesn't even have these methods? https://github.com/web3/web3.js/tree/v1.10.4/packages/web3-utils/src I don't see a `format` or `mergeDeep`

satish-ravi (approved):

PR #5163, opened 2024-03-26
feat(escrow): Adding analytics to the reclaim flow.

dievazqu:
### Description

As title says

### DIFF
--- a/src/analytics/Events.tsx
+++ b/src/analytics/Events.tsx
@@ -663,6 +663,13 @@ export enum JumpstartEvents {
   jumpstart_claim_loading_dismissed = 'jumpstart_claim_loading_dismissed',
   jumpstart_claim_error_dismissed = 'jumpstart_claim_error_dismissed',
   jumpstart_claim_error_contact_support = 'jumpstart_claim_error_contact_support',
+  jumpstart_reclaim_press = 'jumpstart_reclaim_press',
+  jumpstart_reclaim_start = 'jumpstart_reclaim_start',
+  jumpstart_reclaim_failed = 'jumpstart_reclaim_failed',
+  jumpstart_reclaim_succeeded = 'jumpstart_reclaim_succeeded',
+  jumpstart_reclaim_dismiss_error = 'jumpstart_reclaim_dismiss_error',
+  jumpstart_reclaim_fetching_success = 'jumpstart_reclaim_fetching_success',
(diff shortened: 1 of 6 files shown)


src/analytics/Events.tsx
```diff
@@ -663,6 +663,13 @@ export enum JumpstartEvents {
   jumpstart_claim_loading_dismissed = 'jumpstart_claim_loading_dismissed',
   jumpstart_claim_error_dismissed = 'jumpstart_claim_error_dismissed',
   jumpstart_claim_error_contact_support = 'jumpstart_claim_error_contact_support',
+  jumpstart_reclaim_press = 'jumpstart_reclaim_press',
+  jumpstart_reclaim_start = 'jumpstart_reclaim_start',
+  jumpstart_reclaim_failed = 'jumpstart_reclaim_failed',
+  jumpstart_reclaim_succeeded = 'jumpstart_reclaim_succeeded',
+  jumpstart_reclaim_dismiss_error = 'jumpstart_reclaim_dismiss_error',
+  jumpstart_reclaim_fetching_success = 'jumpstart_reclaim_fetching_success',
```
bakoushin, line 671: what do you think if we add the word `status` here to make it more explicit what this fetching is about?

src/analytics/Properties.tsx
```diff
@@ -1553,6 +1553,23 @@ interface JumpstartEventsProperties {
   [JumpstartEvents.jumpstart_claim_loading_dismissed]: undefined
   [JumpstartEvents.jumpstart_claim_error_dismissed]: undefined
   [JumpstartEvents.jumpstart_claim_error_contact_support]: undefined
+  [JumpstartEvents.jumpstart_reclaim_press]: { networkId: NetworkId; transactionHash: string }
+  [JumpstartEvents.jumpstart_reclaim_start]: { networkId: NetworkId; transactionHash: string }
+  [JumpstartEvents.jumpstart_reclaim_failed]: undefined
```
bakoushin, line 1558: just noticed that only those two are missing `networkId` and `transactionHash`, while others have it. what do you think about sending this information with every signle event? i see one pro: we could easily track every single hash trough those events.

bakoushin (approved):

src/analytics/Properties.tsx
dievazqu: Yeah, I didn't add it because I didn't have the information in the saga. But I guess I can pass it.

PR #5169, opened 2024-03-27
chore: add fees to send confirmation

MuckT:
### Description

Adds the line item for local amount display of fees on the send confirmation screen.

| Before iOS stCelo | Before iOS ETH | After iOS stCELO | After iOS ETH |
| ---- | ---- | ---- | ---- |
|  |  |  |  |

### Test plan

- Tested locally on iOS
- Tested locally on Android
- Unit test snapshot updated

### Related issues

- Fixes ACT-1135

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/src/send/SendConfirmation.tsx
+++ b/src/send/SendConfirmation.tsx
@@ -149,38 +149,36 @@ function SendConfirmation(props: Props) {
   const FeeContainer = () => {
     return (
       <View style={styles.feeContainer}>
-        <>
-          <LineItemRow
-            testID="SendConfirmation/fee"
-            title={t('feeEstimate')}
-            amount={
-              maxFeeAmount && (
-                <TokenDisplay
-                  amount={maxFeeAmount}
-                  tokenId={feeTokenInfo?.tokenId}
-                  showLocalAmount={false}
-                />
-              )
-            }
-            isLoading={!maxFeeAmount}
-          />
-          <LineItemRow
-            testID="SendConfirmation/localFee"
-            title=""
-            textStyle={{ ...fontStyles.small }}
-            style={{ marginVertical: 0 }}
-            amount={
-              maxFeeAmount && (
-                <TokenDisplay
-                  amount={maxFeeAmount}
-                  tokenId={feeTokenInfo?.tokenId}
-                  showLocalAmount={true}
-                />
-              )
-            }
-            isLoading={!maxFeeAmount}
-          />
-        </>
+        <LineItemRow
+          testID="SendConfirmation/fee"
+          title={t('feeEstimate')}
+          textStyle={{ ...typeScale.bodyMedium }}
+          amount={
+            maxFeeAmount && (
+              <TokenDisplay
+                amount={maxFeeAmount}
+                tokenId={feeTokenInfo?.tokenId}
+                showLocalAmount={false}
+              />
+            )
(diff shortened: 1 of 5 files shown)


src/send/SendConfirmation.tsx
```diff
@@ -373,6 +388,13 @@ const styles = StyleSheet.create({
     color: colors.infoDark,
     paddingRight: 8,
   },
+  subHeading: {
+    marginVertical: 0,
+  },
+  subHeadingText: {
+    ...fontStyles.small,
```
MuckT, line 395: Used deprecated fontStyles as mock aren't updated and to match subHeading of line below.

src/send/SendConfirmation.tsx
```diff
@@ -152,6 +152,7 @@ function SendConfirmation(props: Props) {
         <LineItemRow
           testID="SendConfirmation/fee"
           title={t('feeEstimate')}
+          textStyle={{ ...typeScale.bodyMedium }}
```
finnian0826, line 155: Why is this being added? Is it different from what it was defaulting to or just ensuring that it is this?

src/send/SendConfirmation.tsx
finnian0826: Do we want to add a clean up note here or check with Nitya about changing the mocks? I worry that if we add the deprecated style here in the future when we change this screen we'll be in the same boat.

finnian0826 (approved):
Left a couple questions but otherwise lgtm

src/send/SendConfirmation.tsx
MuckT: It defaults to `fontStyles.regular`. This was closer to the mocks in regards to font size.

src/send/SendConfirmation.tsx
MuckT: @finnian0826 unfortunately It's already used in several other places throughout `src/send/SendConfirmation.tsx`; however, I do think it's worthwhile to call out uses of the old fonts on new screens and new mocks.

src/send/SendConfirmation.tsx
satish-ravi: does this need a color too? I don't think the new styles default a color like the old ones.

src/send/SendConfirmation.tsx
satish-ravi: is there no equivalent style in typescale? Or is the concern that there's multiple other uses and hard to update now?

src/send/SendConfirmation.tsx
MuckT: `src/components/LineItemRow.tsx` has a default text color set to `Colors.black` and the subtext is set to `colors.gray4`.

src/send/SendConfirmation.tsx
MuckT: My main concern is that there are several usages of `fontStyle` within the file. I'll switch them over to typeScale and see if I can get a similar appearance.

MuckT: Was able to update the styles in 55bcde12049b9f9b5fd74fe7e4c6bcdc5f3b2ee1! The snapshots were pretty helpful in the case to make sure the new fonts were as close to the old ones as possible.

PR #5174, opened 2024-03-27
feat(swaps): add dropdown network filter

jh2oman:
### Description

Dropdown for filtering swap tokens by network

[Figma](https://www.figma.com/file/gFuRy5HJrV34ehkY0dvWWb/Swap-guidance?type=design&node-id=2708-3698&mode=design&t=l0AoyI5xCPHiauyq-0)

[Android](https://github.com/valora-inc/wallet/assets/8432644/7aaf789f-904b-4a06-988a-a3a8fae23713)

[iOS](https://github.com/valora-inc/wallet/assets/8432644/98ff2060-f4fa-446f-98e3-b6b8de4d7496)

### Test plan

Updated unit tests

### Related issues

https://linear.app/valora/issue/ACT-1098/select-token-by-network

### Network scalability


src/components/TokenBottomSheet.tsx
```diff
@@ -160,23 +166,52 @@ function TokenBottomSheet<T extends TokenBalance>({
...
+          }
         }
         return chip
       })
     })
   }
 
+  const handleToggleFilterChip = (toggledChip: FilterChip<TokenBalance>) => {
+    if (isNetworkChip(toggledChip)) {
+      networkChipRef.current?.snapToIndex(0)
```
satish-ravi, line 197: this will never be invoked on close right? like the boolean filters?

src/components/multiSelect/MultiSelectBottomSheet.tsx
```diff
@@ -26,6 +26,7 @@ export interface MultiSelectBottomSheetProps<T extends string> {
   setSelectedOptions: Dispatch<SetStateAction<T[]>>
   selectAllText: string
   title: string
+  allOrOne?: boolean
```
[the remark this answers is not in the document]
jh2oman: It was meant to be that you can only select one or select all. Single select isn't as accurate, but does get the point across better

src/components/TokenBottomSheet.tsx
jh2oman: It's only called when the chip is pressed.

src/components/multiSelect/MultiSelectBottomSheet.tsx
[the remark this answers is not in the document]
satish-ravi: not tied to `singleSelect`, open to other alternatives. but definitely 👎 for `allOrOne` as a boolean value for it doesn't exactly say whether its all or one.

locales/base/translation.json
```diff
@@ -1928,7 +1928,8 @@
       "myTokens": "My tokens",
       "popular": "Popular",
       "recentlySwapped": "Recently swapped",
-      "network": "{{networkName}} Network"
+      "network": "{{networkName}} Network",
+      "selectNetwork": "Network"
```
satish-ravi, line 1932: I think we've been doing version suffixes for new version of the same key, but this works too

src/components/TokenBottomSheet.tsx
```diff
@@ -42,8 +48,8 @@ export interface TokenBottomSheetProps<T extends TokenBalance> {
   tokens: T[]
   TokenOptionComponent?: React.ComponentType<TokenOptionProps>
   showPriceUsdUnavailableWarning?: boolean
-  filterChips?: FilterChip<TokenBalance>[]
   areSwapTokensShuffled?: boolean
+  filterChips?: FilterChip<TokenBalance>[]
```
satish-ravi, line 52: revert this, so git blame doesn't flag this line on this commit 😄

src/swap/useFilterChips.ts
```diff
@@ -23,20 +22,6 @@ export default function useFilterChip(selectingField: Field | null): FilterChip<
   ).popularTokenIds
   const supportedNetworkIds = getSupportedNetworkIdsForSwap()
 
-  const networkIdFilters =
-    supportedNetworkIds.length > 1
```
satish-ravi, line 27: should we retain this condition? seems very unlikely we go back to a single network though

satish-ravi (approved):

src/swap/useFilterChips.ts
jh2oman: Yea, i thought the additional code for a *slightly* better experience wasn't worth it since, like you said, seems very unlikely we will return to single network.

PR #5175, opened 2024-03-27
feat: Add icon overlay with one Eth asset on Arbitrum or Optimism

chriskeating603:
### Description

Currently, when a user has one balance of Eth, users cannot tell the difference between on different networks (Ethereum, Arbitrum, Optimism). This change add an network icon over the token icon for Eth when held on either Arb or Op. 

### Test plan

Tests for presence of TokenIcon and presence / absence of NetworkIcon depending on network, and passes all tests. 

### Related issues

- Fixes ACT-1086

### Backwards compatibility

Passes all previous tests

### Network scalability

If a new NetworkId and/or Network are added in the future, the changes in this PR will continue to work without code changes but will require additional testing for new changes


ios/.xcode.env
MuckT, line 1: I don't think there's harm in committing this file: [ios/.xcode.env](https://github.com/valora-inc/wallet/pull/5175/files#diff-0ac28f6822aef5dd17d112ca8416b314ee808b0fbd1dc9b0139fbe749e81ee27), but it would be good to check with others.

src/components/TokenIcon.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react'
-import { StyleProp, StyleSheet, Text, View, ViewStyle } from 'react-native'
+import { FlexStyle, StyleProp, StyleSheet, Text, View, ViewStyle } from 'react-native'
```
MuckT, line 2: Can be removed when unused prop is removed.

ios/.xcode.env
chriskeating603: Oh I can remove that if it is not standard to add to this file

PR #5200, opened 2024-04-02
chore(jumpstart): support for reclaim from retired jumpstart contracts

kathaypacific:
### Description

This PR adds a `retiredContractAddresses` to the jumpstart dynamic config.

If we need to change the contract for whatever reason, we should still allow users to reclaim their funds that were sent to previous versions of the contract. The only place we need to do this is in the transaction classification stage - we should mark transactions to _any_ jumpstart contract appropriately, from there the jumpstart transaction details screen will allow reclaim from that contract directly.

Of course this logic relies on the abi of the jumpstart contracts staying the same, since a single abi is baked into the app. 

### Test plan

Unit tests

### Related issues

n/a

### Backwards compatibility

Y

### Network scalability

Y

### DIFF
--- a/src/transactions/feed/TransferFeedItem.tsx
+++ b/src/transactions/feed/TransferFeedItem.tsx
@@ -99,10 +99,15 @@ function TransferFeedItem({ transfer }: Props) {
 }
 
 function isJumpstartTransaction(tx: TokenTransfer) {
-  const jumpstartAddress = getDynamicConfigParams(
+  const jumpstartConfig = getDynamicConfigParams(
     DynamicConfigs[StatsigDynamicConfigs.WALLET_JUMPSTART_CONFIG]
-  ).jumpstartContracts[tx.networkId]?.contractAddress
-  return tx.address === jumpstartAddress
+  ).jumpstartContracts[tx.networkId]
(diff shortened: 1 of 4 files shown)


src/transactions/feed/TransferFeedItem.tsx
```diff
@@ -99,10 +99,15 @@ function TransferFeedItem({ transfer }: Props) {
...
-  const jumpstartAddress = getDynamicConfigParams(
+  const jumpstartConfig = getDynamicConfigParams(
     DynamicConfigs[StatsigDynamicConfigs.WALLET_JUMPSTART_CONFIG]
-  ).jumpstartContracts[tx.networkId]?.contractAddress
-  return tx.address === jumpstartAddress
+  ).jumpstartContracts[tx.networkId]
+  const jumpstartAddresses = [
+    jumpstartConfig?.contractAddress,
+    ...(jumpstartConfig?.retiredContractAddresses ?? []),
+  ].filter((address): address is string => !!address)
```
jeanregisser, line 108: nit: we have a helper for this.
```suggestion
  ].filter(isPresent)
```

jeanregisser (approved):
Great! 👍

PR #5204, opened 2024-04-04
fix: xcode 15.3 build issues by upgrading Flipper version

kathaypacific:
### Description

This PR fixes the iOS build issue with xcode 15.3 by specifying an explicit Flipper version in the podfile which contains the [fix needed](https://github.com/facebook/flipper/commit/b3dcdb87f930dbbc9dbacb53ad60996e0111e7d8). [Context](https://valora-app.slack.com/archives/C025V1D6F3J/p1712220215515279).

Android:
it felt bad to leave the android flipper version at 0.125.0, but 0.250.0 does not work on Android. among other things, the flipper dependencies which were synced are [no longer synced after 0.182.0](https://stackoverflow.com/a/76222262). so i've just bumped the version to 0.182.0, which is also what RN 72 uses.

react-native-flipper:
it looks like we are using 2 versions - 0.127 in the dev dependencies, and 0.212 as a resolution. i'm not sure why this is, we don't need both. the version used in the dev dependencies is the one that is used to generate the Pods - when using version 0.212, the iOS release builds failed locally and in the CI. the failure was something to do with Flipper (https://github.com/valora-inc/wallet/actions/runs/8554301604/job/23439350728). since we were not doing anything with version 0.212 really, i removed the resolution so that we only have 1 version.

### Test plan

Tested that iOS and Android builds succeed and Flipper works.

### Related issues

n/a

### Backwards compatibility

Y

### Network scalability

Y


ios/Podfile
```diff
@@ -64,7 +64,10 @@ target "celo" do
         # otherwise the main app fails to link (likely due to the 2 dynamic libraries with the same name)
         # Flipper-Glog >= 0.3.7 uses a vendored dynamic library
         # See https://github.com/CocoaPods/Specs/blob/86d45638f249b0cb766c8e674a98b1b504b1ac0e/Specs/2/7/2/Flipper-Glog/0.3.7/Flipper-Glog.podspec.json#L21
-        'Flipper-Glog' => '0.3.5' 
+        'Flipper-Glog' => '0.3.5',
+        # Use a specific version of Flipper which supports the Xcode 15.3
+        # toolchain: https://github.com/facebook/flipper/commit/b3dcdb87f930dbbc9dbacb53ad60996e0111e7d8
+        'Flipper' => '0.250.0',
```
jeanregisser, line 70: nit: apparently the [latest version](https://github.com/CocoaPods/Specs/blob/master/Specs/a/e/a/Flipper/0.251.0/Flipper.podspec.json) is this:
```suggestion
        'Flipper' => '0.251.0',
```
Does it still work?

android/gradle.properties
```diff
@@ -32,7 +32,7 @@ android.useAndroidX=true
 android.enableJetifier=true
 
 # Version of flipper SDK to use with React Native
-FLIPPER_VERSION=0.125.0
+FLIPPER_VERSION=0.182.0
```
jeanregisser, line 35: Should this match the iOS flipper version used?

jeanregisser (approved):
🙌

android/gradle.properties
jeanregisser: Sorry, just saw your comment in the PR about it 🤦

ios/Podfile
kathaypacific: yes it worked, i changed it to use the latest (it was previously locked to a very old version so i opted for the oldest version that has the fix that we need, but i suppose there's no harm to use the newest version)

PR #5206, opened 2024-04-04
feat(points): Add initial points activity feed saga

jophish:
### Description

For [RET-1032](https://linear.app/valora/issue/RET-1032/[wallet-fake-backend]-points-activity-history). Adds in saga/data fetching for points history. (I also started implementing the bottom sheet itself but left it out here in the interest of size, see videos below. Will get out a PR after this as a fast follow.)

See the related PR for the backend `/getHistory` endpoint [here](https://github.com/valora-inc/points-functions/pull/23).

### Test plan

#### Happy path (no error)

https://github.com/valora-inc/wallet/assets/569401/3fd18685-8f27-4b09-b1b9-227eb1fc92de

#### Error

https://github.com/valora-inc/wallet/assets/569401/72838322-f454-4df0-b49d-e838f984d0d2

### Related issues

- Part of [RET-1032](https://linear.app/valora/issue/RET-1032/[wallet-fake-backend]-points-activity-history).

### Backwards compatibility

Yes.
### Network scalability


src/points/slice.ts
```diff
@@ -0,0 +1,58 @@
...
+  pointsHistory: [],
+  nextPageUrl: null,
+  getHistoryStatus: 'idle',
+}
+
+const slice = createSlice({
+  name: 'points',
+  initialState,
+  reducers: {
+    getInitialHistoryStarted: (state) => ({
```
kathaypacific, line 25: i realise the decision to have this explicit "initial" scenario may be related to the current designs. i started a discussion about that [in slack](https://valora-app.slack.com/archives/C0684TXDR3K/p1712309137807599). i guess i'm not entirely convinced that we should clear the history every time the points screen loads :/ that seems different to how we do the transaction feed...

taking it one step further, if the activity feed state needs to be reset every time the user launches the bottom sheet, we could not go through redux at all (why bother storing stuff in global state when it needs to be constantly cleared? we could instead use a hook that is naturally reset when unmounted)

i also wonder if we can simplify the state to only have 1 type of loading? 'loading-initial' is the same as 'loading' + pointsHistory is empty?

src/web3/networkConfig.ts
```diff
@@ -246,6 +247,9 @@ const CAB_DELETE_ENCRYPTED_MNEMONIC_MAINNET = `${CLOUD_FUNCTIONS_MAINNET}/delete
 const SAVE_CONTACTS_ALFAJORES = `${CLOUD_FUNCTIONS_STAGING}/saveContacts`
 const SAVE_CONTACTS_MAINNET = `${CLOUD_FUNCTIONS_MAINNET}/saveContacts`
 
+const GET_POINTS_HISTORY_ALFAJORES = `${CLOUD_FUNCTIONS_STAGING}/getHistory`
```
kathaypacific, line 250: i just realised how generic these function names are :| this might be a pain for monitoring (e.g. going through GCP logs and wondering which history this getHistory endpoint is talking about)...perhaps we should have "points" in our function names....wdyt?

src/points/saga.ts
```diff
@@ -0,0 +1,82 @@
...
+import Logger from 'src/utils/Logger'
+
+const TAG = 'Points/saga'
+
+export async function fetchHistory(address: string, url?: string): Promise<GetHistoryResponse> {
+  const signedMessage = await retrieveSignedMessage()
+  const response = await fetchWithTimeout(url ?? networkConfig.getPointsHistoryUrl, {
+    method: 'GET',
+    headers: {
+      authorization: `Valora ${address}:${signedMessage}`,
```
kathaypacific, line 24: curious, why we need authentication on this endpoint?

kathaypacific (commented):
Looking great! it'd be great to work with Kayla to decide if the activity history should be persisted between points homepage visits and between app sessions.

if we should clear the history every time the user comes into the bottom sheet, i think we should consider whether redux and saga is the most straightforward approach. naively i think we'd have a less complex solution with `useAsyncCallback`...

if we should show previously fetched history, we might want to think about how much history is sensible to keep in redux. currently we just keep putting more stuff there, but that's not good for memory usage. we could consider something like the homefeed (keep only the first page or X items?)

src/web3/networkConfig.ts
kathaypacific: i realise this is out of scope for this PR

src/points/saga.ts
jophish: Hmm, yeah I suppose we don't, though it means we would need to add another query param for `address`

src/points/slice.ts
jophish: I've simplified this a good deal. I've opted for an approach where, upon subsequent visits, we fetch the latest page of data (while displaying the old stale data) and as soon as that page has been fetched, we clear the stale data and overwrite it with the new, fresh history.

The alternative, where we _don't_ overwrite old data, but instead merge it with the newest page we've fetched, is more complicated and opens us up to potentially weird cases for users. For example, say a user views the points history once, and then the next time they view it is a month later, after earning lots of points. When they view the screen this second time, we'll fetch the latest page, and merge it with the old data, but there could be a big gap between the latest data and the old data. The history screen will render the new + old data side by side as if they're contiguous, and the in-between pages won't be fetched until the user drags down all the way _past_ the old data, which is weird. 

(One way to get around that is to _keep fetching_ fresh pages on subsequent viewings, until one of those pages contains a record that we already have in Redux, though I thought we might want to avoid that complexity.)

src/points/saga.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+  )
+  if (response.ok) {
+    return response.json() as Promise<GetHistoryResponse>
+  } else {
+    throw new Error(await response.text())
+  }
+}
+
+export function* getHistory({ payload: params }: ReturnType<typeof getHistoryStarted>) {
```
jophish, line 30: The idea here is that we'll emit `getHistoryStarted` with `fromPage: false` on all opens of the history bottom sheet. On non-first-time-opens, this will have the effect of rendering the old Redux state + loading indicator while we fetch the new data, and the new data will overwrite the old data (see my other comment for more thoughts on this). When a user drags down to fetch more data, set `fromPage: true`, which will fetch using the page params stored in Redux, and _append_ the results to Redux state rather than overwriting them.

src/points/saga.ts
jophish: (updated the backend to reflect this!)

src/points/PointsHome.tsx
```diff
@@ -48,14 +52,31 @@ export default function PointsHome({ route, navigation }: Props) {
       onPress()
     }
   }
+  const onPressActivity = () => {
+    ValoraAnalytics.track(PointsEvents.points_screen_activity_press)
+    dispatch(getHistoryStarted({ fromPage: false }))
```
kathaypacific, line 57: what do you think about dispatching this action on screen mount, to reduce the chances of users needing to wait to see the activity feed? of course we could argue that the activity feed might not be launched but it's a background task and doesn't require much bandwidth to make the network request

src/points/saga.ts
kathaypacific: this makes sense, essentially we're using each bottom sheet load to clear and refresh the cache. i think that works well, although it makes me even more convinced that we should do a background fetch on the points home screen mount rather than on bottom sheet open, since the user could close and reopen the bottom sheet without leaving the screen. wdyt?

can we rename `fromPage` to `fetchFirstPage` / `refreshCache` / `getNextPage` or similar? it was hard to understand what `fromPage` meant at first

src/points/saga.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  const url = params.fromPage ? yield* select(getPointsHistoryNextPageUrlSelector) : undefined
+
+  try {
+    const history = yield* call(fetchHistory, walletAddress, url)
+    yield* put(
+      getHistorySucceeded({
+        appendHistory: params.fromPage,
+        newPointsHistory: history.data,
+        nextPageUrl: history.hasNextPage ? history.nextPageUrl : null,
```
kathaypacific, line 46: null will be stored for the nextPageUrl, so i assume that we'll take that on the front end to mean we should not fetch more?

src/points/slice.ts
```diff
@@ -0,0 +1,53 @@
...
+import { ClaimHistory } from 'src/points/types'
+
+interface GetPointsHistorySucceededAction {
+  appendHistory: boolean | null
+  newPointsHistory: ClaimHistory[]
+  nextPageUrl: string | null
+}
+
+interface GetPointsHistoryStartedAction {
+  fromPage: boolean | null
```
kathaypacific, line 11: why boolean or null instead of just boolean? same for `appendHistory` (could these variables be named the same? maybe i'm missing something but it seems like `fromPage` becomes `appendHistory` in usage)

src/points/saga.test.ts
```diff
@@ -0,0 +1,126 @@
...
+    mockFetch.mockResponseOnce(JSON.stringify(MOCK_HISTORY_RESPONSE))
+    const address = 'some-address'
+    const result = await fetchHistory(address, 'https://example.com')
+    expect(fetchWithTimeoutSpy).toHaveBeenCalledWith('https://example.com', {
+      method: 'GET',
+    })
+    expect(result).toEqual(MOCK_HISTORY_RESPONSE)
+  })
+  it('throws on non-200', async () => {
+    mockFetch.mockRejectOnce(new Error('failure'))
```
kathaypacific, line 68: nit: this mocks fetch throwing rather than fetch resolving with a non 200 right?

src/points/saga.test.ts
```diff
@@ -0,0 +1,126 @@
...
+      .provide([[matchers.call.fn(fetchHistory), throwError(new Error('failure'))]])
+      .put(getHistoryError())
+      .run()
+  })
+  it('fetches from stored page if requested', async () => {
+    const params = getHistoryStarted({
+      fromPage: true,
+    })
+    await expectSaga(getHistory, params)
+      .withState(createMockStore().getState())
```
kathaypacific, line 115: do you need to pass in a stored page to the store to use it?

src/points/saga.test.ts
```diff
@@ -0,0 +1,126 @@
...
+  it('sets error state if no address found', async () => {
+    const params = getHistoryStarted({
+      fromPage: false,
+    })
+    await expectSaga(getHistory, params)
+      .withState(createMockStore({ web3: { account: null } }).getState())
+      .put(getHistoryError())
+      .run()
+  })
+  it('fetches and sets history from scratch', async () => {
```
kathaypacific, line 84: might be nice to add a test that tests the cache refresh (there were items in the points history store, after the saga runs they are replaced)

kathaypacific (approved):
a few nit comments but LGTM! 🚀

src/points/selectors.ts
```diff
@@ -0,0 +1,5 @@
+import { RootState } from 'src/redux/reducers'
+
+export const getPointsHistoryNextPageUrlSelector = (state: RootState) => {
```
kathaypacific, line 3:
```suggestion
export const nextPageUrlSelector = (state: RootState) => {
```

src/points/saga.ts
jophish: Actually, I had meant to handle this from within the `getHistory` method itself, e.g., silently succeeding without fetching if a refresh is requested but no page information is available; I don't think callers should be expected to check that beforehand.

PR #5208, opened 2024-04-05
fix(jumpstart): ensure send status error is not persisted across transactions

kathaypacific:
### Description

At bugapalooza yesterday, we saw the flow:
1. enter a jumpstart transaction amount
2. navigate to the confirmation screen
3. send transaction -> transaction fails -> the error toast is displayed
4. navigate back to the enter amount screen
5. navigate to the confirmation screen -> the error is still there

The root cause is that we are not resetting the flow when navigating back from the confirmation screen. The send flow can be thought of as complete when the status is 'success' or 'error'. after completing a flow, we should reset the status.

on success, we already navigate to the next screen with no option to navigate back to the confirmation screen (the only way to get back here is to start a new send flow, which resets the status back to "idle"). but on error, we allow the user to navigate back to a previous step without resetting the state. The fix is to ensure a state reset when the user ends the flow in "error" and comes back to the confirmation screen.

### Test plan

In step 5. the error should not be there anymore.

https://github.com/valora-inc/wallet/assets/20150449/3896e363-f0d2-4a94-8d9c-01bdf5df19e6

### Related issues

n/a

### Backwards compatibility

Y

### Network scalability

Y


jophish (approved):
LGTM assuming the build passes!

PR #5221, opened 2024-04-11
refactor: Add new top bar buttons [DRAFT for ACT-1134]

chriskeating603:
### Description

I added a component `TopBarIconButtonV2` which is closer to the solution, but I cannot figure where the height prop is being set. When I change the `borderRadius`, it impacts the width of the ripple, but the height always remains the same. I need to figure out how to set the ripple to be a circle. 

### Related issues

- Fixes [ACT-1134](https://linear.app/valora/issue/ACT-1134/create-new-topbarbuttons-text-icon)

### Network scalability


chriskeating603:
Ok so I added some changes here where I have different padding for horizontal and vertical. I am using     
`paddingVertical: Spacing.Small12` and `paddingHorizontal: Spacing.Regular16` for the BackButton. I've figured out that this is due to the size of the icon inside, so I think ultimately I want to have the padding amount be responsive to the size of the icon. This way, the shape of the ripple (which is caused by the padding) is always a circle

chriskeating603:
Split out TopBarButton file and TopBarIconButtonV2 + updated BackChevron icon to accommodate new TopBarIconButtonV2 component. Will add any necessary tests on Monday!

ios/Podfile.lock
```diff
@@ -1251,4 +1251,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 81b64061bfc811164c7f86731fc3327e217c8ce4
 
-COCOAPODS: 1.12.1
+COCOAPODS: 1.15.2
```
MuckT, line 1254: This looks like an unexpected change.
```suggestion
COCOAPODS: 1.12.1
```

src/navigator/TopBarIconButtonV2.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+export type TopBarIconButtonProps = CommonProps & {
+  icon: JSX.Element
+  containerStyle?: ViewStyle
+}
+
+export function TopBarIconButtonV2(props: TopBarIconButtonProps) {
+  return (
+    <Wrapper {...props}>
+      <View style={{ margin: -Spacing.Small12 }}>
```
MuckT, line 62: Is this needed? A bit strange to have this inline style outside of the containerStyle.

MuckT: This should be from components that need it as a container style from `SendSelectRecipient.tsx`

src/navigator/TopBarIconButtonV2.tsx
```diff
@@ -0,0 +1,88 @@
...
+        ? ValoraAnalytics.track(eventName, eventProperties)
+        : ValoraAnalytics.track(eventName)
+    }
+    onPress()
+  }, [onPress, eventName])
+
+  return (
+    <Touchable
+      disabled={disabled}
+      // testID={`${testID}_wrapper`}
```
MuckT, line 43:
```suggestion
      testID={testID}
```

src/navigator/TopBarIconButtonV2.tsx
```diff
@@ -0,0 +1,88 @@
...
+  containerStyle?: ViewStyle
+}
+
+export function TopBarIconButtonV2(props: TopBarIconButtonProps) {
+  return (
+    <Wrapper {...props}>
+      <View style={{ margin: -Spacing.Small12 }}>
+        <View style={[styles.container, props.containerStyle]}>
+          <Touchable
+            testID={props.testID}
```
MuckT, line 65:
```suggestion
```

MuckT:
Closing in favor of #5276

PR #5230, opened 2024-04-12
fix(cya): ensure profile and celo options navigate correctly

satish-ravi:
### Description

Instead of making back to back navigate calls, use CommonActions.reset to reset the nav stack with the correct screens.

### Test plan

Unit tests, e2e tests, manual

### Related issues

- Fixes ACT-1151

### Backwards compatibility

Yes

### Network scalability

N/A

### DIFF
--- a/e2e/src/usecases/ChooseYourAdventure.js
+++ b/e2e/src/usecases/ChooseYourAdventure.js
@@ -1,0 +1,60 @@
+import { launchApp } from '../utils/retries'
+import { quickOnboarding, sleep, waitForElementId } from '../utils/utils'
+
+export default ChooseYourAdventure = () => {
+  beforeAll(async () => {
+    await device.disableSynchronization()
+  })
+
+  beforeEach(async () => {
+    await device.uninstallApp()
+    await device.installApp()
+    await launchApp({
+      launchArgs: {
+        statsigGateOverrides: `use_tab_navigator=true,show_cloud_account_backup_setup=true,show_cloud_account_backup_restore=true`,
+      },
+    })
+    await quickOnboarding({ stopOnCYA: true, cloudBackupEnabled: true })
+  })
+
+  it('learn about celo navigates to celo token details page', async () => {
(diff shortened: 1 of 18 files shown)


src/onboarding/ChooseYourAdventure.tsx
```diff
@@ -64,7 +69,11 @@ function ChooseYourAdventure() {
     {
       text: t('chooseYourAdventure.options.add'),
       goToNextScreen: () => {
-        navigateHome() // navigate home so that back on the fiat exchange currency screen takes the user back to Home screen
+        // navigate home so that closing the fiat exchange currency bottom sheet
+        // takes the user back to Home screen. Can't use
+        // navigateHomeAndThenToScreen here because it doesn't work for bottom
+        // sheets.
+        navigateHome()
```
satish-ravi, line 76: this has been working fine and didn't need a fix.

e2e/src/usecases/ChooseYourAdventure.js
```diff
@@ -0,0 +1,56 @@
...
+
+  it('learn about celo navigates to celo token details page', async () => {
+    await element(by.text('Learn about Celo')).tap()
+
+    // Check that we are on the Celo token details page
+    await waitForElementId('TokenDetails/AssetValue')
+    await expect(element(by.text('Celo')).atIndex(0)).toBeVisible()
+
+    // Back should go to the home screen
+    await element(by.id('BackChevron')).tap()
```
MuckT, line 24:
```suggestion
    await waitForElementByIdAndTap('BackChevron')
```

This the advantage of waiting for the back chevron before attempting to tap, but this should also be handled by detox's sync.

MuckT (approved):

e2e/src/usecases/ChooseYourAdventure.js
satish-ravi: We wait for other elements of the screen and have assertions, so back should be visible at this point.

PR #5231, opened 2024-04-12
feat(points): fetch points config from getPointsConfig function

kathaypacific:
### Description

This PR:
- add points config and status variables to redux
- on app start, fetch and store relevant points config (remove any unsupported activities)
- the points config should always be refreshed on app start to avoid showing stale points values
- add selector which massages the points config to the `PointsMetata[]` type, so that we can switch out the UI components to use the selector instead of statsig
- add unit tests and redux migration

### Test plan

No user facing changes here

### Related issues

- Related to RET-1044

### Backwards compatibility

Y

### Network scalability

Y


src/points/slice.ts
```diff
@@ -11,16 +12,27 @@ interface GetPointsHistoryStartedAction {
   getNextPage: boolean
 }
 
+export type PointsConfig = {
+  activitiesById: {
+    [activityId in PointsActivity]?: {
+      points: number
```
jophish, line 18: Did we decide on [pointsAmount](https://valora-app.slack.com/archives/C029Z1QMD7B/p1712928606786159) for this?

src/points/PointsHome.tsx
```diff
@@ -1,23 +1,23 @@
 import { NativeStackScreenProps } from '@react-navigation/native-stack'
```
jophish, line 1: Are there meant to be no functional changes in this file?

src/points/PointsHome.test.tsx
```diff
@@ -1,13 +1,13 @@
-import * as React from 'react'
```
jophish, line 1: Same q with this file as for `PointsHome.tsx`

src/points/saga.ts
```diff
@@ -66,10 +75,45 @@ export function* getHistory({ payload: params }: ReturnType<typeof getHistorySta
...
+    const supportedActivities: PointsConfig = {
+      activitiesById: {},
+    }
+    Object.entries(config.activitiesById).forEach(([activityId, activityMetadata]) => {
+      if (isPointsActivity(activityId) && activityMetadata.points > 0) {
+        supportedActivities.activitiesById[activityId] = activityMetadata
+      }
+    })
+    if (Object.keys(supportedActivities.activitiesById).length === 0) {
+      throw new Error('No supported activities found in points config')
```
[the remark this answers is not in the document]
jophish: (also, this will presumably result in an infinite loop of errors that retrying will never fix)

src/points/selectors.ts
```diff
@@ -1,5 +1,44 @@
...
+      }
+
+      // check if there is already a metadata object for this points value,
+      // either add activity to the existing points object or create a new one
+      const existingMetadata = pointsMetadata.find((metadata) => metadata.points === points)
+      if (existingMetadata) {
+        existingMetadata.activities.push({ name: activityId })
+      } else {
+        pointsMetadata.push({
+          points,
```
jophish, line 32: nit: should this also be `pointsAmount`?

src/points/selectors.ts
```diff
@@ -1,5 +1,44 @@
...
+          activities: [{ name: activityId }],
```
jophish, line 33: Should `name` be `activityId` instead?

src/points/selectors.ts
```diff
@@ -1,5 +1,44 @@
...
 }
+
+const pointsConfigSelector = (state: RootState) => state.points.pointsConfig
+
+export const pointsSectionsSelector = createSelector(
+  [pointsConfigSelector],
+  (pointsConfig): PointsMetadata[] => {
+    const pointsMetadata: PointsMetadata[] = []
+    if (!pointsConfig) {
+      return pointsMetadata
```
jophish, line 16: Could we change the type of the `pointsConfig` state from `pointsConfig: PointsConfig | null` to `pointsConfig: PointsConfig` and set the initial state to `{ activitiesById: {}}` to avoid this special check? Not sure if it would make things uglier in the long run/not committed to the idea, just curious.

src/points/saga.ts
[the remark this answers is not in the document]
kathaypacific: yeah good question...i feel weird asking for a design for the points screen for when no points activities are available :/ i guess i had assumed that there would always be points activities for app versions that support points but what you're saying is that's not the case. i think that's fair, i probably overthought this scenario.

i've removed this error throwing block - the scenario where there are no supported activities can then be handled on the UI separately to an error

src/points/slice.ts
kathaypacific: yes! i was planning to make the renaming changes in one go but it makes sense to do this one to avoid the need for a redux migration

src/points/PointsHome.tsx
kathaypacific: yes i think i moved the changes to a different PR but had a hard time turning off the vs code organize imports setting. i've removed these unrelated changes now

src/points/selectors.ts
kathaypacific: i'll make this change separately, since this is declared on the PointsMetadata type and still cause changes in the UI components

src/points/selectors.ts
kathaypacific: you're right, i can't think of any benefits to having an explicit null

src/points/slice.ts
kathaypacific: (i'll need to make this change also in the getPointsConfig CF, in case you're wondering why this branch doesn't work anymore)

jophish: gotcha, sgtm!

src/points/saga.ts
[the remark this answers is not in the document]
jophish: yeah, i don't think this will be a common thing, but prolly good to avoid a constant loop of errors and just deal with it in the frontend as a special case

jophish (approved):
Looks great!

jeanregisser (approved):
👍

PR #5232, opened 2024-04-12
chore(points): use points config from redux rather than Statsig

kathaypacific:
### Description

This PR:
- remove the statsig dynamic config
- use the redux selectors to provide config to existing UI components
- update unit tests (some were removed because the scenarios are tested in the saga and selector)

### Test plan

n/a

### Related issues

- Related to RET-1044

### Backwards compatibility

Y

### Network scalability

Y


src/points/PointsHome.tsx
```diff
@@ -84,13 +87,9 @@ export default function PointsHome({ route, navigation }: Props) {
           <Text style={styles.infoCardTitle}>{t('points.infoCard.title')}</Text>
           <Text style={styles.infoCardBody}>{t('points.infoCard.body')}</Text>
         </View>
-        <ActivityCardSection onCardPress={onCardPress} />
+        <ActivityCardSection onCardPress={onCardPress} pointsSections={pointsSections} />
       </ScrollView>
-      <BottomSheet
-        snapPoints={['50%']}
```
kathaypacific, line 90: this is a personal preference but the vertical spaces triggered me, particularly between the bottom of the button and the bottom of the screen. i think it's better to let the bottom sheet size itself (if we need to increase the height, we should add space above the button)

src/points/ActivityCardSection.tsx
```diff
@@ -1,27 +1,21 @@
...
 }
 
-export default function ActivityCardSection({ onCardPress }: Props) {
+export default function ActivityCardSection({ onCardPress, pointsSections }: Props) {
   const { t } = useTranslation()
 
-  const pointsConfig = getDynamicConfigParams(DynamicConfigs[StatsigDynamicConfigs.POINTS_CONFIG])
-
-  function makeSection(pointsMetadata: PointsMetadata): React.ReactNode {
+  function makeSection(pointsMetadata: PointsMetadata, sectionNumber: number): React.ReactNode {
```
bakoushin, line 18: nit: is that an index? wdyt about naming it `sectionIndex`?

bakoushin (approved):

jophish (approved):

PR #5253, opened 2024-04-15
feat(points): add loading and error states for loading config

kathaypacific:
### Description

As the title.

### Test plan

The load is done on app start so we shouldn't see a loading state unless there was an error and the user tries again. Added an artificial delay for these videos.

Error state:

https://github.com/valora-inc/wallet/assets/20150449/d15a4b5a-b5cb-4bb0-bb07-91a15f2c6975

Success state:

https://github.com/valora-inc/wallet/assets/20150449/ec5a50cf-2873-4d4b-8dbc-f41aa1240b8b

### Related issues

- Fixes RET-1044

### Backwards compatibility

Y

### Network scalability

Y


locales/base/translation.json
```diff
@@ -2244,6 +2244,15 @@
   "points": {
     "title": "Valora Points",
     "activity": "Activity",
+    "loading": {
+      "title": "Gathering your points",
+      "description": "We're in the process of tallying up your points, hang tight"
```
jophish, line 2249: nit: should this copy not have a period, unlike the error description? :eyes: i see that this matches the figma designs but not sure if it's intentional.

jophish (approved):

locales/base/translation.json
kathaypacific: i'll double check! i think the copy is not final yet though, seems reasonable to assume that we'll do a pass through before releasing this

PR #5254, opened 2024-04-15
feat(points): Add points history bottom sheet

jophish:
### Description

Part of [RET-1032](https://linear.app/valora/issue/RET-1032/[wallet-fake-backend]-points-activity-history). Adds in a points history bottom sheet, minus the actual card rendering.

### Test plan

Unit and manual tested. See videos:

https://github.com/valora-inc/wallet/assets/569401/7319dce2-934a-4bf2-95db-4bb583136430

https://github.com/valora-inc/wallet/assets/569401/1766cddb-50d2-43da-bb10-54e5df6a7b4a

### Related issues

Part of [RET-1032](https://linear.app/valora/issue/RET-1032/[wallet-fake-backend]-points-activity-history).

### Backwards compatibility

Yes.

### Network scalability


src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -0,0 +1,182 @@
...
+        />
+      </View>
+    )
+  }
+
+  const sections = useMemo(() => {
+    if (!pointsHistory.length) {
+      return []
+    }
+    return groupFeedItemsInSections([], pointsHistory, (t: ClaimHistory) => Date.parse(t.createdAt))
```
kathaypacific, line 79: could we simplify this by returning a `timestamp` on the history items?

src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const onPressTryAgain = () => {
+    ValoraAnalytics.track(PointsEvents.points_screen_activity_try_again_press)
+    dispatch(
+      getHistoryStarted({
+        getNextPage: false,
+      })
+    )
+  }
+
+  const fetchMoreHistory = () => {
```
kathaypacific, line 46: could we prefix methods consistently with either `on` or `handle`? (my preference is `handle` for methods declared in the component and `on` for callbacks passed as props so that you can quickly understand where the logic is when reading the renders, although we've been super inconsistent about doing this)

src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -0,0 +1,182 @@
...
+      }
+      default: {
+        Logger.error(TAG, `Unknown points history status found: ${pointsHistoryStatus}`)
+      }
+    }
+  }
+
+  return (
+    <BottomSheetBase snapPoints={['80%']} forwardedRef={forwardedRef}>
+      <View style={styles.container}>{renderContents()}</View>
```
kathaypacific, line 133: i wonder if we could simplify this component by always having the BottomSheetFlatList here? i'm not sure if you've already tried and i'm overlooking some complexities....but naively i was thinking we could have the BottomSheetFlatList always, the sections would be always computed, the loader is always there as a child of the BottomSheetFlatList if the status is loading (i _think_ this makes it always appear at the bottom of the list). there is a ListEmptyComponent prop on the flatlist which you could use to create a component that displays either the error state or an empty state saying "you have no history" etc. depending on the history status. 

the advantage i see with this approach is that you can see at a glance all the components that could be present in the render, so you can easily understand what the screen could look like. it's personal preference but i find it much more difficult to visualise the screen by reading a render method that has many layers of logic handlers.

src/points/PointsHistoryBottomSheet.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+import { PointsEvents } from 'src/analytics/Events'
+import { getHistoryStarted } from 'src/points/slice'
+import { GetHistoryResponse } from 'src/points/types'
+
+jest.mock('@gorhom/bottom-sheet', () => {
+  const react = require('react-native')
+  return {
+    __esModule: true,
+    default: react.View,
+    BottomSheetSectionList: react.SectionList,
```
kathaypacific, line 18: can you add this to `__mocks__/@gorhom/bottom-sheet.ts` instead so that it's applied globally?

src/points/PointsHistoryBottomSheet.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+    },
+  ],
+  hasNextPage: false,
+  nextPageUrl: '',
+}
+
+describe(PointsHistoryBottomSheet, () => {
+  const mockFetch = fetch as FetchMock
+  beforeEach(() => {
+    jest.useRealTimers()
```
kathaypacific, line 50: why use real timers?

kathaypacific (commented):
looking great, i had some ideas that might simplify some logic :D

src/points/PointsHistoryBottomSheet.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+  })
+
+  it('Shows error screen if fetch fails', async () => {
+    const tree = renderScreen({ points: { getHistoryStatus: 'error' } })
+    expect(tree.getByTestId('PointsHistoryBottomSheet/ErrorState')).toBeTruthy()
+  })
+
+  it('Dispatches action when try again is pressed', async () => {
+    const { dispatch, getByTestId } = renderScreen({ points: { getHistoryStatus: 'error' } })
+    fireEvent.press(getByTestId('PointsHistoryBottomSheet/TryAgain'))
```
bakoushin, line 114: nit: if we query by text, we ensure that the user will see the correct labels (Kathy already mentioned it in another PR comment)
i personally try to resort to test ids only if it's impossible to query by text.
see also: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-querying-by-text

src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const fetchMoreHistory = () => {
+    ValoraAnalytics.track(PointsEvents.points_screen_activity_fetch_more)
+    dispatch(
+      getHistoryStarted({
+        getNextPage: true,
+      })
+    )
+  }
+
+  const renderError = () => {
```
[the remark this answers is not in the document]
jophish: ohhh good shout, that definitely makes sense!

src/points/PointsHistoryBottomSheet.tsx
jophish: a UNIX timestamp of the `createdAt` time in addition to the `createdAt` field? that would simplify this a bit (the function would be generic on `T extends { timestamp: number }`) though i shy away from it for two reasons:

* we'd be duplicating the "same" data in two different fields, `createdAt` and `timestamp`; having two sources of truth doesn't seem ideal
* we're tightly coupling the API interface with this re-used implementation from the transaction feed, meaning that we're stuck with the name `timestamp` exactly for no great reason (unless we want to apply some transform before this). `timestamp` feels a bit confusing... we already have `createdAt`, so i would assume that `timestamp` might not necessarily _also_ represent creation time

src/points/PointsHistoryBottomSheet.test.tsx
jophish: whoooops, don't think that belongs here :)

src/points/PointsHistoryBottomSheet.tsx
kathaypacific: the getHistory endpoint is meant to make our lives easy on the front end so i was wondering if we could rename createdAt to timestamp actually...

jophish: ohh yeah - @jeanregisser was pulling in the direction of making the "created at" time more human readable, but fwiw i actually started with a UNIX timestamp and i think that's probably my preference anyhow

src/points/PointsHistoryBottomSheet.tsx
jophish: OK, after messing around with this for a while, I think I've come to a working solution. Setting direct children of `BottomSheetSectionList` (or `SectionList` in general, I presume) doesn't work; they never rendered no matter what I did, and I"m guessing they're just ignored. Setting the `ListFooterComponent` however, did just the trick.

There's a _bit_ of complexity in selectively rendering the "My points activity" title (outside of the section list) depending on the history status, but that's not too bad - overall I think this approach is a whole lot cleaner!

src/points/PointsHistoryBottomSheet.tsx
jophish: In order to avoid another migration, I've made this change _down_ the chain of PRs in `jophish/ret-1032-3`; when I merge that branch into this one, the change will take effect.

jeanregisser: I have a strong case for keeping ISO 8601 strings: https://github.com/valora-inc/points-functions/pull/46#discussion_r1570453534

src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -0,0 +1,163 @@
...
+        />
+      </View>
+    </BottomSheetBase>
+  )
+}
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+    padding: Spacing.Thick24,
```
jeanregisser, line 131: Related to https://github.com/valora-inc/wallet/pull/5259#pullrequestreview-2008511618

There should be no padding here, otherwise the scroll indicator is placed incorrectly.
Also the scrollview needs to be fully scrollable without hitting an invisible wall at the bottom.
We should likely use a SafeAreaView or inset to properly handle this on devices with a home bar (iOS).

src/points/PointsHistoryBottomSheet.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+    )
+
+    return {
+      dispatch,
+      store,
+      ...tree,
+    }
+  }
+
+  it('Show empty state when no hstory', async () => {
```
jeanregisser, line 65: nit:
```suggestion
  it('shows empty state when no history', async () => {
```

src/points/PointsHistoryBottomSheet.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+    const tree = renderScreen()
+
+    expect(tree.queryByTestId('PointsHistoryBottomSheet/Error')).toBeNull()
+    expect(tree.queryByTestId('PointsHistoryBottomSheet/Error')).toBeNull()
+    expect(tree.queryByTestId('PointsHistoryBottomSheet/Empty')).toBeTruthy()
+  })
+
+  it('Displays content while loading', async () => {
```
jeanregisser, line 73:
```suggestion
  it('displays content while loading', async () => {
```

src/points/PointsHistoryBottomSheet.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+      },
+    })
+
+    // Trigger end reached to fetch more
+    getByTestId('PointsHistoryList').props.onEndReached()
+
+    expect(dispatch).toHaveBeenCalledWith(getHistoryStarted({ getNextPage: true }))
+  })
+
+  it('Shows error screen if fetch fails', async () => {
```
jeanregisser, line 99:
```suggestion
  it('shows error screen if fetch fails', async () => {
```

src/points/PointsHistoryBottomSheet.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+    const tree = renderScreen({ points: { getHistoryStatus: 'error' } })
+    expect(tree.getByTestId('PointsHistoryBottomSheet/Error')).toBeTruthy()
+  })
+
+  it('Dispatches action when try again is pressed', async () => {
```
jeanregisser, line 104:
```suggestion
  it('dispatches an action when try again is pressed', async () => {
```

src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -0,0 +1,163 @@
...
+  }, [pointsHistory, pointsHistoryStatus])
+
+  // TODO: Render items
+  const renderItem = () => {
+    return <></>
+  }
+
+  const renderLoading = () => {
+    if (pointsHistoryStatus !== 'loading') {
+      return <></>
```
jeanregisser, line 66:
```suggestion
      return null
```

src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -0,0 +1,163 @@
...
+    }
+    return groupFeedItemsInSections([], pointsHistory, (t: ClaimHistory) => Date.parse(t.createdAt))
+  }, [pointsHistory, pointsHistoryStatus])
+
+  // TODO: Render items
+  const renderItem = () => {
+    return <></>
+  }
+
+  const renderLoading = () => {
```
jeanregisser, line 64: We should avoid defining components inline. it's bad for perf and can have unintended effects (like if we use `useEffects` in them they'll be triggered on each render, since the parent comp is created on each reader).

src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -0,0 +1,163 @@
...
+      <ActivityIndicator
+        testID={'PointsHistoryBottomSheet/Loading'}
+        style={styles.loadingIcon}
+        size="large"
+        color={colors.primary}
+      />
+    )
+  }
+
+  const renderEmpty = () => {
```
jeanregisser, line 78: ditto about defining components inline.

jeanregisser (commented):
Nice! 👍

src/points/PointsHistoryBottomSheet.tsx
jophish: Got it, I've pulled these out into their own functional components.

src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -0,0 +1,154 @@
...
+    pointsHistoryStatus === 'loading' ? (
+      <ActivityIndicator
+        testID={'PointsHistoryBottomSheet/Loading'}
+        style={styles.loadingIcon}
+        size="large"
+        color={colors.primary}
+      />
+    ) : null
+
+  const Empty =
```
bakoushin, line 61: super nit: what if we name it `EmptyOrError` to reflect that it can display an error as well?

bakoushin (approved):

src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -0,0 +1,154 @@
...
+          contentContainerStyle={styles.contentContainer}
+          renderItem={renderItem}
+          renderSectionHeader={(item) => (
+            <SectionHead text={item.section.title} style={styles.sectionHead} />
+          )}
+          sections={sections}
+          keyExtractor={(item) => `${item.activity}-${item.timestamp}`}
+          keyboardShouldPersistTaps="always"
+          testID="PointsHistoryList"
+          onEndReached={onFetchMoreHistory}
```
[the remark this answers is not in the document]
jophish: hmm yeah, no harm i think! i'll set it to `.5` so it refreshes when the user is halfway down the list.

src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -0,0 +1,154 @@
...
+          size={BtnSizes.FULL}
+          style={{ width: '100%' }}
+        />
+      </View>
+    ) : (
+      <View testID={'PointsHistoryBottomSheet/Empty'}></View>
+    ) // TODO: Render empty/no history state
+
+  const sections = useMemo(() => {
+    if (!pointsHistory.length || pointsHistoryStatus === 'error') {
```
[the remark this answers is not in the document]
jophish: it's actually slightly trickier than just removing this condition. we want to display a full screen error when the _initial_ fetch fails, but not when a later page fetch fails. the redux `pointsHistory` state isn't reset on an error. from just looking at `pointsHistory` and `pointsHistoryStatus` in the component, we can't tell which of the two above cases we're in, assuming `pointsHistoryStatus === 'error'`.

one way to fix this is:
* wipe the redux state on error _only_ if it was during the initial fetch
* remove this check, like you mention
* add in a check for `pointsHistory.length && pointsHistoryStatus === 'error'` elsewhere in the component (which will tell us if we error'd during a page fetch) and render something TBD

i'll do the first two parts of this now, and talk with kayla about the last point as a followup.

src/points/PointsHistoryBottomSheet.tsx
jophish: Aha, after messing around with this, yeah -- your approach works well, with the appropriate padding applied to individual components (list entries and section headers), and using an inset for the bottom padding.

PR #5255, opened 2024-04-15
chore(points): Use activityId and pointsAmount throughout

jophish:
### Description

Updates the points implementation + redux to use `activityId` and `pointsAmount` everywhere, as discussed in Slack.

### Test plan

Manual/unit tested.

### Related issues

Part of [ACT-1032](https://linear.app/valora/issue/RET-1032/[wallet-fake-backend]-points-activity-history).

### Backwards compatibility

Yes.

### Network scalability


src/points/ActivityCard.tsx
```diff
@@ -11,26 +11,26 @@ import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
 import { PointsEvents } from 'src/analytics/Events'
 
 interface Props {
-  activity: PointsActivity
+  activityId: PointsActivity
   points: number
```
kathaypacific, line 15: should this also be `pointsAmount`?

kathaypacific (approved):
💅🏼🚀

src/points/ActivityCard.tsx
```diff
@@ -11,26 +11,26 @@ import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
 import { PointsEvents } from 'src/analytics/Events'
 
 interface Props {
-  activity: PointsActivity
+  activityId: PointsActivity
```
bakoushin, line 14: should we change the type name to `PointsActivityId` as well?

src/points/ActivityCard.tsx
jophish: yep, good call!

kathaypacific (approved):

PR #5259, opened 2024-04-16
feat(points): Implement Points Activity bottom sheet cards

jophish:
### Description

Part of [RET-1032](https://linear.app/valora/issue/RET-1032/[wallet-fake-backend]-points-activity-history). Implements the points cards on the bottom sheet. Relies on the changes in [this PR](https://github.com/valora-inc/points-functions/pull/44) in the functions repo.

### Test plan

Unit and manual tested. See video.

https://github.com/valora-inc/wallet/assets/569401/675002c6-b74b-4c4b-8bc0-b5e809c4ee0c

### Related issues

- Fixes [RET-1032](https://linear.app/valora/issue/RET-1032/[wallet-fake-backend]-points-activity-history)

### Backwards compatibility

Yes.

### Network scalability


locales/base/translation.json
```diff
@@ -2245,6 +2245,16 @@
     "title": "Valora Points",
     "activity": "Activity",
     "history": {
+      "cards": {
+        "createWallet": {
```
jophish, line 2249: note: note sure about the copy/image here, will check with Kayla

src/icons/SwapArrows.tsx
```diff
@@ -8,7 +8,7 @@ interface Props {
 }
 
 const SwapArrows = ({ size = 24, color = Colors.black }: Props) => (
-  <Svg width={size} height={size} viewBox="0 0 24 24" fill="none">
```
bakoushin, line 11: what's the reason for this not being a square?

I think that having all the icons squared doesn't distort their initial proportions and enables them to be easily reusable in other parts of the app.

note how all the icons in the [figma components library](https://www.figma.com/file/erFfzHvSTm5g1sjK6jWyEH/Working-Design-System-Components?type=design&node-id=317-135&mode=design&t=2wyGE58IAneSuv2f-0) are squared:
image

src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -79,9 +109,16 @@ function PointsHistoryBottomSheet({ forwardedRef }: Props) {
...
-  // TODO: Render items
-  const renderItem = () => {
-    return <></>
+  const renderItem: ListRenderItem<ClaimHistory> = ({ item }: { item: ClaimHistory }) => {
+    try {
+      const historyDefinition = getHistoryDefinition(item)
+      return (
+        <PointsHistoryCard {...historyDefinition} testID={`${item.activityId}-${item.createdAt}`} />
+      )
+    } catch (error) {
```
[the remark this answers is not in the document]
jophish: yeah definitely. i think it's actually a good idea to do this filtering at the saga level _and_ at the selector level, since there's a possibility that we remove an supported activityId (and its associated rendering code), but records with that activityId still exist within redux

src/icons/SwapArrows.tsx
jophish: Ahhh good catch! If you double click on the icons there, it creates a smaller bounding box that directly touches the SVG border, which is what I think I used when exporting the icons.

src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -106,7 +133,7 @@ function PointsHistoryBottomSheet({ forwardedRef }: Props) {
           <Text style={styles.contentHeader}>{t('points.history.title')}</Text>
         )}
         <BottomSheetSectionList
-          contentContainerStyle={styles.contentContainer}
+          contentContainerStyle={sections.length ? {} : styles.contentContainer}
```
jophish, line 136: note that there was some weird behavior here that i couldn't quite figure out; having `flex: 1` breaks scrolling on the section list, however, _without_ `flex: 1`, styling was totally off on the error state. some googling indicates this _might_ be an android thing, but not totally sure.

src/points/cardDefinitions.tsx
```diff
@@ -40,3 +49,44 @@ export default function useCardDefinitions(
...
 }
+
+export function useGetHistoryDefinition(): (history: ClaimHistory) => HistoryCardMetadata {
+  const { t } = useTranslation()
+  const tokensById = useSelector((state) =>
+    tokensByIdSelector(state, getSupportedNetworkIdsForSwap())
+  )
+
+  return (history: ClaimHistory) => {
+    // This pattern gives us static checking that we've exhausted all string union values
```
jeanregisser, line 60: I don't think it does check for exhaustiveness,
We need a `default` with `never` for this.
https://www.typescriptlang.org/docs/handbook/2/narrowing.html#exhaustiveness-checking

src/points/saga.test.ts
```diff
@@ -108,13 +117,13 @@ describe('getHistory', () => {
       .put(
         getHistorySucceeded({
           appendHistory: false,
-          newPointsHistory: MOCK_HISTORY_RESPONSE.data,
+          newPointsHistory: MOCK_HISTORY_RESPONSE.data.slice(0, 2),
```
jeanregisser, line 120: Why are we doing this now?

jeanregisser (commented):
👍 

Small UI nit not directly related to the code changes here, but the video shows the scroll indicator has some right padding.
This is because a padding is added to the container of the ScrollView and this should be avoided.

src/points/saga.test.ts
jophish: Since we now filter out history records whose activityIds are unknown/invalid.

jeanregisser: Ok! Should we maybe introduce a `const MOCK_SUPPORTED_HISTORY = MOCK_HISTORY_RESPONSE.filter(...)` to make it more obvious? possibly with a little comment.

src/points/cardDefinitions.tsx
jophish: I believe this approach does check for exhaustiveness, albeit somewhat indirectly; removing a case statement causes the compiler to recognize the return value as potentially undefined/unassigned; this warning only goes away all possible cases are handled.

I'm not sure which pattern is more idiomatic; I was thinking we should go with whatever pattern gives us a more useful error message, though to be honest, neither of them give a _super_ actionable error at compile time... One added benefit of the pattern you mention is that it avoids needing `break` statements within the cases, which I like to avoid when possible, since it's easy to forget a `break` and have accidental fall-through. Will go with this pattern!

jeanregisser: What I tested is that it doesn't complain if we add a new value to the `claimActivities` array.
Which is important that it does complain if we do that, no?

jophish: Oh interesting... I tested by removing a case from the switch statement and it did complain at me... which I would imagine is sort of equivalent to adding a new value to the `claimActivities` array... Definitely important that it complains in both of those examples. In any case, I've updated this to use the new pattern.

src/points/saga.test.ts
jophish: Yeah! good idea.

src/points/PointsHistoryBottomSheet.tsx
kathaypacific: ah, this rings a bell but i can't remember when i ran into this. the scrolling views are 2 nested containers and `style` controls the outer one and `contentContainerStyle` controls the inner one. perhaps applying flex:1 to one of them makes a difference (but maybe not, what you've got here is unoffencive)

src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -159,5 +185,40 @@ const styles = StyleSheet.create({
...
     paddingHorizontal: 0,
   },
+  historyCard: {
+    flexDirection: 'row',
+    alignItems: 'center',
+    paddingVertical: Spacing.Regular16,
+  },
+  cardIcon: {
+    backgroundColor: colors.successLight,
+    display: 'flex',
```
kathaypacific, line 195: you don't need this, all styles in RN are flexbox

src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -159,5 +185,40 @@ const styles = StyleSheet.create({
...
+    alignItems: 'center',
+    borderRadius: 20,
+    width: 40,
+    height: 40,
+    padding: Spacing.Smallest8,
+    marginRight: Spacing.Regular16,
+  },
+  cardContent: {
+    flex: 1,
+    flexDirection: 'column',
```
kathaypacific, line 206: this is also the default value

src/points/PointsHistoryBottomSheet.tsx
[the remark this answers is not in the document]
kathaypacific: wait i'm not quite following...are you saying we want to remove history records if the activityId is no longer supported? this assumption is baked into this PR but i can imagine a world where we stop rewarding points for creating a wallet, but if you've got points for that then we should still show it no? we don't subtract the points from the balance so we should always show them in the history too? this does imply that we'd need to support more activities for the history than the active claimable activities....

src/points/cardDefinitions.tsx
```diff
@@ -40,3 +49,43 @@ export default function useCardDefinitions(
...
+          title: t('points.history.cards.createWallet.title'),
+          subtitle: t('points.history.cards.createWallet.subtitle'),
+          pointsAmount: history.pointsAmount,
+        }
+      }
+      case 'swap': {
+        const fromToken = tokensById[history.metadata.from]
+        const toToken = tokensById[history.metadata.to]
+        if (!fromToken || !toToken) {
+          throw new Error(`Cannot find tokens ${history.metadata.from} or ${history.metadata.to}`)
```
kathaypacific, line 73: rather than throwing an error, can we just silently omit this entry? or we can do a Logger.error so that it gets sent to Sentry, but idk if it's worth the extra error handling to do that here?

src/points/types.ts
```diff
@@ -35,7 +35,12 @@ export type PointsMetadata = {
   }>
 }
 
-type ClaimActivityId = 'create-wallet' | 'swap'
+const claimActivities = ['create-wallet', 'swap'] as const
+type ClaimActivityId = (typeof claimActivities)[number]
+
+export function isClaimActivityId(activity: unknown): activity is ClaimActivityId {
```
kathaypacific, line 41: nit: i wonder if this should just be `isActivityId` (mainly because i don't want to go back and rename activity -> claimActivity everywhere 😅) or maybe `isSupportedActivityId`?

src/points/types.ts
```diff
@@ -61,3 +66,10 @@ export interface GetHistoryResponse {
   hasNextPage: boolean
   nextPageUrl: string
 }
+
+export interface HistoryCardMetadata {
```
kathaypacific, line 70: i think that any types that are specifically for presentation should live with the presentational component (they're unlikely to be reused outside of that presentational context). it'd be nice for the types file to contain data types only?

src/points/selectors.ts
```diff
@@ -10,10 +15,17 @@ export const pointsHistoryStatusSelector = (state: RootState) => {
...
 
-export const pointsHistorySelector = (state: RootState) => {
+const allPointsHistorySelector = (state: RootState) => {
   return state.points.pointsHistory
 }
 
+export const pointsHistorySelector = createSelector(
+  [allPointsHistorySelector],
+  (pointsHistory): ClaimHistory[] => {
+    return pointsHistory.filter((history) => isClaimActivityId(history.activityId))
```
kathaypacific, line 25: i wonder if we need this here if it can be in the saga already (but there's an open thread about that, i might be missing something)

kathaypacific (commented):
this actually looks great but wanted to check about this assumption that we only want to show history for supported activities...i didn't think through it a lot before, sorry 😞 i think the main thing we'd need to do if we want to go with showing all history, is to keep a separate list of historically active activities....

src/points/PointsHistoryBottomSheet.tsx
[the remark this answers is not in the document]
jophish: so, we actually (with this PR) already have two lists of activities; one is for "active activities" (for lack of a better term), or activities that can show up as _cards on the PointsHome_ screen. the other list are these "claim activities", which are _actually_ just activities that can show up as entries on the history screen. we should be able to modify "active activities" in whatever way we want (adding/removing) without affecting the rendering on the history screen.

> i can imagine a world where we stop rewarding points for creating a wallet, but if you've got points for that then we should still show it no

yeah, so we'd remove `create-wallet` from the regular list of activityIds but keep it in the "claim activities" list (which we should definitely rename, maybe like `historicalActivities`)

to address another one of your comments, the reason why it's important to do this check in _both_ the saga and the selector is because of the case where we _remove_ an historical activity that's already in redux state; when we go to render, the handler to render that historical activity will no longer exist, and stuff will break. (this can also be fixed by a fiddly migration if we want to remove an historical activity.) now, i don't think we should really be removing these historical activities to begin with, but i reckon if we can avoid an edge case now, we may as well do it. in fact, it completely suffices to _just_ do this filtering in the selector and remove it from the saga, i think, but it feels weird to knowingly put things into redux that don't match the schema.

src/points/selectors.ts
jophish: see my other comment!

src/points/PointsHistoryBottomSheet.tsx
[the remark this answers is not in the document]
kathaypacific: ooohhhhh i'm sorry, i didn't understand correctly. i was looking at the diff and thought that we only had 1 activityId array, but i see that you have 2 so i think we're aligned.

i think all i wanted to call out is that we could reduce some complexity in the render logic here by ensuring that it doesn't have to do error handling. so the selector's job is to return render-able data, i think it should just return whatever is in redux (implicitly, we want the redux store to not have any data that act as traps for ourselves)

i don't love the idea of having the same filtering logic in the saga _and_ selector because it's a little nuanced to remember why we have that. from what i understand, we can't avoid some kind of filtering logic in the saga because the points config is for all app versions and we need to ignore stuff that we don't understand. you're also saying that we need filtering logic on the selector in case of persisted bad data (e.g. on a new app release, we accidentally stop supporting a previously supported history item type) - is that right? in that case, i think we're better off adding this logic in the REHYDRATE action which is kicked off on app start. i think this would make the use case more obvious than doing it in the selector, wdyt?

src/points/types.ts
jophish: yeah, this is needed by both `cardDefinitions` and `PointsHistoryBottomSheet`, so i've moved it into `cardDefinitions` as a compromise

src/points/PointsHistoryBottomSheet.tsx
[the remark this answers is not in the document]
jophish: > you're also saying that we need filtering logic on the selector in case of persisted bad data

yep, that's exactly it - i think sticking this logic in the REHYDRATE action is a much better idea; done!

src/points/slice.ts
```diff
@@ -1,10 +1,10 @@
 import { PayloadAction, createSlice } from '@reduxjs/toolkit'
-import { ClaimHistory, PointsActivityId } from 'src/points/types'
+import { RawClaimHistory, PointsActivityId, isClaimActivityId } from 'src/points/types'
 import { REHYDRATE, RehydrateAction, getRehydratePayload } from 'src/redux/persist-helper'
 
 interface GetPointsHistorySucceededAction {
   appendHistory: boolean
-  newPointsHistory: ClaimHistory[]
+  newPointsHistory: RawClaimHistory[]
```
kathaypacific, line 7: small nit: i personally don't like prefixing names with `raw` because it doesn't really help us to understand what it means (aside from the one scenario of being used inside a function that massages data, but even then there are more descriptive names we could use). perhaps we can stick with `ClaimHistory` since that describes the data that we store, and the return type of the selector can be something like `ClaimHistoryCardItem` or `ClaimHistoryItemWithTimestamp`? it would make it clearer what the type is for (i.e. display purposes)

kathaypacific (approved):
LGTM! ✨

src/points/slice.ts
jophish: yeah i didn't love this name either -- updated to use the original meaning of `ClaimHistory`, and the timestamped version now uses `ClaimHistoryCardItem`

src/points/types.ts
```diff
@@ -35,7 +35,12 @@ export type PointsMetadata = {
   }>
 }
 
-type ClaimActivityId = 'create-wallet' | 'swap'
+const claimActivities = ['create-wallet', 'swap'] as const
```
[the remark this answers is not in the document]
jophish: yeah, i wonder that too... there are definitely benefits to using both - i've historically used enums and have only recently started using string literal unions, but i'm kind of warming to them. [this article](https://www.typescriptcourse.com/string-literal-unions-over-enums) is pretty interesting and compares the two. i know @jeanregisser is supporter of using unions of enums.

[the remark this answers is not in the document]
jeanregisser: Yes great article on this, maybe we should bring this up at the next eng team meeting?

src/points/types.ts
```diff
@@ -56,6 +61,12 @@ type SwapClaimHistory = BaseClaimHistory & {
 
 export type ClaimHistory = CreateWalletClaimHistory | SwapClaimHistory
 
+// See https://stackoverflow.com/questions/59794474/omitting-a-shared-property-from-a-union-type-of-objects-results-in-error-when-us
+type DistributiveOmit<T, K extends keyof any> = T extends any ? Omit<T, K> : never
```
jeanregisser, line 65: nit: Should we move this to [src/utils/typescript.ts](https://github.com/valora-inc/wallet/blob/main/src/utils/typescript.ts)?

PR #5262, opened 2024-04-16
fix(qr-scanner): Allow data with network prefixes

jh2oman:
### Description

just ignores the network prefix and reads the address

I considered trying to read and validate the network, but some things stopped me:
1. lack of any kind of plumbing for a 'default network' in the send flow to action upon the prefix. If we want to use it, should probably be another ticket and have specific UX guidelines
2. Lack of consistency for how other wallets mark their data.

| Wallet              | Celo | Ethereum |
| :---------------- | :------: | ----: |
| Metamask        |   ethereum:0x   | ethereum:0x |
| Coinbase           |  celo:0x   | ethereum:0x |
| SafeWallet    |  celo:0x   | ethereum:0x |

Sometimes the wallets mark L2 chains as their own (celo, optimism, arbitrum) and sometimes just label it 'ethereum'. This inconsistency makes it difficult to action upon.

### Test plan

Unit tested with examples pulled from metamask and coinbase

### Related issues

- Fixes https://linear.app/valora/issue/ACT-1155/chain-prefix-support-for-qr-scanner

### DIFF
--- a/src/qrcode/utils.test.tsx
+++ b/src/qrcode/utils.test.tsx
@@ -111,28 +111,31 @@ describe('handleQRCodeDefault', () => {
     })
     expect(ValoraAnalytics.track).toHaveBeenCalledWith(QrScreenEvents.qr_scanned, qrCode)
   })
-  it('navigates to the send amount screen with a qr code with address as the data', async () => {
-    const qrCode: QrCode = { type: QRCodeTypes.QR_CODE, data: mockAccount }
+  it.each([mockAccount, `ethereum:${mockAccount}`, `celo:${mockAccount}`])(
+    'navigates to the send amount screen with a qr code with address as the data',
+    async (data) => {
+      const qrCode: QrCode = { type: QRCodeTypes.QR_CODE, data }
 
-    await expectSaga(handleQRCodeDefault, handleQRCodeDetected(qrCode))
-      .withState(createMockStore({}).getState())
-      .provide([[select(recipientInfoSelector), mockRecipientInfo]])
-      .run()
-    expect(navigate).toHaveBeenCalledWith(Screens.SendEnterAmount, {
-      origin: SendOrigin.AppSendFlow,
-      isFromScan: true,
-      recipient: {
-        address: mockAccount.toLowerCase(),
-        name: mockName,
-        contactId: 'contactId',
-        displayNumber: '14155550000',
-        thumbnailPath: undefined,
-        recipientType: RecipientType.Address,
-      },
-      forceTokenId: false,
-    })
-    expect(ValoraAnalytics.track).toHaveBeenCalledWith(QrScreenEvents.qr_scanned, qrCode)
-  })
+      await expectSaga(handleQRCodeDefault, handleQRCodeDetected(qrCode))
+        .withState(createMockStore({}).getState())
+        .provide([[select(recipientInfoSelector), mockRecipientInfo]])
+        .run()
+      expect(navigate).toHaveBeenCalledWith(Screens.SendEnterAmount, {
+        origin: SendOrigin.AppSendFlow,
+        isFromScan: true,
+        recipient: {
+          address: mockAccount.toLowerCase(),
+          name: mockName,
+          contactId: 'contactId',
+          displayNumber: '14155550000',
+          thumbnailPath: undefined,
+          recipientType: RecipientType.Address,
+        },
+        forceTokenId: false,
+      })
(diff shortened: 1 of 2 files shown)


src/qrcode/utils.ts
```diff
@@ -126,10 +127,17 @@ export function* handleSecureSend(
 }
 
 function* extractQRAddressData(qrCode: QrCode) {
-  // Regex matches any 40 hexadecimal characters prefixed with "0x" (case insensitive)
-  if (/^0x[a-f0-9]{40}$/gi.test(qrCode.data)) {
+  // Check if the QR code is a valid address, does not perform checksum validation
+  if (isAddress(qrCode.data, { strict: false })) {
```
silasbw, line 131: what's this `strict` option?  I don't see it in the docs...

src/qrcode/utils.ts
```diff
@@ -126,10 +127,17 @@ export function* handleSecureSend(
...
     qrCode.data = `celo://wallet/pay?address=${qrCode.data}`
   }
+  // Ignore network prefix and extract address
+  else if (
+    qrCode.data.split(':').length > 1 &&
+    isAddress(qrCode.data.split(':')[1], { strict: false })
+  ) {
+    qrCode.data = `celo://wallet/pay?address=${qrCode.data.split(':')[1]}`
+  }
```
satish-ravi, line 140:
```suggestion
  // strip network prefix if present
  const qrAddress = qrCode.data.replace(/^.*:/, '')
  if (isAddress(qrAddress, { strict: false })) {
    qrCode.data = `celo://wallet/pay?address=${qrAddress}`
  }
```

src/qrcode/utils.test.tsx
```diff
@@ -111,28 +111,31 @@ describe('handleQRCodeDefault', () => {
     })
     expect(ValoraAnalytics.track).toHaveBeenCalledWith(QrScreenEvents.qr_scanned, qrCode)
   })
-  it('navigates to the send amount screen with a qr code with address as the data', async () => {
-    const qrCode: QrCode = { type: QRCodeTypes.QR_CODE, data: mockAccount }
+  it.each([mockAccount, `ethereum:${mockAccount}`, `celo:${mockAccount}`])(
```
satish-ravi, line 114: can we add tests for invalid addresses?

src/qrcode/utils.ts
jh2oman: https://github.com/wevm/viem/blob/main/src/utils/address/isAddress.ts#L12

src/qrcode/utils.ts
```diff
@@ -126,9 +127,10 @@ export function* handleSecureSend(
 }
 
 function* extractQRAddressData(qrCode: QrCode) {
-  // Regex matches any 40 hexadecimal characters prefixed with "0x" (case insensitive)
-  if (/^0x[a-f0-9]{40}$/gi.test(qrCode.data)) {
-    qrCode.data = `celo://wallet/pay?address=${qrCode.data}`
+  // strip network prefix if present
+  const qrAddress = qrCode.data.replace(/^.*:/, '')
```
MuckT, line 131:
```suggestion
  const qrAddress = qrCode.data.split(':').at(-1)
```
This is a bit faster than the regex and has the added benefit of not being regex.

silasbw: +1 on the "not being a regex".  should we check length of array === 2?  to avoid unforeseen stuff like `celo:0xeee:0xfff` (no idea what the might mean, but you get the idea -- kinda a halfhearted suggestion).

src/qrcode/utils.ts
```diff
@@ -126,9 +127,10 @@ export function* handleSecureSend(
...
+  if (isAddress(qrAddress, { strict: false })) {
```
silasbw, line 132: ...so why wouldn't we want `strict: true`?

jh2oman: `Strict: true` would mean that only checksum'd addresses will be valid. I didn't think it wise to add this additional restriction when we don't have any control over the QR codes that are being scanned.

src/qrcode/utils.ts
jh2oman: I'm pretty sure thats not something we'll have to worry about. I believe most of these URL's are based on [ERC-681](https://eips.ethereum.org/EIPS/eip-681) which doesn't include any additional colons

src/qrcode/utils.ts
satish-ravi: I don't think valora address qr codes are checksummed either 😅

satish-ravi (approved):

PR #5267, opened 2024-04-17
feat(points): add valora heart imagery to points screen

kathaypacific:
### Description

As the title.

### Test plan

https://github.com/valora-inc/wallet/assets/20150449/8b2cf3da-54fc-4cb2-abee-cf19cbe3b417

### Related issues

- Fixes RET-1071

### Backwards compatibility

Y

### Network scalability

n/a


branding/celo/src/icons/LogoHeart.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react'
```
kathaypacific, line 1: this is actually just the Logo file, i figure there's no need to create a specific heart version of the non-branded logo

jophish (approved):
yessss! looks so good :)

PR #5297, opened 2024-04-22
chore(deps): upgrade firebase dependencies to v19.1.0

kathaypacific:
### Description

The privacy manifests were included in the firebase iOS [v10.22.0 SDK](https://firebase.google.com/support/release-notes/ios#10.22.0) which is in [v19 of the react native SDK](https://github.com/invertase/react-native-firebase/blob/main/packages/app/CHANGELOG.md#1901-2024-03-07). This is required for Apple privacy policy compliance which goes into effect 1st may.

According to the change log for v18.0.0 and v19.0.0, there should be no breaking changes for us (the breaking changes are around minimum tooling versions).

This PR locks the version to v19.1.0 because this is the last version before the [google `play-services-auth` version is updated to 21](https://github.com/invertase/react-native-firebase/blob/b742500cdd2a2128f4696e4b0b479d901314e2da/packages/app/package.json#L88). [Version 21 of `play-services-auth` removes the Credentials API](https://developers.google.com/android/guides/releases) but unfortunately this is used in `react-native-sms-retriever` which we use in the CPV phone number screen. This lib is not maintained anymore. While this lib declares the version it needs as v17, it seems that 2 versions of `play-services-auth` in one app does not coexist and the newest one is used - the android app crashes on the phone number screen because of the removed Credentials API. Since we cannot upgrade firebase beyond this point until we do something about this sms-retriever lib (probably replace it with an active project, but it's not necessary for this work), I've required the absolute version here.

### Test plan

Rely on CI for app functionality.

Manually test on iOS and Android:
1. remote config values (e.g. the remote config value for `celoNews` should be available in redux in Flipper) ✅
2. rtdb connection (token prices, supercharge remote notification) ✅
3. dynamic link create and decode (generate invite link, verify inviter address on link consume) ✅
4. receive push notification for payment received (using fcm token through firebase messaging dashboard) ✅

### Related issues

- Related to RET-1058

### Backwards compatibility

Y

### Network scalability

Y


src/firebase/notifications.ts
```diff
@@ -57,12 +57,12 @@ export function* handleNotification(
   notificationState: NotificationReceiveState
 ) {
   ValoraAnalytics.track(AppEvents.push_notification_opened, {
-    id: message.data?.id,
+    id: message.data?.id.toString(),
```
kathaypacific, line 60: there was a change to the type declarations where every key in `data` can be a string or object, but it should not affect these particular properties `https://github.com/invertase/react-native-firebase/pull/7316/files`

src/firebase/notifications.ts
```diff
@@ -57,12 +57,12 @@ export function* handleNotification(
   notificationState: NotificationReceiveState
 ) {
   ValoraAnalytics.track(AppEvents.push_notification_opened, {
-    id: message.data?.id,
+    id: message.data?.id as string,
```
kathaypacific, line 60: there was a change to the type declarations where every key in `data` can be a string or object, but it should not affect these particular properties `https://github.com/invertase/react-native-firebase/pull/7316/files`

jophish: i know that these properties should always be strings, so i don't think it's worth adding actual type checks, but stumbling across these assertions always gives me a little pause - maybe add a comment saying that these properties are unaffected in our use case?

jophish (approved):

src/firebase/notifications.ts
bakoushin: what if there _is_ an object for some reason?

I've seen some places with typeguards like this, what do you think @kathaypacific ?

```typescript
if (
  typeof message.data?.id !== 'string' ||
  typeof message.data?.type !== 'string' ||
  typeof message.data?.ou !== 'string'
) {
  return // this should never happen
}
```

bakoushin (approved):

src/firebase/notifications.ts
kathaypacific: TY for the feedback! i've changed it to something like the above :)

bakoushin (approved):
✅

package.json
```diff
@@ -89,13 +89,13 @@
...
-    "@react-native-firebase/dynamic-links": "^17.5.0",
-    "@react-native-firebase/messaging": "^17.5.0",
-    "@react-native-firebase/remote-config": "^17.5.0",
+    "@react-native-firebase/analytics": "19.1.0",
+    "@react-native-firebase/app": "19.1.0",
+    "@react-native-firebase/auth": "19.1.0",
+    "@react-native-firebase/database": "19.1.0",
+    "@react-native-firebase/dynamic-links": "19.1.0",
+    "@react-native-firebase/messaging": "19.1.0",
+    "@react-native-firebase/remote-config": "19.1.0",
```
satish-ravi, line 98: should we temporarily disable renovate for firebase packages? Guess we can also let it open PRs and we can manually close, assuming it will be blocked by CI

kathaypacific: doesn't this change effectively disable renovate by specifying that we want the exact versions of these packages?

the CI failed to catch the crash actually (the previous version of this PR which used a higher version of these packages passed) - i think because the sms retrieval functionality which gets the phone number from the sim card on an emulator doesn't work the same as on a real device. i could consistently repro the crashes on my physical device.

but yes i agree, it was my intention to stop renovate from automatically updating these packages without us knowing (it's pretty bad versioning practice that there is a major version upgrade in an underlying android sdk but on the react native project, it is included in a patch version upgrade)

PR #5308, opened 2024-04-23
feat(wallet-tab): include error message when token fetching fails

satish-ravi:
### Description

The home token balance component included a toast error message when token fetching fails. This adds the same to the wallet tab. Also added more tests for AssetsTokenBalance (used in the wallet tab) and cleaned up some redundant tests in preparation for removing the home token balance (part of cleaning up the drawer nav)

### Test plan

Unit tests, manually failing the token fetch request

### Related issues

- Part of ACT-1133

### Backwards compatibility

Yes

### Network scalability


src/components/TokenBalance.test.tsx
```diff
@@ -713,109 +721,3 @@ describe.each([
     expect(getElementText(tree.getByTestId('TotalTokenBalance'))).toEqual('₱-')
   })
 })
-
-describe('renders the network icon on the home screen to differentiate between ETH on Ethereum, Arbitrum, and Optimism', () => {
```
satish-ravi, line 717: these tests were superfluous, as it was testing the underlying behavior of TokenIcon component using HomeTokenBalance, just updated a test in FiatExchangeTokenBalance to ensure TokenIcon is displayed when there's a single token with balance as HomeTokenBalance is going to be removed as part of the drawer nav cleanup

silasbw (approved):

PR #5318, opened 2024-04-24
refactor: remove drawer navigation

satish-ravi:
### Description

Cleanup drawer navigation. Includes removing:
- DrawerNavigator and other drawer components and icons
- All drawer screens and its associated components including (not exclusive)
  - Removing celo page
  - Removing drawer variants of screens used both in drawer and tabs (Home, Assets/Wallet, Dapps/Discover, Settings, Support, WalletSecurityPrimer, Invite)
  - Removing drawer only home screen components like beta tag, notification spotlight, dapps carousel, home token balance, etc
  - Hiding tx feed balances (now exclusively for wallet tab)
- All other unused items flagged by knip

 Will make a follow up to clean more todos including:
- dropping the statsig gate override launch arg from e2e tests
- dropping settings screen param in the backup quiz flow
- renaming tab screen components 
- other misc todos like EmbeddedNavBar enum, etc

### Test plan

CI, manually going through flows

### Related issues

- Fixes ACT-1133

### Backwards compatibility

Yes

### Network scalability

N/A


src/tokens/AssetList.test.tsx
```diff
@@ -259,42 +254,14 @@ describe('AssetList', () => {
   })
 
   it.each([
-    {
-      name: 'tokens tab on wallet tab when gate is off',
```
finnian0826, line 263: is this test removed, I don't see it below but it looks like it is for tab navigator?

satish-ravi: this test is for when the feature gate is off, we're rolling out the on case, so this specific case isn't needed.

src/backup/BackupPhrase.test.tsx
```diff
@@ -7,38 +7,37 @@ import BackupPhrase from 'src/backup/BackupPhrase'
...
-])('BackupPhrase (settingsScreen: $settingsScreen)', (routeParams) => {
-  it('renders correctly with backup not completed', () => {
-    const tree = render(
-      <Provider store={createMockStore()}>
-        <BackupPhrase {...getMockStackScreenProps(Screens.BackupPhrase)} />
-      </Provider>
-    )
-    expect(tree).toMatchSnapshot()
-  })
+describe.each([{ settingsScreen: Screens.Settings }, {}])(
```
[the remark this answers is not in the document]
satish-ravi: actually this describe.each (added here: https://github.com/valora-inc/wallet/pull/5121/files#diff-6215a8f5c86692eac81da63fc0f0a41ace9c791f2ad68e8d17687a175d817fedR14) isn't really doing anything because `routeParams` is never passed. Just runs the same test multiple times. I am cleaning this up in a follow up PR (#5336) to add better tests

src/tokens/AssetList.test.tsx
finnian0826: ahh okay I was confused because isWalletTab is set to true but then gate is off

satish-ravi: good catch, was confused because there are different gates involved, will add it back

finnian0826 (approved):

src/components/TokenBalance.tsx
```diff
@@ -233,7 +217,7 @@ export function AssetsTokenBalance({
         <TokenBalance
           style={styles.totalBalance}
           singleTokenViewEnabled={false}
-          showBalanceToggle={isWalletTab}
+          showBalanceToggle={true}
```
jh2oman, line 220: I believe this prop can now be removed from TokenBalance

jh2oman (approved):

src/components/TokenBalance.tsx
satish-ravi: FiatExchangeTokenBalance also uses this but that doesn't need it. I think we should just make them different components, but we may change our withdraw spend page, so we can hold off for now

PR #5319, opened 2024-04-24
chore(deps): upgrade Sentry to v5

kathaypacific:
### Description

It seems like the Apple privacy manifest requirements were met with version 5.19.3 (`https://github.com/getsentry/sentry-react-native/issues/3708#issuecomment-2017636873`).

I followed the migration guide [here](https://docs.sentry.io/platforms/react-native/migration/#from-4x-to-5x).

### Test plan

In Sentry, for iOS and Android:
- I can see an [error](https://valora-inc.sentry.io/issues/?end=2024-04-24T10%3A59%3A59&environment=org.celo.mobile.alfajores.dev&project=1250733&query=&referrer=issue-list&start=2024-04-24T09%3A00%3A00&utc=true) and the breadcrumbs
- I can see [performance traces](https://valora-inc.sentry.io/performance/?end=2024-04-24T10%3A59%3A59&environment=org.celo.mobile.alfajores.dev&project=1250733&start=2024-04-24T09%3A00%3A00&utc=true)

### Related issues

- Related to RET-1058

### Network scalability


src/app/App.tsx
```diff
@@ -51,7 +51,7 @@ BigNumber.config({
   },
 })
 
-interface Props {
+interface Props extends Record<string, unknown> {
```
kathaypacific, line 54: this is to satisfy Sentry's updated `wrap` types (we use `Sentry.wrap(App)` below)

bakoushin: surprisingly I don't see TS complaining if I remove that

bakoushin (approved):

src/app/App.tsx
kathaypacific: oh weird, you can see the unit test / iOS and android test jobs failing on the CI in [this commit](https://github.com/valora-inc/wallet/pull/5319/commits/ce81c56aad0bc373799f210fe0e83288b2f6973a) which doesn't have this change. perhaps you need to update dependencies and restart vs code? i got into a weird state with TS in vs code a few times doing all these upgrades

bakoushin: ah, I had to restart vs code. now it yells!

PR #5321, opened 2024-04-24
chore(deps): remove firebase analytics

kathaypacific:
### Description

Removing because I don't think this is used - I can't find references of it in our codebase and looking at the yarn.lock / podfile.lock, it doesn't seem to be needed by any of our dependencies either...

### Test plan

n/a

### Related issues

- Related to RET-1058

### Backwards compatibility

Y

### Network scalability


kathaypacific:
welp, it looks like it is used!

PR #5322, opened 2024-04-24
chore(deps): patch react-native-fs with privacy manifest

kathaypacific:
### Description

The `react-native-fs` project is no longer maintained, and I don't have faith that a new version with the changes we need will be published. It seems like the maintainer of the project has changed but he is now publishing a fork rather than publishing from this project (context `https://github.com/itinance/react-native-fs/issues/1197`). The fork has a lot of [changes](https://github.com/itinance/react-native-fs/compare/master...birdofpreyru:react-native-fs:master) and it seems that most projects have _not_ yet migrated to this fork ([low downloads on npm](https://www.npmjs.com/package/@dr.pogodin/react-native-fs)). I don't want to make a decision now about which lib to migrate to and introduce more changes than necessary given the deadline for these privacy manifest changes. So this PR contains a patch that I made for the lib, and I used the [same declared values](https://github.com/birdofpreyru/react-native-fs/blob/master/ios/PrivacyInfo.xcprivacy) as in the fork (I did also verify that the values seem reasonable).

### Test plan

n/a

### Related issues

- Related to RET-1058

### Backwards compatibility

Y

### Network scalability

### DIFF
--- a/patches/react-native-fs+2.20.0.patch
+++ b/patches/react-native-fs+2.20.0.patch
@@ -1,0 +1,52 @@
+diff --git a/node_modules/react-native-fs/PrivacyInfo.xcprivacy b/node_modules/react-native-fs/PrivacyInfo.xcprivacy
+new file mode 100644
+index 0000000..23f6d7f
+--- /dev/null
++++ b/node_modules/react-native-fs/PrivacyInfo.xcprivacy
+@@ -0,0 +1,26 @@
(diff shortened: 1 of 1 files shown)


bakoushin (approved):
🍀

jeanregisser (approved):

jophish (approved):

PR #5323, opened 2024-04-24
fix(feed): use tx status from the blockchain-api

bakoushin:
### Description

Use transaction status returned from the blockchain-api instead of the hardcoded `TransactionStatus.Success` value.

Does not invalidate local transactions cache expecting it to be updated as soon as there is an internet connection.

### Test plan

* Tested manually

image

### Related issues

* Fixes RET-1057
* Fixes ACT-1167

### Backwards compatibility

Y

### Network scalability


kathaypacific (approved):
LGTM! it'd be great to add a test (i know it might be a pain, but maybe there's some existing test in the TransactionFeed tests we can add some statuses to)

jeanregisser (approved):
🚀

silasbw (approved):

PR #5326, opened 2024-04-24
chore: Switch polygon testnet from mumbai to amoy

finnian0826:
### Description

title

### Test plan

builds

### Related issues

- Fixes #ACT-1164

### Backwards compatibility

Yes

### Network scalability


src/web3/networkConfig.ts
```diff
@@ -561,7 +561,7 @@ export const networkIdToWalletConnectChainId: Record<NetworkId, string> = {
   [NetworkId['op-mainnet']]: 'eip155:10',
   [NetworkId['op-sepolia']]: 'eip155:11155420',
   [NetworkId['polygon-pos-mainnet']]: 'eip155:137',
-  [NetworkId['polygon-pos-mumbai']]: 'eip155:80001',
+  [NetworkId['polygon-pos-amoy']]: 'eip155:80001',
```
MuckT, line 564:
```suggestion
  [NetworkId['polygon-pos-amoy']]: 'eip155:80002',
```

MuckT (changes requested):
Chain id change: https://chainlist.org/chain/80002.

src/web3/networkConfig.ts
finnian0826: good catch, just updated

MuckT (approved):

PR #5332, opened 2024-04-25
chore(deps): patch react-native-cookies with privacy policy

kathaypacific:
### Description

This PR puts the following change into a patch: `https://github.com/react-native-cookies/cookies/pull/189/files` - the last publish for this lib was 2 years ago and I am concerned that this will not be published in time.

### Test plan

n/a

### Related issues

- Related to RET-1058

### Backwards compatibility

Y

### Network scalability


jeanregisser (approved):
🙏

PR #5333, opened 2024-04-25
chore(deps): upgrade some indirect Pod dependencies for privacy policy

kathaypacific:
### Description

There were some libs that we indirectly depend on that were flagged by [this tool](https://github.com/stelabouras/privacy-manifest). This PR bumps the pod versions so that the privacy manifest is included.

### Test plan

n/a

### Related issues

- Related to RET-1058

### Backwards compatibility

Y

### Network scalability


kathaypacific:
This isn't strictly necessary anymore after #5334

PR #5266, opened 2024-04-17
chore(iOS): update privacy policy

kathaypacific:
### Description

Please check the [comment](https://linear.app/valora/issue/RET-1058/apple-build-issues#comment-2b75e79e) in the linear task for detailed commentary on the values chosen.

### Test plan

n/a

### Related issues

- Fixes RET-1058

### Backwards compatibility

Y

### Network scalability

n/a

PR #5334, opened 2024-04-25
chore(iOS): declare privacy manifest

kathaypacific:
### Description

So I think that we _do_ need one of these. The contents are different to my [initial attempt](https://github.com/valora-inc/wallet/pull/5266), it now includes the user defaults api and is linked to the NotificationService plugin.

[Here](https://github.com/valora-inc/wallet/blob/main/ios/celo/AppDelegate.mm#L154) we are using the [user default API](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278401).

I've struggled to understand whether we need to declare on behalf of third party sdks. The [apple docs](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#overview) seem to say no (`If you use the API in your third-party SDK’s code, then you need to report the API in your third-party SDK’s privacy manifest file.`) but the new react native implementation seems to generate the privacy policy docs for new projects as part of the pod install step and add the stuff specific for RN (`https://github.com/facebook/react-native/commit/d39712f54a95e6dcdf1d2f80d9581211ab03c157`). This PR attempts to declare everything on the app level, to test if it passes the review.

### Test plan

n/a

### Related issues

- Related to RET-1058

### Backwards compatibility

Y

### Network scalability

### DIFF
--- a/ios/PrivacyInfo.xcprivacy
+++ b/ios/PrivacyInfo.xcprivacy
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>NSPrivacyAccessedAPITypes</key>
+	<array>
+		<dict>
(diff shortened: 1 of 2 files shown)


jeanregisser (approved):
🤞

PR #5341, opened 2024-04-26
refactor(education): remove unnecessary prop and enum

satish-ravi:
### Description

- Removed EmbeddedNavBar enum (only one value)
- Removed onboarding from EducationTopic
- Cleanup unused analytics events

### Test plan

CI

### Related issues

Part of ACT-1133

### Backwards compatibility

Yes

### Network scalability

N/A


src/analytics/docs.ts
```diff
@@ -72,10 +72,7 @@ export const eventDocs: Record<AnalyticsEventType, string> = {
   [AppEvents.multichain_beta_opt_out]: `When the user taps the No Thanks button on the multichain beta screen`,
   [AppEvents.multichain_beta_contact_support]: `When the user taps the Contact Support button on the multichain beta screen`,
   [AppEvents.handle_deeplink]: `When a deeplink that leads into the app is detected and handled`,
-  [HomeEvents.hamburger_tapped]: ``,
```
MuckT, line 75: Should these be added to the legacy events at the bottom of the file?

MuckT (approved):
Approved with the question about adding to the removed events to Legacy event docs, but as there is no description it also seems fine to exclude.

src/analytics/docs.ts
satish-ravi: yeah I thought about it, but the event names are pretty verbose, so ignored them.

PR #5345, opened 2024-04-26
feat(points): add number ticker for balance

kathaypacific:
### Description

As the title. The approach was largely inspired by [a nice man on youtube](https://www.youtube.com/watch?v=nhshusc5ya0&list=PL7ycojg9_ZoJwnayMF8n-VlhdXnE_grdg&index=1).

I've added support for decimals in case we want to use this for the wallet balance. 

The ticker animation will run any time the value changes, so it will work automatically for when the points balance changes and the user remains on the screen.

The downside with animations is that there's a lot of smoke and mirrors that mess with the dom. For this ticker, we're basically animating columns of text up and down. From the dom it no longer looks like the actual balance and we lose the ability to assert on its value from the unit tests.

### Test plan

Without decimals:

https://github.com/valora-inc/wallet/assets/20150449/a9ce407e-544e-4a69-85c1-9219f629360b

With decimals:

https://github.com/valora-inc/wallet/assets/20150449/4128f00a-0949-4321-98ba-8dda0cd23c89

### Related issues

- Fixes RET-1072

### Backwards compatibility

Y

### Network scalability

### DIFF
--- a/src/components/NumberTicker.tsx
+++ b/src/components/NumberTicker.tsx
@@ -2,29 +2,27 @@ import * as React from 'react'
 import { Animated, StyleProp, StyleSheet, Text, TextStyle, View } from 'react-native'
 import { typeScale } from 'src/styles/fonts'
 
-interface Props {
-  finalValue: string
+interface CommonProps {
   textHeight: number
   textStyles?: StyleProp<TextStyle>
+}
+interface Props extends CommonProps {
+  finalValue: string
   testID?: string
 }
 
-interface TickProps extends Omit<Props, 'finalValue'> {
+interface TickProps extends CommonProps {
   startValue: number
   endValue: number
 }
 
+interface TickTextProps extends CommonProps {
+  value: string
+}
+
 const numberRange = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9']
 
-function TickText({
-  value,
-  textHeight,
-  textStyles,
-}: {
-  value: string
-  textHeight: number
-  textStyles?: StyleProp<TextStyle>
-}) {
+function TickText({ value, textHeight, textStyles }: TickTextProps) {
   return (
     <View style={[styles.tickText, { height: textHeight }]}>
       <Text style={[styles.text, textStyles]}>{value}</Text>
--- a/src/components/NumberTicker.tsx
+++ b/src/components/NumberTicker.tsx
@@ -100,13 +100,13 @@ const styles = StyleSheet.create({
   container: {
     overflow: 'hidden',
     flexDirection: 'row',
+    // This negative gap is a hack to bring the numbers closer together,
+    // otherwise they feel unnatural and far apart
+    gap: -2,
   },
   tickText: {
     alignItems: 'center',
     justifyContent: 'center',
-    // This negative margin is a hack to bring the numbers closer together,
-    // otherwise they feel unnatural and far apart
-    marginHorizontal: -1,
   },
   text: {
     ...typeScale.displaySmall,
--- a/src/components/NumberTicker.tsx
+++ b/src/components/NumberTicker.tsx
@@ -5,6 +5,7 @@ import { typeScale } from 'src/styles/fonts'
 interface CommonProps {
   textHeight: number
   textStyles?: StyleProp<TextStyle>
+  animationDuration?: number
 }
 interface Props extends CommonProps {
   finalValue: string
@@ -30,13 +31,14 @@ function TickText({ value, textHeight, textStyles }: TickTextProps) {
   )
 }
 
-function Tick({ startValue, endValue, textHeight, textStyles }: TickProps) {
+function Tick({ startValue, endValue, textHeight, textStyles, animationDuration }: TickProps) {
   const animatedValue = new Animated.Value(startValue * textHeight * -1)
   const transformStyle = { transform: [{ translateY: animatedValue }] }
+  const duration = animationDuration ?? 1300
 
   Animated.timing(animatedValue, {
     toValue: endValue * textHeight * -1,
-    duration: 1300,
+    duration,
     useNativeDriver: true,
   }).start()
 
@@ -51,7 +53,13 @@ function Tick({ startValue, endValue, textHeight, textStyles }: TickProps) {
   )
 }
 
-export default function NumberTicker({ finalValue, textStyles, textHeight, testID }: Props) {
+export default function NumberTicker({
+  finalValue,
+  textStyles,
+  textHeight,
+  animationDuration,
+  testID,
+}: Props) {
   const finalValueArray = finalValue.toString().split('')
 
   // For the startValueArray, map over each character in the finalValueArray to
@@ -80,6 +88,7 @@ export default function NumberTicker({ finalValue, textStyles, textHeight, testI
             endValue={endValue}
             textHeight={textHeight}
             textStyles={textStyles}
+            animationDuration={animationDuration}
           />
         )
       })}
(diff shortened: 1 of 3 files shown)


src/points/PointsHome.tsx
```diff
@@ -121,7 +122,12 @@ export default function PointsHome({ route, navigation }: Props) {
               />
             </View>
             <View style={styles.balanceRow}>
-              <Text style={styles.balance}>{pointsBalance}</Text>
+              <NumberTicker
+                testID="PointsBalance"
+                textStyles={styles.balance}
+                textHeight={48}
+                finalValue={pointsBalance.toString()}
```
bakoushin, line 129: i wonder if this prop could be simpler?
* named just `value` 
* of a number type

src/components/NumberTicker.tsx
```diff
@@ -0,0 +1,104 @@
...
+      <Text style={[styles.text, textStyles]}>{value}</Text>
+    </View>
+  )
+}
+
+function Tick({ startValue, endValue, textHeight, textStyles }: TickProps) {
+  const animatedValue = new Animated.Value(startValue * textHeight * -1)
+  const transformStyle = { transform: [{ translateY: animatedValue }] }
+
+  Animated.timing(animatedValue, {
```
bakoushin, line 37: what is the reason for not choosing reanimated?

src/points/PointsHome.tsx
```diff
@@ -121,7 +122,12 @@ export default function PointsHome({ route, navigation }: Props) {
               />
             </View>
             <View style={styles.balanceRow}>
-              <Text style={styles.balance}>{pointsBalance}</Text>
+              <NumberTicker
+                testID="PointsBalance"
+                textStyles={styles.balance}
+                textHeight={48}
```
bakoushin, line 128: 1. What's this magic number? I had to drill down into `styles.balance` and then to `typeScale.displaySmall` to find out.
2. It will go out of sync if suddenly the `lineHeight` property in the applied style changes.
3. I guess it could be less painful to measure the height inside the component, instead of forcing the engineer to figure it out.

src/points/PointsHome.tsx
kathaypacific: i used a string in case we want to pass in a very big number in here or a localised number (e.g. 12,34) - i'm not sure if these scenarios are likely to be used, but i need to understand the number of digits to animate and the easiest way to get that is to turn it into a string anyway. i guess i don't see many downsides of using a string?

src/components/NumberTicker.tsx
kathaypacific: no reason, this solution was inspired by someone on youtube who chose Animated from RN. i think given the option, i'd choose RN over other libs anyway because RN will always exist but decreasing our dependencies is always a good thing?

src/points/PointsHome.tsx
kathaypacific: i agree with your concerns but also they're kind of reasons for why i added this prop. for this particular case, the textHeight is equal to the lineHeight of the textStyles but they can also be independent. it is a prop that sets how tall the component is. by allowing the text styles to be customised, i think it becomes dangerous to derive the height based on the passed in styles (what if there are margins and paddings aside from the line height?)

i guess this complexity can be avoided by not allowing the styles to be customised for now? i thought i'd add it in while the animation logic is fresh in my mind, rather than punt it to the future when someone needs to figure it out

src/components/NumberTicker.tsx
```diff
@@ -0,0 +1,114 @@
...
+export default function NumberTicker({
+  finalValue,
+  textStyles,
+  textHeight,
+  animationDuration,
+  testID,
+}: Props) {
+  const finalValueArray = finalValue.toString().split('')
+
+  // For the startValueArray, map over each character in the finalValueArray to
```
bakoushin, line 65: perhaps we could use [accessibility properties](https://reactnative.dev/docs/accessibility#accessible) `accessible` and `accessibilyLabel` to ensure that we have the actual number for the screen readers and unit tests?

src/components/NumberTicker.tsx
```diff
@@ -0,0 +1,114 @@
...
+            textStyles={textStyles}
+            animationDuration={animationDuration}
+          />
+        )
+      })}
+    </View>
+  )
+}
+
+const styles = StyleSheet.create({
```
bakoushin, line 99: while indeed reducing letter spacing, this will shift the whole number 1px to the left, right?
i wonder if we can get the same result using [`columnGap`](https://reactnative.dev/docs/layout-props#columngap) which shouldn't affect the edge elements

bakoushin: ah, apparently `columnGap` wouldn't help.
but maybe we could cancel horizontal shift my adding `marginHorizontal: 1` to the container?
not sure though if 1-2 px shift is even noticeable 😅

src/components/NumberTicker.tsx
kathaypacific: how does this work? would we have to apply custom logic for the values of accessible and accessibilyLabel based on the animation status? (i think we don't get a value like "isVisible" or "isAnimating" after rendering in a unit test environment). i thought about testing with accessiblity labels but ultimately decided that if i need to apply logic to set the values, then i'd be testing that logic rather than the animation. we're also not prioritising accessibility and screen readers (i.e. we're not designing with this in mind and we have not added any type of testing for these topics) so i don't really feel the need to go through extra effort to make sure this fun animation works with screen readers....

src/components/NumberTicker.tsx
kathaypacific: i can apply a negative margin to the right side only, i meant to do that but forgot

bakoushin (approved):
🎉 This is so cool!

I noticed that the points balance is designed to be even bigger than the assets balance. Now with a nice animation, it definitely looks like the most important number! 😁

src/components/NumberTicker.tsx
kathaypacific: i applied a negative gap, thanks for reminding me about this!!

bakoushin: yeah, but this will expect that this component will be always used aligned to the left.
what if suddenly we would like to align it to the right, or center it?

src/components/NumberTicker.tsx
bakoushin: I was thinking of just passing through the final value straight to the `accessibilityLabel`, so we can search for it in the tests, ensuring the element is rendered with this value.

src/components/NumberTicker.tsx
kathaypacific: i think the negative gap would account for that since it's applied to the container containing the numbers

src/points/PointsHome.tsx
bakoushin: yes, this makes totally sense

src/components/NumberTicker.tsx
kathaypacific: i think i prefer no test coverage to test coverage that doesn't directly test the outcome (false sense of security makes us more reckless). i was thinking that (import) animations should be tested manually instead, perhaps in our release testing steps

src/components/NumberTicker.tsx
bakoushin: I agree that having fewer dependencies is good if we can achieve the same goals without them. but we are already using reanimated and I wonder if we should be consistent in our approach to animation.

src/points/PointsHome.tsx
bakoushin: I think it would make sense to have a single number affecting the size (either a prop or a style)

currently, I can provide any arbitrary number in this prop and get something I don't know why I would ever need:

| 100px | 20px |
|--------|--------|
| image | image |

PR #5359, opened 2024-04-29
feat(swap): enable app fee via remote config

jeanregisser:
### Description

This PR allows enabling the swap fee via the remote config.

When enabled, the free Valora swap fee line item is hidden.
The details of the Valora fee are then displayed when hitting the info button of the exchange rate line item.

### Test plan

- Updated tests

| Valora fee disabled | Valora fee enabled (0 fee from API) | Valora fee enabled (positive fee from API) |
|----------|----------|----------|
| image  | image   | image   |
| image | image   | image   |

Notes: 
- copies are not final
- the "Valora fee disabled" case is same as today except it also shows the info icon next to the exchange rate.

### Related issues

- Fixes RET-1061

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/locales/base/translation.json
+++ b/locales/base/translation.json
@@ -1824,7 +1824,7 @@
       "networkFeeInfoDismissButton": "Got it",
       "exchangeRate": "Exchange Rate",
       "exchangeRateInfo": "Exchange Rate\n\nWe've found the best exchange rate for you. Your trade will be completed at a price within {{slippagePercentage}}% of the estimate or the transaction will be cancelled & funds returned to your wallet.",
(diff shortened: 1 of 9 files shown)


kathaypacific (approved):
LGTM!! i wondered if we want to hard code "valora" in variable names and translation keys / params for the fees? we've tried to be brand agnostic in the code so far so wanted to check that this was an intentional move (i was thinking that the readability wouldn't be affected if we omit the "valora" part)

jeanregisser:
> LGTM!! i wondered if we want to hard code "valora" in variable names and translation keys / params for the fees? we've tried to be brand agnostic in the code so far so wanted to check that this was an intentional move (i was thinking that the readability wouldn't be affected if we omit the "valora" part)

It's intentional, but I see your point. 
If we remove the "valora" part I think it's way less clear what this is, cause we already have network fee, gas fee, and soon cross-chain fee.
Just calling it "swap fee" may be a bit ambiguous I think, as there are liquidity pool swap fees already which we don't really see.
Maybe we could call it `appFee`? or `appEditorFee`? or `swapApiFee`?

PR #5361, opened 2024-04-29
feat(points): Implement empty state for points history bottom sheet

jophish:
### Description

Implements the empty state for the points history bottom sheet. See designs [here](https://www.figma.com/file/rXBDplfMHHqYmuu6EkgMEo/Gamification-experiments?type=design&node-id=1486-5899&mode=design&t=gVtyIzOM0LmxJaDy-4).

### Test plan

Units and manual tested. See video below

https://github.com/valora-inc/wallet/assets/569401/b9ed634a-4784-4c6a-a7db-fdbeaf9eb395

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

Yes

### Network scalability


src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -115,8 +120,23 @@ function PointsHistoryBottomSheet({ forwardedRef }: Props) {
...
+          onPress={onPressLearnMore}
+          text={t('points.history.empty.learnMore')}
+          type={BtnTypes.GRAY_WITH_BORDER}
+          size={BtnSizes.FULL}
+          style={{ width: '100%' }}
+        />
+      </View>
+    )
+
+  const renderEmpty = pointsHistoryStatus !== 'loading' && !pointsHistory.length
```
jeanregisser, line 139: nit: an attempt to make it slightly more self-explanatory
```suggestion
  const isEmpty = pointsHistoryStatus !== 'loading' && !pointsHistory.length
```

jeanregisser (approved):
🚀

kathaypacific (approved):
looks good, although the "learn more" button felt a odd to me because it feels like it should do something more than just dismissing the sheet. i wonder if it'd be better for the button text to say something like "got it"? or for us to link to a support article? just my 2c though, if you and Kayla want to leave it like this then that's also fine

PR #5362, opened 2024-04-29
feat(points): Add error banner on failure fetching subsequent points history pages

jophish:
### Description

Adds in an inline notification error when there's a failure fetching subsequent points history page.s See designs [here](https://www.figma.com/file/rXBDplfMHHqYmuu6EkgMEo/Gamification-experiments?type=design&node-id=1486-6438&mode=design&t=gVtyIzOM0LmxJaDy-4).

### Test plan

https://github.com/valora-inc/wallet/assets/569401/c5cb9fc4-50a3-424c-84f6-dab6fb10f981

### Related issues

- Fixes #[issue number here]

### Backwards compatibility

Yes.

### Network scalability


src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -55,6 +57,8 @@ function PointsHistoryBottomSheet({ forwardedRef }: Props) {
   const pointsHistoryStatus = useSelector(pointsHistoryStatusSelector)
   const pointsHistory = useSelector(pointsHistorySelector)
 
+  const [showErrorBanner, setShowErrorBanner] = useState(false)
```
kathaypacific, line 60:
```suggestion
  const [showError, setShowError] = useState(false)
```
nit: technically not a banner (and the type of display component could change)

src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -138,12 +152,16 @@ function PointsHistoryBottomSheet({ forwardedRef }: Props) {
 
   const renderEmpty = pointsHistoryStatus !== 'loading' && !pointsHistory.length
 
-  // TODO: Figure out what to render when error occurs on subsequent page fetch
-
   const sections = useMemo(() => {
     return groupFeedItemsInSections([], pointsHistory)
   }, [pointsHistory, pointsHistoryStatus])
 
+  useMemo(() => {
```
kathaypacific, line 159:
```suggestion
  useEffect(() => {
```

src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -165,11 +183,27 @@ function PointsHistoryBottomSheet({ forwardedRef }: Props) {
...
       />
+      {showErrorBanner && (
+        <InLineNotification
+          variant={NotificationVariant.Error}
+          title={t('points.history.pageError.title')}
+          description={t('points.history.pageError.subtitle')}
+          ctaLabel={t('points.history.pageError.refresh')}
+          onPressCta={onPressRefresh}
+          withBorder={true}
+          style={styles.errorBanner}
```
kathaypacific, line 194: you'll also need to add something like `{ paddingBottom: Math.max(insets.bottom, Spacing.Thick24) }` here, otherwise this content collides with the bottom bar.

Simulator Screenshot - iPhone 14 Pro - 2024-04-30 at 09 49 55

src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -92,6 +96,16 @@ function PointsHistoryBottomSheet({ forwardedRef }: Props) {
     forwardedRef.current?.close()
   }
 
+  const onPressRefresh = () => {
```
kathaypacific, line 99:
```suggestion
  const onPressRefreshNextPage = () => {
```
actually i wonder if you want to consider combining this with `onPressTryAgain`? they're doing the same thing and could probably just take a `getNextPage` as input. we could also consider having the same analytics event with a property for `getNextPage` (since it's the same user action, to retry the fetch?)

src/points/PointsHistoryBottomSheet.tsx
kathaypacific: i didn't notice this before, but it seems like the section headers are sticky? is that expected? the homefeed doesn't do this and i guess it doesn't matter so much but in my screenshot above you can see some clipped content which was unexpected for me

kathaypacific (approved):
LGTM, would be great to address my comments before merging!

src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -165,11 +183,27 @@ function PointsHistoryBottomSheet({ forwardedRef }: Props) {
...
+        />
+      )}
     </BottomSheetBase>
   )
 }
 
 const styles = StyleSheet.create({
+  errorBanner: {
+    marginHorizontal: Spacing.Regular16,
+  },
```
jeanregisser, line 206: I suggest we use absolute positioning (over the content) toward the bottom for this one.
So it doesn't abruptly push the content up when it appears.

jeanregisser (commented):
👍

src/points/PointsHistoryBottomSheet.tsx
jophish: Hmm, this actually isn't something I've ever experienced on Android -- I wonder if this is a platform-specific thing?

src/points/PointsHistoryBottomSheet.tsx
jophish: yeah, good callout, combined them into a single fn that shares an analytics event.

src/points/PointsHistoryBottomSheet.tsx
satish-ravi: I ran into this on a different screen. https://reactnative.dev/docs/sectionlist#stickysectionheadersenabled - defaults to true on iOS

PR #5368, opened 2024-04-30
feat(swap): track valora fee

jeanregisser:
### Description

As the title says.

### Test plan

- Updated tests

### Related issues

- Fixes RET-1061

### Backwards compatibility

Yes

### Network scalability


kathaypacific (approved):
🚀

PR #5369, opened 2024-04-30
chore: simplify the NumberTicker interface

kathaypacific:
### Description

As the title, this addresses some discussions we had in #5345 about possible misuse of the props. This PR makes it harder to screw up the styles.

### Test plan

No visual changes

https://github.com/valora-inc/wallet/assets/20150449/2b203175-13a2-4dc3-bbd5-e53923e82d5a

### Related issues

- Fixes RET-1072

### Backwards compatibility

Y
### Network scalability

### DIFF
--- a/src/components/NumberTicker.tsx
+++ b/src/components/NumberTicker.tsx
@@ -2,36 +2,33 @@ import * as React from 'react'
 import { Animated, StyleProp, StyleSheet, Text, TextStyle, View } from 'react-native'
 import { typeScale } from 'src/styles/fonts'
 
-interface CommonProps {
-  textHeight: number
-  textStyles?: StyleProp<TextStyle>
-  animationDuration?: number
-}
-interface Props extends CommonProps {
+interface Props {
   finalValue: string
+  typeScaleName?: keyof typeof typeScale
+  animationDuration?: number
   testID?: string
 }
 
-interface TickProps extends CommonProps {
-  startValue: number
-  endValue: number
+interface TickTextProps {
+  textStyles: StyleProp<TextStyle>
+  value: string
 }
 
-interface TickTextProps extends CommonProps {
-  value: string
+interface TickProps {
+  startValue: number
+  endValue: number
+  animationDuration: number
+  textHeight: number
+  textStyles: StyleProp<TextStyle>
 }
 
 const numberRange = ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9']
 
-function TickText({ value, textHeight, textStyles }: TickTextProps) {
-  return (
-    <View style={[styles.tickText, { height: textHeight }]}>
-      <Text style={[styles.text, textStyles]}>{value}</Text>
-    </View>
-  )
+function TickText({ value, textStyles }: TickTextProps) {
+  return <Text style={[styles.tickerText, textStyles]}>{value}</Text>
 }
 
-function Tick({ startValue, endValue, textHeight, textStyles, animationDuration }: TickProps) {
+function Tick({ startValue, endValue, textStyles, textHeight, animationDuration }: TickProps) {
   const animatedValue = new Animated.Value(startValue * textHeight * -1)
   const transformStyle = { transform: [{ translateY: animatedValue }] }
   const duration = animationDuration ?? 1300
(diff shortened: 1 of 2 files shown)


src/components/NumberTicker.tsx
```diff
@@ -2,36 +2,33 @@ import * as React from 'react'
...
 import { typeScale } from 'src/styles/fonts'
 
-interface CommonProps {
-  textHeight: number
-  textStyles?: StyleProp<TextStyle>
-  animationDuration?: number
-}
-interface Props extends CommonProps {
+interface Props {
   finalValue: string
```
bakoushin, line 6: do we need to emphasize that this is indeed a _final_ value here?

src/components/NumberTicker.tsx
```diff
@@ -43,23 +40,22 @@ function Tick({ startValue, endValue, textHeight, textStyles, animationDuration
...
 export default function NumberTicker({
   finalValue,
-  textStyles,
-  textHeight,
-  animationDuration,
+  typeScaleName = 'displaySmall',
+  animationDuration = 1300,
   testID,
 }: Props) {
+  const textStyles = typeScale[typeScaleName]
```
bakoushin, line 57: having a prop with a style name to looks too smart to me: why not just provide a `style` as it is? in that case we still can extract a `lineHeight` from it, right?

also, what if the style doesn't have a `lineHeight` property? perhaps we should have a reasonable default? something like 1.2

src/components/NumberTicker.tsx
```diff
@@ -2,36 +2,33 @@ import * as React from 'react'
...
 }
 
-interface TickTextProps extends CommonProps {
-  value: string
+interface TickProps {
+  startValue: number
+  endValue: number
+  animationDuration: number
+  textHeight: number
+  textStyles: StyleProp<TextStyle>
```
bakoushin, line 22: super nit: i wonder why it's plural?

bakoushin (approved):
💅 Looks great!

bakoushin:
Like the idea of extracting the line height from the style!

src/components/NumberTicker.tsx
kathaypacific: ah, good question. i guess not, but i started with this because i wanted to enable a `startValue`. for example, if we want to use this animation, maybe we want to only animate the changes from the first balance to the updated balance. i then decided against allowing a custom startValues because 1) after playing around with the animation, i think starting from a random number is not visually jarring and draws more emphasis to the changing data, 2) the logic for ensuring that the start value is compatible with the final value got very complex (trying to ensure the same number of digits, that the decimal is in the right place in both number, etc). so i can change this back to a simple `value`

src/components/NumberTicker.tsx
kathaypacific: 2 reasons: 
1. once we've described a prop as a StyleProp in typescript, you can't reference the style properties easily anymore
2. exactly what you said - what if the style doesn't have a lineHeight? so i think it is better to have a prop that enforces us to use `typeScale` and not pass in unexpected combinations of styles

bakoushin: but this implies that `typeScale` must have the `lineHeight` prop. if we remove that prop from it, this component will break. looks like we are requiring the `typeScale` to have a specific shape from now on.

what is we use [StyleSheet.flatten()](https://reactnative.dev/docs/stylesheet#flatten)?

then we could extract the line height somehow like this:
```typescript
const textHeight = StyleSheet.flatten(textStyle).lineHeight ?? DEFAULT_TEXT_HEIGHT
```

the main benefit I see the consumer can provide any style and it still works.

kathaypacific: i'm not sure that we can or try to strive for "the consumer can provide any style and it still works". if a consumer applies a style that has no lineHeight + a fontSize that is bigger than DEFAULT_TEXT_HEIGHT, then this still breaks. 

if i make a style prop that requires fontSize and lineHight to be both defined, the consumer can _still_ add a vertical padding or margin that can screw up the styles.

i think this is a byproduct of trying to make something cater for too many scenario (i.e. trying to allow all styles to come through means we need to think about all combinations that could not work) and it's unnecessary complexity to maintain. i prefer to keep the degree of customisability to a minimum, which is similar about design system components. wdyt?

bakoushin: yes, i agree! 
there are still many ways for consumers to screw that up.
i also like the idea of sticking with a design system discouraging arbitrary customization. 

apologies for introducing the confusion -- i was mistakenly treating `lineHeight` numeric values as _ratios_ similar to CSS ones, while they are absolute values in RN 😖

kathaypacific: i did a quick check to make sure that there are typescript errors when `lineHeight` is removed on any object in `typeScale`, so we reduce the risk of accidentally breaking this. (of course styles can be _added_ to `typesScale` to break this component though 😅)

src/components/NumberTicker.tsx
kathaypacific: that's true, i'm not sure why i got into the habit of using plural for styles when they are props. changed it to singular

PR #5373, opened 2024-04-30
feat: add stablecoin earn feature gate

satish-ravi:
### Description

Just setting up the gate in the code, will be used in subsequent PRs

### Test plan

N/A

### Related issues

- Fixes ACT-1183

### Backwards compatibility

Yes

### Network scalability

N/A

### DIFF
--- a/src/statsig/constants.ts
+++ b/src/statsig/constants.ts
@@ -27,6 +27,7 @@ export const FeatureGates = {
   [StatsigFeatureGates.SHOW_NFT_REWARD]: false,
   [StatsigFeatureGates.SHOW_JUMPSTART_SEND]: false,
   [StatsigFeatureGates.SHOW_POINTS]: false,
+  [StatsigFeatureGates.SHOW_STABLECOIN_EARN]: false,
 } satisfies { [key in StatsigFeatureGates]: boolean }
 
 export const ExperimentConfigs = {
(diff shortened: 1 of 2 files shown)


MuckT (approved):

src/statsig/constants.ts
```diff
@@ -27,6 +27,7 @@ export const FeatureGates = {
   [StatsigFeatureGates.SHOW_NFT_REWARD]: false,
   [StatsigFeatureGates.SHOW_JUMPSTART_SEND]: false,
   [StatsigFeatureGates.SHOW_POINTS]: false,
+  [StatsigFeatureGates.SHOW_STABLECOIN_EARN]: false,
```
silasbw, line 30: nit on naming: "SHOW_EARN" instead?

jh2oman: we talked and debated between the two. show_earn is a bit short, but as long as we continue calling it Earn we should be fine.

satish-ravi: I think if we anticipate this to be long lived token, we can change to `show_earn`. But almost all of our features have been specific to one release, which is why I went with `show_stablecoin_earn`. So unless we have a case for this being a long lived token (e.g., if we want to have some region lock for this feature), I lean towards `show_stablecoin_earn`.

PR #5375, opened 2024-05-01
test(e2e): update detox to 20.18.5

MuckT:
### Description

Upgrades Detox to [20.18.5](https://github.com/wix/Detox/releases/tag/20.18.5) from [20.11.3](https://github.com/wix/Detox/releases/tag/20.11.3) - [Changes](https://github.com/wix/Detox/compare/20.11.3...20.18.5).
- Sets `com.wix:detox-legacy:+` to continue using Java 11. 
- Changed `launchDeepLink` to take an object instead of a series of params.
- Remove the`quote` e2e util function which appears to have been causing the failures listed below when upgrading.

#### Updates to Detox version before changes
```
 FAIL  e2e/src/DeepLinkSend.spec.js (74.126 s)
  Given
    Deep Link Send
      ○ skipped :android: Send url while app is in background, back pressed
      When Launching Deeplink - App Closed
        ✕ Then should handle deeplink with all attributes (13686 ms)
        ✕ Then should handle deeplink without amount (14343 ms)
        ✓ Then should error if no address provided (3489 ms)
```

#### Updates to Detox version after changes
```
 PASS  e2e/src/DeepLinkSend.spec.js (94.196 s)
  Given
    Deep Link Send
      ○ skipped :android: Send url while app is in background, back pressed
      When Launching Deeplink - App Closed
        ✓ Then should handle deeplink with all attributes (20126 ms)
        ✓ Then should handle deeplink without amount (24880 ms)
        ✓ Then should error if no address provided (3944 ms)
```

### Test plan

- Tested locally on iOS
- Tested locally on Android
- Tested in CI.

### Related issues

- N/A

### Backwards compatibility

Yes

### Network scalability

N/A


android/app/build.gradle
```diff
@@ -266,7 +266,7 @@ android {
 
 
 dependencies {
-    androidTestImplementation('com.wix:detox:+') {
+    androidTestImplementation('com.wix:detox-legacy:+') {
```
satish-ravi, line 269: I guess you have this in the PR desc, but can we also leave a comment here on why we use `legacy`?

satish-ravi (approved):

android/app/build.gradle
MuckT: Added in aeeca981b7fdd864cda811e04dce48aebf0b46d6!

PR #5376, opened 2024-05-01
feat(earn): Add Crypto bottom sheet

finnian0826:
### Description

Creates new component for the Earn "Add Crypto" bottom sheet. This component should be used in the same way as the TokenDetailsMoreActions component [is used](https://github.com/valora-inc/wallet/blob/6df3859e7271db5e643aba11eda8cde19151135d/src/tokens/TokenDetails.tsx#L119).

Decided to make a new component as to avoid adding extra props to decide copy, choose actions, and add an amount prop which would not be used in the TokenDetailsMoreActions case.

Talked with product and Nitya and confirmed that:
- On the QR code page, have title be "Deposit Crypto" like in the cash-in exchange case
- Use "Arbitrum One" as the network name in copy

Note: Price is not passed to the swap component, there is not support for that and based on discussion in quick sync it seems hard.

### Test plan

Unit tests added.

**Manual testing:**
Bottom sheet:
bottom-sheet

Tapping Buy:

https://github.com/valora-inc/wallet/assets/140328381/2a8985ca-9ae2-41e1-b0f7-abed4071b723

Tapping Transfer:

https://github.com/valora-inc/wallet/assets/140328381/61e38183-d641-43c8-9d2d-39af4236215d

Tapping Swap:

https://github.com/valora-inc/wallet/assets/140328381/393d00a6-41d7-49fc-a0f7-162e1ed15e71

Tapping 

### Related issues

- Fixes #ACT-1177

### Backwards compatibility

Yes, new component so not changing past stuff

### Network scalability


locales/base/translation.json
```diff
@@ -2312,5 +2312,21 @@
         "title": "More coming soon!"
       }
     }
+  },
+  "earnFlow": {
```
satish-ravi, line 2316: this (and a few other files) conflicts with main. I used `earnStablecoin` as the key, but I like `earnFlow` and updated it here https://github.com/valora-inc/wallet/pull/5394/files#diff-d3d67f4b3f8dbec345426f534268e40d9b087dfa5d83750165d63caf136cb24cR2329

locales/base/translation.json
```diff
@@ -2312,5 +2312,21 @@
...
+    "addCrypto": {
```
satish-ravi, line 2317: nit: to be more explicit
```suggestion
    "addCryptoBottomSheet": {
```

src/analytics/Events.tsx
```diff
@@ -673,3 +673,7 @@ export enum PointsEvents {
   points_screen_activity_try_again_press = 'points_screen_activity_try_again_press',
   points_screen_activity_fetch_more = 'points_screen_activity_fetch_more',
 }
+
+export enum EarnEvents {
+  earn_tap_add_crypto = 'earn_tap_add_crypto',
```
satish-ravi, line 678: nit: maybe `earn_tap_add_crypto_action` to be more explicit?

src/earn/EarnAddCrypto.tsx
```diff
@@ -0,0 +1,162 @@
...
+        tokenSymbol: token.symbol,
+        tokenNetwork: NETWORK_NAMES[token.networkId],
+      }),
+      iconComponent: QuickActionsSwap,
+      onPress: () => {
+        navigate(Screens.SwapScreenWithBack, { toTokenId: token.tokenId })
+      },
+      visible:
+        isSwapEnabled &&
+        !!swappableFromTokens.find((tokenInfo) => tokenInfo.networkId === token.networkId),
```
satish-ravi, line 136: this should compare tokenId's right?

src/tokens/types.ts
```diff
@@ -1,15 +1,16 @@
 import Colors from 'src/styles/colors'
 
-export enum TokenDetailsActionName {
+export enum TokenActionName {
   Send = 'Send',
   Swap = 'Swap',
   Add = 'Add',
   Withdraw = 'Withdraw',
   More = 'More',
+  Receive = 'Receive',
```
satish-ravi, line 9: should this just be `Transfer` everywhere to match with the name in the designs?

src/earn/EarnAddCrypto.tsx
```diff
@@ -0,0 +1,162 @@
...
+  },
+  actionTitle: {
+    ...typeScale.labelMedium,
+  },
+  actionDetails: {
+    ...typeScale.bodySmall,
+  },
+  title: {
+    ...typeScale.labelLarge,
+  },
```
satish-ravi, line 154: add text color

satish-ravi:
Also, some styling looks off compared to figma (title, gap between description and first action)

src/earn/EarnAddCrypto.test.tsx
```diff
@@ -0,0 +1,132 @@
...
+import { createMockStore } from 'test/utils'
+
+jest.mock('src/statsig', () => ({
+  getDynamicConfigParams: jest.fn(() => {
+    return {
+      showCico: ['arbitrum-sepolia'],
+      showSwap: ['arbitrum-sepolia'],
+    }
+  }),
+  getFeatureGate: jest.fn().mockReturnValue(true),
```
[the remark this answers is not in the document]
finnian0826: no feature gates are required to be true, but some subcomponent uses feature gates (without this line the tests fail), so I changed it to set them false

src/earn/EarnAddCrypto.tsx
finnian0826: I don't think so, swap should be enabled if we have a token that we can swap into USDC on arbitrum, so we want to check if there is a swappable token on the Arbitrum network, so I am comparing network IDs.

I guess I should also check that there is a match where the token ID is not USDC since we won't swap that for itself.

src/earn/EarnAddCryptoBottomSheet.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+      },
+    },
+  },
+  app: {
+    showSwapMenuInDrawerMenu: true,
+  },
+})
+
+describe('EarnAddCryptoBottomSheet', () => {
+  it('Renders correct actions', () => {
```
satish-ravi, line 73: can we also add a test that asserts the not visible case for actions?

src/analytics/Events.tsx
```diff
@@ -677,4 +677,5 @@ export enum PointsEvents {
 
 export enum EarnEvents {
   earn_cta_press = 'earn_cta_press',
+  earn_tap_add_crypto_action = 'earn_tap_add_crypto_action',
```
satish-ravi, line 680:
```suggestion
  earn_add_crypto_action_press = 'earn_add_crypto_action_press',
```
to be consistent with existing earn event

src/earn/EarnAddCryptoBottomSheet.tsx
```diff
@@ -0,0 +1,164 @@
...
+    {
+      name: TokenActionName.Transfer,
+      title: t('earnFlow.addCryptoBottomSheet.actions.transfer'),
+      details: t('earnFlow.addCryptoBottomSheet.actionDescriptions.transfer', {
+        tokenSymbol: token.symbol,
+        tokenNetwork: NETWORK_NAMES[token.networkId],
+      }),
+      iconComponent: QuickActionsSend,
+      onPress: () => {
+        navigate(Screens.ExchangeQR, { flow: CICOFlow.CashIn, exchanges: [] })
```
satish-ravi, line 80: are exchanges supposed to be empty for the MVP?

satish-ravi (approved):

src/earn/EarnAddCryptoBottomSheet.tsx
finnian0826: adding in a follow-up PR

PR #5378, opened 2024-05-02
fix(points): add required request headers for trackPointsEvent call

kathaypacific:
### Description

The missing headers were causing these calls to always fail.

### Test plan

No more errors for /trackPointsEvent

### Related issues

n/a

### Backwards compatibility

Y

### Network scalability


bakoushin (approved):
✨

PR #5381, opened 2024-05-02
feat(earn): add cta to discover tab

satish-ravi:
### Description

Add CTA component and add to discover tab behind feature gate. 
[Figma](https://www.figma.com/file/E1rC3MG74qEg5V4tvbeUnU/Stablecoin-Enablement?type=design&node-id=2472-13347&mode=design&t=oKYLUKxeE593WxAR-0)

### Related issues

- Part of ACT-1174

### Backwards compatibility

Yes

### Network scalability


src/icons/EarnAave.tsx
```diff
@@ -0,0 +1,43 @@
+import * as React from 'react'
+import Svg, { ClipPath, Defs, G, LinearGradient, Path, Rect, Stop } from 'react-native-svg'
+
+const EarnAave = () => (
+  <Svg width={34} height={34} viewBox="0 0 34 34" fill="none">
+    <Rect width={32} height={32} rx={16} fill="#F1FDF1" />
```
MuckT, line 6: It looks like this is also `Colors.successLight` should we pass this color instead of the hardcoded value?

MuckT (approved):

PR #5385, opened 2024-05-03
fix(points): prevent sagas spawning if points are disabled

kathaypacific:
### Description

As the title. Rather than remembering to check the remote config in each saga, catch it all at the root level. We should not start talking to the points DB yet.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y

### Network scalability


bakoushin (approved):
✨

PR #5391, opened 2024-05-06
feat(points): fetch points balance

kathaypacific:
### Description

This PR adds:
- points balance redux variable + migrations
- points balance saga to fetch balance from endpoint
- some coordination so that the balance only animates when it has changed (the screen state can change without the balance changing, which triggers the animation without extra intervention)
- add the heart on points bottom sheet

The error handling will come in the next PR.

### Test plan

Refreshing when the points balance changes:

https://github.com/valora-inc/wallet/assets/20150449/d9675acf-4461-4794-be5f-e9e768cd2c1c

Refreshing when the points balance did not change:

https://github.com/valora-inc/wallet/assets/20150449/261f36f5-acdb-48a8-b070-59edc5459ebd

### Related issues

- Related to RET-1073

### Backwards compatibility

Y

### Network scalability


src/points/PointsHistoryBottomSheet.tsx
```diff
@@ -64,6 +67,18 @@ function PointsHistoryBottomSheet({ forwardedRef }: Props) {
...
+    if (
+      !hasNextPage ||
+      pointsHistoryStatus === 'loadingNextPage' ||
+      pointsHistoryStatus === 'loadingFirstPage'
+    ) {
+      // prevent fetching more history if there is no next page. onEndReached
+      // will continue to fire if this is not checked. onEndReached also does
+      // not have a throttle so prevent fetching more history when there is
+      // already a request in flight.
+      return
```
kathaypacific, line 79: specifically on my simulator, there were not enough total history items to stop the onEndReached from firing continuously in the background

test/schemas.ts
```diff
@@ -2303,14 +2302,7 @@ export const v131Schema = {
...
     ),
-    activeDapp: v130Schema.dapps.activeDapp
-      ? {
-          ...v130Schema.dapps.activeDapp,
-          categories: v130Schema.dapps.activeDapp.categories ?? [
-            v130Schema.dapps.activeDapp.categoryId,
-          ],
-        }
-      : null,
+    activeDapp: null,
```
kathaypacific, line 2305: i had typescript errors trying to define the new schema, it kept complaining that `points` did not exist on the v211Schema. i'm sure that the issue is not with how i defined the new v212Schema, but i'm not sure exactly where in this waterfall object the type definitions became messed up. i tried to remove the `any` casts further up the waterfall which seems to have done the trick. the migration here for `activeDapp` is actually not needed because this should be a session level state anyway.

src/redux/migrations.test.ts
```diff
@@ -1584,7 +1584,7 @@ describe('Redux persist migrations', () => {
     const oldSchema = {
       ...v203Schema,
       positions: {
-        ...v203Schema.positions,
+        ...(v203Schema.positions as any),
```
kathaypacific, line 1587: these `any` casts became necessary after the changes in the schemas.ts file...we create schemas to migrate from old data types that we don't keep anymore so casting the old data types as `any` seems to make sense

src/points/saga.ts
```diff
@@ -87,6 +90,7 @@ export function* getHistory({ payload: params }: ReturnType<typeof getHistorySta
         appendHistory: params.getNextPage,
         newPointsHistory: history.data.filter((record) => isClaimActivityId(record.activityId)),
         nextPageUrl: history.hasNextPage ? history.nextPageUrl : null,
+        pointsBalance: history.balance,
```
bakoushin, line 93: i realized that we may have a small discrepancy when making the history endpoint returning the balance: when fetching the next history page we may update the balance, but the history feed will remain unchanged until the user pulls to refresh. so we will lose the matching between balance and history.

perhaps we should update the balance only when we are fetching the first page?

bakoushin (approved):
💫 Looking good!

src/points/saga.ts
jophish: what do you mean by "so we will lose the matching between balance and history"? like the sum of the history entries that are rendered on the bottom sheet will not add to the number shown on the home screen? i agree that might be a little strange, especially if the user e.g., redeems all their points (so the home screen shows 0 but the history screen shows positive points) or something like that.

bakoushin: @jophish i mean that if the balance will suddenly update, the top of the history will not:
* from the start, we get the balance and the first page (we see the last balance and the last events)
* but then the user opens the bottom sheet and scrolls down
* we get the _updated balance_ and the next page (but not the page with those events that updated that balance)
* the user notices that the balance is updated
* scrolls to the top of the bottom sheet to see the new entries, but there are no new entries because they have to pull to refresh

this is why I'm proposing to update the balance only in sync with fetching the first page, so the user will see their 0 balance updated only when they pull to refresh

src/points/PointsHome.test.tsx
```diff
@@ -77,6 +77,25 @@ describe(PointsHome, () => {
     expect(store.getActions()).toEqual([getPointsConfigRetry()])
   })
 
+  it('refreshes the balance and history on mount and on pull to refresh', async () => {
```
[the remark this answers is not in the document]
kathaypacific: the pull to refresh can only be triggered manually with a swipe down action so it cannot be triggered on load next history page

src/points/slice.ts
```diff
@@ -46,23 +48,25 @@ const initialState: State = {
...
+      getHistoryStatus: action.payload.getNextPage ? 'loadingNextPage' : 'loadingFirstPage',
     }),
     getHistorySucceeded: (state, action: PayloadAction<GetPointsHistorySucceededAction>) => ({
       ...state,
       pointsHistory: action.payload.appendHistory
         ? [...state.pointsHistory, ...action.payload.newPointsHistory]
         : action.payload.newPointsHistory,
       nextPageUrl: action.payload.nextPageUrl,
-      getHistoryStatus: 'idle',
+      getHistoryStatus: 'success',
```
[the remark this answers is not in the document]
kathaypacific: yes essentially to differentiate between when the balance is up to date or in need of a fetch. the status is now reset to idle on app start :)

PR #5392, opened 2024-05-06
chore(points): add error handling for points balance fetch

kathaypacific:
### Description

This PR adds a warning toast when there is an error fetching the history / balance on the points home screen.

I made some changes to the history error handling so that we have consistent approaches between the 2 screens:
1. don't use the error state to erase data in redux - i think this is an unexpected pattern, usually we'd want to keep the existing data in place and show an error
2. use the Toast component so that we have nice animations

### Test plan

Error on fetch first page:

https://github.com/valora-inc/wallet/assets/20150449/6ff4675c-7127-45b8-9a64-10e60f8abef5

Error on fetch next pages (I reduced the size of the bottom sheet because of the small amount of test data available):

https://github.com/valora-inc/wallet/assets/20150449/4a3dbe37-59ac-412f-a8e2-f7ec45c94021

### Related issues

- Related to RET-1073

### Backwards compatibility

Y

### Network scalability


bakoushin (approved):
🎉

jophish (approved):
:tada:

PR #5393, opened 2024-05-07
feat: add earn active pool component

MuckT:
### Description

Creates an `EarnActivePool.tsx` component to conditionally render in place of the CTA when the user has a balance on the pool.

### Test plan

- [ ] Tested locally on iOS
- [ ] Tested locally on Android
- [ ] Unit tests updated

### Related issues

- Fixes ACT-1179

### Backwards compatibility

Yes

### Network scalability

Yes this will continue to work on Arbitrum as new networks are added; however, it would be nice to abstract some of the logic in `getAavePoolUserBalance` to handle Aave pools on other networks.


MuckT:
Closed in favor of #5397 & #5400.

PR #5397, opened 2024-05-07
chore: add aaveArbUsdcTokenId to network config

MuckT:
### Description

Adds aaveArbUsdcTokenId to network config.
### Test plan

- N/A

### Related issues

- ACT-1179

### Backwards compatibility

Yes

### Network scalability

N/A


src/earn/poolInfo.ts
```diff
@@ -32,3 +36,47 @@ export async function fetchAavePoolInfo(assetAddress: Address) {
...
+/** @beta - Exclude from Knip dep check */
+export async function fetchAavePoolUserBalance({
+  assetAddress,
+  walletAddress,
+}: {
+  assetAddress: Address
+  walletAddress: Address
+}) {
+  try {
+    if (!assetAddress) {
```
silasbw, line 49: would we expect this to happen with a `viem` `Address`?

src/earn/poolInfo.ts
```diff
@@ -32,3 +36,47 @@ export async function fetchAavePoolInfo(assetAddress: Address) {
...
+      throw new Error('No asset address provided')
+    }
+    if (!walletAddress) {
```
silasbw, line 52: same here

src/earn/poolInfo.ts
```diff
@@ -32,3 +36,47 @@ export async function fetchAavePoolInfo(assetAddress: Address) {
...
+    }
+    Logger.debug(TAG, 'Fetching Aave pool user balance', { assetAddress, walletAddress })
+    const result = await publicClient[Network.Arbitrum].readContract({
+      abi: AavePool,
+      address: networkConfig.arbAavePoolV3ContractAddress,
+      functionName: 'getReserveData',
+      args: [assetAddress],
+    })
+
+    const { aTokenAddress } = result as { aTokenAddress: `0x${string}` }
```
silasbw, line 63: use `Address` instead of `0x${string}`?

src/earn/poolInfo.ts
```diff
@@ -32,3 +36,47 @@ export async function fetchAavePoolInfo(assetAddress: Address) {
...
+}) {
+  try {
+    if (!assetAddress) {
+      throw new Error('No asset address provided')
+    }
+    if (!walletAddress) {
+      throw new Error('No wallet address provided')
+    }
+    Logger.debug(TAG, 'Fetching Aave pool user balance', { assetAddress, walletAddress })
+    const result = await publicClient[Network.Arbitrum].readContract({
```
silasbw, line 56: Dumb q: is this the approach aave recommends instead of saving the ERC-20 address (or hard coding it)?

src/earn/poolInfo.ts
MuckT: It was mainly so the usage of hooks can handle possible undefined values for example: `const asyncPoolInfo = useAavePoolInfo(token?.address as Address)`. Since hooks are coming in a later PR this can be removed, but will likely be needed when hooks are added.

src/earn/poolInfo.ts
MuckT: Fixed in 5bc1b87684d4f940a27421f3842738d9ec374230!

src/earn/poolInfo.ts
MuckT: This was the approach used in `fetchAavePoolInfo` and the example shared by @jh2oman in the example code I am not sure why it is preferred.

src/earn/poolInfo.ts
satish-ravi: The type should reflect this if it can be potentially undefined. I don't think we should be doing `token?.address as Address` anywhere. It can lead to nasty bugs down the line, if the function doesn't handle it being potentially nullable

src/earn/poolInfo.ts
silasbw: the downside is that, possibly, the pool contract would return a different address at some point :shrug: 

we already save the aTokenAddress is address-metadata too, so it's unexpected to fetch it here, but in some parts of the experience rely on the address in address-metadata.

src/earn/poolInfo.ts
silasbw: seems worth removing since it's confusing given the function signature/typing

src/earn/poolInfo.ts
```diff
@@ -32,3 +36,47 @@ export async function fetchAavePoolInfo(assetAddress: Address) {
     throw err
   }
 }
+
+/** @beta - Exclude from Knip dep check */
```
satish-ravi, line 40: is this required?

src/earn/poolInfo.ts
MuckT: I was going to do something like this to get the pool info in the component:

```TypeScript
  const token = useTokenInfo(networkConfig.arbUsdcTokenId)
  const asyncPoolInfo = useAavePoolInfo(token?.address as Address)
  const asyncPoolUserInfo = useAavePoolUserBalance({
    assetAddress: token?.address as Address,
    walletAddress: walletAddress as Address
  })
```

src/earn/poolInfo.ts
MuckT: Removed in 3c133d6c3651dd08098da0fccb348aec1821ed50!

src/earn/poolInfo.ts
MuckT: @silasbw It looks like aTokenAddress resolves to `0x724dc807b04555b71ed48a6896b6f41593b8c637` on mainnet which is  [here](https://github.com/valora-inc/address-metadata/blob/main/src/data/mainnet/arbitrum-one-tokens-info.json#L43) in address-metadata. Are you proposing we filter tokens using a hardcoded tokenId or symbol?

silasbw: I'm proposing we have one source of truth for the `aTokenAddress`, yeah.  Filtering by tokenId could work.  How are we going to associate the aave-specific logic with the aArbUSDCn token details screen?

MuckT: At this point I am thinking that a network config variable is probably simplest either `aArbUSDCnTokenId` or `aArbUSDCnAddress`.

MuckT: Added the tokenId to the networkConfig in 9873c6b7f022df5966679eb368d6fb222a6a7f0d and adjusted the fetchAavePoolUserBalance in f79cee8371a1922e4d5d08ff7103aba66f86b4cb.

src/earn/poolInfo.ts
```diff
@@ -32,3 +33,32 @@ export async function fetchAavePoolInfo(assetAddress: Address) {
...
+    const aaveUSDCContract = getContract({
+      abi: erc20.abi,
+      address: tokenAddress,
+      client: {
+        public: publicClient[Network.Arbitrum],
+      },
+    })
+
+    const balance = await aaveUSDCContract.read.balanceOf([walletAddress])
+    const decimals = await aaveUSDCContract.read.decimals()
```
satish-ravi, line 55: the arb usdc token is part of address metadata right? If we have the tokenId can't we just get this by doing `useTokenId` hook instead of making another contract call here?

satish-ravi (approved):

silasbw (approved):

PR #5398, opened 2024-05-07
refactor(enter-amount): refactor EnterAmount to prepare for it to be used with earn

jh2oman:
### Description

* Allow for a custom CTA to be passed in, which will be used with Earn
* New prop `disableBalanceCheck` to allow users to enter in more crypto than they actually have, and allows send to be pressed regardless of the amount relative to the balance
* If the user has no balance for the given token, hide the max button

#### Example with disableBalanceCheck=true

https://github.com/valora-inc/wallet/assets/8432644/58d34b2c-75d8-4b41-9950-ee0ab07ca4af

### Test plan

Added unit tests

### Related issues

https://linear.app/valora/issue/ACT-1175/wallet-earn-cryptofiat-input-box-component

### Backwards compatibility

Backwards compatible to work with existing send / jumpstart flows


src/send/EnterAmount.tsx
```diff
@@ -67,12 +73,38 @@ interface Props {
   onPressProceed(args: ProceedArgs): void
   disableProceed?: boolean
   children?: React.ReactNode
+  CTA: ComponentType<CTAProps>
```
satish-ravi, line 76: naming nit: maybe match with proceed to be consistent with other args? e.g., `ProceedComponent` ? Or rename other args to use Cta to be consistent.

Also, this needs to start with an upper case letter to render as a jsx element right?

src/send/EnterAmount.tsx
```diff
@@ -244,17 +279,22 @@ function EnterAmount({
...
+  const transactionIsPossible =
     !showLowerAmountError &&
     prepareTransactionsResult &&
     prepareTransactionsResult.type === 'possible' &&
     prepareTransactionsResult.transactions.length > 0
 
+  const disabled =
+    disableProceed ||
+    (disableBalanceCheck && tokenAmount?.isZero()) ||
+    (!disableBalanceCheck && !transactionIsPossible)
```
satish-ravi, line 291: maybe this is slightly more readable? don't feel strongly though
```suggestion
    (disableBalanceCheck ? tokenAmount?.isZero() : !transactionIsPossible)
```

satish-ravi (approved):
Looks good, approved once prop name is updated to be consistent with other args. Also, rename PR title from chore to refactor :)

PR #5399, opened 2024-05-07
feat(earn): Add enter amount screen

finnian0826:
### Description

Add EarnEnterAmount screen.

### Test plan
Unit tests added.

**Manual tests:**

When a user has USDC on arbitrum:

https://github.com/valora-inc/wallet/assets/140328381/5103c167-3447-4df5-bb86-2318bf2c021a

When a user doesn't have enough USDC on arbitrum:

https://github.com/valora-inc/wallet/assets/140328381/459a6671-e8bc-42ca-84ef-e8c6d104a427

Clicking the info button:

https://github.com/valora-inc/wallet/assets/140328381/86c47948-61c2-4c1b-89b9-b02be943f534

### Related issues

- Fixes #ACT-1176

### Backwards compatibility

Yes

### Network scalability


src/send/EnterAmount.tsx
```diff
@@ -286,7 +299,10 @@ function EnterAmount({
     prepareTransactionsResult.transactions.length > 0
 
   const disabled =
-    disableProceed || (disableBalanceCheck ? !!tokenAmount?.isZero() : !transactionIsPossible)
+    disableProceed ||
+    (disableBalanceCheck
+      ? !!tokenAmount?.isZero() || (!!tokenAmount?.lte(token.balance) && !transactionIsPossible)
```
finnian0826, line 304: We want to disable the continue button when checking if the prepared transaction is possible, even if we are disabling the warnings.

src/send/EnterAmount.tsx
```diff
@@ -441,14 +465,13 @@ function EnterAmount({
           />
         )}
 
-        {children}
```
finnian0826, line 444: With children here, the bottom sheets were acting super weird. Need to go through and double check that this doesn't mess anything else up.

locales/base/translation.json
```diff
@@ -2351,6 +2351,19 @@
...
+    "enterAmount": {
+      "earnUpToLabel": "You could earn up to:",
+      "rateLabel": "Rate (est.)",
+      "earnUpTo": "~{{fiatSymbol}}{{amount}} / yr",
+      "rate": "{{rate}}% APY",
+      "continue": "Continue",
+      "info": "This pool is powered by Aave",
+      "infoBottomSheet": {
+        "title": "Why this pool?",
+        "description": "This Aave pool has over $150M TVL and more than 6,000 contributors, indicating strong liquidity and trustworthiness. It also uses USDC as its token and Abritrum as its network, which reduces volatility and ensures users will receive inexpensive gas rates.\n\nAll together this makes it a safe and attractive option with competitive APY, ensuring a beneficial investment opportunity for our users.\n\nYou can explore other Aave pools here.",
```
finnian0826, line 2363: The "here" is supposed to be a link, but I'm not sure what it is supposed to link to. Also, this is used as the `description` param for a bottom sheet so I'm not sure how to format links for translations when it is passed as a param.

src/earn/EarnEnterAmount.tsx
```diff
@@ -0,0 +1,293 @@
...
+      <View style={styles.line}>
+        <Text style={styles.valuesText} testID="EarnEnterAmount/EarnUpTo">
+          {t('earnFlow.enterAmount.earnUpTo', {
+            fiatSymbol: localCurrencySymbol,
+            amount:
+              asyncPoolInfo?.result && !!asyncPoolInfo.result.apy && tokenAmount?.gt(0)
+                ? tokenAmount.multipliedBy(new BigNumber(asyncPoolInfo.result.apy)).toFormat(2)
+                : '--',
+          })}
+        </Text>
```
satish-ravi, line 180: I guess the design linked in the ticket shows TVL here, is this from the follow up? Just confirming, should be fine to leave this as is

src/send/EnterAmount.tsx
```diff
@@ -150,10 +154,13 @@ function EnterAmount({
   prepareTransactionError,
   tokenSelectionDisabled = false,
   onPressProceed,
+  onPressInfo,
+  onSetTokenAmount,
```
satish-ravi, line 158:
```suggestion
  onChangeTokenAmount,
```

src/send/EnterAmount.tsx
```diff
@@ -71,10 +72,13 @@ interface Props {
   prepareTransactionError?: Error
   tokenSelectionDisabled?: boolean
   onPressProceed(args: ProceedArgs): void
+  onPressInfo?(): void
```
satish-ravi, line 75: is `onPressInfo` needed in EnterAmount? Seems like it just gets it as a prop and passes in back to proceed component, which can all be done within EarnEnterAmount

satish-ravi (commented):
some padding looks off

src/earn/EarnEnterAmount.tsx
finnian0826: yeah talking with Jacob we decided to just show the earn up to amount here, as it seemed easier than getting TVL which would then just be removed

src/send/EnterAmount.tsx
finnian0826: no, I'll update it to just be within the proceed component

src/send/EnterAmount.tsx
finnian0826: I think the `!!tokenAmount?.isZero()` can be removed, because in that case `!!tokenAmount?.lte(token.balance) && !transactionIsPossible` should be true because 0 should be lte to the token balance and the transaction should not be possible.

locales/base/translation.json
finnian0826: Confirmed with John that that link is good so I'll add it in

src/send/EnterAmount.tsx
finnian0826: actually, I think to do this I would have to have the EarnProceed function inside EarnEnterAmount, and this causes issues with the token amount state value and also rendering APY.

So I think EarnProceed (which needs access to onPressInfo) needs to be outside of EarnEnterAmount so I'm not sure then how it would have access to onPressInfo. @satish-ravi is there some way to access it from outside the function it is declared without passing as a prop?

src/send/EnterAmount.tsx
```diff
@@ -427,7 +454,10 @@ function EnterAmount({
             description={t('sendEnterAmountScreen.notEnoughBalanceForGasWarning.description', {
               feeTokenSymbol: prepareTransactionsResult.feeCurrencies[0].symbol,
             })}
-            style={styles.warning}
+            style={[
+              styles.warning,
+              hideNetworkFee && { marginTop: Spacing.Regular16, marginBottom: 0 },
```
finnian0826, line 459: This is needed to make spacing correct. Usually there is spacing below the network fee but when that is disabled, the top spacing needs to be here. Similarly, there shouldn't be bottom spacing in the earn case because the EarnProceed has top spacing to be in the correct spot below the enter amount component.

src/earn/EarnEnterAmount.tsx
```diff
@@ -0,0 +1,628 @@
...
+            testID="SendEnterAmount/PrepareTransactionError"
+          />
+        )}
+        <EarnProceed
+          tokenAmount={tokenAmount}
+          localAmount={localAmount}
+          token={token}
+          amountEnteredIn={enteredIn}
+          onPressProceed={onPressContinue}
+          onPressInfo={onPressInfo}
```
finnian0826, line 349: Still need to pass this here. I tried having EarnProceed be a const inside of EarnEnterAmount rather than a function outside of it, but this was causing it to re-render everytime the input amount was changed and so the APY was flashing a lot.

finnian0826:
Follow up issue to use EnterAmount: https://linear.app/valora/issue/ACT-1199/placeholder-clean-up-have-earnenteramount-use-enteramount-component

src/earn/prepareTransactions.ts
```diff
@@ -134,6 +138,24 @@ export async function prepareSupplyTransactions({
   })
 }
 
+/**
+ * Hook to prepare transactions for supplying crypto.
+ */
+export function usePrepareSupplyTransactions() {
+  const prepareTransactions = useAsyncCallback(prepareSupplyTransactions, {
+    onError: (error) => {
+      Logger.error(TAG, `usePrepareSupplyTransactions: ${error}`)
```
satish-ravi, line 147:
```suggestion
      Logger.error(TAG, 'usePrepareSupplyTransactions', error)
```
so the error is logged with trace, you may need an ensureError potentially

src/earn/EarnEnterAmount.tsx
```diff
@@ -0,0 +1,626 @@
...
+    prepareTransactionsResult &&
+    prepareTransactionsResult.type === 'not-enough-balance-for-gas'
+  const transactionIsPossible =
+    isAmountLessThanBalance &&
+    prepareTransactionsResult &&
+    prepareTransactionsResult.type === 'possible' &&
+    prepareTransactionsResult.transactions.length > 0
+
+  const disabled =
+    !!tokenAmount?.isZero() || (!!tokenAmount?.lte(token.balance) && !transactionIsPossible) // Should disable if the user enters 0 or has enough balance but the transaction is not possible, shouldn't disable if they enter an amount larger than their balance as they will go to add flow
```
satish-ravi, line 198: nit: this won't be readable in the editor, move the comment to one line above and break it up to multiple lines. would be nice if prettier can do this.

src/earn/EarnEnterAmount.tsx
```diff
@@ -0,0 +1,626 @@
...
+  const localCurrencySymbol = useSelector(getLocalCurrencySymbol) ?? LocalCurrencySymbol.USD
+
+  const {
+    prepareTransactionsResult,
+    refreshPreparedTransactions,
+    clearPreparedTransactions,
+    prepareTransactionError,
+  } = usePrepareSupplyTransactions()
+
+  const walletAddress = useSelector(walletAddressSelector) as Address
```
satish-ravi, line 96: can we avoid the as Address, I think we've been doing a `isAddress` check where required

src/analytics/Properties.tsx
```diff
@@ -1599,6 +1599,16 @@ interface EarnEventsProperties {
...
   [EarnEvents.earn_deposit_submit_cancel]: EarnDepositProperties
   [EarnEvents.earn_view_pools_press]: undefined
+  [EarnEvents.earn_enter_amount_info_press]: undefined
+  [EarnEvents.earn_enter_amount_continue_press]: {
+    userHasFunds: boolean
+    tokenAmount: string | null
+    amountInUsd: string | null
+    amountEnteredIn: AmountEnteredIn
+    tokenId: string | null
+    networkId: string | null
```
satish-ravi, line 1609: do these need to be nulls?

src/earn/EarnEnterAmount.tsx
```diff
@@ -0,0 +1,626 @@
...
+    ValoraAnalytics.track(SendEvents.max_pressed, {
+      tokenId: token.tokenId,
+      tokenAddress: token.address,
+      networkId: token.networkId,
+    })
+  }
+
+  const onPressContinue = ({ tokenAmount, token, amountEnteredIn }: ProceedArgs) => {
+    ValoraAnalytics.track(EarnEvents.earn_enter_amount_continue_press, {
+      userHasFunds: token.balance?.gte(tokenAmount),
```
satish-ravi, line 255: reuse isAmountLessThanBalance

src/earn/EarnEnterAmount.tsx
```diff
@@ -0,0 +1,626 @@
...
+      tokenAmount: tokenAmount.toString(),
+      amountInUsd: tokenAmount.multipliedBy(token.priceUsd ?? 0).toFixed(2),
+      amountEnteredIn,
+      tokenId: token.tokenId,
+      networkId: token.networkId,
+    })
+    tokenAmount?.gt(token.balance)
```
satish-ravi, line 262: same as above

src/earn/EarnEnterAmount.tsx
```diff
@@ -0,0 +1,626 @@
...
+        </View>
+        {showNotEnoughBalanceForGasWarning && (
+          <InLineNotification
+            variant={NotificationVariant.Warning}
+            title={t('sendEnterAmountScreen.notEnoughBalanceForGasWarning.title', {
+              feeTokenSymbol: prepareTransactionsResult.feeCurrencies[0].symbol,
+            })}
+            description={t('sendEnterAmountScreen.notEnoughBalanceForGasWarning.description', {
+              feeTokenSymbol: prepareTransactionsResult.feeCurrencies[0].symbol,
+            })}
```
satish-ravi, line 327: I guess this would just say "ETH" instead of "ETH on Arbitrum" which could be confusing, maybe we can have a follow up that adds network name to this

src/earn/EarnEnterAmount.tsx
```diff
@@ -0,0 +1,626 @@
...
+  token,
+  amountEnteredIn,
+  disabled,
+  onPressProceed,
+  onPressInfo,
+}: ProceedComponentProps) {
+  const { t } = useTranslation()
+  const localCurrencySymbol = useSelector(getLocalCurrencySymbol)
+
+  const asyncPoolInfo = useAsync(
```
satish-ravi, line 382: replace with `useAavePoolInfo` hook

src/earn/EarnEnterAmount.test.tsx
```diff
@@ -0,0 +1,208 @@
...
+        networkId: NetworkId['arbitrum-sepolia'],
+        balance: new BigNumber(10),
+      },
+      walletAddress: mockAccount.toLowerCase(),
+      poolContractAddress: networkConfig.arbAavePoolV3ContractAddress,
+      feeCurrencies: mockFeeCurrencies,
+    })
+  })
+
+  it('should handle navigating to the deposit bottom sheet', async () => {
```
satish-ravi, line 146: can this assert the appropriate bottom sheet was opened?

src/earn/EarnEnterAmount.test.tsx
```diff
@@ -0,0 +1,208 @@
...
+        amountEnteredIn: 'token',
+        amountInUsd: '8.00',
+        networkId: NetworkId['arbitrum-sepolia'],
+        tokenAmount: '8',
+        tokenId: networkConfig.arbUsdcTokenId,
+        userHasFunds: true,
+      }
+    )
+  })
+  it('should handle navigating to the add crypto bottom sheet', async () => {
```
satish-ravi, line 177: same comment as above

src/earn/EarnEnterAmount.tsx
```diff
@@ -0,0 +1,626 @@
...
+      <EarnAddCryptoBottomSheet
+        forwardedRef={addCryptoBottomSheetRef}
+        token={token}
+        tokenAmount={tokenAmount ? tokenAmount.minus(token.balance) : new BigNumber(0)}
+      />
+      {prepareTransactionsResult?.type === 'possible' && (
+        <EarnDepositBottomSheet
+          forwardedRef={reviewBottomSheetRef}
+          preparedTransaction={prepareTransactionsResult}
+          amount={tokenAmount ? tokenAmount.toString() : '0'}
```
satish-ravi, line 362: tokenAmount should be defined with tx is possible right? You can just add this as another condition to the conditional rendering check above instead of the ternary here.

src/earn/EarnEnterAmount.tsx
```diff
@@ -0,0 +1,626 @@
...
+  },
+  label: {
+    ...typeScale.bodySmall,
+    color: Colors.gray4,
+  },
+  continueButton: {
+    paddingVertical: Spacing.Thick24,
+  },
+  valuesText: {
+    ...typeScale.labelSemiBoldSmall,
```
satish-ravi, line 532: add color to all text styles

src/earn/EarnEnterAmount.tsx
```diff
@@ -0,0 +1,626 @@
...
+}: {
+  infoBottomSheetRef: React.RefObject<BottomSheetRefType>
+}) {
+  const { t } = useTranslation()
+  const onPressDismiss = () => {
+    infoBottomSheetRef.current?.close()
+  }
+  const onPressMorePools = () => {
+    ValoraAnalytics.track(EarnEvents.earn_enter_amount_info_more_pools)
+    navigate(Screens.WebViewScreen, { uri: 'https://app.aave.com/markets/' })
```
satish-ravi, line 468: can we move this to the earn_stablecoin_config dynamic config?

src/earn/prepareTransactions.ts
```diff
@@ -7,12 +8,15 @@ import { getDynamicConfigParams } from 'src/statsig'
...
 import { StatsigDynamicConfigs } from 'src/statsig/types'
 import { TokenBalance } from 'src/tokens/slice'
+import Logger from 'src/utils/Logger'
 import { fetchWithTimeout } from 'src/utils/fetchWithTimeout'
 import { publicClient } from 'src/viem'
 import { TransactionRequest, prepareTransactions } from 'src/viem/prepareTransactions'
 import networkConfig, { networkIdToNetwork } from 'src/web3/networkConfig'
 import { Address, encodeFunctionData, isAddress, parseUnits } from 'viem'
 
+const TAG = 'src/earn/prepareTransactions'
```
MuckT, line 18:
```suggestion
const TAG = 'earn/prepareTransactions'
```

nit: usually we don't include src in the tags.

src/analytics/Properties.tsx
finnian0826: no, they should all have values, I'll update to not take null

silasbw (approved):

PR #5400, opened 2024-05-07
feat: add earn active pool component

MuckT:
### Description

Creates the `EarnActivePool.tsx` component and associated `useEarn` hook. This hook is then used on the `TabDiscover.tsx` screen to display either `EarnActivePool.tsx` or `EarnCta.tsx`.

#### Video iOS

https://github.com/valora-inc/wallet/assets/26950305/70528be7-f239-4c7b-9e65-a97409520647

### Test plan

- [x] Tested locally on iOS
- [x] Unit tests added

### Related issues

- ACT-1179

### Backwards compatibility

Yes

### Network scalability


src/dappsExplorer/TabDiscover.tsx
```diff
@@ -76,6 +82,14 @@ function TabDiscover({ navigation }: Props) {
   const nonFavoriteDappsWithCategoryNames = useSelector(nonFavoriteDappsWithCategoryNamesSelector)
   const favoriteDappsWithCategoryNames = useSelector(favoriteDappsWithCategoryNamesSelector)
 
+  // Earning Pool Aave
+  const token = useTokenInfo(networkConfig.arbUsdcTokenId)
+  const walletAddress = useSelector(walletAddressSelector)
+  const asyncPoolUserInfo = useAavePoolUserBalance({
```
silasbw, line 88: should this be conditional on `token?.address` being truthy?  The null dereference implies `assetAddress` might end up being `null` (which is not of type `Address`).

src/dappsExplorer/TabDiscover.tsx
```diff
@@ -237,7 +251,13 @@ function TabDiscover({ navigation }: Props) {
                   </Text>
                 }
                 <DappFeaturedActions onPressShowDappRankings={handleShowDappRankings} />
-                {getFeatureGate(StatsigFeatureGates.SHOW_STABLECOIN_EARN) && <EarnCta />}
+                {getFeatureGate(StatsigFeatureGates.SHOW_STABLECOIN_EARN) ? (
```
silasbw, line 254: nit: i think it's clearer to move the business logic of what's enabled above (outside the ts embedded in the tsx).  it took me a sec to parse the nested ternary operators.  you could do:

ts:
```ts
let earnInstance = null
const showStablecoinEarn = getFeatureGate(StatsigFeatureGates.SHOW_STABLECOIN_EARN)
if (showStablecoinEarn) {
  earnInstance = ...
}
```
tsx:
```tsx
{earnInstance}
```

src/earn/EarnActivePool.tsx
```diff
@@ -0,0 +1,150 @@
...
+              />
+            )}
+          </View>
+          <View style={styles.row}>
+            {asyncPoolInfo.loading && <PoolDetailsLoading />}
+            {asyncPoolInfo.result && (
+              <View style={styles.apyContainer}>
+                <Text style={styles.apyText}>
+                  {t('earnFlow.activePools.apy', {
+                    apy: (asyncPoolInfo.result.apy * 100).toFixed(2) ?? '',
```
silasbw, line 67: we've ran into "percentage naming" issues before: where we name something "percentage" (e.g., APY) when the value is in fact a proportion (i.e., the "decimal value") and then end up passing the value to an API or showing to the user or etc when the expected value is a percentage (or vice versa).  Can we get ahead of it here somehow?

I realize the naming I'm commenting on might predate this PR.  How about `result.apyProporion` (don't think to hard about Annual Percentage Yield Proportion :laughing: ) and `apy` (for the actual percentage)

src/dappsExplorer/TabDiscover.tsx
satish-ravi: +1, and we should avoid `as type` as much as possible in production code

src/dappsExplorer/TabDiscover.tsx
MuckT: Moved the conditional rendering to a hook in 532eac80e94c101bd66770e017f5dccd8f412a26. Hopefully this makes it a bit clearer what the business logic is for displaying the earn component.

src/dappsExplorer/TabDiscover.tsx
MuckT: `token.address` returned from `useTokenInfo` is of type string, but the `readContract` method in `fetchAavePoolInfo` expects an `Address` type. We do a lot of `...as Address` with jumpstart so I am not sure if this is avoidable when using smart contract methods.

MuckT: This has since moved to the `EarnActivePool.tsx` component itself so we could do a check and simply return null for that component with `if(!token?.address) return null`.

src/earn/EarnActivePool.tsx
```diff
@@ -0,0 +1,142 @@
...
+        />
+      </SkeletonPlaceholder>
+    </View>
+  )
+}
+
+export default function EarnActivePool() {
+  const { t } = useTranslation()
+  const token = useTokenInfo(networkConfig.arbUsdcTokenId)
+  const asyncPoolInfo = useAavePoolInfo(token?.address as Address)
```
silasbw, line 35: `null` is not an Address.  You could, for example, use `isAddress` https://viem.sh/docs/utilities/isAddress.html which implements a type guard and narrows the type to an `Address`.

```ts
const tokenAddress = token.address
if (!tokenAddress || !isAddress(tokenAddress)) {
  return null
}
// ts knows tokenAddress is an Address now
```

src/dappsExplorer/TabDiscover.tsx
silasbw: Continued here https://github.com/valora-inc/wallet/pull/5400/files#r1594881960

src/earn/EarnActivePool.tsx
MuckT: Fixed in fa05f358887ebb760308ff5fcba477b12c09bbd0!

src/earn/EarnActivePool.tsx
```diff
@@ -0,0 +1,154 @@
...
+    </View>
+  )
+}
+
+export default function EarnActivePool() {
+  const { t } = useTranslation()
+  const token = useTokenInfo(networkConfig.arbUsdcTokenId)
+  const poolToken = useTokenInfo(networkConfig.aaveArbUsdcTokenId)
+  const asyncPoolInfo = useAsync(async () => {
+    if (!token || !token.address) {
```
silasbw, line 41: if you use the viem type guard here for `Address` `|| !isAddress(token.address)` you can avoid the type assertion down below

silasbw (approved):

src/earn/EarnActivePool.tsx
MuckT: Fixed in f5b7c8e7e7335ddfc9005344dfec0dc7727ca36d!

PR #5401, opened 2024-05-08
feat: Change button color to black

finnian0826:
### Description
- Change primary button color to black
- Set opacity at 25% when disabled
- Set opacity of GRAY_WITH_BORDER to 50% when disabled

### Test plan

Updated unit tests, manual tests:
token-details

https://github.com/valora-inc/wallet/assets/140328381/5fdf0e9c-fef7-4c9b-8608-48354a2ca6c2

https://github.com/valora-inc/wallet/assets/140328381/6afe33f8-7751-42c0-8fce-e13192324e3a

https://github.com/valora-inc/wallet/assets/140328381/2dd9d319-cf81-40c4-b104-bbe1544788cc

### Related issues

- N/A

### Backwards compatibility

Yes

### Network scalability


src/points/PointsHome.tsx
```diff
@@ -145,7 +145,6 @@ export default function PointsHome({ route, navigation }: Props) {
                 onPress={onPressActivity}
                 text={t('points.activity')}
                 type={BtnTypes.GRAY_WITH_BORDER}
-                fontStyle={typeScale.labelXSmall}
```
satish-ravi, line 148: I think we may want to leave the font style as an option because of this. cc @kathaypacific

kathaypacific: naw if it's just for points, i think we can remove it. i'm a fan of not having customisable styles, they can be convenient but they add extra complexity and facilitate inconsistency

src/app/MultichainBeta.tsx
MuckT, line 1: Do we still need this file? I suppose it's outside of the scope of the ticket to delete.

MuckT (approved):

PR #5405, opened 2024-05-10
feat(earn): add function for preparing supply transaction

satish-ravi:
### Description

Mostly based off of swap transaction preparation

### Test plan

Unit tests, manually preparing a tx

### Related issues

- Part of ACT-1178

### Backwards compatibility

Yes

### Network scalability


src/earn/prepareTransactions.ts
```diff
@@ -0,0 +1,75 @@
...
+  walletAddress: Address
+  feeCurrencies: TokenBalance[]
+  poolContractAddress: Address
+}) {
+  const baseTransactions: TransactionRequest[] = []
+
+  // amount in smallest unit
+  const amountToSupply = parseUnits(amount, token.decimals)
+
+  if (!token.address) {
```
silasbw, line 28: i've been nagging people to use the viem type guard in places like this:

```ts
if (!token.address || !isAddress(token.address) {
```

src/earn/prepareTransactions.ts
```diff
@@ -0,0 +1,75 @@
...
+    // should never happen
+    throw new Error('Cannot use a token without address')
+  }
+
+  const approvedAllowanceForSpender = await publicClient[
+    networkIdToNetwork[token.networkId]
+  ].readContract({
+    address: token.address as Address,
```
silasbw, line 36: ...then you can drop the type assertions and maybe feel good about that.

silasbw (approved):

src/earn/prepareTransactions.ts
```diff
@@ -0,0 +1,117 @@
...
+  poolContractAddress: Address
+}) {
+  const baseTransactions: TransactionRequest[] = []
+
+  // amount in smallest unit
+  const amountToSupply = parseUnits(amount, token.decimals)
+
+  if (!token.address || !isAddress(token.address)) {
+    // should never happen
+    throw new Error('Cannot use a token without address')
```
MuckT, line 42: Could we add the tokenId to this error to help with debugging if we run into issues?

src/earn/prepareTransactions.ts
```diff
@@ -0,0 +1,117 @@
...
+  const approvedAllowanceForSpender = await publicClient[
+    networkIdToNetwork[token.networkId]
+  ].readContract({
+    address: token.address,
+    abi: erc20.abi,
+    functionName: 'allowance',
+    args: [walletAddress, poolContractAddress],
+  })
+
+  if (approvedAllowanceForSpender < amountToSupply) {
```
MuckT, line 54: Would this work if `approvedAllowanceForSpender === amountToSupply`?

src/earn/prepareTransactions.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  if (!response.ok) {
+    throw new Error(
+      `Failed to simulate transactions. status ${response.status}, text: ${await response.text()}`
+    )
+  }
+
+  // extract fee of the supply transaction and set gas fields
+  const { simulatedTransactions }: SimulatedTransactionResponse = await response.json()
+  const supplySimulatedTx = simulatedTransactions[simulatedTransactions.length - 1]
```
MuckT, line 100: nit: code golf

```suggestion
  const supplySimulatedTx = simulatedTransactions.at(-1)
```

jh2oman (approved):

src/earn/prepareTransactions.ts
satish-ravi: yes

src/earn/prepareTransactions.ts
satish-ravi: done

src/earn/prepareTransactions.ts
satish-ravi: this would need a type assertion, since `at` returns <array type> | undefined, so leaving as is.

MuckT (approved):

PR #5406, opened 2024-05-13
feat(points): points entry point and intro screen

bakoushin:
### Description

* Add points card to the "Discover" screen
* Add an intro screen to the points home page (shown only once)

https://github.com/valora-inc/wallet/assets/2737872/ef769852-a628-4e4c-8ce1-63ffcbb0345b

### Test plan

Updated unit tests

### Related issues

- Fixes RET-1077

### Backwards compatibility

NA

### Network scalability


bakoushin:
@jeanregisser what kind of animation would you expect?

.gitignore
```diff
@@ -133,6 +133,7 @@ src/brandingConfig.ts
 src/icons/LogoHeart.tsx
 src/icons/Logo.tsx
 src/icons/Home.tsx
+src/icons/PointsCoverImage.tsx
```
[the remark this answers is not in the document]
bakoushin: nope, thanks!

src/points/PointsDiscoverCard.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { Colors } from 'src/styles/colors'
+import { typeScale } from 'src/styles/fonts'
+import { Spacing } from 'src/styles/styles'
+
+interface Props {
+  style?: StyleProp<ViewStyle>
+}
+
+export default function PointsDiscoverCard({ style }: Props) {
+  const showPoints = getFeatureGate(StatsigFeatureGates.SHOW_POINTS)
```
jophish, line 23: Why is the gate checked within the component? I suppose if we're using this card in multiple places then this pattern would reduce some duplication, but my general instinct would be for `<PointsDiscoverCard/>` to _always_ render the card, and to handle feature gating one level above. Not super important to me, but is there any reason in particular to prefer this pattern?

src/points/PointsDiscoverCard.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  const { t } = useTranslation()
+  const pointsBalance = useSelector(pointsBalanceSelector)
+
+  const handlePress = () => {
+    ValoraAnalytics.track(PointsEvents.points_screen_open)
```
jophish, line 29: nit: can this event name be something like `points_discover_press`? If we want the points screen to ever have more entry points, then reusing `points_screen_open` from those new locations, would be ambiguous as to where the user entered from, unless we add on an e.g. `fromScreen` property or something similar.

src/points/slice.ts
```diff
@@ -39,6 +39,7 @@ export interface State {
   pendingPointsEvents: PendingPointsEvent[]
   pointsBalanceStatus: 'idle' | 'loading' | 'error' | 'success'
   pointsBalance: string
+  introHasBeenSeen: boolean
```
jophish, line 42: nit: perhaps `introHasBeenDismissed` is a slightly more accurate name, since you can "see" the intro screen and this can still be `false`. it only becomes `true` once you actually press the dismiss button.

jeanregisser:
> @jeanregisser what kind of animation would you expect?

Not sure. Some kind of navigation animation (slide down, slide left/right).

src/points/PointsDiscoverCard.tsx
bakoushin: yeah, the main reason is modularity -- to encapsulate all related logic within the component, so the parent component doesn't care about conditions. it just provides a place for this card to be rendered in.

currently, we have a single condition -- the feature gate, but it could be many conditions as well. in that case parent component should handle them all and could become kind of bloated.

also, that way we can test the component in isolation.

those overweight for me the possible performance issue caused by unconditional rendering. I think it could be ignored for this particular, relatively lightweight component.

fwiw, the neighboring [EarnCard](https://github.com/valora-inc/wallet/blob/258f77a1958bd88a92bd7efc4cf2c7517b568c48/src/earn/EarnCard.tsx#L16) is implemented in the same way.

I would prefer to keep consistency for now and re-use the `EarnCard` "modular" pattern.

src/points/slice.ts
```diff
@@ -42,6 +42,7 @@ interface State {
   trackOnceActivities: {
     [key in PointsActivityId]?: boolean
   }
+  introHasBeenDismissed: boolean
```
kathaypacific, line 45:
```suggestion
  hasSeenIntro: boolean
```
maybe?

src/points/slice.test.ts
```diff
@@ -59,4 +61,12 @@ describe('pending points events', () => {
 
     expect(newState.pendingPointsEvents).toEqual([pendingSwapEvent])
   })
+
+  it('should set intro has been seen', () => {
+    const initialState = { introHasBeenDismissed: false } as unknown as State
```
kathaypacific, line 66: nit: use `initialState` rather than casting this object to the `State` type - will ensure that we get better type errors (e.g. if you change the name or type of this variable in the slice, this test probably won't complain until you run it)

src/points/PointsIntro.tsx
```diff
@@ -0,0 +1,88 @@
...
+  }, [navigation])
+
+  const onIntroDismiss = () => {
+    ValoraAnalytics.track(PointsEvents.points_intro_dismiss)
+    dispatch(pointsIntroDismissed())
+    replace(Screens.PointsHome)
+  }
+
+  return (
+    <SafeAreaView style={styles.outerContainer} edges={['top']}>
```
kathaypacific, line 43: why not have both top and bottom edges? then maybe you don't need the paddingBottom on the container (which might be too small for some phones)

src/points/PointsDiscoverCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+      navigate(Screens.PointsIntro)
+    }
+  }
+
+  if (!showPoints) {
+    return null
+  }
+
+  return (
+    <Touchable onPress={handlePress} testID="PointsDiscoverCard">
```
kathaypacific, line 43: nit: i think you need to add a `borderRadius` prop here to ensure that the highlighted part on press on android does not exceed the rounded borders

src/points/PointsDiscoverCard.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+  }
+}
+
+describe('PointsDiscoverCard', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  it('renders when feature gate is enabled', () => {
+    jest.mocked(getFeatureGate).mockReturnValue(true)
```
kathaypacific, line 37: nit: maybe this should be in the beforeEach as a default for all tests so you only need to mock it to return false for 1 test case?

src/dappsExplorer/TabDiscover.tsx
```diff
@@ -237,6 +238,7 @@ function TabDiscover({ navigation }: Props) {
                   </Text>
                 }
                 <DappFeaturedActions onPressShowDappRankings={handleShowDappRankings} />
+                <PointsDiscoverCard />
```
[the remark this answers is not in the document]
kathaypacific: argh sorry i discussed this with Laura some time ago but i forgot exactly what the conclusion was...iirc the decision was to eventually remove the `DappFeaturedActions` entirely, which means finding a new home for the claim positions card. so i think let's proceed with this assumption for now, i'll check with Laura/Kayla and can queue up a change if i've remembered wrong!

kathaypacific (approved):
LGTM!!

src/points/slice.ts
bakoushin: this is similar to my initial attempt! 😅

but Joe fairly noted that this flag will only turn on when the user deliberately presses the dismiss button. for instance, they can _see_ the intro and navigate back, or terminate the app. see this [comment](https://github.com/valora-inc/wallet/pull/5406#discussion_r1599312617)

kathaypacific: oh i actually wasn't expecting that behaviour...why would we want to show this screen again if the user navigates back without pressing the button? the user needs to do _something_ to exit the screen, which is an acknowledgement that they've seen the screen (there's no guarantee that a user will actually read + comprehend the messaging on the screen, we tap the big buttons to quickly dismiss things all the time...) i thought the straightforward thing to do is to only show this screen once and mark it as seen once the user navigates away...

src/points/PointsDiscoverCard.tsx
bakoushin: great call! thanks! correct ripple handling required small styles/layout refactoring as well 😅

src/points/PointsIntro.tsx
bakoushin: good catch! i found it not necessary to specify any [`edges`](https://github.com/th3rdwave/react-native-safe-area-context#edges) here at all. defaults works just fine.

PR #5409, opened 2024-05-13
feat(swap): Initial Swap screen redesign

jophish:
### Description

First part of [RET-1065](https://linear.app/valora/issue/RET-1065/[wallet]-swap-screen-redesign). Updates the design of swap screen to match the [designs in Figma](https://www.figma.com/design/fjpfHZdVSLazv330kT2sQz/Cross-chain-swaps?node-id=606-1653&t=Bq4e7bCLkjpGu4D0-4). The rest of the behavior outlined in the ticket will come in followup PRs.

### Test plan

https://github.com/valora-inc/wallet/assets/569401/c078eb34-4cd2-4fe9-a9a2-f5eba85c0c47

### Related issues

- Part of [RET-1065](https://linear.app/valora/issue/RET-1065/[wallet]-swap-screen-redesign).

### Backwards compatibility

Yes.

### Network scalability


locales/base/translation.json
```diff
@@ -1765,10 +1765,11 @@
     "insufficientFunds": "Insufficient {{token}} balance. Try a smaller amount or choose a different asset.",
     "fetchSwapQuoteFailed": "Sorry for the delay! It's taking longer than usual to get the best exchange rate. Please try again in a few minutes.",
     "disclaimer": "Best rate determined by decentralized sources. <0>Learn more</0>",
-    "swapFromTokenSelection": "SWAP FROM",
-    "swapToTokenSelection": "SWAP TO",
```
jeanregisser, line 1769: Ah we can't remove these or older versions won't show it anymore.

src/swap/SwapScreen.tsx
```diff
@@ -983,6 +1005,14 @@ const styles = StyleSheet.create({
     ...typeScale.titleSmall,
     marginTop: -Spacing.Regular16,
   },
+  switchTokens: {
+    position: 'absolute',
+    top: -20,
+    zIndex: 99,
```
jeanregisser, line 1011: 1 should be enough, no?

jeanregisser (commented):
Great!

Things/questions I noted from the video:
- Does the token selection also work when tapping the container of the down arrow?
- The right padding/margin in the skeleton loader seems different from the left one. It should be the same.
- The ripple effect on the switch token button is being clipped by containers above and below it. It should be above them.
- I'm slightly concerned by the fact that the "MAX" button can change position based on the length of the $ value. Maybe something we can improve with Kayla?

src/swap/SwapScreen.tsx
```diff
@@ -953,13 +977,11 @@ const styles = StyleSheet.create({
     flex: 1,
   },
   fromSwapAmountInput: {
-    borderBottomWidth: 0,
-    borderTopLeftRadius: 16,
-    borderTopRightRadius: 16,
+    borderRadius: 16,
+    marginBottom: Spacing.Smallest8,
```
bakoushin, line 981: nit: perhaps we can use `Spacing.Regular16` instead of hardcoded number?

bakoushin:
Looking great!

Noticed a few rendering issues:
1. The switch token button is rendered behind the "to" field on the iOS emulator (see screenshot): 


2. If we put a really long number, everything breaks (see video). This could possibly be a real-life issue for fiats having really large denominations.
Perhaps we could stop the fiat value from growing in width past 40% of the parent width?

https://github.com/valora-inc/wallet/assets/2737872/bd25e8a4-2cff-4039-a72a-82d4f4e83902

3. +1 to Jean:
* I would expect the whole top part of the swap amount box to be pressable. Not only the arrow button.
* I wonder if instead of ripple across the switch token button, we can add a rotation animation instead? Like, the arrow after a tap rotates 360 degrees (don't know though if it is possible to turn off the ripple effect).

jophish:
Responding to some comments:

@jeanregisser 
> Does the token selection also work when tapping the container of the down arrow?
 It did not, but now it does - this required a bit of rejiggering to how Touchables work, since the border radius is uneven depending on the state of the screen.

> The right padding/margin in the skeleton loader seems different from the left one. It should be the same.
Thanks! This has been fixed.

> The ripple effect on the switch token button is being clipped by containers above and below it. It should be above them.
This was a tricky one. I'm somewhat sure that we were running up against [this bug](https://github.com/facebook/react-native/issues/17200), since I _did not_ encounter the adverse overlapping behavior when using `PlatformTouchable.SelectableBackground()` as the background. Fortunately, I was able to work around this using the `foreground` option instead of `background` as described [here](https://github.com/rnc-archive/react-native-platform-touchable?tab=readme-ov-file#additional-props-used-by-touchablenativefeedback--default-android).

> I'm slightly concerned by the fact that the "MAX" button can change position based on the length of the $ value. Maybe something we can improve with Kayla?
The designs have been updated to move the "MAX" button to the right of the fiat estimation.

 @bakoushin
> The switch token button is rendered behind the "to" field on the iOS emulator (see screenshot):
Would you mind testing on iOS again? This is working on Android, but made a few small changes.

> If we put a really long number, everything breaks (see video). This could possibly be a real-life issue for fiats having really large denominations.
I've raised this with Kayla, waiting to hear back about design ideas.

> I would expect the whole top part of the swap amount box to be pressable. Not only the arrow button.
Addressed above!

> I wonder if instead of ripple across the switch token button, we can add a rotation animation instead?
I suppose we could do this, though if the ripple is working correctly (which it should be, now) we could also stick with that. 

See new video here:

https://github.com/valora-inc/wallet/assets/569401/3ee616bc-937d-43d9-87eb-cebbf73a5255

src/swap/SwapScreen.tsx
```diff
@@ -684,14 +698,14 @@ export function SwapScreen({ route }: Props) {
...
-      title: t('swapScreen.swapFromTokenSelection'),
+      title: t('swapScreen.selectTokenTitle'),
       filterChips: filterChipsFrom,
       origin: TokenPickerOrigin.SwapFrom,
     },
     {
       fieldType: Field.TO,
       tokens: swappableToTokens,
-      title: t('swapScreen.swapToTokenSelection'),
+      title: t('swapScreen.selectTokenTitle'),
```
kathaypacific, line 708: since these are the same now, it can probably be passed directly to the component and removed from this map

src/components/Touchable.tsx
```diff
@@ -1,27 +1,41 @@
...
   return borderRadius === 0 ? (
     <PlatformTouchable {...passThroughProps} background={background} />
   ) : (
-    <View style={[{ borderRadius }, !borderless && { overflow: 'hidden' }]}>
-      <PlatformTouchable {...passThroughProps} background={background} />
+    <View style={[{ borderRadius }, touchableStyle, !borderless && { overflow: 'hidden' }]}>
+      <PlatformTouchable
+        {...passThroughProps}
+        background={useForeground ? undefined : background}
+        foreground={useForeground ? background : undefined}
```
kathaypacific, line 37: i got really hung up on this even though it's not important because it's not easy to understand what `useForeground` means. (also reading `if useForeground, use background` is weird). it seems like this prop should be used `if the Touchable has any images as children and you want the ripple to be rendered above the image` - so perhaps a more intuitive prop for this component could be `shouldRenderRippleAbove` or `rippleAboveImages` or `hasImageChildren`?

src/components/Touchable.tsx
```diff
@@ -1,27 +1,41 @@
...
+}: Props) {
   const background = borderless
     ? PlatformTouchable.SelectableBackgroundBorderless()
     : PlatformTouchable.SelectableBackground()
   return borderRadius === 0 ? (
     <PlatformTouchable {...passThroughProps} background={background} />
   ) : (
-    <View style={[{ borderRadius }, !borderless && { overflow: 'hidden' }]}>
-      <PlatformTouchable {...passThroughProps} background={background} />
+    <View style={[{ borderRadius }, touchableStyle, !borderless && { overflow: 'hidden' }]}>
```
kathaypacific, line 33: i have feelings about this because it feels like `touchableStyle` removes the need for `borderRadius`, but at the same time gives too much flexibility to this component (it's definitely harder to screw up the styles by passing in a number for borderRadius as opposed to any stylesheet).

it seems like you're using `touchableStyle` to undo some effects of `borderRadius`. i wonder instead if the borderRadius be a number _or_ an array (like how in css you can apply `padding` in an array)?

src/swap/SwapAmountInput.tsx
```diff
@@ -214,61 +218,51 @@ const styles = StyleSheet.create({
...
     backgroundColor: Colors.white,
     borderWidth: 1,
     borderColor: Colors.gray2,
-    borderRadius: 4,
-    paddingVertical: 2,
-    paddingHorizontal: 4,
-    marginRight: Spacing.Smallest8,
+    borderRadius: Spacing.Tiny4,
+    paddingVertical: Spacing.Tiny4,
+    paddingHorizontal: 6,
```
kathaypacific, line 236: perhaps we can go with 4 or 8 and stick to the design system values?

src/swap/SwapAmountInput.tsx
```diff
@@ -66,143 +67,146 @@ const SwapAmountInput = ({
...
-    padding: Spacing.Regular16,
     backgroundColor: Colors.gray1,
     borderColor: Colors.gray2,
     borderWidth: 1,
   },
-  label: {
-    ...fontStyles.xsmall,
-    fontWeight: 'bold',
+  tokenInfo: {
+    flexDirection: 'row',
```
kathaypacific, line 195: i think you need an alignItems center here so that the icon is vertically centered. if you increase your device font size, the icon is no longer centered currently

src/swap/SwapScreen.tsx
```diff
@@ -989,13 +1017,11 @@ const styles = StyleSheet.create({
...
-    borderBottomWidth: 0,
-    borderTopLeftRadius: 16,
-    borderTopRightRadius: 16,
+    zIndex: -1,
+    marginBottom: Spacing.Smallest8,
   },
   toSwapAmountInput: {
-    borderBottomLeftRadius: 16,
-    borderBottomRightRadius: 16,
+    zIndex: -1,
```
kathaypacific, line 1024: we should minimise changes to the zIndex because it makes the layout unpredictable. i think you can remove all the zIndex styles and apply a single `zIndex: 1` to `switchTokensContainer`. at least this minimises the places that we manually set the layout order

src/swap/SwapAmountInput.tsx
```diff
@@ -214,61 +218,51 @@ const styles = StyleSheet.create({
...
+  },
+  tokenNamePlaceholder: {
+    ...typeScale.labelMedium,
+    paddingHorizontal: 4,
+    color: Colors.gray3,
   },
   fiatValue: {
     ...typeScale.bodyXSmall,
+    paddingLeft: Spacing.Smallest8,
+    maxWidth: '40%',
```
kathaypacific, line 263: if we're adding a max width, we should also add a `numberOfLines={1}` so that the text is ellipsised when the limit is hit rather than spreading to new lines (e.g. if you type a long number like 123455678923624856 in the token amount - unlikely to happen but if the device font size is large then this number can become smaller to reach the limit)

kathaypacific (commented):
looking great!

src/components/Touchable.tsx
jophish: Yeah, I wasn't too pleased with this either. You can't specify border radii using an array/list of values, so I've opted for a similar approach using a style object restricted to certain keys.

src/components/Touchable.tsx
jophish: Yeah, agreed that it's weird as-is... I like `shouldRenderRippleAbove`, since it does have utility outside of using images within the Touchable.

src/swap/SwapAmountInput.tsx
jophish: Aha, good catch.. I had tried to do this but was setting `numberOfLines={1}` at the wrong layer and it wasn't working as intended. Setting it on the `SwapAmountInput/FiatValue` component works well!

src/swap/SwapScreen.test.tsx
```diff
@@ -1180,12 +1177,47 @@ describe('SwapScreen', () => {
...
+    })
+    const { swapFromContainer, swapToContainer, swapScreen, getByTestId } = renderScreen({})
+
+    selectSwapTokens('CELO', 'cUSD', swapScreen)
+
+    expect(within(swapFromContainer).getByTestId('SwapAmountInput/Input')).toBeTruthy()
+    expect(within(swapFromContainer).getByText('CELO')).toBeTruthy()
+
+    expect(within(swapToContainer).getByTestId('SwapAmountInput/Input')).toBeTruthy()
+    expect(within(swapToContainer).getByText('cUSD')).toBeTruthy()
```
kathaypacific, line 1202: nit: `selectSwapTokens` [already includes](https://github.com/valora-inc/wallet/blob/5946f1950f97028282765c31c2319c4cdfd6eabf/src/swap/SwapScreen.test.tsx#L251) these assertions so you don't need to do it again here :)

src/swap/SwapAmountInput.tsx
kathaypacific: bump on this! :)

kathaypacific (approved):
looks great!! 💅🏼

PR #5414, opened 2024-05-15
feat: earn transaction feed items

MuckT:
### Description

Adds the `EarnFeedItem.tsx`, types and associated GraphQL queries to populate the transaction feed.

#### Screenshots

| iOS transaction Feed | iOS Details |
| ----- | ----- |
|  |  |

### Test plan

- [x] Tested locally on iOS
- [x] Unit tests added

### Related issues

- Fixes ACT-1182

### Backwards compatibility

Yes
### Network scalability


src/transactions/feed/EarnFeedItem.tsx
```diff
@@ -0,0 +1,157 @@
...
+      <AmountDisplay transaction={transaction} isLocal={true} />
+    </View>
+  )
+}
+
+interface Props {
+  transaction: EarnWithdraw | EarnDeposit | EarnClaimReward
+}
+
+export default function EarnFeedItem({ transaction }: Props) {
```
silasbw, line 83: Why not make 3 separate components (deposit, withdraw, reward claim)?

MuckT: We could, I just thought this would lead to a lot of duplicate code across the components.

satish-ravi: I guess these are similar enough to be in a single component, but not a fan of the multiple conditional rendering inside title and subtitle, how about something like
```
const TX_TYPE_TO_STRINGS = {
  'EarnDeposit': {
    title: t('earnFlow.transactionFeed.earnDepositTitle'),
    subtitle: t('...'),
  }
  ..
}
```
and then you can do `TX_TYPE_TO_STRINGS[transaction.__typename].title / subtitle` in the JSX
or you can also do a switch case similar to AmountDisplay

src/transactions/feed/EarnFeedItem.tsx
```diff
@@ -0,0 +1,160 @@
...
+    paddingVertical: Spacing.Small12,
+    paddingHorizontal: variables.contentPadding,
+    alignItems: 'center',
+  },
+  contentContainer: {
+    flex: 1,
+    paddingHorizontal: variables.contentPadding,
+  },
+  title: {
+    ...typeScale.labelMedium,
```
satish-ravi, line 140: add a color

src/transactions/feed/EarnFeedItem.tsx
```diff
@@ -0,0 +1,160 @@
...
+  },
+  subtitle: {
+    ...typeScale.bodySmall,
+    color: Colors.gray4,
+  },
+  amountContainer: {
+    maxWidth: '50%',
+  },
+  amountTitle: {
+    ...typeScale.labelMedium,
```
satish-ravi, line 150: add a color

src/transactions/feed/TransactionDetailsScreen.tsx
```diff
@@ -100,12 +106,16 @@ function TransactionDetailsScreen({ route }: Props) {
...
-
       break
     case TokenTransactionTypeV2.SwapTransaction:
       content = <SwapContent exchange={transaction as TokenExchange} />
       retryHandler = () => navigate(Screens.SwapScreenWithBack)
       break
+    case TokenTransactionTypeV2.EarnClaimReward:
+    case TokenTransactionTypeV2.EarnWithdraw:
+    case TokenTransactionTypeV2.EarnDeposit:
+      content = null //TODO details for Earn
```
satish-ravi, line 117: is there a separate ticket for this?

src/transactions/feed/EarnFeedItem.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const isWithdraw = transaction.__typename === 'EarnWithdraw'
+  const isClaim = transaction.__typename === 'EarnClaimReward'
+
+  return (
+    <Touchable
+      testID={`EarnFeedItem/${transaction.transactionHash}`}
+      onPress={() => {
+        ValoraAnalytics.track(EarnEvents.earn_feed_item_select, { origin: transaction.__typename })
+        navigate(Screens.TransactionDetailsScreen, {
+          transaction: { ...transaction, status: TransactionStatus.Complete },
```
satish-ravi, line 100: is there a reason to hardcode complete here? I have a few that have failed already 😄

src/transactions/feed/queryHelper.ts
```diff
@@ -505,6 +508,122 @@ export const TRANSACTIONS_QUERY = gql`
     }
   }
 
+  fragment EarnDepositItemV3 on EarnDeposit {
```
satish-ravi, line 511: Do we need the V3 suffix for these?

src/transactions/feed/TransactionDetailsScreen.tsx
MuckT: There is not. I'll make one, but we are still lacking designs.

src/transactions/feed/queryHelper.ts
MuckT: Fixed in 8dda925decf6f0a1e76868cc7463e000e9690ad5!

src/transactions/feed/EarnFeedItem.tsx
MuckT: Used a switch in be4243dc52188876881cc2b93cc82d22c15e42bc!

src/transactions/feed/EarnFeedItem.tsx
MuckT: Fixed in 16aa133988a508a628bdd9f9bde8e841595bb178!

src/transactions/feed/EarnFeedItem.tsx
MuckT: Fixed in 16aa133988a508a628bdd9f9bde8e841595bb178!

src/transactions/feed/EarnFeedItem.tsx
MuckT: It seems that we're not getting a status back from blockchain-api currently 🤔

MuckT: Fixed in ceeaefcf3d715dcc02c444e1979e5ae709c8b127!

src/transactions/feed/TransactionDetailsScreen.tsx
MuckT: Added ticket and a link to it in 21cfc10839664eebf27f38bf1d4863ad11f8f939!

src/transactions/feed/queryHelper.ts
```diff
@@ -384,6 +384,9 @@ export const TRANSACTIONS_QUERY = gql`
         ...TokenTransferItemV3
         ...NftTransferItemV3
         ...TokenExchangeItemV3
+        ...EarnDepositItem
+        ...EarnWithdrawItem
+        ...EarnClaimRewardItem
```
satish-ravi, line 389: one side effect of starting to query these items is that, even if the gate is off, we'll still be getting and displaying rewards. Maybe that's ok, but just wanted to call it out. cc @silasbw

src/transactions/feed/EarnFeedItem.tsx
```diff
@@ -0,0 +1,175 @@
...
+import TransactionFeedItemImage from 'src/transactions/feed/TransactionFeedItemImage'
+import { EarnClaimReward, EarnDeposit, EarnWithdraw } from 'src/transactions/types'
+
+interface DescriptionProps {
+  transaction: EarnWithdraw | EarnDeposit | EarnClaimReward
+}
+
+function Description({ transaction }: DescriptionProps) {
+  const { t } = useTranslation()
+  const { providerName } = getDynamicConfigParams(
```
satish-ravi, line 27: probably a rare edge case, but in case there are multiple providers in the future, and someone is on the app version that gets the name from the dynamic config, and have earn txs from a different provider, they'll see the Aave name.. maybe one option would be to just filter by provider id (hardcoded aave-v3) for now to be extra safe?

satish-ravi (commented):
Looks good mostly, just a couple of BC related questions

src/transactions/feed/queryHelper.ts
MuckT: In testing we are already displaying them currently just as miscategorized swaps and a payment received.

#### Example from Main

src/transactions/feed/EarnFeedItem.tsx
MuckT: I think we'll have a similar issue with the bottom sheet so I think its fine for now and when we have multiple providers we can have a more robust solution.

silasbw (approved):

PR #5416, opened 2024-05-15
fix(points): enable some activities to only be tracked once

kathaypacific:
### Description

As the title, this PR:
1. adds a redux property that tracks the status of "track once" activities, which is updated after a points event is processed
2. ensure the track event for create-wallet is not fired again after successfully being processed
3. update tests and migrations

I opted not to hook up the create-wallet activity card status with redux because it'd be very confusing for the user in case of tracking failure.

### Test plan

n/a

### Related issues

- Fixes RET-1069

### Backwards compatibility

Y

### Network scalability

### DIFF
--- a/src/home/TabHome.tsx
+++ b/src/home/TabHome.tsx
@@ -125,13 +125,7 @@ function TabHome({ navigation }: Props) {
     // rest of feed to load unencumbered
     setTimeout(tryImportContacts, 500)
 
-    // stale reference to trackOnceActivities is intentional here to avoid
-    // queuing up multiple events with different id's in the case of track
-    // create-wallet failure + successful tracking of another activity in
-    // trackOnceActivities in the same app session
-    if (!trackOnceActivities['create-wallet']) {
-      dispatch(trackPointsEvent({ activityId: 'create-wallet' }))
-    }
+    dispatch(trackPointsEvent({ activityId: 'create-wallet' }))
   }, [])
 
   useEffect(() => {
--- a/src/home/TabHome.tsx
+++ b/src/home/TabHome.tsx
@@ -125,6 +125,10 @@ function TabHome({ navigation }: Props) {
     // rest of feed to load unencumbered
     setTimeout(tryImportContacts, 500)
 
+    // stale reference to trackOnceActivities is intentional here to avoid
+    // queuing up multiple events with different id's in the case of track
+    // create-wallet failure + successful tracking of another activity in
+    // trackOnceActivities in the same app session
     if (!trackOnceActivities['create-wallet']) {
       dispatch(trackPointsEvent({ activityId: 'create-wallet' }))
     }
--- a/src/home/TabHome.tsx
+++ b/src/home/TabHome.tsx
@@ -31,7 +31,6 @@ import { importContacts } from 'src/identity/actions'
 import { Screens } from 'src/navigator/Screens'
 import useScrollAwareHeader from 'src/navigator/ScrollAwareHeader'
 import { StackParamList } from 'src/navigator/types'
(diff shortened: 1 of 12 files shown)


src/home/TabHome.tsx
```diff
@@ -123,7 +125,13 @@ function TabHome({ navigation }: Props) {
...
 
-    dispatch(trackPointsEvent({ activityId: 'create-wallet' }))
+    // stale reference to trackOnceActivities is intentional here to avoid
+    // queuing up multiple events with different id's in the case of track
+    // create-wallet failure + successful tracking of another activity in
+    // trackOnceActivities in the same app session
+    if (!trackOnceActivities['create-wallet']) {
+      dispatch(trackPointsEvent({ activityId: 'create-wallet' }))
+    }
   }, [])
```
jeanregisser, line 135: nit: We could even decouple this one entirely from the UI and manage it just from the saga (waiting for a created account event).

src/points/slice.ts
```diff
@@ -92,12 +98,14 @@ const slice = createSlice({
...
+    pointsEventProcessed: (state, action: PayloadAction<PendingPointsEvent>) => {
+      state.pendingPointsEvents = state.pendingPointsEvents.filter(
         (event) => event.id !== action.payload.id
-      ),
-    }),
+      )
+      if (action.payload.event.activityId in state.trackOnceActivities) {
+        state.trackOnceActivities[action.payload.event.activityId] = true
+      }
+    },
```
jeanregisser, line 108: Shouldn't we add it to `trackOnceActivities` when `trackPointsEvent` is dispatched?
It's guaranteed it will be processed because there's a queue, right?

Otherwise it's possible we dispatch more than once in a session. Though unlikely if requests are fast.

jeanregisser (commented):
🙌

bakoushin (approved):
💫

src/home/TabHome.tsx
kathaypacific: yes i was also thinking about this, since you also brought it up i moved it to the saga level. i thought if we do this on the UI level then we can save some actions being dispatched...but doing it on the saga level is more neat and less error prone so it's probably better

src/points/slice.ts
kathaypacific: yeah you're right, i can move this further up. i think it can't be as far up as `trackPointsEvent` though because we need a chance to check whether the activity has been tracked. i've moved it to be updated on `sendPointsEventStarted`

src/points/slice.ts
```diff
@@ -39,9 +39,12 @@ export interface State {
   pendingPointsEvents: PendingPointsEvent[]
   pointsBalanceStatus: 'idle' | 'loading' | 'error' | 'success'
   pointsBalance: string
+  trackOnceActivities: {
```
jophish, line 42: nit: i think this name is totally fine as-is, but did you consider e.g. `binaryActivities`? might be a bit more clear, since track _could_ be called for one of these activities multiple times if the first call fails

jophish (approved):
LGTM! with optional comment on naming

src/points/slice.ts
kathaypacific: i thought about this for a while (naming this thing has been really hard) but i think i prefer to stick to the current name because `binaryActivities` could read like an object that describes which activities are binary rather than whether we've tracked an activity that should be tracked only once. i hear the point about how these `trackOnce` activities could be tracked multiple times but i guess our intention is to prevent that as much as we can...excluding the reinstall use case, we only call the track endpoint once for these activities (and only multiple times if the track fails, but you could argue that a failed track doesn't count as a track)

PR #5417, opened 2024-05-15
feat(earn): implement deposit submit saga

satish-ravi:
### Description

Submits the deposit transactions, mostly based on the swap and jumpstart sagas

### Test plan

Unit tests, manually
- tx currently fails because of a gas limit issue, will make a follow up for this
- loading state is done in #5415 

Uploading earn-submit-tx.mp4…

### Related issues

- Part of ACT-1178

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/src/earn/saga.ts
+++ b/src/earn/saga.ts
@@ -36,7 +36,7 @@ export function* depositSubmitSaga(action: PayloadAction<DepositInfo>) {
   }
 
   const networkId = tokenInfo.networkId
-  const trackedProperties = {
+  const commonAnalyticsProps = {
     tokenId,
     networkId,
     tokenAmount: amount,
@@ -54,7 +54,7 @@ export function* depositSubmitSaga(action: PayloadAction<DepositInfo>) {
     const createDepositStandbyTxHandlers = []
 
     if (preparedTransactions.length > 1 && preparedTransactions[0].data) {
-      const { functionName, args } = yield* call(decodeFunctionData, {
+      const { functionName, args } = decodeFunctionData({
         abi: erc20.abi,
         data: preparedTransactions[0].data,
       })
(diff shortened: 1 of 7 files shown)


src/earn/saga.ts
```diff
@@ -1,14 +1,177 @@
...
+
+  const tokenInfo = yield* call(getTokenInfo, tokenId)
+  if (!tokenInfo) {
+    Logger.error(`${TAG}/depositSubmitSaga`, 'Token info not found for token id', tokenId)
+    yield* put(depositError())
+    return
+  }
+
+  const networkId = tokenInfo.networkId
+  const trackedProperties = {
```
silasbw, line 39: nit: move this down closer to where it's used?

src/earn/saga.ts
```diff
@@ -1,14 +1,177 @@
...
+  try {
+    Logger.debug(
+      `${TAG}/depositSubmitSaga`,
+      `Starting deposit for token ${tokenId}, total transactions: ${preparedTransactions.length}`
+    )
+
+    const createDepositStandbyTxHandlers = []
+
+    if (preparedTransactions.length > 1 && preparedTransactions[0].data) {
+      const { functionName, args } = yield* call(decodeFunctionData, {
```
silasbw, line 57: probably dumb q: what's the reason for not calling `decodeFunctionData` directly?

silasbw (approved):

src/earn/saga.ts
satish-ravi: this has to be outside the try so it can be used in the catch as I well. I can rename this to something like `sharedAnalyticsProps` or something to make it more obvious

src/earn/saga.ts
satish-ravi: just an optimization I guess? we only need this if there's more than one tx. Just copied from swaps

PR #5419, opened 2024-05-16
feat(swap): show app fee details inline

jeanregisser:
### Description

This shows the app fee details directly as a line item.

### Test plan

- Updated tests

| No quote | Valora fee disabled or 0 fee from API | Valora fee > 0 from API | Valora fee > 0 from API, but no USD price |
|----------|----------|----------|----------|

Notes: 
- copies are not final

### Related issues

- Fixes [RET-1061](https://linear.app/valora/issue/RET-1061)

### Backwards compatibility

Yes

### Network scalability

If a new NetworkId and/or Network are added in the future, the changes
in this PR will:

- [x] Continue to work without code changes, OR trigger a compilation
error (guaranteeing we find it when a new network is added)


src/swap/SwapTransactionDetails.tsx
```diff
@@ -194,15 +200,52 @@ export function SwapTransactionDetails({
...
+            })
+            appFeeInfoBottomSheetRef.current?.snapToIndex(0)
+          }}
+          label={t('swapScreen.transactionDetails.appFee')}
+          testID="SwapTransactionDetails/AppFee/MoreInfo"
+        />
+        <Text testID={'SwapTransactionDetails/AppFee'} style={styles.value}>
+          <Trans
+            i18nKey={'swapScreen.transactionDetails.appFeeValue'}
+            context={
```
bakoushin, line 217: this kind of feels too smart use for [context](https://www.i18next.com/translation-function/context) 😅

it implies this `context` prop conditions and underlying children rendering conditions must be kept in sync (e.g. both have to rely on `appFee.token.priceUsd`). i expect it to be hard/error-prone to quickly adjust in the future: one has to account for this, and the `translation.json` template as well.

wouldn't the good old conditional rendering without translation templates/context be less complex?

bakoushin (approved):
LGTM!

Left one nit about perceived complexity, more to have a discussion.

src/swap/SwapTransactionDetails.tsx
jeanregisser: Yes it is a bit complex. But we need the ability to properly translate this.

The children conditionals aren't entirely needed, I did that to make it easier to understand tests.

We could have 4 very similar renders. Probably easier to understand but with more duplication. Though there's also test coverage.

In the interest of time, I can take a stab at this later to not block the current release further than needed. What do you think?

bakoushin: yes, I agree!
I'm not sure if we actually have to tune it given there is a refactoring of swap details is ahead

PR #5423, opened 2024-05-16
feat(buttons): Change onboarding buttons to new design

finnian0826:
### Description

Update the onboarding buttons

### Test plan

unit/e2e tests pass

Manual Tests:
Simulator Screenshot - iPhone 14 Pro - 2024-05-16 at 12 33 56
Simulator Screenshot - iPhone 14 Pro - 2024-05-16 at 12 34 02
Simulator Screenshot - iPhone 14 Pro - 2024-05-16 at 12 34 25

### Related issues

- N/A

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -17,6 +17,7 @@ export enum BtnTypes {
   SECONDARY = 'Secondary',
   SECONDARY_WHITE_BG = 'SecondaryWhiteBg',
   GRAY_WITH_BORDER = 'GrayWithBorder',
+  ONBOARDING_SECONDARY = 'OnboardingSecondary',
 }
 
 export enum BtnSizes {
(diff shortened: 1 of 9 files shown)


finnian0826:
The paste button on restore wallet looks a bit weird, should that one stay as the old version?

satish-ravi:
> The paste button on restore wallet looks a bit weird, should that one stay as the old version? It feels weird to keep around the style for just one use case, but I'm not seeing any other way to override color to white.

Looks like this sets the a rounded: false property, but the gray with border doesn't seem to honor it and always sets a borderRadius. I'd suggest we fix this if its simple, or leave onboarding_secondary as deprecated and use it in one place, with a follow up ticket

finnian0826:
Kept onboarding secondary as it was used for the paste button and looks weird in a different style, added a follow-up ticket for removing the deprecated style: https://linear.app/valora/issue/ACT-1200/placeholder-remove-deprecated-onboarding-button-style

MuckT (approved):
Thanks for catching the undesired change with the paste button!

PR #5427, opened 2024-05-17
feat(jumpstart): add specific error messaging when link is already claimed

kathaypacific:
### Description

As the title

### Test plan

https://github.com/valora-inc/wallet/assets/20150449/b701d6b9-758e-4742-99f8-02da11641ba9

### Related issues

- Fixes RET-1075

### Backwards compatibility

Y

### Network scalability


src/jumpstart/saga.ts
```diff
@@ -67,10 +67,12 @@ export function* jumpstartClaim(privateKey: string, networkId: NetworkId, wallet
     ValoraAnalytics.track(JumpstartEvents.jumpstart_claim_succeeded)
 
     yield* put(jumpstartClaimSucceeded())
-  } catch (error) {
+  } catch (error: any) {
     Logger.error(TAG, 'Error handling jumpstart link', error)
     ValoraAnalytics.track(JumpstartEvents.jumpstart_claim_failed)
-    yield* put(jumpstartClaimFailed())
+    yield* put(
```
jophish, line 73: I think the pattern of parsing the error message to determine the error type is prolly fine here, since we're only discerning between two types of error, though in general this does seem like it has the potential to break stuff if we're not careful... e.g. if i'm adding a new error, i have to know that it can't contain certain substrings else other code will interpret it as a certain "kind" of error. one alternative is having some custom error type that we stick some error enum value on, which eliminates that confusion. 

i think for this case it's fine since this is simple and we prolly won't add more error types, but wanted to call it out since i've definitely run into error-parsing-hell when dealing with handling errors from external libs that are only distinguished by their (inconsistently formatted) messages.

jophish (approved):
general comment about distinguishing error types, but no changes needed :)

src/jumpstart/saga.ts
kathaypacific: yeah good shout, this is a fair point. i've been mulling this over but i think there's always going to be some level of brittleness with error handling here because there are multiple points that the claim flow can throw (including when we generate the transaction and when we make the network request to claim the transaction). so we don't have control over every error that can exist, and (afaik) it's not possible to enforce any checks on exactly which errors are thrown (i.e. we can declare a map of custom errors, but that relies on the sagas actually using the map when creating the errors) which at least for now seems like extra effort but still leaves the door open for us to screw this up in the future. so i guess for now my preference would be to leave it as is and think about it more when we add more custom errors - i'm not sure that is likely :)

PR #5429, opened 2024-05-17
feat(earn): redux setup for withdraw

satish-ravi:
### Description

Also changed from started -> loading to be consistent with other slices

### Test plan

Unit tests

### Related issues

- Part of ACT-1180

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/src/earn/slice.ts
+++ b/src/earn/slice.ts
@@ -1,14 +1,18 @@
 import { PayloadAction, createSlice } from '@reduxjs/toolkit'
 import { REHYDRATE, RehydrateAction } from 'redux-persist'
-import { DepositInfo } from 'src/earn/types'
+import { DepositInfo, WithdrawInfo } from 'src/earn/types'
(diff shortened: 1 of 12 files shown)


src/earn/slice.ts
```diff
@@ -1,22 +1,26 @@
...
 import { REHYDRATE, RehydrateAction } from 'redux-persist'
-import { DepositInfo } from 'src/earn/types'
+import { DepositInfo, WithdrawInfo } from 'src/earn/types'
 import { getRehydratePayload } from 'src/redux/persist-helper'
 
+type Status = 'idle' | 'loading' | 'success' | 'error'
+
 interface State {
-  depositStatus: 'idle' | 'started' | 'success' | 'error'
+  depositStatus: Status
```
MuckT, line 9: Eventually have to handle multiple pools so deposit and withdraw status could need to be represented differently, but I think for the MVP this is fine.

MuckT (approved):

PR #5430, opened 2024-05-17
feat(earn): dispatch withdraw start on pressing collect

satish-ravi:
### Description

Dispatches the redux action on pressing the collect earnings button and fires analytics event

### Test plan

Unit tests

### Related issues

- Part of ACT-1180

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/src/earn/EarnCollectScreen.tsx
+++ b/src/earn/EarnCollectScreen.tsx
@@ -4,14 +4,18 @@ import React from 'react'
 import { useTranslation } from 'react-i18next'
 import { SafeAreaView, ScrollView, StyleSheet, Text, View } from 'react-native'
 import SkeletonPlaceholder from 'react-native-skeleton-placeholder'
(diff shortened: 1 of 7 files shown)


src/earn/EarnCollectScreen.tsx
```diff
@@ -40,7 +47,34 @@ export default function EarnCollectScreen({ route }: Props) {
     feeCurrencies,
   })
   const onPress = () => {
-    // TODO(ACT-1180): submit the tx
+    if (!asyncRewardsInfo.result || asyncPreparedTransactions.result?.type !== 'possible') {
```
MuckT, line 50:
```suggestion
    if (!asyncRewardsInfo.result || (asyncPreparedTransactions.result && asyncPreparedTransactions.result.type !== 'possible')) {
```

nit: to get rid of optional chaining

MuckT (approved):

PR #5432, opened 2024-05-17
feat: query transactions on networks independently

MuckT:
### Description

PR for independent network queries; `useFetchTransactions` in `src/transactions/feed/queryHelper.ts` has a lot going on to manage the transaction state. In the future, something like [@tanstack/react-query](https://github.com/TanStack/query) might serve us better, but is outside of the scope of this PR. 

#### Video

| iOS Before | iOS After |
| ----- | ----- |
| <video src="https://github.com/valora-inc/wallet/assets/26950305/4cd55c3d-dd0b-4cf5-8144-8a2811526cb8" /> | <video src="https://github.com/valora-inc/wallet/assets/26950305/69414274-895d-4741-a073-4506fbdefeed" /> |

### Test plan

prerequisite: add your wallet address to the overrides for [`base_and_polygon_targeting`](https://console.statsig.com/4plizaPmWwPL21ASV4QAO0/gates/base_and_polygon_targeting) in Statsig.

- [x] Tested locally on iOS
- [x] Unit tests updated

### Related issues

- Fixes ACT-1186

### Backwards compatibility

Yes

### Network scalability


src/transactions/feed/queryHelper.ts
```diff
@@ -187,8 +188,7 @@ export function useFetchTransactions(): QueryHookResult {
   // Query for more transactions if requested
   useAsync(
     async () => {
-      if (!fetchingMoreTransactions || !anyNetworkHasMorePages(fetchedResult.pageInfo)) {
```
MuckT, line 190: Removing `!fetchingMoreTransactions` allows tests to pass, but can result in multiple fetch requests firing. Should likely be included and tests updated.

src/transactions/feed/queryHelper.ts
```diff
@@ -278,35 +298,44 @@ function anyNetworkHasTransactionsOnCurrentPage(hasTransactionsOnCurrentPage: {
...
 
-  return results.reduce((acc, result, index) => {
-    return {
-      ...acc,
-      [params[index].networkId]: result,
-    }
-  }, {})
+  for (const promise of promises) {
+    const { networkId, result } = await promise
+    yield { [networkId]: result }
```
satish-ravi, line 337: If I understand correctly, this would start all the requests in parallel and then serially wait for requests to complete one after the other right? If so, if we have a slow running request at the first one, it would block all subsequent requests right? And if one of them fails, wouldn't all subsequent requests be ignored even if they succeed? Wonder if we can do a non generator based solution where we just invoke some action that can update tx feed for a single n/w

MuckT: All the requests are started in parallel within the map and you are correct that a slow request would block the subsequent requests from being processed; however, they will not be removed from the activeRequests state. During the next polling cycle, the slower request will be skipped, if it's listed as having an active request, and the other networks will be processed accordingly. 

`handleResult` currently updates the transaction feed for a single network; however, it's called from two hooks and has some heavy lifting interacting fetchedResult state. Dispatching it from `queryTransactionsFeed` might be a bit complicated when we have to deal with empty pages and possible refetches.

MuckT: @satish-ravi I think this is now Fixed in #5528 which uses an alternative approach.

MuckT:
Closing in favor of #5528.

PR #5439, opened 2024-05-21
fix(earn): navigation to enter amount on deposit more tap

MuckT:
### Description

Hooks up the deposit more button on the Earn cta to navigation and analytics.

### Test plan

- [x] Tested locally on iOS
- [x] Unit tests added 

### Related issues

- Part of ACT-1176

### Backwards compatibility

Yes

### Network scalability


kathaypacific (approved):
🚀

PR #5441, opened 2024-05-21
chore(points): add track points event to transaction watcher

kathaypacific:
### Description

In case the transaction is not settled within the same app session, we should add the points tracking to the transaction watcher as well.

I added the network to the tracked event, in preparation for network specific activities.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y

### Network scalability


src/points/types.ts
```diff
@@ -84,6 +85,7 @@ interface PointsEventCreateWallet {
 interface PointsEventSwap {
   activityId: 'swap'
   transactionHash: string
+  network: Network
```
jeanregisser, line 88: What do you think of using `NetworkId` to remove any ambiguity?

kathaypacific: yes fair, i updated to use networkId. i thought i was saving some computation by using the network but i was mistaken (reading the wallet code i thought that we need only the network to create the viem client, but we've a couple of layers of mappings in between!)

PR #5445, opened 2024-05-21
fix(earn): hide pools with less than .001

MuckT:
### Description

Hides pools with a token balances below `0.001.`

#### Screenshots

| iOS Before | iOS After |
| ----- | ----- | 
|  |  |

### Test plan

- [x] Tested locally on iOS

### Related issues

N/A

### Backwards compatibility

Yes

### Network scalability


MuckT:
Closing as we will find a better solution to hide dust that doesn't include a hardcoded value.

PR #5447, opened 2024-05-21
feat(earn): add gas related analytics properties

satish-ravi:
### Description

Adding gas fields to analytics properties similar to swaps for the earn deposit transaction

### Test plan

Unit tests

### Related issues

N/A

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/src/transactions/types.ts
+++ b/src/transactions/types.ts
@@ -1,6 +1,8 @@
 import BigNumber from 'bignumber.js'
 import { Nft } from 'src/nfts/types'
+import { TransactionRequest } from 'src/viem/prepareTransactions'
 import { v4 as uuidv4 } from 'uuid'
+import { Hash, TransactionReceipt } from 'viem'
 
 export enum Network {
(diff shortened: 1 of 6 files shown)


src/transactions/types.ts
```diff
@@ -246,3 +248,9 @@ export interface EarnClaimReward {
   providerId: string
   status: TransactionStatus
 }
+
+export interface TrackedTx {
```
satish-ravi, line 252: moved from swap/saga to a shared location

src/transactions/utils.ts
```diff
@@ -46,3 +57,66 @@ export function groupFeedItemsInSections<T extends { timestamp: number }>(
       data: value.data,
     }))
 }
+
+export function getTxReceiptAnalyticsProperties(
```
satish-ravi, line 61: moved from swaps/saga to a shared location

MuckT (approved):

PR #5450, opened 2024-05-21
refactor(earn): consolidate analytics events

satish-ravi:
### Description

Use consistent naming, include some common properties in events

### Test plan

Unit tests

### Related issues

N/A

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/src/analytics/Properties.tsx
+++ b/src/analytics/Properties.tsx
@@ -1581,18 +1581,18 @@ interface PointsEventsProperties {
   [PointsEvents.points_screen_activity_learn_more_press]: undefined
 }
 
-interface EarnDepositProperties {
-  tokenId: string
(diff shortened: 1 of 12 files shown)


silasbw (approved):

src/analytics/Properties.tsx
```diff
@@ -1581,18 +1581,18 @@ interface PointsEventsProperties {
   [PointsEvents.points_screen_activity_learn_more_press]: undefined
 }
 
-interface EarnDepositProperties {
-  tokenId: string
+interface EarnCommonProperties {
+  providerId: 'aave-v3'
```
silasbw, line 1585: nit: should this be `PROVIDER_ID` instead of `'aave-v3'`

PR #5455, opened 2024-05-22
feat(earn): update cta to match new designs

satish-ravi:
### Description

Just the EarnCta card in the Discover tab. The home screen changes will be in a follow up

[Figma](https://www.figma.com/design/E1rC3MG74qEg5V4tvbeUnU/Stablecoin-Enablement?node-id=2906-9161&t=qHZ4KetHUCGCOedp-0)

### Test plan

Unit tests, manual

### Related issues

- Part of ACT-1191

### Backwards compatibility

Yes

### Network scalability


src/earn/EarnCta.tsx
```diff
@@ -1,36 +1,67 @@
...
+  const asyncPoolInfo = useAavePoolInfo({ depositTokenId })
+
+  if (!depositToken) {
+    // should never happen
+    Logger.error(TAG, `No deposit token found: ${depositTokenId}`)
+    return null
+  }
+
+  const apyDisplay = asyncPoolInfo.result
+    ? new BigNumber(asyncPoolInfo.result.apy).multipliedBy(100).toFixed(2)
```
silasbw, line 33: annoying unhelpful nit: multiplying an APY by 100 is an odd thing to do (since "APY" implies it's already a percentage and why would we multiple an percentage by 100?).

i know the naming predates this PR, but similar naming issues have caused small bugs in the swap experience. naming something "percentage" when it's really a proportion, or vice versa.

silasbw (approved):

src/earn/EarnCta.tsx
```diff
@@ -1,36 +1,67 @@
...
+            <TokenIcon token={depositToken} size={IconSize.SMALL} viewStyle={styles.icon} />
             <View style={styles.subtitleContainer}>
-              <Text style={styles.subtitle}>{t('earnFlow.cta.subtitle')}</Text>
-              <Text style={styles.description}>{t('earnFlow.cta.description')}</Text>
+              <Text style={styles.subtitle}>
+                {t('earnFlow.ctaV1_86.subtitle', { symbol: depositToken.symbol })}
+              </Text>
+              <Text style={styles.description}>
+                {t('earnFlow.ctaV1_86.description', {
+                  apy: apyDisplay,
```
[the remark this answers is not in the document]
satish-ravi: yes, spoke to John and we decided to drop the fiat value. The figma is updated to reflect this

src/earn/EarnCta.tsx
satish-ravi: I've always thought multiplying by 100 is a display thing, and the value is always a proportion. (like how excel / google sheets display percentages, the underlying value is always a proportion). This makes it easier for calculations too on wherever we use it. If this has caused issues in the past, we can change it to something else in a follow up issue.

jh2oman (approved):

src/earn/EarnCta.tsx
silasbw: That makes sense.  The issues we've ran into our with 3rd parties.  E.g., `slippagePercentage` here https://0x.org/docs/0x-swap-api/api-references/get-swap-v1-quote.

The convention that the wallet always stores a proportion sounds good.

PR #5458, opened 2024-05-23
chore(points): send token ids with swap events

kathaypacific:
### Description

The core points logic doesn't rely on the to and from tokens in a swap, these are only used for display purposes. Since we already have them in the wallet, it'd save a lot of work to just pass them to the backend (rather than deriving them again using the transactions and adding logic for creating the token id's on the backend)

### Test plan

n/a

### Related issues

- Related to RET-1034

### Backwards compatibility

Y

### Network scalability

### DIFF
--- a/src/points/slice.test.ts
+++ b/src/points/slice.test.ts
@@ -15,7 +15,13 @@ const pendingCreateWalletEvent: PendingPointsEvent = {
 const pendingSwapEvent: PendingPointsEvent = {
   id: 'test-id-2',
   timestamp: '2024-04-22T12:00:00.000Z',
-  event: { activityId: 'swap', transactionHash: '0xTEST', networkId: NetworkId['celo-alfajores'] },
+  event: {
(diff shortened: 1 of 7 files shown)


jophish (approved):
great idea! i suppose if someone _really_ wanted to, they could make their own swap activity feed look a little wonky by passing unrelated token ids 😄

jophish:
actually, mirroring the discussion on the related backend PR, should we also update this to include `toNetworkId` and `fromNetworkId` here? or maybe in a different PR if you want

kathaypacific:
> actually, mirroring the discussion on the related backend PR, should we also update this to include `toNetworkId` and `fromNetworkId` here? or maybe in a different PR if you want

i replied to the comment in the backend PR [here](https://github.com/valora-inc/points-functions/pull/106#discussion_r1612942120) - i don't think we need both networkIds but happy to add it in a follow up if i'm wrong!

PR #5459, opened 2024-05-23
fix(points): ensure balance is displayed timely and correctly

kathaypacific:
### Description

This PR fixes:
1. the points balance was outdated on the discover tab because the balance fetch was only triggered on the points home screen. now we also trigger a balance fetch on app mount.
2. also fetch points config on app mount (consolidate logic, was previously being fetched on home screen visit)
3. disable font scaling on the points balance because the font size is large already, and font scaling messes with the animation.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y

### Network scalability


jophish (approved):

bakoushin (approved):
💅

PR #5463, opened 2024-05-23
style: add jsx expressions linter

satish-ravi:
### Description

ESLint rule that catches conditional rendering bugs. More context [here](https://valora-app.slack.com/archives/C025V1D6F3J/p1714697074743779)

### Test plan

CI

### Related issues

N/A

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/src/icons/BackChevron.tsx
+++ b/src/icons/BackChevron.tsx
@@ -5,20 +5,14 @@ import Svg, { Path } from 'react-native-svg'
 import Colors from 'src/styles/colors'
 
 export interface Props {
-  height?: number
-  color?: ColorValue | Animated.Node<ColorValue | undefined>
+  height: number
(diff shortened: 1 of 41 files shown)


src/icons/BackChevron.tsx
MuckT, line 1: Interesting that this and `ForwardChevron.tsx` where working with a default hight or hight being passed to the Component. Should we provide default values?

MuckT (approved):

src/icons/BackChevron.tsx
MuckT: Tested locally on iOS and didn't notice any changes in the display of `ForwardChevron.tsx`.

silasbw (approved):
:partying_face:

jeanregisser (approved):
Awesome!! Thanks so much 🚀 🚀

PR #5466, opened 2024-05-24
fix(points): app mount sagas have not been running

bakoushin:
### Description

I noticed that points app mount sagas have not been running and apparently they have to be wrapped into `call` methods to run.

To the worse, the initial unit test was wrong and didn't catch this issue -- it relied on mock functions _being called_ within the test, which created illusion of success.

### Test plan

* Tested manually
* Updated unit tests

### Related issues

NA

### Backwards compatibility

Y

### Network scalability

NA


src/points/saga.test.ts
```diff
@@ -560,13 +563,20 @@ describe('watchAppMounted', () => {
     const mockGetPointsConfig = jest.fn()
     const mockAction = { type: AppActions.APP_MOUNTED }
 
+    jest.mocked(safely).mockImplementation((saga) => {
```
kathaypacific, line 566: instead of this, can we do `.provide([[call(getPointsConfig), {}], ...)` in the tests? that is the pattern that we've more commonly been using for mocking

kathaypacific (approved):
thank you for fixing!! sorry for the careless bug 🙈

src/points/saga.test.ts
bakoushin: removed the hacky test for now. will follow up with an additional PR once figure out a proper solution.

bakoushin:
closing in favor of #5479

PR #5477, opened 2024-05-27
feat(swaps): add new fee container with feature flag

kathaypacific:
### Description

This PR includes:
1. a dummy feature flag that is hard coded switch between the new and old swap details UI. i opted for the simple approach here (it seems unnecessary to create a remote feature flag for this, since we want to enable it for everyone as soon as it's code complete)
2. UI for the entry point of the new swap details

I opted for the approach of writing new code rather than changing existing code because the new UI shows less information than the old UI so we need to keep the old UI in place until it is ready. i also think it's maybe easier to review. With that in mind, i've suffixed the existing UI with `Old`.

### Test plan

Happy + sad flow if we are missing some information (for this example, there is no estimated network fee because i have no ETH):

https://github.com/valora-inc/wallet/assets/20150449/c6887f30-390c-4957-bef9-8daed067bc53

### Related issues

- Related to RET-1066

### Backwards compatibility

Y

### Network scalability


src/swap/SwapTransactionDetails.tsx
```diff
@@ -263,7 +295,196 @@ export function SwapTransactionDetails({
...
+    return null
+  }
+
+  if (appFee && appFee.amount.gt(0) && !appFee.token.priceUsd) {
+    Logger.warn(
+      'swap/useTotalSwapFeesInLocalCurrency',
+      `Cannot calculate swap fees due to missing usd price for app fee token ${appFee.token.tokenId}`
+    )
+    return null
+  }
```
jeanregisser, line 349: Though the other cases can't really happen (we should always have a price for the network fee and cross-chain fee), this one for sure happens.
Should we count it as 0 then to at least partially show the "total"?

jeanregisser (approved):
Awesome! 🚀

src/swap/SwapTransactionDetails.tsx
kathaypacific: my concern is that we're only showing this "bite size" version of the fee in fiat, if we don't have the price of some tokens in the fee and set them to 0, then the user could see a low fee value and proceed with the swap thinking it's a great deal. my intention was to show only the total fee if we have the complete picture, if not then don't show anything and require the user to go into the details screen, where they can see the breakdown of the fees to understand the full picture.

kathaypacific:
Closing as the designs have changed

PR #5483, opened 2024-05-28
feat(earn): prepare transactions for supply when gas fee is covered

jh2oman:
### Description

When gas fees are subsidized:
1. use the simulateTransactions endpoint to estimate gas for the approve transaction.
1. ignore limitations of not having enough gas when preparing transactions

### Test plan

Verified that the transaction goes through when using the syndicate RPC node on mainnet and that the UI works regardless of whether or not the user has enough gas when the feature flag is on.

### Related issues

https://linear.app/valora/issue/ACT-1193/new-enter-amount-screen


src/viem/prepareTransactions.test.ts
```diff
@@ -198,6 +198,47 @@ describe('prepareTransactions module', () => {
...
     })
+    it("returns a 'possible' result when the balances for feeCurrencies are too low to cover the fee but isGasSubsidized is true", async () => {
+      mocked(estimateFeesPerGas).mockResolvedValue({
+        maxFeePerGas: BigInt(100),
+        maxPriorityFeePerGas: BigInt(2),
+        baseFeePerGas: BigInt(50),
+      })
+      mocked(estimateGas).mockResolvedValue(BigInt(1_000))
+
+      // max gas fee is 10 * 10k = 100k units, too high for either fee currency
```
satish-ravi, line 209: where do the 10 and 10k come from? should this be 100 (maxFeePerGas) * 1k (gas) ?

src/earn/prepareTransactions.test.ts
```diff
@@ -146,6 +147,84 @@ describe('prepareTransactions', () => {
...
+              status: 'success',
+              blockNumber: '1',
+              gasNeeded: 50000,
+              gasUsed: 49800,
+              gasPrice: '1',
+            },
+          ],
+        })
+      )
+      jest.mocked(getFeatureGate).mockReturnValue(true)
```
satish-ravi, line 175: specify the feature gate here. E.g., `jest.mocked(getFeatureGate).mockImplementation(gate => gate === <gate-name>)`, so this can be cleaned up when removing a gate. Also should this be reset in a beforeEach so it doesn't affect other tests?

jh2oman: already had it reset in the beforeEach, but I'll do the mockImplementation in both

src/viem/prepareTransactions.test.ts
jh2oman: This was copy pasted from the previous test...I think you're right. I'll fix it in both!

satish-ravi (approved):

PR #5485, opened 2024-05-29
refactor(token-bottom-sheet): support bottom sheet screen and use in cico

satish-ravi:
### Description

Refactors the TokenBottomSheet component so that it can be used as a bottom sheet screen for CICO flow. This will allow enabling filters without needing to duplicate much of the code

### Test plan

Confirmed token bottom sheet works correctly on all the flows

### Related issues

- Part of ACT-1710

### Backwards compatibility

Yes

### Network scalability


src/components/TokenBottomSheet.tsx
```diff
@@ -224,76 +230,90 @@ function TokenBottomSheet({
...
-      </BottomSheetBase>
+        )}
+      </View>
+    </View>
+  )
+
+  return (
+    <>
+      {isScreen ? (
+        <View style={{ height: variables.height * 0.9 }}>{content}</View>
```
satish-ravi, line 311: one implication of this is that the bottom sheet will always be 90% high regardless of the number of tokens. If this is not set and the token list exceeds the screen height, the bottom sheet occupies the full screen. The BottomSheetScrollView (which we use for other bottom sheet screens) specifies a maxHeight to the scroll view, but wasn't able to use that here. Open to other ideas.

kathaypacific: what about using just `maxHeight` here? you can remove the `flex:1` on the `container` style to make that work (might need to bring the `container` into the render method and use it only if not `isScreen`)

satish-ravi: this worked. thanks!

jh2oman (approved):

PR #5528, opened 2024-06-10
feat: query transactions on networks independently

MuckT:
### Description

Alternative implementation of #5432 which addresses the comments raised in https://github.com/valora-inc/wallet/pull/5432#discussion_r1622895331 passing state and setState to the `queryTransactionsFeed` function.

### Test plan

- [x] Tested locally on iOS
- [x] Unit tests updated

### Related issues

- Fixes ACT-1186

### Backwards compatibility

Yes

### Network scalability


src/transactions/feed/queryHelper.ts
```diff
@@ -282,31 +304,45 @@ async function queryTransactionsFeed({
...
+  updateTransactionState: (networkId: NetworkId, data: QueryResponse | null) => void
+  setActiveRequests: (updateFunc: (prevState: ActiveRequests) => ActiveRequests) => void
+  activeRequests: ActiveRequests
+}): Promise<void> {
+  // Launch all network requests without waiting for each to finish before starting the next
+  const requests = params.map(async ({ networkId, afterCursor }) => {
+    // Prevent duplicate requests for the same network
+    if (activeRequests[networkId]) {
+      Logger.info(TAG, `Skipping fetch for ${networkId} as it is already active`)
+      return
```
jeanregisser, line 326: I'm wondering if we could avoid that check and the need to pass `setActiveRequests` / `activeRequests`
Is the problem that `useAsync` can call `queryTransactionsFeed` multiple times, but we effectively only care about the last one?

src/transactions/feed/queryHelper.ts
```diff
@@ -282,31 +304,45 @@ async function queryTransactionsFeed({
...
   localCurrencyCode: string
   params: Array<{
     networkId: NetworkId
     afterCursor?: string
   }>
-}): Promise<{ [key in NetworkId]?: QueryResponse }> {
-  const results = await Promise.all(
-    params.map(({ networkId, afterCursor }) =>
-      queryChainTransactionsFeed({
+  updateTransactionState: (networkId: NetworkId, data: QueryResponse | null) => void
```
jeanregisser, line 317: nit: to make it clear it's a callback.
```suggestion
  onNetworkResponse: (networkId: NetworkId, data: QueryResponse | null) => void
```

src/transactions/feed/queryHelper.ts
```diff
@@ -107,76 +125,65 @@ export function useFetchTransactions(): QueryHookResult {
...
   const { loading, error } = useAsync(
     async () => {
-      const result = await queryTransactionsFeed({
+      await queryTransactionsFeed({
         address,
         localCurrencyCode,
         params: allowedNetworkIds.map((networkId) => {
           return { networkId }
         }),
+        updateTransactionState: (networkId, result) => {
```
jeanregisser, line 141: I wonder if we need to worry that is is handled only for the "current" promise of `queryTransactionsFeed`.

But maybe since the result is deduplicated it's fine?

jeanregisser (commented):
Awesome! Thanks for working on this! 🚀

src/transactions/feed/queryHelper.ts
MuckT: I added this because some slow networks, such as polygon, take longer than the polling interval to return a result, so we can end up with multiple pending requests that are slow to resolve.

src/transactions/feed/queryHelper.ts
MuckT: We also handle the pending states for polling and pagination separately so it should only deduplicate on each one of the pending states respectively.

jeanregisser (approved):

PR #5529, opened 2024-06-10
fix(send): Fix bug where you can't send max amount

finnian0826:
### Description

When pressing the max button a user would always get the error saying to enter a lower amount, this fixes it.

Explanation [here](https://valora-app.slack.com/archives/C02KBT0DAHJ/p1718034648230829).

### Test plan

CI

### Related issues

- N/A

### Backwards compatibility

Yes

### Network scalability


MuckT (approved):
Tested by sending max of USDC on Polygon. LGTM!

PR #5542, opened 2024-06-13
feat(viem): support using valora rpc endpoint if gas is subsidized

satish-ravi:
### Description

In preparation for using syndicate's rpc node for earn transactions when gas is subsidized, update the viem functions to use a different rpc endpoint for both estimating gas and submitting transaction

### Test plan

CI, a follow up PR will have comprehensive testing

### Related issues

- Part of ACT-1194

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/src/web3/contracts.ts
+++ b/src/web3/contracts.ts
@@ -38,6 +38,7 @@ let wallet: KeychainWallet | undefined
 let contractKit: ContractKit | undefined
 
 const viemWallets = new Map<Chain, ViemWallet>()
+const valoraViemWallets = new Map<Chain, ViemWallet>()
 
 const keychainLock = new KeychainLock()
 const initContractKitLock = new Lock()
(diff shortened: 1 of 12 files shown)


src/web3/networkConfig.ts
```diff
@@ -94,6 +94,7 @@ interface NetworkConfig {
   arbAavePoolV3ContractAddress: Address
   arbAaveIncentivesV3ContractAddress: Address
   aaveArbUsdcTokenId: string
+  valoraRpcUrl: Record<Network.Arbitrum, string>
```
satish-ravi, line 97: probably can be a single base url from which we can compute per n/w if we go with the url format as below, will clean in a follow up once we confirm this pattern works

src/web3/contracts.ts
```diff
@@ -122,9 +123,10 @@ async function waitForContractKit(tries: number) {
 
 // This code assumes that the account for walletAddress already exists in the Keychain
 // which is a responsibility currently handled by KeychainWallet
-export function* getViemWallet(chain: Chain) {
-  if (viemWallets.has(chain)) {
-    return viemWallets.get(chain) as ViemWallet
+export function* getViemWallet(chain: Chain, useValoraTransport?: boolean) {
```
MuckT, line 126: Should this function take an object instead of multiple parameters to avoid accidentally setting the wrong values?

satish-ravi: I left a comment above about this in general 😄 this will be in a follow up

MuckT (approved):

PR #5543, opened 2024-06-13
fix(points): Update swap points copy

jophish:
### Description

Updates a bit of copy for the swap points activity to be correct!

### Related issues

- Fixes #[issue number here]

### Network scalability


jeanregisser (approved):
👍

PR #5551, opened 2024-06-14
fix(max): respect local decimal separator when setting max

MuckT:
### Description

Discovered with `EarnEnterAmount.tsx`, but also present in `EnterAmount.tsx` when pressing max the local decimilar separator was not set correctly preventing users from being able to delete characters from the input amount.

### Test plan

- [x] Tested locally on iOS
- [x] Unit tests updated

### Related issues

N/A

### Backwards compatibility

Yes

### Network scalability

N/A

### DIFF
--- a/src/send/EnterAmount.test.tsx
+++ b/src/send/EnterAmount.test.tsx
@@ -320,6 +320,26 @@ describe('EnterAmount', () => {
       expect(tokenAmountInput.props.value).toBe(replaceSeparators('1000'))
       expect(localAmountInput.props.value).toBe(replaceSeparators('₱133'))
     })
+
+    it('entering MAX token applies correct decimal separator', async () => {
+      const store = createMockStore(mockStore)
+      const tokenBalances = mockStoreBalancesToTokenBalances([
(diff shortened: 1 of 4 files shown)


src/send/EnterAmount.test.tsx
```diff
@@ -320,6 +320,26 @@ describe('EnterAmount', () => {
       expect(tokenAmountInput.props.value).toBe(replaceSeparators('1000'))
       expect(localAmountInput.props.value).toBe(replaceSeparators('₱133'))
     })
+
+    it('entering MAX token applies correct decimal separator', async () => {
+      const store = createMockStore(mockStore)
+      const tokenBalances = mockStoreBalancesToTokenBalances([
+        { ...mockStoreTokenBalances[mockCeloTokenId], balance: '100000' },
```
satish-ravi, line 327: Can we include decimals in the balance? Otherwise I believe this test will pass even without the fix

MuckT: Fixed in 3bdc56680ff84de59208546f881e760f549d0551!

satish-ravi (approved):

PR #5554, opened 2024-06-17
feat(onboarding): Update welcome screen

finnian0826:
### Description

Add background image, replace text with "valora", update button copy and langauge header color.

Assets added in https://github.com/valora-inc/valora-app-branding/pull/34

### Test plan

Simulator Screenshot - iPhone 14 Pro - 2024-06-17 at 16 04 10

### Related issues

- Fixes #ACT-1215

### Backwards compatibility

Yes

### Network scalability


locales/base/translation.json
```diff
@@ -826,6 +826,7 @@
     "createNewWallet": "Create new wallet",
     "restoreWallet": "Restore wallet",
     "hasWallet": "I already have a wallet",
+    "hasWalletv2": "I have a wallet",
```
satish-ravi, line 829: any reason to use a new variant here? If we need one, should follow the pattern of how other strings are versioned. E.g., `V1_xx`. there should be other examples in this file

src/icons/Valora.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react'
```
satish-ravi, line 1: seems like something that should go in the branding repo, also could use a better name.

src/icons/Valora.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+interface Props {
+  width?: number
+  height?: number
+  style?: ViewStyle
+}
+
+export default function Valora({ width = 243, height = 64, style }: Props) {
+  return (
+    <Svg width={width} height={height} viewBox={`0 0 ${width} ${height}`} fill="none" style={style}>
```
satish-ravi, line 13: this needs a ViewBox if you want to update width and height with props

locales/base/translation.json
finnian0826: I was just following what was in the designs, should I update `hasWallet` to remove the already, or just use hasWallet as is?

src/onboarding/welcome/Welcome.tsx
```diff
@@ -87,49 +88,48 @@ export default function Welcome() {
 
   return (
     <SafeAreaView style={styles.container}>
-      <ScrollView contentContainerStyle={styles.contentContainer}>
```
MuckT, line 90: Should this scroll view be retained?

finnian0826: I think the scroll view was there for the welcome text that was being shown, now it is just the valora icon so not sure if it makes sense to have that be in a scroll view since it should just be in the center of the screen and not change size based on text size on the phone

src/icons/Valora.tsx
finnian0826: Is that something different than the `viewBox={`0 0 ${width} ${height}`}` that I have? In other examples with this like Blob.tsx it is done the same as here.

src/icons/Valora.tsx
MuckT: Is there a way to exclude the `Valora` text from the image and use the translations for that text? Otherwise, this should probably go into the branding repo.

branding/celo/src/icons/WelcomeLogo.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react'
```
finnian0826, line 1: Wasn't sure what the placeholder value should be so just copied over from Logo.tsx and updated the props to match the real component

branding/celo/src/icons/WelcomeLogo.tsx
```diff
@@ -0,0 +1,31 @@
...
+import colors from 'src/styles/colors'
+import Svg, { Path } from 'svgs'
+
+interface Props {
+  width?: number
+  height?: number
+  style?: ViewStyle
+}
+
+export default function Logo({ width, height, style }: Props) {
```
MuckT, line 12:
```suggestion
export default function WelcomeLogo({ width, height, style }: Props) {
```

MuckT (approved):

PR #5560, opened 2024-06-20
fix(earn-enter-amount): local earn up to amount

MuckT:
### Description

Fixes the display of the earn up to amount to display in the users selected local currency: [Slack Thread](https://valora-app.slack.com/archives/C04B61SJ6DS/p1718895183418909).

#### Screenshots

| iOS Before PHP | iOS After PHP |
| ----- | ----- |
|  |  |

### Test plan

- [x] Tested locally on iOS

### Related issues

N/A

### Backwards compatibility

Yes

### Network scalability

N/A

### DIFF
--- a/locales/base/translation.json
+++ b/locales/base/translation.json
@@ -2412,7 +2412,7 @@
       "earnUpToLabel": "You could earn up to:",
       "rateLabel": "Rate (est.)",
       "earnUpTo": "~{{fiatSymbol}}{{amount}} / yr",
-      "earnUpToV2": "<0></0> / yr",
+      "earnUpToV1_87": "<0></0> / yr",
       "rate": "{{rate}}% APY",
       "continue": "Continue",
       "info": "This pool is powered by Aave",
--- a/locales/base/translation.json
+++ b/locales/base/translation.json
@@ -2412,6 +2412,7 @@
       "earnUpToLabel": "You could earn up to:",
       "rateLabel": "Rate (est.)",
       "earnUpTo": "~{{fiatSymbol}}{{amount}} / yr",
+      "earnUpToV2": "<0></0> / yr",
       "rate": "{{rate}}% APY",
       "continue": "Continue",
       "info": "This pool is powered by Aave",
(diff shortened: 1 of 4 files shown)


locales/base/translation.json
```diff
@@ -2412,6 +2412,7 @@
       "earnUpToLabel": "You could earn up to:",
       "rateLabel": "Rate (est.)",
       "earnUpTo": "~{{fiatSymbol}}{{amount}} / yr",
+      "earnUpToV2": "<0></0> / yr",
```
finnian0826, line 2415: should this be `v1_XX` to match the other updated versions of translations?

MuckT: Fixed in af0b259a889916e4d1efa01cde54d79fccdbf0ed!

src/earn/EarnApyAndAmount.tsx
```diff
@@ -24,14 +22,12 @@ export function EarnApyAndAmount({
...
 
-  const localCurrencySymbol = useSelector(getLocalCurrencySymbol) ?? LocalCurrencySymbol.USD
-
   const asyncPoolInfo = useAavePoolInfo({ depositTokenId: token.tokenId })
   const apy = asyncPoolInfo?.result?.apy
 
   const apyString = apy ? (apy * 100).toFixed(2) : '--'
   const earnUpTo =
-    apy && tokenAmount?.gt(0) ? tokenAmount.multipliedBy(new BigNumber(apy)).toFormat(2) : '--'
+    apy && tokenAmount?.gt(0) ? tokenAmount.multipliedBy(new BigNumber(apy)) : new BigNumber(0)
```
finnian0826, line 30: When APY is loading this will now show earn up to 0 / yr instead of -- / yr, right? Just confirming that this is the behavior we want for the loading state. It seems like it would be tricky to have it show something different while using TokenDisplay

finnian0826 (approved):
one question about loading state but otherwise looks good

src/earn/EarnApyAndAmount.tsx
finnian0826: Actually, is there a reason the skeleton placeholder was removed?

MuckT: That's correct, it'll display `0 / yr` instead of `-- / yr` as the `TokenDisplay` component does not contain a loading state.

PR #5579, opened 2024-06-26
chore(points): reset `trackOnceActivities`

bakoushin:
### Description

After [migrating points contract to ERC20](https://github.com/valora-inc/valora-points-contracts/pull/42) and [resetting the DB](https://github.com/valora-inc/points-functions/pull/150) it makes sense to reset `trackOnceActivities` state in redux, so the testers can get their points for creating wallet.

### Test plan

* Tested manually
* Updated unit test

### Related issues

- Related to RET-1093

### Backwards compatibility

Y

### Network scalability


jeanregisser (approved):
Great! 💯

PR #5583, opened 2024-06-27
chore: update es-419 translation

mpgaarciaa:


kathaypacific:
hey @mpgaarciaa we can make updates to localised transactions in Crowdin, only the base translations should go through the wallet to avoid conflicts :) closing this!

PR #5590, opened 2024-06-28
feat(cab): phone code input changes for setup in onboarding

satish-ravi:
### Description

Changes to the help bottom sheet during the cloud backup setup in onboarding

### Test plan

Unit tests, manual

| Onboarding | Settings |
|--------|--------|
| <video src="https://github.com/valora-inc/wallet/assets/5062591/906f4317-9290-4083-8164-468284f58d41" /> | <video src="https://github.com/valora-inc/wallet/assets/5062591/23d38e01-5915-4f2f-b6a2-b19622649140" /> | 

### Related issues

- Fixes ACT-1232

### Backwards compatibility

Yes

### Network scalability


src/keylessBackup/KeylessBackupPhoneCodeInput.tsx
```diff
@@ -21,15 +22,53 @@ import { Spacing } from 'src/styles/styles'
 import VerificationCodeInput from 'src/verify/VerificationCodeInput'
 
 function HelpInfoBottomSheet({
-  onPressHelpGoBack,
```
MuckT, line 24: This was changed from the function being passed to values used within the component to create the appropriate functions. If we passed the functions we would need to conditionally pass different functions, is that correct? If so this seems fine.

MuckT (commented):
Looks like there might be some missing annotations for the docs, but otherwise LGTM!

src/keylessBackup/KeylessBackupPhoneCodeInput.tsx
satish-ravi: Just moved closer to where the functions are really used. We can still pass the 3 functions and conditionally decide what function to use as now, but the parent component doesn't need to define that.

MuckT (approved):

PR #5592, opened 2024-06-28
chore: remove ivory onboarding background

MuckT:
### Description

Removes usages of onboardingBackground as background and sets color name to ivory for those places where it is used. Some screens such as `ChooseYourAdventure.tsx`, which relied on the background to display the buttons, now use a button background color.

| CYA iOS Before | CYA iOS After | Phrase iOS Before | Phrase iOS After |
|  |  |  |  |

### Test plan

- [x] Tested locally on iOS

### Related issues

N/A

### Backwards compatibility

Yes

### Network scalability

N/A

### DIFF
--- a/src/backup/BackupPhraseContainer.tsx
+++ b/src/backup/BackupPhraseContainer.tsx
@@ -179,7 +179,7 @@ const styles = StyleSheet.create({
     flexWrap: 'wrap',
     flexDirection: 'row',
     marginTop: 8,
-    backgroundColor: colors.onboardingBackground,
+    backgroundColor: colors.ivory,
     borderRadius: 8,
     alignContent: 'center',
     justifyContent: 'center',
--- a/src/styles/colors.tsx
+++ b/src/styles/colors.tsx
@@ -4,6 +4,7 @@ export enum Colors {
   primary = '#1AB775',
   primaryDisabled = `${primary}80`, // 50% opacity
   white = '#FFFFFF',
+  ivory = '#F9F6F0',
   black = '#2E3338',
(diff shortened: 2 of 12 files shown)


src/backup/BackupPhraseContainer.tsx
```diff
@@ -179,7 +179,7 @@ const styles = StyleSheet.create({
     flexWrap: 'wrap',
     flexDirection: 'row',
     marginTop: 8,
-    backgroundColor: colors.onboardingBackground,
+    backgroundColor: colors.ivory,
```
finnian0826, line 182: Where is this backup phrase container used? Looking at the new designs for onboarding all of the backup phrase stuff I see looks like it has gray1 background ([example](https://www.figma.com/design/mthUlChSbWXXxLe2AV13Wd/Onboarding-2024?node-id=2512-4691&t=AT3fVvxFFHr78bl5-0))

src/styles/colors.tsx
```diff
@@ -23,8 +24,6 @@ export enum Colors {
   goldBrand = '#FBCC5C',
   /** @deprecated */
   onboardingBrownLight = '#A49B80',
```
finnian0826, line 26: Is `onboardingBrownLight` still used or could we remove it? It looks like it was removed in a lot of the files above

MuckT: I swapped it out in some of the onboarding screens; however, it's still used in several places: [Slack Thread](https://valora-app.slack.com/archives/C03D3K379FE/p1719612125810649).

jh2oman (approved):

PR #5604, opened 2024-07-01
fix(points): fetch points balance and history on backup restore

bakoushin:
### Description

Start fetching points balance and history on backup restore. Otherwise users will not see their points balance when they navigate to the Dicover section right after restoring their wallet.

Context: https://valora-app.slack.com/archives/C04B61SJ6DS/p1719509875907719?thread_ts=1719507689.657979&cid=C04B61SJ6DS

| Before | After |
|--------|--------|
|  |   |

### Test plan

* Tested manually

### Related issues

NA

### Backwards compatibility

Y

### Network scalability


jophish (approved):
ah, nice fix, lgtm - i remember running into this during bugapalooza

src/points/saga.ts
```diff
@@ -275,6 +276,15 @@ function* watchTrackPointsEvent() {
...
   yield* take(AppActions.APP_MOUNTED)
+  yield* call(initializePoints)
+}
+
+export function* watchInitializeAccountSuccess() {
+  yield* take(AccountActions.INITIALIZE_ACCOUNT_SUCCESS)
+  yield* call(initializePoints)
+}
+
+function* initializePoints() {
```
jeanregisser, line 287: This works, note we could also do this (though `getWalletAddress` should probably be updated to reflect what it waits).

```suggestion
  // This waits for the wallet address to exist before proceeding
  yield* call(getWalletAddress)
```

I guess since `initializePoints` `spawns` sagas, it's unclear without looking whether they are long running or not, and whether it's "safe" to spawn multiple instances so maybe preferable to do it only once and wait for the account?

jeanregisser (approved):

src/points/saga.ts
bakoushin: @jeanregisser oh, thanks for this suggestion! it makes the change a lot simpler.

jeanregisser (approved):

PR #5605, opened 2024-07-01
feat(points): Add support for earning Points for sending Live Links

jophish:
### Description

For [RET-1079](https://linear.app/valora/issue/RET-1079/enable-points-for-live-links-send). Adds support in-app for earning points on Live Link sends, and rendering history.

Note that both the backend and the Jumpstart contract support ERC721 sends, though it's currently not possible in-app to use the Jumpstart system to send NFTs. Despite this, since it's supported everywhere else, ERC721 Live Link history is supported in-app.

### Test plan

Unit and manual tested, see video below. Note that the reclaim in the video is meant to demonstrate that points are _not_ granted for anything other than sends, even though sends/claims both interact with the Jumpstart contract.

https://github.com/valora-inc/wallet/assets/569401/1477002f-a1dc-4b84-8300-51a959e73240

### Related issues

- Fixes [RET-1079](https://linear.app/valora/issue/RET-1079/enable-points-for-live-links-send).

### Backwards compatibility

Yes.

### Network scalability


src/jumpstart/saga.ts
```diff
@@ -329,6 +331,17 @@ export function* sendJumpstartTransactions(
...
+      trackPointsEvent({
+        activityId: 'create-live-link',
+        liveLinkType: 'erc20',
+        beneficiaryAddress,
+        transactionHash: jumpstartTxReceipt.transactionHash,
+        networkId,
+        tokenId: sendToken.tokenId,
+        amount: sendAmount,
+      })
+    )
```
jeanregisser, line 344: Something to consider in the future, [to avoid sequential dispatch of actions](https://redux.js.org/style-guide/#avoid-dispatching-many-actions-sequentially), we should listen to the `depositTransactionSucceeded` action (attaching the additional needed data) in the points saga.

src/jumpstart/saga.ts
```diff
@@ -329,6 +331,17 @@ export function* sendJumpstartTransactions(
...
     yield* put(depositTransactionSucceeded())
+    yield* put(
+      trackPointsEvent({
+        activityId: 'create-live-link',
+        liveLinkType: 'erc20',
+        beneficiaryAddress,
+        transactionHash: jumpstartTxReceipt.transactionHash,
+        networkId,
+        tokenId: sendToken.tokenId,
+        amount: sendAmount,
```
jeanregisser, line 342: Shouldn't we let the backend retrieve `beneficiaryAddress`, `liveLinkType`, `tokenId` and `amount` from the sent TX?
I know we did something similar for swaps, but I'm slightly worried we'll keep doing this, making it easier for people to abuse the system and send incorrect data.
i.e. we may forget about this and later on we'll start using this unverified data for another purpose while it could be faked.

jeanregisser (commented):
Great! 🚀

src/jumpstart/saga.ts
jeanregisser: Sorry I should have raised that when we implemented the backend part, similar to this:

https://github.com/valora-inc/points-functions/blob/fd8e68236aee1b1657fcbba2223d9df438a7041b/src/endpoints/trackPointsEvent/types.ts#L26-L33

jophish: @jeanregisser None of these fields are used for purposes of verification; a user can't abuse the system (i.e., claim points unfairly) by messing with these fields. ATM, these are just used for logging, and `tokenId` is used for eventually rendering the history card (which is the same use case as for swaps). I do understand the concern though, in the case that we _eventually_ use these fields for something more meaningful. I think there are a few options if we want to do something now.

1. Remove all "unnecessary" fields from the request (those not used for rendering the history card), keeping just `tokenId`, and do not bother collecting them in the backend, since they're just for logging.
2. Do the same as 1., but collect those fields from the contract
3. Keep things as-is, but add a comment in the backend (like the one we have for swaps) with the caveat that we would need to collect these data from the transaction on-chain in order to use them for verification.

I kind of lean toward 3., since it works for swaps and I think it's relatively clear that client-provided data shouldn't be trusted. The actual transaction verification is very localized within the code, so it wouldn't be difficult to prevent improper usage with a well-placed comment imo. It's also not clear to me that we have plans in the near future to modify the TX validation requirements.

src/jumpstart/saga.ts
jophish: ah yeah, that's a good callout!

bakoushin: oh we had a [similar discussion](https://github.com/valora-inc/wallet/pull/5294/files#r1575826469) with Kathy in the swaps tracker. perhaps we should refactor all tracking to listeners within the points saga to not have this discussion arise over and over again :) i'm sorry i've set the wrong pattern when implementing the first tracking events.

jophish: See [this PR](https://github.com/valora-inc/wallet/pull/5631) for a potential implementation of this pattern

src/jumpstart/saga.ts
jeanregisser: Cool yes 3 sounds good!

src/jumpstart/saga.ts
jeanregisser: Oh thanks for taking a stab at this 👍

jeanregisser (approved):
🙌

PR #5608, opened 2024-07-02
chore(points): points discover card redesign

bakoushin:
### Description

Implement recent designs for points discover card.

Figma: https://www.figma.com/design/rXBDplfMHHqYmuu6EkgMEo/Gamification-experiments?node-id=1907-11167&t=llLXJVN4HRpgRqIr-0

### Test plan

* Tested manually

### Related issues

- Related to RET-1103

### Backwards compatibility

Y

### Network scalability


src/points/PointsDiscoverCard.tsx
```diff
@@ -36,63 +36,68 @@ export default function PointsDiscoverCard() {
...
+    >
+      <>
+        <Image style={styles.image} source={pointsCardBackground} />
+        <View style={styles.content}>
+          <Text style={styles.title}>{t('points.discoverCard.title')}</Text>
           <Text style={styles.description}>{t('points.discoverCard.description')}</Text>
-        </>
-      </Touchable>
-    </View>
+          <View style={styles.pill}>
```
jophish, line 51: When testing this out, for some reason I expected the pill to be independently tappable apart from the card as a whole (like you tap it and it highlights just the pill). Doesn't matter since the whole card is tappable, but I guess my instinct with a pill like that is that it itself is a tappable button :sweat_smile:

src/points/PointsDiscoverCard.tsx
```diff
@@ -36,63 +36,68 @@ export default function PointsDiscoverCard() {
...
-        <>
-          <View style={styles.header}>
-            <Text style={styles.title}>{t('points.discoverCard.title')}</Text>
-            <View style={styles.pill}>
-              <Text style={styles.balance}>{pointsBalance}</Text>
-              <LogoHeart size={Spacing.Regular16} />
-            </View>
-          </View>
+    <Touchable
+      shouldRenderRippleAbove
```
jophish, line 40: Why's the `shouldRenderRippleAbove` here now?

src/points/PointsDiscoverCard.tsx
jophish, line 51: nit: You can remove this outer `View` and merge `styles.pill` and `styles.balance`

bakoushin: yeah, it definitely looks like a button. let me ask Kayla about that.

src/points/PointsDiscoverCard.tsx
bakoushin: it's because it ripples beneath the image and the pill otherwise (see below). 
I would be glad for any suggestions on how to solve that in another way 😅

| `shouldRenderRippleAbove=true` | `=false`  |
|--------|--------|
|  |  |

src/points/PointsDiscoverCard.tsx
bakoushin: apparently, not on iPhone 😢:

image

jophish: Ohhh weird, I tried this on android and it worked fine for me, weird that the behavior is inconsistent!
2024-07-02-130916_950x505_scrot

src/points/PointsDiscoverCard.tsx
jophish: ahh got it, ty!

src/points/PointsDiscoverCard.tsx
bakoushin: so she thinks it is not necessary for this pill behave as a button:  https://valora-app.slack.com/archives/C0684TXDR3K/p1719941062683099?thread_ts=1719940711.678299&cid=C0684TXDR3K

my choice is to not add its own ripple, to keep the code simple.

jophish: okay! sounds good to me :)

jophish (approved):
A'ing (once tests pass) since you've addressed my comments!

src/points/PointsDiscoverCard.tsx
bakoushin: I found that `borderRadius: 100` is rendering weirdly on iOS, while in Android it does the same as on the Web: any big enough value yields perfectly rounded sides. If I set `borderRadius` somewhere equal to the actual height of the element iOS renders it kind of okay (plus `overflow:hidden`). But I prefer to not bother about keeping border radius in sync with the height, just have a big enough value that always works :)

PR #5611, opened 2024-07-04
feat: rework cab for onboarding flow

MuckT:
### Description

Works to adjust the cab flow for similar styles to onboarding. 

#### Brief Summary of Changes:
- Styled bottom buttons to match other onboarding buttons - previously they displayed lower than other onboarding buttons and buttons throughout the app.
- Replaced usages of nav option headers with customer headers to match onboarding headers and allow onboarding step display.
- Placed Icons on Keyless backup a set distance from the top as to avoid icon content jumps during the flow.
- Replaced top-level views`View` with `ScrollView` throughout the `KeylessBackupProgress.tsx`.

### Test plan

- [x] Tested locally on iOS
- [x] Unit tests updated

### Related issues

- Fixes ACT-1233

### Backwards compatibility

Yes

### Network scalability

N/A


src/keylessBackup/KeylessBackupProgress.tsx
```diff
@@ -257,15 +270,24 @@ function Restore() {
...
+  const onboardingProps = useSelector(onboardingPropsSelector)
+  const { step, totalSteps } = getOnboardingStepValues(Screens.SignInWithEmail, onboardingProps)
   const { t } = useTranslation()
+  const { bottom } = useSafeAreaInsets()
+  const insetsStyle = {
+    paddingBottom: Math.max(0, 40 - bottom),
+  }
 
   const navigatedFromSettings = origin === KeylessBackupOrigin.Settings
+  const isOnboarding = origin === KeylessBackupOrigin.Onboarding
```
finnian0826, line 282: Do we need both `navigatedFromSettings` and `isOnboarding`? One being true should imply that the other is false (unless we have some plan to have a 3rd way to get to CAB).

src/keylessBackup/KeylessBackupProgress.tsx
```diff
@@ -285,80 +307,150 @@ function Setup({ origin }: { origin: KeylessBackupOrigin }) {
...
+          firstScreenInCurrentStep: Screens.SignInWithEmail,
+          onboardingProps: { ...onboardingProps, showRecoveryPhraseEducation: true },
+        })
+      : navigate(Screens.AccountKeyEducation)
   }
 
   const onPressSkip = () => {
     ValoraAnalytics.track(KeylessBackupEvents.cab_progress_failed_skip_onboarding)
-    navigate(Screens.ChooseYourAdventure)
+    isOnboarding
```
finnian0826, line 320: We should only see the skip button in onboarding, so shouldn't need this check.

src/keylessBackup/KeylessBackupProgress.tsx
```diff
@@ -285,80 +307,150 @@ function Setup({ origin }: { origin: KeylessBackupOrigin }) {
 
   const onPressManualOnboarding = () => {
     ValoraAnalytics.track(KeylessBackupEvents.cab_progress_failed_manual, { origin })
-    navigate(Screens.AccountKeyEducation)
+    isOnboarding
```
finnian0826, line 310: This also should only be called from onboarding so shouldn't need this check either.

src/keylessBackup/KeylessBackupProgress.tsx
MuckT: We don't need both, but using `isOnboarding` is more explicit and will keep working if we add another origin.

src/keylessBackup/KeylessBackupProgress.tsx
MuckT: Fixed in 03ba34c68ada224b2e007793d3f03ace442c3a25!

src/keylessBackup/KeylessBackupProgress.tsx
MuckT: Fixed in f797fdfcf188e439b71e6ec58f1eb987e9bfe8bc!

src/keylessBackup/KeylessBackupProgress.tsx
```diff
@@ -169,58 +169,71 @@ function Restore() {
                 t('keylessBackupStatus.restore.completed.bodyZeroBalance')
               )}
             </Text>
+          </ScrollView>
+          <View style={styles.buttonContainerRestore}>
```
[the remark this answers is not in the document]
MuckT: Fixed in b18ce9ba701ceafb8052ce1ffd268d81ee1a1184! For places that had multiple buttons, I kept the view.

src/analytics/Properties.tsx
```diff
@@ -256,14 +256,17 @@ interface KeylessBackupEventsProperties {
...
   [KeylessBackupEvents.cab_sign_in_with_google]: CommonKeylessBackupProps
   [KeylessBackupEvents.cab_sign_in_with_google_success]: CommonKeylessBackupProps
+  [KeylessBackupEvents.cab_sign_in_with_email_screen_back]: CommonKeylessBackupProps
   [KeylessBackupEvents.cab_sign_in_with_email_screen_cancel]: CommonKeylessBackupProps
   [KeylessBackupEvents.cab_sign_in_with_email_screen_skip]: CommonKeylessBackupProps
   [KeylessBackupEvents.cab_enter_phone_number_continue]: CommonKeylessBackupProps
-  [KeylessBackupEvents.cab_enter_phone_number_cancel]: CommonKeylessBackupProps
+  [KeylessBackupEvents.cab_enter_phone_number_back]: CommonKeylessBackupProps
   [KeylessBackupEvents.cab_intro_continue]: CommonKeylessBackupProps
+  [KeylessBackupEvents.cab_enter_phone_number_cancel]: CommonKeylessBackupProps
```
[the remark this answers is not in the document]
MuckT: Fixed in 19c59fa7b3967d2b4deeff99b61e157d8e2636cd!

src/keylessBackup/KeylessBackupProgress.tsx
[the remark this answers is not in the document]
MuckT: I added a bit more of a summary to the description.

src/keylessBackup/KeylessBackupProgress.tsx
```diff
@@ -295,70 +323,127 @@ function Setup({ origin }: { origin: KeylessBackupOrigin }) {
...
+        </SafeAreaView>
+      )
     }
     case KeylessBackupStatus.Completed: {
       return (
-        <SafeAreaView style={styles.container}>
-          <View style={styles.finishedContainer}>
-            <View style={styles.iconContainer}>
+        <SafeAreaView style={styles.safeAreaView}>
+          {isOnboarding && (
```
satish-ravi, line 353: this can probably be made more DRY since the same safe area view / header wrapper is duplicated in every case. Could be in a follow up

src/keylessBackup/KeylessBackupProgress.tsx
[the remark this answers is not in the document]
satish-ravi: ty!

src/keylessBackup/KeylessBackupProgress.tsx
```diff
@@ -257,15 +271,29 @@ function Restore() {
...
+  const { bottom } = useSafeAreaInsets()
+  const insetsStyle = {
+    paddingBottom: Math.max(0, 40 - bottom),
+  }
 
   const navigatedFromSettings = origin === KeylessBackupOrigin.Settings
+  const isOnboarding = origin === KeylessBackupOrigin.Onboarding
 
   const onPressContinue = () => {
     ValoraAnalytics.track(KeylessBackupEvents.cab_progress_completed_continue)
```
satish-ravi, line 286: maybe not in scope of this PR, but should this include origin?

src/keylessBackup/KeylessBackupProgress.tsx
```diff
@@ -295,70 +323,127 @@ function Setup({ origin }: { origin: KeylessBackupOrigin }) {
...
             testID="KeylessBackupProgress/Continue"
             onPress={onPressContinue}
             text={t('continue')}
             size={BtnSizes.FULL}
             type={BtnTypes.PRIMARY}
-            style={styles.button}
             touchableStyle={styles.buttonTouchable}
+            style={[
+              styles.buttonContainer,
+              isOnboarding ? insetsStyle : { marginBottom: Spacing.Thick24 },
```
satish-ravi, line 381: wonder if we could drop the conditional padding and just be consistent for both settings and onboarding

satish-ravi (approved):

src/keylessBackup/KeylessBackupProgress.tsx
MuckT: Fixed in 30e1497cf0936fecd68252c2e11295e4d74734da!

src/keylessBackup/KeylessBackupProgress.tsx
MuckT: Removed in 9886de417fa3605fba501d79e259b17ab1544fe7!

src/keylessBackup/KeylessBackupProgress.tsx
MuckT: Since it's within the setup function I don't think we need to specify the `KeylessBackupFlow`.

satish-ravi: origin can still be onboarding / settings right?

src/keylessBackup/KeylessBackupProgress.tsx
```diff
@@ -269,6 +269,32 @@ function Restore() {
...
+  isOnboarding,
+  title,
+  subtitle,
+  children,
+}: {
+  isOnboarding: boolean
+  title: string
+  subtitle: string
+  step: number
+  totalSteps: number
```
satish-ravi, line 282: seems like these 2 are not used. Can we either drop these or drop title / subtitle and construct the header from it (probably more DRY)

MuckT: Removed the props in 5402d6da01b077fb88a9c3cab431dad4a2d4d657!

src/keylessBackup/KeylessBackupProgress.tsx
MuckT: Added the origin to the event in f54214c8af1b190924092e3eb80d056494b78e3f & updated the tests in 779f0172358734d92481ac22a6db0db25f54394e!

src/keylessBackup/KeylessBackupProgress.tsx
MuckT: Replaced instances of `navigatedFromSettings` with `isOnboarding` and swapped the tertiaries.

satish-ravi (approved):

PR #5616, opened 2024-07-05
chore(points): points activities list redesign

bakoushin:
### Description

Implement new points activities list design.

Figma: https://www.figma.com/design/rXBDplfMHHqYmuu6EkgMEo/Gamification-experiments?node-id=1747-6464&t=aOfd2uHuL1gAJfTG-0

Important changes:
1. Activities are rendered as a list
2. No more grouping by points amount; the amount is indicated on the card
3. Activities can have previous amount crossed out
4. "More coming soon" card dividing incomplete and completed activities
5. "More coming soon" card has a new icon (rushing clock instead of rocket)

Implementation notes:
* `ActivityCard` is refactored as a "dumb" component only relying on its props
* Added a specific `CheckCircle` icon instead of a mix of another icon and css styling used previously

| New design | Previous design |
|--------|--------|
| new design | prev design | 

Crossed out previous amount:

| Card | Bottom sheet |
|--------|--------|
| previous amount | image | 

### Test plan

* Tested manually
* Updated unit tests

### Related issues

- Fixes RET-1103

### Backwards compatibility

Y

### Network scalability


src/icons/RushingClock.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
```
bakoushin, line 1: will use the one from #5618 instead once it is merged

src/points/ActivityCardSection.tsx
```diff
@@ -1,132 +1,122 @@
...
+        {incompleteActivities.map(renderActivityCard)}
+        <MoreComingCard />
+        {completedActivities.map(renderActivityCard)}
       </View>
-      {pointsSections.map(makeSection)}
     </View>
   )
 }
 
+function MoreComingCard() {
```
kathaypacific, line 77: this is fine but i wonder if it'd be easier to have this in ActivityCard so that it can share styles?

src/points/cardSort.ts
```diff
@@ -0,0 +1,18 @@
+import { Props as ActivityCardProps } from 'src/points/ActivityCard'
```
kathaypacific, line 1: all good here but it made me think that perhaps we should return an order from the getPointsConfig endpoint instead in the future...

src/points/selectors.ts
```diff
@@ -31,41 +31,13 @@ export const pointsConfigStatusSelector = (state: RootState) => state.points.poi
...
-      if (a.pointsAmount > b.pointsAmount) return -1
-      return 0
-    })
-  }
-)
+export const pointsActivitiesSelector = createSelector([pointsConfigSelector], (pointsConfig) => {
+  return Object.entries(pointsConfig.activitiesById).map(([activityId, metadata]) => ({
+    ...metadata,
+    activityId,
+    completed: activityId === 'create-wallet',
```
kathaypacific, line 38: should we add the `trackOnceActivitiesSelector` to this selector and use that for this `completed` condition? it doesn't matter so much here but it seems like something we could overlook when adding the next binary activity

src/icons/RushingClock.tsx
kathaypacific: i think you can keep this since it's already here and this PR will be merged before the next release

src/points/ActivityCard.test.tsx
```diff
@@ -0,0 +1,59 @@
+import { fireEvent, render } from '@testing-library/react-native'
+import React from 'react'
+import { PointsEvents } from 'src/analytics/Events'
+import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
+import SwapArrows from 'src/icons/SwapArrows'
+import ActivityCard, { Props } from './ActivityCard'
+
+jest.mock('src/analytics/ValoraAnalytics')
+
+describe('ActivityCard', () => {
```
kathaypacific, line 10: i think you can add a `jest.clearAllMocks()` in a `beforeEach` since you check whether mocks have been called in the tests below (without this, the assertions could be on calls to that mock from previous tests)

src/points/ActivityCardSection.tsx
```diff
@@ -1,132 +1,122 @@
...
+        // To catch any missing cases at compile time
+        const assertNever: never = activity.activityId
+        return assertNever
     }
+  }
 
-    if (!cards.length) {
-      return <View key={pointsAmount}></View>
-    }
+  const preparedActivities = pointsActivities.map(mapActivityToCardProps).sort(sortByAmountAndTitle)
```
kathaypacific, line 59:
```suggestion
  const sortedActivities = pointsActivities.map(mapActivityToCardProps).sort(sortByAmountAndTitle)
```
nit: "prepared" makes me think too much of all our prepared transactions stuff....

src/points/PointsHome.test.tsx
```diff
@@ -124,17 +124,12 @@ describe(PointsHome, () => {
   })
 
   it('renders multiple sections', async () => {
-    const { getByTestId, queryByTestId, queryByText } = renderPointsHome()
+    const { queryByText } = renderPointsHome()
 
-    expect(getByTestId('PointsActivitySection-50')).toBeTruthy()
-    expect(getByTestId('PointsActivitySection-20')).toBeTruthy()
+    expect(queryByText('points.activityCards.createWallet.title')).toBeTruthy()
```
kathaypacific, line 129:
```suggestion
    expect(getByText('points.activityCards.createWallet.title')).toBeTruthy()
```
you get better error messages with `getBy*` and truthy assertions if the test fails :D

src/points/cardSort.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export function sortByAmountAndTitle(a: ActivityCardProps, b: ActivityCardProps) {
```
kathaypacific, line 3: nit: i wonder if we should avoid mutations, and therefore have the .sort inside this function rather than on the caller. i think generally we want to avoid mutations to keep things predictable

src/points/ActivityCardSection.tsx
kathaypacific: one thing that was nice about having this sorting in the selector is that we get memoization for free. of course the config is so small here that it doesn't really matter. but i wonder if we should do a useMemo here anyway...

kathaypacific (approved):
Nice job!! I had a few minor suggestions but it looks great!!

src/points/ActivityCardSection.tsx
bakoushin: despite looking like an ActivityCard it behaves quite differently:
* is unable to tap
* doesn't have a `PointActivityId` to track

so instead of branching to handle those specifics in the ActivityCard, I opted for this separate component

kathaypacific: yes totally - i just wondered if it'd be better for it to live in the ActivityCard file, so that it can share the font and spacing styles with the ActivityCard. I agree that it shouldn't be the same component

bakoushin: ah, sure! great idea!

src/points/cardSort.ts
bakoushin: oh, it would be great to get the list already properly sorted from the config!

one challenge I'm thinking about is how to sort activities alphabetically in different languages. alphabetical sort is needed when we have multiple activities of the same points value. 

or we can ask the marketing team to define the order 😅

src/points/selectors.ts
bakoushin: absolutely! thanks for the suggestion.

src/points/ActivityCard.test.tsx
bakoushin: great catch, thanks!

src/points/ActivityCardSection.tsx
bakoushin: yes, sure!

src/points/cardSort.ts
bakoushin: hmm, this function is a comparator, it does not do any sorting. I will update the naming.

when we are actually do sorting in place using this comparator, we are sorting the result of the `.map()` function, which I believe is safe to mutate in that case:
https://github.com/valora-inc/wallet/blob/331b6fc2115e158ffe88316c0f215953c77af776/src/points/ActivityCardSection.tsx#L59

PR #5617, opened 2024-07-05
fix(points): ensure sagas are spawned on statsig refresh

kathaypacific:
### Description

[Context](https://valora-app.slack.com/archives/C04B61SJ6DS/p1720133654144139?thread_ts=1720105286.269059&cid=C04B61SJ6DS).

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

n/a

### Network scalability

### DIFF
--- a/src/points/saga.test.ts
+++ b/src/points/saga.test.ts
@@ -35,6 +35,8 @@ import pointsReducer, {
   trackPointsEvent,
 } from 'src/points/slice'
 import { ClaimHistory, GetHistoryResponse, PointsEvent } from 'src/points/types'
+import { getFeatureGate } from 'src/statsig'
+import { StatsigFeatureGates } from 'src/statsig/types'
 import { NetworkId } from 'src/transactions/types'
 import Logger from 'src/utils/Logger'
 import * as fetchWithTimeout from 'src/utils/fetchWithTimeout'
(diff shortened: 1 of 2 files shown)


bakoushin:
should this fix the issue Jean complained about?

kathaypacific:
> should this fix the issue Jean complained about?

yes i think so!

bakoushin:
now the sagas are spawned no matter what the statsig value is, but despite we [are waiting until the address exists](https://github.com/valora-inc/wallet/blob/d5919782e0d3cc9b88c7a7a74d828d6120ea0102/src/points/saga.ts#L316) it is not enough time and we have to wait (or listen) to when the statsig will finish its update

bakoushin:
I've put some console.logs:
```
[13:51:04.623Z] [90miPhone 15[39m --- WAITING FOR ADDRESS TO EXIST
[13:51:25.493Z] [90miPhone 15[39m --- ADDRESS EXISTS
[13:51:25.495Z] [90miPhone 15[39m --- GETTING POINTS CONFIG {"showPoints": false}
```
notice it tries to get statsig config immediately after it gets an address. i guess it is too early. we have to wait for statsig to react to address update. i wonder if we have any way to handle it.

bakoushin:
or maybe we should watch some event other than `APP_INIT` to spawn initial sagas getting config and balance

kathaypacific:
TY for the ideas @bakoushin ! i clearly didn't test properly before 😅 i also was thinking about triggering the sagas on a different event because the coordination of statsig and account initialisation was feeling too hard....

bakoushin (approved):
🚀 thanks for fixing it!

PR #5625, opened 2024-07-09
chore: update discover dapp and dapp filter styles

kathaypacific:
### Description

[Context](https://valora-app.slack.com/archives/C0684TXDR3K/p1719510075150019) and [design](https://www.figma.com/design/FwRyq9HaKtZgywKVmIg0Zh/Discover-Page?node-id=1-3&t=uHdxyczMjeuYYBO2-0).

This PR only has styles changes. Also fixes the dapp filters appearing on more than 1 row. This PR breaks the dapp rankings styles but that is disabled and earmarked for deletion anyway. 

### Test plan

Simulator Screenshot - iPhone 15 Pro - 2024-07-09 at 12 16 46

https://github.com/valora-inc/wallet/assets/20150449/7f1cd23b-8e5f-497a-88e9-174ad7afdf5a

### Related issues

n/a

### Backwards compatibility

yes

### Network scalability

### DIFF
--- a/src/components/BorderlessButton.tsx
+++ b/src/components/BorderlessButton.tsx
@@ -12,7 +12,7 @@ export type Props = Omit<TouchableProps, 'style'> & {
 export default function BorderlessButton(props: Props) {
   const { style, containerStyle, children, notScaleFont, ...passThroughProps } = props
   return (
-    <Touchable borderless={true} {...passThroughProps} style={containerStyle}>
+    <Touchable {...passThroughProps} borderless={true} style={containerStyle}>
       {notScaleFont ? (
         <Text allowFontScaling={false} style={style}>
           {children}
--- a/src/components/BorderlessButton.tsx
+++ b/src/components/BorderlessButton.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react'
 import { StyleProp, Text, TextStyle, ViewStyle } from 'react-native'
 import Touchable, { Props as TouchableProps } from 'src/components/Touchable'
-import variables from 'src/styles/variables'
 
 export type Props = Omit<TouchableProps, 'style'> & {
   style?: StyleProp<TextStyle>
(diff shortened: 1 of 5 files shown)


src/components/BorderlessButton.tsx
```diff
@@ -12,7 +13,12 @@ export type Props = Omit<TouchableProps, 'style'> & {
 export default function BorderlessButton(props: Props) {
   const { style, containerStyle, children, notScaleFont, ...passThroughProps } = props
   return (
-    <Touchable {...passThroughProps} borderless={true} style={containerStyle}>
+    <Touchable
+      {...passThroughProps}
+      borderless={true}
+      style={containerStyle}
+      hitSlop={variables.iconHitslop}
```
kathaypacific, line 20: i considered adding it to the "explore dapps" text only, but seems like a hitslop is always a good idea to make the touch target bigger

src/dappsExplorer/DappCard.tsx
```diff
@@ -98,22 +91,28 @@ function DappCard({
...
-      style={[styles.card, showBorder ? styles.borderStyle : {}]}
+      style={[styles.card, showBorder ? styles.borderStyle : {}, cardStyle]}
       rounded={true}
       shadow={null}
     >
-      <Touchable onPress={onPressDapp} borderRadius={8} testID={`Dapp/${dapp.id}`}>
+      <Touchable
+        style={[styles.pressableCard, cardContentContainerStyle]}
+        onPress={onPressDapp}
+        borderRadius={showBorder ? 8 : undefined}
```
jophish, line 108: nit: can 8 be `Spacing.Smallest8`?

jophish (approved):
Thank you for this, and sorry that this slipped through!! This is a lot cleaner, and the approach to styling the `DappCard` makes much more sense here :sweat_smile:

src/dappsExplorer/DappCard.tsx
kathaypacific: i prefer not to 🙈 i agree that we prefer variables over hardcoded numbers so that we are restricted to certain consistent values + it's easier to make a global change for the values, but i think borders and spaces may not want to use the same variables...it can be a surprise if a change to "Spacing" values also changes rounded corners. i've always wanted to create a separate style enum for borders but never got round to it 😅

PR #5626, opened 2024-07-09
feat(swap): update swap info UI and add transaction time for cross chain

kathaypacific:
### Description

This PR implements part of [this design](https://www.figma.com/design/fjpfHZdVSLazv330kT2sQz/Cross-chain-swaps?node-id=606-2259&t=7bMhTAB8SD1gWv6z-0). The fees part of this design is not included in this PR.
- updated UI on the swaps screen
- updated bottom sheet styles
- added estimated transaction time for cross chain quotes

### Test plan

https://github.com/valora-inc/wallet/assets/20150449/76c7ca36-ee66-4744-bfd6-d3c6b1787d1c

### Related issues

- Related to RET-1066

### Network scalability


src/swap/SwapScreen.tsx
```diff
@@ -834,6 +836,9 @@ export function SwapScreen({ route }: Props) {
             fetchingSwapQuote={quoteUpdatePending}
             appFee={appFee}
             appFeeInfoBottomSheetRef={appFeeInfoBottomSheetRef}
+            estimatedDurationInSeconds={
+              quote && quote.swapType === 'cross-chain' ? quote.estimatedDuration : undefined
```
jeanregisser, line 840: nit: code golf
```suggestion
              quote?.swapType === 'cross-chain' ? quote.estimatedDuration : undefined
```

jeanregisser, line 840: nit: Maybe we could name it `estimatedDurationInSeconds`?

jeanregisser (approved):
Great! 👍

PR #5631, opened 2024-07-09
chore(points): Reduce trackPointsEvent usage

jophish:
### Description

From discussions [here](https://github.com/valora-inc/wallet/pull/5605#discussion_r1663938104) and [here](https://github.com/valora-inc/wallet/pull/5294/files#r1575826469), we should [avoid sequential dispatch of actions](https://redux.js.org/style-guide/#avoid-dispatching-many-actions-sequentially). This PR attempts to restructure when we call `trackPointsEvent`, namely, not calling it when we _already_ emit other events that we can subscribe to that could contain points event metadata. For now this mainly changes how we deal with swap events, but will also apply to live link events as well.

This does _not_ eliminate all uses of `trackPointsEvent`, since we still have one use case where we want to track a swap event when no other event would otherwise be emitted, namely, when [updating standby transactions](https://github.com/valora-inc/wallet/blob/main/src/transactions/saga.ts#L280).

I'm not sure if this is a better approach, or if it makes sense to require a complete `pointsEvent` object on each payload we watch in the points saga.

### Test plan

Unit and manual tested.

### Related issues

N/A

### Backwards compatibility

Yes.

### Network scalability


kathaypacific (commented):
great! this works, although i have some feelings about other actions needing to know about points because it doesn't feel like clean separation of responsibilities.

one approach i was thinking is for transaction actions (e.g. swapSuccess) to have the transactionHash added to the action params. then in the points saga, we could have a watcher that watches all the points events that we care about...something like:
```
const swapTrackedProps = { activityId: 'swap', ...someotherstuff}
takeLeading(swapSuccess, (action) => trackPointsEventSaga({...action, swapTrackedProps}))
```
i'm not sure how to get the syntax correct for the second line but i think there will be a way where we can essentially take the action that we care about and add extra points params, and then call the track event saga. the benefit i see here is that we don't need to have points logic in more places than in the points folders

jophish:
@kathaypacific Yeah, I also had reservations about including the entire `pointsEvent` object in each action payload; it feels like we're just tacking it on. The thing is, for "transaction actions" (successful swap, successful live link send, earn deposit), the `trackEvent` endpoint payload may require more than just the transaction hash. We can solve this by just adding those particular additional fields to whatever action we already dispatch, and handling them one-by-one in the points saga, extracting the fields as needed and constructing the `pointsEvent` object before passing it along to the `sendPointsEvent` saga.

Does this sound like a reasonable plan?

kathaypacific:
> @kathaypacific Yeah, I also had reservations about including the entire `pointsEvent` object in each action payload; it feels like we're just tacking it on. The thing is, for "transaction actions" (successful swap, successful live link send, earn deposit), the `trackEvent` endpoint payload may require more than just the transaction hash. We can solve this by just adding those particular additional fields to whatever action we already dispatch, and handling them one-by-one in the points saga, extracting the fields as needed and constructing the `pointsEvent` object before passing it along to the `sendPointsEvent` saga.
> 
> Does this sound like a reasonable plan?

yes sounds good, i think we are on the same page! i think the approach of isolating the points logic is more easy to maintain in the long run, in case we need to make any changes or refactors

PR #5636, opened 2024-07-11
feat(swap): add total swap fee on swap screen

kathaypacific:
### Description

This PR includes:
- adding the total "fees" line item on the swap screen details section
- some refactoring to get the network fees / app fees / cross chain fee calculation functions to return the same object schema
- a custom function that creates this total fee string, since it is the summation of potentially 3 different tokens
- skeleton loader for the swap details during quote fetch
- don't render the details section on initial screen mount (this is in the [design](https://www.figma.com/design/fjpfHZdVSLazv330kT2sQz/Cross-chain-swaps?node-id=606-1607&t=5nusKUmZoYSiB3Z9-0))

To keep the PR small, the info icon press + fees info bottom sheet will be in a separate PR.

### Test plan

with font scaling, the values should be right aligned and wrapped to multiple lines
Simulator Screenshot - iPhone 15 Pro - 2024-07-11 at 10 40 11

https://github.com/valora-inc/wallet/assets/20150449/81b65758-3452-434e-950c-02ad943387a1

### Related issues

- Related to RET-1066

### Backwards compatibility

Y

### Network scalability


jeanregisser (approved):
Awesome! Grouping fees refactor is much better! 🚀 

Perhaps, the skeleton loaders shouldn't be as wide in the fees?
Screenshot 2024-07-12 at 10 47 19

kathaypacific:
> Awesome! Grouping fees refactor is much better! 🚀
> 
> Perhaps, the skeleton loaders shouldn't be as wide in the fees? Screenshot 2024-07-12 at 10 47 19

yeah i wasn't sure about that too, we can either cover the whole area where the text can be (usually too large) or cover the existing text (sometimes too small, if the previous fees were $1.00 and the next quote is $0.90 + 0.1 BNB). I guess the second one looks better, i've modified it to look like this:

https://github.com/user-attachments/assets/802842c2-b1a4-4ad6-9521-b37dc8af8445

PR #5637, opened 2024-07-11
feat(swap): add fees info bottom sheet

kathaypacific:
### Description

This PR adds the fees bottom sheet.

### Test plan

https://github.com/valora-inc/wallet/assets/20150449/75e0842a-eb1b-48da-89ff-88e08cd6c5aa

With font scaling
Simulator Screenshot - iPhone 15 Pro - 2024-07-11 at 14 29 22

### Related issues

- Fixes RET-1066

### Backwards compatibility

Y

### Network scalability


src/swap/FeeInfoBottomSheet.tsx
```diff
@@ -0,0 +1,199 @@
...
+              />
+              {` (`}
+              <TokenDisplay
+                amount={displayAmount}
+                tokenId={fee.token.tokenId}
+                showLocalAmount={false}
+                style={styles.bodyText}
+                errorFallback=""
+              />
+              {`)`}
```
jeanregisser, line 62: We're probably copy/pasting this, but we should avoid concatenating strings, it makes localization harder.
We can use a `<Trans>` component.

jeanregisser (approved):
Amazing! 🙌

src/swap/FeeInfoBottomSheet.tsx
kathaypacific: updated!

src/swap/FeeInfoBottomSheet.tsx
```diff
@@ -0,0 +1,199 @@
...
+  const displayAmount = showMaxAmount ? fee?.maxAmount : fee?.amount
+  const isFeeZero = displayAmount && displayAmount.isZero()
+  const loadingText = t('loading')
+  const fallbackText = t('unknown')
+  const zeroFeeText = t('swapScreen.transactionDetails.swapFeeWaived')
+
+  return (
+    <Text style={styles.feeAmountText}>
+      {isLoading ? (
+        loadingText
```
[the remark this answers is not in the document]
kathaypacific: i spent >30 minutes on this but couldn't get the styles working for some reason....i think it's a good idea but i also think that the chances of users seeing this state is small (probably only likely if they are waiting a long time for a quote, but even so it may not be likely that they want info about the fees during the wait). so for now i'm giving up on it but maybe i'll return to it at some point 😅

PR #5638, opened 2024-07-11
chore: remove dapp rankings

kathaypacific:
### Description

As the title

### Test plan

n/a

### Related issues

- Fixes RET-921

### Backwards compatibility

Y

### Network scalability


src/dappsExplorer/DappCard.tsx
```diff
@@ -36,12 +29,16 @@ interface Props {
 // Since this icon exists within a touchable, make the hitslop bigger than usual
 const favoriteIconHitslop = { top: 20, right: 20, bottom: 20, left: 20 }
 
-export function DappCardContent({
```
kathaypacific, line 39: there's no need for this strange abstraction anymore so i removed

jeanregisser (approved):
👍

PR #5639, opened 2024-07-11
chore: remove `as Address` casting as much as possible

kathaypacific:
### Description

This PR removes as much `as Address` casts as possible.
- update our redux types to use `Address` rather than `string` for addresses, e.g. tokens and wallet address
- i added a redux migration to lower case the stored `web3.account` and remove the `rawWalletAddressSelector`, it seems like this is needed for users who have had valora on their devices since before 2020 without restoring the app. it seems low impact to also update the analytics user property to this lower cased address

### Test plan

n/a

### Related issues

- Fixes RET-952

### Backwards compatibility

Y

### Network scalability


package.json
```diff
@@ -299,7 +299,7 @@
     "ts-retry-promise": "^0.6.0",
     "twilio": "^3.69.0",
     "typescript": "^5.4.3",
-    "typescript-json-schema": "^0.59.0",
+    "typescript-json-schema": "^0.64.0",
```
kathaypacific, line 302: i upgraded this because after updating some redux variables from `string` to `Address`, the inferred type in the schema became a string array (and this causes the `store` tests to fail, because the values are not actually arrays)

kathaypacific:
giving up on this, it's too much work

PR #5642, opened 2024-07-12
feat(onboarding): Recovery phrase screen update

finnian0826:
### Description

Update copy and styles to match designs.

Checked with design and confirmed that we should not have the "Why is this important?" line, and rather just keep the old help button + bottom sheet.

### Test plan

Simulator Screenshot - iPhone 14 Pro - 2024-07-12 at 14 46 45

### Related issues

- Fixes #ACT-1221

### Backwards compatibility

Yes

### Network scalability


src/onboarding/registration/OnboardingRecoveryPhrase.tsx
MuckT, line 1: The continue button and bottomSection should be outside of the scroll view to match the other onboarding screens.

PR #5643, opened 2024-07-15
feat(points): Reduce trackPointsEvent usage

jophish:
### Description

Taking another crack at #5631. Updated following a [discussion](https://github.com/valora-inc/wallet/pull/5631#pullrequestreview-2168324821) with @kathaypacific.

### Test plan

Unit and manual tested; points are still earned for performing expected tasks.

### Related issues

- Fixes #[issue number here]

### Network scalability

### DIFF
--- a/src/points/saga.ts
+++ b/src/points/saga.ts
@@ -1,6 +1,7 @@
 import { differenceInDays } from 'date-fns'
 import { isEqual } from 'lodash'
 import { Actions as HomeActions } from 'src/home/actions'
+import { PayloadAction } from '@reduxjs/toolkit'
 import { retrieveSignedMessage } from 'src/pincode/authentication'
 import {
   nextPageUrlSelector,
@@ -39,6 +40,8 @@ import networkConfig from 'src/web3/networkConfig'
 import { walletAddressSelector } from 'src/web3/selectors'
 import { call, put, select, spawn, take, takeEvery, takeLeading } from 'typed-redux-saga'
 import { v4 as uuidv4 } from 'uuid'
+import { depositTransactionSucceeded, DepositTransactionSucceededAction } from 'src/jumpstart/slice'
+import { swapSuccess, SwapResult } from 'src/swap/slice'
 
 const TAG = 'Points/saga'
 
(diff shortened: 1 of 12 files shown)


src/points/saga.ts
```diff
@@ -296,21 +298,57 @@ export function* sendPendingPointsEvents() {
   }
 }
 
-function* watchGetHistory() {
+export function* watchSwapSuccessTransformPayload({
```
jophish, line 301: This pattern is really just for testing purposes, and I don't really like it... I was able to inline this (see [here](https://github.com/valora-inc/wallet/pull/5643/commits/77b808b5bde5466f5a7a9c53655bc9c3c56718aa#diff-4187b1512dcd56a9389e0b7054cdb54b30a8e632563003e5c7068bae36b817adR317)), though I couldn't get the expect-saga matchers to match the `sendPointsEvent` call, likely related to the inlining.

kathaypacific (approved):
looks good to me! 

one thought that came to mind is that we're not really wanting to use the `trackPointsEvent` action anymore, so maybe we can remove it? it simplifies some stuff:
- no need to read the code comment about not using it (easy to miss during a review if that file isn't changed)
- the `sendPointsEvent` saga can take `PointsEvent` as the input type

the benefit of the the second point is the watchers could perhaps be simplified to one saga like 
```
function* watchPointsActivities() {
  yield* takeLeading(
    swapSuccess.type,
    safely(function* (action: ReturnType<typeof swapSuccess>) {
      yield* call(sendPointsEvent, { ...action.payload, activityId: 'swap' as const })
    })
  )
  yield* takeLeading(...)
}
```

jophish:
@kathaypacific We actually do still use `trackPointsEvent` in one place atm, when [fetching transaction receipts](https://github.com/valora-inc/wallet/blob/main/src/transactions/saga.ts#L280) when watching pending transactions. We do this since we don't already emit an event here. We could of course avoid this by emitting a _new_ different event, under e.g., `src/transactions/actions.ts` that has no knowledge of points semantics. We'd have to do this for every such case however, so I don't know if it would necessarily be better than the current solution.

PR #5649, opened 2024-07-17
feat(earn): update earn discover card for multiple pools

satish-ravi:
### Description

Sets up the statsig gate and shows the earn entrypoint card on the discover tab. Only changes the card for when there are no active pools. 
Also refactors the EarnInfoScreen to not take tokenId as a param because it is a generic screen.

[Figma](https://www.figma.com/design/E1rC3MG74qEg5V4tvbeUnU/Stablecoin-Enablement?node-id=4079-14115&t=VKwjwdsC2mZD4NcH-0)

### Related issues

- Part of ACT-1257

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/src/earn/EarnCard.tsx
+++ b/src/earn/EarnCard.tsx
@@ -3,6 +3,7 @@ import { View } from 'react-native'
 import ItemSeparator from 'src/components/ItemSeparator'
 import EarnActivePool from 'src/earn/EarnActivePool'
 import EarnCta from 'src/earn/EarnCta'
+import EarnEntrypoint from 'src/earn/EarnEntrypoint'
 import { getFeatureGate } from 'src/statsig'
 import { StatsigFeatureGates } from 'src/statsig/types'
 import { Spacing } from 'src/styles/styles'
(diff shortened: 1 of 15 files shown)


src/earn/EarnCard.tsx
```diff
@@ -14,9 +15,16 @@ interface Props {
...
-  const showStablecoinEarn = getFeatureGate(StatsigFeatureGates.SHOW_STABLECOIN_EARN)
+  const showMultiplePools = getFeatureGate(StatsigFeatureGates.SHOW_MULTIPLE_EARN_POOLS)
   const poolToken = useTokenInfo(poolTokenId)
 
+  if (showMultiplePools) {
+    // TODO(ACT-1257): show active pools
+    return <EarnEntrypoint />
+  }
+
+  const showStablecoinEarn = getFeatureGate(StatsigFeatureGates.SHOW_STABLECOIN_EARN)
```
finnian0826, line 26: Is `showStablecoinEarn` only going to be used to refer to the MVP now? I'm just wondering if down the road there could be some case where `showStablecoinEarn` is false but `showMultiplePools` is true and if that would have any issues

satish-ravi: yes that's an MVP specific gate. If we wanted to combine gates, we can always do it from statsig, instead of enforcing in code

finnian0826 (approved):

PR #5650, opened 2024-07-18
feat(earn): add active pools card for multiple pools

satish-ravi:
### Description

[Figma](https://www.figma.com/design/E1rC3MG74qEg5V4tvbeUnU/Stablecoin-Enablement?node-id=4079-14167&t=0UIxbMax1NwBgdFf-0)

### Related issues

- Fixes ACT-1257

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/src/earn/EarnActivePools.tsx
+++ b/src/earn/EarnActivePools.tsx
@@ -30,7 +30,7 @@ export default function EarnActivePools() {
     return null
   }
   const poolsSupplied = 1
-  const totalSupplied = `${localCurrencySymbol} ${poolTokenInLocalCurrency ? formatValueToDisplay(poolTokenInLocalCurrency) : '--'}`
+  const totalSupplied = `${localCurrencySymbol}${poolTokenInLocalCurrency ? formatValueToDisplay(poolTokenInLocalCurrency) : '--'}`
 
   return (
     <View style={styles.card} testID="EarnActivePools">
--- a/src/earn/EarnActivePools.tsx
+++ b/src/earn/EarnActivePools.tsx
@@ -3,7 +3,7 @@ import { useTranslation } from 'react-i18next'
 import { StyleSheet, Text, View } from 'react-native'
 import { EarnEvents } from 'src/analytics/Events'
 import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
-import Button, { BtnSizes, BtnTypes } from 'src/components/Button'
+import Button, { BtnSizes, BtnTypes, TextSizes } from 'src/components/Button'
 import { formatValueToDisplay } from 'src/components/TokenDisplay'
 import { useDollarsToLocalAmount } from 'src/localCurrency/hooks'
 import { getLocalCurrencySymbol } from 'src/localCurrency/selectors'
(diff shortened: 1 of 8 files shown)


src/components/Button.tsx
```diff
@@ -55,6 +65,7 @@ export default React.memo(function Button(props: ButtonProps) {
     loadingColor,
     touchableStyle,
     iconMargin = 4,
+    fontStyle = styles.fontStyle,
```
satish-ravi, line 68: Confirming with design on this , but all buttons in the earn flow require buttons with the small font size instead of medium. 
We could also make a TextSizes enum similar to BtnSizes, so that way consumers only have limited ways of tweaking the style. Thoughts?

src/earn/EarnActivePools.tsx
```diff
@@ -0,0 +1,124 @@
...
+import networkConfig from 'src/web3/networkConfig'
+
+const TAG = 'earn/EarnActivePools'
+
+export default function EarnActivePools() {
+  const { t } = useTranslation()
+  const localCurrencySymbol = useSelector(getLocalCurrencySymbol)
+
+  // TODO(ACT-1268): use info from getEarnPositions
+  const poolToken = useTokenInfo(networkConfig.aaveArbUsdcTokenId)
```
satish-ravi, line 25: using the network config here instead of getting this as props from EarnCard because this is only temporary.

src/components/Button.tsx
jh2oman: I think yes to TextSizes enum. These Button components are easy to work with. lets keep it that way

src/earn/EarnCard.tsx
```diff
@@ -19,8 +20,8 @@ export function EarnCardDiscover({ depositTokenId, poolTokenId }: Props) {
   const poolToken = useTokenInfo(poolTokenId)
```
finnian0826, line 20: I'm a bit confused what this would be in the multi-pool case. Should the props to this function be refactored? It seems like we would need to check if any of the pool tokens have a balance to decide to show `EarnActivePools`.

src/earn/EarnActivePools.tsx
```diff
@@ -0,0 +1,124 @@
...
+          <View style={styles.row} testID="EarnActivePools/TotalSupplied">
+            <Text style={styles.labelText}>{t('earnFlow.activePools.totalSupplied')}</Text>
+            <Text style={styles.valueText}>{totalSupplied}</Text>
+          </View>
+        </View>
+        <View style={styles.buttonContainer}>
+          <Button
+            onPress={() => {
+              ValoraAnalytics.track(EarnEvents.earn_active_pools_cta_press, {
+                action: 'exploreOpenPools',
```
finnian0826, line 54: should these actions be pulled out into an enum?

src/earn/EarnActivePools.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+const TAG = 'earn/EarnActivePools'
+
+export default function EarnActivePools() {
+  const { t } = useTranslation()
+  const localCurrencySymbol = useSelector(getLocalCurrencySymbol)
+
+  // TODO(ACT-1268): use info from getEarnPositions
+  const poolToken = useTokenInfo(networkConfig.aaveArbUsdcTokenId)
+  const poolTokenInLocalCurrency = useDollarsToLocalAmount(poolToken?.balance ?? null)
```
finnian0826, line 26: Is poolToken.balance in USD? Should we convert to USD first before doing this?

src/earn/EarnCard.tsx
satish-ravi: For multi pool, this would just be based on the `getEarnPools` output. The tokenId props would just be for the earn MVP. We can remove those when we remove MVP specific code.

src/earn/EarnActivePools.tsx
satish-ravi: I think string literals are fine, we talked about enums vs string literals in one of the eng syncs and I don't think we landed on using only one over the other. Enums seem like an overkill for this case

src/earn/EarnActivePools.tsx
satish-ravi: good catch. It technically should be since its a stablecoin, but could be off by a bit.

finnian0826 (approved):

PR #5658, opened 2024-07-19
fix(hooks): too many claim shortcuts displayed

jeanregisser:
### Description

With the switch of hooks to multichain, shortcuts needed to be filtered by `networkId`.
Also the displayed tokens needed an update to use `tokenId`.

### Test plan

| Before | After |
|-|-|
| image | image |

### Related issues

- Fixes RET-1108

### Network scalability


src/positions/selectors.ts
```diff
@@ -92,7 +92,10 @@ export const positionsWithClaimableRewardsSelector = createSelector(
   (positions, shortcuts, triggeredShortcuts) => {
     const claimablePositions: ClaimablePosition[] = []
     positions.forEach((position) => {
-      const appShortcuts = shortcuts.filter((shortcut) => shortcut.appId === position.appId)
+      const appShortcuts = shortcuts.filter(
+        (shortcut) =>
+          shortcut.appId === position.appId && shortcut.networkIds.includes(position.networkId)
```
jeanregisser, line 97: This is the main needed change to avoid duplicates by network.

satish-ravi (approved):

PR #5659, opened 2024-07-22
feat: allow `tokensByIdSelector` to return position tokens and count tokens once for total balance calculation

jeanregisser:
### Description

This contains the following changes:

- Update `tokensByIdSelector`:
  - When a regular token has no `priceUsd`, we update it with the value sourced from position tokens if we have it
    - This allows showing $ price for complex ERC20 tokens when the position can be decomposed via position hooks.
  - Include tokens from positions when `includePositionTokens: true` is passed
    - `useTokenInfo` and `getTokenInfo` do that now. This shouldn't affect existing behaviors though.
    - This allows `TokenDisplay` to display tokens from positions
- Tokens/positions are only counted once when calculating total balances:
  - The total tokens balance ignores tokens sourced from positions
  - The total positions balance ignores tokens present in the regular list

Note: tokens from positions still can't be seen in the "Tokens" tab, or sent or swapped, unless explicitly added to address-metadata (or manually imported). We can decide later whether or not we want to enable that.

More generally, existing token selectors still return the same data (except with $ price coming from position if we have it), and only places that call `tokensByIdSelector` passing `includePositionTokens: true` get position tokens too.

This was ported from https://github.com/valora-inc/troopo/pull/17

### Test plan

- Updated tests

### Related issues

- Fixes RET-1113
- Fixes RET-1114

### Backwards compatibility

Yes

### Network scalability


src/tokens/selectors.ts
```diff
@@ -42,21 +43,97 @@ const DEFAULT_MEMOIZE_MAX_SIZE = 10
...
     (_state: RootState, networkIds: NetworkId[]) => networkIds,
   ],
-  (storedBalances, networkIds) => {
+  (storedBalances, positionTokens, positionsFetchedAt, networkIds) => {
+    const allStoredBalances = { ...storedBalances }
+
+    // Enrich with position tokens
+    // This allows us to have priceUsd and balance for tokens
+    // decomposed via positions
+    for (const positionToken of Object.values(positionTokens)) {
```
MuckT, line 103: Could we destructure here and adjust the allStoredBalances[tokenId] values to use the new consts? This avoids the `tokenId` and `priceUsd` consts.

```TypeScript
for (const {
  priceUsd,
  tokenId,
  address,
  networkId,
  decimals,
  symbol,
  balance,
} of Object.values(positionTokens)) {
```

src/tokens/selectors.ts
```diff
@@ -42,21 +43,97 @@ const DEFAULT_MEMOIZE_MAX_SIZE = 10
...
+      if (!networkIds.includes(positionToken.networkId)) {
+        continue
+      }
+      const tokenId = positionToken.tokenId
+      const existingToken = allStoredBalances[tokenId]
+      const priceUsd = positionToken.priceUsd != '0' ? positionToken.priceUsd : undefined
+      if (!existingToken) {
```
satish-ravi, line 110: adding all position tokens here slightly makes me nervous 😅  since its a backwards incompatible change and its on the consumers to filter out these tokens as required. Right now I see there are filters in some selectors, but some others don't seem to include it, like cash in / cash out token selectors, tokens with balance selector etc. Maybe these are fine for now, but wonder if down the line we'll run into some bugs, as its not super obvious from the name of the selector on whether it includes or excludes position tokens. Couple of ideas (not fully thought through):
- Make this change backwards compatible, e.g., don't include these by default and only include if requested by consumers. Maybe this is hard to do with selector parameters
- Have some naming pattern that makes it obvious that the selector includes or excludes position tokens.

What do you think?

jeanregisser: Great feedback, I was going back and forth about this.
I guess there are too parts here.

1. Injecting the $ price everywhere if not already present. I think this is the right place for this.
2. Adding all tokens from positions. I added it here for now so all places using `tokenId` could access position tokens (for instance `TokenDisplay` working out of the box with `tokenId` for positions). But the drawback is we don't want them in all selectors down the line. That's why I resorted to adding `isFromPosition`. Which isn't ideal like you mentioned. I tried to be careful, but it's possible it can cause subtle bugs. On a side note, Cash In and Cash Out token selectors don't need more filtering because they already check if `isCashInEligible` / `isCashOutEligible` is present. But it's possible I missed something else or that it will cause confusion in the future.

Anyway to be more on the safe side, I'll take a stab at making it a backward compatible change.

jeanregisser: Changed in https://github.com/valora-inc/wallet/pull/5659/commits/cb86d4322ea709e89bb828fa2f9ba4e44958ec37

Let me know what you think @satish-ravi 🙏

src/tokens/selectors.ts
jeanregisser: Yes we could, I didn't do it to make it a bit clearer as to what is coming from `positionToken` and what is `existingToken`.

src/tokens/selectors.ts
satish-ravi: This looks great! Thanks!

src/tokens/selectors.ts
```diff
@@ -42,21 +43,108 @@ const DEFAULT_MEMOIZE_MAX_SIZE = 10
...
+      for (const token of position.tokens) {
+        visitToken(token, false)
+      }
+    }
+  }
+  return positionTokens
+})
+
+type TokensByIdArgs =
+  | NetworkId[] // For backward compatibility
```
satish-ravi, line 91: we can eventually drop this as an option and we should be able to catch everything with TS errors right?

satish-ravi (approved):

src/tokens/selectors.ts
jeanregisser: yes exactly.

PR #5661, opened 2024-07-22
feat(analytics): add event for insufficient gas warnings

satish-ravi:
### Description

When preparing a tx in the various flows, if it fails with insufficient gas, add an analytics event. 

### Test plan

Unit tests, manual

### Related issues

- Fixes ACT-1256

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/src/earn/EarnCollectScreen.tsx
+++ b/src/earn/EarnCollectScreen.tsx
@@ -4,7 +4,7 @@ import React, { useEffect } from 'react'
 import { useTranslation } from 'react-i18next'
 import { SafeAreaView, ScrollView, StyleSheet, Text, View } from 'react-native'
 import SkeletonPlaceholder from 'react-native-skeleton-placeholder'
-import { EarnEvents } from 'src/analytics/Events'
+import { EarnEvents, TransactionEvents } from 'src/analytics/Events'
 import ValoraAnalytics from 'src/analytics/ValoraAnalytics'
 import Button, { BtnSizes } from 'src/components/Button'
 import InLineNotification, { NotificationVariant } from 'src/components/InLineNotification'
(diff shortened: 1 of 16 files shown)


src/earn/EarnCollectScreen.tsx
```diff
@@ -58,6 +58,19 @@ export default function EarnCollectScreen({ route }: Props) {
     depositTokenId,
     feeCurrencies,
   })
+
+  useEffect(() => {
+    if (
+      asyncPreparedTransactions.result?.type === 'need-decrease-spend-amount-for-gas' ||
+      asyncPreparedTransactions.result?.type === 'not-enough-balance-for-gas'
+    ) {
```
MuckT, line 66:
```suggestion
    if (
      ['need-decrease-spend-amount-for-gas', 'not-enough-balance-for-gas']
      .includes(asyncPreparedTransactions.result?.type)
    ) {
```

src/send/EnterAmount.test.tsx
```diff
@@ -126,6 +126,7 @@ const defaultParams = {
   prepareTransactionError: undefined,
   onPressProceed: onPressProceedSpy,
   ProceedComponent: SendProceed,
+  origin: 'send' as const,
```
MuckT, line 129: Is this type cast needed?

src/walletConnect/saga.test.ts
```diff
@@ -570,6 +572,64 @@ describe('showActionRequest', () => {
       preparedTransaction: mockPreparedTransactions.transactions[0],
       prepareTransactionErrorMessage: undefined,
     })
+    expect(ValoraAnalytics.track).toHaveBeenCalledTimes(1)
+    expect(ValoraAnalytics.track).toHaveBeenCalledWith(
+      WalletConnectEvents.wc_request_propose,
+      expect.anything()
```
MuckT, line 578: Is this because we could have an array of any number of transactions?

MuckT (approved):

src/send/EnterAmount.test.tsx
satish-ravi: yes, otherwise TS infers it as a generic string.

src/walletConnect/saga.test.ts
satish-ravi: this was just me being lazy, it wasn't super obvious what the expected props were, there were too many 😄

kathaypacific (commented):
without thinking deeply about how every screen works, it's hard to say if that useEffect will be rerun unexpectedly especially because there are multiple dependencies in the dependency array. e.g. you might have a prepared transaction that has insufficient gas, then change the token / network id (idk if possible) - if the network id is in the dep array then you're going to get another event triggered, even though the new tx hasn't been prepared yet...again, not sure if this is a real scenario / whether we care about duplicate events. 

did you consider dispatching the event in prepareTransactions directly? it seems easy to forget to add this useEffect block when implementing new screens, so it'd be nice to do this somewhere central

satish-ravi:
@kathaypacific good call, I think if network id changes, we clear the prepared tx but yeah there might be races. I originally added it per flow to have more control per flow on when to fire the event, but the behavior is same as firing from the prepareTransactions method. Made #5671 instead

PR #5668, opened 2024-07-23
feat(swap): add cross chain swap transaction type and store pending tx

kathaypacific:
### Description

This PR:
- add cross chain swap item to the transactions query. the existing infrastructure will store these fetched items in redux / remove any pending transactions of the same hash.
- add pending transaction of cross chain swap type on successful swap
- prevent the transaction watcher from marking the whole cross chain swap as successful
- add swap type (same or cross chain) to analytics events 
- render no UI for cross chain swap items 

### Test plan

Tested using redux debugger that the transactions are created as pending and updated when blockchain-api responses are recieved.

### Related issues

- Fixes RET-1120

### Backwards compatibility

Y

### Network scalability


bakoushin (approved):
✨

PR #5669, opened 2024-07-23
refactor(statsig): pull multichain features into its own function to decrease risk of crashes

jh2oman:
### Description

* Ensure at compile that developers cannot use `getDynamicConfigParams` to get multichain dynamic config
* Add in a new `getMultichainFeatures` function to retrieve the dynamic config for multichain and filter out unknown networkids

Almost all of the file changes are just updating `getDynamicConfigParams` to `getMultichainFeatures`

### Test plan

unit tests updated.

### Related issues

https://linear.app/valora/issue/ACT-1078/decrease-risk-of-wallet-crash-when-adding-new-networks

### Network scalability


src/statsig/index.ts
```diff
@@ -86,6 +88,32 @@ export function getDynamicConfigParams<T extends Record<string, StatsigParameter
   }
 }
 
+export function getMultichainFeatures() {
+  const defaultValues =
+    DynamicConfigs[StatsigMultiNetworkDynamicConfig.MULTI_CHAIN_FEATURES].defaultValues
+  const multichainParams = _getDynamicConfigParams({
+    configName: StatsigMultiNetworkDynamicConfig.MULTI_CHAIN_FEATURES,
+    defaultValues,
+  })
```
satish-ravi, line 97: does this not work?
```suggestion
  const multichainParams = _getDynamicConfigParams(
    DynamicConfigs[StatsigMultiNetworkDynamicConfig.MULTI_CHAIN_FEATURES]
  )
```

src/statsig/types.ts
```diff
@@ -10,6 +9,11 @@ export enum StatsigDynamicConfigs {
   EARN_STABLECOIN_CONFIG = 'earn_stablecoin_config',
 }
 
+// Separating into different enum from StatsigDynamicConfigs to allow for more strict typing
```
satish-ravi, line 12: you can also use `Exclude<StatsigDynamicConfigs, StatsigDynamicConfigs.MULTI_CHAIN_FEATURES>` instead of explicitly defining 2 enums

jh2oman: Exclude makes things a bit weirder because the output is a type not an enum

satish-ravi (approved):

PR #5670, opened 2024-07-23
feat(earn): Hook up and add analytics

finnian0826:
### Description

Add multi-pool analytics events, hook up EarnHome screen to cta components.

### Test plan

Unit tests updated

### Related issues

- Part of ACT-1260

### Backwards compatibility

Yes

### Network scalability


src/analytics/Events.tsx
```diff
@@ -705,4 +705,7 @@ export enum EarnEvents {
   earn_info_learn_press = 'earn_info_learn_press',
   earn_info_earn_press = 'earn_info_earn_press',
   earn_active_pools_cta_press = 'earn_active_pools_cta_press',
+  earn_home_learn_more_press = 'earn_home_learn_more_press',
+  earn_multi_pool_add_to_pool_press = 'earn_multi_pool_add_to_pool_press',
+  earn_multi_pool_exit_pool_press = 'earn_multi_pool_exit_pool_press',
```
satish-ravi, line 710: Can we just reuse the existing events for the single pool card instead of more new events w/ additional props you're adding below?  Guess when we launch, these might also be called deposit / withdraw at the time we launch, so we could name it something more generic if we want new events.

satish-ravi (commented):
Looks good, just a question on potentially reusing events

src/analytics/Events.tsx
finnian0826: I didn't re-use the old events because they use `EarnCommonProperties` which have 'aave-v3' hardcoded, and I would've needed to add the `poolId` prop which I think we would want to know which pool it was pressed for. I also don't think there is an event for first time adding to pool since in the single-pool case there is a cta press that takes you right to enter amount.

I can try and re-use `earn_deposit_more_press` and `earn_exit_pool_press` and modify `EarnCommonProperties` and then just have a new event for adding to an open pool, or I can just update these names to be more generic and add any extra props that we would want.

satish-ravi: eventually `EarnCommonProperties` would have to be modified to work with multiple pools, so don't see that as in issue. Actually, seems cleaner to start with new events since this is a new screen, but would be nice if we can just change names to be more generic, as it looks like we're gonna replace the "Add to pool" / "Exit pool" naming soon with the designs presented today. Maybe we can do `earn_pool_card_cta_press` with prop for deposit / withdraw + EarnCommonProperties. wdyt?

finnian0826: yeah I like that idea of just the single event. I think it would also make sense to have a balance property, that way we have that info for the withdraw event, and also it can let us differentiate between a user adding to a pool for the first time vs. depositing more

finnian0826:
Made a follow-up issue to update analytics across Earn for multipool https://linear.app/valora/issue/ACT-1319/update-earncommonproperties-to-work-for-multi-pool

src/earn/PoolCard.tsx
```diff
@@ -92,6 +95,14 @@ export default function PoolCard({ pool, testID = 'PoolCard' }: { pool: Pool; te
           <View style={styles.buttonContainer}>
             <Button
               onPress={() => {
+                ValoraAnalytics.track(EarnEvents.earn_pool_card_cta_press, {
+                  poolId: pool.poolId,
+                  depositTokenId,
+                  networkId: pool.networkId,
+                  tokenAmount: poolTokenInfo.balance.toString(),
+                  providerId: PROVIDER_ID,
```
satish-ravi, line 103: leave a note that this should be generalized. Pretty easy to miss when adding more pools. Or alternatively, change this to a more common string type in the analytics properties and get this from the pool config file for now (which can eventually be replaced by the hooks endpoint)

finnian0826: added TODOs linking to the analytics update follow-up ticket. Currently providerId isn't part of the `Pool` object, and so would have to update that object which would result in a lot of test files needing to be changed so left it for the follow-up ticket to make this cleaner.

src/earn/EarnCard.test.tsx
```diff
@@ -132,7 +132,7 @@ describe('EarnCardDiscover', () => {
     expect(queryByTestId('EarnCta')).toBeFalsy()
     expect(getFeatureGate).toHaveBeenCalledWith(StatsigFeatureGates.SHOW_MULTIPLE_EARN_POOLS)
     expect(getFeatureGate).toHaveBeenCalledWith(StatsigFeatureGates.SHOW_STABLECOIN_EARN)
-    expect(getFeatureGate).toHaveBeenCalledTimes(2)
+    expect(getFeatureGate).toHaveBeenCalledTimes(3)
```
satish-ravi, line 135: this seems unexpected given there are only 2 assertions above, and the multiple pools gate is needed only for the case when the component is used in the token details section. One way you can avoid this is to request the gate more closer to where it is needed.

src/earn/EarnActivePool.tsx
```diff
@@ -141,7 +145,7 @@ export default function EarnActivePool({ depositTokenId, poolTokenId, cta }: Pro
                   networkId: poolToken.networkId,
                   providerId: PROVIDER_ID,
                 })
-                navigate(Screens.TabDiscover)
+                navigate(showMultiplePools ? Screens.EarnHome : Screens.TabDiscover)
```
satish-ravi, line 148: needs unit tests

satish-ravi (approved):

PR #5671, opened 2024-07-23
feat(analytics): add event for insufficient gas warnings

satish-ravi:
### Description

When preparing a tx in the various flows, if it fails with insufficient gas, add an analytics event.

### Test plan

Unit tests, manual

### Related issues

- Fixes ACT-1256

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/src/walletConnect/saga.ts
+++ b/src/walletConnect/saga.ts
@@ -476,6 +476,7 @@ function* showActionRequest(request: Web3WalletTypes.EventArguments['session_req
         feeCurrencies,
         decreasedAmountGasFeeMultiplier: 1,
         baseTransactions: [normalizedTx],
+        origin: 'wallet-connect' as const,
       })
     } catch (err) {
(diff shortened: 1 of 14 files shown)


src/walletConnect/saga.ts
```diff
@@ -476,6 +476,7 @@ function* showActionRequest(request: Web3WalletTypes.EventArguments['session_req
         feeCurrencies,
         decreasedAmountGasFeeMultiplier: 1,
         baseTransactions: [normalizedTx],
+        origin: 'wallet-connect' as const,
```
jh2oman, line 479: why the 'as const' here?

jh2oman (approved):

src/walletConnect/saga.ts
satish-ravi: sagas + TS, type is inferred as `string`

PR #5677, opened 2024-07-25
feat: cross chain swap transaction feed item

kathaypacific:
### Description

This PR augments the swap tx feed item to also display cross chain swaps. 

There is some coordination needed to show the estimated "in" amount for cross chain swaps, because this information is in the standby transaction but usually we use the blockchain-api transactions as the source of truth. I settled on using standby transactions as the source of truth (if possible) for pending cross chain transactions. we can eventually decide to use standby transactions as the source of truth for _all_ pending transactions, but blockchain-api does not have real examples for this yet.

The above means that:
- when a pending cross chain tx is created from the wallet, we have most things we need to display the pending tx
- when the pending tx comes back from blockchain-api, it contains the swap fee info but _not_ the inAmount value. this problem is resolved in the reducer explicitly by modifying the existing standby transaction.
- when the tx comes back from blockchain-api and is completed, the standby transaction is removed as usual
- if there is a wallet restore and the standby transaction is removed before it is settled, the pending tx from blockchain-api goes through the normal completed tx flow of being rendered using the available info (no inAmount value)

### Test plan

https://github.com/user-attachments/assets/cb4e571a-1930-4c2a-a374-24537f57ef1f

After the transaction has settled:

https://github.com/user-attachments/assets/9f22457d-aa62-40f6-bab2-3cbc46f7ec35

### Related issues

- Fixes RET-1118

### Backwards compatibility

Y

### Network scalability


src/transactions/feed/SwapFeedItem.tsx
```diff
@@ -13,64 +14,74 @@ import { Spacing } from 'src/styles/styles'
...
+    <Touchable testID="SwapFeedItem" onPress={handleOpenTransactionDetails}>
       <View style={styles.container}>
         <TransactionFeedItemImage
-          status={exchange.status}
-          transactionType={exchange.__typename}
-          networkId={exchange.networkId}
+          status={transaction.status}
+          transactionType={transaction.__typename}
+          networkId={transaction.networkId}
+          hideNetworkIcon={isCrossChainSwap}
```
jeanregisser, line 42: nit: Why aren't we keeping the network icon here? Still seems like a useful bit of info. Or do we have another plan for a future PR?

src/transactions/reducer.ts
```diff
@@ -145,12 +145,65 @@ export const reducer = (
...
+              tx.transactionHash === standbyTx.transactionHash &&
+              tx.networkId === standbyTx.networkId
+          )
+        ) {
+          pendingTxsWithStandbyMatch.push(tx)
+        } else {
+          receivedTransactions.push(tx)
+        }
+      })
+
```
jeanregisser, line 164: Maybe we should add a comment of what this logic is trying to do? and why?
What's in the PR description.
Or I guess the tests are good too, and we could point the reader to them?

src/transactions/reducer.ts
```diff
@@ -22,7 +22,7 @@ export interface InviteTransactions {
 type TransactionsByNetworkId = {
   [networkId in NetworkId]?: TokenTransaction[]
 }
-interface State {
+export interface State {
```
jeanregisser, line 25: nit: It seems this is exported only for testing.
We could infer it instead for the test?

src/transactions/reducer.ts
```diff
@@ -46,7 +46,7 @@ export interface KnownFeedTransactionsType {
   [txHash: string]: string | boolean
 }
 
-const initialState = {
+export const initialState = {
```
jeanregisser, line 49: nit: Also this is only exported for testing, we can either use our convention of using `export const _initialState = initialState` or  calling `reducer(undefined, { type: 'does-not-matter' })` to get the same.

To avoid exporting things that shouldn't be exported.

src/transactions/reducer.test.ts
```diff
@@ -0,0 +1,156 @@
...
+    })
+
+    expect(updatedState).toEqual({
+      ...initialState,
+      transactionsByNetworkId: {
+        [NetworkId['celo-mainnet']]: [
+          { ...incomingCrossChainSwap, status: TransactionStatus.Complete },
+        ],
+      },
+      standbyTransactions: [standbyCrossChainSwap], // the standby transaction is not altered
```
jeanregisser, line 153: Shouldn't the standby TX be removed then?
Or will we do that only once we are able to get the inAmount from the explorer TX?

jeanregisser (commented):
Great 🙌

src/transactions/feed/SwapFeedItem.tsx
kathaypacific: the network icon is hidden [on purpose](https://www.figma.com/design/fjpfHZdVSLazv330kT2sQz?node-id=1573-6584#875927421) from the design side. i also asked this question during a refinement and it was Kayla's preference to hide it for now

src/transactions/reducer.test.ts
kathaypacific: the standby tx will be removed indeed, but by the cleanupStandbyTransactions saga (not by this reducer). i'm not sure that it makes sense to mention the sagas in the reducers test. or perhaps the confusion points to a bigger question of why we are modifying state via a saga _and_ reducer? i think that's a legitimate question actually, perhaps i can axe the saga entirely...

jeanregisser (approved):
🚀

PR #5656, opened 2024-07-18
feat(earn): Add EarnHome screen

finnian0826:
### Description

Move over necessary multi-pool earn stuff from earnV2. Two pools shown for testing purposes but the PR only includes the single Aave pool currently supported.

Will add analytics and hook stuff up in follow-up PRs

### Test plan

https://github.com/user-attachments/assets/c6a76424-367f-4d26-84ec-7e0343dd50c0

| Open pools | My pools | Bottom sheet |
| ----- | ----- | ----- | 
| open | my | bottom-sheet |

### Related issues

- Part of ACT-1260

### Backwards compatibility

Yes
### Network scalability

If a new NetworkId and/or Network are added in the future, the changes in this PR will:

- [X] Continue to work without code changes, OR trigger a compilation error (guaranteeing we find it when a new network is added)

PR #5679, opened 2024-07-25
feat(earn): fetch earn positions + selector

jeanregisser:
### Description

Fetch and store earn positions. Also add a selector to list them.

User and earn positions are fetched and stored together.

For each returned earn position:
- if the user already holds the earn position, it is ignored (since it's already present as a user position)
- if not it's added to the list

Next up: use earn positions to drive the earn screen (which was added in https://github.com/valora-inc/wallet/pull/5656)

### Test plan

- Updated tests

### Related issues

- Fixes RET-1121

### Backwards compatibility

Yes

### Network scalability


src/positions/types.ts
```diff
@@ -9,15 +9,28 @@ export interface PositionDisplayProps {
   imageUrl: string
 }
 
+export type DataProps = EarnDataProps
+
+export interface EarnDataProps {
+  apy: number
```
satish-ravi, line 15: I think not every pool will have an apy, some may just use apr. Should we make this optional so this can be compatible with new pools we had that don't have apy? cc @finnian0826 @jh2oman

src/positions/slice.ts
```diff
@@ -22,6 +22,7 @@ export type TriggeredShortcuts = Record<
 interface State {
   positions: Position[]
   positionsFetchedAt?: number
+  earnPositionIds: string[]
```
satish-ravi, line 25: We can also store this in the earn state, but I guess its easier to put this here since it all comes from a single saga

src/positions/types.ts
jeanregisser: Do we want to display APR or APY depending on the pool?
Instead would it be possible to land on one so we are displaying comparable values across our earn pools?

src/positions/slice.ts
jeanregisser: Yes I was hesitating.

src/positions/types.ts
satish-ravi: I believe @finnian0826 has been working with product on this, and from the latest designs presented this week, I think we wanted to include a total yield and a breakdown, although I can't find the link to the designs right now. @finnian0826 do you have it?

finnian0826: yeah the designs for the breakdown are [here](https://www.figma.com/design/E1rC3MG74qEg5V4tvbeUnU/Stablecoin-Enablement?node-id=4761-18541&t=jjFc6ucvG3KMaiKG-0), I think we could include a compounding APY field, and a reward APR field, that should cover all of the different pool types I think.

jeanregisser:
@satish-ravi @finnian0826 ok to still merge this one? We can work on the latest shape for the earn data props separately.

finnian0826 (approved):

PR #5680, opened 2024-07-26
feat: add network overlay icon to dapp positions

silasbw:


src/tokens/PositionIcon.tsx
```diff
@@ -0,0 +1,49 @@
...
+import FastImage from 'react-native-fast-image'
+import { useSelector } from 'src/redux/hooks'
+import { networksIconSelector } from 'src/tokens/selectors'
+import { Spacing } from 'src/styles/styles'
+import { Position } from 'src/positions/types'
+
+const networkImageSize = 12
+const networkImagePosition = 20
+
+export const PositionIcon = ({ position, testID }: { position: Position; testID?: string }) => {
```
satish-ravi, line 12: Would be nice to use https://github.com/valora-inc/wallet/blob/main/src/components/IconWithNetworkBadge.tsx for this, although would need parameterizing size, can maybe add a TODO linking to ACT-1137

satish-ravi (approved):

src/tokens/PositionIcon.tsx
silasbw: Doh -- i missed that one.  I'll add the TODO and then take a pass at a follow up PR.

src/tokens/PositionIcon.tsx
```diff
@@ -0,0 +1,50 @@
...
```
jeanregisser, line 12: nit: easier to parse visually and less chars
```suggestion
export function PositionIcon({ position, testID }: { position: Position; testID?: string }) {
```

jeanregisser (approved):
🙌

PR #5683, opened 2024-07-29
feat: EstimatedNetworkFee with loading state and multiple TXs

jeanregisser:
### Description

Extend `<EstimatedNetworkFee />` so it can show a loading state (while preparing transactions) and use multiple TXs to show the total estimated fee.

It also now uses `getEstimatedFee` instead of `getMaxGasFee`. 
So it's closer to the actual fee.

### Test plan

- Updated tests

https://github.com/user-attachments/assets/e351f231-5885-435f-8555-07df348f9f48

Note: This video is just to show the loading state, this PR doesn't yet add it to the claim shortcut bottom sheet, it will be in another PR.

### Related issues

- Fixes RET-1108

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/src/walletConnect/screens/EstimatedNetworkFee.tsx
+++ b/src/walletConnect/screens/EstimatedNetworkFee.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react'
 import { useTranslation } from 'react-i18next'
 import { StyleSheet, Text, View } from 'react-native'
+import SkeletonPlaceholder from 'src/components/SkeletonPlaceholder'
 import TokenDisplay from 'src/components/TokenDisplay'
 import { useSelector } from 'src/redux/hooks'
 import { NETWORK_NAMES } from 'src/shared/conts'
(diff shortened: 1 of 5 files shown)


src/walletConnect/screens/EstimatedNetworkFee.tsx
```diff
@@ -1,69 +1,112 @@
...
+function getNetworkFee(
+  transactions: SerializableTransactionRequest[],
+  networkId: NetworkId,
+  tokensById: TokenBalances
+) {
+  try {
+    const preparedTransactions = transactions.map(getPreparedTransaction)
+    const feeCurrency = getFeeCurrencyToken(preparedTransactions, networkId, tokensById)
+    if (!feeCurrency) {
+      Logger.warn('WalletConnect', 'No fee token info found', { transactions, networkId })
```
bakoushin, line 40: shall we extract `WalletConnect` to a `TAG` constant? i noticed this pattern in other places.

bakoushin (approved):
🚀 Looking good!

PR #5687, opened 2024-07-30
refactor: consolidate redux state changes on update transactions

kathaypacific:
### Description

[Context](https://github.com/valora-inc/wallet/pull/5677#discussion_r1696537447)

On receiving new transactions from blockchain-api, the redux state that stores transactions is updated via the reducer. We then kick off a saga that dispatches another action to remove corresponding standby transactions via another action. This PR removes that complexity and updates the standby transactions via the original action. Sagas should be reserved for asynchronous side effects. 

### Test plan

n/a

### Related issues

- Related to RET-1118

### Backwards compatibility

Y

### Network scalability


jeanregisser (approved):
Thanks so much! 🚀

PR #5692, opened 2024-07-31
chore(naming): Remove (almost) all in-code references to Valora

jophish:
### Description

Second step in [RET-1154](https://linear.app/valora/issue/RET-1154/%5Bvalora%5D-remove-in-code-valora-references). Refactors the repository, replacing in-code references to `Valora` with generic counterparts. There were a couple of kinds of things that needed to change:

* Variable, type, etc. names - these were straightforward changes
* String literal values. In cases where these can be changed in-place (where the actual value of `"valora"` doesn't matter), they are. In other places where the actual `"valora"` value is relevant, the value has been pulled out into a config variable, so that it can potentially be modified down the line in MS apps.

Some references to valora in the app were _not_ changed, however. This is mainly evident in the deeplink/universal link feature, where pulling those out to be configurable will be done in a separate ticket. Certain code comments that link to Valora repos/Slack threads also remain.

You can get a decent feeling of the valora references that remain with the following `grep` command:

```bash
grep -v -w "//" -r src/ --exclude=\*networkConfig.ts --exclude=\*config.ts --exclude=\*brandingConfig.ts | grep -i -e "valora"
```

Which gives the following output:

```
src/send/hooks.test.tsx:import { NameResolution, ResolutionKind } from '@valora/resolve-kit'
src/send/hooks.ts:import { NameResolution, ResolutionKind } from '@valora/resolve-kit'
src/components/BottomSheetLegacy.tsx: * @deprecated use https://www.npmjs.com/package/@gorhom/bottom-sheet instead, see example here https://github.com/valora-inc/wallet/blob/09c4d4ce98181480f1ed0a299709070d535221ad/src/dappsExplorer/useDappInfoBottomSheet.tsx#L47
src/analytics/InjectTraits.ts: * TODO eventually: remove dependencies and delete this plugin, see https://linear.app/valora/issue/ENG-60/remove-dependencies-on-context-traits-
src/keylessBackup/encryption.ts:  const info = 'valora.keylessBackup.deriveKeyFromKeyShares'
src/account/LicenseDisclaimer.txt:THE FOLLOWING SETS FORTH ATTRIBUTION NOTICES FOR THIRD PARTY SOFTWARE THAT MAY BE CONTAINED IN PORTIONS OF THE @VALORA/WALLET PRODUCT.
src/walletConnect/saga.ts:          universal: 'https://valoraapp.com/wc',
src/walletConnect/walletConnect.ts:const UNIVERSAL_LINK_PREFIX = 'https://valoraapp.com/wc?uri='
src/walletConnect/walletConnect.ts:const UNIVERSAL_LINK_PREFIX_WITHOUT_URI = 'https://valoraapp.com/wc'
src/app/reducers.ts:const PERSISTED_DEEP_LINKS = ['https://valoraapp.com/share', 'celo://wallet/jumpstart']
src/app/saga.test.ts:      link: `https://valoraapp.com/wc?uri=${connectionString}`,
src/app/saga.test.ts:    { name: 'iOS universal link', link: `https://valoraapp.com/wc?uri=${actionString}` },
src/redux/migrations.test.ts:      iconUrl: 'https://raw.githubusercontent.com/valora-inc/dapp-list/main/assets/ubeswap.png',
src/redux/migrations.test.ts:              'https://raw.githubusercontent.com/valora-inc/dapp-list/main/assets/ubeswap.png',
src/redux/migrations.test.ts:            iconUrl: 'https://raw.githubusercontent.com/valora-inc/dapp-list/main/assets/moola.png',
src/redux/migrations.test.ts:              'https://raw.githubusercontent.com/valora-inc/dapp-list/main/assets/ubeswap.png',
src/redux/migrations.test.ts:          iconUrl: 'https://raw.githubusercontent.com/valora-inc/dapp-list/main/assets/ubeswap.png',
src/redux/migrations.test.ts:              'https://raw.githubusercontent.com/valora-inc/dapp-list/main/assets/ubeswap.png',
src/redux/migrations.test.ts:          iconUrl: 'https://raw.githubusercontent.com/valora-inc/dapp-list/main/assets/ubeswap.png',
src/redux/migrations.ts:      ..._.omit(state.recipients, 'valoraRecipientCache'),
src/redux/migrations.ts:      appRecipientCache: state.valoraRecipientCache,
src/redux/migrations.ts:      ..._.omit(state.keylessBackup, 'valoraKeyshare'),
src/redux/migrations.ts:      appKeyshare: state.keylessBackup.valoraKeyshare,
src/utils/safely.ts: * See more details: https://github.com/valora-inc/wallet/tree/main/WALLET.md#redux-saga-pitfalls
```

These all roughly fall into the categories of:
* Imports from `@valora` scoped repos
* Code comments linking to stuff owned by Valora
* Deeplink/Universal Link stuff
* License stuff
* Redux migrations
* A single keyless backup string literal; this is not going to change with MS, though maybe we can pull it into `config.ts` if we want.

### Test plan

Unit tests updated where appropriate.

### Related issues

- Fixes [RET-1154](https://linear.app/valora/issue/RET-1154/%5Bvalora%5D-remove-in-code-valora-references).

### Backwards compatibility

Yes.

### Network scalability


src/analytics/Events.tsx
```diff
@@ -111,9 +111,9 @@ export enum KeylessBackupEvents {
   cab_issue_sms_code_error = 'cab_issue_sms_code_error',
   cab_enter_phone_code_back = 'cab_enter_phone_code_back',
   cab_enter_phone_code_cancel = 'cab_enter_phone_code_cancel',
-  cab_issue_valora_keyshare_start = 'cab_issue_valora_keyshare_start',
-  cab_issue_valora_keyshare_success = 'cab_issue_valora_keyshare_success',
-  cab_issue_valora_keyshare_error = 'cab_issue_valora_keyshare_error',
+  cab_issue_app_keyshare_start = 'cab_issue_app_keyshare_start',
+  cab_issue_app_keyshare_success = 'cab_issue_app_keyshare_success',
+  cab_issue_app_keyshare_error = 'cab_issue_app_keyshare_error',
```
satish-ravi, line 116: this is a non BC change, as it results in creating new analytics events and has analytics implications as we'll have to look at different events for different app versions. Also, depending on how we split out CAB, this maybe continued to be issued by Valora or by individual apps, so this event name maybe slightly misleading. Both are probably fine, but just wanted to call it out

jophish: oh shoot, yeah i made a note of this as i made this change, but this was a big PR so i forgot about it by the time i wrote the description. i can update the description.

re: naming, yeah, we could probably keep the existing naming if we wanted, since our current plan is to share the CAB system between all MS apps (e.g., you can backup in one and restore in another), so calling it the e.g., "Valora CAB" system might be fine.

src/redux/migrations.ts
```diff
@@ -1824,4 +1824,15 @@ export const migrations = {
   }),
   220: (state: any) => state,
   221: (state: any) => state,
+  222: (state: any) => ({
+    ...state,
+    recipients: {
+      ..._.omit(state.recipients, 'valoraRecipientCache'),
+      appRecipientCache: state.valoraRecipientCache,
+    },
+    keylessBackup: {
```
satish-ravi, line 1833: this isn't persisted, so you don't need a migration

src/analytics/Events.tsx
kathaypacific: i'm not sure if these events are being used for marketing segmentation in clevertap or on mixpanel dashboards, if we want to make this non BC change can you let the product / marketing folks know about the change @jophish ?

src/redux/migrations.ts
kathaypacific: should we remove this part then?

kathaypacific (approved):
🧹

src/analytics/Events.tsx
jophish: https://valora-app.slack.com/archives/C025N1VU657/p1722970764905039

PR #5694, opened 2024-07-31
feat(shortcuts): multichain support and UI improvements

jeanregisser:
### Description

Updated the claim shortcuts flow to support multichain with viem.
- uses our `sendPreparedTransactions` saga under the hood
  - this allows sending multiple TXs at once
- updated the confirmation bottom sheet to surface the following:
  - estimated gas
  - preparation failures with a user friendly message that can surface contract call errors (similar to what we do for WalletConnect)
  - out of gas 

### Test plan

- Updated tests

| Prepare success | Prepare failure | Need more gas |
|-|-|-|
| image | image | image |

https://github.com/user-attachments/assets/0a1c7273-4260-4858-942a-b2f7c28a81fd

### Related issues

- Fixes RET-1108

### Backwards compatibility

Yes

### Network scalability


src/positions/slice.ts
```diff
@@ -1,22 +1,25 @@
...
+  | {
+      status: ShortcutStatus
+      transactions: RawShortcutTransaction[]
+      appName: string
+      appImage: string
+      appId: string
+      networkId: NetworkId
+      shortcutId: string
+    }
+  | undefined
```
satish-ravi, line 22: when can this be `undefined`?

satish-ravi (approved):

src/positions/slice.ts
jeanregisser: If we pass an ID that wasn't stored.
More to be on the safe side, to force callers to check for that case.

PR #5695, opened 2024-07-31
feat: add cross chain transaction details screen

kathaypacific:
### Description

This PR adds the transaction details screen for cross chain swaps. Key changes include:
- add the swap networks for cross chain swaps
- add the app fees and cross chain fees (this will also work for same chain swaps once we do the backend work)
- add the axelar scan links

### Test plan

(long video)

https://github.com/user-attachments/assets/924b443f-70d7-4bb0-852d-5968e717cca0

Failed swap:
Simulator Screenshot - iPhone 15 Pro - 2024-07-31 at 15 23 29

### Related issues

- Fixes RET-1119

### Backwards compatibility

Y

### Network scalability


src/transactions/feed/detailContent/FeeRowItem.tsx
```diff
@@ -30,26 +38,29 @@ export default function FeeRowItem({ fees, feeType, transactionStatus }: Props)
   }
 
   return (
-    <View testID="TransactionDetails/FeeRowItem">
-      <View style={styles.row}>
-        <Text style={styles.bodyText}>{label}</Text>
+    <View style={styles.row} testID="TransactionDetails/FeeRowItem">
```
kathaypacific, line 41: moving the Views around so that long labels can wrap to 2 lines without affecting the right hand values

src/transactions/feed/detailContent/SwapContent.tsx
```diff
@@ -4,90 +4,122 @@ import { useTranslation } from 'react-i18next'
...
+      )}
+
       <FeeRowItem
-        fees={exchange.fees}
+        fees={transaction.fees}
         feeType={FeeType.SecurityFee}
-        transactionStatus={exchange.status}
+        transactionStatus={transaction.status}
+      />
+      <FeeRowItem
```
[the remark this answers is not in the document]
jophish: for example, the current setup technically contains the possibility that a [row divider](https://github.com/valora-inc/wallet/pull/5695/files#diff-3ade5aeec77d2dbc8ab35eab85cf5c3063663ee12d23d67909ca5871c312c739R74) will be present with nothing underneath, if `showExchangeRate` is `false` and the only fee that's present is a `FeeType.GatewayFee`, which we don't show (and `FeeRowItem` is in fact incapable of rendering). i don't think this will ever ever happen (and even if it does, doesn't _really_ matter) but it does go to show that the logic feels a little wonky

src/transactions/feed/detailContent/SwapContent.tsx
```diff
@@ -4,90 +4,122 @@ import { useTranslation } from 'react-i18next'
...
+    !new BigNumber(transaction.inAmount.value).isNaN() &&
+    !!fromToken &&
+    !!toToken
 
   return (
     <View style={styles.contentContainer}>
-      <View style={[styles.row, { paddingBottom: Spacing.Smallest8 }]}>
-        <Text style={styles.bodyText}>{t('swapTransactionDetailPage.swapTo')}</Text>
+      <View style={styles.row}>
+        <Text style={styles.bodyText}>{t('swapTransactionDetailPage.swapFrom')}</Text>
```
jophish, line 36: i assume that switching the from/to ordering for _all_ types of swaps (not just cross-chain) is intended here? (i do see this in the cross-chain designs, but just curious)

jophish (approved):
LGTM! some minor nits and a rant about conditional rendering, but looks great :sweat_smile:

src/transactions/feed/detailContent/FeeRowItem.tsx
```diff
@@ -21,6 +22,13 @@ export default function FeeRowItem({ fees, feeType, transactionStatus }: Props)
       transactionStatus === TransactionStatus.Pending
         ? t('transactionFeed.estimatedNetworkFee')
         : t('transactionFeed.networkFee')
+  } else if (feeType === FeeType.AppFee) {
```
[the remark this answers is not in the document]
kathaypacific: i think yes, but we have stuff in the FeeType (gateway fee and encryption fee) that are not actually used. i'm not sure if this is intentional, so i'll check and address separately

src/transactions/feed/detailContent/SwapContent.tsx
kathaypacific: yes i think so!

PR #5698, opened 2024-08-01
fix: prevent duplicate cross chain pending swap tx items

kathaypacific:
### Description

This PR is a "quick fix" so I didn't bother with tests. I want to rework the transaction feed stuff because it's too complex and doesn't work well. There are a ton of rerenders constantly (possibly contributing to app slowness) and most of the issue stems from storing the fetched transactions in state which doesn't preserve the identity of the objects even if they change. It also means that fetched transactions are displayed right away, without being coordinated with other stored redux data.

The last part of this is really hard to work with. For pending cross chain swaps, any standby transaction should be augmented with the data (in effect, the standby transactions are the source of truth when available for pending transactions). With the current setup of the TransactionFeed, anything that comes back from blockchain-api goes straight to the display layer and this causes the standby tx in redux _and_ received tx from blockchain-api to be displayed at the same time if they both exist, like in a cross chain transaction made through the app.

This is what i saw once (note that this can only happen if we receive a pending tx from blockchain-api, for cross chain swaps that are quite fast the first tx we receive from blockchain-api is not in the pending state)
https://github.com/user-attachments/assets/96f82c56-7cc6-442b-af00-fc9ff8cd16ff

### Test plan

The duplicate transaction is not there anymore when doing cross chain swaps.

### Related issues

- Related to RET-1119

### Backwards compatibility

Y

### Network scalability

### DIFF
--- a/src/transactions/feed/TransactionFeed.tsx
+++ b/src/transactions/feed/TransactionFeed.tsx
@@ -23,7 +23,7 @@ import {
   pendingStandbyTransactionsSelector,
   transactionsSelector,
 } from 'src/transactions/reducer'
-import { TokenTransaction } from 'src/transactions/types'
+import { TokenTransaction, TransactionStatus } from 'src/transactions/types'
 import { groupFeedItemsInSections } from 'src/transactions/utils'
 
 function TransactionFeed() {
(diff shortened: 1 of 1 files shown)


src/transactions/feed/TransactionFeed.tsx
```diff
@@ -36,16 +36,37 @@ function TransactionFeed() {
   const allowedNetworks = getAllowedNetworkIds()
 
   const confirmedFeedTransactions = useMemo(() => {
+    // Filter out received pending transactions that are also in the pending
+    // standby array because those will be displayed with the pending
+    // transactions
+    const completedTransactionsOrWithoutPendingStandby = transactions.filter(
+      (tx) =>
+        tx.status === TransactionStatus.Complete ||
+        !allPendingTransactions.find(
```
bakoushin, line 45: how long this list may be? if it's long perhaps we can create a Map to reduce the computational complexity

bakoushin (approved):
LGTM 🚀

src/transactions/feed/TransactionFeed.tsx
```diff
@@ -36,16 +36,37 @@ function TransactionFeed() {
   const allowedNetworks = getAllowedNetworkIds()
 
   const confirmedFeedTransactions = useMemo(() => {
+    // Filter out received pending transactions that are also in the pending
+    // standby array because those will be displayed with the pending
+    // transactions
+    const completedTransactionsOrWithoutPendingStandby = transactions.filter(
```
jophish, line 42: nit: name is a little confusing, maybe `completedOrNotPendingStandbyTransactions`? still kind of a mouthful either way :sweat_smile:

jophish (approved):

src/transactions/feed/TransactionFeed.tsx
kathaypacific: in most cases it should be short (this is only the in-flight transactions, not including the completed and failed ones). since users need to make these transactions in the wallet itself to get the standby transactions, it seems reasonable to assume that there is unlikely to be a big buildup of pending transactions especially when we have the internal tx status watcher.

src/transactions/feed/TransactionFeed.tsx
kathaypacific: haha yeah i like it, thanks!

PR #5703, opened 2024-08-02
refactor: use useAllowedNetworks in a hook

kathaypacific:
### Description

This was a suggestion in #5699 but it was merged before being addressed. This also fixes a bad trailing space that was left in the string, which causes a 400 response from blockchain-api.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y

### Network scalability


jeanregisser (approved):
👍

PR #5704, opened 2024-08-02
fix(swap): ensure long running cross chain swaps eventually shows completed status

kathaypacific:
### Description

This PR fixes a bug where pending cross chains swaps remain pending forever. The root cause is the query helper reconciles already received transactions with newly received transactions by preferring the already received transaction. This means that previously received pending swaps are not updated when the same transactions is received again with the completed status.

This is a bug that affects not only cross chain swaps (we should always prefer the newest information in case other fields in the transaction have changed.

### Test plan

Cross chains swaps can be settled correctly

### Related issues

n/a

### Backwards compatibility

Y

### Network scalability


src/transactions/feed/queryHelper.ts
```diff
@@ -49,18 +49,24 @@ const TAG = 'transactions/feed/queryHelper'
...
   existingTxs: TokenTransaction[],
   incomingTxs: TokenTransaction[]
 ) => {
-  const currentHashes = new Set(existingTxs.map((tx) => tx.transactionHash))
-  const transactionsWithoutDuplicatedHash = existingTxs.concat(
-    incomingTxs.filter((tx) => !isEmpty(tx) && !currentHashes.has(tx.transactionHash))
-  )
-  transactionsWithoutDuplicatedHash.sort((a, b) => {
+  const transactionsByTxHash: { [txHash: string]: TokenTransaction } = {}
+  existingTxs.forEach((transaction) => {
```
jophish, line 60: nit: can do 
```ts
[...existingTxs, ...incomingTxs].forEach((transaction) => {
    transactionsByTxHash[transaction.transactionHash] = transaction
  })
```
to save some space

src/transactions/reducer.ts
```diff
@@ -239,10 +239,10 @@ export const pendingStandbyTransactionsSelector = createSelector(
     return transactions
       .filter((transaction) => transaction.status === TransactionStatus.Pending)
       .map((transaction) => ({
-        ...transaction,
         transactionHash: transaction.transactionHash || '',
         block: '',
         fees: [],
+        ...transaction, // in case the transaction already has the above (e.g. cross chain swaps), use the real values
```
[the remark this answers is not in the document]
jophish: can we cover this new behavior with a test to ensure new TX data overrides old data?

jophish (approved):
LGTM, just small comments

src/transactions/feed/queryHelper.ts
kathaypacific: TY, updated!

PR #5706, opened 2024-08-05
feat(deeplinks): Parameterize deeplink prefix

jophish:
### Description

For [RET-1160](https://linear.app/valora/issue/RET-1160/[deeplinks]-parameterize-deeplink-identifiers-by-app-name). Parameterizes deeplink prefixes, removing in-code references to the `celo:` prefix by adding a config value to `src/config.ts`

### Test plan

Updated unit tests.

### Related issues

- Fixes [RET-1160](https://linear.app/valora/issue/RET-1160/[deeplinks]-parameterize-deeplink-identifiers-by-app-name).

### Backwards compatibility

Y

### Network scalability


kathaypacific (commented):
i think that these changes work for runtime logic that needs the deep link prefix, but we also have it declared as a build time parameter in the AppDelegate and AndroidManifest. i don't have a good solution that works for both at the top of my head, it might be okay to just add a comment that points to the other places that need to match the value in the config file?

kathaypacific (approved):

PR #5708, opened 2024-08-05
fix: display default network explorer in tx details for failed cross chain swap

kathaypacific:
### Description

The only time we mark cross chain swaps as failed is when it fails on the source network. In this case, the cross chain swap hasn't started and AxelarScan does not have a reference to it. We should use the default network explorer which has more information on why the tx failed.

### Test plan

Simulator Screenshot - iPhone 15 Pro - 2024-08-05 at 13 14 48

### Related issues

n/a

### Backwards compatibility

Y

### Network scalability


bakoushin (approved):
🚀

PR #5709, opened 2024-08-05
chore: remove hard coded colours

kathaypacific:
### Description

This PR removes all hard coded color values in the app. Most of the colors had close matches in the `colors` enum already, so nothing much to call out here. 

TODO: I need to check with Kayla about the new green and yellow.

### Test plan

n/a

### Related issues

- Fixes RET-1115

### Backwards compatibility

Y

### Network scalability


bakoushin (approved):
🌈

PR #5716, opened 2024-08-06
fix: handle received pending cross chain swaps correctly

kathaypacific:
### Description

This PR fixes a couple of bugs:
1. when a cross chain swap is pending for a long time, the app vibrates periodically for no apparent reason
2. when a user restores their wallet and there is a pending cross chain swap, and it is actually completed, it will show completed in the app session but on subsequent app sessions, it always starts off as pending.

The root cause for both is due to the query helper attributing the cross chain swap as new or existing incorrectly. For the scenarios:
1. the pending cross chain swap isn't stored with the other received blockchain-api transactions (it is instead stored with the pending standby transactions) so every poll interval, we receive the pending tx from blockchain-api -> it's not in the redux store that we expect -> we count it as new and vibrate
2. the pending tx received from blockchain-api doesn't exist in the standby redux store -> we store it in redux with the other received blockchain-api tx's -> later we receive from blockchain-api the updated status as complete -> we never update the cached transactions because this transaction [doesn't meet the definition of new transaction](https://github.com/valora-inc/wallet/blob/a75e5e5dc5dd5c6bb81d677abe3d0168b089a9b6/src/transactions/feed/queryHelper.ts#L196).

This PR fixes both by including a check for the transaction status when determining whether or not to dispatch the redux action. This is the second time I've had to fix something around the queryHelper so I also tried to add some tests focused on the action dispatch and haptic feedback.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y

### Network scalability

### DIFF
--- a/src/transactions/feed/queryHelper.ts
+++ b/src/transactions/feed/queryHelper.ts
@@ -12,7 +12,7 @@ import { getMultichainFeatures } from 'src/statsig/index'
 import { vibrateSuccess } from 'src/styles/hapticFeedback'
 import { updateTransactions } from 'src/transactions/actions'
 import { transactionHashesByNetworkIdSelector } from 'src/transactions/reducer'
-import { NetworkId, TokenTransaction } from 'src/transactions/types'
+import { NetworkId, TokenTransaction, TransactionStatus } from 'src/transactions/types'
 import Logger from 'src/utils/Logger'
 import { gql } from 'src/utils/gql'
 import config from 'src/web3/networkConfig'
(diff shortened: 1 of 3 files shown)


src/transactions/feed/queryHelper.ts
```diff
@@ -322,6 +291,84 @@ function anyNetworkHasTransactionsOnCurrentPage(hasTransactionsOnCurrentPage: {
...
+      const nonEmptyTransactions = returnedTransactions.filter(
+        (returnedTransaction) => !isEmpty(returnedTransaction)
+      )
+      const knownCompletedTransactionHashes = completedTxHashesByNetwork[networkId]
+      const knownPendingTransactionHashes = pendingTxHashesByNetwork[networkId]
+      let shouldUpdateCachedTransactions = false
+      let hasNewCompletedTransaction = false
+
+      // Compare the new tx hashes with the ones we already have in redux
+      for (const tx of nonEmptyTransactions) {
```
kathaypacific, line 344: the diff is a bit hard to compare but this block is the main change

bakoushin (approved):
LGTM!

PR #5718, opened 2024-08-06
feat: add info disclaimer on points screen

kathaypacific:
### Description

As the title.

### Test plan

https://github.com/user-attachments/assets/f8cc7380-0b3e-495d-a249-ce1b37869ad8

### Related issues

- Fixes RET-1167

### Backwards compatibility

Y

### Network scalability

### DIFF
--- a/locales/base/translation.json
+++ b/locales/base/translation.json
@@ -2341,7 +2341,7 @@
     "allNetworks": "All Networks"
   },
   "points": {
-    "title": "Valora Points",
+    "title": "{{appName}} Points",
     "activity": "Activity",
     "history": {
       "cards": {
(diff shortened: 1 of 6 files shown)


locales/base/translation.json
```diff
@@ -2449,6 +2449,11 @@
       "title": "Earn points effortlessly",
       "description": "Jump into a new way to earn! Your {{appName}} actions now earn points. Start earning today and unlock awesome benefits.",
       "cta": "Start your points journey"
+    },
+    "disclaimer": {
+      "learnMoreCta": "<0>Learn more</0> about {{appName}} Points",
+      "body": "<0>What are Points?</0>\nPoints are {{appName}}'s loyalty program. They're awarded to users when you use our app.\n\n<0>What are they good for?</0>\nYou collect them. They can’t be spent or traded or transferred.\n\n<0>Why are they in my wallet?</0>\nBecause they’re tokens. On chain, digital bits of happiness you earn by using the app.",
```
bakoushin, line 2455: I'm curious, how could they be in the wallet? Unless imported

bakoushin (approved):
🏀

locales/base/translation.json
kathaypacific: i'm going to merge this and update the copy separately once it's ready (i think there will be copy changes for swaps too, so can do those together)

PR #5729, opened 2024-08-07
feat(swaps): cross chain swap success analytics

bakoushin:
### Description

* Do not send the `swap_execute_success` event when the source chain transaction is complete, because we have to wait for the destination chain
* Send this event once the standby transaction changes from `Pending` to `Complete`

Note: for cross-chain swaps, we send to analytics a slightly different set of additional information, than for a same-chain swap, because we don't have so many details when we receive the status change from the blockchain-api.

### Test plan

* Tested manually
* Updated unit tests

### Related issues

- Fixes RET-1163

### Backwards compatibility

Y

### Network scalability


src/swap/saga.ts
```diff
@@ -285,11 +285,17 @@ export function* swapSubmitSaga(action: PayloadAction<SwapInfo>) {
       })
     )
 
-    AppAnalytics.track(SwapEvents.swap_execute_success, {
-      ...defaultSwapExecuteProps,
-      ...getTimeMetrics(),
-      ...getSwapTxsReceiptAnalyticsProperties(trackedTxs, networkId, tokensById),
-    })
+    if (swapType === 'same-chain') {
```
kathaypacific, line 288: i wonder if we want to extend this block to also include the dispatching the `swapSuccess` action and even waiting for the txReceipt?

src/transactions/feed/queryHelper.ts
```diff
@@ -350,6 +367,14 @@ export function handlePollResponse({
             shouldUpdateCachedTransactions = true
             hasNewCompletedTransaction = true
           }
+
+          if (
+            // Track cross-chain swap transaction status change to `Complete`
+            tx.__typename === 'CrossChainTokenExchange' &&
+            pendingStandbyTransactionHashes?.has(tx.transactionHash)
```
kathaypacific, line 374: it's also possible for the pending transaction to have come through blockchain-api only, without having a standby transaction (account restore) - we also need to check `knownPendingTransactionHashes`?

src/transactions/reducer.ts
```diff
@@ -316,3 +316,23 @@ export const completedTxHashesByNetworkIdSelector = createSelector(
     return hashesByNetwork
   }
 )
+
+export const pendingStandbyTxHashesByNetworkIdSelector = createSelector(
```
kathaypacific, line 320: this is fine but we do also have `pendingStandbyTransactionsSelector`, it's less efficient to use i guess, but if the array is always small then i wonder if it's worth having 2 selectors that do basically the same thing?

kathaypacific (approved):
thank you!! 🙏🏻

src/transactions/feed/queryHelper.ts
bakoushin: ah yes, of course!

PR #5730, opened 2024-08-07
feat(earn): Use hooks info for multi-pool earn flow

finnian0826:
### Description

Use hooks earn positions in the multi-pool earn flow rather than hard-coded pool list.

Updated logic so it uses balance from the hooks response rather than tokenInfo, that way we don't need to add new pool tokens to address-metadata.

Deposit and withdraw flows still need to be updated to work for multi-pool.

### Test plan

Tests updated 

https://github.com/user-attachments/assets/bd50e7fa-6e03-416f-9326-5f70c286c586

### Related issues

- Fixes ACT-1268

### Backwards compatibility

Yes, updating code behind feature gate not yet rolled out

### Network scalability


src/positions/selectors.ts
```diff
@@ -43,7 +43,7 @@ export const earnPositionsSelector = createSelector(
   [positionsSelector, earnPositionIdsSelector],
   (positions, earnPositionIds) => {
     const earnPositionIdsSet = new Set(earnPositionIds)
-    return positions.filter((position) => earnPositionIdsSet.has(position.address))
+    return positions.filter((position) => earnPositionIdsSet.has(position.positionId))
```
finnian0826, line 46: I think this is a bug from Jean, should be checking for positionId here to match

src/earn/PoolCard.tsx
```diff
@@ -53,41 +56,40 @@ export default function PoolCard({ pool, testID = 'PoolCard' }: { pool: Pool; te
...
+              ).toFixed(2),
             })}
           </Text>
         </View>
         <View style={styles.keyValueRow}>
           <Text style={styles.keyText}>{t('earnFlow.poolCard.reward')}</Text>
           <Text
             style={styles.valueText}
-          >{`${new BigNumber(pool.reward).multipliedBy(100).toFixed(2)}%`}</Text>
+          >{`${pool.earnItems.reduce((acc, earnItem) => acc.plus(new BigNumber(earnItem.amount)), new BigNumber(0)).toFixed(2)}%`}</Text>
```
satish-ravi, line 69: move these computations into a variable to make this more readable and potentially a useMemo

src/earn/poolHelper.ts
```diff
@@ -0,0 +1,29 @@
+import BigNumber from 'bignumber.js'
+import { Pool } from 'src/earn/types'
+import { Position } from 'src/positions/types'
+import { Address } from 'viem'
+
+export function convertPositionToPool(position: Position): Pool {
```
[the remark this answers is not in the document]
finnian0826: I had thought about that but the `Position` type includes several different types of positions, each that have slightly different fields. If just using the position type I think there would be the need for extra logic in a lot of places where the fields are used to check that they exist to make typescript happy, even though they should always exist.

src/earn/PoolCard.test.tsx
```diff
@@ -1,34 +1,100 @@
...
 import { createMockStore } from 'test/utils'
-import { mockArbEthTokenId, mockArbUsdcTokenId, mockTokenBalances } from 'test/values'
+import {
+  mockArbEthTokenId,
+  mockArbUsdcTokenId,
+  mockTokenBalances,
+  mockUSDCAddress,
+} from 'test/values'
+
+// const AAVE_POOL: EarnPosition = {
```
satish-ravi, line 19: remove

src/earn/PoolCard.tsx
```diff
@@ -18,17 +20,45 @@ import { Spacing } from 'src/styles/styles'
...
+  const poolBalance =
+    useDollarsToLocalAmount(new BigNumber(pool.balance).times(new BigNumber(pool.priceUsd))) ?? null
+  const poolBalanceString = useMemo(
+    () => `${localCurrencySymbol}${poolBalance ? formatValueToDisplay(poolBalance) : '--'}`,
+    [localCurrencySymbol, poolBalance]
+  )
+
+  const rewardPercentageString = useMemo(
+    () =>
+      `${pool.dataProps.earningItems.reduce((acc, earnItem) => acc.plus(new BigNumber(earnItem.amount)), new BigNumber(0)).toFixed(2)}%`,
```
satish-ravi, line 54: nit: do the `reduce` and string formatting in separate lines, probably fine just to have the string formatting outside the useMemo.

Also, use destructuring assignment to extract `earningItems` instead of doing `pool.dataProps.<>` everywhere

src/positions/types.ts
```diff
@@ -11,13 +11,32 @@ export interface PositionDisplayProps {
...
+interface EarningItem {
+  amount: SerializedDecimalNumber
+  label: string
+  tokenId: string
+  subtractFromDeposit?: boolean
+}
+
 interface EarnDataProps {
-  apy: number
+  contractCreatedAt?: number
```
satish-ravi, line 28: would be good to document what number this is. E.g., epoch seconds, or epoch milliseconds, or something else

finnian0826: this hasn't been implemented in hooks yet so we can still choose what we want it to be, is there an advantage of one over the other? It seems like we don't need millisecond accuracy since we will just show how many days old the pool is, so maybe epoch seconds would be okay?

src/earn/PoolCard.tsx
```diff
@@ -120,11 +126,11 @@ export default function PoolCard({
             <Button
               onPress={() => {
                 AppAnalytics.track(EarnEvents.earn_pool_card_cta_press, {
-                  poolId: pool.positionId,
+                  poolId: positionId,
                   depositTokenId,
-                  networkId: pool.networkId,
-                  tokenAmount: pool.balance.toString(),
-                  providerId: pool.appId,
+                  networkId: networkId,
```
satish-ravi, line 131:
```suggestion
                  networkId,
```

src/earn/PoolCard.tsx
```diff
@@ -138,11 +144,11 @@ export default function PoolCard({
...
```
satish-ravi, line 149:
```suggestion
                  networkId,
```

src/earn/PoolCard.tsx
```diff
@@ -159,11 +165,11 @@ export default function PoolCard({
         <Button
           onPress={() => {
             AppAnalytics.track(EarnEvents.earn_pool_card_cta_press, {
-              poolId: pool.positionId,
+              poolId: positionId,
               depositTokenId,
-              networkId: pool.networkId,
+              networkId: networkId,
```
satish-ravi, line 170:
```suggestion
              networkId,
```

satish-ravi (approved):

PR #5732, opened 2024-08-08
fix(homefeed): prevent reordering on app load

kathaypacific:
### Description

It was annoying me that the homefeed reorders itself on app start. The root cause is that we initialise the `useFetchTransactions` hook with empty transactions, then append new transactions to the empty array as we receive results from each network. This means the first page of results from the first network response will overwrite all the transactions that are currently rendered (aside from the standby transactions, which come through a different datasource). This issue was exacerbated by the change to show responses per network as we get them.

Reminder of the problem:

https://github.com/user-attachments/assets/0731a96f-23fb-41bd-aed2-cb823575bc56

### Test plan

Not much to show, aside from the tx feed is stable on app start now.

### Related issues

n/a

### Backwards compatibility

n/a

### Network scalability


src/transactions/feed/TransactionFeed.tsx
```diff
@@ -30,8 +29,7 @@ function TransactionFeed() {
   const { loading, error, transactions, fetchingMoreTransactions, fetchMoreTransactions } =
     useFetchTransactions()
 
-  const cachedTransactions = useSelector(transactionsSelector)
```
kathaypacific, line 33: nothing interesting is changed in this file. the cachedTransactions is removed here because it will come with the `transactions` from the `useFetchTransactions` hook

src/transactions/feed/queryHelper.ts
```diff
@@ -133,7 +134,7 @@ export function useFetchTransactions(): QueryHookResult {
     pageInfo: { [key in NetworkId]?: PageInfo | null }
     hasTransactionsOnCurrentPage: { [key in NetworkId]?: boolean }
   }>({
-    transactions: [],
+    transactions: cachedTransactions,
```
kathaypacific, line 137: this is the key change that prevents the re-ordering.

src/transactions/feed/queryHelper.ts
```diff
@@ -150,16 +151,20 @@ export function useFetchTransactions(): QueryHookResult {
...
+    return () => clearInterval(id)
+  }, [])
 
   const handleError = (error: Error) => {
     Logger.error(TAG, 'Error while fetching transactions', error)
   }
 
   // Query for new transaction every POLL_INTERVAL
-  const { loading, error } = useAsync(
+  const pollNewTransactions = useAsyncCallback(
```
kathaypacific, line 167: this isn't a consequential change, but i think it is an improvement on readability + removing an unnecessary state variable

bakoushin (approved):
Thanks for addressing this! 🎉

PR #5734, opened 2024-08-08
chore: remove deprecated h1 font

kathaypacific:
### Description

As the title. It seems like a straight replace for h1 -> titleMedium, this is what was asked for in RET-1165.

Screenshot 2024-08-08 at 15 10 08

### Test plan

n/a

### Related issues

Related to RET-1174 

### Backwards compatibility

Y

### Network scalability


bakoushin (approved):
✨

PR #5736, opened 2024-08-08
refactor: remove separate branding repo

kathaypacific:
### Description

Key changes:
1. copy contents of branding repo to `/branding` folder
2. remove celo and valora branding files
3. remove unused images and icons
4. update the post install script to just copy the contents of the new `branding` folder into the appropriate locations. this is the smallest change i could think of, and seems to optimise for having all the branding in one place (alternatively we could remove this script and have the branding in the ios / android / src folders)

### Test plan

CI passes, app builds and displays as expected

### Related issues

- Fixes RET-1156

### Backwards compatibility

Y

### Network scalability


bakoushin (approved):
🖼️🖼️🖼️

.gitignore
```diff
@@ -130,13 +129,9 @@ ios/celo/Base.lproj/*
...
 ios/.xcode.env
 src/brandingConfig.ts
-src/icons/LogoHeart.tsx
-src/icons/Logo.tsx
-src/icons/WelcomeLogo.tsx
-src/icons/Home.tsx
+src/brandedIcons/*
 src/images/*
 !src/images/Images.ts
-!src/images/link-arrow.png
```
bakoushin, line 139: should we delete `link-arrow.png` from the repo?

bakoushin:
what is the reason for keeping a separate folder and copying?
the change is already big to create a new folder with all the images. naively thinking a different location shouldn't affect the change size too much.

jophish (approved):
looks good! i also think it would be nice if we could get rid of the script, though having everything centralized in a separate dir is also sort of nice -- either way works for me

.gitignore
kathaypacific: yes it's been removed!

kathaypacific:
> what is the reason for keeping a separate folder and copying? the change is already big to create a new folder with all the images. naively thinking a different location shouldn't affect the change size too much.

is there an alternative solution that allows us to keep the branded files together? the main reason for keeping the separate folder and copying is because there are branded files needed at build time and run time (in the ios, android, src folders separately). for ease of updating branding, it seemed preferable to have to update one folder rather than 3 different places in the project. i think this branding folder isn't bundled with the app so i don't see much downside with having it separate?

PR #5738, opened 2024-08-08
chore: remove deprecated uncommonly used font styles

kathaypacific:
### Description

This PR removes some deprecated fontStyles that are not widely used. Nothing much to point out, most were straight swaps with a few pixels difference in lineHeight.

| fontStyle name (old) | typeScale name (new) | Image diff |
|--------|--------|--------|
| iconText | labelMedium | Screenshot 2024-08-08 at 15 28 48 |
| largeNumber | titleLarge | Screenshot 2024-08-08 at 15 30 04 |
| mediumNumber | bodyMedium | no diff |
| large | bodyLarge | Screenshot 2024-08-08 at 15 39 39|
| notificationHeadline | labelSemiBoldMedium | Screenshot 2024-08-08 at 15 36 36 | 
| displayName | labelSemiBoldMedium | Screenshot 2024-08-08 at 15 37 38 | 
| sectionHeader | labelSmall | no diff | 

### Test plan

n/a

### Related issues

- Related to RET-1174

### Backwards compatibility

Y

### Network scalability


src/components/AccountNumber.tsx
```diff
@@ -73,9 +73,8 @@ const styles = StyleSheet.create({
     marginBottom: 8,
   },
   link: {
-    ...fontStyles.label,
+    ...fontStyles.small,
```
bakoushin, line 76: small moment of confusion: what is the reason to keep `fontStyles` here?

bakoushin (approved):
Awesome! 🎉

src/components/AccountNumber.tsx
kathaypacific: oh it is because i removed the export for the font family, this component was overriding the font family here which really turns it into something similar to the `small` style. this will eventually be removed

jophish (approved):
:tada:

PR #5741, opened 2024-08-08
chore: remove deprecated fonts (large600, label)

kathaypacific:
### Description

The changes include:

| fontStyle name | typeScale name | note |
|--------|--------|--------|
| large600 | labelSemiBoldLarge | +4pt lineHeight |
| label | labelSemiBoldSmall | +1pt fontSize, +4pt lineHeight |

### Test plan

n/a

### Related issues

- Related to RET-1174

### Backwards compatibility

Y

### Network scalability


src/home/NotificationCenter.tsx
```diff
@@ -269,9 +269,7 @@ const styles = StyleSheet.create({
     flex: 1,
   },
   title: {
-    ...fontStyles.large600,
-    fontSize: 24,
-    lineHeight: 32,
+    ...typeScale.titleMedium,
```
jophish, line 272: why `titleMedium` here and not `semiBoldLarge` as the others? because of the custom `fontSize` and `lineHeight` we currently have?

jophish (approved):

PR #5742, opened 2024-08-08
chore: remove deprecated fontStyle (small)

kathaypacific:
### Description

This PR replaces all `fontStyles.small` with `typeScale.bodySmall`. The difference is +2pt on lineHeight.

### Test plan

n/a

### Related issues

- Related to RET-1174

### Backwards compatibility

Y

### Network scalability


src/components/Avatar.tsx
```diff
@@ -62,6 +62,7 @@ const styles = StyleSheet.create({
     marginBottom: 10,
   },
   contactName: {
+    ...typeScale.bodyXSmall,
```
jophish, line 65: why `bodyXSmall` instead of `bodySmall`?

jophish (approved):

src/components/Avatar.tsx
kathaypacific: ah good catch! i think that should be bodySmall indeed, TY for paying attention :D :D

PR #5743, opened 2024-08-08
chore: remove deprecated font (regular)

kathaypacific:
### Description

This PR replaces all `fontStyles.regular` with `typeScale.bodyMedium`. The difference is +2pt on lineHeight.

### Test plan

n/a

### Related issues

- Fixes RET-1174

### Backwards compatibility

Y

### Network scalability


src/components/Carousel.tsx
jophish, line 1: Why delete this file? Is this related somehow?

src/styles/fonts.tsx
```diff
@@ -148,14 +120,3 @@ export const typeScale = StyleSheet.create({
     letterSpacing: 0.2,
   },
 })
-
-/***
- * @deprecated
- * Use typeScale instead
- */
-export const fontStyles = StyleSheet.create({
```
jophish, line 156: :tada:

jophish (approved):

src/components/Carousel.tsx
kathaypacific: yes i was supposed to update the font style in it and then found it wasn't being used, so deleted it rather than make the update

PR #5751, opened 2024-08-09
fix(points): hide live links when user has no jumpstart tokens

bakoushin:
### Description

Hide live links activity when a user has no jumpstart tokens (or live links are disabled).
This is consistent with the send flow behavior when the live link option is not shown when a user has no jumpstart tokens.

If the user tries to create a jumpstart link without having the jumpstart tokens the app will crash.

Context: https://valora-app.slack.com/archives/C04B61SJ6DS/p1723075264724759

### Test plan

* Tested manually
* Updated unit test

### Related issues

NA

### Backwards compatibility

Y

### Network scalability

### DIFF
--- a/src/points/PointsHome.test.tsx
+++ b/src/points/PointsHome.test.tsx
@@ -1,15 +1,18 @@
 import { fireEvent, render, waitFor } from '@testing-library/react-native'
 import * as React from 'react'
 import { Provider } from 'react-redux'
-import { PointsEvents } from 'src/analytics/Events'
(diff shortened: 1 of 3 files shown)


kathaypacific (approved):
i think this will work for the most part, but it probably won't respond to the statsig value changing since it's not part of the `createSelector` inputs. but this is only a problem if we turn off live links

bakoushin:
@kathaypacific great catch, thanks! updated it to mitigate that

PR #5754, opened 2024-08-09
fix(points): dispatch `swapSuccess` event for cross-chain swaps

bakoushin:
### Description

For cross-chain swaps, the `swapSuccess` event wasn't dispatched, and points weren't tracked. This fixes it.

Context: https://valora-app.slack.com/archives/C04B61SJ6DS/p1723220938449899?thread_ts=1723219373.676809&cid=C04B61SJ6DS

### Test plan

Tested manually.

### Related issues

NA

### Backwards compatibility

Y

### Network scalability


kathaypacific (commented):
most of my concerns are more related to my confidence in queryHelper / the tx feed than the changes in this PR, but it doesn't feel right to rely on the tx feed to distribute points? sometimes the tx feed can lag and be unreliable, we've built some internal transaction watching so that we don't have to rely on it...what do you think about awarding points when the source tx succeeds? at that point, the tx is headed for either full or partial success anyway and we can dispatch the event in the swap saga and internal tx watcher as usual? (apologies i didn't think about points when i suggested to skip the whole last section of the swap saga last week...)

bakoushin:
@kathaypacific 
> what do you think about awarding points when the source tx succeeds
yes, this makes sense.

kathaypacific (approved):
thank you!! i have plans to update the transaction time for cross chain swaps in the transaction watcher so i'll also add dispatching this action as part of that work

PR #5756, opened 2024-08-10
chore(deps): restrict commit type to `chore`

bakoushin:
### Description

Committing regular dependency updates with a `fix` semantic commit type seems misleading because those aren't necessarily fixing anything _in the app_, but mostly keeping the dependencies up to date.

For instance: this PR doesn't _fix_ anything in the wallet: https://github.com/valora-inc/wallet/pull/5749

This aims to force renovate commit everything with `chore` type by ignoring the [:semanticPrefixFixDepsChoreOthers](https://docs.renovatebot.com/presets-default/#semanticprefixfixdepschoreothers) preset.

| Before | After |
|--------|--------|
| * Dependency updates committed with `fix` <br> * Dev dependency updates committed with `chore`  | All updates committed with `chore` |

### Test plan

Visual assessment

### Related issues

NA

### Backwards compatibility

NA

### Network scalability

NA


jophish (commented):
I like this change - I think `chore` is semantically closer to what dependency updates are doing, whether or not they're dev deps. Approving since this makes sense to me, but curious to hear what other folks think about this as well.

jophish (approved):

PR #5758, opened 2024-08-12
fix(swaps): hide incompatible selection banner on quote fetch

bakoushin:
### Description

Observed behavior: incompatible selection banner disappears only after the new quote is fetched
Expected behavior: incompatible selection banner disappears right after another token is selected

Context: https://valora-app.slack.com/archives/C04B61SJ6DS/p1723220851805809?thread_ts=1723219373.676809&cid=C04B61SJ6DS

The same applies for other quote-dependent banners as well:
* cross-chain fee warning
* decrease spend amount for gas
* non enough balance for gas
* show missing price impact

| Before | After |
|--------|--------|
| before | after | 

### Test plan

Visual assessment

### Related issues

NA

### Backwards compatibility

Y

### Network scalability

### DIFF
--- a/src/swap/SwapScreen.tsx
+++ b/src/swap/SwapScreen.tsx
@@ -619,11 +619,13 @@ export function SwapScreen({ route }: Props) {
       showUnsupportedTokensWarning:
         !quoteUpdatePending && fetchSwapQuoteError?.message.includes(NO_QUOTE_ERROR_MESSAGE),
       showInsufficientBalanceWarning: parsedSwapAmount[Field.FROM].gt(fromTokenBalance),
-      showCrossChainFeeWarning: crossChainFee?.nativeTokenBalanceDeficit.lt(0),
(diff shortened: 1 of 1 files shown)


kathaypacific (approved):
ah thank you!! i was wondering about this the other day, should we hide all warnings and checks when there is a quote pending? it seems like all checks (apart from the first one) are on the quote - even the first one seems okay to be hidden during the quote fetch...

bakoushin:
@kathaypacific yes, this makes sense.

added the same condition for other quote-dependent banners:
* cross-chain fee warning
* decrease the from amount for gas
* non enough balance for gas
* show missing price impact

these 3 will keep their presence while fetching quote because they aren't dependent on quote:
* switched network (depends on token selection, potentially redundant with the advent of cross-chain swaps)
* insufficient balance (depends on balance/from amount)
* swap max amount warning (depends on balance/from amount)

PR #5760, opened 2024-08-12
fix: cross chain transaction time and points dispatch

kathaypacific:
### Description

This PR fixes:
1. cross chain swaps appear below the corresponding approval transaction on the transaction feed
2. points are not tracked if the cross chain swap is not completed within the app session

### Test plan

1. cross chain swaps should appear after the corresponding approval
2. the trackPointsEvent endpoint is called for a pending cross chain swap on app start that settles in the app session

### Related issues

n/a

### Backwards compatibility

Y

### Network scalability


src/transactions/saga.ts
```diff
@@ -284,20 +303,25 @@ export function* getTransactionReceipt(
...
   })
 
   for (const transaction of filteredPendingTxs) {
-    yield* fork(getTransactionReceipt, transaction, network)
+    yield* fork(getTransactionReceipt, transaction, network, updatedCrossChainSwaps)
   }
 }
 
 export function* watchPendingTransactionsInNetwork(network: Network) {
+  const updatedCrossChainSwaps = new Set<string>()
```
kathaypacific, line 321: this can also be a redux variable but it feels very convoluted...we either choose to persist this so that we don't revisit the swaps across app sessions + add actions to remove the complete swaps, or we have this local variable / redux variable that isn't persisted and accept that for very long running swaps, we'll track the event multiple times (seems like an unlikely scenario)

bakoushin: thanks to idempotency in the contract it is not a big deal, although i would thrive not deliberately sending multiple events.

another option i can think of is to split pending and confirmed standby transactions into separate buckets: 

1) `waitingForReceipt`
2) `waitingForBlockchainAPI`

initially, all transactions land in `waitingForReceipt` bucket. once they get their receipts, they move into the `waitingForBlockchainAPI` bucket with the appropriate statuses and sit there until we get them from the `blockchain-api`

we effectively had this separation by relying on the `pending` status, but with cross-chain swaps, it is not enough anymore.

it is obviously a more complex change. but from my perspective, the resulting data flow could appear more straightforward. i found myself spending more than i wanted trying to wrap my mind over how it is now (and i'm not completely sure that i truly succeeded 😅).

kathaypacific: i also thought about splitting the watchers, but i think the problem remains that we don't know which cross chain transactions have already been tracked. the blockchain-api requests don't go through the sagas currently, so without redoing queryHelper, our only choice would be to listen for the UPDATE_TRANSACTIONS action but that is dispatched with the results of the whole page of transactions. if a user is executing another transaction on the same chain as a cross chain swap, we would either blindly track the cross chain swap even though it's been tracked before or we'd need some kind of tracking for tracked swaps. 

i think the crux of the problem is that the "pending" transaction status now has multiple meanings. in hindsight we should probably have introduced a new transaction status. having said that, i think that we need to be able to differentiate between a cross chain swap that is pending on both chains or on the destination chain only.

i updated the pending standby transaction type to have an extra variable to track this. i think that i like this approach, it doesn't simplify the already complex logic but it feels relatively clean. wdyt?

src/transactions/feed/queryHelper.ts
```diff
@@ -78,9 +78,7 @@ export const deduplicateTransactions = (
     transactionsByTxHash[transaction.transactionHash] = transaction
   })
 
-  return Object.values(transactionsByTxHash).sort((a, b) => {
```
bakoushin, line 81: i wonder if this is safe to remove?

bakoushin (approved):
Thanks for addressing this! 🎉

It turned out that it is not an easy part to handle due to cumulative evolutionary complexity. Maybe we could refactor the whole home feed logic later.

src/transactions/feed/queryHelper.ts
kathaypacific: yes i moved it to the TransactionFeed - we don't actually need to sort the transactions in the query helper, we only need to sort it for displaying the `confirmedFeedTransactions`

PR #5770, opened 2024-08-13
chore(deps): use react-native-restart

MuckT:
### Description

Replaces the old [`react-native-restart-android`](https://www.npmjs.com/package/react-native-restart-android) with the more recently updated [`react-native-restart`](https://www.npmjs.com/package/react-native-restart) and uses this for iOS instead of the previously used `RNExitApp`; This is needed to upgrade to `react-native@0.72.15` & `react-native-reanimated@3.15.0`.

### Test plan

- [x] Tested locally on iOS

### Related issues

- Part of ACT-1274
- Part of ACT-1258

### Backwards compatibility

Yes

### Network scalability

N/A


src/utils/AppRestart.ts
```diff
@@ -1,19 +1,13 @@
...
 export function restartApp() {
   AppAnalytics.track(AppEvents.user_restart)
   Logger.info('utils/AppRestart/restartApp', 'Restarting app')
-  if (Platform.OS === 'android') {
-    RestartAndroid.restart()
-  } else {
-    // We can't restart on iOS, so just exit ;)
-    RNExitApp.exitApp()
-  }
+  RNRestart.restart()
```
satish-ravi, line 12: does this still only exit on iOS or now it can restart? Wonder if we can update the i18n key? Or at least leave a comment if this still only can exit the app on iOS

MuckT: It restarts on iOS now so I'll update the key.

MuckT: Fixed in 485b963bbed0adc2e378db8a2b8c55ef5b90162d!

jh2oman (approved):

PR #5777, opened 2024-08-15
refactor: use viem for live links

kathaypacific:
### Description

As the title.

### Test plan

Live links works as before (no changes for the user)

https://github.com/user-attachments/assets/bfd65fb6-adcf-495d-93ba-86aa15942bad

### Related issues

- Fixes RET-1019
- Related to RET-1173

### Backwards compatibility

Y

### Network scalability


src/jumpstart/jumpstartLinkHandler.test.ts
```diff
@@ -42,26 +45,41 @@ describe('jumpstartLinkHandler', () => {
...
       expect.any(Object),
       expect.any(Number)
     )
   })
 
   it('throws an error if all funds were already claimed', async () => {
-    mockErc20ClaimsCall.mockResolvedValueOnce({ claimed: true })
-    mockErc20ClaimsCall.mockResolvedValueOnce({ claimed: true })
-    mockErc20ClaimsCall.mockRejectedValue(new Error('execution reverted'))
+    jest
```
jophish, line 62: nit: it looks like this mock implementation is identical to the previous one? could this be shared?

jophish (approved):
Looks great! Glad the switch to Viem for this seems relatively painless

src/jumpstart/jumpstartLinkHandler.test.ts
kathaypacific: ah, it's the same except for one line! the first mock returns a true then false for the claims, and this one is false false 🙈

PR #5778, opened 2024-08-15
chore: remove support for old escrow invitations flow

kathaypacific:
### Description

This PR removes support for the old escrow flow, where you could send funds to the escrow contract to invite a friend (using their phone number). Specifically, the following is deleted:
- support for reclaiming sent funds. this includes the UI notification in notification center, as well as all the backend sagas and actions.
- support for showing invitations via escrow in the homefeed. it's been a long time since we removed escrow so not many users should be impacted by this change. those transactions will be omitted from the feed.

### Related issues

- Related to RET-1173

### Network scalability


bakoushin (approved):
🧹 Epic cleanup!

PR #5783, opened 2024-08-16
chore: remove more usages of contractkit, in debug screen

kathaypacific:
### Description

Removing some unused functions and files, and also remove the latest block item in the debug screen because it doesn't seem helpful anymore. If we really wanted to know the latest block, we should print it for all supported networks instead of Celo anyway.

### Test plan

n/a

### Related issues

- Related to RET-1173

### Backwards compatibility

Y

### Network scalability


bakoushin (approved):
LGTM! 🧹

PR #5784, opened 2024-08-16
chore: update earn info screen

MuckT:
### Description

Updates the `EarnInfoScreen.tsx` to use new text, removes references to gas subsidies, makes app name used in text dynamic based on the branding config and adjusts text padding to match button padding.

### Test plan

- [x] Unit tests updated
- [x] Tested locally on iOS

### Related issues

- Fixes ACT-1336

### Backwards compatibility

Yes

### Network scalability

N/A

### DIFF
--- a/locales/base/translation.json
+++ b/locales/base/translation.json
@@ -2484,17 +2484,17 @@
       "details": {
         "earn": {
           "titleGasSubsidy": "Maximize Earnings without Gas Fees*",
-          "title": "Maximize Earnings",
-          "subtitle": "Earn on your stablecoins and get rewards when you supply a pool.",
+          "title": "Explore and Compare Pools",
+          "subtitle": "Compare earning potential and risks. Swap, buy, or transfer tokens to join the best pool for you.",
           "footnoteSubsidy": "*Gas fees are temporarily covered by Valora"
         },
         "manage": {
           "title": "Manage Directly from Valora",
-          "subtitle": "We’ll take care of the setup so you can manage your position directly from Valora."
+          "subtitle": "We'll auto-connect your wallet so you can manage pools without leaving Valora."
         },
         "access": {
           "title": "Easily Access your Funds",
-          "subtitle": "Collect your funds effortlessly with a single tap. Enjoy the same instant access as you would directly through your favorite protocols."
+          "subtitle": "Collect funds with a tap, just like your favorite protocols."
         }
       },
       "action": {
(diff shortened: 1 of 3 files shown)


src/earn/EarnInfoScreen.tsx
```diff
@@ -76,20 +75,13 @@ export default function EarnInfoScreen() {
         <View style={styles.detailsContainer}>
           <DetailsItem
             icon={<EarnCoins size={ICON_SIZE} color={Colors.black} />}
-            title={
-              isGasSubsidized
```
satish-ravi, line 80: we still support gas subsidy right? Are we no longer going to show that message on this screen?

locales/base/translation.json
```diff
@@ -2484,17 +2484,19 @@
       "details": {
         "earn": {
           "titleGasSubsidy": "Maximize Earnings without Gas Fees*",
-          "title": "Maximize Earnings",
-          "subtitle": "Earn on your stablecoins and get rewards when you supply a pool.",
+          "title": "Explore and Compare Pools",
+          "subtitle": "Compare earning potential and risks. Swap, buy, or transfer tokens to join the best pool for you.",
```
satish-ravi, line 2488: Related to the gas subsidy question, this change would mean that for older versions, with gas subsidy on, we'll still show the `titleGasSubsidy` as title, with this new subtile. Is that intended?

src/earn/EarnInfoScreen.tsx
MuckT: Correct, we are no longer showing the gas subsidy message.

locales/base/translation.json
```diff
@@ -2484,17 +2484,19 @@
...
           "footnoteSubsidy": "*Gas fees are temporarily covered by Valora"
         },
         "manage": {
           "title": "Manage Directly from Valora",
-          "subtitle": "We’ll take care of the setup so you can manage your position directly from Valora."
+          "titleV1_92": "Manage Directly from {{appName}}",
```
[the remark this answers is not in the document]
MuckT: Correct since we're adding a new param appName the new translation key is needed.

locales/base/translation.json
MuckT: Checked with product and they're ok showing the gas subsidy title on old app versions.

finnian0826 (approved):

PR #5792, opened 2024-08-19
chore: adjust text for earn and simplify active pool tap action

MuckT:
### Description

Simplies the multipool card in the Discover Tab.

| iOS No Deposits Before | iOS No Deposits After | iOS With Deposits Before | iOS With Deposits After |
|  |  |  |  |

| Android No Deposits Before | Android No Deposits After | Android With Deposits Before | Android With Deposits After |
|  |  |  |  |

### Test plan

- [x] Tested locally on iOS
- [x] Unit tests updated

### Related issues

- Fixes ACT-1333

### Backwards compatibility

Yes

### Network scalability

N/A


src/earn/EarnActivePools.tsx
```diff
@@ -41,51 +41,34 @@ export default function EarnActivePools() {
     [localCurrencySymbol, totalSuppliedValue]
   )
 
+  function handleTap() {
+    AppAnalytics.track(EarnEvents.earn_active_pools_cta_press, {
```
finnian0826, line 45: Does it make sense to continue using this even since the action will always be "myPools"?

MuckT: Fixed in a1595f5e248e6cddf224e335bf8d7e5c2457e18b!

src/analytics/docs.ts
```diff
@@ -606,7 +606,7 @@ export const eventDocs: Record<AnalyticsEventType, string> = {
   [EarnEvents.earn_withdraw_add_gas_press]: `When the user doesn't have enough for gas and clicks on the button to add gas token`,
   [EarnEvents.earn_info_learn_press]: `When the user taps 'Learn More' on the earn info page`,
   [EarnEvents.earn_info_earn_press]: `When the user taps 'Start Earning' on the earn info page `,
-  [EarnEvents.earn_active_pools_cta_press]: `When the user taps one of the CTAs in the active pools card in the discover tab. The action property denotes the CTA, either myPools or exploreOpenPools`,
+  [EarnEvents.earn_active_pools_cta_press]: `When the user taps one of the CTAs in the active pools card in the discover tab.`,
```
finnian0826, line 609:
```suggestion
  [EarnEvents.earn_active_pools_cta_press]: `When the user taps on the active pools card in the discover tab.`,
```

finnian0826: or maybe we just have a new event? it's not really a cta press anymore it's just active pool card press

MuckT: Fixed in 21174eee2f43974eae2b8ae98569fc42a6a858e6!

src/analytics/docs.ts
```diff
@@ -606,7 +606,7 @@ export const eventDocs: Record<AnalyticsEventType, string> = {
   [EarnEvents.earn_withdraw_add_gas_press]: `When the user doesn't have enough for gas and clicks on the button to add gas token`,
   [EarnEvents.earn_info_learn_press]: `When the user taps 'Learn More' on the earn info page`,
   [EarnEvents.earn_info_earn_press]: `When the user taps 'Start Earning' on the earn info page `,
-  [EarnEvents.earn_active_pools_cta_press]: `When the user taps one of the CTAs in the active pools card in the discover tab. The action property denotes the CTA, either myPools or exploreOpenPools`,
```
finnian0826, line 609: Should this be moved down to the commented out deprecated events? (Not sure if we do that for all events we remove or just some.)

finnian0826 (approved):
One small comment but otherwise lgtm!

src/analytics/docs.ts
MuckT: Added docs in fc89dd8637a7d339c9424e127a484ee140eec1ce!

PR #5796, opened 2024-08-20
feat(settings): Update Initial settings screen

jophish:
### Description

For [RET-1180](https://linear.app/valora/issue/RET-1180/[settings]-create-new-settings-screen). First of several PRs for the Settings screen redesign. This PR largely does the following:

* Adds the beginning of a new `SettingsMenu`, which will eventually replace the current `ProfileMenu`. For now, this `SettingsMenu` is inaccessible, and `ProfileMenu` is still used.
* Implements the required redesigns to a subset of the settings submenus/screens, namely the Help and Wallet Address screens.

This PR only makes minor visible changes to the app today. These are:
* Some small design tweaks to the Help submenu and Contact Support screen.
* Some small design tweaks to the QR code screen.
* Self-avatar style is now grey, as called for by designs, and a visual bug with the text icon being off-center has been fixed.
* The margins on the horizontal rules between settings menu items are now equal, rather than extending all the way to the right edge of the screen.

### Test plan

Unit and manual tested. See video:

https://github.com/user-attachments/assets/0f55cb9e-71c1-4334-9be9-39e38063a8f9

### Related issues

Part of [RET-1180](https://linear.app/valora/issue/RET-1180/[settings]-create-new-settings-screen).

### Backwards compatibility

Yes.

### Network scalability


src/components/ContactCircle.tsx
```diff
@@ -55,7 +55,10 @@ function ContactCircle({
       return (
         <Text
           allowFontScaling={false}
-          style={[typeScale.labelMedium, { fontSize: iconSize / 2.0, color: fontColor }]}
+          style={[
+            typeScale.labelMedium,
+            { fontSize: iconSize / 2.0, color: fontColor, lineHeight: iconSize / 1.5 },
```
jophish, line 60: There was an issue here with the capital letter in the avatar being off-center and cut off. Not sure if this was related to the recent port over to using `typeScale`.

src/components/ListItem.tsx
```diff
@@ -33,6 +39,6 @@ const styles = StyleSheet.create({
     paddingVertical: variables.contentPadding,
     borderBottomWidth: 1,
     borderBottomColor: colors.gray2,
-    marginLeft: variables.contentPadding,
+    marginHorizontal: variables.contentPadding,
```
jophish, line 42: I'm not sure why we were doing `marginLeft` only instead of `marginHorizontal` here in the first place. This decision was forcing us to have weird styles elsewhere in the app (basically re-adding the right-padding back in), wherever we use `ListItem` in order to compensate. I manually looked through every instance where we use this and ensured that changes were made to keep styles consistent.

The initial reason for doing it this way _may_ have been to allow us to have horizontal rules on the settings menu that extend all the way to the right of the screen, however, with the new designs, the rules have identical horizontal alignments, so I figured it would be fine to change this sorta hacky style.

src/fiatExchanges/ExternalExchanges.tsx
```diff
@@ -62,7 +62,7 @@ function ExternalExchanges({ route }: Props) {
         })}
       </Text>
 
-      <ScrollView style={styles.exchangesContainer}>
+      <ScrollView>
```
jophish, line 65: This is a fix related to the `ListItem` margin change.

src/fiatExchanges/Spend.tsx
```diff
@@ -56,7 +56,7 @@ function Spend(props: Props) {
     <ScrollView style={styles.container}>
       <SafeAreaView>
         <Text style={styles.pleaseSelectProvider}>{t('useBalanceWithMerchants')}</Text>
-        <View style={styles.providersContainer}>
+        <View>
```
jophish, line 59: Related to the `ListItem` margin change.

src/fiatExchanges/WithdrawSpend.tsx
```diff
@@ -111,7 +111,6 @@ const styles = StyleSheet.create({
   optionsListContainer: {
     flex: 1,
     justifyContent: 'flex-start',
-    paddingRight: variables.contentPadding,
```
jophish, line 114: `ListItem`-related change again.

src/onboarding/registration/PictureInput.tsx
```diff
@@ -76,10 +73,7 @@ function PictureInput({ picture, onPhotoChosen }: Props) {
     )
   }
 
-  const recipient: Recipient = {
-    ...useSelector(currentUserRecipientSelector),
-    thumbnailPath: picture || undefined,
-  }
+  const thumbnailPath = picture || undefined
```
jophish, line 76: This component is never used for anything but the user themselves.

src/navigator/SettingsMenu.tsx
```diff
@@ -0,0 +1,229 @@
+import { NativeStackScreenProps } from '@react-navigation/native-stack'
```
jophish, line 1: Much of this file is just boilerplate, nearly identical to what we're doing in our current settings screen. The only real "new" stuff here is the Profile menu item.

kathaypacific (approved):
🧑🏻‍🎨

src/qrcode/QRTabBar.tsx
```diff
@@ -91,15 +92,15 @@ export default function QRTabBar({
             numberOfLines={1}
             allowFontScaling={false}
           >
-            {state.index === 0 ? t('myCode') : t('scanCode')}
+            {state.index === 0 ? '' : t('scanCode')}
```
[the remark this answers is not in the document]
jophish: The [redesign files for Valora](https://www.figma.com/design/kXwIdFqrwvFOaI63rNRCAb/Mobile-Stack-Run-time?node-id=2433-8624&t=GiUdeEwGyyOWq74N-4) lack this text here, but after referencing the [Mobile stack designs](https://www.figma.com/design/kXwIdFqrwvFOaI63rNRCAb/Mobile-Stack-Run-time?node-id=2521-32801&t=O1lBZT238nnW6EUg-4), it actually appears that they're present. I've updated this (and other screens) to include the correct title in the header, according to the MS design files.

PR #5803, opened 2024-08-20
feat(settings): Implement Profile submenu + redesign, remove profile picture

jophish:
### Description

For [RET-1180](https://linear.app/valora/issue/RET-1180/[settings]-create-new-settings-screen). Implements the profile redesign within the new Settings screen, while removing the profile picture feature. The new settings menu is still not accessible, however some existing components were modified, resulting in a few visible changes to the app. The main ones are:

* Existing profile pictures have been removed, as well as the ability to add new ones.
* The edit profile screen, currently accessible in-app, has been updated to match the Figma designs.
* Some design changes on the connect phone number screen and revoke phone number bottom sheet

### Test plan

Unit and manual tested. See videos below (first video is before removing picture but shows everything, second video just demonstrates removal of profile picture).

https://github.com/user-attachments/assets/d1c086b9-6503-4686-bba2-5a350e831651

https://github.com/user-attachments/assets/34e83696-b542-40a2-9d8b-7120cd22d510

### Related issues

- Part of [RET-1180](https://linear.app/valora/issue/RET-1180/[settings]-create-new-settings-screen)

### Backwards compatibility

Yes

### Network scalability


src/account/ProfileSubmenu.test.tsx
```diff
@@ -0,0 +1,105 @@
+import * as React from 'react'
```
jophish, line 1: These tests were largely ported over from the existing tests on the ProfileMenu screen.

src/account/ProfileSubmenu.tsx
```diff
@@ -0,0 +1,89 @@
...
+import Phone from 'src/icons/Phone'
+import RevokePhoneNumber from 'src/RevokePhoneNumber'
+import { BottomSheetRefType } from 'src/components/BottomSheet'
+import CustomHeader from 'src/components/header/CustomHeader'
+import variables from 'src/styles/variables'
+import BackButton from 'src/components/BackButton'
+
+type Props = NativeStackScreenProps<StackParamList, Screens.ProfileSubmenu>
+
+export default function ProfileSubmenu(props: Props) {
```
jophish, line 24: This whole component is mostly a port of the same existing functionality on the current Settings screen (which will be removed once the new designs are complete)

src/account/Profile.tsx
```diff
@@ -147,21 +121,35 @@ const styles = StyleSheet.create({
...
+    paddingLeft: Spacing.Thick24,
+    paddingBottom: Spacing.Smallest8,
   },
-  disclaimerContainer: {
-    margin: Spacing.Thick24,
+  bottomSection: {
+    flexGrow: 1,
+    justifyContent: 'flex-end',
+    flex: 1,
+    flexDirection: 'column',
```
kathaypacific, line 153: nit: this isn't needed since column is the default value

src/account/ProfileSubmenu.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+  })
+
+  it('shows the expected menu items', () => {
+    const store = createMockStore()
+    const { queryByTestId } = render(
+      <Provider store={store}>
+        <MockedNavigator component={ProfileSubmenu}></MockedNavigator>
+      </Provider>
+    )
+    expect(queryByTestId('ProfileSubmenu/EditProfile')).toBeTruthy()
```
kathaypacific, line 38: probably this is ported over but using the `getBy*` queries would be nice for truthy assertions

src/account/selectors.ts
```diff
@@ -47,8 +43,6 @@ export const recoveringFromStoreWipeSelector = (state: RootState) =>
 export const accountToRecoverSelector = (state: RootState) =>
   state.account.accountToRecoverFromStoreWipe
 
-export const backupCompletedSelector = (state: RootState) => state.account.backupCompleted
```
kathaypacific, line 50: i see this is duplicated in the src/backup folder, but the state is in `account`...so what do you think about keeping this one and removing the other one instead?

src/icons/Phone.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react'
+import Svg, { Path } from 'react-native-svg'
+import colors from 'src/styles/colors'
+
+interface Props {
+  color?: string
```
kathaypacific, line 6:
```suggestion
  color?: Colors
```

src/icons/User.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react'
+import Svg, { Path } from 'react-native-svg'
+import colors from 'src/styles/colors'
+
+interface Props {
+  color?: string
```
kathaypacific, line 6:
```suggestion
  color?: Colors
```
i also wonder if we want to use this User icon with the CircledIcon for DefaultAvatar, these look similar

src/navigator/Headers.tsx
```diff
@@ -192,7 +192,7 @@ export function HeaderTitleWithSubtitle({
   subTitle,
   testID,
 }: {
-  title: string | React.ReactNode
+  title?: string | React.ReactNode
```
kathaypacific, line 195: nothing to do but man, a HeaderTitleWithSubtitle component that does not require a title _or_ a subtitle 🙈

kathaypacific:
sorry if this is the wrong PR to report on, but i see a display bug on the profile picture row when there is no name (the chevron appears right next to the profile icon)

Simulator Screenshot - iPhone 15 Pro - 2024-08-28 at 14 51 38

locales/base/translation.json
```diff
@@ -2008,7 +2011,7 @@
   "profileScreen": {
     "generateName": "Generate a name",
     "namePlaceholder": "Select a name, like \"{{exampleName}}\"",
-    "profilePictureDisclaimer": "Customize {{appName}} with your profile picture. Your image remains private, it is never shared or uploaded."
+    "profilePictureDisclaimer": "Your image remains private, it is never shared or uploaded."
```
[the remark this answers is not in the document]
jophish: Ah yeah, this was from before the removal, ty for the catch

src/account/selectors.ts
jophish: Ahh good catch! We can delete that other file entirely now.

src/icons/User.tsx
jophish: They're nearly imperceptibly different -- I'll replace the default avatar with this, since this is the newest version.

jophish:
TY for the callout with the chevron @kathaypacific! This was easy to miss, since the profile menu doesn't let you set a non-null name after you've already set one prior.
2024-08-28-132155_1011x2160_scrot

PR #5805, opened 2024-08-20
refactor: make generic add assets bottom sheet component

kathaypacific:
### Description

As the title. This change makes it easier to implement a similar component for the points [design](https://www.figma.com/design/6sJ4fCxCptTybHB6ZZ7Pqi/Escrow-Feature?node-id=3341-3388&t=wNp4q9sJM0JyZxXQ-0).

The new component doesn't contain much logic so I didn't invest time in making unit tests.

### Test plan

n/a

### Related issues

- Related to RET-1177

### Backwards compatibility

Y

### Network scalability

### DIFF
--- a/src/components/AddAssetsBottomSheet.tsx
+++ b/src/components/AddAssetsBottomSheet.tsx
@@ -40,15 +40,15 @@ export default function AddAssetsBottomSheet({
   const actionExtraProps = {
     [TokenActionName.Add]: {
       iconComponent: QuickActionsAdd,
-      title: t('earnFlow.addCryptoBottomSheet.actions.add'),
+      title: t('addFundsActions.add'),
     },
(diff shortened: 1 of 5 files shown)


src/components/AddAssetsBottomSheet.tsx
```diff
@@ -0,0 +1,119 @@
...
+  title: string
+  description: string
+  testId: string
+}) {
+  const { t } = useTranslation()
+
+  const actionExtraProps = {
+    [TokenActionName.Add]: {
+      iconComponent: QuickActionsAdd,
+      title: t('earnFlow.addCryptoBottomSheet.actions.add'),
```
jeanregisser, line 43: I wonder if we should rename these translation keys 🤔

jeanregisser (approved):
👍

PR #5806, opened 2024-08-20
feat(jumpstart): show add assets flow when no available Celo tokens

kathaypacific:
### Description

This PR implements [this design](https://www.figma.com/design/6sJ4fCxCptTybHB6ZZ7Pqi/Escrow-Feature?node-id=3324-2382&t=wNp4q9sJM0JyZxXQ-0).
- the live links flow should always work. so at the entry point, launch the intro / add assets screen if the user does not have the correct assets to continue.
- on the points and send screens, allow the live links flow to be displayed even if the user has no tokens on Celo.
- improve the styles on the QR screen (the bottom components were too close to the safe area before, and the address spilled awkwardly to 3 lines)

Not included in this PR:
- analytics
- network pre-selection on the swap and buy flows

### Test plan

https://github.com/user-attachments/assets/98dfac4f-0f6e-42bf-b228-395e9525b57b

https://github.com/user-attachments/assets/ce0ecf90-45b0-424d-82ae-fffc2533a3cf

### Related issues

- Related to RET-1177

### Backwards compatibility

Y

### Network scalability


src/jumpstart/JumpstartAddAssets.tsx
```diff
@@ -0,0 +1,145 @@
...
+      },
+    },
+    {
+      name: TokenActionName.Transfer,
+      details: t('jumpstartIntro.addFundsCelo.actionDescriptions.transfer'),
+      onPress: () => {
+        // TODO analytics
+        navigate(Screens.ExchangeQR, {
+          flow: CICOFlow.CashIn,
+          exchanges: asyncExchanges.result ?? [],
```
jeanregisser, line 62: Should the fallback be an empty array or `undefined`? Or is it the same?

src/fiatExchanges/utils.tsx
```diff
@@ -284,15 +284,15 @@ export const filterLegacyMobileMoneyProviders = (
...
+  tokenId?: string
 ): Promise<ExternalExchangeProvider[] | undefined> {
   // If user location data is not available, default fetching exchanges serving the US
   if (!countryCodeAlpha2) countryCodeAlpha2 = 'us'
   // Standardize cGLD to CELO
 
   try {
     const resp = await fetchWithTimeout(
-      `${networkConfig.fetchExchangesUrl}?country=${countryCodeAlpha2}&tokenId=${tokenId}`,
+      `${networkConfig.fetchExchangesUrl}?country=${countryCodeAlpha2}${tokenId ? `&tokenId=${tokenId}` : ''}`,
```
jeanregisser, line 295: What's the result when no tokenId is provided?

jeanregisser (approved):
Great! 🙌 

One question, what happens after the user has added assets?
They land on the home screen and need to navigate again to live links?

src/jumpstart/JumpstartAddAssets.tsx
kathaypacific: ah, i didn't think when i copied this from the earn stuff but on closer inspection, they are not the same but they should be. i updated `ExchangeQR` to make it the same (we currently allow rendering the exchanges bottom sheet if `exchanges` is defined but what's the point if the array is empty)

src/fiatExchanges/utils.tsx
kathaypacific: thanks for calling this out! i did a (too) quick investigation today and saw that the `getExchanges` CF marks `tokenId` as optional. i didn't dig further, expecting that the CF would then return only exchanges based on the provided query params...but it seems like the CF does default to CELO if the tokenId is not provided 😅 i think for this use case, we can't show the exchanges unless we add token selection first. i see the exchanges part isn't in the design anyway so for now i removed it

PR #5807, opened 2024-08-20
feat(jumpstart): add network preselection for live links buy and swap

kathaypacific:
### Description

This PR adds nav params for the buy and swap flows so that a network can be pre-selected.

### Test plan

https://github.com/user-attachments/assets/cb5895b1-c7ea-4523-bf6b-7dfde631f08a

### Related issues

- Related to RET-1177

### Backwards compatibility

Y

### Network scalability


jeanregisser (approved):
🚀

PR #5809, opened 2024-08-20
refactor: move addAccount and loadStoredAccounts to KeychainLock

jeanregisser:
### Description

Move remaining logic for account management from `KeychainWallet` (ContractKit specific) to `KeychainLock` (library agnostic).
This will make it easier to replace `KeychainWallet` with `KeychainLock` once we remove ContractKit.

### Test plan

- Updated tests

### Related issues

- Related to RET-1182

### Backwards compatibility

Yes

### Network scalability


src/web3/KeychainLock.ts
```diff
@@ -55,15 +51,13 @@ export async function decryptPrivateKey(encryptedPrivateKey: string, password: s
   }
 }
 
-export async function storePrivateKey(
-  privateKey: string,
-  account: KeychainAccount,
-  password: string
-) {
+async function storePrivateKey(privateKey: string, account: KeychainAccount, password: string) {
```
MuckT, line 54: Can we put these in an object instead?

jeanregisser: Yes probably but it's unrelated to the concern of this PR.

jeanregisser:
Yes almost, but we'd need a few modifications to still load the stored accounts once.

PR #5810, opened 2024-08-20
fix: network requests in fb flipper

MuckT:
### Description

Fixes the network plugin not showing requests not showing in iOS on Fb Flipper by removing the explicit initialization in `AppDelegate.mm`.
 
### Test plan

- [x] Tested locally on iOS

### Related issues

https://github.com/facebook/flipper/issues/3945

### Backwards compatibility

Yes

### Network scalability

N/A


jeanregisser (approved):
Thanks 🚀

PR #5812, opened 2024-08-21
chore: display all pools and my pools on earn tab bar

MuckT:
### Description

Changes the pools tab titles and display from:
- Open Pools & My Pools
- All Pools & My Pools

| iOS Before | iOS After |
| ----- | ----- | 
|  |  | 

### Test plan

- [x] Tested locally on iOS
- [x] Unit tests updated

### Related issues

- Fixes ACT-1332

### Backwards compatibility

Yes

### Network scalability

N/A


src/earn/EarnHome.tsx
```diff
@@ -238,7 +238,7 @@ export default function EarnHome({ navigation, route }: Props) {
     return pools.filter((pool) => {
       const depositTokenInfo = allTokens[pool.dataProps.depositTokenId]
       const isMyPool = new BigNumber(pool.balance).gt(0) && !!depositTokenInfo
-      return activeTab === EarnTabType.MyPools ? isMyPool : !isMyPool
+      return activeTab === EarnTabType.MyPools ? isMyPool : pools
```
finnian0826, line 241:
```suggestion
      return activeTab === EarnTabType.MyPools ? isMyPool : true
```
this is just returning a bool for filtering so should be true always for AllPools case

finnian0826, line 241: I think `EarnTabType` enum also should be update to be AllPools rather than OpenPools just for consistency

satish-ravi: maybe just early bail if tab is all pools? E.g.,
```ts
return activeTab === <all pools> ? pools : pools.filter(...)
```

locales/base/translation.json
```diff
@@ -2614,7 +2614,8 @@
     },
     "poolFilters": {
       "openPools": "Open Pools",
```
satish-ravi, line 2616: is this used any longer? if not, can we remove?

MuckT: Removed in ba933e128cc2dc5655a678f3534d6f669e9437b6!

src/earn/EarnHome.tsx
MuckT: Updated the enum in 499eb47ad2f48081a3680818340b064f644073e3!

src/earn/EarnHome.test.tsx
```diff
@@ -159,16 +160,21 @@ describe('EarnHome', () => {
...
     )
 
+    // All Pools
     expect(
       queryByTestId('PoolCard/arbitrum-sepolia:0x460b97bd498e1157530aeb3086301d5225b91216')
-    ).toBeFalsy()
+    ).toBeTruthy()
+    expect(
+      getByTestId('PoolCard/arbitrum-sepolia:0x460b97bd498e1157530aeb3086301d5225b91216')
+    ).toBeTruthy()
```
satish-ravi, line 175: seems like these are asserting the same thing, should these be different?

finnian0826: I think should be `PoolCard/ethereum-sepolia:0xe50fa9b3c56ffb159cb0fca61f5c9d750e8128c8`?

src/earn/EarnHome.test.tsx
```diff
@@ -159,16 +160,21 @@ describe('EarnHome', () => {
...
     fireEvent.press(getByText('earnFlow.poolFilters.myPools'))
+    // My Pools
```
finnian0826, line 177: Maybe should also assert here that we don't see the Eth pool

src/earn/EarnHome.test.tsx
MuckT: Fixed in737593bcfa64e72ec22cea78c982fbbf3b03ee94!

src/earn/EarnHome.test.tsx
MuckT: Added in 737593bcfa64e72ec22cea78c982fbbf3b03ee94!

satish-ravi (approved):

PR #5813, opened 2024-08-21
refactor: remove branding folder

kathaypacific:
### Description

This PR:
1. remove branding folder and move all assets to the intended `android`, `ios`, `src` directories
2. in the `images` folder, put all `.png` files in an `assets` sub folder. these images continue to be exported from `Images.ts`.
3. move "branded" components from `icons` to `images` - these are all files that use custom colours, but eventually should be all icons that are not generic (e.g. Material UI icons) 

### Test plan

n/a

### Related issues

- Fixes RET-1156

### Backwards compatibility

Y

### Network scalability


bakoushin (approved):
✨

PR #5815, opened 2024-08-21
chore(points): fetch points config more often

bakoushin:
### Description

To minimize the chances of showing stale points values during and after a promo period, the wallet now is fetching points config more often:
* on points discovery card mount
* on points home screen mount
* on pull to refresh

### Test plan

* Updated unit test

### Related issues

- Fixes RET-1179

### Backwards compatibility

Y

### Network scalability

NA


branding/valora
jeanregisser, line 1: Looks like this shouldn't have been added.

src/points/slice.ts
```diff
@@ -159,4 +159,6 @@ export const {
 // attach points event data to.
 export const trackPointsEvent = createAction<PointsEvent>('points/trackPointsEvent')
 
+export const pointsDataRefreshStarted = createAction('points/pointsDataRefreshStarted')
```
jeanregisser, line 162: nit: should we name it `refreshPointsData` instead?

jeanregisser (approved):
Great! 🚀

src/points/saga.ts
```diff
@@ -170,7 +171,10 @@ export function* getPointsConfig() {
   }
 
   yield* put(getPointsConfigStarted())
+  yield* fetchPointsConfig()
```
kathaypacific, line 174: i think that we would need to put `yield* put(getPointsConfigStarted())` inside the `fetchPointsConfig` saga...the `pointsDataRefreshStarted` action currently doesn't trigger state changes, so if for whatever reason the config CF hangs, the points home screen (which listens to `pointsConfigStatus`) won't display the loader. having said that, why do we need `fetchPointsConfig` as a separate abstraction? why not just use `getPointsConfig`?

kathaypacific (approved):
LGTM! it took a while to understand why we have this action that treats fetching config + balance + activity as a package deal, but it makes sense that we don't want to dispatch multiple actions on the discover points card but it needs to refresh all 3 things

src/points/saga.ts
kathaypacific: or maybe `pointsDataRefreshStarted` should update the states to loading?

src/points/slice.ts
bakoushin: @jeanregisser here i was following the pattern already established in that module with `tackPointsEvent` being a single exception 😅. and i believe it is following the [Redux Toolkit recommendation](https://redux-toolkit.js.org/usage/migrating-to-modern-redux#reducers-and-actions-with-createslice):
> Ideally, the naming of the reducers and actions should use the past tense and describe "a thing that happened", rather than an imperative "do this thing now", such as todoAdded instead of ADD_TODO

having that said I have no preference between using past tense or an imperative, just trying to be consistent. please let me know if you are feeling strong against the past tense. otherwise, I prefer to keep it for now.

src/points/saga.ts
bakoushin: @kathaypacific great question!
i extracted the config fetching part exactly to not display the config loader.
mainly because this particular config update is used in pull to refresh, which has its own UI to display the loading state.

my way of thinking is that the config loader makes sense until we get some config. because if we can't get any config, we can't proceed with the points at all. and we do display that loader when we request config for the first time on the [home screen mount](https://github.com/valora-inc/wallet/blob/5b451c73b1d61309bae2ca12df2f219903ffaa5e/src/points/saga.ts#L372). the subsequent config updates affect the consistency of config data, but aren't preventing us from proceeding.

PR #5821, opened 2024-08-22
feat(settings): Add Preferences Submenu

jophish:
### Description

For [RET-1180](https://linear.app/valora/issue/RET-1180/%5Bsettings%5D-create-new-settings-screen). Adds in the Preferences submenu to the new Settings screen.

### Test plan

https://github.com/user-attachments/assets/2b6ce860-2493-4300-ae1f-c6f516388e70

### Related issues

- Part of [RET-1180](https://linear.app/valora/issue/RET-1180/%5Bsettings%5D-create-new-settings-screen).

### Backwards compatibility

Yes.

### Network scalability


src/account/PreferencesSubmenu.tsx
```diff
@@ -0,0 +1,71 @@
...
+          title={t('languageSettings')}
+          value={locales[currentLanguage ?? '']?.name ?? t('unknown')}
+          onPress={goToLanguageSetting}
+          showChevron
+        />
+        <SettingsItemTextValue
+          title={t('localCurrencySetting')}
+          testID="PreferencesSubmenu/ChangeCurrency"
+          value={preferredCurrencyCode}
+          onPress={goToLocalCurrencySetting}
```
kathaypacific, line 58: should this also have `showChevron`? in the PR description video i see the chevron but confused about why 😅

src/account/PreferencesSubmenu.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+  it('shows the expected menu items', () => {
+    const store = createMockStore()
+    const { queryByTestId } = render(
+      <Provider store={store}>
+        <MockedNavigator component={PreferencesSubmenu}></MockedNavigator>
+      </Provider>
+    )
+    expect(queryByTestId('PreferencesSubmenu/Language')).toBeTruthy()
+    expect(queryByTestId('PreferencesSubmenu/ChangeCurrency')).toBeTruthy()
+    expect(queryByTestId('PreferencesSubmenu/HapticToggle')).toBeTruthy()
```
kathaypacific, line 26: same comment as before about the `getBy*` and asserting on text rather than test id where possible

kathaypacific (approved):
LGTM!!

src/account/PreferencesSubmenu.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const handleHapticFeedbackToggle = (value: boolean) => {
+    dispatch(hapticFeedbackSet(value))
+    AppAnalytics.track(SettingsEvents.settings_haptic_feedback, {
+      enabled: value,
+    })
+  }
+
+  return (
+    <ScrollView>
+      <SafeAreaView>
```
kathaypacific, line 49: ah, i think we may need to go back to the pattern of SafeAreaView with edges top only -> custom header -> Scroll view with insets bottom. otherwise the header also scrolls with the content.

PR #5823, opened 2024-08-22
refactor: create viem wallet without the need to unlock the keychain first

jeanregisser:
### Description

This changes the viem wallet client creation so it can be called without unlocking the keychain first.
Similar to how ContractKit is behaving. 

Unlocking is now necessary only when invoking an action that involves signing from the account.

This means we can now call any wallet action without manually listing it as a client extension.

Additionally, this tightens the security a bit more around the access to the private key.
With the previous implementation it was possible to bypass unlocking (after the first unlock) by accessing `client.account` directly. 

### Test plan

- Updated tests

### Related issues

- Related to RET-1182

### Backwards compatibility

Yes

### Network scalability


src/viem/getLockableWallet.test.ts
```diff
@@ -29,6 +29,9 @@ jest.mock('src/viem', () => {
   }
 })
 
+const PRIVATE_KEY1 = '0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef'
+const ACCOUNT_ADDRESS1 = privateKeyToAddress(PRIVATE_KEY1).toLowerCase() as Address
```
jeanregisser, line 33: We now use these in several tests, I'll move then to `test/values` in a follow up PR.

src/web3/KeychainLock.ts
jeanregisser, line 1: I was thinking we should rename this one to `KeychainAccounts` or something similar.
It will replace `KeychainWallet` once we remove ContractKit.

src/viem/getLockableWallet.test.ts
jeanregisser: Done in https://github.com/valora-inc/hooks/pull/501

PR #5825, opened 2024-08-22
feat(dev): add screen that shows all image assets dynamically

kathaypacific:
### Description

As the title.

### Test plan

https://github.com/user-attachments/assets/c1c3fff8-6d42-433a-9032-4696e92bf9d4

### Related issues

n/a

### Backwards compatibility

Y

### Network scalability


src/app/DebugImages.tsx
```diff
@@ -0,0 +1,105 @@
...
+}
+
+function DebugImages() {
+  return (
+    <LinearGradient
+      colors={[Colors.white, Colors.gray3]}
+      start={{ x: 0, y: 0 }}
+      end={{ x: 0, y: 1 }}
+    >
+      <ScrollView>
```
jeanregisser, line 37: nit: I thought ScrollView would struggle rendering all these views at once.

metro.config.js
```diff
@@ -25,6 +25,7 @@ const config = {
         inlineRequires: true,
       },
     }),
+    unstable_allowRequireContext: true,
```
jeanregisser, line 28: nit: maybe we should add a comment why we're enabling this?

jeanregisser (approved):
🚀

src/app/DebugImages.tsx
kathaypacific: yeah maybe there might be performance issues if running on a device, but i figure this screen will be most useful when building new features and running on our laptops. i'll keep an eye on it in the nightly and happy to update it to use flatlist or something if there are issues, but i wanted to keep it fast and simple for now

PR #5826, opened 2024-08-22
chore: consolidate and remove unnecessary icons

kathaypacific:
### Description

After implementing the debug screen, I noticed some duplicate icons. This PR removes the duplicates and replaces them with similar alternatives. There should be no visual changes in this PR.

### Test plan

n/a

### Related issues

n/a

### Backwards compatibility

Y

### Network scalability


bakoushin (approved):
🧹

PR #5827, opened 2024-08-22
chore: remove sleep dependency from celo

kathaypacific:
### Description

This PR makes one central `sleep` util function and applies it to the repo (removing the dependency for this on the `@celo` package)

### Test plan

n/a

### Related issues

- Related to RET-1173

### Backwards compatibility

Y

### Network scalability


bakoushin (approved):
🧹

PR #5828, opened 2024-08-22
test: move common private keys and related values to test/values

jeanregisser:
### Description

As the title says.

### Test plan

- Tests pass

### Related issues

- Related to RET-1182

### Backwards compatibility

Yes

### Network scalability


kathaypacific (approved):
💅🏼

jophish (approved):

PR #5829, opened 2024-08-22
feat(earn): Update earn transaction feed items to show correct provider name

finnian0826:
### Description

Gets providerId from transaction and finds pool with matching appId, uses appName.

### Test plan

Unit tests updated

### Related issues

- Part of ACT-1346

### Backwards compatibility

Yes

### Network scalability


src/transactions/feed/detailContent/EarnContent.tsx
```diff
@@ -18,18 +17,19 @@ interface EarnClaimRewardProps {
 
 export function EarnClaimContent({ transaction }: EarnClaimRewardProps) {
   const { t } = useTranslation()
-  const { providerName } = getDynamicConfigParams(
-    DynamicConfigs[StatsigDynamicConfigs.EARN_STABLECOIN_CONFIG]
-  )
+  const pools = useSelector(earnPositionsSelector)
+  const providerName = pools.find((pool) => pool.appId === transaction.providerId)?.appName
```
satish-ravi, line 21: nit: maybe add a reusable hook for this, similar to https://github.com/valora-inc/wallet/blob/9a2df7d045d39f8a0b24adc6a1721f18b9c9daa6/src/earn/hooks.ts#L97-L104

finnian0826: that makes sense! added in [cd2a03f](https://github.com/valora-inc/wallet/pull/5829/commits/cd2a03fbd2828613ce7cac1ebdc0e283501a38e4)

jh2oman (approved):

src/earn/hooks.ts
```diff
@@ -93,6 +93,15 @@ export function useAaveRewardsInfoAndPrepareTransactions({
   return { asyncRewardsInfo, asyncPreparedTransactions }
 }
 
+export function useEarnPositionProviderName(providerId: string) {
+  const pools = useSelector(earnPositionsSelector)
+  const providerName = pools.find((pool) => pool.appId === providerId)?.appName
```
MuckT, line 98: Is there any case we'd want to use a fallback provider name? But that might add to the confusion when debugging.

MuckT (approved):

src/earn/hooks.ts
finnian0826: @MuckT it looks like the provider name is used in the translation string like "Deposited to {{provider}} pool", etc. I guess a default could be "Earn", but then it wouldn't get translated. 

We could add in a fallback subtitle to show if we want because with my implementation here the subtitle is just omitted if providerName isn't found.

MuckT: It's fine to avoid a fallback with the conditional rendering as described.

PR #5836, opened 2024-08-26
refactor: rename KeychainLock -> KeychainAccounts

jeanregisser:
### Description

Rename to better convey the role of this class which manages keychain accounts.

No behavior was changed.

### Test plan

- Tests pass

### Related issues

- Part of RET-1182

### Backwards compatibility

Yes

### Network scalability


kathaypacific (approved):
✨💅🏼

jophish (approved):
:tada:

PR #5838, opened 2024-08-26
feat(earn): Add language to positions hooks request

finnian0826:
### Description

Title

### Test plan

Unit tests updated

Wasn't sure how to test e2e, since the translations files haven't been generated yet so any request to the hooks API would just use the fallback `base.json` file.

### Related issues

- Fixes ACT-1331

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/src/positions/saga.ts
+++ b/src/positions/saga.ts
@@ -3,11 +3,12 @@ import path from 'path'
 import { Alert, Platform } from 'react-native'
 import Toast from 'react-native-simple-toast'
 import { showError } from 'src/alert/actions'
-import { BuilderHooksEvents, DappShortcutsEvents } from 'src/analytics/Events'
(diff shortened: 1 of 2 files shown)


src/positions/saga.ts
```diff
@@ -72,7 +73,7 @@ async function fetchHooks<T>(
   return json.data as T
 }
 
-async function fetchPositions(hooksApiUrl: string, walletAddress: string) {
+async function fetchPositions(hooksApiUrl: string, walletAddress: string, language: string) {
```
MuckT, line 76: Can we make these an object instead?

MuckT, line 76: Can we make this function take an object instead?

MuckT (approved):

PR #5839, opened 2024-08-27
feat(settings): Add in Security submenu to new settings screen

jophish:
### Description

For [RET-1180](https://linear.app/valora/issue/RET-1180/%5Bsettings%5D-create-new-settings-screen). Adds in the Security & Privacy submenu in the new settings screen. Includes the old dev menu as normal by tapping the submenu title five times. Various small styling tweaks to a handful of flows that are accessible from the Security & Privacy screen. 

Do not be alarmed by the diff size! The new `SecuritySubmenu.tsx` and associated test file are both basically entirely ported over from `Settings.tsx`. The next PR in this (very long) chain will hook up the new Settings experience to the rest of the app, and remove our old one entirely, resulting in a lot of code gone.

### Test plan

https://github.com/user-attachments/assets/ea50b1da-3fcd-490c-ac05-6197c8c8515f

### Related issues

- Part of [RET-1180](https://linear.app/valora/issue/RET-1180/%5Bsettings%5D-create-new-settings-screen).

### Backwards compatibility

Yes.

### Network scalability


src/account/SecuritySubmenu.test.tsx
```diff
@@ -0,0 +1,347 @@
+import * as React from 'react'
```
jophish, line 1: This whole file is essentially verbatim copied over from `Settings.test.tsx`, save for some small superficial changes.

src/account/SecuritySubmenu.tsx
```diff
@@ -0,0 +1,514 @@
+import React, { useEffect, useRef } from 'react'
```
jophish, line 1: This file is essentially just copy-pasting subsets of the functionality from the existing `Settings.tsx` file, basically no "new" stuff happening here.

src/account/SecuritySubmenu.tsx
```diff
@@ -0,0 +1,532 @@
...
+          <View style={styles.devSettingsItem}>
+            <TouchableOpacity onPress={wipeReduxStore}>
+              <Text>Wipe Redux Store</Text>
+            </TouchableOpacity>
+          </View>
+          <View style={styles.devSettingsItem}>
+            <TouchableOpacity onPress={confirmAccountRemoval}>
+              <Text>App Quick Reset</Text>
+            </TouchableOpacity>
+          </View>
```
kathaypacific, line 315: i think the debug app assets option got removed here

src/account/SecuritySubmenu.tsx
```diff
@@ -0,0 +1,532 @@
...
+  )
+}
+
+const styles = StyleSheet.create({
+  header: {
+    paddingHorizontal: variables.contentPadding,
+  },
+  title: {
+    ...typeScale.labelSemiBoldMedium,
+    margin: 16,
```
kathaypacific, line 510: can we use the `Spacing` enum? there are several other magic numbers in the styles below

src/keylessBackup/SignInWithEmail.tsx
```diff
@@ -234,8 +234,8 @@ function SignInWithEmail({ route }: Props) {
             onPress={() => onPressSignIn('apple')}
             text={t('signInWithEmail.apple')}
             size={BtnSizes.FULL}
-            type={BtnTypes.PRIMARY}
-            icon={<AppleIcon color={Colors.white} />}
+            type={BtnTypes.SECONDARY}
+            icon={<AppleIcon color={Colors.black} />}
```
kathaypacific, line 238: just checking that we want to make these changes for onboarding too? kind of seems like we just restyled the onboarding flow, so i'm surprised we're changing it again 😅

src/pincode/Pincode.tsx
```diff
@@ -91,11 +91,12 @@ const styles = StyleSheet.create({
     alignItems: 'center',
   },
   subtitle: {
-    ...typeScale.bodyMedium,
+    ...typeScale.titleMedium,
     textAlign: 'center',
   },
   title: {
     ...typeScale.titleMedium,
+    lineHeight: 32,
```
kathaypacific, line 99: 32 is already the defined lineHeight for titleMedium

kathaypacific (approved):
looks good with a few nits

kathaypacific:
playing around with the whole flow in next PR, i realised it's kind of unexpected that the debug menu is inside "security & privacy". IMO the debug stuff is unrelated and maybe it's helpful to have it on the main settings screen instead to be more easily accessible. wdyt? rather than tapping on the settings title 5 times, the taps could be on the version number (like on Android)

src/account/SecuritySubmenu.tsx
```diff
@@ -0,0 +1,532 @@
...
+              </Text>
+            </>
+          }
+        />
+      )
+    }
+  }
+
+  return (
+    <SafeAreaView>
```
kathaypacific, line 379: there's some regression here that i can't exactly point to but the bottom content is colliding into the bottom bar - can we apply the usual pattern of adding a `edges={['top']}` here and adding a padding bottom to the scroll view equal to the insets.bottom?

Simulator Screenshot - iPhone 15 Pro - 2024-08-28 at 14 54 55

kathaypacific: it's also affecting the bottom sheets
Simulator Screenshot - iPhone 15 Pro - 2024-08-28 at 14 56 16

src/keylessBackup/SignInWithEmail.tsx
jophish: According to [the designs](https://www.figma.com/design/kXwIdFqrwvFOaI63rNRCAb/Mobile-Stack-Run-time?node-id=2521-14302&t=IuanyqfGijz4RUl6-4), these buttons are changed in onboarding as well, so I imagine that this change ought to exist in both places.

src/account/SecuritySubmenu.tsx
jophish: Ahh yes! Thank you for catching this. I've gone back and modified every new screen to use the correct pattern and verified that bottom sheets/scrolling etc. works as intended.

jophish:
@kathaypacific Agreed that the current location of the debug menu isn't intuitive. I've moved it over to the main Settings screen. It can be triggered by tapping repeatedly on the "Version" menu option!

PR #5840, opened 2024-08-27
feat(settings): Hook up new settings screen, remove old one

jophish:
### Description

Completes [RET-1180](https://linear.app/valora/issue/RET-1180/%5Bsettings%5D-create-new-settings-screen). Hooks up the new settings screen to the rest of the app, and removes the old one from the codebase entirely.

### Test plan

https://github.com/user-attachments/assets/ec5e53ef-7d4a-42d0-9227-154f3432da8b

### Related issues

- Fixes [RET-1180](https://linear.app/valora/issue/RET-1180/%5Bsettings%5D-create-new-settings-screen).

### Backwards compatibility

Yes.

### Network scalability


src/icons/GearIcon.tsx
```diff
@@ -0,0 +1,18 @@
...
+import Svg, { Path } from 'react-native-svg'
+import colors from 'src/styles/colors'
+
+interface Props {
+  size?: number
+  color?: string
+}
+
+const GearIcon = ({ size = 22, color = colors.black }: Props) => (
+  <Svg width={size} height={size} fill="none">
```
satish-ravi, line 11: for size parameter to work, a viewBox also needs to be specified

src/navigator/Headers.tsx
```diff
@@ -234,7 +234,7 @@ export const tabHeader: NativeStackNavigationOptions = {
   },
   headerLeft: () => (
     <View style={[styles.topElementsContainer, { marginLeft: Spacing.Tiny4 }]}>
-      <AccountCircleButton testID="WalletHome/AccountCircle" />
+      <SettingsGearButton testID="WalletHome/AccountCircle" />
```
kathaypacific, line 237:
```suggestion
      <SettingsGearButton testID="WalletHome/SettingsGearButton" />
```

src/icons/GearIcon.tsx
kathaypacific: +1

kathaypacific (approved):
LGTM, would be good to address the small comments before merging! 🚀

PR #5841, opened 2024-08-27
feat: make quick actions configurable

jeanregisser:
### Description

As the title says.

### Test plan

- Updated tests

| Only send | No actions |
|-|-|
| image | image |

### Related issues

- Fixes RET-1144

### Backwards compatibility

Yes

### Network scalability


kathaypacific (approved):
🚀

should we also create an env.example? or add the config values to the existing env files? (there are a lot of them though...) i was thinking it'd be nice not need to dig into the code (the `HomeActionName` enum) to know which values are accepted in the config values, so providing all the values upfront somewhere central would be nice

jeanregisser:
Yeah maybe .env.example could be used to document all config?

kathaypacific:
> Yeah maybe `.env.example` could be used to document all env config?

yes i think that makes the most sense, we could also do it in README or in Notion...although none of these options give me confidence that we'll continue to update these docs when there are changes to the env variables or values 😅 i don't have a good solution though, so the env.example is probably a good place to start

jeanregisser:
Same feeling yes.

src/config.ts
```diff
@@ -193,3 +194,13 @@ export const LOGGER_LEVEL = Config.LOGGER_LEVEL
...
+
+export const ENABLED_QUICK_ACTIONS = (
+  Config.ENABLED_QUICK_ACTIONS ??
+  // Default to all actions
+  Object.values(HomeActionName).join(',')
+)
+  .split(',')
+  .filter(
+    (value) => !!value && Object.values(HomeActionName).includes(value as HomeActionName)
+  ) as HomeActionName[]
```
bakoushin, line 206: i wonder if we can use Zod to parse the config as we do in points-functions repo? to ensure the config has the correct values

PR #5843, opened 2024-08-27
refactor: remove direct dependency on ethereumjs-util

jeanregisser:
### Description

As the title says.

This was using for sha256 hashing, but we can use viem's implementation instead.

### Test plan

- Updated tests
- Manually tested unlocking via PIN still work for existing accounts

### Related issues

- N/A

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/src/analytics/AppAnalytics.ts
+++ b/src/analytics/AppAnalytics.ts
@@ -3,7 +3,6 @@ import { AdjustPlugin } from '@segment/analytics-react-native-plugin-adjust'
 import { ClevertapPlugin } from '@segment/analytics-react-native-plugin-clevertap'
 import { DestinationFiltersPlugin } from '@segment/analytics-react-native-plugin-destination-filters'
 import { FirebasePlugin } from '@segment/analytics-react-native-plugin-firebase'
-import { sha256FromString } from 'ethereumjs-util'
 import _ from 'lodash'
 import { Platform } from 'react-native'
 import DeviceInfo from 'react-native-device-info'
@@ -28,6 +27,7 @@ import { getSupportedNetworkIdsForTokenBalances } from 'src/tokens/utils'
 import { ensureError } from 'src/utils/ensureError'
 import Logger from 'src/utils/Logger'
 import { Statsig } from 'statsig-react-native'
+import { sha256 } from 'viem'
 
 const TAG = 'AppAnalytics'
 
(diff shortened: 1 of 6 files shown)


src/pincode/authentication.ts
```diff
@@ -164,10 +164,15 @@ async function getPasswordHashForPin(pin: string) {
   return getPasswordHash(password)
 }
 
-function getPasswordHash(password: string) {
-  return sha256(Buffer.from(password, 'hex')).toString('hex')
+// TODO: this existing implementation implies password is in hex (no '0x' prefix)
+// but we should lift that restriction as it's too easy to misuse
```
jeanregisser, line 168: We can change the behavior in a future PR.
It requires resetting the stored hashes, so existing accounts can still unlock.

src/pincode/authentication.test.ts
```diff
@@ -649,3 +650,20 @@ describe(checkPin, () => {
...
+  // Because the current implementation treats the input as hex string (without the '0x' prefix)
+  // But we'll change this to treat the input as a raw string/buffer in the future
+  it.each([
+    ['123456', 'bf7cbe09d71a1bcc373ab9a764917f730a6ed951ffa1a7399b7abd8f8fd73cb4'],
+    ['0x123456', 'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'],
+    ['0x1234567', 'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'],
+    ['abx', '087d80f7f182dd44f184aa86ca34488853ebcc04f0c60d5294919a466b463831'],
+    ['abxy', '087d80f7f182dd44f184aa86ca34488853ebcc04f0c60d5294919a466b463831'],
+    ['ABXY', '087d80f7f182dd44f184aa86ca34488853ebcc04f0c60d5294919a466b463831'],
+    ['0xabxy', 'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'],
```
jeanregisser, line 665: These hashes were captured using the implementation before switching to viem.
So we keep the same behavior.

src/analytics/AppAnalytics.ts
```diff
@@ -125,9 +125,7 @@ class AppAnalytics {
         const deviceInfo = await getDeviceInfo()
         this.deviceInfo = deviceInfo
         uniqueID = deviceInfo.UniqueID
-        this.sessionId = sha256FromString('0x' + uniqueID.split('-').join('') + String(Date.now()))
-          .toString('hex')
-          .slice(2)
+        this.sessionId = sha256(Buffer.from(uniqueID + String(Date.now()))).slice(2)
```
kathaypacific, line 128: do we still need to remove the hypens from `uniqueID`?

kathaypacific (approved):
LGTM!

src/analytics/AppAnalytics.ts
jeanregisser: No it's not needed anymore

PR #5847, opened 2024-08-27
feat(earn): get terms url from positions response

satish-ravi:
### Description

As the title

### Test plan

Manually by going through the deposit flow

### Related issues

Part of ACT-1342

### Backwards compatibility

Yes

### Network scalability


locales/base/translation.json
```diff
@@ -2568,11 +2568,13 @@
     "depositBottomSheet": {
       "title": "Review & Deposit",
       "description": "Before depositing to an Aave pool, please review the transaction details below",
+      "descriptionV1_93": "Before depositing to a {{providerName}} pool, please review the transaction details below",
```
finnian0826, line 2571: are we okay with it saying "... a Aave pool..." rather than "... an Aave pool..."? Not sure how to prevent this.

finnian0826 (approved):
One question but otherwise lgtm!

locales/base/translation.json
satish-ravi: chatted with John and decided to go with `an` for now for the 2 pools. We can revisit this when we add more pools and should be something we can fix OTA

PR #5848, opened 2024-08-28
feat: total deposit and earnings card

MuckT:
### Description

Adds the Total Deposit & Earnings Card on `EarnPoolInfoScreen.tsx`. Also adds a new hook `useTokensInfo` which returns an object with tokenInfo for the specified tokenIds.

| iOS AllBridge | Android AllBridge |
| ----- | ----- |
|  | |

### Test plan

- [x] Tested locally on iOS with and without active pools
- [x] Tested locally on Android with and without active pools
- [x] Unit tests updated

### Related issues

- Fixes ACT-1322

### Backwards compatibility

Yes

### Network scalability

Yes - This should continue to work with new networks supported by hooks.


locales/base/translation.json
```diff
@@ -2657,7 +2657,11 @@
       "ageOfPool": "Age of Pool",
       "learnMoreOnProvider": "View Pool on {{providerName}}",
       "deposit": "Deposit",
-      "withdraw": "Withdraw"
+      "withdraw": "Withdraw",
+      "totalDepositAndEarnings": "Total Deposit & Earnings",
+      "earnings": "Earnings",
```
finnian0826, line 2662: Is this `earnings` key being used? I couldn't find any changes here that use it.

src/earn/EarnPoolInfoScreen.tsx
```diff
@@ -106,14 +109,127 @@ function TokenIcons({
...
+  // Total balance in local currency
+  const depositBalanceInUsd = tokenInfo?.priceUsd?.multipliedBy(balance)
+  const depositBalanceInLocalCurrency = new BigNumber(localCurrencyExchangeRate ?? 0).multipliedBy(
+    depositBalanceInUsd ?? 0
+  )
+  const earningItemsTokenIds = earningItems.map((item) => item.tokenId)
+  const earningItemsTokenInfo = useTokensInfo(earningItemsTokenIds)
+  const totalBalanceInLocalCurrency = depositBalanceInLocalCurrency.plus(
+    earningItems.reduce((acc, item) => {
+      if (!item.subtractFromDeposit) return acc
```
finnian0826, line 176:
```suggestion
      if (item.subtractFromDeposit) return acc
```
I think this should be the other way around. If `subtractFromDeposit` is set, that means that it is already included in the pool balance (probably should have called it subtractFromPoolBalance). So when getting the total Deposit & Earnings, it should be pool balance plus all earning items where this is not set (i.e. add in all external rewards).

src/earn/EarnPoolInfoScreen.tsx
```diff
@@ -106,14 +109,127 @@ function TokenIcons({
...
+          <View style={styles.cardLineLabel}>
+            <Text numberOfLines={1} style={styles.depositAndEarningsCardLabelText}>
+              {t('earnFlow.poolInfoScreen.deposit')}
+            </Text>
+          </View>
+          <View style={styles.flexShrink}>
+            <Text style={styles.depositAndEarningsCardValueText}>
+              {t('earnFlow.poolInfoScreen.lineItemAmountDisplay', {
+                localCurrencySymbol,
+                localCurrencyAmount: formatValueToDisplay(depositBalanceInLocalCurrency),
```
finnian0826, line 218: For this value you should subtract out any earning items where `subtractFromDeposit` is true. What you have here looks to just be the pool balance, but if there is an earning line item for the compounded interest (which is included in pool balance), it should be subtracted out so that the value is not double counted.

src/earn/EarnPoolInfoScreen.tsx
```diff
@@ -106,14 +109,127 @@ function TokenIcons({
...
+                cryptoAmount: formatValueToDisplay(totalBalanceInLocalCurrency),
```
finnian0826, line 219: This should also be adjusted like the fiat amount above.

src/earn/EarnPoolInfoScreen.tsx
```diff
@@ -107,14 +110,156 @@ function TokenIcons({
...
+    }, new BigNumber(0))
+  )
+
+  const totalDepositBalanceInCrypto = new BigNumber(balance).minus(
+    earningItems.reduce((acc, item) => {
+      // Don't include earning items that are not part of the pool balance or are not the deposit token
+      if (!item.includedInPoolBalance || item.tokenId !== depositTokenId) return acc
+      return acc.plus(item.amount)
+    }, new BigNumber(0))
+  )
```
finnian0826, line 208:
```suggestion
  const totalDepositBalanceInCrypto = totalDepositBalanceInLocalCurrency.dividedBy(new BigNumber(localCurrencyExchangeRate ?? 1)).dividedBy(new BigNumber(tokenInfo?.priceUsd ?? 1))
```
I think what you have now would cause an issue if `includedInPoolBalance` is true but `item.tokenId !== depositTokenId`. In that case, we wouldn't subtract it, but we would subtract it for the fiat calculation in `totalDepositBalanceInLocalCurrency`, so the line item would show `fiat (crypto)` but the crypto amount would actually be more than the fiat amount converts to. 

For example, if there is an earningItem where token is USDT and balance is 5, and then the poolBalance is 25 and token is USDC, for the fiat calculation (`totalDepositBalanceInLocalCurrency`) we would do $25 - $5 = $20, but for crypto, since tokens don't match we wouldn't subtract so `totalDepositBalanceInCrypto` would be set to 25. Then for the line item we would show `$20 (25 USDC)`.

I added a suggestion here of what we could do instead.

MuckT: Fixed in 970dd5de7488c541c35d88fca14fb4aa507626db!

locales/base/translation.json
MuckT: Removed in 6e8378565de6bbde237606194c87e703ed26abcd!

src/earn/EarnPoolInfoScreen.tsx
MuckT: Fixed in 9d050848e370f4a61f5a3cc513c262ea78a7ecb9!

src/earn/EarnPoolInfoScreen.tsx
MuckT: Fixed in 9d050848e370f4a61f5a3cc513c262ea78a7ecb9!

src/earn/EarnPoolInfoScreen.tsx
MuckT: Fixed in 9d050848e370f4a61f5a3cc513c262ea78a7ecb9!

src/earn/EarnPoolInfoScreen.test.tsx
```diff
@@ -56,13 +65,82 @@ describe('EarnPoolInfoScreen', () => {
...
-    expect(getByText('earnFlow.poolInfoScreen.deposit')).toBeTruthy()
+    expect(
+      within(getByTestId('ActionButtons')).getByText('earnFlow.poolInfoScreen.withdraw')
+    ).toBeTruthy()
+    expect(
+      within(getByTestId('ActionButtons')).getByText('earnFlow.poolInfoScreen.deposit')
+    ).toBeTruthy()
+  })
+
+  it('renders deposit and earnings card when in pool when deposited in pool', () => {
```
finnian0826, line 76: could add a test where `includedInPoolBalance` is true to test that it doesn't double count it.

finnian0826, line 76: could also add test for when `cantSeparateCompoundedInterest` is true to make sure correct label is shown

finnian0826 (approved):
lgtm, just left two comments for possible tests that could be added

locales/base/translation.json
```diff
@@ -2660,7 +2660,11 @@
       "ageOfPool": "Age of Pool",
       "learnMoreOnProvider": "View Pool on {{providerName}}",
       "deposit": "Deposit",
-      "withdraw": "Withdraw"
+      "depositAndEarnings": "Deposit & Earnings",
+      "withdraw": "Withdraw",
+      "totalDepositAndEarnings": "Total Deposit & Earnings",
+      "titleLocalAmountDisplay": "{{localCurrencySymbol}}{{localCurrencyAmount}}",
+      "lineItemAmountDisplay": "{{localCurrencySymbol}}{{localCurrencyAmount}} ({{cryptoAmount}} {{cryptoSymbol}})"
```
satish-ravi, line 2667: these don't seem to have any language specific stuff that needs to be localized. Can we just do this concatenation in code?

src/earn/EarnPoolInfoScreen.test.tsx
```diff
@@ -7,20 +7,27 @@ import EarnPoolInfoScreen from 'src/earn/EarnPoolInfoScreen'
...
 import { Screens } from 'src/navigator/Screens'
 import { navigateToURI } from 'src/utils/linking'
+import networkConfig from 'src/web3/networkConfig'
 import MockedNavigator from 'test/MockedNavigator'
 import { createMockStore, getMockStackScreenProps } from 'test/utils'
-import { mockEarnPositions } from 'test/values'
+import { mockArbUsdcTokenId, mockEarnPositions, mockTokenBalances } from 'test/values'
 
-const defaultStore = createMockStore({})
+const mockPoolTokenId = networkConfig.aaveArbUsdcTokenId
```
satish-ravi, line 15: can we get this from mockEarnPositions? we are moving away from using networkConfig with the multi pool work

src/tokens/hooks.ts
```diff
@@ -133,6 +133,14 @@ export function useTokenInfo(tokenId?: string): TokenBalance | undefined {
   return tokenId ? tokens[tokenId] : undefined
 }
 
+export function useTokensInfo([...tokenIds]: string[]): (TokenBalance | undefined)[] {
```
satish-ravi, line 136:
```suggestion
export function useTokensInfo(tokenIds: string[]): (TokenBalance | undefined)[] {
```
can we just do this^?

src/earn/EarnPoolInfoScreen.tsx
```diff
@@ -107,14 +110,162 @@ function TokenIcons({
...
+  const depositBalanceInUsd = tokenInfo?.priceUsd?.multipliedBy(balance)
+  const depositBalanceInLocalCurrency = new BigNumber(localCurrencyExchangeRate ?? 0).multipliedBy(
+    depositBalanceInUsd ?? 0
+  )
+
+  // Earning items used to calculate the total balance and total deposited
+  const earningItemsTokenIds = earningItems.map((item) => item.tokenId)
+  const earningItemsTokenInfo = useTokensInfo(earningItemsTokenIds)
+
+  const totalBalanceInLocalCurrency = depositBalanceInLocalCurrency.plus(
```
satish-ravi, line 178: consider using useMemo's for these computations.

src/earn/EarnPoolInfoScreen.tsx
```diff
@@ -107,14 +110,162 @@ function TokenIcons({
...
+    earningItems.reduce((acc, item) => {
+      if (item.includedInPoolBalance) return acc
```
satish-ravi, line 180: nit: consider using filter + reduce to make this more readable, my first assumption was the two reduces were doing the same summation. Also could make this more DRY by making a small function to sum a list of earningItems

src/earn/EarnPoolInfoScreen.tsx
```diff
@@ -107,14 +110,162 @@ function TokenIcons({
...
+      )
+      return acc.plus(amountInLocalCurrency ?? 0)
+    }, new BigNumber(0))
+  )
+
+  const totalDepositBalanceInCrypto = new BigNumber(balance).minus(
+    earningItems.reduce((acc, item) => {
+      if (!item.includedInPoolBalance) return acc
+      const itemToken = earningItemsTokenInfo.find((token) => token?.tokenId === item.tokenId)
+      return acc.plus(
```
satish-ravi, line 206: can we add test coverage for these?

locales/base/translation.json
MuckT: Currency symbols can be after the amount in other languages, I am not sure if this would be the case for crypto symbols, but I decided to air on the side of caution.

src/earn/EarnPoolInfoScreen.test.tsx
MuckT: Changed in /3bf06c071bd628289bfe6b16f48ad3b38f6fc613!

src/tokens/hooks.ts
MuckT: Changed in 60bb96464c22d6311184c76ff7077b999aa700fc!

src/earn/EarnPoolInfoScreen.tsx
MuckT: Used filter and reduce in 76f8603dbc40cde827d8bbf40538dd9900cc5fa2!

src/earn/EarnPoolInfoScreen.tsx
MuckT: I've wrapped these total calculations with a useMemo in f5cc57dca44fe818a41253905b7945fc0006a334!

src/earn/EarnPoolInfoScreen.test.tsx
MuckT: Added additional tests in e00254f5cf5b796566f5a435127bb8bbbb765c8c!

MuckT: Added additional tests in e00254f5cf5b796566f5a435127bb8bbbb765c8c!

src/earn/EarnPoolInfoScreen.tsx
MuckT: These should all have coverage now from the tests added in e00254f5cf5b796566f5a435127bb8bbbb765c8c.

PR #5852, opened 2024-08-28
chore: copy used code from @celo/phone-utils into the repo and remove @celo/phone-utils dependency

sviderock:
### Description
Copy used functions/types from [@celo/phone-utils](https://www.npmjs.com/package/@celo/phone-utils) and also depending functions/types:
- E164Number
- E164PhoneNumberType
- isE164NumberStrict
- Countries
- LocalizedCountry
- getCountryEmoji
- getRegionCode
- getRegionCodeFromCountryCode
- getDisplayNumberInternational
- getPhoneHash from `@celo/phone-utils/lib/getPhoneHash`
- validateInput
- parsePhoneNumber

PhoneNumberUtils was not copied as it is just an object exporting some functions from the list above and the only function used from it was `getPhoneHash` so it was changed to use the function directly.

For `getPhoneHash` a legacy implementation of `getIdentificationHash` dependency was used: 
https://github.com/celo-org/developer-tooling/blob/8572a0f978d1aa01a36775ef4be48c3eafdbb204/packages/sdk/base/src/identifier.ts
Latest version of implementation:
https://github.com/celo-org/social-connect/blob/main/packages/odis-identifiers/src/identifier.ts
P.S. The only difference between the two is `IdentifierPrefix` having an extra `GITHUB` enum value in the latest version.

#### New dependencies:
- [google-libphonenumber](https://github.com/ruimarinho/google-libphonenumber)
- [io-ts](https://github.com/gcanti/io-ts)
- [fp-ts](https://github.com/gcanti/fp-ts)

### Test plan
All tests pass with the new copied functions.

### Related issues

- Relates to RET-1186

### Network scalability


src/utils/indetifier.ts
jeanregisser, line 1: Typo `identifier`

src/country.json
jeanregisser, line 1: Could we source this from an existing dependency?
I see we already have `country-data`, so maybe?

src/utils/Countries.ts
```diff
@@ -0,0 +1,116 @@
+// more countries @ https://github.com/umpirsky/country-list
+import countryData from 'country-data'
+import esData from 'src/country.json'
```
jeanregisser, line 3: I wonder if we could remove this specific one 🤔

jeanregisser (commented):
Awesome work! 🚀 

Could you also add the tests (when they were present) and attribute each copied file to the original file (maybe with the immutable GH link?)

src/country.json
sviderock: @jeanregisser it was used from their internal file `@celo/phone-utils/data/es/country.json`.
I can take a look and if its identical then I'll use it from `country-data`.

src/utils/Countries.ts
sviderock: @jeanregisser Yeah, I'll look into it!

sviderock:
@jeanregisser updated as per the comments

src/utils/phoneNumber.test.ts
jeanregisser, line 1: nit: noting this test file isn't linked to a source file anymore. Like the rest of the unit tests.
Would be nice to split or group the functions that were originally in `phoneNumbers.ts`?

jeanregisser (commented):
Great!

src/utils/phoneNumber.test.ts
sviderock: @jeanregisser the only reason I've kept everything in one file is that those constants are getting used across all the functions. So it's whether splitting into multiple test files and copy/pasting constants or multiple test files and importing constanst from single test file. Would you prefer any of those?

Thought I've forgotten to add comment with a link referenced file at the top, will add that.

sviderock: Although, there's a lint warning for exporting from test file:
Screenshot 2024-08-29 at 15 52 23

jeanregisser: yes since it's seems it's an easy path.

sviderock: @jeanregisser Changes commited!

src/utils/getPhoneHash.ts
```diff
@@ -0,0 +1,29 @@
...
+import { soliditySha3 } from 'web3-utils'
+
+/**
+ * Rerefence implementation uses "web3-utils" version 1.3.6:
+ * https://github.com/celo-org/developer-tooling/blob/8572a0f978d1aa01a36775ef4be48c3eafdbb204/packages/sdk/base/package.json#L30
+ *
+ * In that version the function "soliditySha3" returns "string | null" while the latest version of "web3-utils"
+ * returns "string | undefined". For the sake of compatibility with the original implementation while
+ * also using the latest version of "web3-utils" "?? null" was added to replace possible undefined with null.
+ */
```
jeanregisser, line 19: Nothing to do in this PR, but it looks like sha3 could be replaced with viem's keccak256.

https://github.com/web3/web3.js/blob/9b322052ef266cc57b29080b20053991e0c9a0a9/packages/web3-utils/src/hash.ts#L125

jeanregisser (approved):
Amazing! 🙌

src/utils/getPhoneHash.ts
jeanregisser: Cool, there's a direct replacement with `@noble/hashes/sha3`
And `@noble/hashes` is already used by viem.

PR #5853, opened 2024-08-28
chore: remove max send amount hook and fees reducer

kathaypacific:
### Description

This PR removes the estimated max amount in the FiatExchangeAmount component, and sets the max amount to the user's token balance. The max amount estimation only worked for Celo and was inaccurate. Removing this also allowed removing:
- some unused hooks
- the whole `fees` redux store

The only user flow impacted by this change is the withdraw flow on Celo (since there was effectively no fee estimation for other networks before anyway), specifically when a user tries to cash out without leaving enough balance to pay for gas - most likely when they only have 1 type of token. For a user in this flow, they will reach the end of the withdraw flow and not be able to proceed. I'll add an error state separately.

### Test plan

n/a

### Related issues

- Related to RET-1183

### Backwards compatibility

Y

### Network scalability


bakoushin (approved):
LGTM! 🧹

PR #5857, opened 2024-08-28
refactor(earn): parameterize gas subsidy by network ids

satish-ravi:
### Description

The wallet only supports gas subsidy for arbitrum. 

### Test plan

CI

### Related issues

- Part of ACT-1342

### Backwards compatibility

Yes

### Network scalability


src/earn/utils.ts
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Helper function to check if gas is subsidized for the given networkId
+ * @param networkId - networkId to check
+ * @returns boolean
+ */
+export function isGasSubsidizedForNetwork(networkId: NetworkId) {
+  return (
+    (INTERNAL_RPC_SUPPORTED_NETWORKS as readonly Network[]).includes(
+      networkIdToNetwork[networkId]
+    ) && getFeatureGate(StatsigFeatureGates.SUBSIDIZE_STABLECOIN_EARN_GAS_FEES)
```
satish-ravi, line 16: We can eventually do this via dynamic config, but this is good enough for now

jh2oman: Is there a plan for which ticket we will take care of it? just don't want to forget

jh2oman (approved):

src/earn/utils.ts
satish-ravi: no ticket as of now. Until we plan to support this for other networks, which will need more wallet changes, this can wait.

PR #5858, opened 2024-08-28
refactor(earn): connect pool details to deposit screen

satish-ravi:
### Description

Also fixes a small bug in earn prepare transactions with amount units.

### Test plan

CI

### Related issues

- Part of ACT-1342

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/src/earn/prepareTransactions.ts
+++ b/src/earn/prepareTransactions.ts
@@ -98,7 +98,7 @@ export async function prepareSupplyTransactions({
     feeCurrencies,
     baseTransactions,
     spendToken: token,
-    spendTokenAmount: new BigNumber(amount),
+    spendTokenAmount: new BigNumber(amount).shiftedBy(token.decimals),
     isGasSubsidized,
     origin: 'earn-deposit',
   })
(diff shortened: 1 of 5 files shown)


src/earn/prepareTransactions.ts
```diff
@@ -98,7 +98,7 @@ export async function prepareSupplyTransactions({
     feeCurrencies,
     baseTransactions,
     spendToken: token,
-    spendTokenAmount: new BigNumber(amount),
+    spendTokenAmount: new BigNumber(amount).shiftedBy(token.decimals),
```
satish-ravi, line 101: this is only an issue if the spendToken is also the gas currency. For aave, this can never be the case so there was no user impact

src/viem/prepareTransactions.ts
```diff
@@ -268,7 +268,7 @@ export async function tryEstimateTransactions(
  *
  * @param feeCurrencies
  * @param spendToken
- * @param spendTokenAmount
+ * @param spendTokenAmount BigNumber in smallest unit
```
jh2oman, line 271: what exactly does smallest unit mean?

jh2oman (approved):

src/viem/prepareTransactions.ts
satish-ravi: the smallest unit of the token (e.g., 1ETH => 1e18, 1 usdc => 1e6). I believe we use this term in a few places in the wallet

PR #5860, opened 2024-08-29
feat(earn): use shortcut for deposit tx

satish-ravi:
### Description

Replace the aave supply transaction preparation with the deposit shortcut hook. This enables executing shortcuts for any pool

### Test plan

Manually by going through the deposit flow for different pools

#### Aave on single pool flow
https://github.com/user-attachments/assets/a70b125c-6fad-4c97-b09d-53436741041b

#### Allbridge on multiple pool flow
https://github.com/user-attachments/assets/202dd31e-168f-437c-a8a3-78f05e3176e1

### Related issues

- Fixes ACT-1342

### Backwards compatibility

Yes

### Network scalability


src/earn/saga.ts
```diff
@@ -117,61 +118,70 @@ export function* depositSubmitSaga(action: PayloadAction<DepositInfo>) {
...
+        feeCurrencyId?: string
+      ): BaseStandbyTransaction => {
+        return {
+          context: newTransactionContext(TAG, 'Earn/Deposit'),
+          __typename: 'EarnDeposit',
+          networkId,
+          type: TokenTransactionTypeV2.EarnDeposit,
+          inAmount: {
+            value: amount,
+            tokenId: networkConfig.aaveArbUsdcTokenId,
```
jeanregisser, line 169: Should we use `earnPosition.withdrawTokenId` instead?
Though it's possible it could be wrong in the future, with new type of pools.

jeanregisser (approved):
Awesome! 🚀

src/earn/saga.ts
satish-ravi: oh good catch, I think we can also use the positionId? maybe that's safer that withdrawTokenId?

jeanregisser: Ah yes but the positionId only works in some cases, if the position is an app-token.
For instance it wouldn’t work if we had an earn position for staking celo natively which is modeled using a contract-position as there’s not another erc20 token the user gets back representing it.

withdrawTokenId seems safer then.

jeanregisser: In the future I wonder if we should just show this differently in the feed so we don’t need in and out which don’t apply in all cases.

jh2oman (approved):

PR #5861, opened 2024-08-29
refactor: copy account module from @celo/cryptographic-utils@3.2.0

jeanregisser:
### Description

This is the first step to remove `@celo/cryptographic-utils`.

It copies the [accounts](https://github.com/celo-org/celo-monorepo/blob/sdks-3.2.0/packages/sdk/cryptographic-utils/src/account.ts) module from it.

I've copied the exact same version used from the celo SDK: `3.2.0`.
And listed the exact same NPM packages it was using internally.
This is to ensure it behaves the same.
We'll upgrade/remove these in a follow up PR.

I've slightly modified it to directly include some sub modules, which we didn't need to share.
- `@celo/base/lib/account` (some type definitions)
- `@celo/base/lib/string` (only 1 function used)

### Test plan

- Tests pass

### Related issues

- Part of RET-1185

### Backwards compatibility

Yes

### Network scalability


src/utils/account.ts
```diff
@@ -0,0 +1,495 @@
+// Initially copied from https://github.com/celo-org/celo-monorepo/blob/sdks-3.2.0/packages/sdk/cryptographic-utils/src/account.ts
```
jophish, line 1: Is any part of this file meaningfully modified from the original? What of this ought to be reviewed, if any?

jophish (approved):
LGTM, assuming that the contents of `account.ts` and `account.test.ts` are basically copied verbatim (which, from a brief comparison to the originals, it looks like they are).

src/utils/account.ts
jeanregisser: Nothing was changed except some imports.

PR #5865, opened 2024-08-30
refactor: use react-native-quick-crypto

jeanregisser:
### Description

Use https://github.com/margelo/react-native-quick-crypto to provide fast and secure crypto operations.

Note: We'll be able to gradually remove the following libs which are now redundant:
- `react-native-bip39`
-  `react-native-fast-crypto`
- `react-native-securerandom`
- `node-libs-react-native` and its own native deps (once ContractKit is removed)
- `react-native-crypto`
- `react-native-randombytes`

### Test plan

- Tests pass
- Manually
  -  confirmed mnemonic generation and key derivation is as fast as before and uses the native implementation from the new library, on both iOS and Android
  - importing some existing mnemonics works as before

### Related issues

- Part of RET-1185

### Backwards compatibility

Yes

### Network scalability

### DIFF
--- a/__mocks__/react-native-quick-crypto.ts
+++ b/__mocks__/react-native-quick-crypto.ts
@@ -1,0 +1,2 @@
+// We can just use the actual node crypto module here
+module.exports = jest.requireActual('crypto')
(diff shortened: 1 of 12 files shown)


bakoushin (approved):
🚀

jophish (approved):
🎉

PR #5866, opened 2024-08-30
refactor: remove crypto-js which is now deprecated

jeanregisser:
### Description

This removes [crypto-js](https://github.com/brix/crypto-js) which is now discontinued:
> Active development of CryptoJS has been discontinued. This library is no longer maintained.
>
> Nowadays, NodeJS and modern browsers have a native Crypto module. The latest version of CryptoJS already uses the native Crypto module for random number generation, since Math.random() is not crypto-safe. Further development of CryptoJS would result in it only being a wrapper of native Crypto. Therefore, development and maintenance has been discontinued, it is time to go for the native crypto module.

Bonus, because it now uses the native crypto module (via [react-native-quick-crypto](https://github.com/margelo/react-native-quick-crypto)) AES encryption/decryption should be faster. Though the gains are probably not noticeable given the small use we have.

Note: there's an important change for tests too, they now all use real encryption.
- I feel this is less surprising than what we used to do.
- I was able to refactor tests that expected mocked encryption without too much trouble

### Test plan

- Updated tests (ensuring backward compat with CryptoJS encrypted strings)
- Manually tested existing keychain items encrypted by CryptoJS can still be decrypted in Valora

### Related issues

- Part of RET-1185

### Backwards compatibility

Yes

### Network scalability


src/utils/aes.ts
```diff
@@ -0,0 +1,46 @@
+// This file was copied from https://github.com/RaisinTen/aes-crypto-js/blob/2978af8e004d47539d767e751def003fe134b6e2/index.js
+// and modified slightly for TS compatibility.
```
jeanregisser, line 2: I chose to copy the implementation instead of importing it from NPM as it's critical to the correct behavior of Valora.
And wanted to avoid yet another unmaintained package in the future 😄

src/viem/getLockableWallet.test.ts
```diff
@@ -19,9 +19,6 @@ import {
 } from 'viem/actions'
 import { celoAlfajores, goerli as ethereumGoerli, sepolia as ethereumSepolia } from 'viem/chains'
 
-// Use real encryption
-jest.unmock('crypto-js')
```
jeanregisser, line 23: Real encryption is now the default, since `__mocks__/crypto-js.ts` was removed, which was mocking encrypt and decryption.

package.json
```diff
@@ -124,7 +124,6 @@
     "bignumber.js": "^9.1.2",
     "clevertap-react-native": "^2.2.1",
     "country-data": "^0.0.31",
-    "crypto-js": "^4.2.0",
```
jophish, line 127: 🎉

src/utils/aes.ts
jophish: I'm sure this is all probably totally fine, but something about checking in code that operates at this low-level of abstraction makes me nervous 😅 This is all totally backwards compatible with data that's already been stored and encrypted by our previous library, right? That is, you can encrypt data the old way, decrypt it the new way, and it will decrypt as we'd expect?

src/pincode/authentication.test.ts
```diff
@@ -51,6 +52,16 @@ jest.mock('src/utils/sleep', () => ({
   sleep: jest.fn().mockResolvedValue(true),
 }))
 
+jest.mock('src/utils/aes', () => {
+  const originalAes = jest.requireActual('src/utils/aes')
+
+  return {
+    ...originalAes,
```
jophish, line 59: Why do we need to return the original import here? It seems like the only two functions in that file are `aesEncrypt` and `aesDecrypt`, which we're mocking anyways.

jophish (approved):
Looks fantastic! Glad it was an easy migration, and happy to get rid of this :)

PR #5867, opened 2024-08-30
fix(earn): Update providerId for cached earn txs

finnian0826:
### Description

Updated providerId for earn transactions from `aave-v3` to `aave` in the backend, but it didn't update the cached transactions. This should update the providerId for cached earn transactions stored in root state.

### Test plan

Ran app locally and confirmed errors went away, pool provider name showed correctly

### Related issues

- Fixes N/A

### Backwards compatibility

Yes

### Network scalability


src/redux/migrations.ts
```diff
@@ -1861,4 +1861,23 @@ export const migrations = {
     ...state,
     account: _.omit(state.account, 'pictureUri'),
   }),
+  228: (state: any) => {
```
satish-ravi, line 1864: can we add a unit test for this since this doesn't seem like a straightforward migration?

finnian0826: unit test added in [e005b33](https://github.com/valora-inc/wallet/pull/5867/commits/e005b333701d71b6e2cac1c1aee0a27b640ee957)

src/redux/migrations.ts
```diff
@@ -1861,4 +1861,23 @@ export const migrations = {
...
+    const transactionsByNetworkId: any = {}
+    for (const networkId of Object.keys(state.transactions.transactionsByNetworkId)) {
```
satish-ravi, line 1866: should we do this only for arb networks?

src/redux/migrations.ts
```diff
@@ -1861,4 +1861,23 @@ export const migrations = {
...
+      const transactions = []
+      for (const tx of state.transactions.transactionsByNetworkId[networkId]) {
+        tx.providerId && tx.providerId === 'aave-v3'
```
satish-ravi, line 1869: I guess only earn transactions can have providerId, should we check those?

satish-ravi (commented):
Left a couple of optional optimizations, not strictly necessary I think. And have you tested this against a wallet with cached transactions? Didn't see it in the test plan

src/redux/migrations.ts
finnian0826: I was using the check for `tx.providerId` as a proxy for checking that it was an earn transaction, but I could add an explicit check for the type of transaciton.

src/redux/migrations.ts
```diff
@@ -1877,4 +1877,28 @@ export const migrations = {
       'mtwAddress'
     ),
   }),
+  229: (state: any) => {
+    const transactionsByNetworkId: any = {}
+    for (const networkId of Object.keys(state.transactions.transactionsByNetworkId)) {
+      const transactions = []
+      for (const tx of state.transactions.transactionsByNetworkId[networkId]) {
+        ;(networkId === NetworkId['arbitrum-one'] || networkId === NetworkId['arbitrum-sepolia']) &&
```
MuckT, line 1885: nit: code golf
`[NetworkId['arbitrum-one'], NetworkId['arbitrum-sepolia']].includes(networkId) &&`

MuckT (approved):

src/redux/migrations.ts
finnian0826: @MuckT this gives me a type error since `const networkId of Object.keys(state.transactions.transactionsByNetworkId)` is being typed as a string

PR #5877, opened 2024-09-02
chore: copy used code from @celo/utils into the repo and remove @celo/utils dependency

sviderock:
##### No Equivalents:
- `hexToBuffer` from `@celo/utils/lib/address`
- `ensureLeading0x` from `@celo/utils/lib/address`
- `normalizeAddress` from `@celo/utils/lib/address`
- `normalizeAddressWith0x` from `@celo/utils/lib/address`
- `stringToBoolean` from `@celo/utils/lib/parsing`
- `parseInputAmount` from `@celo/utils/lib/parsing`
- `AddressType` from `@celo/utils/lib/io`
- `appendPath` from `@celo/utils/lib/string`
- `getErrorMessage` from `@celo/utils/lib/displayFormatting`
- `Encrypt` from `@celo/utils/lib/ecies` (only used in `KeychainWallet.test.ts`)

##### Address functions

- `privateKeyToAddress` from `@celo/utils/lib/address`
There is a direct equivalent `privateKeyToAddress` function in `viem/accounts` but previous implementation covered string argument both including and excluding `0x` while `viem`s alternative requires `0x` to always be present. Created abstraction to use `viem`s but ensuring the leading `0x`.

- `privateKeyToPublicKey` from `@celo/utils/lib/address`
There is a potential equivalent `privateKeyToAccount` function in `viem/accounts` which returns object with `publicKey` field but it doesn't match the results of existing function

- `isValidAddress` from `@celo/utils/lib/address`
There is a direct equivalent `isAddress` function in `viem` but it doesn't consider uppercase addresses valid.
Created abstraction that will pass lowercased input to `isAddress`.

- `isValidChecksumAddress` from `@celo/utils/lib/address`
This function was a re-export from `@ethereumjs/util`

##### Various functions

- `ValidatorKind` from `@celo/utils/lib/inputValidation`
Already present at `src/utils/inputValidation.ts` as a copy-paste from `@celo/phone-utils`

- `EIP712TypedData` from `@celo/utils/lib/sign-typed-data-utils`
There is a direct equivalent `EIP712Domain` type at `viem/zksync` but at places where the `EIP712TypedData` is used - it is dependant on other functions from `@celo/wallet-base` that accept only this type. Not migrating to `viem` equivalent for now.

- `sleep` from `@celo/utils/lib/async`
Replaced with existing function from `src/utils/sleep.ts`

- `verifySignature` from `@celo/utils/lib/signatureUtils`
Replaced with `viem` equivalent `verifyMessage` function

- `levenshteinDistance` from `@celo/utils/lib/levenshtein`
Replaced with equivalent from `fast-levenshtein`

##### New dependencies:
- `@ethereumjs/util` (used: `8.0.5`, latest: `9.1.0`)
- `@noble/ciphers ^0.6.0`
- `@noble/curves ^1.5.0`
- `@noble/hashes ^1.5.0`

### Test plan
Added existing tests from `@celo/utils` repo for the used functions

### Related issues

- Relates to RET-1187

### Backwards compatibility
Yes

### Network scalability


src/utils/io.ts
jeanregisser, line 1: nit: Could we add a `TODO` note for later to refactor this one to use `zod` instead of `io-ts`? 🙏

src/utils/displayFormatting.ts
```diff
@@ -0,0 +1,11 @@
+// Ref: https://github.com/celo-org/developer-tooling/blob/master/packages/sdk/base/src/displayFormatting.ts#L2
```
jeanregisser, line 1: nit: would be nice to use the immutable link with the sha, so the link doesn't become dead if the file is moved.

src/web3/KeychainWallet.test.ts
jeanregisser, line 1: Cool, I'll remove this one very soon so we'll be able to remove the `ecies` utils too!

src/web3/KeychainWallet.test.ts
```diff
@@ -348,7 +347,12 @@ describe('KeychainWallet', () => {
               const hexStr: string = mockAddress
               const signedMessage = await wallet.signPersonalMessage(knownAddress, hexStr)
               expect(signedMessage).not.toBeUndefined()
-              const valid = verifySignature(hexStr, signedMessage, knownAddress)
+              // const valid = verifyMessage(hexStr, signedMessage, knownAddress)
```
jeanregisser, line 350: nit:
```suggestion
```

src/web3/saga.test.ts
```diff
@@ -1,4 +1,4 @@
-import { isValidChecksumAddress } from '@celo/utils/lib/address'
+import { isValidChecksumAddress } from '@ethereumjs/util'
```
jeanregisser, line 1: can we replace it with viem's [isAddress](https://viem.sh/docs/migration-guide#utilities-isaddress--getaddress-perform-checksum-validation)?
It can do the checksum too I think.
This way we can avoid pulling `@ethereumjs/util` if I'm not mistaken.

src/utils/address.ts
```diff
@@ -1,9 +1,40 @@
-// Turns '0xce10ce10ce10ce10ce10ce10ce10ce10ce10ce10'
-// into ['ce10','ce10','ce10','ce10','ce10','ce10','ce10','ce10','ce10','ce10']
+/**
+ * Reference files:
+ * https://github.com/celo-org/developer-tooling/blob/master/packages/sdk/base/src/address.ts
+ * https://github.com/celo-org/developer-tooling/blob/master/packages/sdk/utils/src/address.ts
+ */
+
+import { privateToPublic, toChecksumAddress } from '@ethereumjs/util'
```
jeanregisser, line 7: Can we try using the viem equivalent functions?
[getAddress](https://viem.sh/docs/utilities/getAddress#getaddress) and maybe [privateKeyToAccount](https://viem.sh/docs/accounts/local/privateKeyToAccount#privatekeytoaccount)?

Fine to do as a separate PR.

To avoid pulling `@ethereumjs/util`.

jeanregisser (approved):
Awesome work! 🚀

sviderock:
@jeanregisser I'll address all the things in the follow-up PR!

PR #5879, opened 2024-09-02
chore: release automation actions

bakoushin:
### Description

Move release automation actions to the wallet repo.

The actions are moved with minimal changes:
1. removed checkout of another repo since this one is the repo to checkout
2. scripts are ported to TypeScript using ShellJS to match existing scripts in this repo

### Test plan

Manual action run.

### Related issues

- Fixes RET-1161

### Backwards compatibility

NA

### Network scalability

NA


.github/scripts/setWalletEnv.ts
```diff
@@ -0,0 +1,71 @@
...
+    const [key, value] = variable.split('=')
+    if (key && value) {
+      const cleanKey = key.trim()
+      const cleanValue = value
+        .trim()
+        .replace(/^['"]|['"]$/g, '') // Remove surrounding quotes
+        .trim()
+      process.env[cleanKey] = cleanValue
+    }
+  })
```
jeanregisser, line 60: nit: what do you think of using using dotenv instead? From this repo it's easier to use since we already have package.json.
We'd need to move the step that uses this script after `yarn install` but I think this is ok.

PR #5880, opened 2024-09-02
chore: Remove comment encryption, DEK

jophish:
### Description

For [RET-1184](https://linear.app/valora/issue/RET-1184/[contractkit]-remove-comment-encryption-flows). Removes comment encryption and any reference of the DEK from the app.

### Test plan

Unit tests updated, manual testing.

### Related issues

- Fixes [RET-1184](https://linear.app/valora/issue/RET-1184/[contractkit]-remove-comment-encryption-flows).

### Network scalability


src/verify/hooks.ts
```diff
@@ -113,7 +106,6 @@ export function useVerifyPhoneNumber(phoneNumber: string, countryCallingCode: st
           clientPlatform: Platform.OS,
           clientVersion: DeviceInfo.getVersion(),
           clientBundleId: DeviceInfo.getBundleId(),
-          publicDataEncryptionKey: compressedPubKey(hexToBuffer(privateDataEncryptionKey)),
```
jophish, line 116: Note that this change requires us to update the `verifyPhoneNumber` endpoint in `identity-service`, since the public DEK is currently a required parameter. AFAICT, it's not used for anything there, we just log it in the DB as an extra detail, so I think this is fine.

src/transactions/transferFeedUtils.ts
```diff
@@ -111,7 +97,10 @@ export function useTransferFeedDetails(transfer: TokenTransfer, isJumpstart: boo
         subtitle = t('feedItemJumpstartSentSubtitle')
       } else {
         title = t('feedItemSentTitle', { displayName })
-        subtitle = t('feedItemSentInfo', { context: !comment ? 'noComment' : null, comment })
+        subtitle = t('feedItemSentInfo', {
+          context: !subtitleContent ? 'noComment' : null,
```
jophish, line 101: This is the only lingering reference to "comments" ever existing in the codebase, AFAICT. I kept it in since i18n uses it as context to suffix the i18n keys; wasn't sure if we could safely change the key names.

src/verify/hooks.ts
jophish: See PR on `identity-service` [here](https://github.com/valora-inc/identity-service/pull/556/files).

src/transactions/transferFeedUtils.ts
jeanregisser: ah yes, we could rename it to `feedItemSentInfoV1_93` (but we still need to keep the old key name for backward compat.

src/verify/hooks.ts
jeanregisser: Thanks!

src/viem/prepareTransactions.test.ts
```diff
@@ -1153,112 +1150,6 @@ describe('prepareTransactions module', () => {
...
-          estimatedGasFeeInDecimal: new BigNumber(0.05),
-          decreasedSpendAmount: new BigNumber(4),
-        })
-      ).toStrictEqual({
-        feeCurrency: mockCeloTokenBalance,
-        maxFeeAmount: new BigNumber(0.1),
-        estimatedFeeAmount: new BigNumber(0.05),
-      })
-    })
-  })
```
jeanregisser, line 1260: Looks like these tests should have been kept.

jeanregisser (approved):
Epic work! 💥 🚀 

This removes so much code 🙌 

Already approved, with a minor comment about some tests that I think were deleted by mistake.

src/app/saga.ts
```diff
@@ -430,97 +415,6 @@ export function* handleSetAppState(action: SetAppState) {
...
-
-    Logger.debug(
-      `${TAG}@runCentralPhoneVerificationMigration`,
-      'Starting to run central phone verification migration'
-    )
-
-    const phoneHash = yield* call(getPhoneHash, phoneNumber, cachedSalt)
-    const inviterAddress = yield* select(inviterAddressSelector)
-
-    const response = yield* call(fetch, networkConfig.migratePhoneVerificationUrl, {
```
jeanregisser, line 488: This mean we can remove the `migrateASv1Verification` endpoint in `identity-service`.

src/viem/prepareTransactions.test.ts
jophish: good catch!

PR #5886, opened 2024-09-03
feat(earn): Multi-pool earn analytics once over

finnian0826:
### Description

Update to make properties consistent, check that there is test coverage for all events.

Note: No test coverage for `earn_enter_amount_info_more_pools`, but that is for the single pool flow only.

### Test plan

Unit tests updated

### Related issues

- Fixes ACT-1351

### Backwards compatibility

Yes

### Network scalability


src/analytics/Properties.tsx
```diff
@@ -1641,19 +1641,28 @@ interface EarnEventsProperties {
   [EarnEvents.earn_active_pools_card_press]: undefined
   [EarnEvents.earn_home_learn_more_press]: undefined
   [EarnEvents.earn_pool_card_press]: {
-    tokenAmount: string
+    poolAmount: string
     poolId: string
```
satish-ravi, line 1645: can we add poolId to EarnCommonProperties and enforce it for every pool specific event?

jh2oman (approved):

